PR #74, opened 2022-05-20
Implement eth_blockNumber

### LINKED ISSUE #28: Implement eth_blockNumber

ar-conmit:
Signed-off-by: Anton Rusev <anton.rusev.code@gmail.com>

**Description**:

- Retrieve latest block number from the mirror node service
- Update `MIRROR_NODE_URL` in `test.env` to use `http`

**Related issue(s)**:

Fixes #28 

**Notes for reviewer**:
- I have added `http` scheme to `test.env` file in order to connect properly to the local mirror node service, otherwise if its omitted, mirrorNodeClient instance uses `https` by default which points to invalid local mirror node URL
- With the above change, the test case `'baseUrl is exposed and correct'` in `mirrorNodeClient.spec.ts` was failing because it expects the production mirror node url to contains `https` scheme. So I removed the protocol scheme from the url before MirrorNodeClient is instantiated in order to test the same functionality.

**Checklist**

- [] Documented (Code comments, README, etc.)

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -148,18 +148,13 @@ export class EthImpl implements Eth {
   async blockNumber(): Promise<number> {
     this.logger.trace('blockNumber()');
 
-    try {
-      const { blocks } = await this.mirrorNodeClient.getLatestBlock();
-
-      if (Array.isArray(blocks) && blocks.length > 0) {
-        return blocks[0].number;
-      }
-
-      throw new Error('No blocks were found');
-    } catch (error) {
-      this.logger.error(error, 'Error raised during blockNumber');
-      throw(error);
+    const blocksResponse = await this.mirrorNodeClient.getLatestBlock();
+    const blocks = blocksResponse !== null ? blocksResponse.blocks : null;
+    if (Array.isArray(blocks) && blocks.length > 0) {
+      return blocks[0].number;
     }
+
+    throw new Error('Error encountered retrieving latest block');
   }
 
   /**
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -145,9 +145,17 @@ export class EthImpl implements Eth {
   /**
    * Gets the most recent block number.
    */
-  async blockNumber() {
+  async blockNumber(): Promise<number> {
     this.logger.trace('blockNumber()');
-    return await this.mirrorNode.getMostRecentBlockNumber();
+
+    try {
+      const latestBlock = await this.mirrorNodeClient.getLatestBlock();
+
+      return latestBlock.blocks[0].number;
+    } catch (error) {
+      this.logger.error(error, 'Error raised during blockNumber');
+      throw(error);
+    }
   }
 
   /**
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -149,9 +149,13 @@ export class EthImpl implements Eth {
     this.logger.trace('blockNumber()');
 
     try {
-      const latestBlock = await this.mirrorNodeClient.getLatestBlock();
+      const { blocks } = await this.mirrorNodeClient.getLatestBlock();
 
-      return latestBlock.blocks[0].number;
+      if (Array.isArray(blocks) && blocks.length > 0) {
+        return blocks[0].number;
+      }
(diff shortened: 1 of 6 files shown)


mmrt (approved):

mmrt:
your blocknumber test is failing, but other than that it looks good

ar-conmit:
The blockNumber test is failing because it tries to hit mirror node service on http://localhost:5551 which is not available in the CI's environment.

packages/server/tests/server.spec.ts
```diff
@@ -302,6 +302,18 @@ describe('RPC Server', async function() {
 
     BaseTest.unsupportedJsonRpcMethodChecks(res);
   });
+
+  it('should execute "eth_blockNumber"', async function() {
+    const res = await this.testClient.post('/', {
```
Nana-EC, line 307: Please mock the `AxiosInstance (this.testClient)` this will allow you to provide a valid response from the MirrorNode
I provided examples in `mirrorNodeClient.ts`

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -97,6 +97,11 @@ describe('Eth', async function () {
     expect(result).to.eq(false);
   });
 
+  it('should execute "eth_blockNumber"', async function () {
```
Nana-EC, line 100: Add a test case for the failure scenario

Nana-EC (changes requested):
Looking good.
Should be able to mock the MirrorNode and get the test passing.

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -97,6 +97,11 @@ describe('Eth', async function () {
...
+    const blockNumber = await Relay.eth().blockNumber();
```
Nana-EC, line 101: You should actually mock the mirrorNode response for this case.
That's why it's failing.
Please see `mirrorNodeClient.spec.ts` for examples

packages/server/tests/server.spec.ts
Nana-EC: Actually is this even passing?
The mirrorNode call should be failing.
If you add a UT test in the `eth.spec.ts` I think you should be fine to leave this out for now since the response is dynamic.
Once we have integration tests we can have this covered here

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -42,6 +46,55 @@ const validateHash = (hash: string, len?: number) => {
...
+    mock.onGet('blocks?limit=1&order=desc').reply(200, {
+      blocks: [defaultBlock]
+    });
+    const blockNumber = await ethImpl.blockNumber();
+    expect(blockNumber).to.be.eq(blockNumber);
+  });
+
+  it('"eth_blockNumber" should throw an error if no blocks are found', async function () {
+    mock.onGet('blocks?limit=1&order=desc').reply(200, {
+      blocks: []
```
Nana-EC, line 89: See [API Spec for blocks](https://previewnet.mirrornode.hedera.com/api/v1/docs/#/blocks/listBlocks) a missing block will have a status error message

Something like
```
{
  "_status": {
    "messages": [
      {
        "message": "Block not found"
      }
    ]
  }
}
```

Nana-EC (changes requested):
Almost there.
Error vs null response and a test response that needs updating

packages/relay/src/lib/eth.ts
```diff
@@ -145,9 +145,21 @@ export class EthImpl implements Eth {
...
-    return await this.mirrorNode.getMostRecentBlockNumber();
+
+    try {
+      const { blocks } = await this.mirrorNodeClient.getLatestBlock();
+
+      if (Array.isArray(blocks) && blocks.length > 0) {
+        return blocks[0].number;
+      }
+
+      throw new Error('No blocks were found');
```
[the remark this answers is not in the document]
shemnon: If we don't have a most recent block number the chain isn't functioning in a useful way.  I support throwing an error, which will result in an RPC error.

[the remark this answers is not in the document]
Nana-EC: Okay. Good confirmation. Will have to recheck a couple of the other places on error vs null

Nana-EC (approved):
LGTM

PR #82, opened 2022-05-21
Chain ID and FeeHistory Tweaks

shemnon:
* Support reward percentages for FeeHistory
* re-jigger chain_id detection, drive from network if not hard wired
* prettier reformatting

PR #80, opened 2022-05-20
Add support for getBlockBy, getBlockTransactionCountBy and getTransactionByBlock endpoints

Nana-EC:
Signed-off-by: Nana-EC <nana@swirldslabs.com>

**Description**:
Eth developers should be able to retrieve block information via `eth_getBlockByHash` or `eth_getBlockByNumber` from teh relay.
Under the hood this will call a series of Mirror Node APIs as defined in [HIP 482](https://hips.hedera.com/hip/hip-482)

- Simplify `mirroNodeClient. getContractResults ` with interface params
- Rename `ConsensusNodeClient` to `NodeClient` since SDK includes MirrorNode GRPC calls 
- Move node client calls through SDK into `NodeClient` to cleanliness and reuse
- Update `eth.getBlockByHash` and `eth.getBlockByNumber` to attempt real mirror node and as backup pull from cache until transaction hash details are added by MirrorNode
- Add `getTransactionByBlockHashAndIndex` and `getTransactionByBlockNumberAndIndex` implementations
- Update `eth.ts` to utilize `NodeClient` instead of `Client` directly.
- Update `eth_getBlockByNumber` signature in server to accept extra showDetails param

**Related issue(s)**:

Fixes #29
Fixes #30 
Fixes #31 
Fixes #32 
Fixes #53 
Fixes #54 

**Notes for reviewer**:
- Need to try against local node

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #83, opened 2022-05-21
eth_getTransactionReceiptByHash

### LINKED ISSUE #18: Implement eth_getTransactionByHash

shemnon:
**Description**:
Massages the mirror node result to get a transaction receipt

**Related issue(s)**:

Fixes #18 
builds on #80 #82 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -119,16 +132,28 @@ export class EthImpl implements Eth {
...
+        exchangeRates.currentRate.hbars *
+        10_000_000
+    );
   }
 
   /**
    * Gets the most recent block number.
    */
-  async blockNumber() {
+  async blockNumber(): Promise<number> {
```
Nana-EC, line 145:
```suggestion
  async blockNumber(): Promise<number | null> {
```
Will have to update it in `relay/src/index.ts` also

packages/relay/src/lib/eth.ts
```diff
@@ -119,16 +132,28 @@ export class EthImpl implements Eth {
...
     this.logger.trace('blockNumber()');
-    return await this.mirrorNode.getMostRecentBlockNumber();
+    const blockResponse = await this.mirrorNodeClient.getLatestBlock();
```
Nana-EC, line 147: Here's another way that also handles errors.
IF the mirror node doesn't find it a valid 404 with a valid response will be returned so I removed the block undefined check
```ts
return this.mirrorNodeClient.getLatestBlock()
      .then(block => {
        if (block.number === undefined) {
          this.logger.debug('Failed to retrieve latest block');
          // block not found
          return null;
        }

        this.logger.trace('blockNumber=' + block.number);
        return block.number;
      })
      .catch((e: any) => {
        this.logger.error(e, 'Error encountered retrieving latest block');
        return null;
      });
```

packages/relay/src/lib/eth.ts
```diff
@@ -568,34 +674,43 @@ export class EthImpl implements Eth {
...
-    } catch (e) {
-      this.logger.error(
-        e,
-        'Failed to handle getTransactionReceipt cleanly for hash %s',
-        hash
-      );
+    this.logger.trace(`getTransactionReceipt(${hash})`);
+    const receiptResponse = await this.mirrorNodeClient.getContractResult(hash);
+    this.logger.trace(`response - ${JSON.stringify(receiptResponse)}`);
+    if (receiptResponse === undefined) {
```
Nana-EC, line 680: This will never be true.
At worst it's null or receiptResponse.hash is undefined.

```suggestion
    if (receiptResponse === null || receiptResponse.hash === undefined) {
```

Likely causing your test failure

shemnon: done

packages/relay/src/lib/eth.ts
```diff
@@ -568,34 +569,35 @@ export class EthImpl implements Eth {
...
+        from: receiptResponse.from,
+        cumulativeGasUsed: EthImpl.prepend0x(receiptResponse.block_gas_used.toString(16)),
+        gasUsed: EthImpl.prepend0x(receiptResponse.gas_used.toString(16)),
+        logs: receiptResponse.logs,
+        logsBloom: receiptResponse.logs_bloom,
+        transactionHash: receiptResponse.hash,
+        transactionIndex: EthImpl.prepend0x(receiptResponse.transaction_index.toString(16)),
+        effectiveGasPrice: EthImpl.prepend0x((Number.parseInt(effectiveGas) * 10_000_000_000).toString(16)),
+        contractAddress: receiptResponse.to,
+        status: receiptResponse.status,
```
Nana-EC, line 597: I don't think such a value exist on the response. This is supposed to be 0 or 1 right?

packages/relay/src/lib/eth.ts
shemnon: done

packages/relay/src/lib/eth.ts
shemnon: It exists on the mirror node response.  0x0 or 0x1 as appropriate

Nana-EC: Ah you're right.
Open API Spec wasn't ordered so i forgot it was in there

Nana-EC (approved):
LGTM

Nana-EC (approved):

PR #84, opened 2022-05-23
Add support for eth_getTransactionByHash with mocked data

Ivo-Yankov:
Signed-off-by: Ivo Yankov <ivo@devlabs.bg>

**Description**:
Adds support for `eth_getTransactionByHash`. Tests are implemented with mocked data.

Fixes #
https://github.com/hashgraph/hedera-json-rpc-relay/issues/18

**Notes for reviewer**:

**Checklist**


packages/relay/tests/lib/eth.spec.ts
```diff
@@ -258,4 +352,40 @@ describe('Eth', async function () {
       expect(receipt).to.deep.eq(cachedReceipt);
     });
   });
+
+  describe('eth_getTransactionByHash', async function () {
+    it('returns `null` for non-existing hash', async function () {
+      // mirror node request mocks
+      mock.onGet(`contracts/results/${defaultTxHash}`).reply(200, defaultDetailedContractResultByHash);
```
Nana-EC, line 359: Based on your description this should match a 404 with a status error message returned from mirror node
```suggestion
      mock.onGet(`contracts/results/${defaultTxHash}`).reply(404, {
      '_status': {
        'messages': [
          {
            'message': 'No such detailed contract result exists'
          }
        ]
      }
      });
```

packages/server/tests/server.spec.ts
```diff
@@ -74,6 +74,19 @@ describe('RPC Server', async function() {
     expect(res.data.result).to.be.equal('relay/' + process.env.npm_package_version);
   });
 
+  it('should execute "eth_getTransactionByHash"', async function() {
```
Nana-EC, line 77:
```suggestion
  it('should execute "eth_getTransactionByHash missing transaction"', async function() {
```

packages/relay/src/lib/mirrorNode.ts
```diff
@@ -134,4 +134,62 @@ export class MirrorNode {
     const block = this.store.get(hash);
     return block === undefined ? null : block;
   }
+
+  public async getContractResult(hash: string): Promise<any> {
```
Nana-EC: I don't believe this is called anywhere, and it should be as it should be phased out by the actual implementation you've provided in `mirrorNodeClient`

packages/relay/src/lib/eth.ts
```diff
@@ -447,6 +447,42 @@ export class EthImpl implements Eth {
...
   }
 
+  /**
+   * Gets a transaction by the provided hash
+   *
+   * @param hash
+   */
+  async getTransactionByHash(hash: string) {
+    this.logger.trace('getTransactionByHash(hash=%s)', hash);
+    const contractResult = await this.mirrorNodeClient.getContractResult(hash);
```
Nana-EC, line 457: Please add a check for where the fields are not defined because mirror node returned a 404 or other error.
You can add the following check and then the below if  check and the else contents won't be needed
```js
    if (contractResult === null || contractResults.hash === undefined) {
      // contract result not found
      return null;
    }
```
or
```js
    if (contractResult === null || contractResults._status) {
      // contract result not found
      return null;
    }
```

packages/relay/src/lib/eth.ts
```diff
@@ -447,6 +447,42 @@ export class EthImpl implements Eth {
...
+    if (contractResult) {
+      return new Transaction({
+        accessList: contractResult.access_list,
+        blockHash: contractResult.block_hash,
```
Nana-EC, line 461:
```suggestion
        blockHash: contractResult.block_hash.substring(0, 66),
```

packages/relay/src/lib/eth.ts
```diff
@@ -447,6 +447,42 @@ export class EthImpl implements Eth {
...
+        blockNumber: contractResult.block_number,
+        chainId: contractResult.chain_id,
+        from: contractResult.from,
```
Nana-EC, line 464:
```suggestion
        from: contractResult.from.substring(0, 42),
```

packages/relay/src/lib/eth.ts
```diff
@@ -447,6 +447,42 @@ export class EthImpl implements Eth {
...
+        gas: contractResult.gas_used,
+        gasPrice: contractResult.gas_price,
+        hash: contractResult.hash,
```
Nana-EC, line 467:
```suggestion
        hash: contractResult.hash.substring(0, 66),
```

packages/relay/src/lib/eth.ts
```diff
@@ -447,6 +447,42 @@ export class EthImpl implements Eth {
...
+        input: contractResult.function_parameters,
+        maxPriorityFeePerGas: contractResult.max_priority_fee_per_gas,
+        maxFeePerGas: contractResult.max_fee_per_gas,
+        nonce: contractResult.nonce,
+        r: contractResult.r,
```
Nana-EC, line 472:
```suggestion
        r: contractResult.r.substring(0, 66),
```

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -258,4 +352,40 @@ describe('Eth', async function () {
       expect(receipt).to.deep.eq(cachedReceipt);
     });
   });
+
+  describe('eth_getTransactionByHash', async function () {
+    it('returns `null` for non-existing hash', async function () {
```
Nana-EC, line 357: nit: mirror Node won't actually return null. It'll be a valid response with a message.
See [API Spec for getContractResultByTransactionIdOrHash](https://previewnet.mirrornode.hedera.com/api/v1/docs/#/contracts/getContractResultByTransactionIdOrHash)

packages/relay/src/lib/eth.ts
```diff
@@ -447,6 +447,41 @@ export class EthImpl implements Eth {
...
+    if (contractResult === null || contractResult.hash === undefined) {
+      return null;
+    }
+
+    return new Transaction({
+      accessList: contractResult.access_list,
+      blockHash: contractResult.block_hash.substring(0, 66),
+      blockNumber: contractResult.block_number,
+      chainId: contractResult.chain_id,
+      from: contractResult.from.substring(0, 42),
```
shemnon, line 467: Is the substring defensive or do we get longer strings from mirror node?  For the hashes we do get 384 bit hashes.

packages/relay/src/lib/eth.ts
```diff
@@ -447,6 +447,41 @@ export class EthImpl implements Eth {
...
+      gas: contractResult.gas_used,
+      gasPrice: contractResult.gas_price,
+      hash: contractResult.hash.substring(0, 66),
+      input: contractResult.function_parameters,
+      maxPriorityFeePerGas: contractResult.max_priority_fee_per_gas,
+      maxFeePerGas: contractResult.max_fee_per_gas,
+      nonce: contractResult.nonce,
+      r: contractResult.r.substring(0, 66),
+      s: contractResult.s.substring(0, 66),
+      to: contractResult.to.substring(0, 42),
```
shemnon, line 477: Same question as from:, is this just defensive?

packages/relay/src/lib/model.ts
```diff
@@ -122,4 +122,48 @@ export class Receipt {
...
+    public readonly nonce!: string;
+    public readonly r!: string;
+    public readonly s!: string;
+    public readonly to!: string | null;
+    public readonly transactionIndex!: string | null;
+    public readonly type!: string;
+    public readonly v!: string;
+    public readonly value!: string;
+
+    constructor(args: any) {
```
shemnon, line 148: For the functions that sonar complains about too-many-params, should we adopt this pattern? @Nana-EC @mmrt

shemnon (approved):

packages/relay/src/lib/eth.ts
Nana-EC: This was my suggestion and just defensive. Actually checked MirrorNode response and min size is 40 and max is 42.
I can fix this in my following PR's

packages/relay/src/lib/eth.ts
Nana-EC: This was my suggestion and just defensive. Actually checked MirrorNode response and min size is 40 and max is 42.
I can fix this in my following PR's

packages/relay/src/lib/model.ts
Nana-EC: Yes. In my PR I created a few interfaces so we're passing expandable objects instead of individual params.
Also had this class in my PR but Ivo gets to get it in first :)

PR #93, opened 2022-05-25
Add prometheus support and health endpoints

### LINKED ISSUE #87: Add liveness and readiness endpoint to relay
### Problem

When deployed in a k8s instance a liveness and readiness endpoint is necessary for correct deployment coordination.
No such endpoints exist right now

### Solution

Add a `/liveness` and `/readiness` endpoint
- `/health/liveness` - return HTTP 200 post startup
- `/health/readiness` - return HTTP 200 post simple check e.g. valid chainId is set

### Alternatives

_No response_

### LINKED ISSUE #88: Add Prometheus monitoring support
### Problem

The relay currently doesn't have monitoring support embedded into the API.
This will make it challenging to assess and view health at scale.

### Solution

Utilize [Prometheus](https://prometheus.io/docs/introduction/overview/) monitoring support to allow for great log value tracing capabilities. 

- Add a `/metrics` endpoint for aggregations to be polled from
- Add npm [prom-client](https://github.com/siimon/prom-client) to allow logs aggregation
- Capture request latency
- Capture log levels
- Capture response status

### Alternatives

A non prometheus option e.g. swagger

Nana-EC:

**Description**:
Relay currently lacks any monitoring support.

- Add a `/metrics` endpoint to support prometheus metrics retrieval
- Add simple request latency measurements
- Add initial method and success counters
- Add health endpoint support
  - `health/readiness` - ensures valid chainId can be retrieved
  - `/health/liveness` - returns success when up

**Related issue(s)**:

Fixes #87 
Fixes #88 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/src/server.ts
+++ b/packages/server/src/server.ts
@@ -18,9 +18,10 @@
  *
  */
 
+import { Relay, RelayImpl } from '@hashgraph/json-rpc-relay';
 import Koa from 'koa';
 import koaJsonRpc from 'koa-jsonrpc';
-import { Relay, RelayImpl } from '@hashgraph/json-rpc-relay';
+import { collectDefaultMetrics, Counter, Registry } from 'prom-client';
 
 import pino from 'pino';
 const mainLogger = pino({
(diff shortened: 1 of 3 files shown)


packages/server/src/server.ts
```diff
@@ -104,6 +175,8 @@ rpc.use('eth_getCode', async (params: any) => {
  * returns: Chain ID - integer
  */
 rpc.use('eth_chainId', async () => {
+  methodCounter.inc();
```
shemnon, line 178: We don't want this. Front ends can aggregate across tags.

shemnon (changes requested):
Generaly speaking we only want
```
    methodCounter.inc({ method: '<method>' });
```
and not the un-tagged increment.  UI tools will aggregate across the tags.

packages/server/src/server.ts
Nana-EC: Good point. Removed

shemnon (approved):
good enough to merge, although in this or a later PR we should instrument _every_ rpc to register a metric.

PR #98, opened 2022-05-26
Fix blocks transactions and r and s signatures assignments

### LINKED ISSUE #97: Fix getContractResultsByAddressAndTimestamp error on null r and s sigs
### Description

For HBAR transfers a simple CrytoTransfer is submitted. The contract results details will as such not have r and s signatures.
This was missing and an expection is thrown when we attempt to format the value

### Steps to reproduce

Issue an HBAR transfer
Relay will report 
```
Failed to retrieve contract result details for contract address ....
```

### Additional context

_No response_

### Hedera network

previewnet

### Version

v0.1.0

### Operating system

_No response_

Nana-EC:
**Description**:
Previewnet testing highlighted scenarios where the r and s signatures were not set, as well as the fact that transactions in blocks weren't getting set.

- Guard against unset r and s signatures
- Replace async foreach with simple for and async to ensure transactions are set
- Split block caching logic out from Block Model

**Related issue(s)**:

Fixes #97 

**Notes for reviewer**:
The Block caching logic should see an update that queries the mirror node based on transaction hash.
The exact response for a duplicate transaction needs to be considered.

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -22,7 +22,7 @@ import { Eth } from '../index';
 import { Status } from '@hashgraph/sdk';
 import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
 import { Logger } from 'pino';
-import { Block, Transaction } from './model';
+import { Block, CachedBlock, Transaction } from './model';
 import { MirrorNode } from './mirrorNode';
 import { MirrorNodeClient, SDKClient } from './clients';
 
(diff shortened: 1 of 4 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -668,9 +670,10 @@ export class EthImpl implements Eth {
     let gasUsed = 0;
     let maxGasLimit = 0;
     let timestamp = 0;
-    const transactions: Transaction[] = [];
+    const transactionObjects: Transaction[] = [];
     const transactionHashes: string[] = [];
-    contractResults.results.forEach(async (result) => {
+
+    for (const result of contractResults.results) {
```
shemnon, line 676: Hmm... this may collide with #100.  But we are in essence doing the same thing.

shemnon (approved):

PR #100, opened 2022-05-26
Bring getBlockBy... into execution-api schema compliance

### LINKED ISSUE #99: Block returned from eth_getBlockBy... does not conform to schema
The blocks returned from `eth_getBlockByNumber` and all realted APIs do not conform to the https://github.com/ethereum/execution-apis schemas.  Specifically...

* Always use hex-string numbers instead of JSON numbers
* logsBloom must have a value of 512 hex digits
* all hashes must have full-length values
* baseFeePerGas must be present, even if zero.

shemnon:
Bring the getBlockBy... apis into compliance
* always use hex string instead of numbers
* use full hashes
* supply required fields

**Description**:

**Related issue(s)**:

Fixes #99 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -19,7 +19,7 @@
  */
 
 import { Eth } from '../index';
-import { Status } from '@hashgraph/sdk';
+import { ContractId, Status } from '@hashgraph/sdk';
 import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
 import { Logger } from 'pino';
 import { Block, Transaction } from './model';
@@ -39,6 +39,8 @@ export class EthImpl implements Eth {
   static emptyHex = '0x';
   static zeroHex = '0x0';
   static emptyArrayHex = '0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347';
+  static zeroAddressHex = '0x0000000000000000000000000000000000000000';
+  static emptyBloom = "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000";
   static defaultGas = 0x3d0900;
 
   /**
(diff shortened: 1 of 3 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -526,8 +528,11 @@ export class EthImpl implements Eth {
       return null;
     }
 
+    const maxPriorityFee = contractResult.max_priority_fee_per_gas === EthImpl.emptyHex ? undefined : contractResult.max_priority_fee_per_gas;
+    const maxFee = contractResult.max_fee_per_gas === EthImpl.emptyHex ? undefined : contractResult.max_fee_per_gas;
+    
     return new Transaction({
-      accessList: contractResult.access_list,
+      accessList: undefined, // we don't support access lists, so punt for now
```
Nana-EC, line 535: Q: Why set `undefined` vs `null`?

packages/relay/src/lib/eth.ts
```diff
@@ -566,14 +571,18 @@ export class EthImpl implements Eth {
         receiptResponse.max_fee_per_gas === undefined || receiptResponse.max_fee_per_gas == '0x'
           ? receiptResponse.gas_price
           : receiptResponse.max_fee_per_gas;
+      const createdContract =
+        receiptResponse.created_contract_ids.length > 0
+          ? EthImpl.prepend0x(ContractId.fromString(receiptResponse.created_contract_ids[0]).toSolidityAddress())
```
Nana-EC, line 576: Is it a safe assumption for pull from index 0?
What if the createdContract isn't in that index?

You should be able to use `receiptResponse.contract_id` instead

packages/relay/src/lib/eth.ts
```diff
@@ -670,7 +679,8 @@ export class EthImpl implements Eth {
     let timestamp = 0;
     const transactions: Transaction[] = [];
     const transactionHashes: string[] = [];
-    contractResults.results.forEach(async (result) => {
+    for (const i in contractResults.results) {
```
Nana-EC, line 682: Was gonna suggest the below but rebasing should bring this in for you
```suggestion
    for (const result of contractResults.results) {
```

Nana-EC (commented):
Looks good `undefined` question and `contract_id` suggestion

packages/relay/src/lib/eth.ts
shemnon: Schema doesn't support null values.  Undefined keeps it out of the serialized form and makes it clear we need to fix it when we support access  lists.

packages/relay/src/lib/eth.ts
shemnon: There is a deeper issue with detecting "contract create" transacitons with mirror nodes I need to open a bug for.  It's the best we can do in the interim, we can't distinguish null 'to' addresses because mirror node populates it with the created contract.

packages/relay/src/lib/eth.ts
shemnon: ahh... `of` (values) vs `in` (indexes).  I learned new JS language today.

packages/relay/src/lib/eth.ts
Nana-EC: Okay, wanted to make sure the omission was intended behaviour.

packages/relay/src/lib/eth.ts
Nana-EC: Okay.
I was concerned about cases where there's multiple contacts created. 0 might not be the primary contract.
Fine for now I guess

Nana-EC (approved):
LG

Issue #20, opened 2022-04-15
Implement eth_getLogs 


Kalina-Todorova:
Implement eth_getLogs (limited topics passed)

PR #103, opened 2022-05-26
eth_getLogs

Ivo-Yankov:
**Description**:

**Related issue(s)**:

Fixes #
https://github.com/hashgraph/hedera-json-rpc-relay/issues/20

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -759,11 +759,12 @@ export class EthImpl implements Eth {
       }
     }
     else if (filters.fromBlock && filters.toBlock) {
-      const blocks = await this.mirrorNodeClient.getBlocks([
+      const blocksResult = await this.mirrorNodeClient.getBlocks([
           `gte:${filters.fromBlock}`,
           `lte:${filters.toBlock}`
       ]);
 
+      const blocks = blocksResult?.blocks;
       if (blocks?.length) {
         const firstBlock = blocks[0];
         const lastBlock = blocks[blocks.length - 1];
@@ -780,14 +781,18 @@ export class EthImpl implements Eth {
       }
     }
 
-    let logs;
+    let result;
     if (filters.address) {
-      // /api/v1/contracts/{address}/results/logs
-      logs = await this.mirrorNodeClient.getContractResultsLogsByAddress(filters.address, params);
+      result = await this.mirrorNodeClient.getContractResultsLogsByAddress(filters.address, params);
     }
     else {
-      logs = await this.mirrorNodeClient.getContractResultsLogs(params);
+      result = await this.mirrorNodeClient.getContractResultsLogs(params);
     }
+
+    if (!result || !result.logs) {
+      return [];
+    }
+    const logs = result.logs;
 
     // Find all unique contractId and timestamp pairs and for each one make mirror node request
     const promises: Promise<any>[] = [];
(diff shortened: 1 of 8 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -825,4 +825,96 @@ export class EthImpl implements Eth {
...
+
+    if (!result || !result.logs) {
+      return [];
+    }
+    const logs = result.logs;
+
+    // Find all unique contractId and timestamp pairs and for each one make mirror node request
+    const promises: Promise<any>[] = [];
+    const uniquePairs: string[] = [];
+    logs.forEach(log => {
```
Nana-EC, line 879: In this solution you iterate over logs 3 times.
Would prefer a solution that does it once. Also you us a list instead of a object/dict to verify presence.

Notably `forEach` doesn't work well with async so just use a simple for loop
On every `logs` entity, check for the absence of the unique key in `uniquePairs` dictionary.
If present skip, if absent then `await` on the `async` call to get `this.mirrorNodeClient.getContractResultsDetails()`.
Then using `logs` element and `contractDetails` build a new `model/Log` and add it to a list.

At the end you should have a list of logs to return.

packages/relay/src/lib/eth.ts
```diff
@@ -825,4 +825,96 @@ export class EthImpl implements Eth {
...
+    }
+
+    if (!result || !result.logs) {
+      return [];
+    }
+    const logs = result.logs;
+
+    // Find all unique contractId and timestamp pairs and for each one make mirror node request
+    const promises: Promise<any>[] = [];
+    const uniquePairs: string[] = [];
```
Nana-EC, line 878: Not sure how large logs count could be so better to use a dictionary lookup instead
```suggestion
    const uniquePairs = {};
```
Then later on check for `uniquePairs[pair] === undefined` for presence. 
On absence you can just set it equal to true or some valid but small default value

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -673,6 +753,146 @@ describe('Eth calls using MirrorNode', async function () {
...
+      expect(res.data).to.eq(defaultLogs.logs[3].data);
+      expect(res.logIndex).to.eq(defaultLogs.logs[3].index);
+      expect(res.removed).to.eq(false);
+      expect(res.topics).to.exist;
+      expect(res.topics).to.deep.eq(defaultLogs.logs[3].topics);
+      expect(res.transactionHash).to.eq(defaultDetailedContractResults3.hash);
+      expect(res.transactionIndex).to.eq(defaultDetailedContractResults3.transaction_index);
+    };
+
+    it('no filters', async function () {
```
Nana-EC, line 811: Please add a no match case and a case where mirrorNode response is 400 or 404 to ensure `eth.getLogs` handles the response well

Nana-EC (changes requested):
Looking good.
Some improvement suggestion of result build and request for 2 minor non happy path tests

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -673,6 +755,164 @@ describe('Eth calls using MirrorNode', async function () {
     const result = await ethImpl.getTransactionByBlockNumberAndIndex(defaultBlock.number.toString(), defaultBlock.count);
     expect(result).to.equal(null);
   });
+
+  describe('eth_getLogs', async function () {
+
+    const expectLogData1 = (res) => {
```
Nana-EC, line 761: nit:  A few shared checks in here can be pulled out. 
```ts
const expectLogData = (res) => {
      expect(res.blockHash).to.eq(defaultDetailedContractResults.block_hash);
      expect(res.blockNumber).to.eq(defaultDetailedContractResults.block_number);
      expect(res.removed).to.eq(false);
      expect(res.topics).to.exist;
      expect(res.transactionHash).to.eq(defaultDetailedContractResults.hash);
      expect(res.transactionIndex).to.eq(defaultDetailedContractResults.transaction_index);
    };
```

Then each `expectLogDataX(res)` can call `expectLogData(res)`

Nana-EC (approved):
LG.
One nit.

Daniel-K-Ivanov (approved):

PR #106, opened 2022-05-27
fix issue returning transaction hash when submission passed precheck

### LINKED ISSUE #75: Return Ethereum Hash for sendRawTransaction that do not fail precheck
### Problem

Metamask is not properly incrementing the nonce value because an error is being thrown inside of the relay in certain scenarios when sendRawTransaction fails.

### Solution

Calculate the Ethereum Hash from the transaction bytes and return it unless the transaction failed precheck.

mmrt:
closes #75 

Returns computed ethereum hash unless transaction submission fails precheck. This results in proper nonce tracking in metamask for transactions that fail for other reasons (revert, etc)

Signed-off-by: Mitchell Martin <mitch@swirldslabs.com>


packages/relay/src/lib/eth.ts
```diff
@@ -481,36 +483,49 @@ export class EthImpl implements Eth {
...
+        }
+        const computedHash = EthImpl.prepend0x(createHash('keccak256').update(transactionBuffer).digest('hex'));
+
+		if (computedHash != txHash) {
+			this.logger.error("EthereumHash didn't match computed hash");
+		}
+        return txHash;
+      } catch (e) {
+        this.logger.error(e,
+            'Failed to handle sendRawTransaction cleanly for transaction %s, returning computed hash', transaction);
```
Nana-EC, line 523: Need to update error message. There's more specific logic here which includes the record retrieval and a mirror node call and hash calculation.
So it should differentiate from the general error handle at the end of the method

packages/relay/src/lib/eth.ts
```diff
@@ -481,36 +483,49 @@ export class EthImpl implements Eth {
...
+        return EthImpl.prepend0x(createHash('keccak256').update(transactionBuffer).digest('hex'));
```
Nana-EC, line 524: If we're gonna compute this either ways they do it once up before the try.

packages/relay/src/lib/eth.ts
```diff
@@ -481,36 +483,49 @@ export class EthImpl implements Eth {
...
+          this.logger.error('Failed to get a block for transaction');
+          return '';
+        }
+        const computedHash = EthImpl.prepend0x(createHash('keccak256').update(transactionBuffer).digest('hex'));
+
+		if (computedHash != txHash) {
+			this.logger.error("EthereumHash didn't match computed hash");
+		}
+        return txHash;
+      } catch (e) {
```
Nana-EC, line 521: We really need 2 test case coverage for this kind of a flow.
Would have been great to have had that but would likely cause more code churn for this PR.
Eventually we'd wanna mock the sdkclient calls also and have one test for the happy path and one for the error handling.

No-op, just something to note

Nana-EC (changes requested):
Looking good, just 2 simple suggestions

packages/relay/src/lib/eth.ts
mmrt: the one inside of the try was there for debugging, it has been removed

packages/relay/src/lib/eth.ts
mmrt: This error is for a failed handle, the outer error is for a failed submission. What would you like for this to contain?

shemnon: The inner one implies a non-precheck error.  The comment should include something alluding to that like "failed at submit" and this one is "failed at receipt retrieval"

shemnon (approved):
Code looks fine once the two exception messages are no longer identical. 

We need (probably in another PR) unit tests covering both failure modes, not sure if the sdk has sufficient mocking however.

Nana-EC (dismissed):

Nana-EC:
> Code looks fine once the two exception messages are no longer identical.
> 
> We need (probably in another PR) unit tests covering both failure modes, not sure if the sdk has sufficient mocking however.

Yep, agreed x 2
Looked into SDK mocking and have some ideas for next sprint

shemnon (approved):

Nana-EC (approved):

Issue #17, opened 2022-04-15
CI/CD pipeline to run e2e tests


Kalina-Todorova:
Update the pipeline such that we can run e2e tests

PR #120, opened 2022-05-31
Acceptance test CICD

Ivo-Yankov:
**Description**:
Add support for Acceptance tests to run in CI/CD flow

- Add a `localAcceptance.env` that provides configurations for local-node runs
- Update package to pull latest npm version of `hedera-local-node`

**Related issue(s)**:

Addresses #17 

**Notes for reviewer**:

**Checklist**


packages/server/package.json
```diff
@@ -20,7 +20,7 @@
         "pino-pretty": "^7.6.1"
     },
     "devDependencies": {
-        "@hashgraph/hedera-local": "^1.0.1",
+        "@hashgraph/hedera-local": "github:hashgraph/hedera-local-node",
```
Nana-EC, line 23: Q: Was this to just ensure latest bits for now?
It may not be necessary since the tests populate the desired docker version for the configurable local nodes

packages/server/package.json
```diff
@@ -40,7 +40,7 @@
         "compile": "tsc -b tsconfig.json",
         "acceptancetest": "ts-mocha ./tests/acceptance.spec.ts",
         "start": "node dist/index.js",
-        "test": "nyc ts-mocha --recursive ./tests/*.spec.ts"
+        "test": "nyc ts-mocha --recursive ./tests/server.spec.ts"
```
Nana-EC, line 43: Good catch.
If this becomes an issue i can move acceptance tests into its own folder later.

Nana-EC (commented):
Looking good.
One nit suggestion and a question

Nana-EC:
Updated PR comment so that merging this doesn't close #17 and will update my PR to automatically close that noting this PR

packages/server/package.json
Ivo-Yankov: It is necessary, at least until a new version of the package is published to npm. In v1.0.1 the env var `MIRROR_IMAGE_TAG` doesn't exist and the mirror node image version is hardcoded.

Nana-EC: You're right. Good catch

Nana-EC (approved):
LGTM

PR #133, opened 2022-06-01
Improve duplicate transaction management

### LINKED ISSUE #112: Improve duplicate transaction management
### Problem

The current duplicate transaction logic can be improved

### Solution

We want an improved approach to duplicate transaction submission to the relay 
- throw out the transaction with the wrong nonce - note this is not always the first duplicate
- may require interaction with mirror node
- may also explore in app cache for short period 

### Alternatives

_No response_

natanasow:

**Description**:

**Related issue(s)**:

Fixes #112 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -476,13 +477,44 @@ export class EthImpl implements Eth {
...
+    const decodedTx = ethers.utils.RLP.decode(transaction);
+    const [v, r, s] = decodedTx.slice(-3);
+    const signature = ethers.utils.joinSignature({ r, s, v });
+    let accountInfo;
+    try {
+      accountInfo = await this.mirrorNodeClient.getAccount(
+        ethers.utils.recoverAddress(ethers.utils.arrayify(transaction), signature)
+      );
+    } catch (e) {
+      // account not found, handle the error
```
mmrt, line 502: add a FIXME

mmrt: if there's an error in account retrieval should we not also throw?

mmrt (approved):

packages/relay/src/lib/eth.ts
```diff
@@ -517,13 +518,48 @@ export class EthImpl implements Eth {
...
+      // @ts-ignore
+      ethers.utils.joinSignature({ 'r': tx.r, 's': tx.s, 'v': tx.v })
+    );
+    const accountInfo = await this.mirrorNodeClient.getAccount(recoveredAddress);
+
+    if (accountInfo && accountInfo['ethereum_nonce'] > tx.nonce) {
+      throw predefined.NONCE_TOO_LOW;
+    }
+
+    if (accountInfo && accountInfo['ethereum_nonce'] + 1 < tx.nonce) {
```
shemnon, line 550: I don't think we can reliably handle this in the relay.  Consider where a robot is spamming really fast.  They may want to queue up multiple transactions before the account info settles to the new value on the mirror node.  This will ruin the TPS if they need to wait for the TX to do a full transit into the ledger and then through the mirror node.

shemnon: Nonce too low is fine.  Number only ever goes up.  So once it's too low it's always too low.

Daniel-K-Ivanov: If we add a 2-4 second cache for "latest nonce used by account" do you think that it would be sufficient?

shemnon: That won't scale horizontally.  Plus we need to consider robots sending txes to other relay clusters, or directly to the consensus nodes.

Nana-EC (approved):
LG.
Non blocking request to move to it's own class

mmrt (approved):

packages/relay/src/lib/precheck.ts
```diff
@@ -0,0 +1,58 @@
...
+      // @ts-ignore
+      ethers.utils.joinSignature({ 'r': tx.r, 's': tx.s, 'v': tx.v })
+    );
+    const accountInfo = await this.mirrorNodeClient.getAccount(recoveredAddress);
+
+    if (accountInfo && accountInfo['ethereum_nonce'] > tx.nonce) {
+      throw predefined.NONCE_TOO_LOW;
+    }
+  }
+}
```
Nana-EC, line 58: nit: new line

packages/relay/src/lib/precheck.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+    const raw = ethers.utils.serializeTransaction(rsTx);
+    const recoveredAddress = ethers.utils.recoverAddress(
+      ethers.utils.arrayify(ethers.utils.keccak256(raw)),
+      // @ts-ignore
+      ethers.utils.joinSignature({ 'r': tx.r, 's': tx.s, 'v': tx.v })
+    );
+    const accountInfo = await this.mirrorNodeClient.getAccount(recoveredAddress);
+
+    if (accountInfo && accountInfo['ethereum_nonce'] > tx.nonce) {
```
Nana-EC, line 54: nit: make `ethereum_nonce` const
or
```suggestion
    if (accountInfo && accountInfo.ethereum_nonce > tx.nonce) {
```

Nana-EC (dismissed):
LG.
Added 2 nits if i circle back

Nana-EC (approved):
LGTM

Issue #125, opened 2022-06-01
Add acceptance test support for eth_getBalance


Nana-EC:
### Problem

The current acceptance tests implemented in https://github.com/hashgraph/hedera-json-rpc-relay/pull/119 were not able to include `eth_getBalance` since it returned `0x0` instead of a valid balance

### Solution

Add a test that calls `eth_getBalance` for the primary and secondary accounts setup

PR #139, opened 2022-06-02
Acceptance tests - eth_getBalance, eth_getTransactionCount

Ivo-Yankov:
**Description**:
Refactor the current implementation of `eth_getBalance` to use different methods depending on the entity type. For accounts use `AccountBalanceQuery().setAccountId(...)` and for contracts - `AccountBalanceQuery().setContractId(...)`. 
`eth_getTransactionCount` is refactored in a similar way.

**Related issue(s)**:

Fixes #
[125](https://github.com/hashgraph/hedera-json-rpc-relay/issues/125)
[124](https://github.com/hashgraph/hedera-json-rpc-relay/issues/124)

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -156,4 +172,10 @@ export class SDKClient {
             ? input.substring(2)
             : input;
     }
+
+    private static toTinyBar(balance: AccountBalance): BigNumber {
+        return balance.hbars
+            .to(HbarUnit.Tinybar)
+            .multipliedBy(10_000_000_000);
```
Daniel-K-Ivanov, line 179: I assume that we have this constant used somewhere else. We can declare it as variable

packages/server/tests/acceptance.spec.ts
```diff
@@ -209,6 +214,28 @@ describe('RPC Server Integration Tests', async function () {
         expect(blockResult.transactions.length).to.be.greaterThan(0);
     });
 
+    it('should execute "eth_getBalance" for primary account', async function () {
```
Daniel-K-Ivanov, line 217: Lets add tests a long prefixed `0` (e.g 0x000000000123) account & contract

Daniel-K-Ivanov (changes requested):

Nana-EC (commented):
LG. Just a question

packages/relay/src/lib/eth.ts
```diff
@@ -479,8 +479,16 @@ export class EthImpl implements Eth {
...
     } else {
-      const accountInfo = await this.sdkClient.getAccountInfo(address);
-      return Number(accountInfo.ethereumNonce);
+      const result = await this.mirrorNodeClient.resolveEntityType(address);
+      if (result && result.type === 'account') {
+        const accountInfo = await this.sdkClient.getAccountInfo(result.entity.account);
+        return Number(accountInfo.ethereumNonce);
+      }
+      else if (result && result.type === 'contract') {
+        return 1;
```
Daniel-K-Ivanov: We need to dig deeper into this one. 
Some contract addresses return `0x1` and others return `0x2`

Daniel-K-Ivanov: @shemnon do you know the implementation details of `eth_getTransactionCount` when it comes to `contract` addresses?

packages/server/tests/acceptance.spec.ts
```diff
@@ -209,6 +214,68 @@ describe('RPC Server Integration Tests', async function () {
...
+        const res = await callSupportedRelayMethod(this.relayClient, 'eth_getBalance', [mirrorPrimaryAccount.evm_address, 'latest']);
+        expect(res.data.result).to.not.be.equal('0x0');
+        expect(res.data.result).to.eq('0x566527b339630a400');
```
Daniel-K-Ivanov, line 220: Why do we have 2 separate `expect`s for the same value. If we are doing `expect(res.data.result).to.eq('0x566527b339630a400');` do we need `expect(res.data.result).to.not.be.equal('0x0');`?

packages/server/tests/acceptance.spec.ts
```diff
@@ -209,6 +214,68 @@ describe('RPC Server Integration Tests', async function () {
...
+        const res = await callSupportedRelayMethod(this.relayClient, 'eth_getBalance', [idToEvmAddress(mirrorPrimaryAccount.account), 'latest']);
+        expect(res.data.result).to.eq('0x566527b339630a400');
+    });
+
+    it('should execute "eth_getBalance" for contract with id converted to evm_address', async function () {
+        const res = await callSupportedRelayMethod(this.relayClient, 'eth_getBalance', [idToEvmAddress(contractId.toString()), 'latest']);
+        expect(res.data.result).to.eq('0x56bc75e2d63100000');
+    });
+
+    it('should execute "eth_getTransactionCount" primary', async function () {
```
Daniel-K-Ivanov, line 249: Lets add a test in which we test accounts that do have transactions executed, thus the nonce is not `0x0`

Daniel-K-Ivanov (changes requested):

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -36,9 +36,10 @@ import {
     TransactionId
 } from '@hashgraph/sdk';
 import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
+import constants from './../../constants';
 
 export class SDKClient {
-    private static EXCHANGE_RATE_FILE_ID = "0.0.112";
+    private static EXCHANGE_RATE_FILE_ID = constants.EXCHANGE_RATE_FILE_ID;
```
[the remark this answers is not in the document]
Ivo-Yankov: The static is not really necessary. We can just use the constants from the file.

packages/relay/src/lib/eth.ts
```diff
@@ -473,8 +481,16 @@ export class EthImpl implements Eth {
...
     } else {
-      const accountInfo = await this.sdkClient.getAccountInfo(address);
-      return EthImpl.numberTo0x(Number(accountInfo.ethereumNonce));
+      let nonce = 0;
+      const result = await this.mirrorNodeClient.resolveEntityType(address);
+      if (result?.type === 'account') {
+        const accountInfo = await this.sdkClient.getAccountInfo(result?.entity.account);
+        nonce = Number(accountInfo.ethereumNonce);
+      } else if (result?.type === 'contract') {
+        nonce = 1;
```
Daniel-K-Ivanov, line 490: We will have a problem with Contract `nonces`. 
Nonces for a contract is `1 + n` where `n` is the number of times the contract _created other contracts through CREATE or CREATE2_. The deployment of the contract itself is considered part of the `nonce`, that's why the nonce starts from `1`.

Since we are not able to query `getAccountInfo` for contracts, I think the best option is to hardcode the value to `1` and add TODO comment that "currently contract nonces are hardcoded to 1, thus factory contracts will not get their `transactionCount` updated/reflected" 

The long-term solution IMO is to update the nonce of the `contract` for every `CREATE/CREATE2` opcode executed from it and expose it to mirror nodes as information.

@shemnon do you share the same opinion?

Nana-EC: Interesting. I'm okay with the constant for now.
Does that in itself cause errors or blockers?

Also instead of a TODO comment please open a ticket capturing the details of the issue with your suggested solution or other options.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -158,4 +174,10 @@ export class SDKClient {
             ? input.substring(2)
             : input;
     }
+
+    private static toTinyBar(balance: AccountBalance): BigNumber {
```
mmrt, line 178: the code that follows seems to go from hbar -> tiny bar -> weibar but the function name is toTinyBar. Am I missing something simple?

packages/server/tests/acceptance.spec.ts
```diff
@@ -559,10 +625,28 @@ describe('RPC Server Integration Tests', async function () {
...
     const numberTo0x = (input: number): string => {
-        return `0x${input.toString(16)}`;
+        return `0x${toHex(input)}`;
     };
 
     const prune0x = (input: string): string => {
         return input.startsWith('0x') ? input.substring(2) : input;
     };
+
+    const toHex = (num) => {
```
natanasow, line 635: nit: All the utils methods can be added in a Utils class to be usable outside the acceptance.spec.ts

natanasow (approved):
lgtm

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -288,4 +292,23 @@ export class MirrorNodeClient {
             queryParamObject[key] = value;
         }
     }
+
+    public async resolveEntityType(entityIdentifier: string) {
+        const contractResult = await this.getContract(entityIdentifier);
+        if (contractResult) {
+            return {
+                type: 'contract',
```
Nana-EC, line 300: nit: move to constant

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -288,4 +292,23 @@ export class MirrorNodeClient {
...
+                entity: contractResult
+            };
+        }
+        const accountResult = await this.getAccount(entityIdentifier);
+        if (accountResult) {
+            return {
+                type: 'account',
```
Nana-EC, line 307: nit: move to constant

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -81,8 +76,19 @@ export class SDKClient {
             .execute(this.clientMain);
     }
 
+    async getContractBalance(contract: string): Promise<AccountBalance> {
+        return (new AccountBalanceQuery()
+            .setContractId(ContractId.fromString(contract)))
+            .execute(this.clientMain);
+    }
+
+    async getContractBalanceInWeiBar(account: string): Promise<BigNumber> {
```
Nana-EC, line 85:
```suggestion
    async getContractBalanceInTinyBar(account: string): Promise<BigNumber> {
```

packages/relay/src/lib/eth.ts
```diff
@@ -313,7 +313,15 @@ export class EthImpl implements Eth {
     this.logger.trace('getBalance(account=%s, blockNumberOrTag=%s)', account, blockNumberOrTag);
     const blockNumber = await this.translateBlockTag(blockNumberOrTag);
     try {
-      const weibars = await this.sdkClient.getAccountBalanceInWeiBar(account);
+      let weibars: BigNumber | number = 0;
+      const result = await this.mirrorNodeClient.resolveEntityType(account);
+      if (result?.type === 'account') {
```
Nana-EC, line 318: nit: move to constant or utilize constant noted prior

packages/relay/src/lib/eth.ts
```diff
@@ -313,7 +313,15 @@ export class EthImpl implements Eth {
...
+        weibars = await this.sdkClient.getAccountBalanceInWeiBar(result.entity.account);
+      }
+      else if (result?.type === 'contract') {
```
Nana-EC, line 321: nit: move to constant or utilize constant noted prior

packages/relay/src/lib/eth.ts
```diff
@@ -473,8 +481,16 @@ export class EthImpl implements Eth {
     if (blockNumber === 0) {
       return '0x0';
     } else {
-      const accountInfo = await this.sdkClient.getAccountInfo(address);
-      return EthImpl.numberTo0x(Number(accountInfo.ethereumNonce));
+      let nonce = 0;
+      const result = await this.mirrorNodeClient.resolveEntityType(address);
+      if (result?.type === 'account') {
```
Nana-EC, line 486: same note for `contract` and `account` here

Nana-EC (changes requested):
Looking good.
Waiting on some minor changes, a rebase and a requested test noted by Danni

packages/relay/src/lib/clients/sdkClient.ts
Ivo-Yankov: This should stay as it is. The function `toTinyBar` was named incorrectly.

packages/server/tests/acceptance.spec.ts
Ivo-Yankov: This is blocked by the issue described here: https://github.com/hashgraph/hedera-json-rpc-relay/issues/164

packages/server/tests/acceptance.spec.ts
```diff
@@ -185,384 +229,223 @@ describe('RPC Server Integration Tests', async function () {
...
     it('should execute "eth_submitWork"', async function () {
-        const res = await callSupportedRelayMethod(this.relayClient, 'eth_submitWork', []);
+        const res = await utils.callSupportedRelayMethod(this.relayClient, 'eth_submitWork', []);
         expect(res.data.result).to.be.equal(false);
     });
 
-    it('should execute eth_sendRawTransaction legacy', async function () {
-        const res = await callSupportedRelayMethod(this.relayClient, 'eth_sendRawTransaction', ['0x' + legacyTransactionHex]);
-        expect(res.data.result).to.be.equal('0x9ffbd69c44cf643ed8d1e756b505e545e3b5dd3a6b5ef9da1d8eca6679706594');
+    it('should execute "eth_sendRawTransaction" for legacy transactions', async function () {
```
Daniel-K-Ivanov, line 403: I think that it would be best to assert that actually the sender got his balanced reduced and the `to` address got its balance increased with `1`. This would verify that the `sendRawTransaction` was processed, it did not error and returned the manually computed eth hash

packages/server/tests/acceptance.spec.ts
```diff
@@ -185,384 +229,223 @@ describe('RPC Server Integration Tests', async function () {
...
+            to: mirrorContract.evm_address,
+            nonce: 1
+        }, ethCompPrivateKey3);
+
+        const res = await utils.callFailingRelayMethod(this.relayClient, 'eth_sendRawTransaction', [signedTx]);
+        expect(res.data.error.message).to.be.equal('Internal error');
+        expect(res.data.error.code).to.be.equal(-32603);
+    });
+
+    it('should execute "eth_sendRawTransaction" for London transactions', async function () {
```
Daniel-K-Ivanov, line 426: I think that it would be best to assert that actually the sender got his balanced reduced and the `to` address got its balance increased with `1`. This would verify that the `sendRawTransaction` was processed, it did not error and returned the manually computed eth hash

Daniel-K-Ivanov (changes requested):
Looks good. I've added 2 new comments for asserting the result of the transactions executed through `sendRawTransaction`. 

We should add the acceptance test for `nonce != 1`.
Other than that, I think that the rest of the acceptance tests in the pipeline could be added in a separate PR in order to get this one merged already and

Daniel-K-Ivanov (approved):

packages/server/tests/acceptance.spec.ts
```diff
@@ -200,388 +248,257 @@ describe('RPC Server Integration Tests', async function () {
...
+        }, ethCompPrivateKey3);
+
+        const res = await utils.callFailingRelayMethod(this.relayClient, 'eth_sendRawTransaction', [signedTx]);
+        expect(res.data.error.message).to.be.equal('Internal error');
+        expect(res.data.error.code).to.be.equal(-32603);
+    });
 
-        return resp;
-    };
+    it('should execute "eth_sendRawTransaction" for London transactions', async function () {
```
Nana-EC, line 430: We're still missing coverage for EIP 155 right?
Maybe subsequent PR

packages/relay/src/lib/eth.ts
Nana-EC: Missed using `constants.TYPE_ACCOUNT` here

packages/relay/src/lib/eth.ts
```diff
@@ -520,8 +528,16 @@ export class EthImpl implements Eth {
...
       return '0x0';
     } else {
-      const accountInfo = await this.sdkClient.getAccountInfo(address);
-      return EthImpl.numberTo0x(Number(accountInfo.ethereumNonce));
+      const result = await this.mirrorNodeClient.resolveEntityType(address);
+      if (result?.type === 'account') {
+        const accountInfo = await this.sdkClient.getAccountInfo(result?.entity.account);
+        return EthImpl.numberTo0x(Number(accountInfo.ethereumNonce));
+      }
+      else if (result?.type === 'contract') {
```
Nana-EC, line 536: And `constants.TYPE_CONTRACT` here

Nana-EC (approved):
Looks mostly good.
I'm okay with the minor comments going in a subsequent PR

Issue #164, opened 2022-06-07
Add transaction generation support in acceptance tests


Nana-EC:
### Problem

Currently the `sendRawtransaction` flows relies on static examples of raw transaction data.
This is limiting as it results in errors such as `INVALID_ACCOUNT_ID` and could result in a `WRONG_NONCE` error

### Solution

Add raw transaction generation logic with RLP encoding for greater applicability.

- Add a transaction generation class that uses ethers or another library to RLP encode
- Add support for Legacy, EIP155 and EIP1559 
- Accept object param to configure `gasPrice`, `gasLimit`, `value`, `nonce`, `data`, `chainId`, `to`...
- For `data` param eventually support HBAR transfer, token associate , Token transfer and contract function call 

### Alternatives

Since this serves as a dependency for other acceptance tests coverage this should not drag on
- Start with EIP1559/London support for now. Open tickets for Legacy and EIP155 if this ticket drags on
- Start with HBAR transfer and open tickets for token associate , Token transfer and contract function call

PR #172, opened 2022-06-08
Signed transaction generator

Ivo-Yankov:
**Description**:

**Related issue(s)**:

Fixes #
https://github.com/hashgraph/hedera-json-rpc-relay/issues/164

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance.spec.ts
+++ b/packages/server/tests/acceptance.spec.ts
@@ -76,6 +76,8 @@ let ethCompPrivateKey1;
 let ethCompAccountInfo1;
 let ethCompPrivateKey2;
 let ethCompAccountInfo2;
+let ethCompPrivateKey3;
+let ethCompAccountInfo3;
 
 describe('RPC Server Integration Tests', async function () {
     this.timeout(180 * 1000);
@@ -103,13 +105,17 @@ describe('RPC Server Integration Tests', async function () {
         // set up mirror node contents
         logger.info('Submit eth account create transactions via crypto transfers');
         // 1. Crypto create with alias - metamask flow
-        const { accountInfo: primaryAccountInfo, privateKey: primaryKey } = await utils.createEthCompatibleAccount(client);
+        const { accountInfo: primaryAccountInfo, privateKey: primaryKey } = await utils.createEthCompatibleAccount(client, 'a3e5428dd97d479b1ee4690ef9ec627896020d79883c38e9c1e9a45087959888');
         ethCompPrivateKey1 = primaryKey;
         ethCompAccountInfo1 = primaryAccountInfo;
 
-        const { accountInfo: secondaryAccountInfo, privateKey: secondaryKey } = await utils.createEthCompatibleAccount(client);
+        const { accountInfo: secondaryAccountInfo, privateKey: secondaryKey } = await utils.createEthCompatibleAccount(client, '93239c5e19d76c0bd5d62d713cd90f0c3af80c9cb467db93fd92f3772c00985f');
         ethCompPrivateKey2 = secondaryKey;
         ethCompAccountInfo2 = secondaryAccountInfo;
+
+        const ethCompatibleAccount3 = await utils.createEthCompatibleAccount(client, '3e389f612c4b27de9c817299d2b3bd0753b671036608a30a90b0c4bea8b97e74');
+        ethCompPrivateKey3 = ethCompatibleAccount3.privateKey;
+        ethCompAccountInfo3 = ethCompatibleAccount3.accountInfo;
 
         logger.info(`Setup Client for AccountOne: ${primaryAccountInfo.accountId.toString()}`);
         accOneClient = utils.setupClient(primaryKey.toString(), primaryAccountInfo.accountId.toString());
@@ -224,7 +230,7 @@ describe('RPC Server Integration Tests', async function () {
 
     it('should execute "eth_getBalance" for primary account', async function () {
         const res = await utils.callSupportedRelayMethod(this.relayClient, 'eth_getBalance', [mirrorPrimaryAccount.evm_address, 'latest']);
-        expect(res.data.result).to.eq('0x1095793487fe22cac00');
+        expect(res.data.result).to.eq('0x1095793487d8e20c800');
     });
 
     it('should execute "eth_getBalance" for secondary account', async function () {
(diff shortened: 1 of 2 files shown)


packages/server/tests/utils.ts
```diff
@@ -175,8 +177,8 @@ export default class TestUtils {
             .setOperator(AccountId.fromString(id), opPrivateKey);
     };
 
-    createEthCompatibleAccount = async (client: Client) => {
-        const privateKey = PrivateKey.generateECDSA();
+    createEthCompatibleAccount = async (client: Client, privateKeyHex: string) => {
```
Daniel-K-Ivanov, line 180: Is there a reason why we are not generating the ECDSA but rather hardcoding them?

packages/server/tests/utils.ts
```diff
@@ -319,4 +321,9 @@ export default class TestUtils {
         return {contractExecuteTimestamp, contractExecutedTransactionId};
     };
 
+    signRawTransaction = async (tx: TransactionRequest, privateKey) => {
+        const provider = new ethers.providers.JsonRpcProvider('http://localhost:7546');
```
Daniel-K-Ivanov, line 325: I think that we should define the provider in the `describe` within the acceptance test file and pass it as an argument here. We have the localhost:7546 defined there

Daniel-K-Ivanov (changes requested):

packages/server/tests/utils.ts
Ivo-Yankov: Because if they are randomly generated the transaction hash in the result will be different every time and that makes it harder to assert.

packages/server/tests/acceptance.spec.ts
```diff
@@ -360,14 +372,50 @@ describe('RPC Server Integration Tests', async function () {
...
-    it('should execute eth_sendRawTransaction legacy', async function () {
-        const res = await utils.callSupportedRelayMethod(this.relayClient, 'eth_sendRawTransaction', ['0x' + legacyTransactionHex]);
-        expect(res.data.result).to.be.equal('0x9ffbd69c44cf643ed8d1e756b505e545e3b5dd3a6b5ef9da1d8eca6679706594');
-    });
-
-    it('should execute "eth_sendRawTransaction" london', async function () {
-        const res = await utils.callSupportedRelayMethod(this.relayClient, 'eth_sendRawTransaction', ['0x' + londonTransactionHex]);
-        expect(res.data.result).to.be.equal('0xcdbbfb6400aab319f97d32c38e285f0d0399c2b48b683f04878b5f07eb0d50e3');
+    it('should execute "eth_sendRawTransaction" for legacy transactions', async function () {
+        const signedTx = await utils.signRawTransaction({
```
Nana-EC, line 376: nit: move object out to `defaultLegacyTransactionData`. Then you override `to` here

packages/server/tests/acceptance.spec.ts
```diff
@@ -360,14 +372,50 @@ describe('RPC Server Integration Tests', async function () {
...
+        }, ethCompPrivateKey3);
+
+        const res = await utils.callFailingRelayMethod(this.relayClient, 'eth_sendRawTransaction', [signedTx]);
+        expect(res.data.error.message).to.be.equal('Internal error');
+        expect(res.data.error.code).to.be.equal(-32603);
+    });
+
+    it('should execute "eth_sendRawTransaction" for London transactions', async function () {
+        const signedTx = await utils.signRawTransaction({
+            to: mirrorContract.evm_address,
```
Nana-EC, line 407: nit: move object out to `defaultLondonTransactionData`. Then you override `to` here

packages/server/tests/acceptance.spec.ts
```diff
@@ -360,14 +372,50 @@ describe('RPC Server Integration Tests', async function () {
...
+            chainId: 0x12a,
+            gasPrice: 720000000000,
+            gasLimit: 3000000
+        }, ethCompPrivateKey3);
+
+        const res = await utils.callSupportedRelayMethod(this.relayClient, 'eth_sendRawTransaction', [signedTx]);
+        expect(res.data.result).to.be.equal('0x93c4b87f7fe3d6071a9c58acf5b64ec976c60ca2017f21fac42f445472885727');
+    });
+
+    it('should fail "eth_sendRawTransaction" for eip155 transactions', async function () {
```
Nana-EC, line 388: Q: Is there a passing eip155 test support?

Nana-EC (changes requested):
Looking good, some suggestions and potential missing test case

packages/server/tests/acceptance.spec.ts
Ivo-Yankov: Not at this time. I suggest we add it in a later PR

packages/server/tests/acceptance.spec.ts
```diff
@@ -48,17 +48,40 @@ const testLogger = pino({
...
+const defaultLondonTransactionData = {
+    value: 1,
+    chainId: 0x12a,
+    maxPriorityFeePerGas: 720000000000,
+    maxFeePerGas: 720000000000,
+    gasLimit: 3000000,
+    type: 2,
+};
+
+const defaultEIP115TransactionData = {
```
Daniel-K-Ivanov, line 78: It should not be `EIP115` transaction but `2930`
Lets rename it to `defaultLegacy2930TransactionData`

packages/server/tests/acceptance.spec.ts
Daniel-K-Ivanov: The specified acceptance test should be different.
EIP155 stands for transactions including `chain_id` and not for London/Legacy transactions.
Here we are testing that if `type=1` is used, the transaction will fail which is something correct that we need to test for.

Legacy Transactions = `type` is undefined and we have that test on line 401
Post EIP2930, transaction types were introduced and now transactions can have type = 1 (again legacy type of transaction) and type = 2 (London type).
We should rename the acceptance test to `should fail "eth_sendRawTransaction" for Legacy 2930 transactions"`.
We cannot have a passing test for that use case because Hedera does not support legacy 2930 type of transactions

We can add completely separate tests for EIP155 which Imo can be in another PR.

Daniel-K-Ivanov (changes requested):

packages/server/tests/acceptance.spec.ts
Daniel-K-Ivanov: And lets include `type=1` within the `defaultLegacy2930TransactionData` object

Daniel-K-Ivanov (approved):

Daniel-K-Ivanov:
@Nana-EC I am merging this PR in order to continue with the next acceptance tests within the `125-acceptance-test..` branch. If you have any other comments, please leave them in #139

PR #173, opened 2022-06-08
154 document token transfer

### LINKED ISSUE #154: Document Relay Token Transfer Flow using a wallet
### Problem

The relay demo has a few setup steps that are necessary to see a run that highlights the true value of the relay.
Since this can be involving we should document in a separate markdown file for easy reference and reproduction to allow eth devs to hit the ground running

The token transfer flow is one that should be highlighted

**Scenario 2: Transfer HBAR between 2 users**
- Setup up 2 hedera accounts
- Create token
- Associate token with both users 
- Transfer token to them
- Configure separate wallets  to point to a deployed Hedera JSON RPC Relay - could be 2 different relays for same network
- Connect a wallet to each Hedera account
- Add associated token to each account in wallet
- Send Tokens between users

### Solution

Add documentation highlighting the flow.
May include
- Steps
- Code snippets
- Screen shots - if applicable

### Alternatives

Video snips, but this can be done later

mmrt:
**Description**:
Create guide for transferring tokens between two accounts using metamask and associated setup script.

**Related issue(s)**:

Closes #154


docs/token-transfer.md
```diff
@@ -0,0 +1,102 @@
+# Transferring Hbar Using the Hedera JSON-RPC Relay
+
+For illustration purposes, Metamask is used but this guide should be applicable to any JSON-RPC enabled wallet.
+
+## Add Hedera Network to Metamask
+
+1. Open Metamask and click on the circle in the upper right corner
```
Nana-EC, line 7: In this doc no need to repeat the others.
Either 
1. Use the same doc just add a separate section
2. Reference the HBAR doc in here.

My preference would be for a single doc with different scripts where applicable

Nana-EC (changes requested):
Looking good.
I'm thinking 1 doc, what do you think?

docs/token-transfer.md
mmrt: what do you think of the current form?

docs/examples/token-transfer-setup.js
```diff
@@ -0,0 +1,128 @@
...
+    // Assuming that the target shard and realm are known.
+    // For now they are virtually always 0 and 0.
+    const aliasAccountId = publicKey.toAccountId(0, 0);
+
+    /*
+     * Note that no queries or transactions have taken place yet.
+     * This account "creation" process is entirely local.
+     */
+
+    console.log("  Transferring some Hbar to the new account");
```
Nana-EC, line 96: nit: make hbar constant and note in
```suggestion
    const hbarAmount = new Hbar(10);
    console.log(`  Transferring ${hbarAmount} to the new account`);
```

docs/examples/token-transfer-setup.js
```diff
@@ -0,0 +1,128 @@
...
+    const response = await new TransferTransaction()
+        .addHbarTransfer(client.operatorAccountId, new Hbar(10).negated())
```
Nana-EC, line 98: nit: Per comment noted above
```suggestion
        .addHbarTransfer(client.operatorAccountId, hbarAmount.negated())
```

Nana-EC (dismissed):
LGTM
Any nits should be addressed in the follow up contract PR

Nana-EC (approved):

PR #184, opened 2022-06-09
Add ERC20 allowance Interface call to erc spec tests

Nana-EC:
**Description**:
A useful flow to verify would be an ERC 20 interface interacting call

- Add a test to very approve and allowance flow

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance.spec.ts
```diff
@@ -320,6 +344,37 @@ describe('RPC Server Integration Tests', async function () {
         callUnsupportedRelayMethod(this.relayClient, 'eth_protocolVersion', []);
     });
 
+    it('erc20 test', async function () {
```
Daniel-K-Ivanov, line 347: I think that we should move away from the concept of writing our own implementations for calling contracts, creating them etc.

If possible (contract is <6kb) we should do everything inside the acceptance tests as if they were executed from a client dapp script. Example for that would be the following:


```TypeScript
const provider = new ethers.provider.JsonRpcProvider('http://localhost:8545');
const wallet =  new ethers.Wallet(evmCompPK, provider);

// deploy the contract through Hedera SDK since it is > 6kb

const erc20 = new ethers.Contract('0 prefixed address', erc20Abi, wallet);
// now you can do things like
const approveTx = await erc20.approve('some-account-0-prefixed', 100000);

```

packages/server/tests/erc20Contract/ERC20Contract.sol
```diff
@@ -0,0 +1,52 @@
...
+        IERC20(token).transfer(recipient, amount);
+    }
+
+    function delegateTransfer(address token, address recipient, uint256 amount) public {
+        (bool success, bytes memory result) = address(IERC20(token)).delegatecall(abi.encodeWithSignature("transfer(address,uint256)", recipient, amount));
+    }
+
+    //Not supported operations - should return a failure
+
+    function transferFrom(address token, address sender, address recipient, uint256 amount) public {
```
Daniel-K-Ivanov, line 38: transferFrom / approval fails because you have `approved` the `msg.sender` to spend tokens, but after that when you call `transferFrom` you are spending them on the `erc20contract` behalf

Nana-EC:
Addressed in `erc20spec.ts` tests

PR #190, opened 2022-06-10
Ignore vscode and devcontainer tracking

rafaelkallis:
Signed-off-by: Rafael Kallis <rk@rafaelkallis.com>

**Description**:

Devcontainers are a feature of vscode that improve the developer onboarding experience. Essentially, the editor runs inside a container that has all development dependencies preinstalled.

This PR adds devcontainer configuration so that (vscode) developers have a better development experience
* Add devcontainer configuration

**Related issue(s)**:

**Notes for reviewer**:
image

**Checklist**


shemnon:
How does this differ from checking in IDE configurations?

rafaelkallis:
> How does this differ from checking in IDE configurations?

This approach does not require the developer to install any required development dependencies to work on the app, other than docker. This has the following use cases:

1. As a developer I want to work on multiple projects that require different nodejs versions, e.g., one with node v16 and another with v18.
2. As a developer I want a fast project onboarding experience, i.e., have a batteries-included environment that has all runtime dependencies installed so that I can run the app as fast as possible.

I personally suffer from the first issue. I work on many projects, each of which has different dependencies and I have to work inside isolated environments in order to not have dependencies interfere with eachother.

The vscode devcontainers provide the best industry solution so far in my experience. A few more arguments are listed on microsoft docs https://code.visualstudio.com/docs/remote/remote-overview. The devcontainers configurations can also be used with the github codespaces, but I haven't tried that feature yet.

shemnon:
My concern with checkin in IDE configurations is that it may lock us into only being able to build and develop on one tool.  Which is a concern when they don't live forever, like what happened to atom.  A similar argument was made for eclipse years ago and it has now fallen out of "industry standard" status.

How can we insure that checking this in won't result in devcontainers/VSCode being the only way to build and develop?  

As a dev can we keep these files in our own setup and not check them in?  i.e. Is there a reason these must be checked in?

rafaelkallis:
@shemnon 

> How can we insure that checking this in won't result in devcontainers/VSCode being the only way to build and develop?

Convenience vs vendor-lock problem. In the end of the day, the decision of what system dependencies you have and how much of a burden it is to run your app is on you. If your app only requires nodejs 16 and pnpm to run out-of-the-box, the devcontainer setup certainly wont get more complex than what I'm proposing in this PR. 

I agree with you that adding more system dependencies only because of the convenience of devcontainers is not ok.

> As a dev can we keep these files in our own setup and not check them in? i.e. Is there a reason these must be checked in?

Reproducible environment across devs.

If you choose not to include devcontainers, let's at least add an entry to the gitignore so that I can use devcontainers locally and continue working on this project.

Daniel-K-Ivanov:
I personally do not have experience with `devcontainers` but it is true that a lot of web3 devs work with VSCode. I think that the _least_ we can do is add it in the `gitignore` as @rafaelkallis suggests.

shemnon:
Is it just the entire `.devcontainer` directory?  

Do you want me to create a new PR or can this PR be morphed into the `.gitignore` changes?

Nana-EC (approved):
LGTM

Daniel-K-Ivanov (approved):

PR #192, opened 2022-06-10
Acceptance tests for eth_getTransactionReceipt

Ivo-Yankov:
**Description**:
Add tests for the method `eth_getTransactionReceipt` for:
- Negative test for Legacy transaction - contains `value`, `gasPrice`, `gasLimit`, `to`, `nonce`
- Legacy 155 transaction - contains `value`, `gasPrice`, `gasLimit`, `to` and `chainId`, `nonce`
- London transaction - contains `value`, `maxPriorityFeePerGas`, `maxFeePerGas`, `gasLimit`, `to`, `chainId`, `nonce` and `type=2`
- Negative test for Legacy 2930 transaction - contains `value`, `gasPrice`, `gasLimit`, `to`, `chainId`and `type=1`

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance.spec.ts
+++ b/packages/server/tests/acceptance.spec.ts
@@ -19,10 +19,7 @@
  */
 
 // external resources
-import {
-    Client,
-    PrivateKey,
-} from "@hashgraph/sdk";
+import { Client } from "@hashgraph/sdk";
 import Axios from 'axios';
 import axiosRetry from 'axios-retry';
 import { expect } from 'chai';
@@ -103,7 +100,6 @@ let mirrorSecondaryAccount;
 let ethCompPrivateKey3;
 let ethCompAccountInfo3;
 let ethCompAccountEvmAddr3;
-let ethCompNonce = 0;
 
 describe('RPC Server Integration Tests', async function () {
     this.timeout(180 * 1000);
@@ -401,7 +397,6 @@ describe('RPC Server Integration Tests', async function () {
         expect(utils.subtractBigNumberHexes(senderInitialBalance, senderEndBalance).toHexString()).to.not.eq(oneHbarInWeiHexString);
         expect(BigNumber.from(senderInitialBalance).sub(BigNumber.from(senderEndBalance)).gt(0)).to.eq(true);
 
-        ethCompNonce++;
     });
 
     it('should fail "eth_sendRawTransaction" for Legacy 2930 transactions', async function () {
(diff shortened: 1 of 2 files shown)


packages/server/tests/acceptance.spec.ts
```diff
@@ -487,4 +493,55 @@ describe('RPC Server Integration Tests', async function () {
...
+        ethCompNonce++;
+
+        await utils.sleep(3000);
+
+        const submittedLegacyTransactionHash = txRes.data.result;
+        const res = await utils.callSupportedRelayMethod(this.relayClient, 'eth_getTransactionReceipt', [submittedLegacyTransactionHash]);
+
+        utils.assertReceiptNotNull(res.data.result);
+
+        expect(res.data.result.status).to.eq('0x1');
```
Daniel-K-Ivanov, line 514: Instead of having duplicated checks, why dont' we create `assertTransactionReceipt(transactionRequest, transactionReceipt)` and have all the checks there.

packages/server/tests/acceptance.spec.ts
```diff
@@ -398,14 +400,16 @@ describe('RPC Server Integration Tests', async function () {
         expect(senderInitialBalance).to.not.eq(senderEndBalance);
         expect(utils.subtractBigNumberHexes(senderInitialBalance, senderEndBalance).toHexString()).to.not.eq(oneHbarInWeiHexString);
         expect(BigNumber.from(senderInitialBalance).sub(BigNumber.from(senderEndBalance)).gt(0)).to.eq(true);
+
+        ethCompNonce++;
```
Daniel-K-Ivanov, line 404: I don't think that adding dependency between tests (in this case `ethCompNonce`) is a good approach. I would rather get the `nonce` prior to submitting the transaction.

Daniel-K-Ivanov (changes requested):

packages/server/tests/utils.ts
```diff
@@ -351,5 +351,50 @@ export default class TestUtils {
         return BigNumber.from(hex1).sub(BigNumber.from(hex2));
     };
 
+    sleep = async (ms) => {
+        return new Promise(resolve => setTimeout(resolve, ms));
+    };
+
+    getAccountNonce = async (evmAddress) => {
```
Nana-EC, line 358: nit: Add return signature

packages/server/tests/utils.ts
```diff
@@ -351,5 +351,50 @@ export default class TestUtils {
...
+        const nonce = await this.JsonRpcProvider.send('eth_getTransactionCount', [evmAddress, 'latest']);
+        return Number(nonce);
+    };
+
+    sendRawTransaction = async (tx, privateKey) => {
```
Nana-EC, line 363: return signature so it's clear the response is a string hash

Nana-EC (changes requested):
Looking good.
Seeking some clarification

packages/server/tests/utils.ts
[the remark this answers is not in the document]
Ivo-Yankov: Yes, that is correct

Nana-EC (approved):
LG

Daniel-K-Ivanov (approved):

Issue #183, opened 2022-06-09
Refactor Acceptance tests to use ethers.provider


Daniel-K-Ivanov:
**Problem**
Currently the e2e tests are filled with custom logic implemented by us. We can reduce the code size by reusing what ethers providers.

**Solution**
Ethers provides a very simple interface for interacting with JSON RPC providers:
https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#JsonRpcProvider-send

Instead of using the current: 
```
const res = await utils.callSupportedRelayMethod(this.relayClient, 'eth_getTransactionByBlockNumberAndIndex', [mirrorContractDetails.block_number, mirrorContractDetails.transaction_index]);
```
or 
```
const res = await utils.callSupportedRelayMethod(this.relayClient, 'eth_getBalance', [nonExistingAddress, 'latest']);
```

We can use:
```
ethers.provider.send('eth_getBalance', [nonExistingAddress, 'latest']);
```

PR #198, opened 2022-06-13
Workaround for varying eth_getBalance acceptance test result on primary account

Nana-EC:

**Description**:
Work around for `eth_getBalance` results for primary account check
- Updated tests to check for a minimum
- Increase timeout period
- Do a local node restart to ensure there's a clean up of resources prior to a start

**Related issue(s)**:

Workaround for #192 

**Notes for reviewer**:
Coverage exists via the secondary account check and contract balance check. This is simply test logic failure

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance.spec.ts
+++ b/packages/server/tests/acceptance.spec.ts
@@ -116,6 +116,7 @@ describe('RPC Server Acceptance Tests', async function () {
 
             // start local-node
             logger.debug('Start local node');
+            shell.exec('docker volume prune -f');
             shell.exec('npx hedera-local restart');
             logger.trace('Hedera Hashgraph local node env started');
         }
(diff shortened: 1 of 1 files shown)


Daniel-K-Ivanov:
I think that it would be better to handle the balance problem by creating an account explicitly before that and funding it with certain amount of HBARs. I think that this was the approach taken in #183

Nana-EC:
> I think that it would be better to handle the balance problem by creating an account explicitly before that and funding it with certain amount of HBARs. I think that this was the approach taken in #183

Agreed. From the PR description this was just supposed to be a work around to unblock all the unrelated PRs, doesn't actually fix it. Agree with your suggestion though. 
Right now this is failing with docker related issues that i'm able to work around locally but not in github actions yet

Ivo-Yankov (approved):

Issue #166, opened 2022-06-07
Improve eth_estimateGas cached value logic


Nana-EC:
### Problem

The current `eth_estimateGas` implementation uses a single cached value.
However, in some cases this may result in an overly expensive gas cost

### Solution

Add logic to factor in items of the transaction such as value presence. In this case an improved default may be used

Issue #122, opened 2022-06-01
Implement eth_estimateGas with precompiles


Nana-EC:
### Problem

Currently the `eth_estimateGas` endpoint uses a cached value. 
This is due to the inability to delve into contract details and estimate cost of operation

### Solution

The mirror node traceability story needs completion.

Once this is implemented
- Updated `eth_estimateGas` to call the Mirror Node to estimate gas needed to execute a contract method

PR #208, opened 2022-06-14
Improve transfer-only gas estimate

shemnon:
Improve the transfer only gas cost.  If no data is detected return 21k
gas, which is the base tx fee.

**Description**:

**Related issue(s)**:

for #166

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -49,8 +49,8 @@ export class EthImpl implements Eth {
   static emptyArrayHex = '0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347';
   static zeroAddressHex = '0x0000000000000000000000000000000000000000';
   static emptyBloom = "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000";
-  static defaultGas = 0x3d0900;
-  static gasTxBaseCost = 21_000;
+  static defaultGas = EthImpl.numberTo0x(400_000);
+  static gasTxBaseCost = EthImpl.numberTo0x(21_000);
   static ethTxType = 'EthereumTransaction';
   static ethEmptyTrie = '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421';
 
@@ -206,13 +206,12 @@ export class EthImpl implements Eth {
    * Estimates the amount of gas to execute a call.
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  async estimateGas(transaction: any, blockParam: string | null) {
+  async estimateGas(transaction: any, _blockParam: string | null) {
     this.logger.trace('estimateGas()');
     if (!transaction || !transaction.data || transaction.data === '0x') {
-      return EthImpl.numberTo0x(EthImpl.gasTxBaseCost);
+      return EthImpl.gasTxBaseCost;
     } else {
-      // FIXME: For now, we are going to have an arbitrary estimate for non-transfers but in the future we will do something more sophisticated.
-      return EthImpl.numberTo0x(EthImpl.defaultGas);
+      return EthImpl.defaultGas;
     }
   }
 
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -18,16 +18,16 @@
  *
  */
 
-import { Eth } from '../index';
+import {Eth} from '../index';
 import { ContractId, Status, Hbar } from '@hashgraph/sdk';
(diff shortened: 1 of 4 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -203,20 +204,16 @@ export class EthImpl implements Eth {
...
-    // FIXME: For now, we are going to have a rough estimate but in the future we can do something more sophisticated.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async estimateGas(transaction: any, blockParam: string | null) {
     this.logger.trace('estimateGas()');
-    return EthImpl.numberTo0x(EthImpl.defaultGas);
+    if (!transaction || !transaction.data || transaction.data === '0x') {
+      return EthImpl.numberTo0x(EthImpl.gasTxBaseCost);
+    } else {
+      // FIXME: For now, we are going to have an arbitrary estimate for non-transfers but in the future we will do something more sophisticated.
```
Nana-EC, line 214: remove FIXME.
Issue #122 captures the final implementation so we're good on not forgetting

packages/relay/src/lib/eth.ts
```diff
@@ -203,20 +204,16 @@ export class EthImpl implements Eth {
...
+      return EthImpl.numberTo0x(EthImpl.defaultGas);
```
Nana-EC, line 215: hmm how is this working. `defaultGas` is already a hex

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -911,6 +911,7 @@ describe('Eth calls using MirrorNode', async function () {
     mock.onGet(`network/fees`).reply(200, defaultNetworkFees);
     const feeHistory = await ethImpl.feeHistory(1, 'latest', [25, 75]);
     expect(feeHistory).to.exist;
+    if (feeHistory == null) return;
```
Nana-EC, line 914: Q: Why is this and the check below on line 927 necessary? 
It shouldn't be hitting the backup flow where the sdkClient is used.

Nana-EC (changes requested):
Some clarifications

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -935,6 +937,27 @@ describe('Eth calls using MirrorNode', async function () {
     expect(firstFeeHistory).to.equal(secondFeeHistory);
   });
 
+  it('eth_estimateGas contract call returns default', async function() {
+    const gas = await ethImpl.estimateGas({input:"0x01"}, null);
```
Nana-EC, line 941: Your test will fail since you're giving a valid value to transaction which is `{input:"0x01"}`
As a result the first 2 checks you make will be
!transaction => false
!transaction.data => true
hence the non default gas value is used

packages/relay/src/lib/eth.ts
Nana-EC: This will remove 1 of the 2 code smells

packages/relay/src/lib/eth.ts
```diff
@@ -203,20 +204,16 @@ export class EthImpl implements Eth {
...
-   *     "gas": "0x76c0",
-   *     "gasPrice": "0x9184e72a000",
-   *     "value": "0x9184e72a",
-   *     "data": "0xd46e8dd67c5d32be8d46e8dd67c5d32be8058bb8eb970870f072445675058bb8eb970870f072445675"}],
-   */
-  async estimateGas() {
-    // FIXME: For now, we are going to have a rough estimate but in the future we can do something more sophisticated.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async estimateGas(transaction: any, blockParam: string | null) {
```
Nana-EC, line 209: `blockParam` is currently unused hence the other code smell.
If you update the log below to note the params this should fix it

Nana-EC (changes requested):
Suggestions for failing test and code smells

packages/relay/src/lib/eth.ts
shemnon: The param is there because an external API says we may pass it in.  Removing it or un-naming it as `_` will remove the clarity for what the parameter represents.  Automated analysis lacks that context.

packages/relay/src/lib/eth.ts
shemnon: It's still  a javascript number, not a hex string. Needs quotes.  So I went back and put it in quotes and remove the repeated encoding.

packages/relay/src/lib/eth.ts
shemnon: fixed

packages/relay/src/lib/eth.ts
shemnon: fixed

packages/relay/tests/lib/eth.spec.ts
shemnon: Intellij's typescript compiler requires it.  The compiler cannot prove that the 913 line will keep nulls from passing, so at 914 it may be null.  This catch guarantees it.

packages/relay/tests/lib/eth.spec.ts
shemnon: fixed

Nana-EC (approved):
LGTM

Daniel-K-Ivanov (approved):

mmrt (commented):
looks good

PR #211, opened 2022-06-15
Refactor acceptance tests structure

Ivo-Yankov:
**Description**:
Refactor the code structure of the Acceptance tests.
- Add new helper class `ServicesClient`
- Add new helper class `Assertions`
- Consolidates test logic into the TestUtils class. It now has instances of a mirror node axios client and a services client and a JsonRpcProvider.
- All relay calls are performed via JsonRpcProvider.
- `createEthCompatibleAccount` now returns `accountInfo`, `privateKey` and a new `TestUtils` instance, configured to call the services network through a client associated with the new account. To make calls as the new account use  `accounts[n].utils.callRelay(methodName, params)`.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


Daniel-K-Ivanov:
Marking this PR as Draft. I will add a couple commits to prepare the acceptance tests for additional erc20 tests

package.json
```diff
@@ -20,7 +20,7 @@
         ]
     },
     "scripts": {
-        "acceptancetest": "ts-mocha ./packages/server/tests/acceptance.spec.ts --exit",
+        "acceptancetest": "ts-mocha packages/server/tests/acceptance/rpc.spec.ts --exit",
```
Nana-EC, line 23: Can make this pick up all the test files in this directory
```suggestion
        "acceptancetest": "ts-mocha packages/server/tests/acceptance/*.spec.ts --exit",
```

packages/server/package.json
```diff
@@ -41,9 +41,9 @@
         "build": "pnpm clean && pnpm compile",
         "clean": "rm -rf ./dist && rm -rf tsconfig.tsbuildinfo",
         "compile": "tsc -b tsconfig.json",
-        "acceptancetest": "ts-mocha ./tests/acceptance.spec.ts",
+        "acceptancetest": "ts-mocha tests/acceptance/rpc.spec.ts",
```
Nana-EC, line 44:
```suggestion
        "acceptancetest": "ts-mocha tests/acceptance/*.spec.ts",
```

packages/server/tests/clients/mirrorClient.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        this.client = mirrorNodeClient;
+    }
+
+    async get(path: string) {
+        this.logger.debug(`[GET] MirrorNode ${path} endpoint`);
+        return (await this.client.get(path)).data;
+    };
+
+}
```
Nana-EC, line 66: nit: new line

packages/server/tests/clients/relayClient.ts
```diff
@@ -0,0 +1,114 @@
...
+        this.provider = new ethers.providers.JsonRpcProvider(relayUrl);
+    }
+
+    /**
+     * Calls the specified methodName with the provided params
+     * @param methodName
+     * @param params
+     */
+    async call(methodName: string, params: any[]) {
+        this.logger.debug(`[POST] to relay '${methodName}' with params [${params}]`);
```
Nana-EC, line 42: nit: no need for first log as if it fails other logic will make it clear and in success the response line is there

packages/server/tests/clients/relayClient.ts
```diff
@@ -0,0 +1,114 @@
...
+        const result = await this.provider.send(methodName, params);
+        this.logger.trace(`[POST] to relay '${methodName}' with params [${params}] returned ${JSON.stringify(result)}`);
+        return result;
```
Nana-EC, line 45: SHould we include expected checks here
```ts
expect(result).to.have.property('data');
        expect(result.data).to.have.property('id');
        expect(result.data).to.have.property('jsonrpc');
        expect(result.data.jsonrpc).to.be.equal('2.0');
```

Nana-EC: Actually moved my above suggestion into an `Assertions.successResponse()`

packages/server/tests/clients/relayClient.ts
```diff
@@ -0,0 +1,114 @@
...
+     * @param params
+     */
+    async callFailing(methodName: string, params: any[]) {
+        try {
+            const res = await this.call(methodName, params);
+            this.logger.trace(`[POST] to relay '${methodName}' with params [${params}] returned ${JSON.stringify(res)}`);
+            Assertions.expectedError();
+        } catch (err) {
+            const parsedError = JSON.parse(err.body);
+            expect(parsedError.error.message).to.be.equal('Unknown error invoking RPC');
```
Nana-EC, line 60: Can move this to a `Assertions.unknownResponse(err)`

packages/server/tests/clients/relayClient.ts
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * Returns: Transaction hash
+     * @param signedTx
+     */
+    async sendRawTransaction (signedTx): Promise<string> {
+        this.logger.debug(`[POST] to relay for eth_sendRawTransaction`);
+        return this.provider.send('eth_sendRawTransaction', [signedTx]);
+    };
+
+}
```
Nana-EC, line 114: nit: new line

packages/server/tests/clients/servicesClient.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    public readonly client: Client;
+
+    constructor(network: string, accountId: string, key: string, logger: Logger) {
+        this.logger = logger;
+        this.network = network;
+
+        if (!network) network = '{}';
+        const opPrivateKey = PrivateKey.fromString(key);
+        if (network.toLowerCase() in supportedEnvs) {
```
Nana-EC, line 61: Verify that this works cause i'm not sure it does.
In my other PR i had to change it to 
```suggestion
        if (supportedEnvs.includes(network.toLowerCase())) {
```

packages/server/tests/clients/servicesClient.ts
```diff
@@ -0,0 +1,246 @@
...
+            this.logger.info(`Execute ${query.constructor.name} query`);
+            return query.execute(this.client);
+        } catch (e) {
+            this.logger.error(e, `Error executing ${query.constructor.name} query`);
+        }
+    };
+
+    async executeTransaction(transaction: Transaction) {
+        try {
+            this.logger.info(`Execute ${transaction.constructor.name} transaction`);
```
Nana-EC, line 80: nit: drop this line but add transaction type to response line

packages/server/tests/clients/servicesClient.ts
```diff
@@ -0,0 +1,246 @@
...
+    };
+
+    async createToken() {
+        const symbol = Math.random().toString(36).slice(2, 6).toUpperCase();
+        this.logger.trace(`symbol = ${symbol}`);
+        const resp = await this.executeAndGetTransactionReceipt(new TokenCreateTransaction()
+            .setTokenName(`relay-acceptance token ${symbol}`)
+            .setTokenSymbol(symbol)
+            .setDecimals(3)
+            .setInitialSupply(new Hbar(1000).toTinybars())
```
Nana-EC, line 113: nit: make this configureable with a pass in but set default as is

packages/server/tests/clients/servicesClient.ts
```diff
@@ -0,0 +1,246 @@
...
+        return contractId;
+    };
+
+    async executeContractCall(contractId) {
+        // Call a method on a contract exists on Hedera, but is allowed to mutate the contract state
+        this.logger.info(`Execute contracts ${contractId}'s createChild method`);
+        const contractExecTransactionResponse =
+            await this.executeTransaction(new ContractExecuteTransaction()
+                .setContractId(contractId)
+                .setGas(75000)
```
Nana-EC, line 188: make this configureable but with current default

packages/server/tests/clients/servicesClient.ts
```diff
@@ -0,0 +1,246 @@
...
+
+        // The contract ID is located on the transaction receipt
+        const contractId = contractReceipt?.contractId;
+
+        this.logger.info(`new contract ID: ${contractId?.toString()}`);
+
+        return contractId;
+    };
+
+    async executeContractCall(contractId) {
```
Nana-EC, line 182: I would prefer passing in the functionName and `ContractFunctionParameters` so this method can be reused

packages/server/tests/clients/servicesClient.ts
```diff
@@ -0,0 +1,246 @@
...
+        this.logger.info(`Balance of the new account: ${balance.toString()}`);
+
+        const accountInfo = await this.executeQuery(new AccountInfoQuery()
+            .setAccountId(aliasAccountId));
+        this.logger.info(`New account Info: ${accountInfo.accountId.toString()}`);
+        return {
+            alias: aliasAccountId,
+            accountId: accountInfo.accountId,
+            address: accountInfo.contractAccountId,
+            client: new ServicesClient(
```
Nana-EC, line 230: Why the whole client?

packages/server/tests/clients/servicesClient.ts
```diff
@@ -0,0 +1,246 @@
...
+        await aliasCreationResponse?.getReceipt(this.client);
+
+        const balance = await this.executeQuery(new AccountBalanceQuery()
+            .setAccountId(aliasAccountId));
+        this.logger.info(`Balance of the new account: ${balance.toString()}`);
+
+        const accountInfo = await this.executeQuery(new AccountInfoQuery()
+            .setAccountId(aliasAccountId));
+        this.logger.info(`New account Info: ${accountInfo.accountId.toString()}`);
+        return {
```
Nana-EC, line 226: Can create an interface in here that this maps to and update the method signature so other classes know what to expect

packages/server/tests/clients/servicesClient.ts
```diff
@@ -0,0 +1,246 @@
...
+                privateKey.toString(),
+                this.logger.child({ name: `services-client` })
+            ),
+            wallet: new ethers.Wallet(privateKey.toStringRaw())
+        };
+
+    };
+
+    _thisAccountId() {
+        return this.client.operatorAccountId || AccountId.fromString('0.0.0');
```
Nana-EC, line 242: WHen is the fall back ever used. Shouldn't all tests fail if the client isn't configured right on startup?

packages/server/tests/helpers/assertions.ts
```diff
@@ -0,0 +1,96 @@
...
+export default class Assertions {
+    static assertId = (id) => {
+        const [shard, realm, num] = id.split('.');
+        expect(shard).to.not.be.null;
+        expect(realm).to.not.be.null;
+        expect(num).to.not.be.null;
+    };
+
+    static failedResponce = (resp: any) => {
+        expect(resp).to.have.property('error');
```
Nana-EC, line 32: nit: can verify it has a code

packages/server/tests/helpers/assertions.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    static expectedError = () => {
+        expect(true).to.eq(false);
+    };
+
+    public static block(relayResponse, mirrorNodeResponse) {
+        expect(relayResponse.hash).to.be.equal(mirrorNodeResponse.hash.slice(0, 66));
+        expect(relayResponse.number).to.be.equal(Utils.numberTo0x(mirrorNodeResponse.number));
+        // expect(relayResponse.transactions.length).to.equal(mirrorNodeResponse.count); // TODO this assertion fails
+        // TODO add more assertions
```
Nana-EC, line 49: not a fan of `TODO`'s
Just create a ticket with these and below captured inside so it doesn't get lost

Nana-EC (changes requested):
Lots of good stuff.
Added suggestions.
Let me know if you want to address nit stuff in a  follow up related PR

packages/server/tests/clients/relayClient.ts
Daniel-K-Ivanov: Actually, the `provider` parses the response and returns only the actual information within `data`. Meaning if we console.log `result` we will get the underlying `data` and not the whole `json` object.

packages/server/tests/clients/servicesClient.ts
Daniel-K-Ivanov: updated it although I think that we should migrate to using the `ethers.Contract` instance for other types of tests.

Ivo-Yankov (commented):
Looks good to me

packages/server/tests/clients/servicesClient.ts
Daniel-K-Ivanov: I've added the client in order to have an easy access to a client instance for the newly created alias account. It is used within the specs and simplifies developer experience. The concept is -> Alias account has different "connected clients", one of them is the `services client` and the other one is the `ethers.Wallet`

packages/server/tests/clients/servicesClient.ts
Daniel-K-Ivanov: Practically `this.client.operatorAccountId` will/should never be `unidentified`, however the SDK implementation returns `AccountId || null` for `.operatorAccountId` so the IDE marks it as an error and says that the operator account ID is potentially `null` so it **should not** be used within other functions f.e adding it as a transfer sender when doing HBAR transfer. 

I've added `|| AccountId.fromString('0.0.0')` just to remove the error from the IDE that was all over the place

Nana-EC (approved):
LG

PR #225, opened 2022-06-16
Acceptance tests different references - getCode, getBalance, getTransactionCount

Ivo-Yankov:
**Description**:
- Add acceptance tests for `eth_getCode`
- Add additional test scenarios for `eth_getBalance` and `eth_getTransactionCount` for different account/contract references.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance.spec.ts
```diff
@@ -374,13 +383,23 @@ describe('RPC Server Acceptance Tests', async function () {
     });
 
     it('should execute "eth_getBalance" for newly created account with 5000 HBAR', async function () {
-        const { accountInfo } = await utils.createEthCompatibleAccount(client, null, 5000);
+        const accountResponse = await utils.createEthCompatibleAccount(client, null, 5000);
```
Nana-EC, line 386: Can we reduce the 5000 everywhere to like 10 hbar.
I want to make it cheaper to run when pointing to a remote deployment

packages/server/tests/acceptance.spec.ts
```diff
@@ -374,13 +383,23 @@ describe('RPC Server Acceptance Tests', async function () {
...
+        const accountInfo = accountResponse.accountInfo;
         const mirrorResponse = await utils.callMirrorNode(`accounts/${accountInfo.accountId}`);
         const mirrorAccount = mirrorResponse.data;
         const res = await utils.callSupportedRelayMethod('eth_getBalance', [mirrorAccount.evm_address, 'latest']);
         expect(res.data.result).to.eq('0x10f0777f464e77a2400');
     });
 
+    it('should execute "eth_getBalance" with account alias', async function () {
+        const accountResponse = await utils.createEthCompatibleAccount(client, null, 5000);
```
Nana-EC, line 395: Q: could you use mirrorAccount3 here or what's the reason to create another account?

packages/server/tests/utils.ts
```diff
@@ -306,6 +306,24 @@ export default class TestUtils {
         this.logger.info(`Balances of the new account: ${balance.toString()}`);
     };
 
+    deployContract = async(contract, client: Client, privateKey) => {
+        const fileCreateTx = new FileCreateTransaction()
```
Nana-EC, line 310: nit: Meant to double check but i think the SDK has a `ContractCreateFLow` (or something similar) that might do all this for us.
Try it out once, if it doesn't work immediately with like 5 mins trying then revert and leave as is :)

Nana-EC (changes requested):
Looking good.
2 mainly cost related requests

packages/server/tests/utils.ts
Ivo-Yankov: This function will be refactored when [PR 211](https://github.com/hashgraph/hedera-json-rpc-relay/pull/211) gets merged. I would prefer to try it then.

packages/server/tests/acceptance.spec.ts
Nana-EC: Okay. Just want to reduce cost run but this is fine for now

packages/server/tests/utils.ts
Nana-EC: Works for me

packages/server/tests/acceptance.spec.ts
Nana-EC: If you want to address this in PR 211 that works for me

packages/relay/src/lib/eth.ts
```diff
@@ -393,11 +393,11 @@ export class EthImpl implements Eth {
     } catch (e: any) {
       // handle INVALID_CONTRACT_ID
       if (e?.status?._code === Status.InvalidContractId._code) {
-        this.logger.debug('Unable to find code for contract %s in block "%s", returning 0x0', address, blockNumber);
+        this.logger.debug('Unable to find code for contract %s in block "%s", returning 0x0, err code: %s', address, blockNumber, e?.status?._code);
         return '0x0';
       }
 
-      this.logger.error(e, 'Error raised during getCode for address %s', address);
+      this.logger.error(e, 'Error raised during getCode for address %s, err code: %s', address, e?.status?._code);
```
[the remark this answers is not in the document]
Nana-EC: Hmm not sure, i'm seeing some weirdness locally where my local-node isn't running anymore so not sure but let's watch out for it

Nana-EC (approved):
LG

Nana-EC (approved):

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -110,11 +113,15 @@ describe('RPC Server Acceptance Tests', function() {
         // get contract details
         mirrorContract = await mirrorNode.get(`/contracts/${contractId}`);
 
+        logger.info('Create basic contract');
+        basicContractReceipt = await servicesNode.deployContract(basicContractJson, process.env.OPERATOR_KEY_MAIN);
```
Daniel-K-Ivanov, line 117: Why do we need to pass the operator key? `servicesNode` contains a fully working client.

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -316,6 +323,11 @@ describe('RPC Server Acceptance Tests', function() {
             expect(res).to.be.equal('0x0');
         });
 
+        it('should execute "eth_getTransactionCount" for account with non-zero nonce with alais', async function () {
```
Daniel-K-Ivanov, line 326: We do not actually need this test. In `should execute "eth_getTransactionCount" for account with non-zero nonce` we are querying the `transactionCount` for `account.address` which is a alias created account.

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -110,11 +113,15 @@ describe('RPC Server Acceptance Tests', function() {
...
 
+        logger.info('Create basic contract');
+        basicContractReceipt = await servicesNode.deployContract(basicContractJson, process.env.OPERATOR_KEY_MAIN);
+
         // // get contract details
         mirrorContractDetails = await mirrorNode.get(`/contracts/${contractId}/results/${contractExecuteTimestamp}`);
 
         mirrorPrimaryAccount = (await mirrorNode.get(`accounts?account.id=${accounts[0].accountId}`)).accounts[0];
         mirrorSecondaryAccount = (await mirrorNode.get(`accounts?account.id=${accounts[1].accountId}`)).accounts[0];
+        mirrorThirdAccount = (await mirrorNode.get(`accounts?account.id=${accounts[2].accountId}`)).accounts[0];
```
Daniel-K-Ivanov, line 124: We do not need this query / property

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -368,12 +380,12 @@ describe('RPC Server Acceptance Tests', function() {
 
     it('should execute "eth_getBalance" for contract', async function() {
         const res = await relay.call('eth_getBalance', [Utils.idToEvmAddress(contractId.toString()), 'latest']);
-        expect(res).to.eq('0x56bc75e2d63100000');
+        expect(res).to.eq('0xde0b6b3a7640000');
```
Daniel-K-Ivanov, line 383: I think that it would be better to introduce var at the top of the file
```TypeScript
const ONE_WEIBAR = ethers.utils.parseEther('1', 18)
```
and here assert:
```TypeScript
expect(res).to.eq(ethers.utils.hexValue(ONE_WEIBAR));
```

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -368,12 +380,12 @@ describe('RPC Server Acceptance Tests', function() {
...
     });
 
     it('should execute "eth_getBalance" for contract with id converted to evm_address', async function() {
         const res = await relay.call('eth_getBalance', [Utils.idToEvmAddress(contractId.toString()), 'latest']);
-        expect(res).to.eq('0x56bc75e2d63100000');
+        expect(res).to.eq('0xde0b6b3a7640000');
```
Daniel-K-Ivanov, line 388: Same comment as above

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -476,6 +488,36 @@ describe('RPC Server Acceptance Tests', function() {
         });
     });
 
+    describe('eth_getCode', () => {
```
Daniel-K-Ivanov, line 491: Since we will be using the `basic` contract only for the `eth_getCode` rpc calls I think that it is better to add a `before` here and deploy the contract within this scope

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -476,6 +488,36 @@ describe('RPC Server Acceptance Tests', function() {
...
+        });
+
+        it('should fail "eth_getCode" for account evm_address', async function () {
+            const evmAddress = Utils.idToEvmAddress(accounts[2].accountId.toString());
+            const res = await relay.call('eth_getCode', [evmAddress]);
+            expect(res).to.eq('0x0');
+        });
+
+        it('should fail "eth_getCode" for account alias', async function () {
+            const res = await relay.call('eth_getCode', [mirrorPrimaryAccount.alias]);
```
Daniel-K-Ivanov, line 516: in order to be consistent, lets use the same `accounts[2]` but do `Utils.idToEvmAddress(accounts[2].alias)`

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -476,6 +488,36 @@ describe('RPC Server Acceptance Tests', function() {
...
+            expect(res).to.eq(basicContractJson.deployedBytecode);
+        });
+
+        it('should execute "eth_getCode" for contract with id converted to evm_address', async function () {
+            const evmAddress = Utils.idToEvmAddress(basicContractReceipt.contractId.toString());
+            const res = await relay.call('eth_getCode', [evmAddress]);
+            expect(res).to.eq(basicContractJson.deployedBytecode);
+        });
+
+        it('should fail "eth_getCode" for non-existing contract', async function () {
```
Daniel-K-Ivanov, line 504: nit: rename to `should return 0x for...`

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -476,6 +488,36 @@ describe('RPC Server Acceptance Tests', function() {
...
+            const res = await relay.call('eth_getCode', [NON_EXISTING_ADDRESS]);
+            expect(res).to.eq('0x0');
+        });
+
+        it('should fail "eth_getCode" for account evm_address', async function () {
```
Daniel-K-Ivanov, line 509: nit: rename to `should return 0x for...`

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -476,6 +488,36 @@ describe('RPC Server Acceptance Tests', function() {
...
+            const evmAddress = Utils.idToEvmAddress(accounts[2].accountId.toString());
+            const res = await relay.call('eth_getCode', [evmAddress]);
+            expect(res).to.eq('0x0');
+        });
+
+        it('should fail "eth_getCode" for account alias', async function () {
```
Daniel-K-Ivanov, line 515: nit: rename to `should return 0x for...`

packages/server/tests/clients/servicesClient.ts
```diff
@@ -236,6 +236,24 @@ export default class ServicesClient {
         );
     };
 
+    async deployContract(contract, privateKey, gas = 100000) {
```
Daniel-K-Ivanov, line 239: This will not work for every `contract` provided. Let's refactor the function so that it uses the [EthereumFlow](https://github.com/hashgraph/hedera-sdk-js/blob/e696b77d3ee60b6004ef03723654a13aeb540812/src/EthereumFlow.js) we can then use the same function for deploying `basic` contract and `parentContract` and delete `createParentContract` all together

Daniel-K-Ivanov (changes requested):

packages/server/tests/clients/servicesClient.ts
Daniel-K-Ivanov: Lets address this in the PR for erc20 contract deployment #231

Issue #188, opened 2022-06-10
JSON RPC Acceptance Test Plan


Daniel-K-Ivanov:
### Higher Priority
**eth_blockNumber**
- [ ] #217 

**eth_call** 
- [ ] #218

**getTransactionByHash**
- [x] #220

**eth_getTransactionReceipt**
- [ ] #219

### Low Priority

**eth_getBlockByHash**
- [x] #221

**eth_getBlockByNumber**
- [ ] #222

**eth_getBlockTransactionCountByNumber**
- [x] Add a negative test that queries the block transaction count for non-existing block number. The response must be asserted to:
```
{
  "result": null,
  "jsonrpc": "2.0",
  "id": 0
}
```
**eth_getBlockTransactionCountByHash**
- [x] Add a negative test that queries the block transaction count for non-existing block hash. The response must be asserted to:
```
{
  "result": null,
  "jsonrpc": "2.0",
  "id": 0
}
```
**eth_getUncleByBlockHashAndIndex**
- [x] Enhance the current test to pass arguments -> valid block hash and `0` index. Assert that the response is:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```
- [x] Add new test that uses **non existing** block hash and index `0`. Assert that the response is:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```
**eth_getUncleByBlockNumberAndIndex**
- [x] Enhance the current test to pass arguments -> valid block number and `0` index. Assert that the response is:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```
- [x] Add new test that uses **non existing** block number and index `0`. Assert that the response is:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```

**eth_feeHistory**
- [x] Add positive test
**eth_getCode**
- [ ] Deploy ERC20 and assert the code with eth_getCode
- [x] Execute getCode for non-contract address and assert the empty 0x response
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": "0x"
}
```

**eth_getTransactionByBlockNumberAndIndex**
- [x] Add negative test by providing invalid block number. Assert that the response is
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```
- [x] Add negative test by providing invalid index. Assert that the response is
```
{
  "jsonrpc": "2.0",
  "id": 0,
[...]

PR #233, opened 2022-06-17
Acceptance test plan

### LINKED ISSUE #217: Add eth_blockNumber acceptance test
### Problem

No test exists to confirm `eth_blockNumber` and its return

### Solution

Add an acceptance test that asserts the returned response from `eth_blockNumber` is valid and is an integer.

### Alternatives

Add as an acceptance test in `eth.spec.ts`

### LINKED ISSUE #219: Update eth_getTransactionReceipt to verify return schema
### Problem

Currently the `eth_getTransactionReceipt` does not confirm the return response schema for a non existing transaction hash 

### Solution

Update the `eth_getTransactionReceipt` no match test to confirm the response matches
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```

### Alternatives

_No response_

### LINKED ISSUE #220: Add eth_ getTransactionByHash acceptance tests
### Problem

Currently no `getTransactionByHash` acceptance tests exist

### Solution

- Add negative test getting the transaction by non existing hash. Assert the empty response
- Submit a transaction, save the transaction hash and query the getTransactionByHash endpoint to assert the response

### Alternatives

_No response_

### LINKED ISSUE #221: Add additional eth_getBlockByHash acceptance tests
### Problem

The initial coverage was scarce, additonal coverage can be added.

### Solution

- [x] Add negative test for non-existing block hash with `hydrated transactions set to **false**`. You can use `0x52db1930f5ca0a47423a62637cf9632823fdd24d61dc97c3263417b38c31eee6` f.e. The response of the call must be asserted to be:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```
- [x] Add negative test for non-existing block hash with `hydrated transactions set to **true**`. You can use `0x52db1930f5ca0a47423a62637cf9632823fdd24d61dc97c3263417b38c31eee6` f.e. The response of the call must be asserted to be:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```
- [x] Add a test that queries **existing block** by hash but sets `hydrated transactions to **false**. Assert that the response contains `transactions` array but the values are strings of the **transaction hash**
- [x] Enhance the already existing `getBlockByHash` test by asserting that the returned object has `transactions` array **containing the full transaction** and not only the string of their hash

### Alternatives

Where it makes sense some of these tests may be integration tests

### LINKED ISSUE #222: Add additional eth_getBlockByNumber acceptance tests
### Problem

The initial eth_getBlockByNumber coverage was scarce

### Solution

Add move coverage

- [x] Enhance the already existing `getBlockByNumber` test by adding assertion of the returned object that it has `transactions` array **containing the full transaction** and not only the string of their hash (testing the "hydrated transactions flag")
- [x] Add a test that queries **existing block** by **number** but sets `hydrated transactions to **false**. Assert that the response contains `transactions` array but the values are strings of the **transaction hash**
- [x] Add negative test for non-existing block number with `hydrated transactions set to **true**`. The response of the call must be asserted to be:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```
- [x] Add negative test for non-existing block number with `hydrated transactions set to **false**`. The response of the call must be asserted to be:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
Adds all tests described in [188](https://github.com/hashgraph/hedera-json-rpc-relay/issues/188), apart from 2, which will be addressed in a separate PR

**Related issue(s)**:

Fixes #217 
Fixes #220
Fixes #219
Fixes #221
Fixes #222

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -120,31 +124,106 @@ describe('RPC Server Acceptance Tests', function() {
...
+            Assertions.block(blockResult, mirrorBlock, mirrorTransactions, false);
+        });
+
+        it('should execute "eth_getBlockByNumber", hydrated transactions = true', async function() {
             const blockResult = await relay.call('eth_getBlockByNumber', [mirrorBlock.number, true]);
-            Assertions.block(blockResult, mirrorBlock);
+            Assertions.block(blockResult, mirrorBlock, mirrorTransactions, true);
+        });
+
+        it('should execute "eth_getBlockByNumber", hydrated transactions = true', async function() {
```
Nana-EC, line 182: Updated test names to make sure they are all unique and describe the unique test

packages/server/tests/helpers/utils.ts
```diff
@@ -43,4 +47,8 @@ export class Utils {
         return BigNumber.from(hex1).sub(BigNumber.from(hex2));
     };
 
+    static tinyBarsToWeibars = (value) => {
+        return Utils.numberTo0x(Number.parseInt(value) * 10_000_000_000);
```
Nana-EC, line 51: nit: make `10_000_000_000` const

Nana-EC (commented):
Looking good. Final nit updates and will do one more pass

packages/server/tests/helpers/utils.ts
```diff
@@ -27,6 +27,10 @@ export class Utils {
         return parseInt(num).toString(16);
     };
 
+    static numberTo0x = (input: number): string => {
```
Daniel-K-Ivanov, line 30: Instead of using `numberTo0x` lets use the `ethers.utils.hexValue`

packages/server/tests/helpers/utils.ts
Daniel-K-Ivanov: Lets use directly the util method from ethers:
```TypeScript
return ethers.utils.hexValue(ethers.utils.parseUnits(value, 10));
```

Daniel-K-Ivanov (changes requested):

Daniel-K-Ivanov (dismissed):

Nana-EC (approved):

Nana-EC (approved):

PR #236, opened 2022-06-17
ECRecover Proof of Concept

### LINKED ISSUE #167: Investigate limitation of eth public address
### Problem

Currently eth ecosystem uses the public address (which is the last 20 bytes of the Keccak-256 hash of the public key with a 0x prefix) to identify accounts.
However, Hedera requires the public key not the address to serve as an alias for account creation and reference.
This makes it challenging to use the public address as is desired

### Solution

Explore the limitations and possibilities of using the public address as
- an alias for account creation
- a reference during transactions

Deliverables would be
- Document the limitations
- Open tickets for the possibilities

### Alternatives

_No response_

shemnon:
Proof of concept to demonstrate ECRECOVER precompile

**Description**:

**Related issue(s)**:

Fixes #167 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/dapp-example/src/ExtractPublicKey/EtractPublicKey.html
+++ b/dapp-example/src/ExtractPublicKey/EtractPublicKey.html
@@ -1,0 +1,56 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+</head>
+
+<body>
+
+<button type="button" class="btn btn-default" id="conenctButton" onclick="connect()">
+    Connect
+</button>
+<br><br>
+<label for="contract">Connected Address:</label>
+<input type="text" class="form-control" id="wallet"
+       placeholder="Press Connect" value="" size="50" enabled="false">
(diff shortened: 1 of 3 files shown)


Nana-EC:
Wondering if we should move this over to the `smart-contracts` repo and reference in documentation here that wallet users may utilize this as one of the many flows to extract the public key.

dapp-example/src/ExtractPublicKey/EtractPublicKey.html
```diff
@@ -0,0 +1,56 @@
+<!DOCTYPE html>
```
natanasow, line 1: nit: typo in the file name `EtractPublicKey.html`

natanasow (approved):
LG

Nana-EC:
Will take over moving this to appropriate location in this repo or in the Smart Contracts repo.

PR #240, opened 2022-06-18
Handle Mirror Node network fee errors

### LINKED ISSUE #239: getFeeWeibars failing on mirror node network/fees 500
### Problem

The current version of the relay handles 404 for Mirror Node network/fees calls.
However, it doesn't have logic to support a 500 failure 

### Solution

Add logic to handle all error from mirror node call

### Alternatives

_No response_

Nana-EC:

**Description**:
Expand mirror node network fees fallback to encompass all errors
- Add warning log
- Expand test integration capabilities using `sinon`
- Add `eth.spec.ts` coverage for mirror node failures for gas endpoints

**Related issue(s)**:

Fixes #239 

**Notes for reviewer**:
Locally pointed against testnet backends to confirm fix

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -128,6 +128,7 @@ export class EthImpl implements Eth {
    * Gets the fee history.
    */
   async feeHistory(blockCount: number, newestBlock: string, rewardPercentiles: Array<number> | null) {
+    console.log(`feeHistory(blockCount=${blockCount}, newestBlock=${newestBlock}, rewardPercentiles=${rewardPercentiles})`);
     this.logger.trace(`feeHistory(blockCount=${blockCount}, newestBlock=${newestBlock}, rewardPercentiles=${rewardPercentiles})`);
     try {
       let feeHistory: object | undefined = cache.get(constants.CACHE_KEY.FEE_HISTORY);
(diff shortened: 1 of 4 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -147,7 +147,13 @@ export class EthImpl implements Eth {
   }
 
   private async getFeeWeibars() {
-    let networkFees = await this.mirrorNodeClient.getNetworkFees();
+    let networkFees;
+
+    try {
+      networkFees = await this.mirrorNodeClient.getNetworkFees();
+    } catch (e: any) {
+      this.logger.debug(`Failed to retrieve Network fees from Mirror Node`);
```
steven-sheehy, line 155: Should log at the underlying error message

packages/relay/src/lib/eth.ts
```diff
@@ -147,7 +147,13 @@ export class EthImpl implements Eth {
...
+    }
 
     if (_.isNil(networkFees)) {
       this.logger.debug(`Mirror Node returned no fees. Fallback to network`);
```
steven-sheehy, line 159: This will log two statements for one error scenario, which is not ideal.

packages/relay/src/lib/eth.ts
steven-sheehy, line 155: Should be warn level if an upstream has an error.

steven-sheehy (changes requested):

packages/relay/src/lib/eth.ts
Nana-EC: Captured underlying error

packages/relay/src/lib/eth.ts
Nana-EC: Moved up so only 1 of the 2 logs show up

packages/relay/src/lib/eth.ts
Nana-EC: Updated to warn

packages/relay/src/lib/eth.ts
```diff
@@ -147,10 +147,16 @@ export class EthImpl implements Eth {
   }
 
   private async getFeeWeibars() {
-    let networkFees = await this.mirrorNodeClient.getNetworkFees();
+    let networkFees;
 
-    if (_.isNil(networkFees)) {
+    try {
+      networkFees = await this.mirrorNodeClient.getNetworkFees();
       this.logger.debug(`Mirror Node returned no fees. Fallback to network`);
```
steven-sheehy, line 154: This log would also occur if the mirror node did return fees successfully

Nana-EC: Missed that. Updated

steven-sheehy (approved):

Issue #231, opened 2022-06-17
Add `erc20.spec.test`


Daniel-K-Ivanov:
Add a second suite in the `acceptance` test folder. The `erc20.spec.test` must utility `ethers`, `wallet` and `provider` constructs fully in order to "simulate" developer interactions with the relay. 

erc20 tests from the OZ repo: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/token/ERC20/ERC20.test.js

- Use the Hedera JS SDK to deploy the contract within the `before` flow
- Use the `servicesClient` to create `aliasAccount`
- Add the tests from the OZ repo and make sure that the described tests pass

Issue #193, opened 2022-06-10
Precompute contract address on ethers deployment


natanasow:
### Description

On contract deployment, most of the tools pre-compute the contract address on the client-side, based on `sender address` and `nonce`. In the Hedera ecosystem, it's not like that, where it's just the next available id. 

[https://docs.ethers.io/v5/api/utils/address/#utils--contract-addresses](
https://docs.ethers.io/v5/api/utils/address/#utils--contract-addresses)
```
ethers.utils.getContractAddress( transaction )
from + nonce
 ```
In the end, the contract deployment is not 100% compatible with the Ethereum workflow.

### Steps to reproduce

Deploy a contract with an ECDSA account and alias.

### Hedera network

other

### Version

v0.1.0

Daniel-K-Ivanov:
This issue cannot be resolved from our side. Unless a change is introduced in ethers, there is no workaround. My suggestion is to **explicitly** state in the developer documentation of the relay that this is one of the **incompatibilities** with existing web3 tooling. That being said, I do think that we might not have this issue with truffle/web3js

Nana-EC:
@Daniel-K-Ivanov let's spend some time on this and see what work around may be possible and what the final solution should be

- Workaround add documentation to the etherjs tool dir with code snippet and workaround e.g [reinitialization](https://github.com/hashgraph/hedera-json-rpc-relay/blob/a66f3df3ba66418151546e310d11d270721ea917/packages/server/tests/acceptance/erc20.spec.ts#L339 )
- Final solution. Explore PR on etherjs or open a local ticket with described fix steps

PR #241, opened 2022-06-20
Add erc20.spec.test

Ivo-Yankov:
**Description**:
Add a spec test that deploys a generic ERC20 contract and performs contract calls on it. Tests a real world use case where a developer would use `ethers.js` to interact with the relay.

Fixes #
https://github.com/hashgraph/hedera-json-rpc-relay/issues/269
https://github.com/hashgraph/hedera-json-rpc-relay/issues/231

**Notes for reviewer**:

**Checklist**


Daniel-K-Ivanov:
General rule of thumb -> if we are using the mirror node endpoint, something is wrong / should be addressed :)

packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -0,0 +1,625 @@
...
+
+
+    const initialSupply = BigNumber.from(10000);
+
+    const createContract = async (constructorArgs:any[] = [], contractJson) => {
+        // contract = await servicesNode.deployContract(erc20Json, 2000000, new Uint8Array(), initialSupply);
+        // const initialBalance = BigNumber.from(new Hbar(100).toTinybars().toString()).mul(10 ** 10);
+
+        const factory = new ethers.ContractFactory(contractJson.abi, contractJson.bytecode, accounts[0].wallet);
+        let contract = await factory.deploy(...constructorArgs, {
```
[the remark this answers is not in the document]
Nana-EC: My current understanding is `gas` and `gasPrice` need to be set for `ethers` right. 
However, if no additional items are needed that'd be great

packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+    };
+
+    const deployErc20 = async (constructorArgs:any[] = [], contractJson) => {
+        const factory = new ethers.ContractFactory(contractJson.abi, contractJson.bytecode, accounts[0].wallet);
+        let contract = await factory.deploy(...constructorArgs);
+
+        await contract.deployed();
+
+        // FIXME mirror node calls should not be made
+        const contractResult = await mirrorNode.get(`/contracts/results/${contract.deployTransaction.hash}`);
```
Ivo-Yankov, line 292: When deploying through `ethers` the pre-deployed address is saved in the contract object. That address is actually incorrect, since it is not yet submitted to the consensus node. Unfortunatelly the `.deployed` method does not update the address, so the contract has to be re-initiated with the correct one.

packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -0,0 +1,295 @@
...
+
+                        beforeEach(async function () {
+                            amount = initialSupply;
+                            to = ethers.constants.AddressZero;
+                            tokenOwnerWallet = accounts[0].wallet;
+                            await contract.connect(tokenOwnerWallet).approve(spender, amount);
+                        });
+
+                        it('reverts', async function () {
+                            await expectRevert(contract.connect(spenderWallet).transferFrom(tokenOwner, to, amount),
```
Daniel-K-Ivanov, line 264: Ideally we should make the revert messages be parsed correctly by client libraries. Logged it as a separate issue:
https://github.com/hashgraph/hedera-json-rpc-relay/issues/279

packages/server/tests/acceptance/erc20.spec.ts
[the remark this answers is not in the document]
Daniel-K-Ivanov: if `gas` and `gasPrice` are not set, the libraries does:
`eth_gasPrice` and `eth_estimateGas` RPC calls under the hood and populates those prior to signing the transaction

packages/server/tests/acceptance/erc20.spec.ts
Daniel-K-Ivanov: yes. it is logged as #193

Daniel-K-Ivanov (approved):

packages/server/tests/acceptance/index.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+        // stop relay
+        logger.info('Stop relay');
+        if (relayServer !== undefined) {
+            relayServer.close();
+        }
+    });
+
+    describe("Acceptance tests", async () => {
+
+        // require('./erc20.spec');
```
Nana-EC, line 110: are these comments or leftover commented out code. If the latter please remove

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -19,710 +19,605 @@
...
+                    expect(res).to.eq('0x0');
+                });
+
+                it('should return 0x for account evm_address on eth_getCode', async function () {
+                    const evmAddress = Utils.idToEvmAddress(accounts[2].accountId.toString());
+                    const res = await relay.call('eth_getCode', [evmAddress]);
+                    expect(res).to.eq('0x0');
+                });
+
+                it('should return 0x for account alias on eth_getCode', async function () {
```
Nana-EC, line 615:
```suggestion
                it('should return 0x0 for account alias on eth_getCode', async function () {
```

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -19,710 +19,605 @@
...
+                    const res = await relay.call('eth_getCode', [evmAddress]);
+                    expect(res).to.eq(basicContractJson.deployedBytecode);
+                });
+
+                it('should return 0x for non-existing contract on eth_getCode', async function () {
+                    const res = await relay.call('eth_getCode', [NON_EXISTING_ADDRESS]);
+                    expect(res).to.eq('0x0');
+                });
+
+                it('should return 0x for account evm_address on eth_getCode', async function () {
```
Nana-EC, line 609:
```suggestion
                it('should return 0x0 for account evm_address on eth_getCode', async function () {
```

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -19,710 +19,605 @@
...
+                    expect(res).to.eq(basicContractJson.deployedBytecode);
+                });
+
+                it('should execute "eth_getCode" for contract with id converted to evm_address', async function () {
+                    const evmAddress = Utils.idToEvmAddress(basicContract.contractId.toString());
+                    const res = await relay.call('eth_getCode', [evmAddress]);
+                    expect(res).to.eq(basicContractJson.deployedBytecode);
+                });
+
+                it('should return 0x for non-existing contract on eth_getCode', async function () {
```
Nana-EC, line 604:
```suggestion
                it('should return 0x0 for non-existing contract on eth_getCode', async function () {
```

packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -0,0 +1,295 @@
...
+                    describe('when the recipient is not the zero address', function () {
+                        let to, toWallet;
+                        before(async function () {
+                            to = anotherAccount;
+                            toWallet = accounts[2].wallet;
+                        });
+
+                        describe('when the spender has enough allowance', function () {
+                            before(async function () {
+                                await contract.approve(spender, initialSupply, { from: initialHolder });
```
Nana-EC, line 139: Q: should the approve calls be verified?
What's the expected response since this is a feature we provide

Nana-EC (dismissed):
LG
Non blocking suggestions.
Fine to address them in followup PR if there's one.

Nana-EC:
@Ivo-Yankov please run this against a remote env also. Would like it to pass against remote envs also.
If there's non compatibility please open tickets to help address those gaps with priority over current acceptance tests tickets

packages/server/tests/acceptance/erc20.spec.ts
Nana-EC: I believe an event is emitted, you can verify that

Ivo-Yankov: Contract calls return a transaction result. I don't think we should make a full assertion for that in those tests.

An Approval event is emitted and tested [here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/241/files#diff-31050a1f9515fb6f9d6e5d3b684b0b5aaa81a6e2c7d1c2e7b2552195d79d067dR169)

Daniel-K-Ivanov (approved):

PR #254, opened 2022-06-22
Add support for empty ENV variables on deployment

rustyShacklefurd:
**Description**:

Without this change, when attempting to deploy a helm chart with an empty value for `OPERATOR_ID_ETH_SENDRAWTRANSACTION`,`OPERATOR_KEY_ETH_SENDRAWTRANSACTION`, or `CHAIN_ID` resulted in this error 
```
Error: UPGRADE FAILED: error validating "": error validating data: [unknown object type "nil" in Secret.data.CHAIN_ID, unknown object type "nil" in Secret.data.OPERATOR_ID_ETH_SENDRAWTRANSACTION, unknown object type "nil" in Secret.data.OPERATOR_KEY_ETH_SENDRAWTRANSACTION]
helm.go:84: [debug] error validating "": error validating data: [unknown object type "nil" in Secret.data.CHAIN_ID, unknown object type "nil" in Secret.data.OPERATOR_ID_ETH_SENDRAWTRANSACTION, unknown object type "nil" in Secret.data.OPERATOR_KEY_ETH_SENDRAWTRANSACTION]
```

This change corrects that error by setting a default of `""` but using the printf %q to ensure the default is safely escaped.  "%q a double-quoted string safely escaped" from [https://helm.sh/docs/chart_template_guide/function_list/#printf]

### DIFF
--- a/helm-chart/templates/secret.yaml
+++ b/helm-chart/templates/secret.yaml
@@ -8,9 +8,9 @@ data:
   HEDERA_NETWORK: {{ .Values.config.HEDERA_NETWORK | b64enc }}
   OPERATOR_ID_MAIN: {{ .Values.config.OPERATOR_ID_MAIN | b64enc }}
   OPERATOR_KEY_MAIN: {{ .Values.config.OPERATOR_KEY_MAIN | b64enc }}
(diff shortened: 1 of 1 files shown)


Nana-EC (approved):
LG

Issue #255, opened 2022-06-23
Add eth_feeHistory and eth_gasPrice acceptance tests


natanasow:
### Problem

Currently there is no test for `eth_feeHistory` and `eth_gasPrice`.

### Solution

- Test out the different combinations of params
```typescript
feeHistory(blockCount: number, newestBlock: string, rewardPercentiles: Array<number>|null)
```
Params are not implemented and can not be tested [click](https://github.com/hashgraph/hedera-json-rpc-relay/blob/81de6a6bf3d36c5e37623f85e310a71a6317f241/packages/relay/src/lib/mirrorNode.ts#L58).

PR #257, opened 2022-06-23
Add eth_feeHistory and eth_gasPrice acceptance tests

natanasow:

**Description**:

**Related issue(s)**:

Fixes #

- https://github.com/hashgraph/hedera-json-rpc-relay/issues/255

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -668,6 +668,19 @@ describe('RPC Server Acceptance Tests', function () {
...
     });
 
+    describe('eth_feeHistory', () => {
+        it('should call eth_feeHistory', async function() {
+            const res = await relay.call('eth_feeHistory', []);
+
+            expect(res.baseFeePerGasArray).to.exist.to.be.an('Array');
+            expect(res.gasUsedRatioArray).to.exist.to.be.an('Array');
+            expect(res.oldestBlockNumber).to.exist;
+            expect(res.baseFeePerGasArray[0]).to.equal('0xa7a3582000');
```
Daniel-K-Ivanov, line 678: - What is this hardcoded value? I think that it is generally better to represent the value in decimal and hex it using `ethers.util.hexValue`
- Is this value something that is hardcoded / used within the relay? if it is, lets use the value from the relay (maybe config property?) because if we change the value in the relay, the tests will start to fail

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -668,6 +668,19 @@ describe('RPC Server Acceptance Tests', function () {
         });
     });
 
+    describe('eth_feeHistory', () => {
```
Daniel-K-Ivanov, line 671: Nit: Lets rename this to `Gas Price related RPC endpoints` and include the acceptance tests for `eth_gasPrice` as-well :)

Daniel-K-Ivanov (changes requested):

Daniel-K-Ivanov (approved):

PR #278, opened 2022-06-28
Update feeHistory exchangeRate logic to use MirrorNode

### LINKED ISSUE #85: Update feeHistory exchangeRate logic to use MirrorNode
### Problem

Currently the `feeHistory` call uses the mocked `mirrorNode.getFeeHistory()`
This includes mocked logic that shouldn't be relied on to get the most recent block

### Solution

Update this to use a valid implementation 
Given our fees don't change often result should be as follows
```
{
  baseFeePerGas: Array of all the fees.servicedata.gas for each record in weibars
  gasUsedRatio: Array of n - m  values of 0.5 i.e. [0.5, 0.5, 0.5 ...]
  oldestBlock: n
}
```

Current the `mirrorNode` file has similar logic. Adopt this and move into `eth.ts`

### Alternatives

_No response_

ar-conmit:
Signed-off-by: Anton Rusev <anton.rusev.code@gmail.com>

**Description**:
- Add `getFeeHistory()` to `eth.ts` to fetch all the fees
- Add constant var for maximum results returned in a single request

**Related issue(s)**:

Fixes #85 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/constants.ts
```diff
@@ -38,4 +38,6 @@ export default {
 
     TYPE_CONTRACT: 'contract',
     TYPE_ACCOUNT: 'account',
+
+    MAX_FEE_HISTORY_RESULTS: 1024
```
Nana-EC, line 42: Let's make this number smaller for now. 
Say 10 since fees don't change often they are likely to be all the same.
Also given the rate of blocks produced i don't even think this number would capture the full cycle of fee updates even if it was done every hour.

Add an environment config for this so we can modify on the fly if needed

packages/relay/src/lib/eth.ts
```diff
@@ -144,14 +150,79 @@ export class EthImpl implements Eth {
...
+    // get latest block fee
+    let nextBaseFeePerGas = _.last(feeHistory.baseFeePerGas);
+
+    if (latestBlockNumber > newestBlockNumber) {
+      // get next block fee if the newest block is not the latest
+      nextBaseFeePerGas = await this.getFeeByBlockNumber(newestBlockNumber + 1);
+    }
+    
+    if (nextBaseFeePerGas) {
+      feeHistory.baseFeePerGas.push(nextBaseFeePerGas);
```
Nana-EC, line 211: will this result in the length of `baseFeePerGas` and `gasUsedRatio` being different?
Should `gasUsedRatio` be updated also?

Nana-EC (commented):
Looking good.
Early comments

packages/relay/src/lib/eth.ts
ar-conmit: Yes, the length of both arrays is different by 1 element. The `baseFeePerGas` array always includes the next block after the newest but `gasUsedRatio` does not.

natanasow (commented):
LG so far

natanasow (approved):

packages/relay/src/lib/eth.ts
Nana-EC: Okay

packages/relay/src/lib/eth.ts
```diff
@@ -54,7 +54,9 @@ export class EthImpl implements Eth {
   static gasTxBaseCost = EthImpl.numberTo0x(21_000);
   static ethTxType = 'EthereumTransaction';
   static ethEmptyTrie = '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421';
-
+  static defaultGasUsedRatio = EthImpl.numberTo0x(0.5);
+  static feeHistoryZeroBlockCountResponse = {gasUsedRatio:null,oldestBlock:EthImpl.zeroHex};
```
Nana-EC, line 58:
```suggestion
  static feeHistoryZeroBlockCountResponse = {gasUsedRatio:null, oldestBlock:EthImpl.zeroHex};
```

packages/relay/src/lib/eth.ts
```diff
@@ -54,7 +54,9 @@ export class EthImpl implements Eth {
...
+  static feeHistoryEmptyResponse = {baseFeePerGas:[],gasUsedRatio:[],reward:[],oldestBlock:EthImpl.zeroHex};
```
Nana-EC, line 59:
```suggestion
  static feeHistoryEmptyResponse = {baseFeePerGas:[], gasUsedRatio:[], reward:[], oldestBlock:EthImpl.zeroHex};
```

Nana-EC (approved):
LG

Issue #129, opened 2022-06-01
Add chain_id precheck to eth_sendRawTransaction


Nana-EC:
### Problem

Currently `eth_sendRawTransaction` does not verify that the `chain_id` matches the `chain_id` of the targeted consensus node

### Solution

Verify that the provided `chain_id` matches the `chain_id` of the configured consensus node. If no, return an error `-32000, "ChainId not supported"`

PR #288, opened 2022-07-01
ChainId precheck

Ivo-Yankov:
**Description**:
Adds a precheck for `chainId` in `eth_sendRawTransaction`. If the `chainId` in the submitted transaction is different than the one specified in the relay environment an error is thrown: `-32000: ChainId not supported`

**Related issue(s)**:

Fixes #:
https://github.com/hashgraph/hedera-json-rpc-relay/issues/129

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/errors.ts
+++ b/packages/relay/src/lib/errors.ts
@@ -70,5 +70,10 @@ export const predefined = {
     name: 'Incorrect nonce',
     code: -32006,
     message: 'Incorrect nonce'
+  }),
+  'UNSUPPORTED_CHAIN_ID': new JsonRpcError({
+    name: 'ChainId not supported',
+    code: -32000,
+    message: 'ChainId not supported'
   })
 };
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -555,15 +555,10 @@ export class EthImpl implements Eth {
     this.logger.trace('sendRawTransaction(transaction=%s)', transaction);
     await this.precheck.nonce(transaction);
 
-    this.logger.trace('Performing chainId precheck for sendRawTransaction(transaction=%s)', transaction);
-
-    const precheckRes = this.precheck.chainId(transaction);
-    this.logger.debug(precheckRes.toString());
-    if ( !precheckRes ) {
-      this.logger.trace('FAILED chainId precheck for sendRawTransaction(transaction=%s)', transaction);
+    if ( !this.precheck.chainId(transaction) ) {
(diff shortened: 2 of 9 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -555,6 +555,11 @@ export class EthImpl implements Eth {
     this.logger.trace('sendRawTransaction(transaction=%s)', transaction);
     await this.precheck.nonce(transaction);
 
+    if ( !this.precheck.chainId(transaction) ) {
+      this.logger.trace('Failed chainId precheck for sendRawTransaction(transaction=%s)', transaction);
```
Nana-EC, line 559: Move this to the `precheck` and log the incorrect chainId for reference

packages/relay/src/lib/errors.ts
```diff
@@ -70,5 +70,10 @@ export const predefined = {
     name: 'Incorrect nonce',
     code: -32006,
     message: 'Incorrect nonce'
+  }),
+  'UNSUPPORTED_CHAIN_ID': new JsonRpcError({
+    name: 'ChainId not supported',
+    code: -32000,
+    message: 'ChainId not supported'
```
Nana-EC, line 77: Q: do you think we should note the incorrect and correct chainId in the message.
Non blocking comment we can do later if it makes sense.
For now so far as we get some documentation with the right chainId on the repo it should be fine

Nana-EC: Can you add this simple table below [here](https://github.com/hashgraph/hedera-json-rpc-relay#configuration-1). This will provide the reference I noted above
```
The following table highlights some initial configuration values to consider
| Config      | Default | Description           |
| ------------|---------|-----------------------|
| `CHAIN_ID`  | '0x12a' | The netowrk chain id. Local and previewnet envs should use `0x12a`. Mainnet and Testnet should use 0x127 and `0x128` respectively |
```

Nana-EC (changes requested):
Looking good, please move the log as commented for better troubleshooting experiences and add the ref table in the README

packages/relay/src/lib/errors.ts
Ivo-Yankov: Yes, that makes sense. It would be really helpful to anyone trying to use the relay to see what the correct chainId is in the error message.

Nana-EC (dismissed):
LG

natanasow (dismissed):
LG

natanasow (approved):

Issue #140, opened 2022-06-02
Add acceptance test support for getLogs


Nana-EC:
### Problem

#103 added support for getLogs.
However , this preceded the acceptance tests so support was not added

### Solution

Add a tests that calls `eth_getLogs` with coverage for a combination of the following params
- fromBlock
- toBlock
- address
- topics
- blockhash

Daniel-K-Ivanov:
- If nothing is passed in the filter (filter is empty) f.e:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "method": "eth_getLogs",
  "params": [
    {}
  ]
}
```
The response must be all logs from the most recent block
- If only `fromBlock` is specified, we must query all logs from the beginning (specified block) until now for all contracts.
- If only `toBlock` is specified, we must query all blocks until the `toBlock`
- Add tests with 1, 2, 3, and 4 topics filter -> topics must be considered OR filter not AND
- If we are passing only `address` -> we must return all logs for that address (address can be account or contract). If no events are found the response should be:

```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": []
}
```

Kalina-Todorova:
We can use this contract to test all options:
`// SPDX-License-Identifier: GPL-3.0
pragma solidity ^0.8.0;

contract Logs {

    event Log0(uint256 num1) anonymous; // Does not include topic
    event Log1(uint256 indexed num0);
    event Log2(uint256 indexed num0, uint256 indexed num1);
    event Log3(uint256 indexed num0, uint256 indexed num1, uint256 indexed num2);
    event Log4(uint256 indexed num0, uint256 indexed num1, uint256 indexed num2, uint256 num3);

    function log0(uint n) public {
        emit Log0(n);
    }

    function log1(uint n) public {
        emit Log1(n);
    }

    function log2(uint n0, uint n1) public {
        emit Log2(n0, n1);
    }

    function log3(uint n0, uint n1, uint n2) public {
        emit Log3(n0, n1, n2);
    }

    function log4(uint n0, uint n1, uint n2, uint n3) public {
        emit Log4(n0, n1, n2, n3);
    }
}`

PR #292, opened 2022-07-04
Add acceptance test support for eth_getLogs

natanasow:

**Description**:

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/140

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -96,6 +97,125 @@ describe('RPC Server Acceptance Tests', function () {
...
+                expect(logs.length).to.be.greaterThan(0);
+
+                log0Block = await relay.call('eth_getTransactionByHash', [log0.contractExecutedTransactionId]);
+                expect(log0Block).to.have.property('blockNumber');
+
+                log4Block = await relay.call('eth_getTransactionByHash', [log4.contractExecutedTransactionId]);
+                expect(log4Block).to.have.property('blockNumber');
+            });
+
+            it('should be abe to use `fromBlock` param', async () => {
```
Nana-EC, line 136:
```suggestion
            it('should be able to use `fromBlock` param', async () => {
```

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -96,6 +97,125 @@ describe('RPC Server Acceptance Tests', function () {
...
+                }]);
+                expect(logs.length).to.be.greaterThan(0);
+
+                const log4BlockInt = parseInt(log4Block.blockNumber);
+                for(let i in logs) {
+                    expect(logs[i].blockNumber).to.be.greaterThanOrEqual(log4BlockInt);
+                }
+            });
+
+            it('should be abe to use `toBlock` param', async () => {
```
Nana-EC, line 148:
```suggestion
            it('should be able to use `toBlock` param', async () => {
```

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -96,6 +97,125 @@ describe('RPC Server Acceptance Tests', function () {
...
+                const log4 = await accounts[1].client.executeContractCall(logsContract.contractId, 'log4', params);
+
+                await new Promise(r => setTimeout(r, 5000));
+
+                const logs = await relay.call('eth_getLogs', [{}]);
+                expect(logs.length).to.be.greaterThan(0);
+
+                log0Block = await relay.call('eth_getTransactionByHash', [log0.contractExecutedTransactionId]);
+                expect(log0Block).to.have.property('blockNumber');
+
```
Nana-EC, line 131: `log.address` and `log.logIndex` should also be checked for now.
Seems like we also have the blockNumber and blockHash so we should verify those here and below where possible also.
We can expand further in future

Nana-EC (changes requested):
LG
Some suggestions

packages/server/tests/acceptance/rpc.spec.ts
Nana-EC: I think you covered logIndex but not sure you covered the value of `address`

Nana-EC (approved):

PR #293, opened 2022-07-05
Add balance precheck to eth_sendRawtransaction

### LINKED ISSUE #130: Add balance precheck to eth_sendRawtransaction
### Problem

Currently the `eth_sendRawTransaction` does not verify  the `sender` has balance >= `value` specified in the transaction body

### Solution

Verify if the sender does not have enough balance compared to the `value` specified for transfer, return an error `insufficient funds for transfer`

### Alternatives

_No response_

ar-conmit:
**Description**:
* Add balance precheck
* Add optional chaining to `getTransactionByHash` for null value of `contractResult.to` causing "multiple done() method invoked" error in acceptance tests
* Refactor `jsonRpcError` assertion to accept predefined error object

**Related issue(s)**:

Fixes #130 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance/rpc.spec.ts
+++ b/packages/server/tests/acceptance/rpc.spec.ts
@@ -351,6 +351,19 @@ describe('RPC Server Acceptance Tests', function () {
                     expect(balanceChange.toString()).to.eq(ONE_TINYBAR.toString());
                 });
 
+                it('should fail "eth_sendRawTransaction" for legacy EIP 155 transactions (with insufficient balance)', async function () {
+                    const balanceInWeiBars = await servicesNode.getAccountBalanceInWeiBars(accounts[2].accountId)
+
+                    const transaction = {
+                        ...default155TransactionData,
+                        to: mirrorContract.evm_address,
+                        value: balanceInWeiBars,
+                        nonce: await relay.getAccountNonce(accounts[2].address)
+                    };
+                    const signedTx = await accounts[2].wallet.signTransaction(transaction);
+                    await relay.callFailing('eth_sendRawTransaction', [signedTx], -32000, 'Insufficient funds for transfer');
+                });
+
                 it('should fail "eth_sendRawTransaction" for Legacy transactions (with no chainId)', async function () {
                     const transaction = {
                         ...defaultLegacyTransactionData,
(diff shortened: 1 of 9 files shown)


packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -351,6 +351,19 @@ describe('RPC Server Acceptance Tests', function () {
...
+                    const balanceInWeiBars = await servicesNode.getAccountBalanceInWeiBars(accounts[2].accountId)
+
+                    const transaction = {
+                        ...default155TransactionData,
+                        to: mirrorContract.evm_address,
+                        value: balanceInWeiBars,
+                        nonce: await relay.getAccountNonce(accounts[2].address)
+                    };
+                    const signedTx = await accounts[2].wallet.signTransaction(transaction);
+                    await relay.callFailing('eth_sendRawTransaction', [signedTx], -32000, 'Insufficient funds for transfer');
```
Nana-EC, line 364: nit: Should pull code and error message from `errors.ts` itself to avoid breaking test case on changes

packages/server/tests/clients/servicesClient.ts
```diff
@@ -275,6 +275,13 @@ export default class ServicesClient {
         const receipt = await response.getReceipt(this.client);
         this.logger.info(`File ${fileId} updated with status: ${receipt.status.toString()}`);
     }
+
+    async getAccountBalanceInWeiBars(account: string | AccountId) {
+        const balance = await this.executeQuery(new AccountBalanceQuery()
+            .setAccountId(account));
+
+        return ethers.BigNumber.from(balance.hbars.toTinybars().toString()).mul(10 ** 10);
```
Nana-EC, line 283: set multiplier as static

packages/server/tests/clients/servicesClient.ts
```diff
@@ -275,6 +275,13 @@ export default class ServicesClient {
         const receipt = await response.getReceipt(this.client);
         this.logger.info(`File ${fileId} updated with status: ${receipt.status.toString()}`);
     }
+
+    async getAccountBalanceInWeiBars(account: string | AccountId) {
+        const balance = await this.executeQuery(new AccountBalanceQuery()
```
Nana-EC, line 280: Pull out  `AccountBalanceQuery` execution into a separate method as it's used multiple times in this class
Will need to update it to handle `AccountId` also
```suggestion
        const balance = await this.getAccountBalance(account)
```

e.g.
```
    async getAccountBalance(account: string | AccountId): Promise<AccountBalance> {
        return this.executeQuery(new AccountBalanceQuery()
            .setAccountId(AccountId.fromString(account)), this.clientMain);
    }
```

packages/server/tests/acceptance/rpc.spec.ts
Nana-EC: Here and below tests

Nana-EC (changes requested):
LG, minor suggestions then should be good to go

Nana-EC (approved):
LG

PR #297, opened 2022-07-06
Improve logging, metrics and error handling

### LINKED ISSUE #272: Add account_id to rpc_relay_operator_balance metric
### Problem

The `rpc_relay_operator_balance` metric can be configured to run on different accounts. Due to a key problem for example the testnet is running from a different account than it normally is.

We use the `rpc_relay_operator_balance` to monitor the balance of the account, so that it doesn't run to 0 without anyone noticing. It'd be great if we had this exposed as a label in the prometheus metric as well so that we can easily tell which account must be funded.

### Solution

Add account_id to the `rpc_relay_operator_balance` metric data

### Alternatives

Asking Nana politely which account we are using :)

Nana-EC:

**Description**:
Improve logging metrics and error handling
- Update balance metric to show operator id
- Fix some logging issues
- Add api callerName to sdkClient metrics to allow for cost per API analysis
- Fix issues around contract results revert and empty results scenarios from mirror node

**Related issue(s)**:

Fixes #272 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -39,7 +39,6 @@ import {
     Query,
     Transaction,
     TransactionRecord,
-    TransactionReceipt,
     Status
 } from '@hashgraph/sdk';
 import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -74,12 +74,14 @@ export class SDKClient {
 
     private consensusNodeClientHistorgram;
     private operatorAccountGauge;
+    private operatorAccountId;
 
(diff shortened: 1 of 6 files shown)


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -105,67 +106,67 @@ export class SDKClient {
                     const accountBalance = await (new AccountBalanceQuery()
                         .setAccountId(clientMain.operatorAccountId!))
                         .execute(clientMain);
```
ar-conmit, line 108: nit: class method `getAccountBalance` can be used here.

ar-conmit (approved):
Looking good

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC: Thanks will address this in follow up

PR #298, opened 2022-07-06
HTS acceptance tests

### LINKED ISSUE #280: Extend/change the `erc20.spec.ts` to `tokens.spec.ts` to include `HTS` tokens
Currently we are executing acceptance tests for an ERC20 deployed contract in Hedera SCS. We need to extend those tests to work with HTS tokens as well. As HTS tokens will be compatible with the ERC20 interface, we can do the following to avoid code duplication:

```TypeScript

describe('Token tests', async () => {

    let tokens [];
    before() {
          tokens.push(await deployErc20());
          tokens.push(await createHTS());
    }

    tokens.forEach(tokenAddress, i) => {
         tokenTests(tokenAddress);
    })
    
    tokenTests(tokenAddress: string) => {
    
       it(...);
       it(...);
       it(...);
    
    }

})

```

This test will formalise / fully test that HTS tokens are ERC20 compatible.
Keep in mind that we might have issues with expecting events, because currently the `event emitter` of the `transfer` event for HTS tokens is the `HTS precompile 0x167` address -> https://github.com/hashgraph/hedera-services/issues/3530

Ivo-Yankov:
**Description**:

**Related issue(s)**:

Fixes #280 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance/erc20.spec.ts
+++ b/packages/server/tests/acceptance/erc20.spec.ts
@@ -25,28 +25,30 @@ chai.use(solidity);
 
 import { AliasAccount } from '../clients/servicesClient';
 import { ethers, BigNumber } from 'ethers';
-
 import ERC20MockJson from '../contracts/ERC20Mock.json';
-import ERC20DecimalsMockJson from '../contracts/ERC20DecimalsMock.json';
-
 import Assertions from '../helpers/assertions';
 import {Utils} from '../helpers/utils';
 
 
-describe('ERC20 Acceptance Tests', function () {
+describe('ERC20 Acceptance Tests', async function () {
     this.timeout(240 * 1000); // 240 seconds
     const {servicesNode, relay} = global;
 
     // cached entities
     const accounts: AliasAccount[] = [];
-    let contract;
     let initialHolder;
     let recipient;
     let anotherAccount;
+    const contracts:[any] = [];
 
     const name = Utils.randomString(10);
     const symbol = Utils.randomString(5);
     const initialSupply = BigNumber.from(10000);
+
+    const testTitles = [
+        'ERC20 Contract',
+        'HTS token'
+    ];
 
     before(async () => {
         accounts[0] = await servicesNode.createAliasAccount(30, relay.provider);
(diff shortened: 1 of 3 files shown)


packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -57,193 +62,258 @@ describe('ERC20 Acceptance Tests', function () {
...
-                                    await expect(tx)
-                                        .to.emit(contract, 'Transfer')
-                                        .withArgs(tokenOwnerWallet.address, toWallet.address, amount);
                                 });
 
-                                it('emits an approval event', async function () {
-                                    await expect(tx)
-                                        .to.emit(contract, 'Approval')
-                                        .withArgs(tokenOwnerWallet.address, spenderWallet.address, await contract.allowance(tokenOwner, spender));
+                                // FIXME
```
Ivo-Yankov, line 156: There is an issue with the emitted events. Should be resolved with services v0.27.0

packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -57,193 +62,258 @@ describe('ERC20 Acceptance Tests', function () {
...
 
-                            describe('when the token owner does not have enough balance', function () {
-                                it('reverts', async function () {
-                                    await contract.transfer(to, 1, { from: tokenOwner });
-                                    await expectRevert(
-                                        contract.connect(spenderWallet).transferFrom(tokenOwner, to, initialSupply),
-                                        'CALL_EXCEPTION'
-                                    );
+                            // FIXME
+                            if (testTitles[i].testName !== HTS) {
```
Ivo-Yankov, line 175: There is an issue with the allowance method. Should be resolved with services v0.27.0

packages/server/tests/acceptance/erc20.spec.ts
Nana-EC: Please open a ticket to track.
0.27 should be available for local-node and previewnet at least.

packages/server/tests/clients/servicesClient.ts
```diff
@@ -289,6 +292,72 @@ export default class ServicesClient {
 
         return ethers.BigNumber.from(balance.hbars.toTinybars().toString()).mul(ServicesClient.TINYBAR_TO_WEIBAR_COEF);
     }
+
+    async createHTS( args = {
+        tokenName: 'Default Name',
+        symbol: 'HTS',
+        treasuryAccountId: '0.0.2',
```
Nana-EC, line 299: Make this a separate account newly created with the operator public key - allowing for future modifications if needed.
In mirror node acceptance tests we eventually hit a state where a treasury account had too many tokens associated. 
Given we want these tests to run against deployed envs also we should add this support.

Nana-EC (changes requested):
Looks good, treasury account needs to be changed

packages/server/tests/acceptance/erc20.spec.ts
Ivo-Yankov: I opened an [issue](https://github.com/hashgraph/hedera-json-rpc-relay/issues/331)
An image for 0.27 is available, but it seems to be build with a wrong `chainId: 0x0127`, which is the value for Testnet.

Nana-EC: Hmm, thanks

Nana-EC (approved):
LG

PR #299, opened 2022-07-06
Helm chart chain id fix

### LINKED ISSUE #300: Helm deployments don't handle integer CHAIN_ID values
### Description

Hex numbers are often converted (or attempted to be converted) and this makes the code return 0xNaN as the CHAIN_ID.

### Steps to reproduce

Set CHAIN_ID to e.g. `298` instead of `0x12a` (both are supported by relay)

### Additional context

_No response_

### Hedera network

other

### Version

v0.2.0

### Operating system

_No response_

rustyShacklefurd:
**Description**:
Cleanup the way CHAIN_ID is set and transported into the k8s deployments.  Hex numbers are often converted (or attempted to be converted) and this makes the code return 0xNaN as the CHAIN_ID.  Moving away from setting this in a k8s secret because there is less conversion/encoding.

Fixes #300

### DIFF
--- a/helm-chart/templates/secret.yaml
+++ b/helm-chart/templates/secret.yaml
@@ -10,7 +10,6 @@ data:
   OPERATOR_KEY_MAIN: {{ .Values.config.OPERATOR_KEY_MAIN | b64enc }}
   OPERATOR_ID_ETH_SENDRAWTRANSACTION: {{ .Values.config.OPERATOR_ID_ETH_SENDRAWTRANSACTION | default (printf "%q" "") }}
   OPERATOR_KEY_ETH_SENDRAWTRANSACTION: {{ .Values.config.OPERATOR_KEY_ETH_SENDRAWTRANSACTION | default (printf "%q" "") }}
-  CHAIN_ID: {{ printf "%s" .Values.config.CHAIN_ID | default (printf "%q" "") |b64enc }}
   MIRROR_NODE_URL: {{ .Values.config.MIRROR_NODE_URL | b64enc }}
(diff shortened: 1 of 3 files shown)


Nana-EC (approved):
LG

helm-chart/templates/secret.yaml
```diff
@@ -10,7 +10,6 @@ data:
   OPERATOR_KEY_MAIN: {{ .Values.config.OPERATOR_KEY_MAIN | b64enc }}
   OPERATOR_ID_ETH_SENDRAWTRANSACTION: {{ .Values.config.OPERATOR_ID_ETH_SENDRAWTRANSACTION | default (printf "%q" "") }}
   OPERATOR_KEY_ETH_SENDRAWTRANSACTION: {{ .Values.config.OPERATOR_KEY_ETH_SENDRAWTRANSACTION | default (printf "%q" "") }}
```
steven-sheehy, line 12: Why are these operator properties not base64 encoded and are just escaped? Better to use `stringData` instead of `data` to avoid all the extra base64 conversion anyway.

Nana-EC: Agreed. @rustyShacklefurd feel free to open a separate PR that addresses this properly.
This satisfies the work around

Nana-EC: Added #301 to capture this

rustyShacklefurd: good suggestion.  I will put this change in new branch and get it PR'd

PR #304, opened 2022-07-07
Implement gasLimit Precheck

### LINKED ISSUE #128: Add gasLimit precheck to eth_sendRawTransaction
### Problem

Currently the `eth_sendRawTransaction` does not performa. precheck to verify the `gasLimit`

### Solution

`gasLimit` must be `intrinsic_gas <= specified_gas_limit <= block gas limit`, where `intrinsic_gas` is computed by this [algorithm](https://github.com/hashgraph/hedera-services/blob/23f0548ed5aaa466242f9262466a0fa68c095f6b/hedera-node/src/main/java/com/hedera/services/contracts/gascalculator/GasCalculatorHederaV22.java#L48) and `block gas limit` is hardcoded to `15_000_000` gas. If `specified_gas_limit > block_gas_limit`, return an error `-32005, "Transaction gas limit exceeds block gas limit"`; If `specified_gas_limit < intrinsic_gas_cost`, return an error `-32003, "Intrinsic gas exceeds gas limit"` 

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
Implements gasLimit over/under checks before submitting in sendRawTransactions

**Related issue(s)**:

Fixes #128 

**Notes for reviewer**:

**Checklist**


Nana-EC (approved):
LG
Hopefully this addresses the DApp test timeouts also

packages/relay/src/lib/precheck.ts
```diff
@@ -116,11 +114,55 @@ export class Precheck {
...
+
+    const dataBytes = Buffer.from(data, "hex");
+
+    for (const c of dataBytes) {
+      if (c == 0) {
+        zeros++;
+      }
+    }
+
+    const nonZeros = data.replace('0x', '').length - zeros;
```
Ivo-Yankov, line 164: This was causing a wrong calculation of the gas cost. If `0x` is not removed from `data`, that counts as 2 `nonZero` bytes, and the calculated gas cost is `TX_BASE_COST + 2 * ISTANBUL_TX_DATA_NON_ZERO_COST`, which meant that transactions with empty data (hbar transfer transactions) could not pass the precheck if the default gasLimit (21000) was used.

Nana-EC (approved):
LG

Issue #216, opened 2022-06-16
Add cache logic to capture static INVALID entity transaction failures


Nana-EC:
### Problem

Some wallets will reissue the same calls over and over again for extended periods when they receive errors.
In many cases the responses are static (e.g. INVALID_CONTRACT_ID and INVALID_ACCOUNT_ID) and will not change over time or with successive or other transactions

### Solution

This can be optimized by capturing calls and their error response in a cache (LRU) that returns the static response if called again.
Care should be taken to only do this for static responses whosE success is not impacted by future calls

PR #310, opened 2022-07-08
Add cache logic to capture static INVALID entity transaction failures

natanasow:

**Description**:

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/216

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/tests/lib/eth.spec.ts
+++ b/packages/relay/tests/lib/eth.spec.ts
@@ -893,6 +893,61 @@ describe('Eth calls using MirrorNode', async function () {
     expect(result).to.equal(null);
   });
 
+  describe('eth_getBalance', async function() {
+    it('should return cached value', async () => {
+      mock.onGet(`blocks?limit=1&order=desc`).reply(200, {
+        blocks: [{
+          number: 10000
+        }]
+      });
+      mock.onGet(`contracts/0x000000000000000000000000000000000000055f`).reply(200, null);
+      mock.onGet(`accounts/0x000000000000000000000000000000000000055f`).reply(200, {
+        account: '0x000000000000000000000000000000000000055f'
+      })
+      sdkClientStub.getAccountBalanceInWeiBar.throws({
+        status: {
+          _code: 15
+        }
+      });
+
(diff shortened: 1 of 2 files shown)


Nana-EC (changes requested):
Looking good.
Missing tests to show that when cached the mirror node or sdk client are not again called.
Using their mocks should support this check

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -893,6 +893,61 @@ describe('Eth calls using MirrorNode', async function () {
...
+      })
+      sdkClientStub.getAccountBalanceInWeiBar.throws({
+        status: {
+          _code: 15
+        }
+      });
+
+      const resNoCache = await ethImpl.getBalance(contractAddress1, null);
+      const resCached = await ethImpl.getBalance(contractAddress1, null);
+      expect(resNoCache).to.equal('0x0');
```
Nana-EC, line 915:
```suggestion
      expect(resNoCache).to.equal(EthImpl.zeroHex);
```

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -893,6 +893,61 @@ describe('Eth calls using MirrorNode', async function () {
...
+      mock.onGet(`blocks?limit=1&order=desc`).reply(200, {
+        blocks: [{
+          number: 10000
+        }]
+      });
+      mock.onGet(`contracts/0x000000000000000000000000000000000000055f`).reply(200, null);
+      mock.onGet(`accounts/0x000000000000000000000000000000000000055f`).reply(200, {
+        account: '0x000000000000000000000000000000000000055f'
+      })
+      sdkClientStub.getAccountBalanceInWeiBar.throws({
```
Nana-EC, line 907: Can you add a check that ensures the stub for `getAccountBalanceInWeiBar` is only called once.
This verifies the cache was picked up

Nana-EC (changes requested):
LG
Still missing the `getCode` tests in `eth.spec.ts` but also lots of room for reuse

Nana-EC (approved):
LG

PR #315, opened 2022-07-08
Fix schema mismatch for nums in eth_getTransactionReceipt response

### LINKED ISSUE #302: `eth_getTransactionReceipt` invalid response: hex string instead of uint
### Description

When performing `eth_getTransactionReceipt` the response is not valid. According to https://github.com/ethereum/execution-apis/blob/main/src/schemas/receipt.yaml#L50-L101 the `blockNumber`, `cumulativeGasUsed`, `gasUsed`, `transactionIndex`, `effectiveGasPrice` and `status` properties are `uint`, the relay response uses hex strings instead. 

A json-rpc client with client-side validation, such as go-ethereum, throws validation exceptions in such cases.

### Steps to reproduce

```sh
curl -X POST $BASE_URL \
                                       -d "{\"id\":\"2\",\"jsonrpc\":\"2.0\",\"method\":\"eth_getTransactionReceipt\",\"params\":[\"$TRANSACTION_HASH\"]}" --silent
```
`BASE_URL=http://localhost:7546`
`TRANSACTION_HASH=0x9040d4244cfeb7c5fc96ac378308d9df8a6616c550e9d02162b3ef3fe889df86`

### Additional context

image

image

### Hedera network

testnet

### Version

main

### Operating system

Linux

Nana-EC:

**Description**:
Fix schema mismatch for nums in eth_getTransactionReceipt response

**Related issue(s)**:

Fixes #302 

**Notes for reviewer**:

**Checklist**


Ivo-Yankov:
There seems to be a lot of inconsistency with the json rpc implementations. Even if we change the format of those fields, other clients may expect them to be in hex. 
For example the [Alchemy's web interface](https://dashboard.alchemyapi.io/composer) returns those values in hex.

Since anyone will be able to host their own instance of the relay, I suggest we add an `.env` setting for the format of specific fields. This will add a lot of overhead, but it will make the relay more flexible.

natanasow (approved):
LG

Nana-EC:
> There seems to be a lot of inconsistency with the json rpc implementations. Even if we change the format of those fields, other clients may expect them to be in hex. For example the [Alchemy's web interface](https://dashboard.alchemyapi.io/composer) returns those values in hex.
> 
> Since anyone will be able to host their own instance of the relay, I suggest we add an `.env` setting for the format of specific fields. This will add a lot of overhead, but it will make the relay more flexible.

Yes agreed, this makes it tricky for us.
Right now we're trying to play it safe by following what the execution apis dictate, but i do think exploring some configureable option either on demand or by relay might be useful in the future

PR #332, opened 2022-07-12
Align receipt with execution apis

### LINKED ISSUE #302: `eth_getTransactionReceipt` invalid response: hex string instead of uint

shemnon:
**Description**:
Move uints back to hex strings

**Related issue(s)**:

Fixes #302

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/model.ts
+++ b/packages/relay/src/lib/model.ts
@@ -20,6 +20,7 @@
 
 // Used for fake implementation of block history
 import {Status, TransactionRecord} from "@hashgraph/sdk";
+import {EthImpl} from "./eth";
 
 export class Block {
     public readonly timestamp:string = '0x' + new Date().valueOf().toString(16);
@@ -44,7 +45,7 @@ export class Block {
     public readonly totalDifficulty:string = '0x1';
     public readonly transactions:string[] | Transaction[] = [];
     public readonly transactionsRoot:string = '0x0';
-    public readonly uncles:any[] = [];
+    public readonly uncles:string[] = [];
 
     constructor(args?:any) {
         if (args) {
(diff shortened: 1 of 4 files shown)


packages/relay/src/lib/model.ts
```diff
@@ -103,36 +104,36 @@ export class CachedBlock extends Block {
...
         // TODO this.to = record.contractFunctionResult?.contractId;
-        this.cumulativeGasUsed = Number(gasUsed);
-        this.gasUsed = Number(gasUsed);
+        this.cumulativeGasUsed = gasUsed;
+        this.gasUsed = gasUsed;
         this.contractAddress = contractAddress;
         this.logs = [];
         this.logsBloom = '';
-        this.status = record.receipt.status == Status.Success ? 1 : 0;
+        this.status = record.receipt.status == Status.Success ? "0x1" : "0x0";
```
Nana-EC, line 136: Also move `0x1` to a static class variable
```suggestion
        this.status = record.receipt.status == Status.Success ? "0x1" : EthImpl.zeroHex;
```

packages/relay/src/lib/model.ts
```diff
@@ -20,6 +20,7 @@
 
 // Used for fake implementation of block history
 import {Status, TransactionRecord} from "@hashgraph/sdk";
+import {EthImpl} from "./eth";
```
Nana-EC, line 23: Hmm, seems weird to be importing this just for a formatting method.
Eventually (not a fix for now) I think we should move the converter functions out to the `formatter.ts` class

Nana-EC (changes requested):
Looking good.
Some suggestions also i think you're missing test update.
Can see [my PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/315/files#diff-febf1674c0e8b4be0c0f626009660fb57c4160a97d7be3f97b80e38a93aa5016) that introduced the change so it's easy to see what to revert

packages/relay/src/lib/model.ts
```diff
@@ -103,36 +104,36 @@ export class CachedBlock extends Block {
...
 
     constructor(txHash:string, record:TransactionRecord, block:Block) {
-        const gasUsed = record.contractFunctionResult == null ? 0 : record.contractFunctionResult.gasUsed;
+        const gasUsed = EthImpl.numberTo0x(record.contractFunctionResult == null ? 0 : record.contractFunctionResult.gasUsed.toNumber());
         const contractAddress = record.contractFunctionResult == undefined ? undefined : "0x" + record.contractFunctionResult.contractId?.toSolidityAddress();
 
         this.transactionHash = txHash;
-        this.transactionIndex = 0;
-        this.blockNumber = Number(block.number);
+        this.transactionIndex = "0x0";
```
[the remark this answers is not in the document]
shemnon: I opted for a direct rollback on this line.

packages/relay/src/lib/model.ts
shemnon: I opted for a direct rollback on this line.

Nana-EC: Gotcha. `0x0` and `0x1` should still be static class constants

Nana-EC: NO biggie. We can clean this and other files up later

Nana-EC (commented):
LG.
"should execute "eth_getTransactionReceipt" for hash..."  acceptance tests need updating with your new logic

Nana-EC (approved):

Issue #349, opened 2022-07-19
Optimize parse transaction on prechecks


natanasow:
### Problem

On each precheck, we parse the transaction instead of parsing it once and pass it to all checks parsed.

### Solution

- Parse the tx in the sendRawTransaction method
- Pass the parsed tx to the prechecks

PR #350, opened 2022-07-19
Optimize parse transaction on prechecks

natanasow:

**Description**:

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/349

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -28,7 +28,6 @@ import { MirrorNodeClient, SDKClient } from './clients';
 import { JsonRpcError, predefined } from './errors';
 import constants from './constants';
 import { Precheck } from './precheck';
-import { ethers } from 'ethers';
 
 const _ = require('lodash');
 const cache = require('js-cache');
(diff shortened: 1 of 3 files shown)


Ivo-Yankov (dismissed):
Looks good

packages/relay/src/lib/eth.ts
```diff
@@ -646,23 +647,24 @@ export class EthImpl implements Eth {
   async sendRawTransaction(transaction: string): Promise<string | JsonRpcError> {
     this.logger.trace('sendRawTransaction(transaction=%s)', transaction);
 
+    const parsedTx = ethers.utils.parseTransaction(transaction);
```
Nana-EC, line 650: Actually I would prefer an approach where `ethers` stays in the `precheck` file.
Instead could we update `precheck` with helper methods for each of the calling `eth_x` methods that encompasses the needed prechecks.

Example design below

```
  async sendRawTransactionCheck(transaction: string) {
    const parsedTx = Precheck.parseTxIfNeeded(transaction);
    this.gasLimit(parsedTx);
    await this.nonce(parsedTx);
    this.chainId(parsedTx);
    this.value(parsedTx);
  }
```

packages/relay/src/lib/eth.ts
```diff
@@ -646,23 +647,24 @@ export class EthImpl implements Eth {
...
     try {
-      await this.precheck.gasLimit(transaction);
-      await this.precheck.nonce(transaction);
-      this.precheck.chainId(transaction);
-      this.precheck.value(transaction);
+      this.precheck.gasLimit(parsedTx);
```
Nana-EC, line 652: Based on the above this would simplify to 
```suggestion
      this.precheck.sendRawTransactionCheck(transaction);
```

packages/relay/src/lib/precheck.ts
```diff
@@ -38,8 +38,14 @@ export class Precheck {
     this.logger = logger;
   }
 
-  value(transaction: string) {
-    const tx = ethers.utils.parseTransaction(transaction);
+  private static parseTxIfNeeded(transaction: string | ethers.Transaction): ethers.Transaction {
+    return typeof transaction === 'string'
```
Nana-EC, line 42: Not sure if string type is needed with use of helper methods.
Just make all prechecks take in `Transaction` unless that make testing more difficult.

Nana-EC (changes requested):
Looking good.
Suggestion to have logic encompassed in `precheck` file instead

PR #353, opened 2022-07-19
readme updated with new instructions for installing the helm chart

rustyShacklefurd:
Readme update to use the helm repo instead of the existing helm-charts directory.  Helm repo allows for multiple versions to be stored.


Nana-EC (approved):
LG

shemnon (approved):

PR #361, opened 2022-07-20
Complete eth_getStorageAt implementation

### LINKED ISSUE #355: Complete `eth_getStorageAt` implementation
### Problem

Current the `eth_getStorageAt` endpoint exists but is not correctly filled out for the relays capability.
We should return some kind of response

### Solution

Given the mirror node support endpoints don't yet exist we should return unsupported method

- Uncomment `getStorageAt` in `index.ts`
- Implement `eth.getStorageAt` w expected params as noted in [](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getstorageat) and pass to 
  - add trace log
  - return static response of `UNSUPPORTED_METHOD` (See other unsupported methods for examples)
- Update `eth_getStorageAt` in `server.ts` to call `eth.getStorageAt`. 
  - Remove log
- Update `eth.spec.ts` to verify unsupported response

### Alternatives

_No response_

lukelee-sl:
**Description**:
Implement the eth_getStorageAt call as an unsupported method.

**Related issue(s)**:

Fixes #355 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -427,6 +427,11 @@ export class EthImpl implements Eth {
     return predefined.UNSUPPORTED_METHOD;
   }
 
+  getStorageAt(_address: string, _slot: string, _blockNumber: string | null): JsonRpcError {
```
Nana-EC, line 430: nit: we don't tend to use the underscored variables.
No biggie just FYI

packages/relay/src/lib/eth.ts
```diff
@@ -427,6 +427,11 @@ export class EthImpl implements Eth {
...
+    this.logger.trace('getStorageAt()');
```
Nana-EC, line 431: nit: could log the incoming params.

Nana-EC (dismissed):
Nice first PR. 
2 non-blocking nits

Nana-EC (approved):
LGTM

lukelee-sl:
@Nana-EC one more approval please.  For whatever reason, the email used when signing my last 2 commit was incorrect.  I've addressed the issue and resubmitted.

Nana-EC (approved):

PR #375, opened 2022-07-22
Support 'from' in 'eth_call'

### LINKED ISSUE #373: `eth_call` does not respect the `from` field all the time
### Description

From @shemnon 
_When calling an `eth_call` with a `from` field the call returns as though the relay is doing the call.  When sending the query to the consensus node we should 

- set the senderAccountId to the corresponding from field
    - if it is missing or invalid, set senderAccountId to all zeros._

### Steps to reproduce

1. Deploy the contract:
```
pragma solidity >= 0.5.0 < 0.9.0;

contract CoinCollector {
  mapping(address => uint) private balances;
    address public owner;

    event LogDeposit(address indexed accountAddress, uint amount);
    event LogWithDraw(address indexed accountAddress, uint amount);

  function deposit() public payable {
    balances[msg.sender] += msg.value;
        emit LogDeposit(msg.sender, msg.value);
  }

  function balance() public view returns(uint) {
    return balances[msg.sender];
  }
}
```

2. Call deposit() with some hbar (logged results below)
```
[
  {
    "from": "0x0000000000000000000000000000000002D84A2E",
    "topic": "0x1b851e1031ef35a238e6c67d0c7991162390df915f70eaf9098dbf0b175a6198",
    "event": "LogDeposit",
    "args": {
      "0": "0xC2A3C271e2D133ea12ea2BC747fA0780da134136",
      "1": "1000000000",
      "accountAddress": "0xC2A3C271e2D133ea12ea2BC747fA0780da134136",
      "amount": "1000000000"
    }
  }
]
```
3. Call balance() (logged results below)
```
{
	"0": "uint256: 0"
}
```

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.27

### Operating system

macOS

shemnon:
**Description**:
When a from value is in the eth_call params, set it to the senderId.

**Related issue(s)**:

Fixes #373 

**Notes for reviewer**:

**Checklist**


packages/relay/tests/lib/eth.spec.ts
```diff
@@ -1318,6 +1319,77 @@ describe('Eth calls using MirrorNode', async function () {
...
+  describe('eth_call', async function () {
+    it('eth_call with no gas', async function () {
+      sdkClientStub.submitContractCallQuery.returns({
+            asBytes: function () {
+              return Uint8Array.of(0)
+            }
+          }
+      );
+
+      const result = await ethImpl.call({
```
Nana-EC, line 1332: For completeness of coverage this methods should be checking that the result is expected.
Given the return above i think the expected result would be `0x0` or `0x`.

Should make this change for the 4 tests added.

packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -356,8 +358,6 @@ describe('ERC20 Acceptance Tests', async function () {
         // Setup initial balance of token owner account
         await servicesNode.transferHTSToken(accounts[0].accountId, htsResult.receipt.tokenId, initialSupply, htsResult.client);
         const evmAddress = Utils.idToEvmAddress(htsResult.receipt.tokenId.toString());
-        const contract = new ethers.Contract(evmAddress, abi, accounts[0].wallet);
-
-        return contract;
+        return new ethers.Contract(evmAddress, abi, accounts[0].wallet);
     };
 });
```
Nana-EC, line 363: nit: new line since you're here.

Nana-EC (changes requested):
LG.
Just expand the coverage on the added `eth.spec.ts`

Daniel-K-Ivanov (approved):
LGTM! We might want to resolve @Nana-EC 's comments and then merge :)

packages/relay/tests/lib/eth.spec.ts
shemnon: Done

packages/server/tests/acceptance/erc20.spec.ts
shemnon: Done

Nana-EC (approved):
LGTM

PR #419, opened 2022-08-04
return 0x instead of 0x0 for non existing contract addresses given to…

### LINKED ISSUE #394: eth_getCode returning 0x0 instead of 0x for non-contract addresses
### Description

When making a request using the eth_getCode method for a non-contract address, the result value returned is 0x0. Other relays (such as Ethereum or BSC relays) return 0x. This can cause issues if applications or libraries are expecting 0x but not 0x0. 

It might be worth modifying the result value to return 0x to cover applications or libraries that do not expect 0x0, especially if this is standard for relays on other chains.

### Steps to reproduce

1. Make a request to the JSON RPC relay with the eth_getCode method for a non-contract address

`curl -X POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":"2","method":"eth_getCode","params":["0x0000000000000000000000000000000000000000", "latest"]}' https://testnet.hashio.io/api`

Response received:
`{"result":"0x0","jsonrpc":"2.0","id":"2"}`

Response expected:
`{"result":"0x","jsonrpc":"2.0","id":"2"}`

### Additional context

_No response_

### Hedera network

### Version

v0.3.0

### Operating system

macOS

lukelee-sl:
… eth_getCode

**Description**:
In order to be more consistent with other relays, return 0x instead of 0x0 when sending a non existing contract address to the eth_getCode call.

Modified code in the error condition when we are given an invalid contract.

**Related issue(s)**:

Fixes #394 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -494,8 +494,8 @@ export class EthImpl implements Eth {
         // handle INVALID_CONTRACT_ID
         if (e.isInvalidContractId()) {
           this.logger.debug('Unable to find code for contract %s in block "%s", returning 0x0, err code: %s', address, blockNumber, e.statusCode);
-          cache.set(cachedLabel, '0x', constants.CACHE_TTL.ONE_HOUR);
-          return '0x';
+          cache.set(cachedLabel, EthImpl.emptyHex, constants.CACHE_TTL.ONE_HOUR);
+          return EthImpl.emptyHex;
         }
         this.logger.error(e, 'Error raised during getCode for address %s, err code: %s', address, e.statusCode);
       } else {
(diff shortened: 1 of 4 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -494,8 +494,8 @@ export class EthImpl implements Eth {
         // handle INVALID_CONTRACT_ID
         if (e.isInvalidContractId()) {
           this.logger.debug('Unable to find code for contract %s in block "%s", returning 0x0, err code: %s', address, blockNumber, e.statusCode);
-          cache.set(cachedLabel, '0x0', constants.CACHE_TTL.ONE_HOUR);
-          return '0x0';
+          cache.set(cachedLabel, '0x', constants.CACHE_TTL.ONE_HOUR);
+          return '0x';
```
Nana-EC, line 498: nit:
```suggestion
          return EthImpl.emptyHex;
```

Nana-EC: Here and anywhere else it's easily available in your changes

Nana-EC (changes requested):
LG, minor nit

Nana-EC (approved):
LGTM

PR #424, opened 2022-08-05
Add acceptance test support for hts precompile

### LINKED ISSUE #398: Add acceptance test support for HederaTokenService precompile
### Problem

Currently there's no test support for the `IHederaTokenService.sol` for HTS precompile

### Solution

Add support for acceptance tests be
- Add contract implementation [HederaTokenService.sol](https://github.com/hashgraph/hedera-services/blob/master/test-clients/src/main/resource/contract/solidity/hip-206/HederaTokenService.sol) to branch in test section
- Add new htsPrecompile.spec.ts file
- Add logic to deploy contract and fund 2 accounts

This should serve as a basis that other PRs will utilize to test the hts precompile functions

### Alternatives

_No response_

natanasow:

**Description**:

**Related issue(s)**:

Fixes #398

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance/htsPrecompile.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+describe('HTS Precompile Acceptance Tests', async function() {
+  this.timeout(240 * 1000); // 240 seconds
+  const { servicesNode, relay } = global;
+
+  const accounts: AliasAccount[] = [];
+  let baseHTSContract;
+  let HTSTokenContract;
+
+  before(async () => {
+    accounts[0] = await servicesNode.createAliasAccount(1000, relay.provider);
```
Ivo-Yankov, line 42: nit: Could you lower the initial balance of the alias accounts. Somewhere in the range of 10-30 hbars.

Ivo-Yankov (changes requested):
Looking good, just a small nit.

packages/server/tests/acceptance/htsPrecompile.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+  it('should create a HTS token', async function() {
+    const tx = await baseHTSContract.createToken(accounts[0].wallet.address, {
+      value: ethers.BigNumber.from('20000000000000000000'),
+      gasLimit: 1000000
+    });
+    const { tokenAddress } = (await tx.wait()).events.filter(e => e.event = 'CreatedToken')[0].args;
+
+    expect(tokenAddress).to.not.be.null;
+
+    HTSTokenContract = new ethers.Contract(tokenAddress, ERC20MockJson.abi, accounts[0].wallet);
```
Nana-EC, line 67: Q: does this need to be done in the `before()`?

packages/server/tests/acceptance/htsPrecompile.spec.ts
Nana-EC: Yep, as low as possible to pass in remote envs like previewnet/testnet

Nana-EC (commented):
LG
Waiting on reduction in initial hbar balance funded to test accounts.

packages/server/tests/acceptance/htsPrecompile.spec.ts
natanasow: As long as they are async calls we can't initialize them before `describe` and in the `describe` body. So I think that the best place to create these accounts is in `before` hook.

Nana-EC: The accounts is fine, i'm referring to the Contract deployment.
That should be in the before().
Turn this into a normal function (not a test) and call it in before()

Same logic is used in `erc20.spec.ts` and `rpc.spec.ts` i.e. create main contract in before/beforeAll.
In this way future tests are assured it's present and no tests are run if it fails

Nana-EC (changes requested):
LG. Just move the contract deployments (`baseHTSContract` and `HTSTokenContract` into the before to assure no tests run if it fails and that tests have it available prior to run

packages/server/tests/acceptance/htsPrecompile.spec.ts
natanasow: Oh, got it. :pray:

Nana-EC (approved):
LGTM

Ivo-Yankov (approved):

PR #473, opened 2022-08-25
add trace id to error response messages

### LINKED ISSUE #431: Pass back logs request id to caller
### Problem

While adding trace id to logs, a discussion was started in this PR: https://github.com/hashgraph/hedera-json-rpc-relay/pull/423 about sending back the trace id to the caller without breaking the schema. 
One explored option was to pass it through the headers, but the caller receives the response directly so the id won't be received in the response.

### Solution

Find a way to pass back the request id to the caller in so that they can easily do the correlation of request to stack trace.

### Alternatives

_No response_

lukelee-sl:

**Description**:
Add trace id to error response messages.  
There is no obvious place to return the trace id when the call is successful without changing the schema.

**Related issue(s)**:

Fixes #431 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/src/server.ts
+++ b/packages/server/src/server.ts
@@ -146,22 +146,25 @@ const logAndHandleResponse = async (methodName, methodFunction) => {
     ms = Date.now() - start;
     methodResponseHistogram.labels(methodName, status).observe(ms);
     logger.info(`${messagePrefix} ${status} ${ms} ms `);
-    return response;
+    if (response instanceof JsonRpcError) {
+      return new JsonRpcError({name: response.name, code: response.code, message:`${requestIdPrefix} ` + response.message});
+    } return response;
   } catch (e: any) {
     ms = Date.now() - start;
     methodResponseHistogram.labels(methodName, responseInternalErrorCode).observe(ms);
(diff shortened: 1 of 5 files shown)


packages/server/src/server.ts
```diff
@@ -146,22 +146,26 @@ const logAndHandleResponse = async (methodName, methodFunction) => {
...
+        error = predefined.REQUEST_TIMEOUT;
       }
     }
     else if (e instanceof JsonRpcError) {
-      return e;
+      error = e;
     }
 
-    return predefined.INTERNAL_ERROR;
+    return new JsonRpcError({name: error.name, code: error.code, message:`${requestIdPrefix} ` + error.message});
```
Nana-EC, line 168: Q: Would it have been easier to just add a param value to `predefined.INTERNAL_ERROR` that takes a requestId?

Nana-EC, line 168: Add a `JsonRpcError` constructor that centralizes this logic

Nana-EC (changes requested):
Some suggestions.

Also please add some logs to the PR comment to show case where it looks like

Nana-EC (approved):
LGTM
Add an example of the new error response format in the PR comments

PR #485, opened 2022-08-31
add acceptance tests for cryptoTransfer function

lukelee-sl:

**Description**:
https://github.com/hashgraph/hedera-services/issues/3758

The issue referenced above in hedera-services describes several scenarios where using the cryptoTransfer function fails.  

Currently, cryptoTransfer succeeds only when the token type being transferred is of homogenous (fungible -> fungible or nft -> nft).  Also there is no support for sending hbars as one of the legs of the transfer.

Added several acceptance tests to show use cases where the function succeeds and also when it fails due to heterogeneous token types. 

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/issues/3758

**Notes for reviewer**:

**Checklist**


lukelee-sl:
Here are some results after some exploration.

- In services code there is a checks that ensures that fungibles transfers add up to zero.
- The solidity `cryptoTransfer` function takes `IHederaTokenService.TokenTransferList[]` as an argument.
    - `TokenTransferList` is a struct that contains both and array of `AccountAmount` (for fungible tokens) and an array of `NftTransfer` (for NFT).  A user could take this to mean that a transfer of heterogeneous types of tokens is possible but services code seems to indicate that this is not possible.
- In my test - homogeneous transfers using `crytoTransfer` worked 
- but heterogeneous transfers failed (as expected).
- I attempted to **swap** NFT tokens in two separate accounts in a single call and was not successful.
- Also attempted to transfer tokens from a wallet after it was given an allowance but a separate account and was not successful.
- Possible future feature improvements
    - Allow for heterogeneous token type `cryptoTransfer` 
    - Allow HBARs to be transferred as part of the `cryptoTransfer`

packages/server/tests/contracts/BaseHTS.sol
```diff
@@ -140,6 +140,22 @@ contract BaseHTS is FeeHelper {
         }
     }
 
+    function realTransferTokenPublic(address token, address sender, address receiver, int64 amount) public returns (int responseCode) {
```
Nana-EC, line 143: I'd actually rename as below so it's explicit.
`transferTokenPublic` -> `cryptoTransferTokenPublic`
`realTransferTokenPublic ` -> `transferTokenPublic`

packages/server/tests/acceptance/htsPrecompile.spec.ts
```diff
@@ -604,4 +604,162 @@ describe('HTS Precompile Acceptance Tests', async function () {
...
+            amount: -10,
+          },
+        ],
+        nftTransfers: [],
+      }];
+      const txXfer = await baseHTSContract.cryptoTransferPublic(tokenTransferList);
+      expect((await txXfer.wait()).events.filter(e => e.event === 'ResponseCode')[0].args.responseCode).to.equal(TX_SUCCESS_CODE);
+    });
+
+    xit('should be able to swap non-fungible tokens', async function() {
```
Nana-EC, line 710:
```suggestion
    xit('should be able to swap approved non-fungible tokens', async function() {        
```

Nana-EC, line 710: Dupe this test for a FT approved case let's see.

Nana-EC, line 710: Let it run and check for failure like you do in the other.
Then when it's fixed we'll know to just update
```suggestion
    it('should be able to swap non-fungible tokens', async function() {        
```

Nana-EC (changes requested):
Some suggestions

stoqnkpL:
There are a couple of problems with the setup:

1. The HTS precompiled contract function `cryptoTransfer(TokenTransferList[] memory tokenTransfers)` does not support transfers using allowances, it was never designed to. Currently, the only way to make a transfer using allowances through the precompile is to use the redirect functions `transferFrom`:
`IERC20(token).transferFrom(sender, recipient, amount);`
`IERC721(token).transferFrom(from, to, tokenId);`

I'm not sure why the transferFrom functions were not exposed to HTS precompile top-level like the other functions from ERC20 and ERC721 were with [PR3638](https://github.com/hashgraph/hedera-services/pull/3638). In any case, the cryptoTransfer function will not work with allowances.

2. Transfers of HBars via `cryptoTransfer` is also not supported, only transfers of Fungible tokens and NFTs.

For context, we have existing EETs covering the transfer of tokens (including mixing fungible and nfts) via cryptoTransfer [here](https://github.com/hashgraph/hedera-services/blob/master/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/CryptoTransferHTSSuite.java).
We have existing EETs covering the transferFrom ERC redirect functions [here](https://github.com/hashgraph/hedera-services/blob/master/test-clients/src/main/java/com/hedera/services/bdd/suites/contract/precompile/ERCPrecompileSuite.java).

packages/server/tests/acceptance/htsPrecompile.spec.ts
```diff
@@ -604,4 +604,208 @@ describe('HTS Precompile Acceptance Tests', async function () {
...
+      try{
+        const txXfer = await baseHTSContract.cryptoTransferPublic(tokenTransferList);
+        expect((await txXfer.wait()).events.filter(e => e.event === 'ResponseCode')[0].args.responseCode).to.equal(TX_SUCCESS_CODE);
+      } catch (error: any) {
+        expect(error.code).to.equal("CALL_EXCEPTION");
+        expect(error.reason).to.equal("transaction failed");
+      }
+    });
+
+    it('should be able to swap approved non-fungible tokens', async function() {
```
Nana-EC, line 750:
```suggestion
    it('should fail to swap approved non-fungible tokens', async function() {        
```

Nana-EC (dismissed):
LG

Nana-EC (approved):

PR #489, opened 2022-08-31
Rolling restarts done via cronjob

rustyShacklefurd:
Signed-off-by: Matt Halder <rustyShacklefurd@users.noreply.github.com>

**Description**:
This PR adds the ability to enable or disable a rolling restart of the hedera-json-rpc-relay deployment

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/helm-chart/environments/minikube.yaml
+++ b/helm-chart/environments/minikube.yaml
@@ -14,5 +14,6 @@ config:
   OPERATOR_KEY_ETH_SENDRAWTRANSACTION: ""
 
 rolling_restart:
-  enabled: true  
+  enabled: true
+  schedule: "0 0 * * *"
 
--- a/helm-chart/environments/minikube.yaml
+++ b/helm-chart/environments/minikube.yaml
@@ -8,10 +8,11 @@ config:
   LOG_LEVEL: ""
   local:
     HEDERA_NETWORK: {"127.0.01:50211":"0.0.3"}
-
-secret:
   OPERATOR_ID_MAIN: ""
   OPERATOR_KEY_MAIN: ""
   OPERATOR_ID_ETH_SENDRAWTRANSACTION: ""
   OPERATOR_KEY_ETH_SENDRAWTRANSACTION: ""
+
+rolling_restart:
+  enabled: true  
 
--- a/helm-chart/templates/cronjob.yaml
+++ b/helm-chart/templates/cronjob.yaml
@@ -22,6 +22,7 @@ spec:
                 - bash
                 - -c
                 - >-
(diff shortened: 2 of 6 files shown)


helm-chart/templates/cronjob.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.rolling_restart.enabled }}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "json-rpc-relay.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  concurrencyPolicy: Forbid
+  schedule: '0 0 * * *'
```
Nana-EC, line 9: Can we go with `@daily` or `@midnight` to make this more readable

helm-chart/values.yaml
```diff
@@ -95,9 +95,14 @@ config:
...
   # hosted:
   #   HEDERA_NETWORK: ""
-secret:
   OPERATOR_ID_MAIN: ""
   OPERATOR_KEY_MAIN: ""
   OPERATOR_ID_ETH_SENDRAWTRANSACTION: ""
-  OPERATOR_KEY_ETH_SENDRAWTRANSACTION: ""  
+  OPERATOR_KEY_ETH_SENDRAWTRANSACTION: ""
+
+# Enable rolling_restarts if SDK calls fail this is usually due to stale connections that get cycled on restart
```
Nana-EC, line 103: Can we change the replicaCount to 2 with this PR?
Or would k8s handle this by spinning one pod up first and moving traffic there before dropping the current pod?

Nana-EC (changes requested):
LG, some clarifying questions

helm-chart/templates/cronjob.yaml
rustyShacklefurd: For this I have variablized the `rolling_restart.schedule` and the default is set to `@daily` in values.yaml

helm-chart/environments/minikube.yaml
```diff
@@ -8,10 +8,12 @@ config:
...
-
-secret:
   OPERATOR_ID_MAIN: ""
   OPERATOR_KEY_MAIN: ""
   OPERATOR_ID_ETH_SENDRAWTRANSACTION: ""
   OPERATOR_KEY_ETH_SENDRAWTRANSACTION: ""
 
+rolling_restart:
+  enabled: true
+  schedule: "0 0 * * *"
```
Nana-EC, line 18: If it lets you
```suggestion
  schedule: "@daily"
```

Nana-EC (approved):
LG

PR #510, opened 2022-09-10
update acceptance tests to use new signatures.  Create new acceptance…

### LINKED ISSUE #509: Add tests to cover the updated type signatures for HederaToken and createToken functions
### Problem

The types of several fields were updated to better align with protobuf and documentation as outlined in these two tasks.
https://github.com/hashgraph/hedera-smart-contracts/issues/56
https://github.com/hashgraph/hedera-services/issues/3902

We need to ensure that the acceptance test exercise both the new types and the old as the old types signatures are still supported in hedera services.

### Solution

Ensure that acceptance tests run correctly with the new type signatures.
Write a new acceptance test that exercise the create token functions with the old  type signatures.

### Alternatives

_No response_

lukelee-sl:
… test to exercise old create token function signatures

**Description**:
The following pull requests updated the type signatures for functions and structures to better match protobuf and documentation.

https://github.com/hashgraph/hedera-smart-contracts/pull/59
https://github.com/hashgraph/hedera-services/pull/3910

Modify the solidity files to reflect the changes in the smart contracts repo.
Ensure that the acceptance tests run as expected with the new type signatures.
Create a new acceptance test to ensure that the old signature also work as expected.

**Related issue(s)**:

Fixes #509 

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance/htsPrecompile_v1.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+
+// external resources
+import { solidity } from 'ethereum-waffle';
+import chai, { expect } from 'chai';
+
+chai.use(solidity);
+
+import { AliasAccount } from '../clients/servicesClient';
+import { ethers } from 'ethers';
+import ERC20MockJson from '../contracts_v1/ERC20Mock.json';
```
Nana-EC, line 29: Let's keep a single main `contract` folder and have sub folders instead
```suggestion
import ERC20MockJson from '../contracts/htsPrecompileV1/ERC20Mock.json';
```

packages/server/tests/acceptance/htsPrecompile_v1.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  this.beforeAll(async () => {
+    accounts[0] = await servicesNode.createAliasAccount(60, relay.provider);
+    accounts[1] = await servicesNode.createAliasAccount(15, relay.provider);
+    accounts[2] = await servicesNode.createAliasAccount(15, relay.provider);
+
+    // alow mirror node a 2 full record stream write windows (2 sec) and a buffer to persist setup details
+    await new Promise(r => setTimeout(r, 5000));
+
+    BaseHTSContractAddress = await deployBaseHTSContract();
```
Nana-EC, line 59: since this is the v1 test class where we're only testing the older stuff we can split these up a bit.
More the appropriate create call into the test method.
This way the test methods are actually testing create and should there be a regression we clearly see which one failed.

Nana-EC (changes requested):
LG
I think we can rearrange the dir structure and tests case setup

packages/server/tests/acceptance/htsPrecompile_v1.spec.ts
```diff
@@ -55,23 +54,11 @@ describe('HTS Precompile V1 Acceptance Tests', async function () {
...
-    BaseHTSContractAddress = await deployBaseHTSContract();
-    HTSTokenContractAddress = await createHTSToken();
-    NftHTSTokenContractAddress = await createNftHTSToken();
-    HTSTokenWithCustomFeesContractAddress = await createHTSTokenWithCustomFees();
-
-    HTSTokenContract = new ethers.Contract(HTSTokenContractAddress, ERC20MockJson.abi, accounts[0].wallet);
-    baseHTSContract = new ethers.Contract(BaseHTSContractAddress, BaseHTSJson.abi, accounts[0].wallet);
-
-    baseHTSContractOwner = baseHTSContract;
-    baseHTSContractReceiverWalletFirst = baseHTSContract.connect(accounts[1].wallet);
```
Nana-EC: Could have left the wallet inits here

packages/server/tests/acceptance/htsPrecompile_v1.spec.ts
```diff
@@ -103,47 +90,72 @@ describe('HTS Precompile V1 Acceptance Tests', async function () {
...
-      gasLimit: 10_000_000
+      gasLimit: 1_000_000
     });
     const txReceipt = await tx.wait();
     const { tokenAddress } = txReceipt.events.filter(e => e.event === 'CreatedToken')[0].args;
 
     return tokenAddress;
   }
 
+  it('should deploy baseHTSContract', async function () {
```
Nana-EC, line 101: I guess what i meant was you could create the the token, instantiate the contract from the address and associate in one test for the 3 scenarios.

Basically 
- `should createHTSToken` & `should associate to a token` would merge
- `should createNftHTSToken` & `should associate to a nft` would merge
- `should createHTSTokenWithCustomFees` & `should associate to a token with custom fees` would merge

`should finsh setting up contracts` shouldn't really be a thing.
The first parts can go into the tests where the Address is used and the wallet creations would be in the beforeAll

Just trying to make things contained but also ensure that it doesn't matter in what order tests are run

Nana-EC (changes requested):
I think we should reorganize to ensure tests are more contained and easy to follow but also not rely on sequential ordering

Nana-EC (approved):
LG

Issue #511, opened 2022-09-12
Set gas limit to 15_000_000


Kalina-Todorova:
### Problem

The gas limit for the Hedera network is currently 15_000_000. 
This came out as a follow up to: https://github.com/hashgraph/hedera-json-rpc-relay/issues/437

### Solution

Cap the gas to 15_000_000 in the relay

PR #513, opened 2022-09-12
Limit the ethCall gas to 15_000_000, otherwise the call will fail

dimitrovmaksim:

**Description**:
I'm not sure if this is the best way to do this, but ethCalls with gas over 15_000_000 will fail with BUSY error, because the Hedera limit is set to 15_000_000. This breaks the integration with TheGraph, when executing `eth_call`s from the event handlers, because the graph-node reserves all 50_000_000 gas.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/511

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -792,7 +792,13 @@ export class EthImpl implements Eth {
           gas = call.gas;
         }
       }
-      
+
+      // Gas limit for `eth_call` is 50_000_000, but the Hedera network limit is 15_000_000
```
Nana-EC, line 796: Assuming it's safe to cap this we can leave it here for now.
Please add a trace log.
Also add a test  to verify the submitted transaction has the gas capped.

Nana-EC (changes requested):
Looking good.
Missing log and test

packages/relay/src/lib/eth.ts
```diff
@@ -792,7 +792,14 @@ export class EthImpl implements Eth {
...
+      // With values over 15_000_000 the call will fail with BUSY error
+      if (gas > 15_000_000) {
+        this.logger.trace(`${requestIdPrefix} eth_call gas amount (${gas}) exceeds network limit, capping gas to 15_000_000`);
+        gas = 15_000_000;
```
Nana-EC, line 800: nit: could have made `15_000_000` a const
Can make this change in a future PR, no biggie

Nana-EC (approved):
LG

packages/relay/src/lib/eth.ts
dimitrovmaksim: I'll still probably have to merge main at some point to fix the dapp tests, I can push that change too

Nana-EC (approved):

Nana-EC:
@dimitrovmaksim, all your commits need to be signed

dimitrovmaksim:
> @dimitrovmaksim, all your commits need to be signed

All of them seem to have the "Signed-off-by:" message

dimitrovmaksim:
Closing in favor of https://github.com/hashgraph/hedera-json-rpc-relay/pull/526

Issue #472, opened 2022-08-25
`eth_getLogs` not handling non int block params


Nana-EC:
### Description

Currently the `eth_getLogs` endpoint parses the `fromBlock` and `toBlock` without consideration for non int values.
This results in the block time frames not being honoured

See [eth.getLogs](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1153) where `parseInt(toBlock)` and `parseInt(fromBlock)` are used.

These should apply the `blockNumberOrTag` logic used by the other methods

### Steps to reproduce

1. `curl -X POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":"2","method":"eth_getLogs","params":[{ "fromBlock": "0x0", "toBlock": "latest", "addresss": "0x0000000000000000000000000000000002dC46Db"}]}' http://localhost:7546`
2. Results in a `blocks?block.number=lte:NaN&block.number=gte:0&order=asc 400 status`

### Hedera network

previewnet

### Version

v0.6.0

Nana-EC:
Fix will like take the form of updating [eth.getLogs()](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1153)
before
```ts
      if (toBlock) {
        // @ts-ignore
        filters.push(`lte:${parseInt(toBlock)}`);
        order = constants.ORDER.DESC;
      }
      if (fromBlock) {
        // @ts-ignore
        filters.push(`gte:${parseInt(fromBlock)}`);
        order = constants.ORDER.ASC;
      }
```

after
```ts
      if (toBlock) {
        const blockNumber = await this.translateBlockTag(toBlock, requestId);
        // @ts-ignore
        filters.push(`lte:${blockNumber}`);
        order = constants.ORDER.DESC;
      }
      if (fromBlock) {
        const blockNumber = await this.translateBlockTag(fromBlock, requestId);
        // @ts-ignore
        filters.push(`gte:${blockNumber}`);
        order = constants.ORDER.ASC;
      }
```

Nana-EC:
> Wouldn't `translateBlockTag` potentially add more requests to the mirror node? I was wondering if we could do it another way. So earliest is `0` this one should be easy, and `latest` and `pending` should be the last block at the time of the query(?)
> 
[...]

PR #518, opened 2022-09-14
getLogs add support for block tags

dimitrovmaksim:

**Description**:

1. Fixes getLogs not being able to handle block tags (latest, earliest)
2. Adds block range limit (default 500). Can be configured with by setting `ETH_GET_LOGS_BLOCK_RANGE_LIMIT` env variable

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/472

**Notes for reviewer**:

Maybe we should set a max block range limit?

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -1146,28 +1146,59 @@ export class EthImpl implements Eth {
...
-        order = constants.ORDER.DESC;
+        try {
+          const blockResponse = await this.getHistoricalBlockResponse(toBlock, true, requestId);
+          toBlockTimestamp = blockResponse.timestamp.to;
+          toBlockNum = parseInt(blockResponse.number);
+        } catch (e) {
+          // Check if block error is RESOURCE_NOT_FOUND, the most likely scenario this will happen if a block number > height is passed
+          // In this case eth_getLogs() returns logs up to the latest block.
+          // @ts-ignore
+          if (e.statusCode != 404) {
```
Nana-EC, line 1163: If 404 throw `predefined.RESOURCE_NOT_FOUND`

packages/relay/src/lib/eth.ts
```diff
@@ -1146,28 +1146,59 @@ export class EthImpl implements Eth {
...
-        order = constants.ORDER.ASC;
+        try {
+          const blockResponse = await this.getHistoricalBlockResponse(fromBlock, undefined, requestId);
+          fromBlockTimestamp = blockResponse.timestamp.from;
+          fromBlockNum = parseInt(blockResponse.number);
+        } catch (e) {
+          // Check if block error is RESOURCE_NOT_FOUND, the most likely scenario this will happen if a block number > height is passed
+          // In this case eth_getLogs() returns an empty array.
+          // @ts-ignore
+          if (e.statusCode != 404) {
```
Nana-EC, line 1177: if 404 use `predefined.RESOURCE_NOT_FOUND`

Nana-EC (changes requested):

packages/relay/src/lib/eth.ts
```diff
@@ -1146,28 +1146,59 @@ export class EthImpl implements Eth {
...
       if (toBlock) {
-        // @ts-ignore
-        filters.push(`lte:${parseInt(toBlock)}`);
-        order = constants.ORDER.DESC;
+        try {
+          const blockResponse = await this.getHistoricalBlockResponse(toBlock, true, requestId);
+          toBlockTimestamp = blockResponse.timestamp.to;
+          toBlockNum = parseInt(blockResponse.number);
+        } catch (e) {
+          // Check if block error is RESOURCE_NOT_FOUND, the most likely scenario this will happen if a block number > height is passed
```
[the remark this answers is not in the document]
dimitrovmaksim: Currently if toBlock number > height it will not set `toBlockTimestamp ` and `lte` timestamp filter, which means it will get logs up to the latest block. Then here https://github.com/hashgraph/hedera-json-rpc-relay/blob/9b845579ee4661eef4054b998f723f40c73f68da/packages/relay/src/lib/eth.ts#L1194-L1195 it fetches the latest block number just to check that the block range is inside of the set limit here https://github.com/hashgraph/hedera-json-rpc-relay/blob/9b845579ee4661eef4054b998f723f40c73f68da/packages/relay/src/lib/eth.ts#L1198-L1203

packages/relay/src/lib/eth.ts
dimitrovmaksim: In this case we should not throw anything if a block is not found

packages/relay/src/lib/eth.ts
dimitrovmaksim: Same as the previous comment

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
Nana-EC: Great that works.

packages/relay/src/lib/eth.ts
Nana-EC: Makes sense

packages/relay/src/lib/eth.ts
Nana-EC: Okay.
Maybe add a comment

packages/relay/src/lib/eth.ts
Nana-EC: Add a comment

packages/relay/src/lib/eth.ts
```diff
@@ -1146,28 +1146,59 @@ export class EthImpl implements Eth {
...
+          // In this case eth_getLogs() returns logs up to the latest block.
+          // @ts-ignore
+          if (e.statusCode != 404) {
+            throw e;
```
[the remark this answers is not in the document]
Nana-EC: Ah correct. Ignore.

dimitrovmaksim:
I'll address the comments and open a new PR, because the commits in this one aren't verified as well.

dimitrovmaksim:
Closing in favour of https://github.com/hashgraph/hedera-json-rpc-relay/pull/532

PR #522, opened 2022-09-14
Helm testdrive cleanup

rustyShacklefurd:
**Description**:
Removed annotations JSON support in favor of annotation [NEG](https://cloud.google.com/kubernetes-engine/docs/how-to/container-native-load-balancing) support.  Removed an duplicate and incorrect CHAIN_ID config in the configmap.

**Notes for reviewer**:
Templated these changes using `helm template test ./helm-chart --debug`

```
---
# Source: hedera-json-rpc-relay/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: integration-json-rpc-relay
  labels:
    app:  integration-json-rpc-relay

    helm.sh/chart: hedera-json-rpc-relay-0.1.1
    app.kubernetes.io/name: integration-json-rpc-relay
    app.kubernetes.io/instance: test
    app.kubernetes.io/version: "0.7.0"
    app.kubernetes.io/managed-by: Helm
data:
  HEDERA_NETWORK: {XXX}
  MIRROR_NODE_URL: http://xx.xx.xx.xx
  LOCAL_NODE: false
  SERVER_PORT: 7546
  CHAIN_ID:

---
# Source: hedera-json-rpc-relay/templates/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: json-rpc-relay
  labels:
    helm.sh/chart: hedera-json-rpc-relay-0.1.1
    app.kubernetes.io/name: son-rpc-relay
    app.kubernetes.io/instance: test
    app.kubernetes.io/version: "0.7.0"
    app.kubernetes.io/managed-by: Helm
# This addes support for an NEG setup where the key is a string but the value is JSON
  annotations:
    cloud.google.com/neg: "{\"exposed_ports\":{ \"7546\":{\"name\": \"json-rpc-relay\"}}}"
    cloud.google.com/neg:: "{\"ingress\": true}"
spec:
  type: ClusterIP
  ports:
    - port: 7546
      targetPort: jsonrpcrelay
      protocol: TCP
      name: jsonrpcrelay
  selector:
    app: json-rpc-relay
    app.kubernetes.io/name: json-rpc-relay
    app.kubernetes.io/instance: test
```
**Checklist**

### DIFF
--- a/helm-chart/Chart.yaml
+++ b/helm-chart/Chart.yaml
@@ -2,6 +2,6 @@ apiVersion: v2
 name: hedera-json-rpc-relay
 description: Helm chart deployment of the hashgraph/hedera-json-rpc-relay
 type: application
-version: 0.8.0-SNAPSHOT
-appVersion: "0.8.0-SNAPSHOT"
+version: 0.1.1
+appVersion: "0.7.0"
 home: https://github.com/hashgraph/hedera-json-rpc-relay
--- a/helm-chart/templates/service.yaml
+++ b/helm-chart/templates/service.yaml
@@ -8,9 +8,12 @@ metadata:
   annotations:
 {{ toYaml .Values.service.annotations | nindent 4 }}
 {{- end }}
-{{- if .Values.service.annontations_JSON }}
+# This addes support for an NEG setup where the key is a string but the value is JSON
+{{- if .Values.service.annotations_NEG }}
   annotations:
-{{ .Values.service.annontations_JSON | toJson | nindent 4 }}
+{{- range  $key, $value := .Values.service.annotations_NEG }}
+    {{ $key }}: {{ toJson $value }}
+{{- end }}      
 {{- end }}  
 spec:
   type: {{ .Values.service.type }}
--- a/helm-chart/templates/configmap.yaml
+++ b/helm-chart/templates/configmap.yaml
@@ -13,7 +13,6 @@ data:
 {{- else }}
   HEDERA_NETWORK: {{ required "A valid HEDERA_NETWORK must be present in either .Values.config.local.HEDERA_NETWORK or .Values.config.hosted.HEDERA_NETWORK " (or .Values.config.local.HEDERA_NETWORK .Values.config.hosted.HEDERA_NETWORK) }}
 {{- end }}
-  CHAIN_ID: {{ .valu}} 
   MIRROR_NODE_URL: {{ .Values.config.MIRROR_NODE_URL  }}
   LOCAL_NODE: {{ .Values.config.LOCAL_NODE }}
   SERVER_PORT: {{ .Values.config.SERVER_PORT }}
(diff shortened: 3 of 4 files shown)


helm-chart/templates/service.yaml
```diff
@@ -8,9 +8,12 @@ metadata:
   annotations:
 {{ toYaml .Values.service.annotations | nindent 4 }}
 {{- end }}
-{{- if .Values.service.annontations_JSON }}
+# This addes support for an NEG setup where the key is a string but the value is JSON
+{{- if .Values.service.annotations_NEG }}
   annotations:
-{{ .Values.service.annontations_JSON | toJson | nindent 4 }}
+{{- range  $key, $value := .Values.service.annotations_NEG }}
```
steven-sheehy, line 14: Why is a separate `annotations_NEG`/`annotations_JSON` necessary? There's no need to escape the JSON, just use double quotes wrapped in single quotes like the mirror node does in [production](https://github.com/hashgraph/hedera-mirror-node/blob/1625a0414c28f0e3b82b0bb3f4765f659000439d/clusters/mainnet-na/common/helmrelease.yaml#L58) to populates its NEG.

helm-chart/Chart.yaml
```diff
@@ -2,6 +2,6 @@ apiVersion: v2
 name: hedera-json-rpc-relay
 description: Helm chart deployment of the hashgraph/hedera-json-rpc-relay
 type: application
-version: 0.8.0-SNAPSHOT
-appVersion: "0.8.0-SNAPSHOT"
+version: 0.1.1
+appVersion: "0.7.0"
```
steven-sheehy, line 6: Something seems wrong with either the PR or with the release automation to have such version skew. I see in the helm repo the last is `0.1.0` with `0.7.0` app. I think the `appVersion=0.8.0-SNAPSHOT` is correct and should be reverted since this is main and `0.7.0` has been released already. 

But the `version=0.8.0-SNAPSHOT` may or may not be correct. Depends if the relay is moving to synchronize appVersion and version like the mirror node did or whether it's possible for the release lifecycle to vary for the two artifacts.

steven-sheehy (changes requested):

helm-chart/Chart.yaml
Nana-EC: version & appVersion should be synchronized. 
`0.8.0-SNAPSHOT` is the right value.

Is this just a bad merge?

rustyShacklefurd: I changed this b/c I didn't see any release of 0.8.0-SNAPSHOT in the release list.

Nana-EC: Ahh you're right.
The SNAPSHOTS refer to the latest main code.
Deployment should use GAd versions under releases

PR #526, opened 2022-09-16
Limit gas to 15m for eth call

dimitrovmaksim:

**Description**:
I'm not sure if this is the best way to do this, but ethCalls with gas over 15_000_000 will fail with BUSY error, because the Hedera limit is set to 15_000_000. This breaks the integration with TheGraph, when executing `eth_call`s from the event handlers, because the graph-node reserves all 50_000_000 gas.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/511

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -483,7 +483,7 @@ export class EthImpl implements Eth {
 
     return result;
   }
-  
+
   /**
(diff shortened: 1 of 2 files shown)


Nana-EC (approved):
LG.
Please simplify the PR comments to simply state the issue and a bullet points of the changes you made

PR #529, opened 2022-09-17
Update eth_getBalance to use mirror node

### LINKED ISSUE #448: Update `eth_getBalance` with logic to call mirror node account endpoint
### Problem

Mirror node added logic tot eh `api/v1/accounts/{accountIdOrEvmAddress}` to return the up to date balance.
However, the relay still calls the consensus node

### Solution

Update `eth_getBalance` with logic to call mirror node account endpoint

### Alternatives

_No response_

Nana-EC:

**Description**:
`eth_getBalance` currently uses the consensus nodes to retrieve balance since mirror node only update balance every 15 mins.
With mirror node fix to update on transfers relay can now point to mirror node

- Update `eth_getBalance` logic to call Mirror Node /api/v1/accounts` endpoint and extract balance
- Fix metric discrepancy on `resolvedCost`
- Update sdkClient flow to remove unnecessary error check since default is UNKNOWN

**Related issue(s)**:

Fixes #448 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -507,25 +507,18 @@ export class EthImpl implements Eth {
 
     try {
       let weibars: BigNumber | number = 0;
-      const result = await this.mirrorNodeClient.resolveEntityType(account, requestId);
-      if (result?.type === constants.TYPE_ACCOUNT) {
-        weibars = await this.sdkClient.getAccountBalanceInWeiBar(result.entity.account, EthImpl.ethGetBalance, requestId);
-      }
-      else if (result?.type === constants.TYPE_CONTRACT) {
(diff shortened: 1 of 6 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -507,25 +507,18 @@ export class EthImpl implements Eth {
```
georgi-l95, line 507: 10 lines above this, there is a comment, which I think can be removed now with this PR.

georgi-l95 (changes requested):
LG, but one acceptance test is failing.
`@release should execute "eth_getBalance" for newly created account with 10 HBAR`

packages/relay/src/lib/eth.ts
Nana-EC: Thanks

georgi-l95 (approved):
LGTM

lukelee-sl (approved):

PR #532, opened 2022-09-19
Support block tags in getLogs()

dimitrovmaksim:

**Description**:

1. Fixes getLogs not being able to handle block tags (latest, earliest)
2. Adds block range limit (default 500). Can be configured with by setting `ETH_GET_LOGS_BLOCK_RANGE_LIMIT` env variable

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/472

**Notes for reviewer**:

Maybe we should set a max block range limit?

**Checklist**


Nana-EC (approved):
LG

PR #549, opened 2022-09-21
eth_getBalance search by blockNumberOrTag

### LINKED ISSUE #531: Update `eth_getBalance` to filter search based on blockNumberOrTag
### Problem

Current implementation does not yet utilize blockNumber

### Solution

Update `eth_getBalance` to filter search based on blockNumberOrTag.

Note that Mirror node balances endpoint still relies on balance file produced every 15 mins.
As such any blockNumberOrTag that occurred in the last 15 mins may be outdated.

At an error for blockNumbers that fall within this range to prevent false responses

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
Implements searching for the balance of an account or contract at a specific block number.

**Related issue(s)**:

Fixes #531 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -504,11 +504,50 @@ export class EthImpl implements Eth {
...
       let weibars: BigNumber | number = 0;
-
       const mirrorAccount = await this.mirrorNodeClient.getAccount(account, requestId);
-      if (mirrorAccount && mirrorAccount.balance) {
-        weibars = mirrorAccount.balance.balance * constants.TINYBAR_TO_WEIBAR_COEF
-      } else {
+      let balanceFound = false;
+
+      // A blockNumberOrTag has been provided. If it is `latest` or `pending` retrieve the balance from /accounts/{account.id}
+      if (mirrorAccount && blockNumberOrTag !== null && blockNumberOrTag !== EthImpl.blockLatest && blockNumberOrTag !== EthImpl.blockPending) {
```
Nana-EC, line 511: Move the blockNumberTag checks to a method.
I think the same check for a valid non latest value exists somewhere else also

Nana-EC (approved):
LG minor improvement suggestion

Nana-EC (approved):
LG

packages/relay/src/lib/eth.ts
```diff
@@ -504,11 +504,50 @@ export class EthImpl implements Eth {
...
+      let balanceFound = false;
+
+      // A blockNumberOrTag has been provided. If it is `latest` or `pending` retrieve the balance from /accounts/{account.id}
+      if (mirrorAccount && !EthImpl.blockTagIsLatestOrPending(blockNumberOrTag)) {
+        const blockNumberHex = EthImpl.numberTo0x(blockNumber);
+        const latestBlockNumber = await this.blockNumber(requestId);
+
+        // If the parsed blockNumber is the same as the one from the latest block retrieve the balance from /accounts/{account.id}
+        if (blockNumberHex !== latestBlockNumber) {
+          const blockPromises = [
```
dimitrovmaksim, line 517: If I understand correctly, currently this code does up to 4 requests for the blocks. Once to eventually translate the tag [here ](https://github.com/hashgraph/hedera-json-rpc-relay/pull/549/files#diff-dd8c7990f7764960d05dcb4c1b5ae7eee48b425b1b83d6dcc7e7ccc442330ef0R497), once to get the latest block number [here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/549/files#diff-dd8c7990f7764960d05dcb4c1b5ae7eee48b425b1b83d6dcc7e7ccc442330ef0R513) and 2 more here, if the block number and the lates block number are the same. Can't we just use `getHistoricalBlockResponse` for both the from he start and skip the previous 2 requests. This way we will have both the block numbers and the blocks data with just 2 requests. Also shouldn't this [cached request](https://github.com/hashgraph/hedera-json-rpc-relay/pull/549/files#diff-dd8c7990f7764960d05dcb4c1b5ae7eee48b425b1b83d6dcc7e7ccc442330ef0R499-R503) be performed before the once to the mirror node?

natanasow (dismissed):

packages/relay/src/lib/eth.ts
```diff
@@ -494,48 +494,50 @@ export class EthImpl implements Eth {
...
-              balanceFound = true;
-              if (balance.balances?.length) {
-                weibars = balance.balances[0].balance * constants.TINYBAR_TO_WEIBAR_COEF;
+      if (!EthImpl.blockTagIsLatestOrPending(blockNumberOrTag)) {
+        const block = await this.getHistoricalBlockResponse(blockNumberOrTag, true, requestId);
+        if (block) {
+          blockNumber = block.number;
+
+          // A blockNumberOrTag has been provided. If it is `latest` or `pending` retrieve the balance from /accounts/{account.id}
+          if (mirrorAccount && !EthImpl.blockTagIsLatestOrPending(blockNumberOrTag)) {
```
Nana-EC, line 516: Isn't this already inferred by the initial check can this be simplified to
```suggestion
          if (mirrorAccount)
```

packages/relay/src/lib/eth.ts
```diff
@@ -494,48 +494,50 @@ export class EthImpl implements Eth {
...
+
+            const latestBlock = await this.getHistoricalBlockResponse(EthImpl.blockLatest, true, requestId);
+
+            // If the parsed blockNumber is the same as the one from the latest block retrieve the balance from /accounts/{account.id}
+            if (latestBlock && block.number !== latestBlock.number) {
```
Nana-EC, line 521: There's no else to the if check using timestamps below.
Since time 
Does the check above by number add value or can it be made to this instead
```suggestion
            if (latestBlock && block.timestamp && latestBlock.timestamp) {
```

Nana-EC (commented):
Some simplification suggestions

packages/relay/src/lib/eth.ts
Ivo-Yankov: Actually the timestamp checks were not neaded. The missing else statement is intentional.

packages/relay/src/lib/eth.ts
```diff
@@ -494,21 +494,58 @@ export class EthImpl implements Eth {
...
+      if (!EthImpl.blockTagIsLatestOrPending(blockNumberOrTag)) {
+        const block = await this.getHistoricalBlockResponse(blockNumberOrTag, true, requestId);
+        if (block) {
+          blockNumber = block.number;
 
-      const mirrorAccount = await this.mirrorNodeClient.getAccount(account, requestId);
-      if (mirrorAccount && mirrorAccount.balance) {
-        weibars = mirrorAccount.balance.balance * constants.TINYBAR_TO_WEIBAR_COEF
-      } else {
+          // A blockNumberOrTag has been provided. If it is `latest` or `pending` retrieve the balance from /accounts/{account.id}
```
Nana-EC, line 515: nit: based on the outer check on like 510 `blockNumberOrTag` should never be latest or pending

Nana-EC (approved):
LG

PR #551, opened 2022-09-21
return the evm address in the from field when calling getTransactionByHash

### LINKED ISSUE #505: ECDSA alias not returned in eth_getTransactionByHash response
### Description

A user created a pre-signed transaction to deploy a smart contract using the hedera-local-node setup. The transaction was signed using the ECDSA private key for account 0.0.1012 and the signerAddress was the alias address for that account (`0x67D8d32E9Bf1a9968a5ff53B87d777Aa8EBBEe69`).

The transaction executed successfully and the relay returned the transaction hash in the response. When querying the relay using eth_getTransactionByHash, the "from" value in the response is the hex encoded Hedera account id (`0x00000000000000000000000000000000000003f4`), not the alias address for the account (`0x67D8d32E9Bf1a9968a5ff53B87d777Aa8EBBEe69`).

After looking at HIP-482, I believe it should return the alias address. This is also what the user expected. Is this the expected behaviour and I'm misunderstanding how it works, or is it a bug?

### Steps to reproduce

1. Run the hedera-local-node setup using the CLI tool to generate accounts
2. Submit the pre-signed transaction using the relay

**Request**
`curl -X POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":"1","method":"eth_sendRawTransaction","params":["0xf8a78085a7a35820008302e6308080b853604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3820278a015a5e6900dadf4a75d924aba5ba296df568ac3c119395c22d2d3186627fe4212a026871f3f89f88a0d096b43555c981bbba70ce707754bf9f99b581ce82582ee15"]}' http://127.0.0.1:7546`

**Response**
`{
   "result": "0x5400634c3bc22439a8919defca8acd2222d68576e08212324a659f1d0f8c9499",
   "jsonrpc": "2.0",
   "id": "1"
}`

3. Query the relay with the transaction hash from the previous response

**Request**
`curl -X POST -H 'Content-Type: application/json' -d '{
   "method": "eth_getTransactionByHash",
   "params": ["0x5400634c3bc22439a8919defca8acd2222d68576e08212324a659f1d0f8c9499"],
   "id": 1,
   "jsonrpc": "2.0"
 }' http://127.0.0.1:7546`

**Response**
`{
   "result": {
       "blockHash": "0x75b6a360587c648d60695b38d32f35d54363c000f37e3a6af13fd1ac3fe96350",
       "blockNumber": "0x5c",
       "chainId": "0x12a",
       "from": "0x00000000000000000000000000000000000003f4",
       "gas": "0x251c0",
       "gasPrice": "0x48",
       "hash": "0x5400634c3bc22439a8919defca8acd2222d68576e08212324a659f1d0f8c9499",
       "input": "0x604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3",
       "nonce": "0x0",
       "r": "0x15a5e6900dadf4a75d924aba5ba296df568ac3c119395c22d2d3186627fe4212",
       "s": "0x26871f3f89f88a0d096b43555c981bbba70ce707754bf9f99b581ce82582ee15",
       "to": "0x0000000000000000000000000000000000000408",
       "transactionIndex": "0x0",
       "type": "0x0",
       "v": "0x1",
       "value": "0x0"
   },
   "jsonrpc": "2.0",
   "id": 1
}
`
 
 
 
 

### Additional context

_No response_

### Hedera network

other

### Version

v0.6.0

### Operating system

macOS

lukelee-sl:

**Description**:
Currently we are returning the long-zero address in the from field.  We need to return the evm address for the account if one exists.

**Related issue(s)**:

Fixes #505 

### DIFF
--- a/packages/relay/tests/lib/eth.spec.ts
+++ b/packages/relay/tests/lib/eth.spec.ts
@@ -1678,6 +1678,8 @@ describe('Eth', async function () {
     ethImpl = new EthImpl(null, mirrorNodeInstance, logger);
   });
 
+  const defaultFromLongZeroAddress = '0x0000000000000000000000000000000000001f41';
+  const defaultEvmAddress = '0x67D8d32E9Bf1a9968a5ff53B87d777Aa8EBBEe69';
   const defaultTxHash = '0x4a563af33c4871b51a8b108aa2fe1dd5280a30dfb7236170ae5e5e7957eb6392';
   const defaultTransaction = {
     "accessList": undefined,
@@ -1939,6 +1941,10 @@ describe('Eth', async function () {
     it('returns correct transaction for existing hash', async function () {
       // mirror node request mocks
       mock.onGet(`contracts/results/${defaultTxHash}`).reply(200, defaultDetailedContractResultByHash);
+      mock.onGet(`accounts/${defaultFromLongZeroAddress}`).reply(200, {
+        from: `${defaultEvmAddress}`
+      });
+
       const result = await ethImpl.getTransactionByHash(defaultTxHash);
 
       expect(result).to.exist;
@@ -1974,6 +1980,9 @@ describe('Eth', async function () {
       };
 
       mock.onGet(`contracts/results/${defaultTxHash}`).reply(200, detailedResultsWithNullNullableValues);
+      mock.onGet(`accounts/${defaultFromLongZeroAddress}`).reply(200, {
+        from: `${defaultEvmAddress}`
+      });
       const result = await ethImpl.getTransactionByHash(defaultTxHash);
       if (result == null) return;
 
--- a/packages/relay/tests/lib/eth.spec.ts
+++ b/packages/relay/tests/lib/eth.spec.ts
@@ -30,7 +30,11 @@ dotenv.config({ path: path.resolve(__dirname, '../test.env') });
 import {RelayImpl, predefined, MirrorNodeClientError} from '@hashgraph/json-rpc-relay';
 import { EthImpl } from '../../src/lib/eth';
 import { MirrorNodeClient } from '../../src/lib/clients/mirrorNodeClient';
-import { expectUnsupportedMethod } from '../helpers';
+import { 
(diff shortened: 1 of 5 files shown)


packages/relay/tests/lib/eth.spec.ts
```diff
@@ -1678,6 +1678,8 @@ describe('Eth', async function () {
     ethImpl = new EthImpl(null, mirrorNodeInstance, logger);
   });
 
+  const defaultFromLongZeroAddress = '0x0000000000000000000000000000000000001f41';
```
Nana-EC, line 1681: Don't need this since it's specified in `helpers.ts`

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -1939,6 +1941,10 @@ describe('Eth', async function () {
     it('returns correct transaction for existing hash', async function () {
       // mirror node request mocks
       mock.onGet(`contracts/results/${defaultTxHash}`).reply(200, defaultDetailedContractResultByHash);
+      mock.onGet(`accounts/${defaultFromLongZeroAddress}`).reply(200, {
+        from: `${defaultEvmAddress}`
```
Nana-EC, line 1945: Can you update `defaultTransaction.from` to use the `defaultEvmAddress` and this, otherwise it's unclear why in the following assertions the comparisons would be true
```suggestion
        from: `${defaultTransaction.from}`
```

Nana-EC: Same below also

Nana-EC (changes requested):
Looks good.
Just some test clarity

Nana-EC (approved):
LG

PR #565, opened 2022-09-26
Add workaround to ensure contractcallquery is sufficient

### LINKED ISSUE #564: Add workaround for inconsistent cost query on ContractCall
### Problem

Seeing some inconsistencies with cost query amounts in previewnet using the sdk-js.

```
[Request ID: a20e7234-74a5-4451-9533-532eb3b4e924] call(hash={"from":"0xd623443ab1e2b587e42a76c6ca8608b0cefa6bd5","to":"0x000000000000000000000000000000000000403d","data":"0x06fdde03"}, blockParam=latest)
[Request ID: a20e7234-74a5-4451-9533-532eb3b4e924] Consensus Node query response: ContractCallQuery INSUFFICIENT_TX_FEE
[Request ID: a20e7234-74a5-4451-9533-532eb3b4e924] 0.0.902@1664221028.915318756 eth_call query cost: 0.01239914 ℏ
[Request ID: a20e7234-74a5-4451-9533-532eb3b4e924] Failed to successfully submit contractCallQuery
```

As part of logic in the relay with set the from, to and data values noted above in a ContractCallQuery and then use a getCost() call to get the appropriate costs to ensure we don't overpay.
```
const cost = await contractCallQuery
            .getCost(this.clientMain);
```

However, as of previewnet upgrade to 0.31.0-alpha.0 services the sdk-js started producing consistent `INSUFFICIENT_TX_FEE` errors.
It seems using the returned cost is never enough.
If I modify the cost with a factor of as little as 1.001 I can get it to pass and no longer get the `INSUFFICIENT_TX_FEE` for the `ContractCallQuery`
`const updatedCost = cost._valueInTinybar.multipliedBy(1.001).integerValue();`

### Solution

Add a workaround to unblock calls in the outgoing 0.8.0 whiles the issue is investigated.

### Alternatives

_No response_

Nana-EC:

**Description**:
`contractCallQuery.getCost()` seems to not be sufficient to cover the cost of the contract call query being made.
To workaround this while the issue is addressed add a factor of 1.001 which has been shown to be a sufficient bump

**Related issue(s)**:

Fixes #564 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -44,6 +44,7 @@ import {
     FileInfoQuery,
     EthereumTransaction,
     EthereumTransactionData,
+    Hbar,
 } from '@hashgraph/sdk';
 import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
 import { Logger } from "pino";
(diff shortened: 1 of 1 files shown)


Nana-EC:
Workaround not need. dynamic property was the likely issue

PR #568, opened 2022-09-27
Nonce to 0 for Hedera native transactions

### LINKED ISSUE #566: Nonce is null for hedera native transactions
### Description

As described here: https://github.com/hashgraph/hedera-services/issues/3987 Hedera native transactions do not update the account nonce field which is why the nonce field is null for non-eth transactions. This will break the integration with some tools, for example TheGraph which rely on the nonce field. 

As per conversations and additional validation - the fix here is to set the nonce to 0.

### Steps to reproduce

Integration with TheGraph

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.7.0

### Operating system

_No response_

georgi-l95:

**Description**:

- Check if nonce is null and change it to 0.
- Expand tests to cover this use case.
- Bump mirror node version.
- Bump SDK version in dapp-example.
- Bump timeout for tests in dapp-example.

**Related issue(s)**:

Fixes #566 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -945,6 +945,10 @@ export class EthImpl implements Eth {
     return input === null ? null : EthImpl.numberTo0x(input);
   }
 
+  static nullableNonceNumberTo0x(input: number | BigNumber): string | null {
```
dimitrovmaksim, line 948: I think `nullable` in the name implies that it can be `null`. Also I'm not sure it's worth adding another function for that, instead we can do `EthImpl.numberTo0x(contractResultDetails.nonce || 0)`

georgi-l95: Good point.

Nana-EC (approved):

dimitrovmaksim (approved):

shemnon (approved):

Nana-EC (approved):
LG

PR #593, opened 2022-10-06
Add TransactionRecordQuery calls to capture failed transaction costs

### LINKED ISSUE #592: Capture missing costs from  failed consensus node transactions
### Problem

Currently the metrics are missing costs from failed transactions

### Solution

Add `TransactionRecordQuery` calls in the `executeTransaction` and `executeGetTransactionRecord`

### Alternatives

_No response_

Nana-EC:

**Description**:
Add TransactionRecordQuery calls to capture fail transaction costs

- update `executeTransaction` catch with `TransactionRecordQuery` logic
- update `executeGetTransactionRecord` catch with `TransactionRecordQuery` logic

**Related issue(s)**:

Fixes #592 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -44,6 +44,8 @@ import {
     FileInfoQuery,
     EthereumTransaction,
     EthereumTransactionData,
+    TransactionRecordQuery,
+    Hbar,
 } from '@hashgraph/sdk';
 import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
 import { Logger } from "pino";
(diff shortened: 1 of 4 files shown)


georgi-l95 (dismissed):
LG

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -364,27 +377,42 @@ export class SDKClient {
...
-            this.captureMetrics(
-                SDKClient.transactionMode,
-                transactionName,
-                sdkClientError.status,
-                0,
-                callerName);
+            let transactionFee: number | Hbar = 0;
+            if (sdkClientError.isValidNetworkError()) {
+                // pull transaction record for fee
+                const transctionRecord = await new TransactionRecordQuery()
```
lukelee-sl, line 398: Wondering if we should handle exceptions thrown by these calls?

Nana-EC: There shouldn't be exceptions since `validateReceiptStatus` is false (which i need to update for the instance above).
I can wrap this in a try catch

Nana-EC: Addressed

lukelee-sl (approved):

PR #604, opened 2022-10-09
Revert reason in eth_call

### LINKED ISSUE #492: Revert reason not sent back in reverted contract call
## Description

When sending a call to a contract that would revert the relayer respond with a `"Unknown error invoking RPC"` instead of responding that the query result ended up in a revert with the revert reason. The json-relayer is actually aware that the error from the query is due to a contract call reverted it seem, in the log I see the following error
transaction 0.0.47814462@1661172942.961021307 failed precheck with status CONTRACT_REVERT_EXECUTED status: 33
but this information is not returned.

Example call
```
curl https://node_endpoint/ \
  -X POST \
  --data '{"method":"eth_call","params":[{"to":"0x1b68bA912B7AC0Cf771F36ac356Bbceda299fbcf","data":"0x20965255"}, "latest"],"id":1,"jsonrpc":"2.0"}'
```
response: 
```
{"code":3,"data":"0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000d53657420746f2072657665727400000000000000000000000000000000000000","message":"execution reverted: Set to revert"}}
```
0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000d53657420746f2072657665727400000000000000000000000000000000000000
which indicate that the call was reverted and includes the revert reason

### Steps to reproduce

- Send a call on a contract that would revert to the json relayer 

### Additional context

_No response_

### Hedera network

testnet

### Version

testnet

### Operating system

_No response_

Ivo-Yankov:
**Description**:
- Returns the error data and a readable reason when a contract call is reverted in `eth_call`.
- Adds a new `revertReason` field in the response of `eth_getTransactionReceipt`.

**Related issue(s)**:

Fixes #492 

**Notes for reviewer**:

**Checklist**


packages/relay/src/formatters.ts
```diff
@@ -24,11 +24,34 @@ const hashNumber = (num) => {
...
 }
 
-export { hashNumber, formatRequestIdMessage };
+/**
+ * Converts an EVM ErrorMessage to a readable form. For example this :
+ * 0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000d53657420746f2072657665727400000000000000000000000000000000000000
+ * will be converted to "Set to revert"
+ * @param message
+ */
+const extractReadableMessage = (message?: string): string => {
```
Nana-EC, line 49: nit: this and `hexToASCII()` need unit tests

packages/relay/src/lib/eth.ts
```diff
@@ -972,6 +972,10 @@ export class EthImpl implements Eth {
         status: receiptResponse.status,
       };
 
+      if (receiptResponse.error_message) {
```
Nana-EC, line 975: Update `eth.spec.ts` with logic to confirm this

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -309,6 +310,10 @@ export class SDKClient {
                 this.hbarLimiter.addExpense(cost, currentDateNow);
             }
 
+            if (e instanceof PrecheckStatusError && e.contractFunctionResult?.errorMessage) {
```
Nana-EC, line 313: add unit test check for this

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -137,4 +140,10 @@ export const predefined = {
     code: -32007,
     message: 'Historical balance data is available only after 15 minutes.'
   }),
+  'CONTRACT_REVERT': (errorMessage?: string) => new JsonRpcError({
+    name: 'Contract revert executed',
+    code: 3,
```
Nana-EC, line 145: Where are we setting this code off of as it seems to break current convention

Nana-EC (changes requested):
Looking good.
Some missing UTs and i'm not sure about the error code value

packages/relay/src/lib/errors/JsonRpcError.ts
Ivo-Yankov: I tested the same use case with the Alchemy's rpc relay and received this code. It's also the same code mentioned in the linked issue.

Nana-EC (dismissed):
LG

natanasow (dismissed):
LG

natanasow (approved):

PR #624, opened 2022-10-17
Better error handling

dimitrovmaksim:
**Description**:

1. Better handle timeout errors - Currently any error that does not have a `response` property is handled as `unknownServerErrorHttpStatusCode` (567), but `axios` provides `error.code` that indicate (I'm assuming server side errors) for example `ECONNABORTED` for timeouts. Handling those errors as `unknownServerErrorHttpStatusCode` would sometimes result in test passing for the wrong reason, because other errors, like `no mock for ... detected` would also fail with status code `567`. Also update the tests to check the error message.
2. There were couple of cases when `404 not found` responses were not handled correctly and resulted in the code breaking because it was trying to access properties of `null` values. For example here https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1190 because 404 is an acceptable status code, the value of contractResultDetails can be null and the code would break when trying to access the `r` property. Although this error is caught, it seems it was more of a coincidence.
3. Updates tests, removes obsolete mocks, adds missing mocks
4. Refactor the error handling to throw errors only for unexpected behaviours or when the eth method would throw an error instead of returning `null` value

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/625

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -155,7 +155,7 @@ export class MirrorNodeClient {
             return response.data;
         } catch (error: any) {
             ms = Date.now() - start;
-            const effectiveStatusCode = error.response !== undefined ? error.response.status : MirrorNodeClient.unknownServerErrorHttpStatusCode;
+            const effectiveStatusCode = error.response?.status || MirrorNodeClientError.ErrorCodes[error.code] || MirrorNodeClient.unknownServerErrorHttpStatusCode;
             this.mirrorResponseHistogram.labels(pathLabel, effectiveStatusCode).observe(ms);
             this.handleError(error, path, effectiveStatusCode, allowedErrorStatuses, requestId);
         }
@@ -240,7 +240,7 @@ export class MirrorNodeClient {
         const queryParams = this.getQueryParams(queryParamObject);
         return this.request(`${MirrorNodeClient.GET_CONTRACT_RESULTS_ENDPOINT}${queryParams}`,
             MirrorNodeClient.GET_CONTRACT_RESULTS_ENDPOINT,
-            [400],
+            [400, 404],
             requestId);
     }
 
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -203,7 +203,7 @@ export class MirrorNodeClient {
     public async getBlock(hashOrBlockNumber: string | number, requestId?: string) {
         return this.request(`${MirrorNodeClient.GET_BLOCK_ENDPOINT}${hashOrBlockNumber}`,
             MirrorNodeClient.GET_BLOCK_ENDPOINT,
-            [400],
+            [400, 404],
             requestId);
     }
 
@@ -247,7 +247,7 @@ export class MirrorNodeClient {
     public async getContractResultsDetails(contractId: string, timestamp: string, requestId?: string) {
         return this.request(`${this.getContractResultsDetailsByContractIdAndTimestamp(contractId, timestamp)}`,
             MirrorNodeClient.GET_CONTRACT_RESULTS_DETAILS_BY_CONTRACT_ID_ENDPOINT,
-            [400],
+            [400, 404],
             requestId);
     }
 
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -479,17 +479,14 @@ export class EthImpl implements Eth {
       await this.mirrorNodeClient.getContractResultsDetails(address, contractResult.results[0].timestamp)
         .then(contractResultDetails => {
           if(contractResultDetails?.state_changes != null) {
-            // loop through the state changes to match slot and return value
-            for (const stateChange of contractResultDetails.state_changes) {
-              if(stateChange.slot === slot) {
-                result = stateChange.value_written;
-              }
-            }
+            // filter the state changes to match slot and return value
+            const stateChange = contractResultDetails.state_changes.find(stateChange => stateChange.slot === slot);
+            result = stateChange.value_written;
           } else {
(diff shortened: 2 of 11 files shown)


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -240,7 +240,7 @@ export class MirrorNodeClient {
         const queryParams = this.getQueryParams(queryParamObject);
         return this.request(`${MirrorNodeClient.GET_CONTRACT_RESULTS_ENDPOINT}${queryParams}`,
             MirrorNodeClient.GET_CONTRACT_RESULTS_ENDPOINT,
-            [400],
+            [400, 404],
```
Nana-EC, line 243: Please update mirrorNodeClient.spec.ts with this coverage

packages/relay/src/lib/eth.ts
```diff
@@ -1187,29 +1183,34 @@ export class EthImpl implements Eth {
...
-          s: sSig,
-          to: contractResultDetails.to.substring(0, 42),
-          transactionIndex: EthImpl.numberTo0x(contractResultDetails.transaction_index),
-          type: EthImpl.nullableNumberTo0x(contractResultDetails.type),
-          v: EthImpl.nullableNumberTo0x(contractResultDetails.v),
-          value: EthImpl.numberTo0x(contractResultDetails.amount),
-        });
+        // 404 is allowed return code so it's possible for contractResultDetails to be null
+        if(contractResultDetails == null) {
+          return null;
```
Nana-EC, line 1188: Doe eth.spec.ts have this coverage already or does it need to be added?

Nana-EC (changes requested):
Looks good, just potentially needs some missing test cases for coverage

packages/relay/src/lib/eth.ts
dimitrovmaksim: I think it was, but I'll double-check.

I found something else in for that catch, that may be a potential issue. Currently we are catching any errors and returning `null` for that transaction. For example if a `contractResultDetails` property is `null` and some operation is done on it, the code will break with an error. For example if the block_hash here https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1194 is null for some reason (which is an unexpected behaviour and we would like to be aware if it happens) the error will be caught by the `catch` block and it will just return null. I've tested it in the `openrpc` specs and if I set the block_hash to `null` in the mock here https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/tests/lib/openrpc.spec.ts#L124, the `eth_getBlockByNumber` tests will still return a response, but with one transaction missing. 

Usually it's a good practice in the catch blocks to catch specific error or behaviours that are expected to happen, but let the rest throw an error and break, so we are aware that there's an unexpected issue or behaviour. While currently it will log the error to the console, it seems the response still succeeds, and the issue can easily be missed.

dimitrovmaksim: In regards with the tests, mocks for `getContractResultsByAddressAndTimestamp` were missing for the `eth_getBlockByNumber`, which resulted in a `no mock` error, which was handled by the `catch` block and the `getTransactionFromContractResult` function returned `null`. Currently I've set all mocks to return 404, because the specs only test that there's a result, aka a block is returned, and that the block number is the expected one.

packages/relay/src/lib/clients/mirrorNodeClient.ts
dimitrovmaksim: Actually now i tested that 404 is returned only for no matches when querying block hash + index, for block number + index it will return 200 and empty array. So maybe the issue here is with the mirror node, not the relay?

Nana-EC: 404's are applicable expected error codes.
For example asking for an account that doesn't exist will return a 404 and this is expected

packages/relay/src/lib/eth.ts
Nana-EC: Agreed. We should catch expected errors and throw unexpected so we fix them.

Nana-EC: You can update this PR or open a ticket highlighting the improved logic

packages/relay/src/lib/eth.ts
```diff
@@ -634,7 +641,7 @@ export class EthImpl implements Eth {
     this.logger.trace(`${requestIdPrefix} getBlockByHash(hash=${hash}, showDetails=${showDetails})`);
     return this.getBlock(hash, showDetails, requestId).catch((e: any) => {
       this.logger.error(e, `${requestIdPrefix} Failed to retrieve block for hash ${hash}`);
-      return null;
+      throw predefined.INTERNAL_ERROR(e.message);
```
Nana-EC, line 644: Q: Do these new internal errors leak SDK or network information?
If so we should be putting static string here since we have the logging

Nana-EC (commented):
LG.
Just curious about if the internal error response is too details with `e.message`

packages/relay/src/lib/eth.ts
dimitrovmaksim: It will probably forward the error message if the error is MirrorNodeError or SDKError, but I don't know what info the messages contain. If it's a problem I can revert it. Or we could send the requestId and have something like `Internal error for request ${requestId}`. I was thinking that having more info about the error would be better to then locate it in the logs, when/if someone reports an issue, especially for the officially hosted relay, where dapp devs don't have access to the logs themselves.

Nana-EC: Yeah just log for now, we wouldn't want dependency (mirror node or sdk) related info showing up in the relay errors.
You can put static string in the errors to inform clients on actionable steps but if not leave it empty.
You can leave the logic to support a message in there as it's useful

e.g.
```
throw predefined.INTERNAL_ERROR('Downstream issue retrieving block information from network nodes');
```

Nana-EC (approved):

georgi-l95 (approved):
LGTM

PR #637, opened 2022-10-18
Fix rate limit not reseting

### LINKED ISSUE #623: Rate limiting cache occasionally not resetting after long periods
### Description

Some users have reported consistent rate limiting on their IP even when their rps is much lower than the threshold.

The issue was able to be resolved after restarts on the relay thus suggesting the the caching system may have an issue resetting

### Steps to reproduce

1. Run the relay for an extended period
2. Run high rpm to hit the rate limit
3. After some period after the window should have been refreshed run more rps

Is has occasionally been seen that users are rate limited when they should be

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.10.0

### Operating system

_No response_

georgi-l95:

**Description**:
If env variable `LIMIT_DURATION` is not set, rate limit duration is NaN and can't reset. Now defaults to `60000 ms`, if it's not set.

**Related issue(s)**:

Fixes #623 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -125,10 +125,10 @@ export const predefined = {
     code: -32000,
     message: `Exceeded maximum block range: ${blockRange}`
   }),
-  'IP_RATE_LIMIT_EXCEEDED': new JsonRpcError({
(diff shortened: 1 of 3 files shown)


lukelee-sl (approved):

Nana-EC (approved):
LG

shemnon (approved):

PR #650, opened 2022-10-23
Pull evm_address for contract logs

### LINKED ISSUE #644: The log object should return the evm address not the hedera entity address
### Description

According to the [eth json rpc spec](https://github.com/etclabscore/ethereum-json-rpc-specification/blob/97178e7dc3b417318e2977171254f6e21d080076/openrpc.json#L1250) the log object address(i.e. `Log.address`) should be the evm address of the contract that emitted the event. However, the relay currently returns the Hedera entity id for the contract in evm address format(e.g. `0x0000000000000000000000000000000000000409` i.e. _0.0.1033_) when it should instead return the actual evm address of the contract i.e. either the `create1` or `create2` address.

A known impact is on the [subgraph data source templates](https://thegraph.com/docs/en/developing/creating-a-subgraph/#instantiating-a-data-source-template) that enables the subgraph to index logs on created contracts. For example, a `Factory` contract creates a `Child` contract and the `Factory` emits an event(say `ChildCreated`) containing the `create2` address of the `Child` contract. The subgraph handler for this `ChildCreated` event has access to the `create2` address from the `event.params` and creates a template index for that `Child` contract. However since the Hedera relay returns the Hedera entity id for the `Child` contract instead of the evm address on `Log.address` the `Child` contract's logs aren't indexed(at least for `create2` contract's at this point since `create1` addresses are currently identical to the contract's entity id).

### Steps to reproduce

1. Pull and checkout to [this branch]()
2. Start up the hedera local node
3. Start up the local graph node
4. Run `npm run test:hgf` to deploy and test the `GreeterFactory` and created `Greeter` contracts
5. In the `subgraph-example` update the `GreeterFactory` `address` and `startBlock` accordingly
6. create and deploy the subgraph to the local graph node
7. Observe that the child Greeter contract using `create1` has been [templated correctly](https://github.com/hashgraph/hedera-json-rpc-relay/pull/642/files#r999654754) as it's `handleGreetingSet` event handler has been called accordingly and observe that the child `Greeter` contract using `create2` has [not been templated correctly](https://github.com/hashgraph/hedera-json-rpc-relay/pull/642/files#r999656054) as it's handleGreetingSet is not triggered due to the address mismatch.

### Additional context

_No response_

### Hedera network

other

### Version

@hashgraph/hedera-local@2.0.1

### Operating system

macOS

Nana-EC:

**Description**:
Provide a workaround to pull evm_address for contract logs
- for logs retrieve evm_address from mirror node

**Related issue(s)**:

Fixes #644 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1345,8 +1345,6 @@ export class EthImpl implements Eth {
 
   private async getContractEvmAddress(address: string, requestId: string | undefined) {
     const result = await this.mirrorNodeClient.getContract(address, requestId);
-    console.log("HERE");
-    console.log(result.evm_address);
     return result.evm_address || address;
   }
 }
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1320,9 +1320,9 @@ export class EthImpl implements Eth {
       return [];
     }
 
-    return result.logs.map(log => {
+    return result.logs.map(async log => {
       return new Log({
-        address: log.address,
+        address: await this.getContractEvmAddress(log.address, requestId),
         blockHash: EthImpl.toHash32(log.block_hash),
         blockNumber: EthImpl.numberTo0x(log.block_number),
         data: log.data,
@@ -1339,5 +1339,10 @@ export class EthImpl implements Eth {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     this.logger.trace(`${requestIdPrefix} maxPriorityFeePerGas()`);
     return EthImpl.zeroHex;
+  }
+
+  private async getContractEvmAddress(address: string, requestId: string | undefined) {
+    const result = await this.mirrorNodeClient.getContract(address, requestId);
+    return result.evm_address || address; 
   }
 }
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1320,10 +1320,11 @@ export class EthImpl implements Eth {
       return [];
     }
 
-    return Promise.all(
-      result.logs.map(async log => {
-        return new Log({
-          address: await this.getContractEvmAddress(log.address, requestId),
+    const logs: Log[] = [];
+    for(const log of result.logs) {
+      logs.push(
+        new Log({
+          address: await this.getLogEvmAddress(log.address, requestId) || log.address,
           blockHash: EthImpl.toHash32(log.block_hash),
           blockNumber: EthImpl.numberTo0x(log.block_number),
           data: log.data,
@@ -1332,9 +1333,11 @@ export class EthImpl implements Eth {
           topics: log.topics,
           transactionHash: EthImpl.toHash32(log.transaction_hash),
           transactionIndex: EthImpl.numberTo0x(log.transaction_index)
-        });
-      })
-    );
+        })
+      );
+    }
+
+    return logs;
   }
 
   async maxPriorityFeePerGas(requestId?: string): Promise<string> {
@@ -1343,8 +1346,18 @@ export class EthImpl implements Eth {
     return EthImpl.zeroHex;
   }
 
-  private async getContractEvmAddress(address: string, requestId: string | undefined) {
-    const result = await this.mirrorNodeClient.getContract(address, requestId);
-    return result.evm_address || address;
+  private async getLogEvmAddress(address: string, requestId: string | undefined) {
+    const cachedLabel = `getLogEvmAddress.${address}`;
+    let contractAddress = cache.get(cachedLabel);
+
+    // If contractAddress === undefined it means there's no cache record
+    // and a mirror node request should be executed.
+    if(contractAddress === undefined) {
+      const contract = await this.mirrorNodeClient.getContract(address, requestId);
+      contractAddress = contract.evm_address;
+      cache.set(cachedLabel, contractAddress, constants.CACHE_TTL.ONE_HOUR);
+    }
+
+    return contractAddress;
   }
 }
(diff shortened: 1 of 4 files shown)


dimitrovmaksim:
I just realised that this may not fix the Graph issue. The problem is that when using dynamic dataSource templates for child contracts, and the child contract is created with create2, the dynamic datasource will be created with the child contract's `evm_address`, then when the graph node does a `eth_getLogs()` method call for a datasource it would pass `address`, `topic0`, `fromBlock` and `toBlock` filters. So if the mirror node does not filter the logs by the `evm_address` we still wont have matches for that evm_address.

packages/relay/src/lib/eth.ts
```diff
@@ -1320,24 +1320,31 @@ export class EthImpl implements Eth {
...
-        removed: false,
-        topics: log.topics,
-        transactionHash: EthImpl.toHash32(log.transaction_hash),
-        transactionIndex: EthImpl.numberTo0x(log.transaction_index)
-      });
-    });
+    return Promise.all(
+      result.logs.map(async log => {
+        return new Log({
+          address: await this.getContractEvmAddress(log.address, requestId),
```
steven-sheehy, line 1326: The log address can repeat multiple times within the payload. `/api/v1/contracts/{id}` is a heavyweight call that returns large byte arrays. We should minimize calls to this API by collecting `log.address` into a set then calling the distinct addresses outside this loop. It's a performance optimization that makes the code not as clean but is necessary to reduce the load on mirror node.

packages/relay/src/lib/eth.ts
```diff
@@ -1320,24 +1320,31 @@ export class EthImpl implements Eth {
...
   }
 
   async maxPriorityFeePerGas(requestId?: string): Promise<string> {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     this.logger.trace(`${requestIdPrefix} maxPriorityFeePerGas()`);
     return EthImpl.zeroHex;
   }
+
+  private async getContractEvmAddress(address: string, requestId: string | undefined) {
+    const result = await this.mirrorNodeClient.getContract(address, requestId);
```
steven-sheehy, line 1347: Can we cache this lookup with some fixed size LRU cache?

steven-sheehy (changes requested):

packages/relay/src/lib/eth.ts
Nana-EC: Can use something like
- check for `contractEvmAddress` i.e 
```
 let contractEvmAddress = cache.get(cachedLabel);
```
- if absent make mirror node call and when done.
```
cache.set(cachedLabel, <contractEvmAddress>, constants.CACHE_TTL.ONE_HOUR);
```

packages/relay/src/lib/eth.ts
Nana-EC: Agreed.
Prefer the caching option since this is just a workaround till Mirror Node adopts similar logic.
Then this work would go away.

packages/relay/src/lib/eth.ts
dimitrovmaksim: I've added the caching, added a test. 

One question, do we need the `result.evm_address || address`, it seems `evm_address` can't be `null` unless `result` is null, in which case `result.evm_address` will throw a `TypeError: Cannot read properties of null (reading 'evm_address')`. Also in this case `result` would be `null` if the mirror node returns 400 or 404, wouldn't we want  to throw an error in the case when the request result is `null`?

steven-sheehy: I think it might be better to catch any errors during `getContractEvmAddress()` and return the `log.address` as a fallback.

packages/relay/src/lib/eth.ts
```diff
@@ -1316,24 +1316,44 @@ export class EthImpl implements Eth {
...
+  private async getLogEvmAddress(address: string, requestId: string | undefined) {
+    const cachedLabel = `getLogEvmAddress.${address}`;
+    let contractAddress = cache.get(cachedLabel);
+
+    // If contractAddress === undefined it means there's no cache record
+    // and a mirror node request should be executed.
+    if(contractAddress === undefined) {
+      const contract = await this.mirrorNodeClient.getContract(address, requestId);
+      contractAddress = contract.evm_address;
+      cache.set(cachedLabel, contractAddress, constants.CACHE_TTL.ONE_HOUR);
```
steven-sheehy, line 1354: It seems `js-cache` doesn't support a fixed size? I hope you don't get a lot of requests in that one hour and cause a memory overflow. This is a security issue. You guys should investigate an alternative library in a follow up.

packages/relay/src/lib/eth.ts
```diff
@@ -1316,24 +1316,44 @@ export class EthImpl implements Eth {
...
     return EthImpl.zeroHex;
   }
+
+  private async getLogEvmAddress(address: string, requestId: string | undefined) {
+    const cachedLabel = `getLogEvmAddress.${address}`;
+    let contractAddress = cache.get(cachedLabel);
+
+    // If contractAddress === undefined it means there's no cache record
+    // and a mirror node request should be executed.
+    if(contractAddress === undefined) {
```
steven-sheehy, line 1351: nit: Spacing here and elsewhere doesn't match the relay standard.

steven-sheehy (dismissed):

steven-sheehy (approved):

Nana-EC (commented):
LG.
Can't approve since i started the PR so consider this my signoff

packages/relay/src/lib/eth.ts
Nana-EC: Added https://github.com/hashgraph/hedera-json-rpc-relay/issues/658 for next sprint

georgi-l95 (approved):
LGTM

PR #659, opened 2022-10-28
Set Request ID on Acceptance Tests

### LINKED ISSUE #608: Set requestId on acceptance test calls
### Problem

Currenlty the aceptance tests can be run against remote envs, however, when doing so it's difficult to determine which calls it's making.

### Solution

The relay supports using the provided requestid on a call when provided.
Using this we should create a requestId, log it and use that value during the call.
This will allow easier tracing of acceptance test calls

### Alternatives

_No response_

### LINKED ISSUE #662: Investigate test runs in CI
### Description

Spend some time and go through the latest runs of the tests in our CI

### Steps to reproduce

Investigate and update

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.10.0

### Operating system

_No response_

georgi-l95:
**Description**:

- Set `requestId` for acceptance tests (see logs from test runs to check how it looks).
- Fix acceptance tests.
- Make tests run parallel and retry 3 times before failing.
- Fix `eth_getLogs` bug, when passing only `toBlock` as parameter. When only `toBlock` is passed, `fromBlock` is being defaulted to `'latest'`, which results in `toBlock < fromBlock`. This returns empty array. Now if only `toBlock` is passed, `fromBlock` defaults to `toBlock`, are returns logs only for this block.
- Fix `eth_getLogs` bug, when passing wrong `blockHash`.

**Related issue(s)**:

Fixes #608 
Fixes #662 

**Notes for reviewer**:

**Checklist**


.github/workflows/acceptance-workflow.yml
```diff
@@ -38,12 +38,8 @@ jobs:
...
-        continue-on-error: true
-        timeout-minutes: 15
-        run: npm run acceptancetest:${{ inputs.testfilter }}
-        
-      - name: Re-run acceptance tests (without nick-fields/retry@v2)
-        if: ${{ failure() }}
-        timeout-minutes: 15
-        run: npm run acceptancetest:${{ inputs.testfilter }}
-
+        uses: nick-fields/retry@v2
```
Nana-EC, line 41: Glad to see this works now.
Previously the tests would never end and so we couldn't use it.

georgi-l95: Yep, now they're retrying 3 times before showing fail.

packages/relay/src/lib/eth.ts
```diff
@@ -1288,7 +1290,7 @@ export class EthImpl implements Eth {
       } else {
           params.timestamp = [];
 
-          const fromBlockResponse = await this.getHistoricalBlockResponse(fromBlock || "latest", true, requestId);
+          const fromBlockResponse = await this.getHistoricalBlockResponse(toBlock && !fromBlock ? toBlock : fromBlock || "latest", true, requestId);
```
Nana-EC, line 1293: Pull the first param calculation out into a variable for improved readability.

packages/server/tests/clients/servicesClient.ts
```diff
@@ -349,32 +361,34 @@ export default class ServicesClient {
...
 
             const rec = await tokenTransfer.getReceipt(this.client);
-            this.logger.info(`${amount} of HTS Token ${tokenId} can be spent by ${accountId}`);
-            this.logger.debug(rec);
+            this.logger.info(`${requestIdPrefix} ${amount} of HTS Token ${tokenId} can be spent by ${accountId}`);
+            this.logger.debug(`${requestIdPrefix} ${rec}`);
         }
         catch(e) {
-            this.logger.debug(e);
+            this.logger.debug(`${requestIdPrefix} ${e}`);
```
Nana-EC, line 391: nit:
```suggestion
            this.logger.error(e, `${requestIdPrefix} TransferTransaction failed`);
```

Nana-EC (approved):
LG. Some suggestions

packages/server/tests/clients/servicesClient.ts
georgi-l95: Done.

packages/relay/src/lib/eth.ts
georgi-l95: Good idea. Done.

packages/relay/src/lib/eth.ts
```diff
@@ -1290,7 +1290,8 @@ export class EthImpl implements Eth {
       } else {
           params.timestamp = [];
 
-          const fromBlockResponse = await this.getHistoricalBlockResponse(toBlock && !fromBlock ? toBlock : fromBlock || "latest", true, requestId);
+          const onlyToBlockPassed = toBlock && !fromBlock;
```
Nana-EC, line 1293: nit: More like
```suggestion
          // Use the `toBlock` if it is the only passed tag, if not utilize the `fromBlock` or default to "latest"
          const blockTag = toBlock && !fromBlock ? toBlock : fromBlock || "latest";
          const fromBlockResponse = await this.getHistoricalBlockResponse(blockTag, , true, requestId);
```

Nana-EC (approved):
LG, nit

packages/relay/src/lib/eth.ts
georgi-l95: Yep, you're right. Fixed.

Nana-EC (approved):
LGTM

Issue #651, opened 2022-10-24
Add eth methods arguments/parameters validations


dimitrovmaksim:
### Problem

The relay does not validate the method parameters. There isn't required params validations, or param type/format validation of any kind. The relay will just forward whatever is passed as params to the mirror node and usually the mirror node will return `not found` and the relay will return `null` response. 

This can be a problem, because it could be hard for developers to realise they have made an error when sending a request to the relay, because the relay will not return a proper error, just a `null` response. 

```
Relay:

curl http://localhost:7546 -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"eth_getTransactionByBlockNumberAndIndex","params": ["0x190BDD8"],"id":1}'
{"result":null,"jsonrpc":"2.0","id":1}% 

Infura:

curl https://mainnet.infura.io/v3/.... -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"eth_getTransactionByBlockNumberAndIndex","params": ["0x190BDD8"],"id":1}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"missing value for required argument 1"}}%                                                        
```                                                          

```
Relay:

curl http://localhost:7546 -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionByBlockHashAndIndex","params":["0x0", "0xc6ef2fc5426d6ad6fd9e2a26abeab0aa2411b7ab17f30a99d3cb96aed1d1254f"],"id":1,"jsonrpc":"2.0"}' 
{"result":null,"jsonrpc":"2.0","id":1}%      

Infura:
curl https://mainnet.infura.io/v3/.... -X POST -H "Content-Type: application/json" --data '{"method":"eth_getTransactionByBlockHashAndIndex","params":["0x0", "0xc6ef2fc5426d6ad6fd9e2a26abeab0aa2411b7ab17f30a99d3cb96aed1d1254f"],"id":1,"jsonrpc":"2.0"}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"invalid argument 0: json: cannot unmarshal hex string of odd length into Go value of type common.Hash"}}
```

Also not sure if this should be by design or just a coincidence, but the relay accepts both `hex` and `dec` numbers

Relay:
```
[...]

PR #663, opened 2022-10-31
651 add eth methods validations

dimitrovmaksim:
**Description**:
Adds input validations for the `eth` methods. The validation are based on multiple rpc specifications and [openrpc.json](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/hashgraph/hedera-json-rpc-relay/main/docs/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false)

There are some cases where the validation is not as strict as Alchemy for example.

Also It adds temporary fixes to the acceptance tests, because they were inconsistent. There were various cases where hex values were sent without `0x` prefix, and even cases where hedera format (`0.0.xxx`) account, contract or transaction ID were sent as parameters.

I think it would be good to test this more extensively before releasing it as official image version as there may be some missing edge cases.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/651

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -99,7 +99,7 @@ describe('@erc20 Acceptance Tests', async function () {
             });
 
             it('Relay can execute "eth_getCode" for ERC20 contract with evmAddress', async function () {
-                const res = await relay.call('eth_getCode', [contract.address], requestId);
+                const res = await relay.call('eth_getCode', [contract.address, "latest"], requestId);
```
Nana-EC, line 102:
```suggestion
                const res = await relay.call('eth_getCode', [contract.address, 'latest'], requestId);
```

Nana-EC (commented):
1st pass looking good. Will circle back for 2nd pass

packages/server/src/validator.ts
```diff
@@ -0,0 +1,278 @@
+import { JsonRpcError, predefined } from '@hashgraph/json-rpc-relay';
```
georgi-l95, line 1: Good progress. My opinion is that this file way too big, with way too many responsibilities. You can put all const variables in separated `constants.ts` file. Pull out all static methods in separate file as well and keep only one class per file. This way when 3rd person sees the files will know - okay, here are the constants, here we have this class, which does that and so on.. I personally try to aim for higher cohesion.

dimitrovmaksim:
> 1st pass looking good. Will circle back for 2nd pass

Please also consider that I'm not fully sure about the Transaction object validations, there are various cases depending on the method or type of transaction and I can't find definitive answer how it should be, especially regarding required params

packages/server/src/server.ts
```diff
@@ -21,6 +21,7 @@
 import { Relay, RelayImpl, JsonRpcError, predefined, MirrorNodeClientError } from '@hashgraph/json-rpc-relay';
```
georgi-l95, line 21: Nice, but my approach would be to make this validations in `koaJsonRpc` with the other checks. There we already have some kind of invalid params check, we can just extend it and not cluster `server.ts` with unnecessary code.
Overall looks good, it's important to have validations.

georgi-l95 (commented):
LG, sharing my opinion.

packages/server/src/server.ts
dimitrovmaksim: Yeah, I wasn't sure for the best place for this.

packages/server/src/validator.ts
dimitrovmaksim: The issue I see here is that this will probably cause circular dependencies and will probably require some level of code/flow refactoring to escape this.

packages/server/src/server.ts
```diff
@@ -217,6 +218,18 @@ app.useRpc('eth_blockNumber', async () => {
  * returns: Gas used - hex encoded integer
  */
 app.useRpc('eth_estimateGas', async (params: any) => {
+  const validationError = Validator.validateParams(params, {
```
Nana-EC, line 221: Validating too early may prevent the log and metrics of the endpoint call from occurring.
Please check this is not the case.
if it is maybe you want to pass the Validator into the `logAndHandleResponse()` so that it gets run before the service logic is called

Nana-EC (commented):
Nice validations. 
Please check they occur after initial entry point logs and metrics so we can still observe the calls but users get the response as to what they missed on the call

packages/server/src/server.ts
dimitrovmaksim: If the validation happens in `logAndHandleResponse()`, we will only get part of the logs 
```
logger.debug(`${requestIdPrefix} ${methodName}`);
```
because trace logging the called method and the params passed
```
```
happens in function of each method the `eth.ts` module in the `relay` package.
Would this be enough?

lukelee-sl: Good question.  Let me ask a question in response.  Does the validation error message give you enough information about all of the parameters were passed into the function and which one was failing validation?  If so, perhaps there is enough information for the reader of the log to piece together what went wrong?

One alternate suggestion which is kind of ugly as it involves a fair amount of refactoring is to pass in an object instead of the `requestId` string.  The object could contain the `validator` and the `requestId`.  I'm not sure what a good name for such an object might be.

dimitrovmaksim: It either gives you the index of the parameter, for example in
```
{ "method" : "eth_getBlockByNumber" ,"params" : ["1", false], "id": 1, "jsonrpc": "2.0" }
```
It will return that `param at index 0` is not a hex number.

In the case where one of the params itself is an object (Filter or Transaction)
```
{ "method" : "eth_getLogs", "params" : [ {"address": "0x1234"} ], "id":1 , "jsonrpc":"2.0" }
```
It will return that the `address` property of the `Filter` object is not a valid address

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -1067,7 +1067,7 @@ describe('@api RPC Server Acceptance Tests', function () {
             });
         });
 
-        describe('@release Unsupported RPC Endpoints', () => {
+        describe.only('@release Unsupported RPC Endpoints', () => {
```
Ivo-Yankov, line 1070: Remove `.only`

Ivo-Yankov (approved):

lukelee-sl (approved):
LGTM

natanasow (approved):
LG

PR #664, opened 2022-11-01
Add acceptance test for getStorageAt

### LINKED ISSUE #449: Add `eth_getStorageAt` acceptance test coverage
### Problem

The relay added support to `eth_getStorageAt` to call the mirror node.
However, no E2E tests could be added at the time

### Solution

Add `eth_getStorageAt` acceptance test coverage

- ensure the services node is updated to support side car propogation of state changes

### Alternatives

_No response_

lukelee-sl:

**Description**:

Fixes #449


packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -1149,6 +1150,49 @@ describe('@api RPC Server Acceptance Tests', function () {
...
+                const transactionHash = await relay.call('eth_sendRawTransaction', [signedTx]);
+
+                const getStoragAtParams = {
+                    address: contractId,
+                    slot: 0,
+                    blockNumber: "latest",
+                }
+
+                const storageVal = await relay.call('eth_getStorageAt', getStoragAtParams );
+                expect(storageVal).to.eq(8)
```
Nana-EC, line 1192: Add a check for the original value before calling the update method

Nana-EC (dismissed):
LG.
Just add an extra call to get the state before you update and verify .
That way we get coverage for before and after

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -1149,6 +1150,49 @@ describe('@api RPC Server Acceptance Tests', function () {
...
+                    data: STORAGE_CONTRACT_UPDATE,
+                    maxPriorityFeePerGas: gasPrice,
+                    maxFeePerGas: gasPrice,
+                    type: 2
+                };
+
+                const signedTx = await accounts[1].wallet.signTransaction(transaction);
+                const transactionHash = await relay.call('eth_sendRawTransaction', [signedTx]);
+
+                const getStoragAtParams = {
```
dimitrovmaksim, line 1185: `eth_getStorageAt` params should be an array of strings where position 0 is an address, 1 is the slot in hex and 3 is the block number in hex or tag

packages/server/tests/acceptance/rpc.spec.ts
dimitrovmaksim: Also the result value will be in hex

Nana-EC (approved):
LG

PR #666, opened 2022-11-02
Replace js-cache with lru-cache

### LINKED ISSUE #658: Explore js-cache alternative
### Problem

`js-cache` is currently used for caching static responses to speed up responses and reduce unnecessary load on nodes.
However, it has limitation around size limit and other items.

### Solution

Explore alternatives and replace `js-cache` with an option that increase capability including
- cache size limitations
- automatic flushing
- [Stretch] LRU properties

### Alternatives

_No response_

georgi-l95:

**Description**:
Replace `js-cache` with `lru-cache`. A cache object that deletes the least-recently-used items with customisable options.

Options currently set during initialization are:
  - `max` - The maximum number (or size) of items that remain in the cache (assuming no TTL pruning or explicit deletions).
  - `ttl` - Max time to live for items before they are considered stale.
  
**Related issue(s)**:

Fixes #658 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -30,9 +30,8 @@ import { MirrorNodeClientError } from './errors/MirrorNodeClientError';
 import constants from './constants';
 import { Precheck } from './precheck';
 import { formatRequestIdMessage } from '../formatters';
-
+const LRU = require('lru-cache');
 const _ = require('lodash');
-const cache = require('js-cache');
 const createHash = require('keccak');
 
 /**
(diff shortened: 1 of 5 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -164,13 +184,13 @@ export class EthImpl implements Eth {
...
 
-      let feeHistory: object | undefined = cache.get(constants.CACHE_KEY.FEE_HISTORY);
+      let feeHistory: object | undefined = this.cache.get(constants.CACHE_KEY.FEE_HISTORY);
       if (!feeHistory) {
 
         feeHistory = await this.getFeeHistory(blockCount, newestBlockNumber, latestBlockNumber, rewardPercentiles, requestId);
 
         this.logger.trace(`${requestIdPrefix} caching ${constants.CACHE_KEY.FEE_HISTORY} for ${constants.CACHE_TTL.ONE_HOUR} ms`);
-        cache.set(constants.CACHE_KEY.FEE_HISTORY, feeHistory, constants.CACHE_TTL.ONE_HOUR);
+        this.cache.set(constants.CACHE_KEY.FEE_HISTORY, feeHistory);
```
Nana-EC, line 193: With the new logic does one have to create a new cache object if the max time period is not the default 1 hour set in options?
Or is it flexible enough to pass in a time param to override the default option?

Nana-EC (approved):
LG. 
Question on configurability of cache ttl.

packages/relay/src/lib/eth.ts
georgi-l95: Yes, according to their documentation it's possible to pass options, which override default ones: https://github.com/isaacs/node-lru-cache#setkey-value--size-sizecalculation-ttl-nodisposeonset-start-

PR #668, opened 2022-11-02
Developer mode

Ivo-Yankov:
**Description**:

Adds a `developer mode` to the relay. Currently it only changes the behaviour of `eth_getTransactionByHash`, but in the future it can be used for enabling special settings in the local node, which will make it run faster.

The following code is a standard way to assert for the revert message of a contract call when using `hardhat` with its internal evm or any Ethereum JsonRpcRelay:

```
await expect(someContract.revertWithString()).to.be.revertedWith("Some revert message");
```

But the same code does not work when it is configured to use the relay. The difference is in `eth_estimateGas` - normally it throws a JsonRpc error, but in Hedera's context that is not an option. `Developer mode` changes the behaviour of `eth_getTransactionByHash` to throw that same error after it has retrieved the `error_message` from the Mirror node.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


README.md
```diff
@@ -80,6 +80,7 @@ LIMIT_DURATION = 60000
 HBAR_RATE_LIMIT_TINYBAR = 6000000000
 HBAR_RATE_LIMIT_DURATION = 60000
 RATE_LIMIT_DISABLED = false
+DEV_MODE = false
```
Nana-EC, line 83: Add a sentence after here that points to `docs/dev-mode.md` for more info

packages/relay/src/lib/eth.ts
```diff
@@ -940,6 +940,11 @@ export class EthImpl implements Eth {
     const rSig = contractResult.r === null ? null : contractResult.r.substring(0, 66);
     const sSig = contractResult.s === null ? null : contractResult.s.substring(0, 66);
 
+    if (process.env.DEV_MODE && process.env.DEV_MODE === 'true' && contractResult.result === 'CONTRACT_REVERT_EXECUTED') {
```
Nana-EC, line 943: Add a `docs/dev-mode.md` file that highlights dev mode.
Under it add a specific section for ethersjs revert check that highlights the goal of this workaround

Nana-EC: Update the https://github.com/hashgraph/hedera-json-rpc-relay/tree/main/tools#supported-tools section with a note that points to the dev mode feature for ethersjs support

Nana-EC (changes requested):
Changes are great,
Just need to fill in the documentation holes so it's easier understood. Especially since you have a vision for dev mode

This is my suggestion to make sure everyone understands and it's exposed to users.

1. Add a docs/dev-mode.md file as part of your PR
   a. It should have an initial summary with your vision for dev mode.
   b. It should have a sub section that has the above summary focused on ethersjs revert error work around.
2. Add a link under the main README where you set the `DEV_MODE = false` that points to the `dev-mode.md` for more details
3. Update the https://github.com/hashgraph/hedera-json-rpc-relay/tree/main/tools#supported-tools section with a note that points to the dev mode feature for ethersjs support

Nana-EC (approved):
LG

PR #679, opened 2022-11-08
Retry on failed request to mirror node

### LINKED ISSUE #669: Error when deploying on testnet and previewnet with truffle
### Description

Cannot deploy contracts to previewnet or testnet when using truffle. No problem when deploying with hardhat.

Error: 
```
@eth-block-tracker/src/polling.js:51
        const newErr = new Error(`PollingBlockTracker - encountered an error while attempting to update latest block:\n${err.stack}`)
                       ^
Error: PollingBlockTracker - encountered an error while attempting to update latest block:
undefined
    at PollingBlockTracker._performSync (@eth-block-tracker/src/polling.js:51:24)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
UnhandledRejections detected
Promise {
  <rejected> {
    code: -32603,
    message: 'getaddrinfo ENOTFOUND previewnet.hashio.io',
    data: { originalError: [Object] }
  }
} {
  code: -32603,
  message: 'getaddrinfo ENOTFOUND previewnet.hashio.io',
  data: {
    originalError: {
      errno: -3008,
      code: 'ENOTFOUND',
      syscall: 'getaddrinfo',
      hostname: 'previewnet.hashio.io'
    }
  }
}
```

Error from relay:
```
    err: {
      "message": "getaddrinfo ENOTFOUND testnet.mirrornode.hedera.com",
      "stack":
          Error: getaddrinfo ENOTFOUND testnet.mirrornode.hedera.com
              at MirrorNodeClient.handleError (/Users/georgi-lazarov/Documents/GitHub/hedera-json-rpc-relay/packages/relay/dist/lib/clients/mirrorNodeClient.js:126:15)
              at MirrorNodeClient.<anonymous> (/Users/georgi-lazarov/Documents/GitHub/hedera-json-rpc-relay/packages/relay/dist/lib/clients/mirrorNodeClient.js:112:22)
              at rejected (/Users/georgi-lazarov/Documents/GitHub/hedera-json-rpc-relay/packages/relay/dist/lib/clients/mirrorNodeClient.js:25:65)
              at processTicksAndRejections (node:internal/process/task_queues:96:5)
      "statusCode": 567
    }
```
### Steps to reproduce

1. Setup a truffle project with at least 2 big contracts
2. Deploy them to testnet or previewnet

### Additional context

_No response_

### Hedera network

testnet, previewnet

### Version

0.10

### Operating system

_No response_

georgi-l95:

**Description**:

- Introduce axios-retry for retrying failed requests to the mirror node
- Add environmental variables for `delay` and `retry` counts ( 0 `retry` counts disables `axios-retry` )
- Change hbar rate limiter test to one which spends more. This increases the speed, because it's hitting limiter faster.
- Bump timeout for rata limiter acceptance tests, because ofter 240 seconds are not enough.
- Bump hbar rate limiter duration in local acceptance test, because it's failing otherwise.
- Set delays on several places, because some of the tests are hitting rate limit

**Related issue(s)**:

Fixes #669 

**Notes for reviewer**:

**Checklist**


packages/relay/tests/lib/eth.spec.ts
```diff
@@ -536,6 +536,25 @@ describe('Eth calls using MirrorNode', async function () {
...
+      '_status': {
+        'messages': [
+          {
+            'message': 'Block not found'
+          }
+        ]
+      }
+    });
+
+    mock.onGet('blocks?limit=1&order=desc').reply(200, {
```
dimitrovmaksim, line 550: Мисля, че тези можеш да ги чейнеш по този начин `mock.onGet('blocks?limit=1&order=desc').replyOnce(...).onGet('blocks?limit=1&order=desc').replyOnce(...)`

georgi-l95: Yep, that can works as well. Thanks.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -105,6 +106,19 @@ export class MirrorNodeClient {
             timeout: 10 * 1000
         });
 
+        //@ts-ignore
+        axiosRetry(axiosClient, {
+            retries: parseInt(process.env.MIRROR_NODE_RETRIES!) || 3,
+            retryDelay: () => {
```
lukelee-sl, line 112: Should we consider and exponential backoff strategy?

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -105,6 +106,19 @@ export class MirrorNodeClient {
...
+                const delay = parseInt(process.env.MIRROR_NODE_RETRY_DELAY!) || 500;
+                return delay;
+            },
+            retryCondition: (error) => {
+                return !error?.response?.status || error?.response?.status === 404 || error?.response?.status === 504;
```
lukelee-sl, line 117: There might be other codes we care about? For example, maybe retry on 408, 425 and 500 but not 429?

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -105,6 +106,19 @@ export class MirrorNodeClient {
...
+            },
+            shouldResetTimeout: true
```
lukelee-sl, line 119: should we add an onRetry to at least log the retries?

packages/relay/src/lib/clients/mirrorNodeClient.ts
dimitrovmaksim: Should it even retry on 404, in most cases the mirror node returning 404 is an expected behavior?

Nana-EC: Yeah 404 retries make sense since Mirror Node may not have yet synced the files with the data.

packages/relay/src/lib/clients/mirrorNodeClient.ts
Nana-EC: Trace log :)

Nana-EC (commented):
LG. Agreed with some comments that you address.

packages/relay/tests/lib/eth.spec.ts
dimitrovmaksim: Another thing, I believe the way the mocks are currently set, the second one overwrites the first one. If you switch them around the test will fail, even though the first mock would return 200. That's why you have to use `replyOnce`

dimitrovmaksim: Also currently it does not retry at all, because here https://github.com/hashgraph/hedera-json-rpc-relay/blob/b81fe859f329ac7e97a4122833a487a8e9a9b730/packages/relay/tests/lib/eth.spec.ts#L89 an axios instance is created which does not have the retry option. This isntance is passed to the MirrorNodeClient and the Axios mock, so the MirrorNodeClient. We should not create and pass an axios client instance to both so the MirrorNodeClient will create it's own which includes the retry logic. And then the mock instance should be initialised like this `mock = new MockAdapter(mirrorNodeInstance.client, { onNoMatch: "throwException" });`

dimitrovmaksim (changes requested):
Currently the retry test does not work, please look at the comments I left on the `eth.spec.ts` file

packages/relay/src/lib/clients/mirrorNodeClient.ts
georgi-l95: Okay, added array with error codes.

packages/relay/src/lib/clients/mirrorNodeClient.ts
georgi-l95: Added trace log 
Ex. 
`[Request ID: 9263aa9e-605e-4ad3-950e-33665935f381] Retry delay 500 ms`

packages/relay/src/lib/clients/mirrorNodeClient.ts
georgi-l95: Added.

packages/relay/tests/lib/eth.spec.ts
georgi-l95: Done. Now it's using the proper instance.

dimitrovmaksim (approved):

packages/relay/src/lib/clients/mirrorNodeClient.ts
Nana-EC: Okay. Start simple, open a ticket for exponential backoff to come back to later

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -104,7 +105,23 @@ export class MirrorNodeClient {
...
-
+        //@ts-ignore
+        axiosRetry(axiosClient, {
+            retries: parseInt(process.env.MIRROR_NODE_RETRIES!) || 3,
+            retryDelay: (retryCount, error) => {
+                const request = error?.request?._header;
+                const requestId = request ? request.split('\n')[3].substring(11,47) : '';
+                const requestIdPrefix = formatRequestIdMessage(requestId);
+                const delay = (parseInt(process.env.MIRROR_NODE_RETRY_DELAY!) || 500);
+                this.logger.info(`${requestIdPrefix} Retry delay ${delay} ms`);
```
Nana-EC, line 116:
```suggestion
                this.logger.trace(`${requestIdPrefix} Retry delay ${delay} ms`);
```

packages/relay/src/lib/clients/mirrorNodeClient.ts
Nana-EC: Should be trace

Nana-EC (dismissed):
LG. Just make the info log trace

lukelee-sl (dismissed):

packages/relay/src/lib/clients/mirrorNodeClient.ts
georgi-l95: Opened ticket https://github.com/hashgraph/hedera-json-rpc-relay/issues/689

packages/relay/src/lib/clients/mirrorNodeClient.ts
georgi-l95: Yes, you're right. My bad, sorry.

Nana-EC (approved):

PR #693, opened 2022-11-12
Add logic to handle missing transaction index

Nana-EC:

**Description**:
It seems occasionally `transaction_index` can be undefined from mirror node.
Currently the relay throws errors in this case.
While we figure out why

- add logic to manage undefined transaction_index
- add tests

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1068,7 +1068,7 @@ export class EthImpl implements Eth {
   }
 
   static nullableNumberTo0x(input: number | BigNumber): string | null {
-    return input === null || input === undefined ? null : EthImpl.numberTo0x(input);
+    return input == null ? null : EthImpl.numberTo0x(input);
   }
 
   static nanOrNumberTo0x(input: number | BigNumber): string {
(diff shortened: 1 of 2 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1068,7 +1068,7 @@ export class EthImpl implements Eth {
   }
 
   static nullableNumberTo0x(input: number | BigNumber): string | null {
-    return input === null ? null : EthImpl.numberTo0x(input);
+    return input === null || input === undefined ? null : EthImpl.numberTo0x(input);
```
dimitrovmaksim, line 1071: The condition can be combined into `input == null`, this way it checks for both null and undefined

dimitrovmaksim (approved):
LG

georgi-l95 (dismissed):
LG

packages/relay/src/lib/eth.ts
Nana-EC: AH good call, let me make this change and push

dimitrovmaksim (approved):

georgi-l95 (approved):

PR #701, opened 2022-11-16
Support pagination on getLogs requests

### LINKED ISSUE #673: eth_getLogs returns only 25 results
### Description

Currently `eth_getLogs` returns only 25 results, because we are not setting a limit and it defaults to 25

To fix this we should check if there's a `next` query in the mirror node response and do those queries until we get all the logs. To limit the number of request we can also set the limit to 100, which is the current max for the mirror-node.
One problem with this solution is that in combination with this Issue https://github.com/hashgraph/hedera-json-rpc-relay/issues/675 and filtering by block range (fromBlock -> toBlock) this could make getLogs very inefficient and slow

### Steps to reproduce

Execute an `eth_getLogs` request for a block range that contains more than 25 logs

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet

### Version

all

### Operating system

_No response_

georgi-l95:
**Description**:
By default mirror node returns only 25 entries on request and next link for the next entries.

- Use `ILimitOrderParams` by default if none are set. Default: `limit: 100, order: ASC`
- Can pass `ILimitOrderParams` on `eth_getLogs` method, which will override default ones.
- Use next links if there is one and until there is one.
- Add unit and acceptance tests for this functionality.

**Related issue(s)**:

Fixes #673 

**Notes for reviewer**:

**Checklist**


packages/server/src/server.ts
```diff
@@ -410,7 +410,8 @@ app.useRpc('eth_getLogs', async (params: any) => {
     params?.toBlock || null,
     params?.address || null,
     params?.topics || null,
-    requestId
+    requestId,
+    params?.limitParams || undefined
```
dimitrovmaksim, line 414: Per the specifications, `eth_getLogs` does not accept a `limitParams` argument,  so I don't think we should include custom parameters to the Filter object (at this point params value is overwritten with params[0] which is the filter object). And I think this should be internal for the `getLogs` function in the `eth` module.

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -262,6 +264,17 @@ describe('@api RPC Server Acceptance Tests', function () {
...
                 }
             });
+
+            it('should be able to return more than 2 logs with limit of 2 logs per request', async () => {
+                //for the purpose of the test, we are settings limit to 2, and fetching all. 
+                const blocksBehindLatest = Number(await relay.call('eth_blockNumber', [], requestId)) - 50;
+                const logs = await relay.call('eth_getLogs', [{
+                    'fromBlock': blocksBehindLatest,
+                    'toBlock': 'latest',
+                    'limitParams': { limit: 2, order: 'desc'}
```
dimitrovmaksim, line 274: As I mentioned in the previous comment, in my opinion we should not change the `eth_getLogs` by adding custom properties to the Filter object

Nana-EC (changes requested):

packages/server/src/server.ts
Nana-EC: Good point. Agreed, we should be exposing a new prop. Rather we should just always default to ask for the max the mirror node will allow for to reduce the number of trips

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -339,6 +339,17 @@ export class MirrorNodeClient {
             requestId);
     }
 
+    public async getContractResultsLogsByNextLink(
+        link: string,
+        requestId?: string
+    ) {
+        const nextLink = link.split('/api/v1/')[1];
```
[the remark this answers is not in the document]
dimitrovmaksim: It seems because `/api/v1/` is in the beginning js will populate the position 0 with a en empty string ```
```
"/api/v1/some/other/url".split('/api/v1/')
[ '', 'some/other/url' ]
```

We can use `.replace('/api/v1/', '')` instead

packages/server/src/server.ts
georgi-l95: This is optional parameter and it's needed for testing. Otherwise we have to log more than 100 logs per test. This way, I'm limiting it to 2 (for example) and test the functionality easy.

packages/server/tests/acceptance/rpc.spec.ts
georgi-l95: This is optional parameter and it's needed for testing. Otherwise we have to log more than 100 logs per test. This way, I'm limiting it to 2 (for example) and test the functionality easy.

packages/relay/src/lib/clients/mirrorNodeClient.ts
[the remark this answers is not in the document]
georgi-l95: Good point.

packages/relay/src/lib/eth.ts
```diff
@@ -1349,18 +1349,31 @@ export class EthImpl implements Eth {
...
     else {
-      result = await this.mirrorNodeClient.getContractResultsLogs(params, undefined, requestId);
+      result = await this.mirrorNodeClient.getContractResultsLogs(params, limitParams, requestId);
     }
 
     if (!result || !result.logs) {
       return [];
     }
 
+    let unproccesedLogs = result.logs;
```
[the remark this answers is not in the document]
georgi-l95: Yep, good call. Also was wondering about refactoring and pulling out some logic into separate methods for better readability, because now getLogs is overfilled with if statements. Not necessary in this PR, can be in separate. What do you think ?

packages/server/src/server.ts
dimitrovmaksim: Changing the specifications or the code so you can ease testing seems as a wrong approach. This feels more like adjusting the code to the test case. At best it should be an env var or the default, and in the tests the module should be reinitialized with the env var set to a specific number for this specific test.

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
dimitrovmaksim: Was going to say that we should refactor this method, but maybe in a separate PR?

packages/server/tests/acceptance/rpc.spec.ts
dimitrovmaksim: This seems as a wrong approach, imo. There should be another way to do this, like setting the limit as an env var and set the env var before the test case.

packages/server/src/server.ts
dimitrovmaksim: Actually don't you just have to simulate that there's a link in the response in the mock? It shouldn't matter how many logs you return as long as there is a link for the next query.

Nana-EC: Yeah. Agreed with Maksim, add an env variable for default log results request amount. We can default to 100, in tests you can set it smaller. This also allows for easy modification should mirror node modify the limit

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
Nana-EC: Okay, i'm fine with the separate PR if you capture the ticket and note it here :)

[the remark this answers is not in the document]
georgi-l95: Done, i've extracted this logic in separate method in the mirror node client. Opened issue for the rest https://github.com/hashgraph/hedera-json-rpc-relay/issues/704

packages/relay/src/lib/clients/mirrorNodeClient.ts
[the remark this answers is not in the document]
georgi-l95: Done

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -423,6 +434,9 @@ export class MirrorNodeClient {
         if (limitOrderParams) {
             this.setQueryParam(queryParamObject, 'limit', limitOrderParams.limit);
             this.setQueryParam(queryParamObject, 'order', limitOrderParams.order);
+        }else {
```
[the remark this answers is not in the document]
georgi-l95: Done

packages/server/tests/acceptance/rpc.spec.ts
georgi-l95: Done. Added `.env` variable

packages/server/src/server.ts
georgi-l95: Done. Added `.env` variable

Nana-EC (approved):

PR #717, opened 2022-11-24
eth_getBalance workaround

### LINKED ISSUE #694: Add workaround for balances in last 15 mins
### Problem

Currently the Mirror Node doesn't not provide accurate balance to the given timestamp for values in the last 15 mins.
This provides an unexpected error for tools

### Solution

Explore a clean way to get balance for account within last 15 mins. 
Potentially when `eth_getBalance` for a block from the past 15 mins is called

- get the last balance and timestamp for an account from api/v1/balances based on a block timestamp
- get all the transactions for the account between the requested block timestamp and the last balance file using `api/v1/transactions?account.id={accountId}&timestamp=lte:{blockTimestamp}&timestamp=gt:{lastBalanceRecordTimestamp}`. Note this will require looping through 
- Sum up the credits and debits involving the account to get the last balance
- On the relay cache this lastBalance and timestamp with a `ttl` up to the last balance file
- Consult the balance cache on future calls. if present call transaction for credit/debits since that time

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
- Adds a workaround that enables `eth_getBalance` to return historical balance for a block that is up to 15 minutes before the current one. This is done by calculating the value sum of all transactions in the time frame between the END of the specified block and the timestamp for the latest balance update in mirror node at `/balances/{address}` and subtracting it from the current balance.
- Adds an utility function for looping though all pages of a mirror node result (`getPaginatedResults`) .
- Removes a similar utility that was only applicable to `eth_getLogs`. Refactors `eth_getLogs` to use `getPaginatedResults`.

**Related issue(s)**:

Fixes #694 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -196,6 +197,22 @@ export class MirrorNodeClient {
         throw new MirrorNodeClientError(error.message, effectiveStatusCode);
     }
 
+    async getPaginatedResults(url: string, pathLabel: string, resultProperty: string, allowedErrorStatuses?: number[], requestId?: string, results = [], page = 1) {
```
georgi-l95, line 200: There is already `pageAllResult` method for this functionality. It's done via iteration, while this one is done via recursion. Can you decide which one we can use for all operations needing pagination and remove the other one ?

Nana-EC: Would also prefer just one solution here. Use iteration for simplicity.

Nana-EC: Update `pageAllResult` to use this improved logic that takes in a mirror endpoint etc.
Let's open a ticket to handle this in 0.14.
For now let's get this into 0.13 as is since it's functional

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -210,10 +228,28 @@ export class MirrorNodeClient {
...
+            'transactions',
+            [400, 404],
+            requestId
+        );
+    }
+
+    public async getBalanceAtTimestamp(accountId: string, timestamp?: string, requestId?: string) {
+        const queryParamObject = {};
+        this.setQueryParam(queryParamObject, 'account.id', accountId);
+        if (timestamp && timestamp !== "") {
```
Nana-EC, line 250: Move this check into `setQueryParam()` so it's done for all cases

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -231,8 +267,12 @@ export class MirrorNodeClient {
     public async getBlocks(blockNumber?: number | string[], timestamp?: string, limitOrderParams?: ILimitOrderParams, requestId?: string) {
         const queryParamObject = {};
         this.setQueryParam(queryParamObject, 'block.number', blockNumber);
-        this.setQueryParam(queryParamObject, 'timestamp', timestamp);
-        this.setLimitOrderParams(queryParamObject, limitOrderParams);
+        if (timestamp) {
```
Nana-EC, line 270: These checks should be handled by `setQueryParam()` already.
If not move it into there

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -316,10 +356,14 @@ export class MirrorNodeClient {
         limitOrderParams?: ILimitOrderParams,
         requestId?: string) {
         const queryParams = this.prepareLogsParams(contractLogsResultsParams, limitOrderParams);
-        return this.request(`${MirrorNodeClient.GET_CONTRACT_RESULT_LOGS_ENDPOINT}${queryParams}`,
+
+        return this.getPaginatedResults(
+            `${MirrorNodeClient.GET_CONTRACT_RESULT_LOGS_ENDPOINT}${queryParams}`,
             MirrorNodeClient.GET_CONTRACT_RESULT_LOGS_ENDPOINT,
+            'logs',
```
Nana-EC, line 363: Should be constant

packages/relay/src/lib/eth.ts
```diff
@@ -571,7 +571,34 @@ export class EthImpl implements Eth {
...
+                let balanceFromTxs = 0;
+
+                if (balance.balances?.length) {
+                  currentBalance = balance.balances[0].balance;
+                  currentTimestamp = balance.timestamp;
+                }
+
+                let transactionsInTimeWindow = await this.mirrorNodeClient.getTransactionsForAccount(
+                    mirrorAccount.account,
+                    block.timestamp.from,
```
Nana-EC, line 587: Q: should this be `from` or `to`?
Just checking we don't replay the transaction within a block if the retrieved balance already includes that.
We should verify this in an E2E test.

Nana-EC (commented):
Looking good.
Clarifying question and suggestions for test coverage.
Let's just make sure there's no offset in the balance and we can modify the approach in 0.14

Nana-EC:
Also @Ivo-Yankov please update the PR with descriptions of changes

packages/relay/src/lib/eth.ts
Ivo-Yankov: After some further consideration I changed it to `to`, since it makes more sence. When requesting the balance at block X, all transactions from within that block should be accounted for in the calculated balance, hence we should use `to` in this case.

packages/relay/src/lib/eth.ts
```diff
@@ -571,7 +571,34 @@ export class EthImpl implements Eth {
...
+                for(const tx of transactionsInTimeWindow) {
+                  for (const transfer of tx.transfers) {
+                    if (transfer.account === mirrorAccount.account && !transfer.is_approval) {
+                      balanceFromTxs += transfer.amount;
+                    }
+                  }
+                }
+
+                balanceFound = true;
+                weibars = (currentBalance - balanceFromTxs) * constants.TINYBAR_TO_WEIBAR_COEF;
```
Nana-EC, line 601: Assuming a balance file timestamp of T1 and a block timestamp of T2 and these are within 15 mins of each other.
Can we make sure we have test coverage for the following cases

- T1 < T2
- T1 > T2
- T1 = T2

Also the above where
- it's a mixture of debits and credits
- it's all debits
- it's all credits

Nana-EC: Can do these in `eth.spec.ts` no need for it in acceptance test

Nana-EC (approved):
LG
Unblocking

Nana-EC (approved):

PR #719, opened 2022-11-28
eth_getLogs - add support for multiple addresses

Ivo-Yankov:
**Description**:
- Allows filtering of `eth_getLogs` results by multiple addresses. It is now possible to pass an array value for the `address` param.
- Refactors `eth_getLogs` by reducing the function complexity.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -1281,81 +1281,117 @@ export class EthImpl implements Eth {
...
+        return false;
       }
+
+      throw e;
+    }
+
+    return true;
+  }
+
+  private async applyBlockRangeParams(params: any, fromBlock: string | null, toBlock: string | null, requestId?: string) {
```
Nana-EC, line 1307: Can we name this something more suggesting of what it does.
My understanding is it resolves the `to` and `from` block tags to determine the appropriate timestamp values to use.

The `params` is then utilized later by reference.
You're essentially returning 2 items from the function.

Please rename and maybe add some comments to make it clear what's going on

Nana-EC (commented):
Logic is good, just a bit cryptic so we should update the noted method name and comments to make it easy to understand

packages/relay/src/lib/eth.ts
```diff
@@ -1281,81 +1281,117 @@ export class EthImpl implements Eth {
...
+      params.timestamp = [];
 
-      if (!fromBlock && !toBlock) {
-        const blockResponse = await this.getHistoricalBlockResponse("latest", true, requestId);
-        fromBlockNum = parseInt(blockResponse.number);
-        toBlockNum = parseInt(blockResponse.number);
-        params.timestamp = [`gte:${blockResponse.timestamp.from}`, `lte:${blockResponse.timestamp.to}`];
-      } else {
-          params.timestamp = [];
+      // Use the `toBlock` if it is the only passed tag, if not utilize the `fromBlock`
```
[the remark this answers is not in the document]
Nana-EC: Let's return an error like Alchemy but make sure the error notes that it's due to a missing `fromBlock` when a `toBlock` was provided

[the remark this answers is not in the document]
dimitrovmaksim: For context:
When `toBlock` param value is the tag "latest" or "pending" -> returns logs for `latest` block
When `toBlock` param is hex number >= "latest" block number -> returns logs for `latest` block
When `toBlock` param is hex number < "latest" block number -> `{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"One of the blocks specified in filter (fromBlock, toBlock or blockHash) cannot be found."}}`

Nana-EC (approved):
LG

packages/relay/src/lib/eth.ts
```diff
@@ -1305,85 +1305,127 @@ export class EthImpl implements Eth {
...
+
+  private async validateBlockRangeAndAddTimestampToParams(params: any, fromBlock: string | null, toBlock: string | null, requestId?: string) {
+    const blockRangeLimit = Number(process.env.ETH_GET_LOGS_BLOCK_RANGE_LIMIT) || constants.DEFAULT_ETH_GET_LOGS_BLOCK_RANGE_LIMIT;
+    if (!fromBlock && !toBlock) {
+      const blockResponse = await this.getHistoricalBlockResponse("latest", true, requestId);
+      params.timestamp = [`gte:${blockResponse.timestamp.from}`, `lte:${blockResponse.timestamp.to}`];
     } else {
-      const blockRangeLimit = Number(process.env.ETH_GET_LOGS_BLOCK_RANGE_LIMIT) || constants.DEFAULT_ETH_GET_LOGS_BLOCK_RANGE_LIMIT;
+      // toBlock is not specified or is `latest` or `pending`
+      if (EthImpl.blockTagIsLatestOrPending(toBlock)) {
```
dimitrovmaksim, line 1338: Can't we combine this if with the else if, for example something like `if (!EthImpl.blockTagIsLatestOrPending(toBlock) && toBlockNumber < latestBlockNumber`, imo it is unnecessary to assign "latest" to the `toBlock` if it already is either `latest` or `pending`

Ivo-Yankov: It could also be `null` in this case.

dimitrovmaksim: The name `blockTagIsLatestOrPending` does not suggest it checks for null, and personally I don't think it should.

dimitrovmaksim: Maybe it's better to handle this here, https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/server/src/server.ts#L409-L410 instead of `|| null` to be `|| "latest"`

dimitrovmaksim (approved):

Nana-EC (approved):

georgi-l95 (approved):

PR #733, opened 2022-12-02
Handle empty state_changes array in contract details

### LINKED ISSUE #732: Failed to send/recv `storage` err=GetStorage() when testing with Foundry
### Description

I have successfully set up Foundry to work with Hedera by configuring it to work with npm modules. We can run unit test, and complete fuzz testing. Github repo [here](https://github.com/a-ridley/hedera-smart-contract-testing-with-foundry).

The issue lies when attempting to use Foundry's fork testing to test against a smart contract that is deployed on testnet. In my test set up, I provided my smart contract's evm address (which I grabbed from hashscan).

Below is a snippet of my test set up.
```
function setUp() public {
        counter = Counter(0x0000000000000000000000000000000002Eba2A0);
        //counter.setNumber(0);
    }
```

The expected output is for my tests to run successfully. However, the result of running my tests gives the following error:

```
2022-12-01T18:28:36.945764Z ERROR sharedbackend: Failed to send/recv `storage` err=GetStorage(0x0000000000000000000000000000000002eba2a0, 0, 
(code: -32603, message: [Request ID: 6153887f-a84f-42e9-93c7-8fcb32047a21] Error invoking RPC: , data: None)) address=0x0000000000000000000000000000000002eba2a0 index=0

Running 1 test for test/foundry/Counter.t.sol:CounterTest
[FAIL. Reason: EvmError: FatalExternalError] testIncrement() (gas: 9223372036854754743)
Test result: FAILED. 0 passed; 1 failed; finished in 38.01s

Failing tests:
Encountered 1 failing test in test/foundry/Counter.t.sol:CounterTest
[FAIL. Reason: EvmError: FatalExternalError] testIncrement() (gas: 9223372036854754743)
```

### Steps to reproduce

[simulate-swap-hedera.zip](https://github.com/hashgraph/hedera-json-rpc-relay/files/10137040/simulate-swap-hedera.zip)
2. edit the .env file
4. run `HEDERA_RPC_URL=https://testnet.hashio.io/api `
5. run `forge test --fork-url $HEDERA_RPC_URL -vv --fork-block-number <block number>`

### Additional context

**note: block number needs to be at least 15 minutes in the past or you'll get the error below.**
`(code: -32007, message: [Request ID: 3804b205-6ef8-471d-92a2-e640d9131558] Historical balance data is available only after 15 minutes., data: None)) address=0x0000000000000000000000000000000002eba2a0`

### Hedera network

testnet

### Version

v2.18.5

### Operating system

macOS

lukelee-sl:

Handle the case when the state_changes field returned as an empty array. 

Fixes #732 

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -482,6 +482,7 @@ export class EthImpl implements Eth {
    */
   async getStorageAt(address: string, slot: string, blockNumberOrTag?: string | null, requestId?: string) : Promise<string> {
     const requestIdPrefix = formatRequestIdMessage(requestId);
+    this.logger.trace(`${requestIdPrefix} getStorageAt(address=${address}, slot=${slot}, blockNumberOrTag=${blockNumberOrTag})`);
     let result = EthImpl.zeroHex32Byte; // if contract or slot not found then return 32 byte 0
     const blockResponse  = await this.getHistoricalBlockResponse(blockNumberOrTag, false);
 
@@ -498,7 +499,7 @@ export class EthImpl implements Eth {
       // retrieve the contract result details
       await this.mirrorNodeClient.getContractResultsDetails(address, contractResult.results[0].timestamp)
         .then(contractResultDetails => {
-          if (contractResultDetails?.state_changes != null) {
+          if (contractResultDetails?.state_changes != null && Array.isArray(contractResultDetails?.state_changes) && contractResultDetails?.state_changes.length > 0) {
             // filter the state changes to match slot and return value
             const stateChange = contractResultDetails.state_changes.find(stateChange => stateChange.slot === slot);
             result = stateChange.value_written;
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -499,7 +499,7 @@ export class EthImpl implements Eth {
       // retrieve the contract result details
       await this.mirrorNodeClient.getContractResultsDetails(address, contractResult.results[0].timestamp)
         .then(contractResultDetails => {
-          if (contractResultDetails?.state_changes != null && Array.isArray(contractResultDetails?.state_changes) && contractResultDetails?.state_changes.length > 0) {
+          if (EthImpl.isArrayNonEmpty(contractResultDetails?.state_changes)) {
             // filter the state changes to match slot and return value
             const stateChange = contractResultDetails.state_changes.find(stateChange => stateChange.slot === slot);
             result = stateChange.value_written;
@@ -1401,4 +1401,9 @@ export class EthImpl implements Eth {
 
     return contractAddress;
   }
+
+  static isArrayNonEmpty(input: any): boolean {
+    return Array.isArray(input) && input.length > 0;
+  }
+
 }
--- a/packages/relay/tests/lib/eth.spec.ts
+++ b/packages/relay/tests/lib/eth.spec.ts
@@ -378,6 +378,18 @@ describe('Eth calls using MirrorNode', async function () {
     }
   };
 
+  const defaultDetailedContractResultsNullStateChange = {
+    ...defaultDetailedContractResults, ...{
+      'state_changes' : null
+    }
+  };
+
+  const defaultDetailedContractResultsEmptyArrayStateChange = {
+    ...defaultDetailedContractResults, ...{
+      'state_changes' : []
+    }
+  };
+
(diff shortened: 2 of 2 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -498,7 +499,7 @@ export class EthImpl implements Eth {
       // retrieve the contract result details
       await this.mirrorNodeClient.getContractResultsDetails(address, contractResult.results[0].timestamp)
         .then(contractResultDetails => {
-          if (contractResultDetails?.state_changes != null) {
+          if (contractResultDetails?.state_changes != null && Array.isArray(contractResultDetails?.state_changes) && contractResultDetails?.state_changes.length > 0) {
```
david-bakin-sl, line 502: this one-liner should be a utility function somewhere, isn't it somewhere else already?  (typescript doesn't have it as a built-in?)

Nana-EC, line 502: `Array.isArray(null)` and `Array.isArray(undefined)` returns false so we can drop the initial null check
```suggestion
          if (Array.isArray(contractResultDetails?.state_changes) && contractResultDetails?.state_changes.length > 0) {
```

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -2219,6 +2231,38 @@ describe('Eth calls using MirrorNode', async function () {
...
+      defaultDetailedContractResultsNullStateChange
+      mock.onGet(`blocks/${blockNumber}`).reply(200, defaultBlock);
+      mock.onGet(`contracts/${contractAddress1}/results?timestamp=lte:${defaultBlock.timestamp.to}&limit=1&order=desc`).reply(200, defaultContractResults);
+      mock.onGet(`contracts/${contractAddress1}/results/${contractTimestamp1}`).reply(200, defaultDetailedContractResultsNullStateChange);
+      let hasError = false;
+      try {
+        await ethImpl.getStorageAt(contractAddress1, defaultDetailedContractResults.state_changes[0].slot, EthImpl.numberTo0x(blockNumber));
+      } catch (e: any) {
+        hasError = true;
+        expect(e.code).to.equal(-32001);
```
david-bakin-sl, line 2244: Is this magic number something I should just know were I working with typescript?

Nana-EC (commented):
LG, offered a simplification

packages/relay/tests/lib/eth.spec.ts
Nana-EC: It's the [RESOURCE_NOT_FOUND error code](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/errors/JsonRpcError.ts#L118)

Nana-EC: Might be good to use `predefined.RESOURCE_NOT_FOUND.code` in the check instead

packages/relay/src/lib/eth.ts
lukelee-sl: no built-in function that I am aware of

packages/relay/tests/lib/eth.spec.ts
lukelee-sl: done

packages/relay/src/lib/eth.ts
david-bakin-sl: (stupid mis-hit button ate my comment) the concept being expressed here "array must not be empty" is needlessly complicated when written out due to javascript/java/c# stupidity of allowing arrays to be "null" as a distinct thing to test syntactically than array being/not being empty.  So here, even with this simplification, it takes an entire line to express and a pattern match in your brain to figure out the idiom - again complicated because it is done w.r.t. to a named field of an object which itself takes 40 characters to express and you have to say it twice.  My suggestion would be to extract it into a named method - even though the method itself might be a one-liner - so that every _use_ is simplified and the name alone is sufficient to tell you what's going on - you don't have to recognize the pattern.  (Except of course the language might not actually allow you to easily express the method _call_.)

lukelee-sl: I understand the sentiment.  With issues like this, I like to ask myself "does doing this lessen the likelihood of a mistake?".  If so then I create the helper static method.  I think one needs to be careful or you could spend so much time creating helper/macro methods, the code becomes difficult to follow (I've followed too many #define references in header files in c++ to find at the end something trivial).  I'll take your council in this case.

david-bakin-sl:
LGTM

Nana-EC (approved):
LGTM

PR #734, opened 2022-12-02
Handle empty state_changes array in contract details (#733)

lukelee-sl:
* Handle empty state_change array in contract details

cherry pick #733 

**Description**:
Cherry pick commit which addresses bug in issue #732


Nana-EC (approved):
LG

PR #736, opened 2022-12-02
Bump 0.13.0-rc3

lukelee-sl:

**Description**:

Bump version to 0.13.0-rc3


Nana-EC (approved):
LG

david-bakin-sl:
Late to the party but confirmed that only 6 files have `0.13.0-rc` string so all necessary files were bumped.

PR #737, opened 2022-12-02
Update the subgraph example

dimitrovmaksim:

**Description**:
Updates the subgraph example with ERC20, ERC721 and GravatarRegistry contract examples

**Related issue(s)**:

Depends on https://github.com/hashgraph/hedera-json-rpc-relay/pull/719

**Notes for reviewer**:

**Checklist**


tools/subgraph-example/contracts/GravatarRegistry.sol
```diff
@@ -0,0 +1,68 @@
...
+
+    gravatars[id].imageUrl =  _imageUrl;
+    emit UpdatedGravatar(id, msg.sender, gravatars[id].displayName, _imageUrl);
+  }
+
+  // the gravatar at position 0 of gravatars[]
+  // is fake
+  // it's a mythical gravatar
+  // that doesn't really exist
+  // dani will invoke this function once when this contract is deployed
```
Nana-EC, line 63: `anyone` or `someone` may invoke this post deployment

tools/subgraph-example/package.json
```diff
@@ -1,20 +1,49 @@
 {
-  "name": "@hedera-subgraph-greeter-example",
-  "license": "Apache-2.0",
+  "name": "hardhat-project",
+  "devDependencies": {
```
Nana-EC, line 3: Q: lots of new `devDependencies`.
Just making sure but do we need all them or was this added by a dependency package

Nana-EC (commented):
LG on 1st pass.
Will circle back for 2nd

tools/subgraph-example/contracts/GravatarRegistry.sol
dimitrovmaksim: I updated the contract, in the original contracts the function looks like this 
```
  // the gravatar at position 0 of gravatars[]
  // is fake
  // it's a mythical gravatar
  // that doesn't really exist
  // dani will invoke this function once when this contract is deployed
  // but then no more
  function setMythicalGravatar() public {
    require(msg.sender == 0x8d3e809Fbd258083a5Ba004a527159Da535c8abA);
    gravatars.push(Gravatar(0x0, " ", " "));
  }
```

I updated it to be executed once in the constructor and no-one can execute it outside of the contract

tools/subgraph-example/package.json
dimitrovmaksim: Most of them are auto-generated by hardhat

dimitrovmaksim:
Also opened a PR with some tests that is based on this PR's branch https://github.com/hashgraph/hedera-json-rpc-relay/pull/764

tools/subgraph-example/README.md
```diff
@@ -1,35 +1,58 @@
 # Subgraph example
 
-Simple subgraph indexing the `GreeterSet` event of emitted by the [modified] Greeter contract
+Hardhat based project containing example subgraph indexing the following contracts and events:
+
+ExampleERC721 -> Transfer
```
Nana-EC, line 5:
```suggestion
ExampleERC721 -> Mint
```

tools/subgraph-example/subgraph/subgraph.yaml
```diff
@@ -0,0 +1,69 @@
+specVersion: 0.0.5
+schema:
+  file: ./schema.graphql
+dataSources:
+  - kind: ethereum
```
Nana-EC, line 5: Q: Could this be changed to Hedera if not in summary what would be needed to make this work for a self hosted option

tools/subgraph-example/subgraph/subgraph.yaml
```diff
@@ -0,0 +1,69 @@
...
+    name: ExampleERC721
+    network: local
```
Nana-EC, line 7: Q: Does this work is pointing to previewnet for example and if so do you change the network name?

Nana-EC (dismissed):
LG
Just some questions to help capture possibilities

tools/subgraph-example/README.md
dimitrovmaksim: When minting an NFT it still emits a `Transfer` event where `from` is the zero address

tools/subgraph-example/subgraph/subgraph.yaml
dimitrovmaksim: If you change the `network` property it should also change the network which you want to target - https://github.com/graphprotocol/graph-node/blob/master/docs/subgraph-manifest.md#15-data-source

If you run the graph-node using the `docker-compose` file, I think it only supporst a single network and you have to change this line `https://github.com/graphprotocol/graph-node/blob/master/docker/docker-compose.yml#L22`, where you need to replace `mainnet` with the name of the network, and the url with the network endpoint. 

If you buid it from source, you can specify multiple networks https://github.com/graphprotocol/graph-node/blob/master/docs/config.md#configuring-ethereum-providers, then just chaning the `network` property to the ones of the the specified should tharget that network automatically

tools/subgraph-example/subgraph/subgraph.yaml
dimitrovmaksim: Here `ethereum` is the network type or as they also call it the `protocol`, it actually means it's an `evm compatible` network. Basically it tells all the tools how to treat that subgraph, what triggers it should look for (events, calls, blocks), how to convert Solidity types to AssemblyScript types (the language the subgraphs are written) and vice versa, how to call Solidity contracts, etc. 

It will most likely require forking and renaming it in the graph-node, graph-cli and graph-ts

tools/subgraph-example/README.md
```diff
@@ -1,35 +1,58 @@
...
 
-Run `npm install` or `yarn [install]`
+#### ERC20:
+`npx hardhat deployERC20`
+`npx hardhat transferERC20`
+
+#### Gravatar:
+`npx hardhat deployGravatar`
+`npx hardhat createGravatar`
+
```
georgi-l95, line 50: nit: Add `&&` between commands, or put them on new line. VS Code preview shows them together.

tools/subgraph-example/hardhat.config.ts
```diff
@@ -0,0 +1,137 @@
...
+  await contract.deployed();
+  const deployTx = await contract.deployTransaction.wait();
+  const address = deployTx.contractAddress;
+
+  await hre.run("graph", { contractName, address });
+  updateStartBlock(contractName, deployTx.blockNumber, hre);
+
+  console.log("CONTRACT DEPLOYED AT:")
+  console.log(address);
+  console.log(contract.address);
```
georgi-l95, line 33: nit: Those are the same in my experience. Same with the ones in ERC721 and Gravatar.

georgi-l95 (approved):
LG, couple of nits.

tools/subgraph-example/hardhat.config.ts
dimitrovmaksim: Not on the mirror node version used in the custom config, because the latest mirror node image does not include an important fix, so I can't currently use it.

georgi-l95: Yep, I figured that the reason was mirror-node version. We can revisit this, when the fix is released.

tools/subgraph-example/subgraph/subgraph.yaml
Nana-EC: Can we open a ticket with these comments in it some we can have a later README update that provides documentation to help in this scenario

tools/subgraph-example/subgraph/subgraph.yaml
Nana-EC: Gotcha, wish they'd name it `evm-protocol` instead :)

tools/subgraph-example/README.md
Nana-EC: Ah you're correct, this is highlighting events.

Nana-EC (approved):

tools/subgraph-example/hardhat.config.ts
dimitrovmaksim: @Nana-EC https://github.com/hashgraph/hedera-mirror-node/issues/5005 and the PR https://github.com/hashgraph/hedera-mirror-node/pull/5006, but the fix did not get into 0.70.1

georgi-l95 (approved):
LG

PR #738, opened 2022-12-03
Return 32 byte 0 on empty state_changes array

### LINKED ISSUE #732: Failed to send/recv `storage` err=GetStorage() when testing with Foundry

lukelee-sl:

**Description**:
Return 32 byte 0 string when the returned state_changes array is empty.  This is similar behavior to other test chains.

**Related issue(s)**:

Fixes #732 

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -499,13 +499,14 @@ export class EthImpl implements Eth {
       // retrieve the contract result details
       await this.mirrorNodeClient.getContractResultsDetails(address, contractResult.results[0].timestamp)
         .then(contractResultDetails => {
+          if(contractResultDetails === null) {
+            throw predefined.RESOURCE_NOT_FOUND(`Contract result details for contract address ${address} at timestamp=${contractResult.results[0].timestamp}`);
+          }
           if (EthImpl.isArrayNonEmpty(contractResultDetails?.state_changes)) {
             // filter the state changes to match slot and return value
             const stateChange = contractResultDetails.state_changes.find(stateChange => stateChange.slot === slot);
             result = stateChange.value_written;
-          } else {
-            throw predefined.RESOURCE_NOT_FOUND(`Contract result details for contract address ${address} at timestamp=${contractResult.results[0].timestamp}`);
-          }
+          } 
         })
         .catch((e: any) => {
           this.logger.error(
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -502,7 +502,7 @@ export class EthImpl implements Eth {
           if(contractResultDetails === null) {
             throw predefined.RESOURCE_NOT_FOUND(`Contract result details for contract address ${address} at timestamp=${contractResult.results[0].timestamp}`);
           }
-          if (EthImpl.isArrayNonEmpty(contractResultDetails?.state_changes)) {
+          if (EthImpl.isArrayNonEmpty(contractResultDetails.state_changes)) {
             // filter the state changes to match slot and return value
             const stateChange = contractResultDetails.state_changes.find(stateChange => stateChange.slot === slot);
             result = stateChange.value_written;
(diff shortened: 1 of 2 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -499,13 +499,14 @@ export class EthImpl implements Eth {
       // retrieve the contract result details
       await this.mirrorNodeClient.getContractResultsDetails(address, contractResult.results[0].timestamp)
         .then(contractResultDetails => {
+          if(contractResultDetails === null) {
```
lukelee-sl, line 502: we do want to do this null check here because we want to throw the RESOURCE_NOT_FOUND exception if the contract details are not found at all.

packages/relay/src/lib/eth.ts
```diff
@@ -499,13 +499,14 @@ export class EthImpl implements Eth {
...
+            throw predefined.RESOURCE_NOT_FOUND(`Contract result details for contract address ${address} at timestamp=${contractResult.results[0].timestamp}`);
+          }
           if (EthImpl.isArrayNonEmpty(contractResultDetails?.state_changes)) {
             // filter the state changes to match slot and return value
             const stateChange = contractResultDetails.state_changes.find(stateChange => stateChange.slot === slot);
             result = stateChange.value_written;
-          } else {
```
lukelee-sl, line 506: removing this else clause causes us to fall through to return the default 32byte 0 hex value.

Nana-EC (approved):
LG

packages/relay/src/lib/eth.ts
david-bakin-sl: BTW, now can lose the `?` on line 502 ...

lukelee-sl: yes - nice catch

Nana-EC (approved):

PR #763, opened 2022-12-09
Refactor eth_call to use mirror node

### LINKED ISSUE #721: Update the relay to call mirror-node for eth_call
### Problem

As per https://hips.hedera.com/hip/hip-584:
`As a user, I want to perform eth_call calls from the Ethereum JSON-RPC standard`
Information will be made available via new REST API from the mirror-node

### Solution

A new /api/v1/contracts/call POST REST API will be added to execute read-only queries. We want to start calling this API. 
We also want this done behind an environment variable in order to have it configurable on different environments.

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
- Adds a new env variable that controls the behaviour of `eth_call`: `ETH_CALL_CONSENSUS`
- Adds the option for `eth_call` to query the mirror node instead of consensus.
- Adds logic to `mirrorNodeClient` for `POST` requests.
- To enable mirror node integration set `ETH_CALL_CONSENSUS=false`

**Related issue(s)**:

Fixes #721 

**Notes for reviewer**:
Depends on mirror node EVM module implementation.
Still needs some code cleanup.

**Checklist**


Nana-EC:
@Ivo-Yankov for now let's do let of a refactor and more of an additional pathway
- add a config flag that says whether the relay should call the consensus node or not.
- If true then the behavior remain the same
- IF false then let it call the new flow you're adding
- Moving the logic into 2 separate methods in `eth.ts` will make this change simpler

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -130,23 +135,25 @@ export class MirrorNodeClient {
...
 
-    constructor(baseUrl: string, logger: Logger, register: Registry, axiosClient?: AxiosInstance) {
-        if (axiosClient !== undefined) {
-            this.baseUrl = '';
-            this.client = axiosClient;
-        } else {
-            if (!baseUrl.match(/^https?:\/\//)) {
-                baseUrl = `https://${baseUrl}`;
-            }
+    constructor(restUrl: string, logger: Logger, register: Registry, restClient?: AxiosInstance, web3Url?: string, web3Client?: AxiosInstance) {
```
Nana-EC, line 138: This works but i wonder in future whether it'd be better to have separate `mirrorNodeRESTClient.ts` and `mirrorNodeWeb3Client.ts` files.
Since they're separate modules and can be changed and deployed separate from the other might be good to have then separate.
No need to address here.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -130,23 +135,25 @@ export class MirrorNodeClient {
...
-            }
+        if (restClient !== undefined) {
+            this.restUrl = '';
+            this.web3Url = '';
 
-            baseUrl = `${baseUrl}api/v1/`;
+            this.restClient = restClient;
+            this.web3Client = !!web3Client ? web3Client : restClient;
+        } else {
+            restUrl = this.buildUrl(restUrl);
```
Nana-EC, line 150: I forget why we try to update the param passed in.
Why not
```suggestion
            this.restUrl = this.buildUrl(restUrl);
```

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -162,47 +169,79 @@ export class MirrorNodeClient {
...
         let ms;
         try {
-            const response = await this.client.get(path, {
-                headers:{
-                    'requestId': requestId || ''
-                }
-            });
-            ms = Date.now() - start;
-            this.logger.debug(`${requestIdPrefix} [GET] ${path} ${response.status} ${ms} ms`);
+            let response;
```
Nana-EC, line 192: nit: Define headers once
```suggestion
            let response;
            const headers = {
                    headers:{
                        'requestId': requestId || ''
                    }
                };
```

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -160,5 +160,13 @@ export const predefined = {
     name: 'Value too low',
     code: -32602,
     message: 'Value below 10_000_000_000 wei which is 1 tinybar'
-  })
+  }),
+  'INVALID_CONTRACT_ADDRESS':  (address) => {
+    const message = address && address.length ? ` Expected length of 42 chars but was ${address.length}.` : ''
```
Nana-EC, line 165: nit:
```suggestion
    const message = address && address.length ? `Expected length of 42 chars but was ${address.length}.` : ''
```

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -160,5 +160,13 @@ export const predefined = {
...
+    return new JsonRpcError({
+      name: 'Invalid Contract Address',
+      code: -32012,
+      message: `Invalid Contract Address: ${address}.${message}`
```
Nana-EC, line 169: nit: per above
```suggestion
      message: `Invalid Contract Address: ${address}. ${message}`
```

packages/relay/src/lib/errors/JsonRpcError.ts
Ivo-Yankov: This prevents an empty space char at the end of the error message if `address` is null

Nana-EC:
@Ivo-Yankov can you address the comments in a separate PR when you get a moment

lukelee-sl:
Hi @Ivo-Yankov,  I tried to rebase to main in order to try to get this into 0.17.0-rc1.  Unfortunately, I was not able to get the acceptance test batch 2 to pass.  Can you please take a look?  Feel free to revert my rebase if you feel that's the right thing to do.

lukelee-sl (approved):
LGTM.  It's nice you put in the environment variable for turning this on/off.

Nana-EC (approved):
LG
@Ivo-Yankov you didn't address some of the more format related comments I had.
Approving so this can get cherry picked to 0.17 but please address the other items in a separate PR

Issue #703, opened 2022-11-17
Refactor eth_getStorageAt


Ivo-Yankov:
### Problem

Currently `eth_getStorageAt` retrieves the slot value from `/api/v1/contracts/{evmAddress}/results` endpoint, which only contains `state_changes`. If the slot was not changed in the `ContractResult` the returned value would be `0x00000....0`, although it may have an actual value.

### Solution

Implement the mirror node method `/contracts/evmAddress/state`.

dimitrovmaksim:
Also if the slots are zero padded in the mirror-node, we should check the padding and add the missing zeros. E.g. eth_getStorageAt accepts `0x1`, `0x00001` and so on, which should be resolved to format that is returned by the mirror node.

stoqnkpL:
We need to verify this use case:
- contact x slot y is updated in block b
- blocks c, d, e don't change contract x's state
- block f updates contract x slot y again
- we make a query to eth_getStorageAt asking for contract x's state in block d
- verify the returned value is the latest slot value updated in block b

stoqnkpL:
Verify if `/contracts/{evmAddress}/state` supports providing the blocknumber or timestamp as argument.

PR #783, opened 2022-12-19
Refactor `eth_getStorageAt` to use new endpoint

georgi-l95:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -64,6 +64,7 @@ export class MirrorNodeClient {
     private static GET_CONTRACT_RESULT_ENDPOINT = 'contracts/results/';
     private static GET_CONTRACT_RESULT_LOGS_ENDPOINT = 'contracts/results/logs';
     private static GET_CONTRACT_RESULT_LOGS_BY_ADDRESS_ENDPOINT = `contracts/${MirrorNodeClient.ADDRESS_PLACEHOLDER}/results/logs`;
+    private static GET_STATE_ENDPOINT = '/state';
     private static GET_CONTRACT_RESULTS_ENDPOINT = 'contracts/results';
     private static GET_NETWORK_EXCHANGERATE_ENDPOINT = 'network/exchangerate';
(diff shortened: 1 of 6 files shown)


dimitrovmaksim:
Wouldn't this still have the same issue as described here https://github.com/hashgraph/hedera-json-rpc-relay/issues/703, when the state at specific block is requested?

georgi-l95:
> Wouldn't it still have the same issue as described here #703, when the state at specific block is requested?

When specific block is passed, we are searching for state change from this exact block. We can't iterate through all previous blocks until we find a state change and mirror node is not supporting this as well. A change in mirror node must be introduced in order to support - if there is no state change in specific block, give me the latest before this block.

dimitrovmaksim:
> When specific block is passed, we are searching for state change from this exact block. We can't iterate through all previous blocks until we find a state change and mirror node is not supporting this as well. A change in mirror node must be introduced in order to support - if there is no state change in specific block, give me the latest before this block.

Yeah, there should be an endpoint that filters state_changes by address, slot and timestamp

georgi-l95:
> IMO, we should either add this somewhere in the docs so it is known that it does not work on 100%, or we should disable the option to request state for specific blocks until it is fixed, and return some kind of error/message
> 
Could be added to the docs. Agreed.
> > A change in mirror node must be introduced in order to support - if there is no state change in specific block, give me the latest before this block.
> 
> Yeah, there should be an endpoint that filters state_changes by address, slot and timestamp

Yes, current endpoint for state, should be extended to accept timestamp and support previous state changes.

georgi-l95:
> > IMO, we should either add this somewhere in the docs so it is known that it does not work on 100%, or we should disable the option to request state for specific blocks until it is fixed, and return some kind of error/message
> 
> Could be added to the docs. Agreed.
> 
> > 
> > 
> > Yeah, there should be an endpoint that filters state_changes by address, slot and timestamp
> 
> Yes, current endpoint for state, should be extended to accept timestamp and support previous state changes.

lukelee-sl (approved):
LG. Let's add a ticket to track updating the state endpoint to support specific blocks on mirror node and maybe one for updating the doc

PR #788, opened 2022-12-21
Optimise requests to mirror node to speed up relay

### LINKED ISSUE #784: Optimise requests to mirror node to speed up relay
Some requests to the mirror node take longer, because of the address type and retries on error. Review, benchmark and if possible optimise requests to the mirror node or make some of the parallel. Example:
We lose a lot of time here, when resolving address type: https://github.com/hashgraph/hedera-json-rpc-relay/blob/0aa3c5a3ca1d8239ae941cbe916979c6e3d4bac6/packages/relay/src/lib/clients/mirrorNodeClient.ts#L490-L514

natanasow:

**Description**:

**Related issue(s)**:

Fixes #784 

**Notes for reviewer**:

**Checklist**


georgi-l95 (approved):
Super nice change. This will speed up the relay.

dimitrovmaksim:
Wouldn't this return the first resolved request, regardless of whether it found something or not (400 and 404 are acceptable responses for all 3 requests)?

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -488,29 +488,40 @@ export class MirrorNodeClient {
...
-            };
+        const promises = [
+            this.getContract(entityIdentifier, requestId),
+            this.getAccount(entityIdentifier, requestId),
+            this.getTokenById(`0.0.${parseInt(entityIdentifier, 16)}`, requestId)
+        ];
+        // maps the promises with indices of the promises array
+        // because there is no such method as Promise.raceWithIndex in js
+        // the index is needed afterward for detecting the resolved promise type (contract, account, or token)
+        const data = await Promise.race(promises.map((promise, index) => promise.then(value => ({ value, index }))));
```
dimitrovmaksim, line 499: Wouldn't this return the first resolved request, regardless of whether it found something or not (400 and 404 are acceptable responses for all 3 requests)? Meaning that if an address/contract_id is a contract, but token request returns null first, we will assume that this address does not exists at all

dimitrovmaksim: So it relies on the wait time between retries to wait for a 200 from another endpoint if the first one that completes rejects?

natanasow: There is no difference between 0 and 100 retries, except the execution time when the entity can not be found, which will be equal to **N * ms**.
```
Valid account

-------------------------------------------------------->  t
getContract ----- 404 (wait 500ms) - canceled
getAccount  ----------- 200 
getToken    -------- 404 (wait 500ms) - canceled
```

```
Valid token

-------------------------------------------------------->  t
getContract ----- 404 (wait 500ms) - canceled
getAccount  -------- canceled
getToken    -------- 200 
```

```
Not found

-------------------------------------------------------------------------------------------->  t
getContract ----- 404 (wait 500ms) ------ 404 (wait 500ms) --- 404 (not resolved)
getAccount  ----------- 404 (wait 500ms) ---- 404 (wait 500ms) -------- 404 (not resolved)
getToken    -------- 404 (wait 500ms) -- 404 (wait 500ms) ------- 404 (not resolved)
```

dimitrovmaksim: Ok, so in the first case with the valid account, if retires are set to 0, if the first to complete(reject) is the `getContracts` request, won't it reject the whole promise and not return anything for `getAccount`?

natanasow: I guess not because the whole promise will be resolved only by a successful request, the rejected request will be skipped. It might be a good idea to write some tests to cover these cases.

dimitrovmaksim: Are you sure, in the docs for [Promise.race()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) it states `it fulfills if the first promise to settle is fulfilled, and rejects if the first promise to settle is rejected` (see the last of the usage [examples](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race#using_promise.race)). 

I agree we should have tests for that, because if anything changes in how we execute request to the mirror node it may affect how this promise resolves

natanasow: @dimitrovmaksim just tested it and you are right that without the axios-retry package it's not working properly, and I have to modify it. Good catch :pray: .

Ivo-Yankov (approved):

PR #792, opened 2022-12-23
Configurable function input size

### LINKED ISSUE #765: function call with long input data length fails
### Description

I would like to make a large number of calls(via multicall) in a single call, however there appears to be limitations to the length of data for a call as demonstrated [here](https://github.com/mshakeg/hedera-json-rpc-relay/pull/7).

The first discrepancy is that the hedera local-node appears to work with a much higher input length compared to the hashio testnet relay. However even the local-node does not allow inputs as high as the hardhat local node.

### Steps to reproduce

Checkout to the branch in the aforementioned mentioned [PR](https://github.com/mshakeg/hedera-json-rpc-relay/pull/7)

(I) Demo the issue on the hedera local node
1. cd into the `hardhat-example` project's root then start the local node: `npm run hedera-start`
2. Run the `multicaller.test.ts` script against the hedera local-node: `npm run test:multicaller:local`
3. Observe that the local relay returns valid responses for calls where `iterations <= 1637` and fails for calls where `iterations >= 1638`

(II) Demo the issue on the hashio testnet relay
1. Run the `multicaller.test.ts` script against the hashio testnet relay: `npm run test:multicaller:testnet`
2. Observe that the testnet relay returns valid responses for calls where `iterations <= 42` and fails for calls where `iterations >= 43`

(III) Demo the expected multicall capacity on the hardhat local node
1. start the hardhat local node: `npm run hardhat-start`
1. Run the `multicaller.test.ts` script against the hardhat local node: `npm run test:multicaller`
2. Observe that the hardhat relay returns valid responses for calls where `iterations <= 2_936` and fails for calls where `iterations >= 2_937`

### Additional context

_No response_

### Hedera network

testnet, other

### Version

hashio-relay@0.13.0 & @hashgraph/hedera-local@2.3.2

### Operating system

None

Ivo-Yankov:
**Description**:
Adds an `env` variable to control the maximum allowed input size (in mb). By default it is 1mb

**Related issue(s)**:

Fixes #765 

**Notes for reviewer**:

**Checklist**


packages/server/src/koaJsonRpc/index.ts
```diff
@@ -52,7 +52,8 @@ export default class KoaJsonRpc {
 
   constructor(logger: Logger, register: Registry, opts?) {
     this.koaApp = new Koa();
-    this.limit = '1mb';
+    const limit = process.env.INPUT_SIZE_LIMIT || 1;
+    this.limit = limit + 'mb';
```
lukelee-sl, line 56: Kind of weird that `this.limit` can be overwritten if you send in `opts` and it has a `limit` below in line 62.  We should probably only use one potential mechanism for setting this configuration value.

georgi-l95: Yes, I agree with @lukelee-sl

lukelee-sl (dismissed):
LGTM

Nana-EC (approved):

georgi-l95 (approved):
LGTM

Issue #751, opened 2022-12-06
Return appropriate http response codes


mshakeg:
### Description

From what I can tell the response code for all requests is set to 200 even for calls that hit rate limits, according to [this comment](https://forum.thegraph.com/t/are-contract-calls-always-deterministic/1216/3) the graph node expects a NOT OK(specifically 429) response code to retry a call that hit a rate limit.

I'm not sure if there's a spec that defines all response codes, but last I checked the eth json rpc spec doesn't have this detail.

### Steps to reproduce

1. Observe that a 200 ok response is returned for json rpc calls that "reverted" due to the rate limit.

### Hedera network

other

### Version

relay@v0.12.0

### Operating system

None

dimitrovmaksim:
IMO, in general the graph should not be rate limited, even if the appropriate response code is returned, there's a retry limit anyways, so there's the possibility of blocks/data being skipped.

mshakeg:
@dimitrovmaksim do you know what happens once the retry limit is reached?

Also it's got more to do with an event handler making an eth_call request and the hedera relay returning a call "revert" response instead of a rate limit response.

dimitrovmaksim:
For consuming blocks at least it think it skips the block after 10 or 15 retries

mshakeg:
@dimitrovmaksim hmm, does it ever try to recover and reconstruct subgraphs on the skipped blocks. It doesn't make much sense that it would skip blocks and consume later blocks that may only make sense given txs in previous skipped blocks.

dimitrovmaksim:
Tbf it's hard to track all that is happening with the logs spam. Will have to test it again somehow. I may be wrong for the skipping, I may be confusing with fetching the transaction receipt

PR #793, opened 2022-12-23
Return appropriate http codes

dimitrovmaksim:
**Description**:

Currently we return 200 status code in most cases. This PR updates the relay to return the appropriate codes:

InternalError => 500
InvalidParams (JsonRpc request params) => 400
MethodNotFound or UnsupportedMethod => 400
RateLimit => 429
JsonRpcError =>  500 for InternalError, 400 for the rest

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/751

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/src/koaJsonRpc/index.ts
+++ b/packages/server/src/koaJsonRpc/index.ts
@@ -36,6 +36,7 @@ import {
 } from './lib/RpcError';
 import Koa from 'koa';
 import { Registry } from 'prom-client';
+import { JsonRpcError } from '@hashgraph/json-rpc-relay';
 
 const hasOwnProperty = (obj, prop) => Object.prototype.hasOwnProperty.call(obj, prop);
 dotenv.config({ path: path.resolve(__dirname, '../../../../../.env') });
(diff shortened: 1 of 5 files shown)


packages/server/src/koaJsonRpc/index.ts
```diff
@@ -99,18 +100,21 @@ export default class KoaJsonRpc {
         ctx.request.method !== 'POST'
       ) {
         ctx.body = jsonResp(body.id || null, new InvalidRequest(), undefined);
+        ctx.status = 400;
```
lukelee-sl, line 103: does it make sense to add the status as an optional(?) forth parameter to the jsonResp method?

dimitrovmaksim: I'm not sure I understand

lukelee-sl: Ah, I see my suggestion does not make sense.  My mistake.  

But I do have another question.  What happens when calling jsonResp and it throws an exception?

dimitrovmaksim: Hm, good question. It seems like it's not being handled by us anywhere. cc: @georgi-l95

dimitrovmaksim:
```
curl localhost:7546 -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"nonExistingMethod","params": [],"id":1}'
Internal Server Error%                                                                  
```

lukelee-sl: sounds reasonable

lukelee-sl (approved):
LG

georgi-l95 (approved):
LGTM

Issue #533, opened 2022-09-19
Size limit on call result return value to view pure function


mshakeg:
### Description

There appears to be an issue(or internal constraint) on the length of the call result to `view`/`pure` functions. The limit appears to be 32*126=4032 bytes(~4KB). This is too low as some applications require reading a call result >1 MB.

### Steps to reproduce

1. Create `ReadData.sol` with the following contents:
```solidity
// SPDX-License-Identifier: GPL-3.0-or-later
pragma solidity >=0.8.0;

contract ReadData {

    struct LotsOfData {
        uint number1;
        uint number2;
    }

    function getLotsOfData(
        uint24 count
    ) external pure returns (LotsOfData[] memory) {
        LotsOfData[] memory lotsOfData = new LotsOfData[](count);
        return lotsOfData;
    }

    struct NotLotsOfData {
        uint number;
    }

    function getNotLotsOfData(
        uint24 count
    ) external pure returns (NotLotsOfData[] memory) {
        NotLotsOfData[] memory notLotsOfData = new NotLotsOfData[](count);
        return notLotsOfData;
    }
}
```
2. Call `getLotsOfData` and `getNotLotsOfData` with `count` <=63 and <=126 respectively and observe the successful null call result
3. Repeat the same call but now with `count` >63 and >126 respectively and observe the call fails

### Hedera network

testnet

### Version

testnet@latest

mshakeg:
The limit on the hedera local node seems to be 59392 bytes(~59KB) which is still too small.

Ivo-Yankov:
There are 2 limiting factors for the result size - `gasLimit` and the transaction execution time. When increasing the `gasLimit` I managed to receive a result of >700 kb: 
```
    const tx = await contract.getLotsOfData(11175, {
      gasLimit: 15000000
    });
```
However it is possible that due to differences in hardware or OS the local node performs differently for you and you could be reaching the hardcoded maximum of 10s for transaction execution in the SDK. There is a [logged issue](https://github.com/hashgraph/hedera-sdk-js/issues/1276) to make that value configurable.

Ivo-Yankov:
The issue is resolved in the latest release of the SDK - v2.18.6

mshakeg:
Hi @Ivo-Yankov , I don't see a pull request associated with this issue for the SDK version update to the relay.

[...]

PR #812, opened 2023-01-11
ContractCallQuery retry logic

Ivo-Yankov:
**Description**:

In some rare cases `eth_call` is rejected by a precheck with `INSUFFICIENT_TX_FEE`. This PR adds a retry mechanism, that increases the cost with every iteration by a multiplier of `1.1`.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


mshakeg:
Hi @Ivo-Yankov , just wanted to ask what's the maximum return data size you've got using this approach? And how many retry attempts were required?

Ivo-Yankov:
> Hi @Ivo-Yankov , just wanted to ask what's the maximum return data size you've got using this approach? And how many retry attempts were required?

Hi @mshakeg, I tested it with your contract from #533 with this version of the relay connected to testnet and managed to get a result from `getNotLotsOfData(19150, {gasLimit: 15000000})` with a single retry attempt. And with >19200 I started receiving `INSUFFICIENT_GAS`, so the only limiting factor is the gasLimit.

mshakeg:
@Ivo-Yankov fantastic! 19150*32 = 612800 bytes or 612.8 KB, which will be more than enough for most use cases.

Does it work on the first attempt or does it have to retry? Retrying will obviously entail greater latency which is not ideal.

Ivo-Yankov:
@mshakeg It takes one retry attempt, I haven't tested the increased latency, but the `INSUFFICIENT_TX_FEE` error is thrown as a `precheck` by the consensus node, so it doesn't wait for the full execution time.

mshakeg:
@Ivo-Yankov I see, would it not be better to change the base multiplier from 1.1 to 1.21(i.e. 1.1*1.1) to avoid that retry? even if a single retry only marginally increases latency.

Ivo-Yankov:
The multiplier is only applied if it has failed at least once. I.e. if `currentRetry === 0` then the multiplier will be 1: 
```
const multiplier = Math.pow(baseMultiplier, currentRetry); // 1.1 ^ 0 = 1
```

This is done because most Contract Calls do not require a cost increase. If we apply the 1.1 multiplier every time that would effectively increase the expense of operating the relay by roughly 10%.

mshakeg:
I see, that's fair, I thought it wouldn't be an issue as relay operators receive a refund.

natanasow (approved):

georgi-l95 (approved):
LGTM

Issue #626, opened 2022-10-17
Unable to query contract state in subgraph


mshakeg:
### Description

In the _subgraph-example_ project I'm trying to call the `greet` _view_ function on the `Greeter` contract in the [handleGreetingSet](https://github.com/hashgraph/hedera-json-rpc-relay/blob/91a813d7d11b7cb738681b3cb717ffddc2f358d5/tools/subgraph-example/src/greeter.ts#L24) event handler using the [ethereum-api](https://thegraph.com/docs/en/developing/assemblyscript-api/#access-to-smart-contract-state) as follows(1.):
```ts
export function handleGreetingSet(event: GreetingSet): void {
  log.error('in handleGreetingSet', []); // does not reach this point
  const greeterContract = Greeter.bind(event.address);
  log.error('bound greeterContract', []); // does not reach this point
  const greetingValue = greeterContract.greet(); // fails to call greet()
  log.error('greetingValue ' + greetingValue, []); // does not reach this point
  
  ....
}  
```

### Steps to reproduce

1. In `greeter.ts` add `log` to the list of imports from `@graphprotocol/graph-ts`
2. Add the above code snippet(1.) inside the `handleGreetingSet` event handler
3. Run the test in the _hardhat-example_ to deploy the `Greeter` contract and execute the `setGreeting` function
4. Update the subgraph-example's `subgraph.yaml` accordingly setting the correct `startBlock` and Greeter contract address
5. Start the local graph node(i.e. `npm run create-local`) and create and deploy subgraph to local graph node(use scripts `create-local` and `deploy-local`).
6. In the graph node's logs observe that the **"in handleGreetingSet"** and **"bound greeterContract"** logs are visible but not the **"greetingValue"** log.

### Hedera network

other

### Version

@hashgraph/hedera-local@2.0.1

### Operating system

Linux

dimitrovmaksim:
Are you using the relay from the local-node?

mshakeg:
@dimitrovmaksim I got a log relating to the failed call though it appeared much after the call was made, (996-439)*2s/block = 1114s ~ 18min.
NB: The logs inside the handler are omitted as they were no longer in range, due to the significantly elapsed time.
```
[...]

PR #817, opened 2023-01-17
Expand `eth_call` validation to support blockParam object

### LINKED ISSUE #816: Indexing eth_call requests from graph-node fail in release 0.15.0
### Description

Upon upgrading from release `0.14.0` to `0.15.0` eth_call requests are failing when issued from a graph-node indexer. The error is:
```
Invalid parameter 1: Expected 0x prefixed hexadecimal block number, or the string "latest", "earliest" or "pending"
```

### Steps to reproduce

1) Setup the `0.15.0` relayer against hedera testnet
2) Attempt the following eth_call POST request and note the failure
```json
{
    "jsonrpc": "2.0",
    "method": "eth_call",
    "params": [
        {
            "from": "0x5cbaf7669e448e7e4cba7af78da66d16ecb5e2fa",
            "to": "0x0000000000000000000000000000000002ef16ec",
            "gas": "0x186A0",
            "gasPrice": "0x0",
            "value": "0x0",
            "data": "0x95d89b41"
        },
        "26096263"
    ],
    "id": 1
}
```
3) Modify the "26096263" param in the above request to its hex equivalent "0x18E3287" and reissue the call

### Additional context

```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32602,
        "name": "Invalid parameter",
        "message": "[Request ID: 7740f2fc-9257-4f8b-af20-7c4cbfc55007] Invalid parameter 1: Expected 0x prefixed hexadecimal block number, or the string \"latest\", \"earliest\" or \"pending\""
    }
}
```

### Hedera network

testnet

### Version

0.15.0

### Operating system

Windows

georgi-l95:

**Description**:
Expand `eth_call` validation to support `blockParam` object, introduced with [EIP-1898](https://eips.ethereum.org/EIPS/eip-1898). The Graph is passing blockHash as parameter and currently validations only accepts `blockNumber` or `tag`. More info can be found in this issue from this [comment](https://github.com/hashgraph/hedera-json-rpc-relay/issues/626#issuecomment-1383554578) onwards.

Fix some minor intermittent bugs in ERC20 tests.

Bump network images to `0.33.2` and mirror-node images to `0.72.0`.

**Related issue(s)**:

Fixes #816 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/src/validator/types.ts
+++ b/packages/server/src/validator/types.ts
@@ -30,6 +30,18 @@ export const TYPES = {
     test: (param: boolean) => param === true || param === false,
     error: 'Expected boolean type'
   },
+  'blockParams': {
+    test: (param: any) => {
+      if(Object.prototype.toString.call(param) === "[object Object]") {
+        if (param.hasOwnProperty('blockHash')) {
+          return new Validator.BlockHashObject(param).validate();
+        }
+        return new Validator.BlockNumberObject(param).validate();
+      }
+      return /^0[xX]([1-9A-Fa-f]+[0-9A-Fa-f]{0,13}|0)$/.test(param) && Number.MAX_SAFE_INTEGER >= Number(param) || ["earliest", "latest", "pending"].includes(param)
+    },
+    error: Constants.BLOCK_PARAMS_ERROR
+  },
   "filter": {
     test: (param: any) => {
       if(Object.prototype.toString.call(param) === "[object Object]") {
--- a/packages/server/src/validator/types.ts
+++ b/packages/server/src/validator/types.ts
@@ -23,7 +23,7 @@ export const TYPES = {
     error: Constants.BLOCK_HASH_ERROR
   },
   'blockNumber': {
-    test: (param: string) => /^0[xX]([1-9A-Fa-f]+[0-9A-Fa-f]{0,13}|0)$/.test(param) && Number.MAX_SAFE_INTEGER >= Number(param) || ["earliest", "latest", "pending"].includes(param),
+    test: (param: string) => /^0[xX]([1-9A-Fa-f][0-9A-Fa-f]{0,13}|0)$/.test(param) && Number.MAX_SAFE_INTEGER >= Number(param) || ["earliest", "latest", "pending"].includes(param),
     error: Constants.BLOCK_NUMBER_ERROR
   },
   'boolean': {
--- a/packages/server/tests/acceptance/rpc.spec.ts
+++ b/packages/server/tests/acceptance/rpc.spec.ts
@@ -1300,7 +1300,7 @@ describe('@api RPC Server Acceptance Tests', function () {
                 expect(res).to.eq(BASIC_CONTRACT_PING_RESULT);
             });
 
-            it('should fail to execute "eth_call" with correct block number', async function () {
+            it('should execute "eth_call" with correct block number', async function () {
                 const callData = {
                     from: '0x' + accounts[2].address,
                     to: evmAddress,
@@ -1311,7 +1311,7 @@ describe('@api RPC Server Acceptance Tests', function () {
                 expect(res).to.eq(BASIC_CONTRACT_PING_RESULT);
             });
 
-            it('should fail to execute "eth_call" with correct block hash', async function () {
+            it('should execute "eth_call" with correct block hash object', async function () {
                 const blockHash = '0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3';
                 const callData = {
                     from: '0x' + accounts[2].address,
@@ -1323,7 +1323,7 @@ describe('@api RPC Server Acceptance Tests', function () {
                 expect(res).to.eq(BASIC_CONTRACT_PING_RESULT);
             });
 
-            it('should fail to execute "eth_call" with correct block hash', async function () {
+            it('should execute "eth_call" with correct block number object', async function () {
                 const callData = {
                     from: '0x' + accounts[2].address,
                     to: evmAddress,
--- a/packages/server/tests/acceptance/rpc.spec.ts
+++ b/packages/server/tests/acceptance/rpc.spec.ts
@@ -1299,6 +1299,101 @@ describe('@api RPC Server Acceptance Tests', function () {
                 const res = await relay.call('eth_call', [callData, 'latest'], requestId);
                 expect(res).to.eq(BASIC_CONTRACT_PING_RESULT);
             });
+
+            it('should fail to execute "eth_call" with correct block number', async function () {
+                const callData = {
+                    from: '0x' + accounts[2].address,
+                    to: evmAddress,
+                    data: BASIC_CONTRACT_PING_CALL_DATA
+                };
+                
+                const res = await relay.call('eth_call', [callData, '0x1'], requestId);
+                expect(res).to.eq(BASIC_CONTRACT_PING_RESULT);
+            });
+
+            it('should fail to execute "eth_call" with correct block hash', async function () {
+                const blockHash = '0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3';
+                const callData = {
+                    from: '0x' + accounts[2].address,
+                    to: evmAddress,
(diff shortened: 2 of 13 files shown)


packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -1299,6 +1299,101 @@ describe('@api RPC Server Acceptance Tests', function () {
...
+                    from: '0x' + accounts[2].address,
+                    to: evmAddress,
+                    data: BASIC_CONTRACT_PING_CALL_DATA
+                };
+                
+                const res = await relay.call('eth_call', [callData, {'blockHash' : blockHash}], requestId);
+                expect(res).to.eq(BASIC_CONTRACT_PING_RESULT);
+            });
+
+            it('should fail to execute "eth_call" with correct block hash', async function () {
```
lukelee-sl, line 1326: is this test calling with blockNumber rather than blockHash?

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -1299,6 +1299,101 @@ describe('@api RPC Server Acceptance Tests', function () {
                 const res = await relay.call('eth_call', [callData, 'latest'], requestId);
                 expect(res).to.eq(BASIC_CONTRACT_PING_RESULT);
             });
+
+            it('should fail to execute "eth_call" with correct block number', async function () {
```
lukelee-sl, line 1303: It looks like this test is supposed to succeed rather than fail?

packages/server/src/validator/types.ts
```diff
@@ -30,6 +30,18 @@ export const TYPES = {
...
   },
+  'blockParams': {
+    test: (param: any) => {
+      if(Object.prototype.toString.call(param) === "[object Object]") {
+        if (param.hasOwnProperty('blockHash')) {
+          return new Validator.BlockHashObject(param).validate();
+        }
+        return new Validator.BlockNumberObject(param).validate();
+      }
+      return /^0[xX]([1-9A-Fa-f]+[0-9A-Fa-f]{0,13}|0)$/.test(param) && Number.MAX_SAFE_INTEGER >= Number(param) || ["earliest", "latest", "pending"].includes(param)
```
lukelee-sl, line 41: The regex seems odd to me.  It allows for 12 bytes and the first nibble cannot be 0?  I tested 0x1204567890abcdef123456 for example the regex does not match for me.

packages/server/tests/integration/server.spec.ts
```diff
@@ -1111,7 +1111,38 @@ describe('RPC Server', async function() {
 
           Assertions.expectedError();
         } catch (error) {
-          BaseTest.invalidParamError(error.response, Validator.ERROR_CODE, `Invalid parameter 1: ${Validator.BLOCK_NUMBER_ERROR}`);
+          BaseTest.invalidParamError(error.response, Validator.ERROR_CODE, `Invalid parameter 1: ${Validator.BLOCK_PARAMS_ERROR}`);
+        }
+      });
+
+      it('validates Block param is valid block hash', async function() {
```
Nana-EC, line 1118: These should be labelled "non valid" right?
The current logic expects failures

packages/server/src/validator/types.ts
Nana-EC: Maybe also add some of the test cases noted in the EIP for additional coverage

Nana-EC (changes requested):

packages/server/src/validator/types.ts
georgi-l95: First nibble cannot be 0, because block numbers are not starting with 0, that would mean to be padded, which we have to trim afterwards. For example infura doesn't accepts padded blocknumbers, but alchemy accepts them. We can change the regex to accepts, but we may have to trim it afterwards. I mean it could work both ways. 
It allows for this many bytes only, because otherwise it will exceeds javascript max integer size, this is the reason your example is not working. If you cut it down to `0x1204567890abcde`, it should work with this regular expression.

packages/server/tests/integration/server.spec.ts
georgi-l95: Yep, you are correct. Fixed.

packages/server/tests/acceptance/rpc.spec.ts
georgi-l95: Fixed.

packages/server/tests/acceptance/rpc.spec.ts
georgi-l95: Fixed.

packages/server/tests/acceptance/rpc.spec.ts
```diff
@@ -1299,6 +1299,101 @@ describe('@api RPC Server Acceptance Tests', function () {
...
+                    from: '0x' + accounts[2].address,
+                    to: evmAddress,
+                    data: BASIC_CONTRACT_PING_CALL_DATA
+                };
+                
+                const res = await relay.call('eth_call', [callData, '0x1'], requestId);
+                expect(res).to.eq(BASIC_CONTRACT_PING_RESULT);
+            });
+
+            it('should fail to execute "eth_call" with correct block hash', async function () {
```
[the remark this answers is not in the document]
georgi-l95: Fixed.

georgi-l95:
@Nana-EC Tests case examples in the EIP are about `eth_getStorageAt`, I've added similar about `eth_call`, but we are currently not using [blockParam](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L973) parameter, so we can't add all, for now. 
According to the EIP we should make the following endpoints accept block parameter ( tag, block number or object with block hash or block number ):

- eth_getBalance
- eth_getStorageAt
- eth_getTransactionCount
- eth_getCode
- eth_call
- eth_getProof

But only `eth_call` is accepting those in [Alchemy](https://docs.alchemy.com/reference/eth-call), for example. The other endpoints only accept tag or block number (the same as us, currently), both in Alchemy and Infura. Should we follow the EIP fully or only for `eth_call` ? If yes, maybe in a another PR, as this one will get a lot larger ?

Nana-EC:
> @Nana-EC Tests case examples in the EIP are about `eth_getStorageAt`, I've added similar about `eth_call`, but we are currently not using [blockParam](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L973) parameter, so we can't add all, for now. According to the EIP we should make the following endpoints accept block parameter ( tag, block number or object with block hash or block number ):
> 
> * eth_getBalance
> * eth_getStorageAt
> * eth_getTransactionCount
> * eth_getCode
> * eth_call
> * eth_getProof
> 
> But only `eth_call` is accepting those in [Alchemy](https://docs.alchemy.com/reference/eth-call), for example. The other endpoints only accept tag or block number (the same as us, currently), both in Alchemy and Infura. Should we follow the EIP fully or only for `eth_call` ? If yes, maybe in a another PR, as this one will get a lot larger ?

@georgi-l95 Let's match Alchemy and Infura for now.
Open a ticket for the backlog to capture this for other endpoints are you've noted.

Nana-EC (dismissed):

packages/server/src/validator/types.ts
dimitrovmaksim: I may have "over-insured" in this case. Generally it should be fine to have a longer than 16 character hex for block number, my reason was that at one point where hex values will always convert to JS's MAX_SAFE_INTEGER, but then again it would take quite the long amout of time until we reach that point, so in the end that size limit may be uneeded.

georgi-l95: Yes, you're right, I'll change the regex to `/^0[xX]([1-9A-Fa-f][0-9A-Fa-f]{0,13}|0)$/`, which will work for our needs.

lukelee-sl: Thanks for double checking.

lukelee-sl (approved):

lukelee-sl:
LGTM

packages/server/src/validator/types.ts
georgi-l95: Thank you for the feedback, Luke.

PR #820, opened 2023-01-18
Fix cache on getBalance

### LINKED ISSUE #815: Fix cache on getBalance
### Description

Currently if we try to getBalance on account which does not exist, it will be cached as 0. With the changes from HIP-583, this will be a problem, because if we try to getBalance on non existing account, transfer HBAR and then check again, the balance will always be 0( not always but for some time). This might be a problem for metamask for example, because when we create new account with it, metamask will automatically try to get it's balance and will lock it at 0.

### Steps to reproduce

1. Get balance on nonexisting account
2. Create the account
3. Get balance again

### Additional context

_No response_

### Hedera network

other

### Version

latest

### Operating system

None

natanasow:

**Description**:
Remove cache on getBalance, when account doesn't exist, because after HIP 583 we can create accounts via metamask, which asks for balance, before creation and can get stuck on 0.

**Related issue(s)**:

Fixes #815

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -603,12 +603,6 @@ export class EthImpl implements Eth {
       }
     }
 
-    // Cache is only set for `not found` balances
-    const cachedLabel = `getBalance.${account}.${blockNumberOrTag}`;
-    const cachedResponse: string | undefined = this.cache.get(cachedLabel);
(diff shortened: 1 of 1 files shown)


georgi-l95 (approved):
LG

Nana-EC (approved):

PR #832, opened 2023-01-24
eth_feeHistory wrongly returns error

### LINKED ISSUE #813: eth_feeHistory wrongly returns error Incorrect block
### Description

Occasionally when making `eth_feeHistory` calls the relay returns the error **Incorrect block** such as in the following error response:
```
{
   "reason":"bad response",
   "code":"SERVER_ERROR",
   "status":400,
   "headers":{
      "date":"Wed, 11 Jan 2023 10:29:29 GMT",
      "content-type":"application/json; charset=utf-8",
      "content-length":"200",
      "connection":"close",
      "access-control-allow-origin":"*",
      "access-control-allow-methods":"GET,HEAD,PUT,POST,DELETE",
      "via":"1.1 google",
      "cf-cache-status":"DYNAMIC",
      "server":"cloudflare",
      "cf-ray":"787cf878ecf68023-IAD"
   },
   "body":{
      "jsonrpc":"2.0",
      "id":3799,
      "error":{
         "code":-32000,
         "name":"Incorrect block",
         "message":"[Request ID: 2c192c09-7422-4204-a87e-665b2e09ba79] Request beyond head block: requested 29328315, head 29328309"
      }
   },
   "requestBody":{
      "method":"eth_feeHistory",
      "params":[
         "0x4",
         "latest",
         [
            50
         ]
      ],
      "id":3799,
      "jsonrpc":"2.0"
   },
   "requestMethod":"POST",
   "url":"https://testnet.hashio.io/api"
}
```
The block isn't passed explicitly in the `eth_feeHistory` request rather `latest` is as indicated above. Additionally, the error response doesn't add up as at the time of the request i.e. 11 Jan 2023 10:29:29 GMT [Block 29328315](https://hashscan.io/testnet/block/29328315) had already been created and the head block was definitely NOT the earlier block 29328309 as the error message states.

### Steps to reproduce

As stated the error happens occasionally, but I think the above explanation is sufficient evidence of the issue.

### Additional context

_No response_

### Hedera network

testnet

### Version

hashio-relay/0.15.0

### Operating system

None

natanasow:

**Description**:

**Related issue(s)**:

Fixes #813

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -172,7 +172,9 @@ export class EthImpl implements Eth {
 
     try {
       const latestBlockNumber = await this.translateBlockTag(EthImpl.blockLatest, requestId);
-      const newestBlockNumber = await this.translateBlockTag(newestBlock, requestId);
+      const newestBlockNumber = (newestBlock == EthImpl.blockLatest || newestBlock == EthImpl.blockPending)
+        ? latestBlockNumber
+        : await this.translateBlockTag(newestBlock, requestId);
 
       if (newestBlockNumber > latestBlockNumber) {
         return predefined.REQUEST_BEYOND_HEAD_BLOCK(newestBlockNumber, latestBlockNumber);
--- a/packages/relay/tests/lib/eth.spec.ts
+++ b/packages/relay/tests/lib/eth.spec.ts
@@ -2052,6 +2052,88 @@ describe('Eth calls using MirrorNode', async function () {
     expect(rewards[1]).to.equal('0x0');
   });
 
+  it('eth_feeHistory with latest param', async function () {
+    const previousBlock = {...defaultBlock, number: blockNumber2, timestamp: {
+      from: '1651560386.060890948',
+      to: '1651560389.060890948'
+    }};
+    const latestBlock = {...defaultBlock, number: blockNumber3};
+    const previousFees = JSON.parse(JSON.stringify(defaultNetworkFees));
+    const latestFees = JSON.parse(JSON.stringify(defaultNetworkFees));
+
+    mock.onGet('blocks?limit=1&order=desc').reply(200, {blocks: [latestBlock]});
+    mock.onGet(`blocks/${previousBlock.number}`).reply(200, previousBlock);
+    mock.onGet(`blocks/${latestBlock.number}`).reply(200, latestBlock);
+    mock.onGet(`network/fees?timestamp=lte:${previousBlock.timestamp.to}`).reply(200, previousFees);
+    mock.onGet(`network/fees?timestamp=lte:${latestBlock.timestamp.to}`).reply(200, latestFees);
+
+    const feeHistory = await ethImpl.feeHistory(1, 'latest', [25, 75]);
+    expect(feeHistory).to.exist;
+    expect(feeHistory['oldestBlock']).to.eq('0x' + blockNumber3);
+  });
+
+  it('eth_feeHistory with pending param', async function () {
+    const previousBlock = {...defaultBlock, number: blockNumber2, timestamp: {
+      from: '1651560386.060890948',
+      to: '1651560389.060890948'
+    }};
+    const latestBlock = {...defaultBlock, number: blockNumber3};
+    const previousFees = JSON.parse(JSON.stringify(defaultNetworkFees));
+    const latestFees = JSON.parse(JSON.stringify(defaultNetworkFees));
+
+    mock.onGet('blocks?limit=1&order=desc').reply(200, {blocks: [latestBlock]});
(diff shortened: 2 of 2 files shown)


georgi-l95 (approved):
LG

packages/relay/src/lib/eth.ts
```diff
@@ -172,7 +172,9 @@ export class EthImpl implements Eth {
 
     try {
       const latestBlockNumber = await this.translateBlockTag(EthImpl.blockLatest, requestId);
-      const newestBlockNumber = await this.translateBlockTag(newestBlock, requestId);
+      const newestBlockNumber = (newestBlock == EthImpl.blockLatest || newestBlock == EthImpl.blockPending)
```
Nana-EC, line 175: Please add a test that ensure the previous case does not occur.
That is a test based on the logic that
- if `newestBlock == EthImpl.blockLatest` then `newestBlockNumber` = `latestBlockNumber`
- if `newestBlock == EthImpl.blockPending` then `newestBlockNumber` = `this.translateBlockTag(newestBlock, requestId)`

Nana-EC (changes requested):
missing tests

packages/relay/src/lib/eth.ts
AlfredoG87: > if newestBlock == EthImpl.blockPending then newestBlockNumber =

Should not be also = `latestBlockNumber` or do you mean that this case should not happen?

AlfredoG87:
Noticed that Acceptance Tests check failed due to 1 error on API Job with error:

 

> 1) RPC Server Acceptance Tests
       Acceptance tests
         @api RPC Server Acceptance Tests
           RPC Server Acceptance Tests
             eth_getBalance
               @release should correctly execute "eth_getBalance" with block number in the last 15 minutes with several txs around that time:

      AssertionError: expected '2999479740000000000' to equal '2999479750000000000'
      + expected - actual

      -2999479740000000000
      +2999479750000000000
      
      at Context.<anonymous> (packages/server/tests/acceptance/rpc.spec.ts:1041:73)


is this difference expected?

packages/relay/src/lib/eth.ts
ebadiere: Good question, since a call to translateBlockTag with a block in a pending state should also return the latest block.

ebadiere: Good question since the call to translateBlockTag with a newestBlock in a pending state will return the "most recent block number" by calling  the getLatestBlock on the mirrorNodeClient.  The tests seem return the same result when passed "latest" or "pending".

georgi-l95:
> 



> Noticed that Acceptance Tests check failed due to 1 error on API Job with error:
> 
> > 1. RPC Server Acceptance Tests
> >    Acceptance tests
> >    @api RPC Server Acceptance Tests
> >    RPC Server Acceptance Tests
> >    eth_getBalance
> >    @Release should correctly execute "eth_getBalance" with block number in the last 15 minutes with several txs around that time:
> 
> ```
>   AssertionError: expected '2999479740000000000' to equal '2999479750000000000'
>   + expected - actual
> 
>   -2999479740000000000
>   +2999479750000000000
>   
>   at Context.<anonymous> (packages/server/tests/acceptance/rpc.spec.ts:1041:73)
> ```
> 
> is this difference expected?

No, but because we are doing several transactions and fetching with multiple requests, even if one fails, the whole scenario fails. So it's bound to happen sometime. It won't be a bad idea to try and fix this, but it won't be that easy.

AlfredoG87:
> > 
> 
> > Noticed that Acceptance Tests check failed due to 1 error on API Job with error:
> > > 1. RPC Server Acceptance Tests
> > >    Acceptance tests
> > >    @api RPC Server Acceptance Tests
> > >    RPC Server Acceptance Tests
> > >    eth_getBalance
> > >    @Release should correctly execute "eth_getBalance" with block number in the last 15 minutes with several txs around that time:
> > 
> > 
> > ```
> >   AssertionError: expected '2999479740000000000' to equal '2999479750000000000'
> >   + expected - actual
> > 
> >   -2999479740000000000
> >   +2999479750000000000
> >   
> >   at Context.<anonymous> (packages/server/tests/acceptance/rpc.spec.ts:1041:73)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is this difference expected?
> 
> No, but because we are doing several transactions and fetching with multiple requests, even if one fails, the whole scenario fails. So it's bound to happen sometime. It won't be a bad idea to try and fix this, but it won't be that easy.

I just ran the test again and this time it completed without issues, thanks!!

Nana-EC (approved):

PR #844, opened 2023-01-26
Fix eth_estimateGas for hollow account creation

natanasow:

**Description**:

The basic transfer needs 21k gas but hollow account creation needs 587k gas

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


Nana-EC (approved):
LG.
Do the current images support testing this E2E in an acceptance test?
If so let's add one for hollow account creation when the appropriate gas for account creation is provided in a transfer transaction

georgi-l95:
Yes, the current image support is allowing us to write some E2E tests. I think it's a good idea.

lukelee-sl (approved):

lukelee-sl:
My guess is that there is a small chance of a race condition where we will give the wrong gas estimate for a toAccount that is being create.

packages/relay/src/lib/eth.ts
```diff
@@ -326,7 +327,8 @@ export class EthImpl implements Eth {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     this.logger.trace(`${requestIdPrefix} estimateGas()`);
     if (!transaction || !transaction.data || transaction.data === '0x') {
```
Nana-EC, line 329: Can we add a comment up here to highlight this check implies this is a transfer transaction and not a contract function execution

packages/relay/src/lib/eth.ts
```diff
@@ -326,7 +327,8 @@ export class EthImpl implements Eth {
...
-      return EthImpl.gasTxBaseCost;
+      const toAccount = await this.mirrorNodeClient.getAccount(transaction.to);
+      return toAccount ? EthImpl.gasTxBaseCost : EthImpl.gasTxHollowAccountCreation;
```
Nana-EC, line 331: nit: comment.
```suggestion
      // when account exists return default base gas, otherwise return the minimum amount of gas to create an account entity
      return toAccount ? EthImpl.gasTxBaseCost : EthImpl.gasTxHollowAccountCreation;
```

Nana-EC (commented):
LG. You can take this out of draft.
Minor nits

Nana-EC:
> My guess is that there is a small chance of a race condition where we will give the wrong gas estimate for a toAccount that is being create.

Should be fine as network will charge the lower gas amount if the it determines the account is existent by the time one of the relay submitted accounts hits the network.

Nana-EC (approved):

packages/relay/tests/lib/openrpc.spec.ts
```diff
@@ -196,6 +196,7 @@ describe("Open RPC Specification", function () {
     });
 
     it('should execute "eth_estimateGas"', async function () {
+        mock.onGet(`accounts/undefined`).reply(404);
```
lukelee-sl, line 199: not sure why this is necessary?

lukelee-sl (approved):
LGTM

packages/relay/tests/lib/openrpc.spec.ts
Nana-EC: I also had to read into this but I believe it's since there's no value for `transaction.to` object, so when it's put into the UT here it results in the `undefined` value literally getting stringified.

Could be better structured with a default address but fortunately we do have coverage in acceptances tests where address is a valid hex address

PR #860, opened 2023-01-30
Fix eth_estimateGas for hollow account creation (#844)

Nana-EC:
**Description**:
Cherry pick #844 
Hollow account creation requires an accurate gas amount response

- add a `TX_HOLLOW_ACCOUNT_CREATION_GAS` to capture the necessary gas amount
- Update `eth_estimateGas` to check in the case of a transfer transaction if the recipient exists or not. If so return update gas, if not default to current flow
- Update unit and acceptance tests

Co-authored-by: georgi-l95 <glazarov95@gmail.com>

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


lukelee-sl (approved):
LGTM

AlfredoG87:
Why is the DCO Check failing with: _one commit is incorrectly signed off_, when the commit in this PR has "Signed-off-by" as part of the message, could it be because it was expecting that at the last line of the commit message, and in this case is "Co-authored-by" instead?

image

lukelee-sl (approved):

PR #862, opened 2023-01-31
Cache results from resolveEntityType

### LINKED ISSUE #853: Cache results from resolveEntityType
### Problem

`resolveEntityType` makes a lot of requests to the Mirror node. Once the function has resolved the type of an entity it can be cached to reduce future requests.

### Solution

Apply caching logic to `resolveEntityType`, but only when it has actually found an entity.

### Alternatives

_No response_

natanasow:

**Description**:

**Related issue(s)**:

Fixes #853 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -508,6 +512,12 @@ export class MirrorNodeClient {
       requestId?: string,
       searchableTypes: any[] = [constants.TYPE_CONTRACT, constants.TYPE_ACCOUNT, constants.TYPE_TOKEN]
     ) {
+        const cachedLabel = `resolveEntityType.${entityIdentifier}`;
+        const cachedResponse: { type: string, entity: any } | undefined = this.cache.get(cachedLabel);
+        if (cachedResponse != undefined) {
```
Nana-EC, line 517: Will this mean we'd cache an account that's not found to the pint that it could be created and we'd still return not found?

Let's add a test that follow the hollow account flow i.e. a check if done, the account doesn't exist, then the account is created, then a check is made and the account exists.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -550,10 +562,12 @@ export class MirrorNodeClient {
             }
         }
 
-        return {
+        const response = {
             type,
             entity: data.value
         };
+        this.cache.set(cachedLabel, response);
```
Nana-EC, line 569: maybe we should only cache on accounts that exist? Let's see

Nana-EC (changes requested):
test request to ensure we don't cache and break new account flow

packages/relay/src/lib/clients/mirrorNodeClient.ts
natanasow: If we look at the entire function, we will see that there are 3 possibilities:

1. The function receives a valid contract address
    - We query for the contract address, if the contract exists, then return and cache it. The tricky part here is that we MUST check for the contract existence first because the mirror node should return a valid response for both `/contracts/<contractAddress>` and `/accounts/<contractAddress>` because in hedera a contract is treated the same way as an account entity

2. The function receives a valid account or token (we run them with `Promise.any` which will fail only when the 2 promises have been rejected)
    - The account or token has been found, we return it to the user and cache it

3. The function receives a non-existing contract/account/token
    - The contract and token endpoints return null - no entity has been found we return null from the function and we do not cache the response
    - The account and token promises were rejected because of `Promise.any` - no entity has been found, do not cache

Nana-EC: Gotcha.
Thanks for the reminders and explanation.
I think the fact that the whole function is hard to follow to this point is something to be address.
Either with future refactoring or with a few more comments in the function

Nana-EC (approved):

PR #875, opened 2023-02-05
Replace hethers.js with ethers.js in hardhat-example

georgi-l95:

**Description**:
Replace hethers.js with ethers.js in our hardhat example in tools folder.
Bump network and mirror node images in tests

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


tools/hardhat-example/hardhat.config.js
```diff
@@ -46,12 +28,31 @@ task('contract-call', async (taskArgs) => {
...
-    }
-  }
+    local: {
+      url: process.env.RELAY_ENDPOINT,
+      accounts: [process.env.OPERATOR_PRIVATE_KEY, process.env.RECEIVER_PRIVATE_KEY],
+      chainId: 298,
+    },
+    testnet: {
+      url: 'https://testnet.hashio.io/api',
+      accounts: [process.env.OPERATOR_PRIVATE_KEY_TESTNET, process.env.RECEIVER_PRIVATE_KEY_TESTNET],
```
Nana-EC, line 54: Leave this array empty.
Let's not add another config just for testnet. Let devs customize this part or reuse the same config name with the intent they'd fill it in appropriately as needed

tools/hardhat-example/hardhat.config.js
```diff
@@ -1,34 +1,16 @@
-/*-
```
Nana-EC, line 1: Restore license header

tools/hardhat-example/README.md
```diff
@@ -3,22 +3,27 @@
...
 ## Configuration
 
 Create `.env` file based on `.env.example`
+
 ```
 # Alias accounts keys
 OPERATOR_PRIVATE_KEY=
 RECEIVER_PRIVATE_KEY=
-
+OPERATOR_PRIVATE_KEY_TESTNET=
```
Nana-EC, line 19: See comment below. I don't think we need to add another config variable

Nana-EC (commented):
Looking good on actual changes.
Some bits on license and the new testnet config

tools/hardhat-example/hardhat.config.js
georgi-l95: You're right. Left it empty array.

tools/hardhat-example/hardhat.config.js
georgi-l95: Missed that one. Restored.

tools/hardhat-example/README.md
georgi-l95: Removed it.

tools/hardhat-example/hardhat.config.js
georgi-l95: Missed that one. Restored

natanasow (approved):
LGTM

tools/hardhat-example/scripts/deployContract.js
```diff
@@ -18,18 +18,14 @@
...
+const { ethers } = require('hardhat');
 
 module.exports = async () => {
-  const provider = new hre.ethers.providers.JsonRpcProvider(process.env.RELAY_ENDPOINT);
-  const wallet = new hre.ethers.Wallet(process.env.OPERATOR_PRIVATE_KEY, provider);
-  const Greeter = await hre.ethers.getContractFactory('Greeter', wallet);
-  const greeter = await Greeter.deploy('initial_msg');
+  let wallet = (await ethers.getSigners())[0];
+  const Greeter = await ethers.getContractFactory('Greeter', wallet);
+  const greeter = await Greeter.deploy('initial_msg', { gasLimit: 1000000 });
```
Nana-EC, line 26: Why the explicit gasLimit here and above?
I sit just to indicate such calls don't cost more or is it required now for some reason?
If this needs to be explicit we should be clear why with a comment or some kind of an issue etc

Nana-EC (approved):
LG
Curious on the gasLimit detail

tools/hardhat-example/scripts/deployContract.js
georgi-l95: Actually, it's not needed. I've tested and probably forgot to remove it. Will do it now. Also ethers will automatically determine the gas, no need for explicitly writing the gasLimit.

natanasow (approved):
LGTM

PR #878, opened 2023-02-07
Add simple eth_call cache

### LINKED ISSUE #877: Add a simple eth_call cache
### Problem

In some cases DApps may poll the relay `eth_call` endpoint quite often even when there are no contract changes.
In the current iteration prior to HIP 584 completion is may be beneficial to consider a cache to speed up responses

### Solution

Add an appropriate cache

- Check the Mirror Node `api/v1/contracts/{contractIdOrAddress}` for the timestamp of the last change
- Cache the timestamp
- On future calls query the endpoint. If the timestamp returned is greater than the cached value then it's worth making a full state call

- Also add a new metric for gas used by an endpoint. Start with a proxy of 0.8 * gasLimit provided

### Alternatives

_No response_

Nana-EC:
**Description**:
Add simple cache to eth_call method

- Add `getContractCurrentStateByAddress()` to `MirrornNodeClient.ts`
- Add `hasContractStateSeenAnUpdate()` to `eth.ts` to check if a contract state has seen a change
- Update `eth_call` to utilize a cache and call `hasContractStateSeenAnUpdate()` to decide if to return cache. Idea is on a successful call the timestamp and response are cached. On successive calls if the timestamp from mirror is greater than the cache then make the complete state call

**Related issue(s)**:

Fixes #877 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -480,6 +480,16 @@ export class MirrorNodeClient {
         return this.getContractResultsByAddress(address, contractResultsParams, limitOrderParams, requestId);
     }
 
+    public async getContractCurrentStateByAddress(address: string, requestId?: string) {
+        return this.getPaginatedResults(
+            `${MirrorNodeClient.GET_CONTRACT_ENDPOINT}${address}${MirrorNodeClient.GET_STATE_ENDPOINT}`,
+            MirrorNodeClient.GET_STATE_ENDPOINT,
+            MirrorNodeClient.CONTRACT_STATE_PROPERTY,
+            [400, 404],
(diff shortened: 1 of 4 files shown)


Ivo-Yankov:
What would happen if a contract's state remains unchanged but it has a method that returns something that depends on the state of another contract?

Nana-EC:
Will pick this up again once 0.36 goes out

Nana-EC:
> 

Ohhhh good catch.
Ugh, let me think on that. The answer might be this solution not longer works :)
It would if we had a list of all contracts that may be affected by a functions execution, but that's a whole thing on its own

Nana-EC:
Closing as the inter contract behaviors mean the cached value may be wrong.

PR #880, opened 2023-02-07
Add simple cache on `eth_call`

### LINKED ISSUE #877: Add a simple eth_call cache

georgi-l95:

**Description**:
Adds a cache on eth_call request, based on `to` and `data` parameters for 200ms.

**Related issue(s)**:

Fixes #877 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -1017,8 +1017,17 @@ export class EthImpl implements Eth {
         }
       }
 
+      let cachedResponse = this.cache.get(`eth_call: ${call.data} from ${call.to}`);
```
Nana-EC, line 1020: nit: add const for label. Maybe shorten key also
```suggestion
      const cacheKey = `eth_call:.${call.to}.${call.data}`;
      let cachedResponse = this.cache.get(cacheKey);
```

packages/relay/src/lib/eth.ts
```diff
@@ -1017,8 +1017,17 @@ export class EthImpl implements Eth {
...
+      if (cachedResponse != undefined) {
+        this.logger.debug(`${requestIdPrefix} eth_call returned cached response: ${cachedResponse}`);
+        return cachedResponse;
+      }
+
       const contractCallResponse = await this.sdkClient.submitContractCallQuery(call.to, call.data, gas, call.from, EthImpl.ethCall, requestId);
-      return EthImpl.prepend0x(Buffer.from(contractCallResponse.asBytes()).toString('hex'));
+      const formattedCallReponse = EthImpl.prepend0x(Buffer.from(contractCallResponse.asBytes()).toString('hex'));
+      
+      this.cache.set(`eth_call: ${call.data} from ${call.to}`, formattedCallReponse, { ttl: 200 });
```
Nana-EC, line 1029: use `cacheKey` here

packages/relay/src/lib/eth.ts
Nana-EC, line 1020: I'm concerned the `call.data` can get quite long.
In my exploration I went down the path of taking a hashing of the data to reduce it's length.

Maybe since the cache is so short lived it's not a biggie but let's make sure the key itself doesn't have length limits.
Maybe a simple approach of taking the first n characters also works if the hash approach is costly

Nana-EC: I just wanna avoid any potential error cases from a long data. maybe add a test case with a long data field if you stay with this approach.

Nana-EC (commented):
LG
Some suggestions

packages/relay/src/lib/eth.ts
Nana-EC, line 1029: Please make the ttl value a config to allow for on demand tuning

Nana-EC (commented):
Let's make the ttl configureable

Ivo-Yankov:
I think this cache could cause issues in the case where the result of a method from a contract depends not on its own state, but on the state of another contract.

Such is the case of a Multicaller contract, where it only aggregates and returns results from multiple other contracts, but it doesn't update its own state.

georgi-l95:
Hmm, I think you're right, but we have a ttl of 200ms, which is not so long, because those multicalls might take some time. Also this is a temporary solution and it's configurable if there is indeed problem with multicall queries.

Nana-EC (approved):
LG

packages/relay/src/lib/eth.ts
```diff
@@ -1017,8 +1019,24 @@ export class EthImpl implements Eth {
         }
       }
 
+      let data = call.data;
+      if (data) {
+        data = crypto.createHash('sha1').update(call.data).digest('hex'); // NOSONAR
+      }
+
+      const cacheKey = `eth_call:.${call.to}.${data}`;
```
lukelee-sl, line 1027: what happens if `data` is null/unassigned?

georgi-l95: No problems there we continue with the query as normal, I just didn’t want to calculate hash for a undefined, now that I think, I should’ve added a or call.data.length is bigger than 64(for example).

Issue #866, opened 2023-02-01
Review application metrics story


Nana-EC:
### Problem

Currently we have a good set of metrics but they can be improved and need to be reviewed for missing details

### Solution

- review current metrics and dashboards
- determine and add additional metrics - consider server E2E, consensus calls, mirror calls
- update dashboards  to better answer activity based questions related to RPS etc

PR #887, opened 2023-02-08
Needed changes for metrics improvements

AlfredoG87:
**Description**:

This PR defines bucket ranges for the histogram of the following metrics:

-rpc_relay_mirror_response
-rpc_relay_method_response
from using the default bucket values of [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10] that are measuring latency in seconds to [5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000] that are measuring latency in milliseconds, witch is the unit of the observations taken.

And also removed line that decreases the gauge (operatorAccountGauge), that its purpose is to have visibility to the costs per account, but that is also being tracker by rpc_relay_consensusnode_response. So better remove to decrease complexity, unneeded redundancy.

**Related issue(s)**:
https://github.com/hashgraph/hedera-json-rpc-relay/issues/866

Fixes #

**Notes for reviewer**:

**Current Metrics with old buckets:**
image

**After this PR:**
image

**Current Gauge Metrics:**
image

**After this PR:**
image

**Checklist**


david-bakin-sl:
Good catch on bucketing by seconds vs milliseconds!

david-bakin-sl (dismissed):
LGTM!

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -167,7 +167,8 @@ export class MirrorNodeClient {
             name: metricHistogramName,
             help: 'Mirror node response method statusCode latency histogram',
             labelNames: ['method', 'statusCode'],
-            registers: [register]
+            registers: [register],
+            buckets: [5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000]
```
Nana-EC, line 171: nit: add comment to let people know what unit you are specifying

packages/relay/tests/lib/hbarLimiter.spec.ts
```diff
@@ -38,12 +40,12 @@ describe('HBAR Rate Limiter', async function () {
...
 
     it('should be disabled, if we pass invalid total', async function () {
-        rateLimiter = new HbarLimit(logger, currentDateNow, invalidTotal, validDuration);
+        rateLimiter = new HbarLimit(logger, currentDateNow, invalidTotal, validDuration, registry);
 
         const isEnabled = rateLimiter.isEnabled();
         const limiterResetTime = rateLimiter.getResetTime();
         const limiterRemainingBudget = rateLimiter.getRemainingBudget();
-        const shouldRateLimit = rateLimiter.shouldLimit(currentDateNow);
+        const shouldRateLimit = rateLimiter.shouldLimit(currentDateNow, 'mode', 'methodName');
```
Nana-EC, line 48: nit: use valid mode and method values in the test cases. We'll never pass 'mode' or 'methodName'.
Makes it less confusing for someone new reading and wondering when we'd pass those values

packages/server/src/server.ts
```diff
@@ -59,7 +59,8 @@ const methodResponseHistogram = new Histogram({
   name: metricHistogramName,
   help: 'JSON RPC method statusCode latency histogram',
   labelNames: ['method', 'statusCode'],
-  registers: [register]
+  registers: [register],
+  buckets: [5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000]
```
Nana-EC, line 63: nit: same comment on noting units here

Nana-EC (changes requested):
LG, mostly recommended nits

Nana-EC (approved):
LG

PR #885, opened 2023-02-07
cherry pick #880 - simple cache for eth_call

lukelee-sl:
Signed-off-by: lukelee-sl [luke.lee@swirldslabs.com](mailto:luke.lee@swirldslabs.com)

**Description**:

Cherry pick #880 into release/0.17 branch in order to create a 0.17.1 version

PR #892, opened 2023-02-10
856-update-logs-and-metrics-to-capture-IP-and-HBAR-rate-limiting.  Ad…

ebadiere:
This PR modifies rate limited logs in order to include the requestId in a standard log message.

* Adds a Koa requestId package that supports storing the requestId in the application context.

Fixes #885 

**Notes for reviewer**:
 Provides the request ID in the logs right as the rateLimit for a given IP is triggered, in the standard log format.

- Existing tests cover rateLimited requests.


packages/server/src/rateLimit/index.ts
```diff
@@ -42,7 +42,7 @@ export default class RateLimit {
     });
   }
 
-  shouldRateLimit(ip: string, methodName: string, total: number): boolean {
+  shouldRateLimit(ip: string, requestId: string, methodName: string, total: number): boolean {
```
Nana-EC, line 45: nit: always best to add new params at the end of a method.

packages/server/package.json
```diff
@@ -7,6 +7,7 @@
     "author": "Hedera Smart Contracts Team",
     "dependencies": {
         "@hashgraph/json-rpc-relay": "file:../relay",
+        "@kasa/koa-request-id": "^2.0.1",
```
Nana-EC, line 10: were there other options?
Doesn't seem heavily utilized or maintained.
Guess the package itself is simple and it seems to only `require('uuid')` so it's safe

Nana-EC (commented):
LG, question and nit comment

packages/server/package.json
david-bakin-sl: "safe" - now, yes, but isn't this the kind of dependency that has caused, for other applications, unnecessary trauma at later dates?  For such a small thing?  Can't we take advantage of the MIT license and just clone it into a repo that we control? Or just rewrite it, or include it?  MIT license!?  (Basically I think Nana is right to question this kind of thing.)

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -110,9 +114,10 @@ export default class KoaJsonRpc {
         return;
       }
 
+      this.requestId = ctx.state.reqId;
       const methodName = body.method;
       const methodTotalLimit = this.registryTotal[methodName];
-      if (this.rateLimit.shouldRateLimit(ctx.ip, methodName, methodTotalLimit)) {
+      if (this.rateLimit.shouldRateLimit(ctx.ip, ctx.state.reqId, methodName, methodTotalLimit)) {
```
david-bakin-sl, line 120: you could also have used here `this.requestId` (see line 117 just above) - don't know which is better readability-wise.

packages/server/src/server.ts
```diff
@@ -136,7 +135,8 @@ app.getKoaApp().use(async (ctx, next) => {
 const logAndHandleResponse = async (methodName: any, methodParams: any, methodFunction: any) => {
   const start = Date.now();
   let ms;
-  const requestId = generateRequestId();
+
+  const requestId = app.getRequestId();
   const requestIdPrefix = requestId ? `[${REQUEST_ID_STRING}${requestId}]` : '';
```
david-bakin-sl, line 140: Over in server/tests/helpers.utils.ts someone made a nice tiny method `formatRequestIdMessage`. I don't think it's a bad idea to encapsulate this in a method, even of one line.

david-bakin-sl (approved):
LGTM though didn't build/test it myself.

AlfredoG87:
DCO check failed, you need to sign messages with pgp key and include message "signed-off-by..." as [specified](https://github.com/hashgraph/hedera-json-rpc-relay/pull/892/checks?check_run_id=11257165206)  on the details of the check task.

packages/server/package.json
georgi-l95: Agree with both Nana and David, doesn't seem necessary to use a whole package for a id generator. IMO, it would be better to implement our own solution. We can use the `koa-request-id` as reference, if needed.

packages/server/src/server.ts
georgi-l95: Agree, let's use this method for consistency.

packages/server/src/rateLimit/index.ts
ebadiere: Agreed.  Updated.

Nana-EC:
@ebadiere based on comments above sounds like we're suggesting go a simple code addition that we're in control of and doesn't need another npm package.

packages/server/src/server.ts
ebadiere: Added a formatters.ts to the server package that follows the pattern for use of the formatters in the relay package.  Updated the existing logging in the servers and rateLimit to use it.

packages/server/src/koaJsonRpc/index.ts
ebadiere: Updated.

packages/server/package.json
ebadiere: Applied requestId functionality to the server as a middleware.

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -110,9 +113,10 @@ export default class KoaJsonRpc {
         return;
       }
 
+      this.requestId = ctx.state.reqId;
       const methodName = body.method;
       const methodTotalLimit = this.registryTotal[methodName];
-      if (this.rateLimit.shouldRateLimit(ctx.ip, methodName, methodTotalLimit)) {
+      if (this.rateLimit.shouldRateLimit(ctx.ip, methodName, methodTotalLimit, this.requestId)) {
         ctx.body = jsonResp(body.id, new IPRateLimitExceeded(methodName), undefined);
         ctx.status = 409;
```
georgi-l95, line 121:
```suggestion
        ctx.status = 429;
```
nit: should we change, as 429 error code is for too many requests

georgi-l95 (approved):
LG, nice work

packages/server/src/koaJsonRpc/index.ts
ebadiere: Agreed, however the relay client used for testing is built using ethers.js.  The ethers.js provider consumes the 429 error code and uses it to implement throttling.  We would need to implement our own ethers.js provider, much like infura or alchemy have done.

georgi-l95: Gotcha. That seems like a work for a new issue, maybe.

packages/server/src/formatters.ts
```diff
@@ -0,0 +1,30 @@
+/*-
+ *
+ * Hedera JSON RPC Relay
+ *
+ * Copyright (C) 2022 Hedera Hashgraph, LLC
```
Nana-EC, line 5: 2023

packages/server/src/koaJsonRpc/index.ts
Nana-EC: Does this mean we're still throwing a 409 instead of the more appropriate 429 that we agreed on?

If so yeah agree with Georgi looks like future work.
Please open a ticket around updating to 429.
We'll probably have to use a separate test flow not ethersjs.
Maybe just a raw HTTP request client that confirms it

packages/server/src/rateLimit/index.ts
```diff
@@ -51,8 +52,13 @@ export default class RateLimit {
         return false;
       }
 
+
+      const requestIdPrefix = requestId ? formatRequestIdMessage(requestId) : '';
+      this.logger.warn(`${requestIdPrefix}, Rate limit call to ${methodName}, ${this.database[ip].methodInfo[methodName].remaining} out of ${total} calls remaining`);
```
Nana-EC, line 57: This double the logs in this flow.
I also think it's a duplicate.
Just use the original.
Honestly I forget why we check for a valid requestId. In deployed flows it will be valid. If it's missing an if in test case it's probably not a big deal

packages/server/src/server.ts
```diff
@@ -134,11 +135,48 @@ app.getKoaApp().use(async (ctx, next) => {
   }
 });
 
+app.getKoaApp().use(async (ctx, next) => {
```
Nana-EC, line 138: I feel like this section can be better formatted to read like 3 different properties being checked and potentially set in 3 different method with 1 method managing the orchestration of the 3.
Not a biggie for this PR but something to note for easier testing and reading

Nana-EC: That being said I don't see any tests for this logic.
Is it possible to test?

Nana-EC (commented):
Looks good.
Can we quickly improve the coverage?
If not then let's open up follow up items to address in the next sprint

ebadiere:
Still needs to resolve the gpg signed commits

PR #896, opened 2023-02-14
Design Document for `eth_subscribe` and `eth_unsubscribe`

### LINKED ISSUE #861: Research `eth_subscribe` to expose
### Problem

Currently the relay does not expose `eth_subscribe` and `eth_unsubscribe`, therefore these not support any subscription functionality. 

### Solution

It works by subscribing to particular events and for each subscription an id is returned. For each event that matches the subscription a notification with relevant data is sent together with the subscription id. 
For now we need to explore only subscribing to event logs. 

Example Request:
```
>{"jsonrpc":"2.0", "id": 1, "method": "eth_subscribe", "params": ["logs", {"address":"0x8320fe7702b96808f7bbc0d4a888ed1468216cfd", "topics": ["0xd78a0cb8bb633d06981248b816e7bd33c2a35a6089241d099fa519e361cab902"]}]}
```

Example Response for successful subscription:

```
{
    "id": 1, 
    "jsonrpc": "2.0", 
    "result": "0x9cef478923ff08bf67fde6c64013158d"
}
```

Example Response for new log:
```
// logs Subscription
{
    "jsonrpc":"2.0",
    "method":"eth_subscription",
    "params": {
        "subscription":"0x4a8a4c0517381924f9838102c5a4dcb7",
        "result": {
            "address":"0x8320fe7702b96808f7bbc0d4a888ed1468216cfd","blockHash":"0x61cdb2a09ab99abf791d474f20c2ea89bf8de2923a2d42bb49944c8c993cbf04",
            "blockNumber":"0x29e87","data":"0x00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000003",
            "logIndex":"0x0",
            "topics":["0xd78a0cb8bb633d06981248b816e7bd33c2a35a6089241d099fa519e361cab902"],"transactionHash":"0xe044554a0a55067caafd07f8020ab9f2af60bdfe337e395ecd84b4877a3d1ab4",
            "transactionIndex":"0x0"
        }
    }
}
```

### Alternatives

_No response_

georgi-l95:

**Description**:
Adds design doc for subscription and introducing new methods - `eth_subscribe` and `eth_unsubscribe`

**Related issue(s)**:

Fixes #861 

**Notes for reviewer**:

**Checklist**


docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,281 @@
...
+      }],
+      lastUpdated: BlockNumber // The latest block number at which this subscription was updated
+    }
+  ]
+
+```
+
+3. Create a scheduled job that loops through all saved subscriptions and calls the Mirror node Rest API and sends the response to all `subscribers`. It should be possible to configure the time interval. A throttling mechanism can be implemetented that limits the consecutive calls to the Mirror node, and at every interval only X subscriptions with the oldest `lastUpdated` should be polled.
+
+## Error Codes
```
Nana-EC, line 252: There should be an error case also for
- max subscription reached
- duplicate subscription id
- non matching subscription params e.g if no addresses are matched to any existing contracts
- subscription aborted (maybe when the relay is forced shut down - can we do this?)
- ... (any more)

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,281 @@
...
+
+| Error Codes |                                                       Error message                                                        |                                    Solution                                     |
+| :---------: | :------------------------------------------------------------------------------------------------------------------------: | :-----------------------------------------------------------------------------: |
+|    32600    | "The maximum batch size that can be sent over a websocket connection is 10. Please decrease the batch size and try again.” | Occurs when user attempts to send high-volume JSON-RPC traffic over Websockets. |
+
+## Limits:
+
+1. WebSockets in general are resource intensive. We could implement a throttling mechanism. Every live socket connection would also result in additional requests to the Mirror Node every n seconds. This means that we need to introduce a rate limiting feature, like 5 concurrent connections by IP.
```
Nana-EC, line 260: 100%, let's provide a suggestion for a max number of socket connections and maybe the max subscription time to ensure resources aren't locked up for too long.
The last suggestion might be problematic but let's explore it

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,281 @@
...
+#### Milestone 1 (done)
+
+1. Finalize design document
+
+#### Milestone 2
+
+1. Implement library for websocket support in the server.
+2. Extend current koa server logic.
+3. Introduce `eth_subscribe` and `eth_unsubscribe` to both server and relay part.
+4. Add logic to save all subsciption IDs and it's corresponding filter to a cache.
```
Nana-EC, line 273: Add a step for filter param support.
1st maybe single address, then topics, then multiple contracts etc.
Just a suggestion but I'd like us to take a staged approach here.

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,281 @@
...
+5. Add scheduled job function for looping through all saved filters and returns result to it's corresponding socket with the respective subscription id.
+6. Extend current ratelimiter class to support websocket limitation.
+7. Add logic to delete subsciption upon connection break or request.
+
+#### Milestone 3 (use Mirror-Node GraphQL API)
+
+1. Remove scheduled job, as this is going to be handled by the mirror-node graphql module.
+2. Forward all request for every coresponsing subscription id to the mirror node and the response back to the requester.
```
Nana-EC, line 281:
```suggestion
2. Forward all requests for every corresponding subscription id to the mirror node and the return the response back to the requester.
```

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,281 @@
...
+
+1. Implement library for websocket support in the server.
+2. Extend current koa server logic.
+3. Introduce `eth_subscribe` and `eth_unsubscribe` to both server and relay part.
+4. Add logic to save all subsciption IDs and it's corresponding filter to a cache.
+5. Add scheduled job function for looping through all saved filters and returns result to it's corresponding socket with the respective subscription id.
+6. Extend current ratelimiter class to support websocket limitation.
+7. Add logic to delete subsciption upon connection break or request.
+
+#### Milestone 3 (use Mirror-Node GraphQL API)
```
Nana-EC, line 278: Note that the Mirror Node has not yet design this feature at the time of writing and there is no current schedule for when the relay would adopt this

Nana-EC (changes requested):
Looking good.
Some suggestions

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,331 @@
...
+## Goals
+
+1. Implement websocket subscription in the relay
+2. Introduce two new methods - `eth_subscribe` and `eth_unsubscribe`
+3. Support subscription for event logs
+
+## Non-Goals
+
+1. Add long-term memory to the relay
+2. Reduce the efficiency of the relay or mirror node due to high resource usage
```
Nana-EC, line 16:
```suggestion
2. Reduce the efficiency of the relay or mirror node due to high resource usage
3. Support subscription for new blocks and pending transactions
```

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,331 @@
+# Subscription design
+
+## Purpose
+
+The JSON-RPC Relay currently doesn't support subscription using websocket as is common in many other relays and is expected by many web3 tool, which makes polling for event logs, new blocks and transactions very hard for developers.
```
Nana-EC, line 5: nit: maybe i missed this in my previous suggestion
```suggestion
The JSON-RPC Relay currently doesn't support subscription using websocket as is common in many other relays and is expected by many web3 tools, which makes polling for event logs, new blocks and transactions very hard for developers.
```

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,331 @@
...
+
+## Architecture
+
+### Subscription Types
+
+|   Subscription Type    |                                  Description                                   |           Support           |
+| :--------------------: | :----------------------------------------------------------------------------: | :-------------------------: |
+|          logs          |      Emits logs attached to a new block that match certain topic filters.      |       Yes, initially        |
+|        newHeads        |               Emits new blocks that are added to the blockchain.               | No, maybe in future release |
+| newPendingTransactions | Emits transaction hashes that are sent to the network and marked as `pending`. | No, maybe in future release |
```
Nana-EC, line 26: Probably not right for this suggested reason
```suggestion
| newPendingTransactions | Emits transaction hashes that are sent to the network and marked as `pending`. | No, as Hedera does not have pending transactions on a node. |
```

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,331 @@
...
+
+### Initiating a subscription
+
+#### Request
+
+```javascript
+// initiate websocket stream first
+wscat -c wss://testnet.hashio.io/api
```
Nana-EC, line 34: Don't hardcode testnet, so people know they can configure it per their needs
```suggestion
wscat -c wss://<env>.hashio.io/api
```

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,332 @@
...
+2. Implement a global limit for maximum concurrent connections.
+3. Explore the possibility of adding a TTL limit for every connection.
+4. Explore the option to add a limit for the total subscribed time per IP ( `sum(connection * subscribed_time)` ). There could theoretically be a way to charge users to increase this limit, for example through via a smart contract. Alternatively users could be encouraged to run their own instances of the relay.
+5. Users should be required to renew their subscription in the case of an error. If for some reason the relay is restarted it should not automatically try to restore all previous connections and subscriptions.
+
+## Open Questions
+
+1. How will we limit connections and subscriptions?
+2. Is it reasonable to add a TTL timer to subscriptions?
+
```
ebadiere, line 302: Have we thought through the expectations around subscriptions and performance?  Ethereum's block times are approximately 13 seconds.  Will this architecture support the polling of the mirror node every 13 seconds with out triggering rate limits and throttling?  I am asking because some of this transaction related event information can be extremely time sensitive when maintaining asset prices on various protocols, e.g. arbitrage between liquidity pools, oracles adjusting prices based on external events, etc.

Ivo-Yankov: I'm not sure what the exact rate limits are for the mirror node, but I did a quick test and it can easily handle at least 100 requests per second.

ebadiere: Okay, that's good news.

lukelee-sl:
Has any consideration been given to the idea of isolating this functionality in it's own process/pod so that any issues with websockets in particular do not affect the functionality of the rest of the json rpc relay?

Ivo-Yankov:
> Has any consideration been given to the idea of isolating this functionality in it's own process/pod so that any issues with websockets in particular do not affect the functionality of the rest of the json rpc relay?
> 
> If we reject this idea then let's call out the explicit considered but rejected idea.

I haven't considered that idea, but it does make sense. The websocket connection will need to listen on a separate port. And if it runs in a different process then the logs would be easier to read and manage.

Another option is to just have a feature flag that enables the whole socket logic.

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,335 @@
...
+1. Implement an IP based limit for maximum concurrent connections.
+2. Implement a global limit for maximum concurrent connections.
+3. Explore the possibility of adding a TTL limit for every connection.
+4. Explore the option to add a limit for the total subscribed time per IP ( `sum(connection * subscribed_time)` ). There could theoretically be a way to charge users to increase this limit, for example through via a smart contract. Alternatively users could be encouraged to run their own instances of the relay.
+5. Users should be required to renew their subscription in the case of an error. If for some reason the relay is restarted it should not automatically try to restore all previous connections and subscriptions.
+
+## Open Questions
+
+1. How will we limit connections and subscriptions?
```
Nana-EC, line 302: Have we answered this or is it still open?
The `Non-Functional Requirements` above should answer some of this right?

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,335 @@
...
+## Non-Functional Requirements
+
+WebSockets in general are resource intensive. We could implement a throttling mechanism. Every live socket connection would also result in additional requests to the Mirror Node every n seconds. This means that we need to introduce a rate limiting feature, like 5 concurrent connections by IP.
+
+1. Implement an IP based limit for maximum concurrent connections.
+2. Implement a global limit for maximum concurrent connections.
+3. Explore the possibility of adding a TTL limit for every connection.
+4. Explore the option to add a limit for the total subscribed time per IP ( `sum(connection * subscribed_time)` ). There could theoretically be a way to charge users to increase this limit, for example through via a smart contract. Alternatively users could be encouraged to run their own instances of the relay.
+5. Users should be required to renew their subscription in the case of an error. If for some reason the relay is restarted it should not automatically try to restore all previous connections and subscriptions.
+
```
AlfredoG87, line 299: Should we also explore the possibility of auto-ending some subscriptions when they are responded, for example:
 1. User subsribes to a transaction by hash, that is on pending state.
 2. Relay receives the transaction once is mined and included in a block and returns the value to the subscriber.
 3. since that tx is final, is certain there wont be more changes, user should unsubscribe, but in case he does not, we automatically do at the time of response.

AlfredoG87: I already answered this from the meeting.

AlfredoG87:
As per our conversation we should include on the design/HIP the metrics we would like to capture for this feature:

1. How many `eth_subscribe` and `eth_unsubscribe` do we get.
2. How long subscriptions stay alive/open
3. How many request de we perform to the mirror node for subscriptions
4. How many logs do we process/consume from the mirror node for subscriptions

AlfredoG87:
@Ivo-Yankov 
cc: @Nana-EC @georgi-l95 @ebadiere 

This are the offered metrics of `eth_getLogs`

**Mainnet:**
Last 24 hrs: 55,464 requests
Last 48 hrs: 114,064 requests

**Last 7 Days:** 
**Mean:** 58,985
**Min:** 56,907 
**Max:** 61,742

image

Last 30 days:
image

This graph only shows that we don't have enough historical data to accurately predict the actual usage of the feature, taking into consideration the pushes for more adoption and unexpected waves of growth that could happen in the next couple months.






**Last 24 hrs, as relative percentage of total requests:**
image

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,366 @@
...
+### Logs
+
+     The `logs` subscription type emits logs that match a specified topic filter and are included in newly added blocks.
+
+#### Parameters
+
+An object with the following fields:
+
+- address (optional): [`string`] or [`array of strings`] Singular address or array of addresses. Only logs created from one of these addresses will be emitted.
```
Nana-EC, line 57: As part of release we can prioritize `string` and have `array of strings` follow in a future release.
These should be separate tickets

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,366 @@
...
+      "logIndex": "0x0",
+      "topics": ["0xd78a0cb8bb633d06981248b816e7bd33c2a35a6089241d099fa519e361cab902"],
+      "transactionHash": "0xe044554a0a55067caafd07f8020ab9f2af60bdfe337e395ecd84b4877a3d1ab4",
+      "transactionIndex": "0x0"
+    }
+  }
+}
+```
+
+### newHeads
```
Nana-EC, line 95: Drop the `newHeads` and `newPendingTransactions` sections since they aren't part of the design

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,366 @@
...
+    "jsonrpc":"2.0",
+    "method":"eth_subscription",
+    "params":{
+        "subscription": SUBSCRIPTION_ID,
+        "result":"0xd6fdc5cc41a9959e922f30cb772a9aef46f4daea279307bc5f7024edc4ccd7fa"
+    }
+}
+```
+
+## Mock code
```
Nana-EC, line 171: If this mock code was inspired we should credit the inspiration source here.

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,366 @@
...
+  {
+    address: '0x123456..',
+    topics: ['0x12345...'],
+  },
+  function (error, result) {
+    if (!error) console.log(result);
+  }
+);
+
+// unsubscribes the subscriptionsubscription.unsubscribe(function(error, success){
```
Nana-EC, line 251:
```suggestion
// unsubscribes the subscription
subscription.unsubscribe(function(error, success){
```

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,366 @@
...
+if (success) {
+  console.log('Successfully unsubscribed!');
+}
+```
+
+### Mirror node polling:
+
+1. Maintain a list of all live socket connections. This can be done with a `key-value` store where the `key` is a unique identifier (some ws libraries provide such an id, but if needed it can be generated by the relay) of the socket object, and the `value` is the socket object itself.
+2. Maintain a list of all active subscriptions and the corresponding socket objects. It should have the following structure:
```
Nana-EC, line 260: To confirm is this implying `count(live)` >= `count(active)`.
Also for every new subscription are we picking up from the list of remaining live connections. Meaning live connections always remain open and it's more of a queue system to assign and release usage of it?

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,366 @@
...
+|     32609     |                                                 Invalid subscription id                                                  | Occurs when user tries to unsubscribe from a non-existing subscription or one that is not associated with the active socket connection. |
+|     32610     |                                    Invalid subscription parameters: {detailed reason}                                    |               Occurs when there is an error due to the provided subscription parameters, i.e. the address does not exist.               |
+|     32611     |                                                   Subscription aborted                                                   |                          Occurs when the subscription is aborted by the relay, i.e. a maximum TTL is reached.                           |
+
+## Rate Limits
+
+Active subscriptions should be limited in some way. We should create a flexible and configurable system for limiting subscriptions on a global and an IP level. This should be done by extending the existing `RateLimit` class as needed.
+
+1. Add a configurable global limit on the total active subscriptions.
```
Nana-EC, line 294:
```suggestion
1. Add a configurable global limit on the total active subscriptions. Initial default of 10.
```

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,366 @@
...
+2. Add a configurable limit on the total active subscriptions per IP.
```
Nana-EC, line 295:
```suggestion
2. Add a configurable limit on the total active subscriptions per IP. Initial default of 1.
```

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,366 @@
...
+3. Add a TTL to subscriptions, i. e. automatically terminate a subscription after a configurable time period.
```
Nana-EC, line 296:
```suggestion
3. Add a TTL to subscriptions, i. e. automatically terminate a subscription after a configurable time period. Initial default of 5 mins. 
```

docs/design/eth-subscribe.md
```diff
@@ -0,0 +1,366 @@
...
+8. Unsubscribing due to automatic subscription termination. 
+9. E2E test using popular libraries (`ethers.js WebSocketProvider`) that include - connecting, subscribing, receiving data, unsubscribing
+
+## Non-Functional Requirements
+
+Users should be required to renew their subscription in the case of an error or if the subscription was closed from the relay side. If for some reason the relay is restarted it should not automatically try to restore all previous connections and subscriptions.
+
+## Open Questions
+
+1. How is the deployment going to be handled?
```
Nana-EC, line 329: Do we have answers yet?
Initially we'll want the same container instance to serve these with a future option to split out as needed.
This means there should be an item to ensure container start manages both processes.

Nana-EC (commented):
Looking good, minor suggestions and we should be able to merge this.

docs/design/eth-subscribe.md
Ivo-Yankov: It wasn't inspired, we wrote it when researching how `eth_subscribe` can be implemented.

docs/design/eth-subscribe.md
Ivo-Yankov: A single socket connection can have 0, 1 or multiple subscriptions. It is created and destroyed when a client connects/disconnects from the ws-server. Subscriptions are created when `eth_subscribe` is called and are destroyed on `eth_unsubscribe` or when the associated connection is broken.

Nana-EC (approved):

docs/design/eth-subscribe.md
Nana-EC: Great, thanks

Ivo-Yankov (approved):

PR #898, opened 2023-02-15
865-update-logs-and-metrics.  Added the requestId to the server context.

ebadiere:
This PR modifies rate limited logs in order to include the requestId in a standard log message.

Adds a Koa requestId package that supports storing the requestId in the application context.
Fixes https://github.com/hashgraph/hedera-json-rpc-relay/pull/885

Notes for reviewer:
Provides the request ID in the logs right as the rateLimit for a given IP is triggered, in the standard log format.

Existing tests cover rateLimited requests.

### DIFF
--- a/packages/server/src/formatters.ts
+++ b/packages/server/src/formatters.ts
@@ -1,0 +1,30 @@
+/*-
+ *
+ * Hedera JSON RPC Relay
+ *
+ * Copyright (C) 2022 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


packages/server/src/formatters.ts
```diff
@@ -0,0 +1,30 @@
+/*-
+ *
+ * Hedera JSON RPC Relay
+ *
+ * Copyright (C) 2022 Hedera Hashgraph, LLC
```
Nana-EC, line 5: 2023

packages/server/src/server.ts
```diff
@@ -134,11 +135,48 @@ app.getKoaApp().use(async (ctx, next) => {
   }
 });
 
+app.getKoaApp().use(async (ctx, next) => {
```
Nana-EC, line 138: Let's look to rework this method next sprint.
Could see readability and testability improvements

Nana-EC (approved):
LG
DCO also

packages/server/src/formatters.ts
ebadiere: Updated.

packages/server/src/server.ts
ebadiere: Okay

Nana-EC (approved):

PR #902, opened 2023-02-15
Bump 0.19.0-SNAPSHOT

Nana-EC:
**Description**:
Bump 0.19.0-SNAPSHOT

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -3,7 +3,7 @@
     "info": {
         "title": "Hedera JSON-RPC Specification",
         "description": "A specification of the implemented Ethereum JSON RPC APIs interface for Hedera clients and adheres to the Ethereum execution APIs schema.",
-        "version": "0.18.0-SNAPSHOT"
+        "version": "0.19.0-SNAPSHOT"
     },
     "methods": [
         {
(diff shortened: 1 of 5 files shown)


david-bakin-sl (approved):
Looks like they're all here.  Do we do this with a tool? Or manually?  I see these version bumps _all the time_ and each time I have the same question: are they all here?  I do a full text search to find out.  (Probably there's a better way.)  Anyway: Isn't there some way we can set up a CI check to _automate_ knowing whether _all_ needed files were changed (off some manifest file maybe)?

Nana-EC:
> Looks like they're all here. Do we do this with a tool? Or manually? I see these version bumps _all the time_ and each time I have the same question: are they all here? I do a full text search to find out. (Probably there's a better way.) Anyway: Isn't there some way we can set up a CI check to _automate_ knowing whether _all_ needed files were changed (off some manifest file maybe)?

Thanks @david-bakin-sl this has been in my head to do but just captured it in an issue here - https://github.com/hashgraph/hedera-json-rpc-relay/issues/903

PR #919, opened 2023-02-17
893 Added support to configure SDK requestTimeOut.

### LINKED ISSUE #893: Add support to configure SDK requestTimeOut
### Problem

Currently we do not configure the timeout value used by the SDK on calls to the consensus node

### Solution

We should support configuration

- Update the configs and docs with an SDK_REQUEST_TIMEOUT value. Add this value to the startup logs emitted 
- Update `relay. initClient()` to use `setRequestTimeOut` on the SDK client with the env param

See https://github.com/hashgraph/hedera-services/issues/4930#issuecomment-1426184333 for the recommended values

### Alternatives

_No response_

ebadiere:
Currently we do not configure the timeout value used by the SDK on calls to the consensus node

This PR modifies the client in the relay in order to support setting the requestTimeout parameter.

Fixes #893


packages/relay/src/lib/relay.ts
```diff
@@ -124,7 +124,11 @@ export class RelayImpl implements Relay {
 
     client.setTransportSecurity(process.env.CLIENT_TRANSPORT_SECURITY === 'true' || false);
 
-    logger.info(`SDK client successfully configured to ${JSON.stringify(hederaNetwork)} for account ${client.operatorAccountId}`);
+    const sdkRequestTimeout = parseInt(process.env.SDK_REQUEST_TIMEOUT || '');
+    client.setRequestTimeout(Number.isInteger(sdkRequestTimeout) ? sdkRequestTimeout : 1000);
```
lukelee-sl, line 128: Is 1000 (I assume this is milliseconds) a good default value?  Do we know how much time calls to the consensus nodes typically take?

ebadiere: Yes it's milliseconds and good question.  Locally it seems to work with no issues.  Tests pass.  I am currently running/testing against hashio services and will have results back soon.

Nana-EC, line 128: Is this a simpler approach to then skip the `Number.isInteger`?
```suggestion
    client.setRequestTimeout(parseInt(process.env.SDK_REQUEST_TIMEOUT || '1000'));
```

Nana-EC: Check w services and sdk team

Nana-EC (changes requested):
Slight improvement and also curious on Luke's question of the valid default

packages/relay/src/lib/relay.ts
```diff
@@ -123,8 +123,10 @@ export class RelayImpl implements Relay {
     }
 
     client.setTransportSecurity(process.env.CLIENT_TRANSPORT_SECURITY === 'true' || false);
+    client.setRequestTimeout(parseInt(process.env.SDK_REQUEST_TIMEOUT || ''));
```
georgi-l95, line 126: Add this new env var to env.example, readme, localAcceptance, helm-chart. Also if this is not set it's going to return NaN, because `parseInt('')` is NaN.

packages/relay/src/lib/relay.ts
ebadiere: I asked around and it seems that 10000 would be a good default value to start with.  I updated the README, test .env files and helm charts accordingly.

packages/relay/src/lib/relay.ts
ebadiere: Updated.

Nana-EC: I still see `parseInt('')` did you push?

Nana-EC: Should it be?
```suggestion
    client.setRequestTimeout(parseInt(process.env.SDK_REQUEST_TIMEOUT || '10000'));
```

packages/relay/src/lib/relay.ts
ebadiere: Updated.

Nana-EC (approved):

PR #931, opened 2023-02-21
Cache the account mirror node query on  `estimateGas` and `getTransactionByHash`

### LINKED ISSUE #915: Various caching improvements
### Problem

There are several places where it is possible to do some minor performance optimizations.

### Solution

1. `eth_sendRawTransaction` and `eth_getTransactionCount` make 2 calls to `/api/v1/accounts/{accountId}` with the same id.
2. The `gasPrice` is cached, but `eth_sendRawTransaction` doesn't use it, and instead calls the mirror node api again.
3. Cache the latest block number for a short time period that is than 2 seconds (500 ms should be fine).
4. We could cache the result of `mirrorNodeClient.getAccount` and use it in places where only permanent data is required:
- `estimateGas` uses it to check if the account exists
- `getTransactionByHash` uses it to get the `evm_address`

### Alternatives

_No response_

natanasow:
**Description**:

On `estimateGas` we're checking only the account's existence, so we can cache if the account has been found. On `getTransactionByHash` we can use the same cached account because there we are using only the `evm_address` value which is immutable.

**Related issue(s)**:

Fixes #915 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -332,9 +332,21 @@ export class EthImpl implements Eth {
     this.logger.trace(`${requestIdPrefix} estimateGas(transaction=${JSON.stringify(transaction)}, _blockParam=${_blockParam})`);
     //this checks whether this is a transfer transaction and not a contract function execution
     if (!transaction || !transaction.data || transaction.data === '0x') {
-      const toAccount = await this.mirrorNodeClient.getAccount(transaction.to);
+      const accountCacheKey = `account_${transaction.to}`;
+      let toAccount: object | null = this.cache.get(accountCacheKey);
+      if (!toAccount) {
+        toAccount = await this.mirrorNodeClient.getAccount(transaction.to);
+      }
+
       // when account exists return default base gas, otherwise return the minimum amount of gas to create an account entity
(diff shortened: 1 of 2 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -332,9 +332,21 @@ export class EthImpl implements Eth {
     this.logger.trace(`${requestIdPrefix} estimateGas(transaction=${JSON.stringify(transaction)}, _blockParam=${_blockParam})`);
     //this checks whether this is a transfer transaction and not a contract function execution
     if (!transaction || !transaction.data || transaction.data === '0x') {
-      const toAccount = await this.mirrorNodeClient.getAccount(transaction.to);
+      const accountCacheKey = `account_${transaction.to}`;
+      let toAccount: object | null = this.cache.get(accountCacheKey);
```
Nana-EC, line 336: This is a cache addition not an improvement.
Please update the PR description with the goal and changes here

Nana-EC (changes requested):
looking good, missing test cases to verify these 2 caches are hit

Nana-EC (approved):

georgi-l95 (approved):
LGTM

PR #934, opened 2023-02-22
Add exponential backoff to axios retry

### LINKED ISSUE #689: Introduce exponential backoff for axios-retry
### Problem

We have implemented `axios-retry` for retrying failed request to the mirror node. However if we add exponential backoff, some of tests are not working -  Dapp tests, for example.

### Solution

Investigate why adding exponential backoff breaks test and fix it.

### Alternatives

_No response_

georgi-l95:
**Description**:
Adds exponential backoff to axios retry and change configurable retry time to the most efficient.
Screenshot 2023-02-22 at 18 54 00

With this change, we'll have most of the optimisations merged. From here onwards we'll have to look for optimisations to other components (mirror-node, network-node).

**Related issue(s)**:

Fixes #689 

**Notes for reviewer**:

**Checklist**


lukelee-sl:
I'm not exactly sure what the chart means but seem like we are waiting more with each iteration and thus the total time increases exponentially.

lukelee-sl (approved):

Nana-EC (approved):

georgi-l95:
@lukelee-sl Yeah, sorry, I should've explained the charts better, those are E2E test runs before this change and after. As we are aiming to make local node development faster, this is a step in that direction.

PR #938, opened 2023-02-22
865 update logs and metrics

ebadiere:
This PR modifies the koaJsonRpc in order to log errors with response codes other than 200s.


packages/server/src/server.ts
```diff
@@ -189,7 +188,7 @@ const logAndHandleResponse = async (methodName: any, methodParams: any, methodFu
 
     const response = await methodFunction(requestId);
     const status = response instanceof JsonRpcError ? response.code.toString() : responseSuccessStatusCode;
-    ms = Date.now() - start;
+    ms = Date.now() - app.getStartTimestamp();
```
Nana-EC, line 191: This is here just for metrics right?
Should we be capturing metrics in `index.ts` instead?
I think the measurements would more accurate since this is the outer logic

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -128,16 +144,22 @@ export default class KoaJsonRpc {
         if (e instanceof InvalidParamsError) {
           ctx.body = jsonResp(body.id, new InvalidParamsError(e.message), undefined);
           ctx.status = 400;
+          ms = Date.now() - this.startTimestamp;
+          this.logger.error(`${messagePrefix} ${ctx.status} (INVALID PARAMS ERROR) ${ms} ms`);
```
Nana-EC, line 148: We've added error and warning case. 
I think we're missing bringing the info case for successful calls that would produce a 200.
Might be good to have all of the logs in one file

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -104,21 +115,26 @@ export default class KoaJsonRpc {
       ) {
         ctx.body = jsonResp(body.id || null, new InvalidRequest(), undefined);
         ctx.status = 400;
+        ms = Date.now() - this.startTimestamp;
+        this.logger.error(`${messagePrefix} ${ctx.status} (INVALID REQUEST) ${ms} ms`);
```
Nana-EC, line 119: nit: `INVALID REQUEST`, 1METHOD NOT FOUND` and the rest should be const class variables.

Nana-EC (changes requested):
Looking good. Suggestions on movement and

packages/server/src/koaJsonRpc/index.ts
ebadiere: Updated.

packages/server/src/koaJsonRpc/index.ts
ebadiere: The case for successful calls that produce a 200 is logged in the server.ts in the logAndHandleResponse function.  The format for the message logged was taken from that.

Nana-EC: Per offline convo, seems like having all errors in this file is okay with success in `server.ts`

packages/server/src/server.ts
ebadiere: The index.ts is as minimal as it gets.  The server.ts executes the outer shell.

Nana-EC (approved):

PR #939, opened 2023-02-23
Add a gas/s metric for gas consuming endpoints

### LINKED ISSUE #876: Add a gas/s metric for gas consuming endpoints
### Problem

Currently we capture the HBAR spent on gas consuming calls but we don't capture the actual gas

### Solution

Add a metric that captures the amount gas consumed on calls.

Note: This may require a mirror node call to view contract results for "Gas Used"
Note 2: Contract calls do not emit records so it's difficult to capture gas used in that case.
Note 3: gas charged = 0.8 * gas limit. Also gas charged >= gas burnt. Example from Danno. "If I did a 35k gas defi trade, the gas limit at 300k, the gas used will be 240k.  We should call this "gas charged" even though the EVM only used 35k, "gas burnt."..."

### Alternatives

_No response_

natanasow:
**Description**:

Currently, we capture the HBAR spent on gas-consuming calls but we don't capture the actual gas

Note: This may require a mirror node call to view contract results for "Gas Used"
Note 2: Contract calls do not emit records so it's difficult to capture gas used in that case.
Note 3: gas charged = 0.8 * gas limit. Also gas charged >= gas burnt. Example from Danno. "If I did a 35k gas defi trade, the gas limit at 300k, the gas used will be 240k.  We should call this "gas charged" even though the EVM only used 35k, "gas burnt."..."

**Related issue(s)**:

Fixes #876 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -487,15 +491,23 @@ export class SDKClient {
...
         const resolvedCost = cost ? cost : 0;
+        const resolvedGas = typeof gas === 'object' ? gas.toInt() : 0;
         this.consensusNodeClientHistorgram.labels(
             mode,
             type,
             status,
             caller,
             interactingEntity)
             .observe(resolvedCost);
+        this.consensusNodeClientHistorgram.labels(
```
Nana-EC, line 504: I'm unclear how we'll differentiate between the hbar cost and then the gas cost being observed.
Using the same labels will mix up items no? Did you mean to add a new metric or label?

Nana-EC (changes requested):

Nana-EC:
Please add a description

packages/relay/src/lib/clients/sdkClient.ts
AlfredoG87: I agree that we should separate concerns for Spent HBar by Operator and Gas Costs incurred by user in separate metrics.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -113,7 +113,7 @@ export class SDKClient {
         this.consensusNodeClientHistorgram = new Histogram({
             name: metricHistogramName,
             help: 'Relay consensusnode mode type status cost histogram',
-            labelNames: ['mode', 'type', 'status', 'caller', 'interactingEntity'],
+            labelNames: ['mode', 'type', 'status', 'caller', 'interactingEntity', 'type'],
```
AlfredoG87, line 116: we already have a type label, if we were to use the same metric, we should reuse the type label, and registered as gas.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -487,15 +491,25 @@ export class SDKClient {
...
+            interactingEntity,
+            'cost')
             .observe(resolvedCost);
+        this.consensusNodeClientHistorgram.labels(
+            mode,
+            type,
+            status,
+            caller,
+            interactingEntity,
+            'gas')
```
AlfredoG87, line 511: same as above, if 'gas' is a type, it should be sent on the already type.
But since we have already a use-case for type (Transaction and Query) probably we would want to create a new metric for gas capturing.

When I mean a new metric I mean a new:
 `const metricHistogramName = 'rpc_relay_consensusnode_response';
        register.removeSingleMetric(metricHistogramName);
        this.consensusNodeClientHistorgram = new Histogram({`
        
Name could be: `rpc_relay_consensusnode_gasfee`

Nana-EC: Yeah, i think a separate metric would be cleaner.
If any aggregation is done without filtering for labels the values will be off.
Prior to this all observed values would be for hbar transaction cost.
As coded it'll mix hbar cost and gas charged.

Nana-EC (changes requested):
Per other suggestions let's make this clean and have a separate metric.
Preserve the same labels as i believe they are applicable

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -108,14 +109,22 @@ export class SDKClient {
         this.operatorAccountId = clientMain.operatorAccountId ? clientMain.operatorAccountId.toString() : 'UNKNOWN';
 
         // clear and create metrics in registry
-        const metricHistogramName = 'rpc_relay_consensusnode_response';
-        register.removeSingleMetric(metricHistogramName);
-        this.consensusNodeClientHistorgram = new Histogram({
-            name: metricHistogramName,
+        const metricHistogramCost = 'rpc_relay_consensusnode_cost';
```
AlfredoG87, line 112: On principle I agree that the current name "rpc_relay_consensusnode_response" is not descriptive enough, sounds more like if we were capturing the time of the response as opposed as the costs of the call, however, this name is already being used on the dashboard and if we were to change it, there would be a `before` and `after` on the data on prometheus + grafana, so we would not be able to see anything before this change. and also we would need to change all the widgets that currently are using the old metric name and update them to use the new metric name.

So am not sure if is worth it to change the name, or perphas we can leave it as is, and just enrich the description of the metric.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -108,14 +109,22 @@ export class SDKClient {
...
+            name: metricHistogramCost,
             help: 'Relay consensusnode mode type status cost histogram',
             labelNames: ['mode', 'type', 'status', 'caller', 'interactingEntity'],
             registers: [register]
         });
+        const metricHistogramGasFee = 'rpc_relay_consensusnode_gasfee';
+        register.removeSingleMetric(metricHistogramGasFee);
+        this.consensusNodeClientHistorgramGasFee = new Histogram({
+            name: metricHistogramGasFee,
+            help: 'Relay consensusnode mode type status gas fee histogram',
```
AlfredoG87, line 124: same as above, the Help field, is a text description that helps the consumer of the metrics to understand what this metric is being used for, I would change it for something like:
"Relay consensus node histogram for capturing amount of gas fees charged by mode, type, status, caller and interactingEntity"

AlfredoG87 (changes requested):

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC: Yeah, description captures that it's cost so let's leave the name as is

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -89,7 +89,8 @@ export class SDKClient {
      */
     private readonly hbarLimiter: HbarLimit;
 
-    private consensusNodeClientHistorgram;
+    private consensusNodeClientHistorgramCost;
+    private consensusNodeClientHistorgramGasFee;
```
Nana-EC, line 93:
```suggestion
    private consensusNodeClientHistogramGasFee;
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -89,7 +89,8 @@ export class SDKClient {
      */
     private readonly hbarLimiter: HbarLimit;
 
-    private consensusNodeClientHistorgram;
+    private consensusNodeClientHistorgramCost;
```
Nana-EC, line 92:
```suggestion
    private consensusNodeClientHistogramCost;
```

Nana-EC (dismissed):
Looking good, removing block.
Left some final suggestions.
Please get @AlfredoG87 additional sign off before merging.

packages/relay/src/lib/clients/sdkClient.ts
ebadiere: Agreed.  A more descriptive hint in the help if the text can fit, is preferable.

packages/relay/src/lib/clients/sdkClient.ts
AlfredoG87, line 112: Can we leave the name of the metric, name of the const is fine, but the name of the metric is important for the dashboards that we already have in place.

```suggestion
        const metricHistogramCost = 'rpc_relay_consensusnode_response';
```

AlfredoG87 (approved):
LGTM 👍 

Thank you for addressing all the comments 💯

PR #961, opened 2023-03-01
Add subscription handling class

### LINKED ISSUE #923: Save subscription data to memory

Ivo-Yankov:
**Description**:
Adds a `SubscriptionController` class, which is hidden behind a feature flag: `SUBSCRIPTIONS_ENABLED`

**Related issue(s)**:

Fixes #923 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/subscriptionController.ts
```diff
@@ -0,0 +1,134 @@
...
+import LRU from "lru-cache";
+import constants from "./constants";
+import crypto from "crypto";
+
+export interface Subscriber {
+    connection: any,
+    subscriptionId: string
+}
+
+const LOGGER_PREFIX = 'Subscriptions:';
```
georgi-l95, line 32: nit: I think we need to add a requestId, like the rest of the relay. But this might be done at a later stage ?

georgi-l95 (approved):
LG, just a minor nit

packages/relay/src/lib/subscriptionController.ts
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// import { Poller } from './poller';
```
Nana-EC, line 21:
```suggestion
```

packages/relay/src/lib/subscriptionController.ts
```diff
@@ -0,0 +1,134 @@
...
+
+export interface Subscriber {
+    connection: any,
+    subscriptionId: string
+}
+
+const LOGGER_PREFIX = 'Subscriptions:';
+
+export class SubscriptionController {
+    // private poller: Poller;
```
Nana-EC, line 35:
```suggestion
```

packages/relay/src/lib/subscriptionController.ts
```diff
@@ -0,0 +1,134 @@
...
+    private logger: Logger;
+    private subscriptions: {[key: string]: Subscriber[]};
+    private cache;
+
+    // constructor(poller: Poller, logger: Logger) {
```
Nana-EC, line 40:
```suggestion
```

packages/relay/src/lib/subscriptionController.ts
```diff
@@ -0,0 +1,134 @@
...
+    constructor(logger: Logger) {
+        // this.poller = poller;
```
Nana-EC, line 42:
```suggestion
```

packages/relay/src/lib/subscriptionController.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    subscribe(connection, event: string, filters?: {}) {
+
+        let tag: any = {event};
+        if (filters && Object.keys(filters).length) {
+            tag.filters = filters;
+        }
+
+        tag = JSON.stringify(tag);
```
Nana-EC, line 64: Q: are we concerned about the size of the tag string? Does it need to be shortened?

packages/relay/src/lib/subscriptionController.ts
```diff
@@ -0,0 +1,134 @@
...
+        const {id} = connection;
+
+        if (subId) {
+            this.logger.info(`${LOGGER_PREFIX} Unsubscribing connection ${id} from subscription ${subId}`);
+        }
+        else {
+            this.logger.info(`${LOGGER_PREFIX} Unsubscribing all instances of connection ${id}`);
+        }
+
+        for (const [tag, subs] of Object.entries(this.subscriptions)) {
```
Nana-EC, line 93: Not an issue at this scale but would it be an option to have a reverse map i.e. `subId` -> `tag` so that future traversals only need to cover 1 list of connections?

packages/relay/src/lib/subscriptionController.ts
```diff
@@ -0,0 +1,134 @@
...
+
+        if (!this.subscriptions[tag]) {
+            this.subscriptions[tag] = [];
+        }
+
+        const subId = this.generateId();
+
+        this.logger.info(`${LOGGER_PREFIX} New subscription ${subId}, listening for ${tag}`);
+
+        this.subscriptions[tag].push({
```
Nana-EC, line 74: Q: Might not be in this PR but are we adding a configureable  max size to this list?

Nana-EC (changes requested):

packages/relay/src/lib/subscriptionController.ts
Ivo-Yankov: Not really. The idea of this tag is to be unique to every event and filter combination and to provide an easy way to extract that data from it. We could create some more compressed format and optimize on some of the commas, quotation marks and curly brackets, but that would be ~5% of the total length at most, I think converting it to a simple json should be fine.

packages/relay/src/lib/subscriptionController.ts
Ivo-Yankov: Yes, but that would be at a later point.

packages/relay/src/lib/subscriptionController.ts
Ivo-Yankov: That would mean that we have to maintain 2 maps, one for `subId -> tag` and one for `tag -> subId`, which would mean that we keep twice as much data in memory.

Having only 1 map means that we have to loop either in `unsubscribe` or in `notifySubscribers`. And since `notifySubscribers` will be called every time when there is new data, I think looping in `unsubscribe` is the better option.

Nana-EC (dismissed):

natanasow (approved):

PR #962, opened 2023-03-01
Add Mirror node poller

### LINKED ISSUE #925: Add cron job for mirror node polling

Ivo-Yankov:
**Description**:

**Related issue(s)**:

Fixes #925 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/poller.ts
```diff
@@ -0,0 +1,132 @@
...
+                }
+                else if (event === 'newHeads') {
+                    // not supported
+                }
+                else if (event === 'newPendingTransacitons') {
+                    // not supported
+                }
+                else {
+                    // invalid event
+                }
```
georgi-l95, line 72: nit: should we return proper json rpc error here ?

packages/relay/src/lib/poller.ts
```diff
@@ -0,0 +1,132 @@
...
+import { Eth } from '../index';
+import { Logger } from 'pino';
+
+export interface Poll {
+    tag: string,
+    callback: Function,
+    lastPolled?: string
+}
+
+const LOGGER_PREFIX = 'Poller:';
```
georgi-l95, line 30: nit: I think we need to add a requestId, like the rest of the relay. But this might be done at a later stage ?

georgi-l95 (approved):
LG, minor nits

packages/relay/src/lib/poller.ts
```diff
@@ -0,0 +1,132 @@
...
+                        null,
+                        poll.lastPolled || this.latestBlock || 'latest',
+                        'latest',
+                        filters.address || null,
+                        filters.topics || null
+                    );
+
+                    poll.lastPolled = this.latestBlock;
+                }
+                else if (event === 'newHeads') {
```
Nana-EC, line 64: Remove `newHeads` and `newPendingTransacitons` and log `UNSUPPORTED` case as Georgi mentioned in another PR for else

packages/relay/src/lib/poller.ts
```diff
@@ -0,0 +1,132 @@
...
+            }
+        });
+    }
+
+    start() {
+        this.logger.info(`${LOGGER_PREFIX} Starting polling`);
+        this.interval = setInterval(async () => {
+            this.latestBlock = await this.eth.blockNumber();
+            this.poll();
+        }, 500);
```
Nana-EC, line 92: Make polling interval configureable

packages/relay/src/lib/poller.ts
Nana-EC: Seems like we should in this PR

Nana-EC (changes requested):

packages/relay/src/lib/poller.ts
Ivo-Yankov: There is no request in this case. The Poller constantly checks for new data and returns it to `SubscriptionController`, which in turn sends that data to all subscribed clients. The data sent to the clients is actually not in the JSON RPC format - it doesn't have the `id` and `version` fields.

.env.example
```diff
@@ -27,4 +27,5 @@ INPUT_SIZE_LIMIT=
 ETH_CALL_CONSENSUS=
 CONSENSUS_MAX_EXECUTION_TIME=
 ETH_CALL_CACHE_TTL=
-SDK_REQUEST_TIMEOUT=
\ No newline at end of file
+SDK_REQUEST_TIMEOUT=
+POLLING_INTERVAL=
```
Nana-EC, line 31: nit: new line

Nana-EC (approved):

natanasow (approved):

Issue #933, opened 2023-02-22
Request ID is missing from some mirror node retry logs


Nana-EC:
### Description

`createAxiosClient.axiosRetry.retryDelay()` has a log to share teh retry details on a mirror node call.
In some cases we're seeing the request Id missing from the logs

Instead of the below 
`[Request ID: c8368103-17dc-4884-8292-efcdcf3fdddf] Retry delay 500 ms`

we see
`[Request ID: ] Retry delay 500 ms`
or
`Retry delay 500 ms`

### Steps to reproduce

Search container logs with `textPayload:"Retry delay "` to see

### Hedera network

testnet

### Version

v0.18.0-rc2

### Operating system

None

ebadiere:
I can see this in the logs on mainnet.  I was able to reproduce this locally with calls to eth_estimateGas.  The estimateGas call to getAccount in the mirror node client was missing the requestId.

PR #964, opened 2023-03-01
993-Request-ID-is-missing-from-some-mirror-node-relay-logs.  This was…

ebadiere:
**Description**:

This PR fixes issue #933. It modifies the eth.ts in the relay to include the requestId in the mirror node client call, when calling eth_estimateGas.  This fixes the missing requestId in the mirror nodes delayed retry calls.


Nana-EC (approved):

AlfredoG87 (approved):
LGTM 👍

Issue #903, opened 2023-02-15
Automate git branch process for main and release branch


Nana-EC:
### Problem

The current release process sees the following manual steps
1. A PR to bump the SNAPSHOT version in `main`
2. Creation of a new `release/<major>.<minor>` branch and a PR to bump the version to an `alpha-1`/`rc-1` version

This results in us continuously questioning if we got every file in the bump PRs

### Solution

- [x] automate the git branch creation process on an rc1 or alpha1
- [x] automate the find and replace of sem ver values x.y.z with considerations for snapshot and rc or alphas
- [x] expose automation for bump version in a git hub action
- [x] expose git tag and draft release automation in a github action

Resources
- [Mirror Node Automated Release Process](https://github.com/hashgraph/hedera-mirror-node/blob/main/.github/workflows/release-automation.yml)
- https://cli.github.com/manual/gh_release

AlfredoG87:
@Nana-EC, I think this is completed, most of the items done by you 😅 
Will close ticket, but if you think otherwise feel free to re-open it.

PR #975, opened 2023-03-03
Added script and npm task for bumping version for all the needed file…

AlfredoG87:
**Description**:
Added script and npm task for bumping version for all the needed files, for easy bumping when doing releases and as intermediary step into automating release process using GHA

**Related issue(s)**:
#903 

Fixes #

**Notes for reviewer**:

https://user-images.githubusercontent.com/123040664/222847707-4498f0cb-59b5-4e25-ae17-3f6eb2ea4a87.mov

Due to how the *yaml* files are edited, they are re-written in a consistent manner, so the first time the script runs some parts of the file will have modifications that don't impact anything and are only of format, this side effect is nice in my opinion.

**Second Video after improvements:** 

https://user-images.githubusercontent.com/123040664/223514872-8079bee7-9296-4937-bfd8-4b2c5857ca2a.mov

**Checklist**

### DIFF
--- a/.bump-version.js
+++ b/.bump-version.js
@@ -1,0 +1,56 @@
+const replace = require("replace");
+const yaml = require('js-yaml');
+const fs = require('fs');
+const {execSync} = require('child_process')
+
+const versionRegex = /\d+\.\d+\.\d+(-\w+)?/i;
+const newVersion = process.env.SEM_VER;
+
+function checkVersion(semver){
+
+    if(semver == undefined || semver == '') {
+        console.log("semver cannot be blank");
+        process.exit(1);
+    }
+
+    if(semver.match(versionRegex) == undefined) {
+        console.log(`semver: ${semver} must be Semantic Version <Major>.<Minnor>.<Patch>[-<type>], examples: '0.20.0', '0.20.0-rc1'`)
+        process.exit(1);
+    }
+
+    return semver;
+}
+
+// check is a valid version
+checkVersion(newVersion);
+
+console.log(`Bumping version to: ${newVersion}`);
+
+// bumping version using replace for 'packages/relay/package.json', 'packages/server/package.json', 'docs/openrpc.json'
+replace({
+    regex: '\"version\": \"\\d+\\.\\d+\\.\\d+(-\\w+)?\"',
+    replacement: `"version": "${newVersion}"`,
+    paths: ['packages/relay/package.json', 'packages/server/package.json', 'docs/openrpc.json'],
+    recursive: false,
+    silent: false,
+});
+
+// bump docker-compose.js version
+const dockerComposePath = 'docker-compose.yml';
+let dockerComposeYml = yaml.load(fs.readFileSync(dockerComposePath));
+dockerComposeYml.services.relay.image=`ghcr.io/hashgraph/hedera-json-rpc-relay:${newVersion}`
+fs.writeFileSync(dockerComposePath, yaml.dump(dockerComposeYml));
+console.log(dockerComposePath);
+
+// bump helm chart versions
+const helmChartPath = 'helm-chart/Chart.yaml';
+let doc = yaml.load(fs.readFileSync(helmChartPath, 'utf-8'));
+doc.version = newVersion;
+doc.appVersion = `${newVersion}`;
+fs.writeFileSync(helmChartPath, yaml.dump(doc));
+console.log(helmChartPath);
+
+// bump package-lock.json using npm itself, by running install it will update packages versions based on previous files
+execSync("npm install")
+console.log("package-lock.json");
+console.log("All files were bumped successfully!")
--- a/scripts/.bump-version.js
+++ b/scripts/.bump-version.js
@@ -5,6 +5,7 @@ const {execSync} = require('child_process')
 
 const versionRegex = /\d+\.\d+\.\d+(-\w+)?/i;
 const newVersion = process.env.SEM_VER;
+const isSnapshot = process.env.SNAPSHOT ? process.env.SNAPSHOT == 'true' : false;
 
 function checkVersion(semver){
 
@@ -25,7 +26,7 @@ function checkVersion(semver){
 checkVersion(newVersion);
 
 console.log(`Bumping version to: ${newVersion}`);
-
+console.log(`is Snapshot: ${isSnapshot}`);
 // bumping version using replace for 'packages/relay/package.json', 'packages/server/package.json', 'docs/openrpc.json'
 replace({
     regex: '\"version\": \"\\d+\\.\\d+\\.\\d+(-\\w+)?\"',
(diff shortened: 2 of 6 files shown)


lukelee-sl (approved):
Looks awesome!  Thanks for working on this.

.bump-version.js
```diff
@@ -0,0 +1,56 @@
...
+
+// bump helm chart versions
+const helmChartPath = 'helm-chart/Chart.yaml';
+let doc = yaml.load(fs.readFileSync(helmChartPath, 'utf-8'));
+doc.version = newVersion;
+doc.appVersion = `${newVersion}`;
+fs.writeFileSync(helmChartPath, yaml.dump(doc));
+console.log(helmChartPath);
+
+// bump package-lock.json using npm itself, by running install it will update packages versions based on previous files
```
Nana-EC, line 53: there should be a bump for the `README.md` itself. 
There's a a link to the Open RPC spec.
When we branch release it should see a bump to the path e.g. https://github.com/hashgraph/hedera-json-rpc-relay/pull/951/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R18

This should only happen once so future bumps with the same major and minor versions would see no change

Nana-EC (changes requested):
Looks good. Minor suggestion and check for cases where version should remain or should be a partial replacement

Nana-EC:
Oh and I really like the video :)

.bump-version.js
```diff
@@ -0,0 +1,56 @@
+const replace = require("replace");
```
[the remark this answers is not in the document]
AlfredoG87: ✅  done

.bump-version.js
```diff
@@ -0,0 +1,56 @@
...
+let doc = yaml.load(fs.readFileSync(helmChartPath, 'utf-8'));
+doc.version = newVersion;
+doc.appVersion = `${newVersion}`;
+fs.writeFileSync(helmChartPath, yaml.dump(doc));
+console.log(helmChartPath);
+
+// bump package-lock.json using npm itself, by running install it will update packages versions based on previous files
+execSync("npm install")
+console.log("package-lock.json");
+console.log("All files were bumped successfully!")
```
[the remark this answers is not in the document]
AlfredoG87: ✅  done

.bump-version.js
```diff
@@ -0,0 +1,56 @@
...
+// bumping version using replace for 'packages/relay/package.json', 'packages/server/package.json', 'docs/openrpc.json'
+replace({
+    regex: '\"version\": \"\\d+\\.\\d+\\.\\d+(-\\w+)?\"',
+    replacement: `"version": "${newVersion}"`,
+    paths: ['packages/relay/package.json', 'packages/server/package.json', 'docs/openrpc.json'],
+    recursive: false,
+    silent: false,
+});
+
+// bump docker-compose.js version
```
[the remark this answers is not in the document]
AlfredoG87: No, the intention is to support all 3 types of version bumps (GA, RC and SNAPSHOTS), talking with Luke and taking a look at old bump PRs determined that the only one that change are the SNAPSHOTS since they don't change "main" in files `README.md` and `docker-compose.yml`

So I have made the change to add another parameter to the script to determine if is a snapshot bump, and bump those 2 files only when is not a snapshot.

.bump-version.js
AlfredoG87: handling README bumps now, only when is not an snapshot, as `main` branch has as "main" the version in the `README.md` file, similar to the `docker-compose.yml` file.

Nana-EC (dismissed):

georgi-l95 (commented):
When is this script going to be started ? Upon releasing ? 
I'm not sure if I'm a fan, because most of the packages have newer versions with breaking changes and we need to test extensively before updating them.
Also newer alpha versions of the network node and mirror node images sometimes have bugs, which needs to be tested.
If this is running on PRs, it's okay. But on releasing, we can very easily break the relay.

AlfredoG87:
> When is this script going to be started ? Upon releasing ? I'm not sure if I'm a fan, because most of the packages have newer versions with breaking changes and we need to test extensively before updating them. Also newer alpha versions of the network node and mirror node images sometimes have bugs, which needs to be tested. If this is running on PRs, it's okay. But on releasing, we can very easily break the relay.

@georgi-l95 
Right now it will be used manually whenever a new release needs a bump PR, to avoid missing some package or file that needs bumping, and in the future the idea is to generate an automated PR that bumps the version together with the release process, but always the bot/job would raise a PR that has to be reviewed and merged manually by one or more mainteners.

scripts/.bump-version.js
```diff
@@ -0,0 +1,72 @@
+const replace = require("replace");
+const yaml = require('js-yaml');
+const fs = require('fs');
+const {execSync} = require('child_process')
+
+const versionRegex = /\d+\.\d+\.\d+(-\w+)?/i;
```
david-bakin-sl, line 6: can this `versionRegex` _also_ be used in other places, e.g., L32 at least?  maybe the regex at L32 can be be constructed via concatenation or string interpolation or something?

david-bakin-sl (approved):
brief look!

david-bakin-sl:
Based on a conversation in standup right now, a further suggestion:  Maybe this script should include a grep over the repo to find files that contain lines that have a _match_ to a version number - and compare that to the list of files the script is modifying.  Then it could output a list of files that maybe _should_ have been bumped but weren't.  So that the user could know a) something was potentially missed and b) maybe this script should be updated.  (Might need an "ignore" list for false positives.)

ebadiere (commented):
Thanks for doing this.

scripts/.bump-version.js
AlfredoG87: in theory could be possible, but L6 regex is not a string is literally a regex type, and L32 regex is an string that gets converted into a regex by the replace package, and also they are 2 different regex, in practice, they are very similar and string concatenation could indeed work, but I prefer to have them as separate for ease of readability and peace of mind.
if you have a strong opinion on this I might try and make it work as suggested.

AlfredoG87:
> Based on a conversation in standup right now, a further suggestion: Maybe this script should include a grep over the repo to find files that contain lines that have a _match_ to a version number - and compare that to the list of files the script is modifying. Then it could output a list of files that maybe _should_ have been bumped but weren't. So that the user could know a) something was potentially missed and b) maybe this script should be updated. (Might need an "ignore" list for false positives.)

I think is better to keep tight control, not everyday we are adding packages or changing the release process, there are literally hundreds of ocurrences for versions and false positives could become a bigger problem than having to update the script everytime a new package is added or the release process is updated. imo.

AlfredoG87:
@Nana-EC @ebadiere @lukelee-sl @georgi-l95 
I just updated the script to include the newly added package: `ws-server` to the list of files that need bumping, in doing so, I also bring up the version of that package that was still as 0.18.0 to 0.20.0 that is the current one right now on main.

Other than that, everything is the same:

Files that get bumped now (with snapshot false and upgrading to a minor version), also the package-lock.json, now has 3 modifications, one for each sub-package.

image

Nana-EC (approved):

georgi-l95 (approved):
LGTM

PR #978, opened 2023-03-09
927 - Added socket connection limits.

ebadiere:
Description:

This PR add socket connection limits to the websocket server.

CONNECTION_LIMIT was added to the env.example and localAcceptance.env
CONNECTION_LIMIT was listed in the README.md
Notes for reviewer:
This was tested by setting the CONNECTION_LIMIT to a low value and adding connections. It was agreed to start with a connection limit of 10.


packages/ws-server/src/webSocketServer.ts
```diff
@@ -46,11 +46,20 @@ const LOGGER_PREFIX = 'WebSocket:';
...
 app.ws.use((ctx) => {
     ctx.websocket.id = relay.subs()?.generateId();
     logger.info(`${LOGGER_PREFIX} new connection ${ctx.websocket.id}`);
 
+    if (connectedClients >= MAX_CONNECTIONS) {
+        ctx.websocket.send("Connections exceeded");
+        ctx.websocket.close();
+        return;
+    }
+    connectedClients++;
```
Ivo-Yankov, line 61: I'm curious if the ordering here is correct. When the 11th client connects, the connection would be closed and the `close` event will be triggered, which will decrement `connectedClients`. I think then `connectedClients` will become `9` instead of remaining unchanged.

Could you please add a test to verify this?

david-bakin-sl: It's surprisingly hard to find documentation on whether the `onclose` event (or any other websocket event) is fired before `close()` returns or if it can be fired sometime later.  Is this something that's known for websocket `close()` in particular, or is a general behavior of JavaScript that I don't know?

If the answer is: "can be fired sometime after the return of `close()`" then a test with the existing code may sometimes work and sometimes not due to a race condition.  Maybe?

Nana-EC, line 61: Q: does the server or middleware itself not have a method to get the number of connected clients?
If it does we could just utilize that instead of our own counter which gets us into the counting problem noted below.

If not then maybe the counter should increment prior to the max connection check since it is alive.
In this way the close decrement logic will be consistent

packages/ws-server/src/webSocketServer.ts
```diff
@@ -46,11 +46,20 @@ const LOGGER_PREFIX = 'WebSocket:';
...
 const DEFAULT_ERROR = predefined.INTERNAL_ERROR();
+const MAX_CONNECTIONS = parseInt(process.env.CONNECTION_LIMIT || '10');
+let connectedClients = 0;
 
 app.ws.use((ctx) => {
     ctx.websocket.id = relay.subs()?.generateId();
     logger.info(`${LOGGER_PREFIX} new connection ${ctx.websocket.id}`);
 
+    if (connectedClients >= MAX_CONNECTIONS) {
+        ctx.websocket.send("Connections exceeded");
```
Nana-EC, line 57: We need a test for this.

Nana-EC (changes requested):

packages/ws-server/src/webSocketServer.ts
ebadiere: The koa-websockets package is a koa wrapper around the JS websockets (ws) package.  I am not seeing anything that tracks the number of connections in the WS package.  The only related identifier is the individual client IP address, which will not work behind a proxy.

AlfredoG87: @david-bakin-sl, does this helps answer your question? I think is fired **after always**.

https://github.com/websockets/ws/blob/7.4.6/lib/websocket-server.js#L146

Nana-EC: Sounds good then incrementing before  check should fix the issue

ebadiere: Yes, it does, but it's not in the koa-websockets, it's in the node websockets package that the koa-websockets wraps.  The latest commit now uses the verifyClient attribute in the constructor to pass in a callback function that handles the connections.

packages/ws-server/src/webSocketServer.ts
ebadiere: Done.

packages/ws-server/src/webSocketServer.ts
ebadiere: Some tests were added to verify the connection limits.  Some refactoring of the implementation was also done.

Ivo-Yankov (dismissed):
Looks good, I added just one nit comment.

packages/server/tests/acceptance/ws/subscribe.spec.ts
```diff
@@ -20,39 +20,121 @@
...
+        wsProvider = await new ethers.providers.WebSocketProvider(
             `ws://localhost:${process.env.WEB_SOCKET_PORT}`
         );
-    });
 
-    this.beforeEach(async () => {
         requestId = Utils.generateRequestId();
+        // Stabilizes the initial connection test.
+        await new Promise(resolve => setTimeout(resolve, 1000));
+        console.log("----------------------");
```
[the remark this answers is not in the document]
ebadiere: Done

packages/ws-server/src/webSocketServer.ts
```diff
@@ -98,6 +111,7 @@ app.ws.use((ctx) => {
     ctx.websocket.on('error', console.error);
 
     ctx.websocket.on('close', function () {
+        connectedClients--;
```
Nana-EC, line 114: Very interesting that it handles increments but not decrements of the connected client count.

packages/server/tests/acceptance/ws/subscribe.spec.ts
```diff
@@ -20,39 +20,120 @@
...
+            secondProvider.destroy();
+        });
+
+        it('Does not allow more connections than the connection limit', async function() {
+            // We already have one connection
+            for (let i = 1; i < parseInt(process.env.CONNECTION_LIMIT); i++) {
+                await establishConnection();
+            }
+
+            expect(server._connections).to.equal(parseInt(process.env.CONNECTION_LIMIT));
```
Nana-EC, line 129: Can we add a test goes through checking 0, n then 0 connections in a single session to make sure the increment and decrement logic in a single session is good when connection are added and closed.

packages/server/tests/acceptance/ws/subscribe.spec.ts
```diff
@@ -20,39 +20,120 @@
...
+import { finished } from "stream";
+import bodyParser from "koa-bodyparser";
 const {ethers} = require('ethers');
 
-describe('@web-socket Acceptance Tests', async function () {
+const FOUR_TWENTY_NINE_RESPONSE = 'Unexpected server response: 429';
+
+const establishConnection = async () => {
+        const provider = await new ethers.providers.WebSocketProvider(
+            `ws://localhost:${process.env.WEB_SOCKET_PORT}`
```
Nana-EC, line 35: Nit: make this a class variable set in the before All maybe.
We use it quite a few times.

Nana-EC (dismissed):
Looks good. One additional test suggestion

packages/server/tests/acceptance/ws/subscribe.spec.ts
ebadiere: Done.

packages/ws-server/src/webSocketServer.ts
ebadiere: Ugh, I thought I removed that. It's removed now.  Detailed explanation below.

packages/server/tests/acceptance/ws/subscribe.spec.ts
ebadiere: That counter decrement line "connectedClients--;" is confusing the conversation.  Apologies.  I thought I removed it earlier.  The number of connections is now simply read from the server context on line 65, 
```connectedClients = ctx.app.server._connections;``` and then checked in the verifyClient on line 49:
```
if (connectedClients >= MAX_CONNECTIONS) {
                return done(false, 429, 'Connection limit exceeded');
} 
```
If the limit has been reached, then the next connection is rejected on verifyClient before it even reached the decrement of the counter.  The tests pass because the limit breaking connection is immediately rejected and closed.

Ivo-Yankov (approved):

PR #997, opened 2023-03-20
Fix `eth_getBalance` returning wrong balance

### LINKED ISSUE #992: `eth_getBalance` is wrong in rare cases
### Description

In some cases when fetching balance of account using ethers, we are receiving wrong one.

### Steps to reproduce

1. Get one of the already created accounts with local-node
2. Fetch balance
3. Send some amount to this account
4. Compare new balance with old, which should be old+amount.

### Additional context

_No response_

### Hedera network

other

### Version

latest

### Operating system

None

georgi-l95:
**Description**:
There is a bug with `eth_getBalance` returning wrong balances. This happens because the number or BigNumber that's being calculated exceeds `MAX_SAFE_INTEGER` and calculation comes out wrong. We need to use data type that allows to do some basic calculation without exceeding `MAX_SAFE_INTEGER`, BigInt allows that.

**Related issue(s)**:

Fixes #992 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -621,7 +621,7 @@ export class EthImpl implements Eth {
 
     let blockNumber = null;
     let balanceFound = false;
-    let weibars: BigNumber | number = 0;
+    let weibars: BigInt = BigInt(0);
     const mirrorAccount = await this.mirrorNodeClient.getAccount(account, requestId);
 
     try {
@@ -665,7 +665,7 @@ export class EthImpl implements Eth {
                 }
 
                 balanceFound = true;
-                weibars = (currentBalance - balanceFromTxs) * constants.TINYBAR_TO_WEIBAR_COEF;
+                weibars = BigInt(currentBalance - balanceFromTxs) * BigInt(constants.TINYBAR_TO_WEIBAR_COEF);
               }
 
               // The block is NOT from the last 15 minutes, use /balances rest API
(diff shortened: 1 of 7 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -665,15 +665,15 @@ export class EthImpl implements Eth {
                 }
 
                 balanceFound = true;
-                weibars = (currentBalance - balanceFromTxs) * constants.TINYBAR_TO_WEIBAR_COEF;
+                weibars = BigInt(currentBalance - balanceFromTxs) * BigInt(constants.TINYBAR_TO_WEIBAR_COEF);
```
Nana-EC, line 668: Q: If a lot of these numbers were just BigInt by type definition would we need all the casting?
Or is this an optimization of sort?

Nana-EC (commented):
Looking good. Casting and type question

packages/relay/src/lib/eth.ts
georgi-l95: No optimization in mind. I haven't defined them as BigInt, because in some other places where we are using those variables and constants, they are needed as numbers and that means converting them back to number. Problem is that when we multiply them by the weibar coefficient, we are losing the precision, when they are numbers or bignumbers. I've cached this coefficient as bigInt at the beginning of the tests to avoid casting, but for the other numbers I can't avoid the casting either from BigInt to number or the other way around.

Nana-EC (approved):

natanasow (approved):
LGTM

PR #998, opened 2023-03-20
941 add unit tests for the poller class

### LINKED ISSUE #941: Add unit tests for the Poller class

ebadiere:
**Description**:
This PR adds unit test to the poller class.

This PR modifies the relay test suite by adding the poller.spec.ts.

Fixes #941

### DIFF
--- a/packages/relay/src/lib/poller.ts
+++ b/packages/relay/src/lib/poller.ts
@@ -119,6 +119,7 @@ export class Poller {
     }
 
     hasPoll(tag): boolean {
+        // Return boolean true if the polls array contains this tag
         return !!this.polls.filter(p => p.tag === tag).length;
     }
 
--- a/packages/relay/src/lib/poller.ts
+++ b/packages/relay/src/lib/poller.ts
@@ -119,7 +119,7 @@ export class Poller {
     }
 
     hasPoll(tag): boolean {
-        return !!this.polls.filter(p => p.tag !== tag).length;
+        return !!this.polls.filter(p => p.tag === tag).length;
     }
 
     isPolling() {
--- a/packages/relay/tests/lib/poller.spec.ts
+++ b/packages/relay/tests/lib/poller.spec.ts
@@ -1,0 +1,72 @@
+/*-
+ *
+ * Hedera JSON RPC Relay
+ *
+ * Copyright (C) 2023 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import { expect } from 'chai';
+import pino from 'pino';
+import { Poller } from '../../src/lib/poller';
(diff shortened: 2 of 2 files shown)


packages/relay/tests/lib/poller.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+           
+        });
+
+        it('should stop polling', (done) => {
+            const loggerSpy = sandbox.spy(logger, 'info');            
+            poller.remove(tag);
+            done();
+
+            expect(poller.isPolling()).to.be.false;
+            expect(loggerSpy.callCount).to.equal(5);
```
Nana-EC, line 88: It's not clear from the logic why the tests should expect 5 calls here and 2 above.
I'm guessing you're using the logs as a proxy to measure progression through the stack.

The number here isn't as useful as the actual log check.
If removed a log and added it back it would make no different to this test.

I'd say remove the count check and verify the actual logs expected as you are below.
In which case you're missing 4 and 5

packages/relay/tests/lib/poller.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+                expect(loggerSpy.getCall(0).args[0]).to.equal(FETCHING_DATA);                
+            };
+
+            await poll();   
+            expect(loggerSpy.getCall(1).args[0]).to.equal(ARRAY_OF_LOGS);
+        });
+
+    });
+
+});
```
Nana-EC, line 141: nit:  new line

Nana-EC (commented):
Looking good.
Would suggest testing the actual log vs the count

packages/relay/tests/lib/poller.spec.ts
ebadiere: Done.

packages/relay/tests/lib/poller.spec.ts
ebadiere: This is a good catch.  That line of code should have failed, but the evaluations were not run because the done() was called before them.  I originally added the done() to finish the asynchronous function call.  I refactored both tests using done() to use the more robust async/await.  This exposed what I think is a bug.  @Ivo-Yankov please review.  This is what I was trying to explain when you showed us how to run the websockets server.  

In the poller.ts the hasPoll function should return true when it finds a matched tag in the Poll array.  I updated line 122 from:
```return !!this.polls.filter(p => p.tag !== tag).length;```
 that returns false when the tags match to:
```return !!this.polls.filter(p => p.tag === tag).length;```

The first test in the poller.spec.ts, lines 71 to 73 cover this.

Comment (AlfredoG87) [reply packages/relay/tests/lib/poller.spec.ts]: @ebadiere

Yes this is indeed a bug, I found it during acceptance test for subscription, when doing multiple subscriptions, and today I found the issue and then thought that this might be present on your Unit Tests, and then realized you found it and fix it there too... 
https://github.com/hashgraph/hedera-json-rpc-relay/issues/988

**Review by Nana-EC — COMMENTED** [review]
Comment and new line.
LG otherwise

File: packages/relay/src/lib/poller.ts [hunk 3]
```diff
@@ -119,7 +119,7 @@ export class Poller {
     }
 
     hasPoll(tag): boolean {
-        return !!this.polls.filter(p => p.tag !== tag).length;
+        return !!this.polls.filter(p => p.tag === tag).length;
```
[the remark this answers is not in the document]
ebadiere: Done.

AlfredoG87 (approved):
LGTM 👍

packages/relay/tests/lib/poller.spec.ts
Nana-EC: Yeah there should be a line 135 with nothing on it.
Just makes it easier to git diff in the future.
It's a nit

Nana-EC (approved):

PR #1005, opened 2023-03-21
Switch eth_call to Mirror Node

Ivo-Yankov:
**Description**:
- Refactor of eth_call with added condition for using only consensus node for calls or mirror node and fallback to consensus node.
- Helm charts set for different env.
- Fixes erc20 tests as some of the tests actions need more time propagated to the mirror node.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
Depends on https://github.com/hashgraph/hedera-mirror-node/issues/5634
All the needed changes are merged in the mirror-node release 0.77.

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -970,7 +970,7 @@ export class EthImpl implements Eth {
    * @param call
    * @param blockParam
    */
-  async call(call: any, blockParam: string | null, requestId?: string): Promise<string | JsonRpcError> {
+  async call(call: any, blockParam: string | null, requestId?: string, isFallback?: boolean): Promise<string | JsonRpcError> {
```
Nana-EC, line 973: Hmm not a fan of modifying the method signature just to support an internal retry.
Might be better to pull out the 2 pathways into 2 method e.g. contractCallMirrorNode() & contractCallConsenssusNode()
Then something like
```
if (process.env.ETH_CALL_CONSENSUS == 'false') {
  try {
    ...
    contractCallMirrorNode();
    ...
  } catch (e: any) {
    ...
    if (e instanceof MirrorNodeClientError && e.isNotSupported()) {
        return contractCallConsenssusNode(...);
    }
  }
} else {
  ...
  contractCallConsenssusNode(...);
  ...
}
```

AlfredoG87: I agree on this, to improve code readability and maintenability

georgi-l95: Agree, I've refactored `eth_call` method.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -238,7 +239,14 @@ export class MirrorNodeClient {
         }
 
         this.logger.error(new Error(error.message), `${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`);
-        throw new MirrorNodeClientError(error.message, effectiveStatusCode);
+
+        const mirrorError = new MirrorNodeClientError(error, effectiveStatusCode);
+
+        if (mirrorError.isContractReverted() && !mirrorError.isNotSupported()) {
+            throw predefined.CONTRACT_REVERT(mirrorError.errorMessage);
```
Nana-EC, line 246: nit: comment here highlighting we want to distinguish unsupported mirror node evm functionality.

packages/relay/src/lib/errors/MirrorNodeClientError.ts
```diff
@@ -21,30 +21,49 @@
...
 
+    public isContractReverted(): boolean {
+        return this.statusCode === MirrorNodeClientError.ErrorCodes.CONTRACT_REVERT_EXECUTED;
+    }
+
     public isNotFound(): boolean {
       return this.statusCode === MirrorNodeClientError.statusCodes.NOT_FOUND;
     }
+
+    public isNotSupported(): boolean {
```
Nana-EC, line 66: Q: Is it okay to have an and here or is the name incorrect?
Are we aware of any unsupported cases that aren't precompile?
Should this instead be `isUnsupportedSystemContractOperaton()` with a separate `isNotSupported()`?

packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -67,7 +67,7 @@ describe('@erc20 Acceptance Tests', async function () {
         recipient = accounts[1].address;
         anotherAccount = accounts[2].address;
 
-        // alow mirror node a 2 full record stream write windows (2 sec) and a buffer to persist setup details
+        // allow mirror node a 5 full record stream write windows (5 sec) and a buffer to persist setup details
```
Nana-EC, line 70:
```suggestion
        // allow mirror node 5 full record stream write windows (2 secs each) and a buffer to persist setup details
```

packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -163,6 +163,8 @@ describe('@erc20 Acceptance Tests', async function () {
 
                                 it ('@release contract owner transfers tokens', async function () {
                                     tx = await contract.connect(tokenOwnerWallet).transfer(to, amount);
+                                    // allow mirror node a 5 full record stream write windows (5 sec) and a buffer to persist setup details
```
Nana-EC, line 166: same fix as above and those below

packages/server/tests/acceptance/erc20.spec.ts
Nana-EC: Also if the comment is true then the timeout would become 11 secs.
That is each window is 2 secs. 5 windows is 10 secs with a 1 sec buffer.
Is the intent still applicable and the effect still desired?

Nana-EC (changes requested):
Looking good.
Some nits, suggestions and questions

helm-chart/values.yaml
```diff
@@ -125,6 +125,7 @@ config:
   CONSENSUS_MAX_EXECUTION_TIME: 15000
   SDK_REQUEST_TIMEOUT: 10000
   SUBSCRIPTIONS_ENABLED: false
+  ETH_CALL_CONSENSUS: true
```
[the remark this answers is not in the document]
georgi-l95: This is a little confusing, but `eth_call_consensus` set to true means, use only consensus for contract call, that's why i've set it to true here. Do you think we should rename the variable?

packages/server/tests/acceptance/erc20.spec.ts
georgi-l95: Hmm, i've copied this comment around the code, but basically we are setting a 5 second sleep here, nothing more. I've noticed that now it needs around 4-5 seconds for the changes to reach web3 module. Earlier we were asking the consensus node, but now that we are getting this from the mirror node it has a delay.

packages/server/tests/acceptance/erc20.spec.ts
georgi-l95: I've corrected the comment here and in other places

packages/relay/src/lib/errors/MirrorNodeClientError.ts
```diff
@@ -21,30 +21,49 @@
...
     static statusCodes = {
       NOT_FOUND: 404
     };
 
-    constructor(message: string, statusCode: number) {
-      super(message);
-      this.statusCode = statusCode;
+    constructor(error: any, statusCode: number) {
+        if (error.response?.data?._status?.messages?.length) {
+            const msg = error.response.data._status.messages[0];
```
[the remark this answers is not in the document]
georgi-l95: Added.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -25,6 +25,7 @@ import constants from './../constants';
 import { Histogram, Registry } from 'prom-client';
 import { formatRequestIdMessage } from '../../formatters';
 import axiosRetry from 'axios-retry';
+import {predefined} from "../errors/JsonRpcError";
```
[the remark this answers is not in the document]
georgi-l95: Fixed.

packages/relay/src/lib/clients/mirrorNodeClient.ts
georgi-l95: Added a comment.

packages/relay/src/lib/errors/MirrorNodeClientError.ts
georgi-l95: Hm, yes, i've added a separate error for this.

packages/relay/src/lib/errors/MirrorNodeClientError.ts
```diff
@@ -21,30 +21,54 @@
...
 
     static statusCodes = {
       NOT_FOUND: 404
     };
 
-    constructor(message: string, statusCode: number) {
-      super(message);
-      this.statusCode = statusCode;
+    constructor(error: any, statusCode: number) {
+        //web3 module sends errors in this format, this is why we need a check to distinguish
```
Nana-EC, line 39: nit: format
```suggestion
        // web3 module sends errors in this format, this is why we need a check to distinguish
```

packages/relay/src/lib/errors/MirrorNodeClientError.ts
Nana-EC: Gotcha, i thought maybe they were connected but this works for now

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -238,7 +239,15 @@ export class MirrorNodeClient {
         }
 
         this.logger.error(new Error(error.message), `${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`);
-        throw new MirrorNodeClientError(error.message, effectiveStatusCode);
+
+        const mirrorError = new MirrorNodeClientError(error, effectiveStatusCode);
+
+        // we only need contract revert errors here as it's not the same as not supported
+        if (mirrorError.isContractReverted() && !mirrorError.isNotSupported() && !mirrorError.isNotSupportedSystemContractOperaton()) {
```
Nana-EC, line 246: I thought maybe it would end up looking like this but this is okay for now.
Maybe revisit how their tied together later
```suggestion
        if (mirrorError.isContractReverted() && !mirrorError.isNotSupportedSystemContractOperaton()) {
```

packages/server/tests/contracts/HederaTokenServiceImpl.sol
```diff
@@ -0,0 +1,117 @@
...
+
+    function getTokenKeyPublic(address token, uint256 keyType) public returns (IHederaTokenService.KeyValue memory)
+    {
+        (int256 responseCode,IHederaTokenService.KeyValue memory key) = HederaTokenService.getTokenKey(token, keyType);
+        if (responseCode != HederaResponseCodes.SUCCESS) {
+            revert();
+        }
+        return key;
+    }
+}
```
Nana-EC, line 117: nit: add new line after

Nana-EC (approved):
LG, some nits and rename suggestion.

packages/relay/src/lib/clients/mirrorNodeClient.ts
georgi-l95: Yes, I think one will be removed after we support precompiles, right ?

helm-chart/values.yaml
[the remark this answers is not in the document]
georgi-l95: Done, changed to `ETH_CALL_DEFAULT_TO_CONSENSUS_NODE`

packages/server/tests/contracts/HederaTokenServiceImpl.sol
georgi-l95: Added.

packages/relay/src/lib/errors/MirrorNodeClientError.ts
georgi-l95: Fixed.

Nana-EC (approved):

PR #1008, opened 2023-03-23
Return existing SubId for repeated subscriptions from same connection and same tag/filters.

### LINKED ISSUE #987: Repeated Subscriptions by same connection
### Problem

A single Client can open a web socket connection and then subscribe several times to the same address and topic, I did 3 equal subscriptions, and when a log is generated, I received the same message from the ws-server 3 times. 

### Solution

maybe we can ignore subsequent requests and return the same subscriptionId when the subscription is the same?

### Alternatives

_No response_

AlfredoG87:
**Description**:
When receiving repeated subscriptions from the same connection, for the same Tag, return the already existing subId instead of a new one.

**Related issue(s)**:

Fixes #987 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/subscriptionController.ts
+++ b/packages/relay/src/lib/subscriptionController.ts
@@ -66,6 +66,13 @@ export class SubscriptionController {
             this.subscriptions[tag] = [];
         }
 
+        // Check if the connection is already subscribed to this event
+        const existingSub = this.subscriptions[tag].find(sub => sub.connection.id === connection.id);
+        if (existingSub) {
+            this.logger.info(`${LOGGER_PREFIX} Connection ${connection.id} already subscribed to ${tag}`);
+            return existingSub.subscriptionId;
+        }
+
         const subId = this.generateId();
 
         this.logger.info(`${LOGGER_PREFIX} New subscription ${subId}, listening for ${tag}`);
--- a/packages/relay/tests/lib/subscriptionController.spec.ts
+++ b/packages/relay/tests/lib/subscriptionController.spec.ts
@@ -221,4 +221,14 @@ describe("subscriptionController", async function() {
         expect(status).to.be.eq(true);
     });
 
+    it('Subscribing to the same event and filters should return the same subscription id', async function () {
+        const wsConnection = new MockWsConnection("7");
+        const tag1 = { event: "logs", filters:{"topics": ["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef"]}};
+        const subId = subscriptionController.subscribe(wsConnection, tag1.event);
+        const subId2 = subscriptionController.subscribe(wsConnection, tag1.event);
+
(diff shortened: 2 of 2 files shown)


Ivo-Yankov (approved):
LGTM

packages/relay/src/lib/subscriptionController.ts
```diff
@@ -66,6 +66,13 @@ export class SubscriptionController {
             this.subscriptions[tag] = [];
         }
 
+        // Check if the connection is already subscribed to this event
+        const existingSub = this.subscriptions[tag].find(sub => sub.connection.id === connection.id);
+        if (existingSub) {
+            this.logger.info(`${LOGGER_PREFIX} Connection ${connection.id} already subscribed to ${tag}`);
```
Nana-EC, line 72: nit: debug log maybe?

packages/relay/tests/lib/subscriptionController.spec.ts
```diff
@@ -141,10 +117,10 @@ describe("subscriptionController", async function() {
 
 
     it('notifySubscribers should notify multiple subscribers with data', async function () {
-        const wsConnection1 = new MockWsConnection("2");
+        const wsConnection1 = new MockWsConnection("12");
```
Nana-EC, line 120: Q: what was the significance of the 12 and 13 change?
Something to comment or turn into a self describing variable?

Nana-EC (commented):
looking good, a nit and a question

packages/relay/tests/lib/subscriptionController.spec.ts
AlfredoG87: This change broke this other UT because we were having 2 connections with the same ID and the same exact filters, so naturally the subscriptionController recognized it as repeated subscription from same connection 😅.
With changing just 1 of 2 the issue was resolved, since the tests are independent from one another, but I like to keep them with numbers different than the others just to play safe.

packages/relay/src/lib/subscriptionController.ts
AlfredoG87: yes you are right! 😅

packages/relay/tests/lib/subscriptionController.spec.ts
Nana-EC: Ah the number is a subscription Id.
Okay we should make that clear by setting the value to a const with an appropriate name in the tests so it's clear.

Nana-EC (dismissed):
LG.
Suggestion on adding a const to make subscription id use in tests clear

packages/relay/tests/lib/subscriptionController.spec.ts
AlfredoG87: actually is the connection Id. but yeah, you are right... it would be nice.

Nana-EC (approved):
LG

georgi-l95 (approved):
LGTM

PR #1009, opened 2023-03-23
936 add a max block number range check

### LINKED ISSUE #936: Add a max block number range check
### Problem

```
    err: {
      "stack":
      "statusCode": 504
    }
    code: -32603
```

### Solution

### Alternatives

_No response_

ebadiere:
**Description**:

This PR modifies the block query to the mirror node(eth_getBlockByNumber) by checking if the block number is greater than three above the latest block number.  If so, it returns null immediately.

Fixes #936

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -80,7 +80,7 @@ export class EthImpl implements Eth {
   static blockLatest = 'latest';
   static blockEarliest = 'earliest';
   static blockPending = 'pending';
-  static fewBlocksGreater = 3;
+  static fewBlocksGreater = 5;
 
   /**
    * Configurable options used when initializing the cache.
@@ -1335,8 +1335,6 @@ export class EthImpl implements Eth {
       return null;
     }
 
-    // If blocknumber requested is more than a few blocks greater than the latest, fail fast. 
-    // const latestBlock = await this.mirrorNodeClient.getLatestBlock(requestId);
     if(!returnLatest){
       const response = await this.mirrorNodeClient.getLatestBlock(requestId);
       blockResponse = response.blocks[0];
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -80,6 +80,9 @@ export class EthImpl implements Eth {
   static blockLatest = 'latest';
   static blockEarliest = 'earliest';
   static blockPending = 'pending';
+  static fewBlocksGreater = 3;
+
+
 
   /**
    * Configurable options used when initializing the cache.
@@ -1230,7 +1233,7 @@ export class EthImpl implements Eth {
 
   private static blockTagIsLatestOrPending = (tag) => {
     return tag == null || tag === EthImpl.blockLatest || tag === EthImpl.blockPending;
-  }
+  };
 
   /**
    * Translates a block tag into a number. 'latest', 'pending', and null are the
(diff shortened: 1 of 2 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1334,10 +1335,21 @@ export class EthImpl implements Eth {
       return null;
     }
 
+    // If blocknumber requested is more than a few blocks greater than the latest, fail fast. 
+    // const latestBlock = await this.mirrorNodeClient.getLatestBlock(requestId);
```
AlfredoG87, line 1339: we could remove this comment

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -842,6 +868,10 @@ describe('Eth calls using MirrorNode', async function () {
     expect(result.number).equal(blockNumberHex);
   });
 
+  it('eth_getBlockByNumber should fail fast when a block number greater than current block number from the mirror node', async function () {
+    console.log('Test to fail');
```
AlfredoG87, line 872: This test case is incomplete?

packages/relay/src/lib/eth.ts
```diff
@@ -1334,10 +1335,21 @@ export class EthImpl implements Eth {
...
+    if(!returnLatest){
+      const response = await this.mirrorNodeClient.getLatestBlock(requestId);
```
AlfredoG87, line 1341: is this returnLatest correct? I mean returnLatest = true, it wont enter this condition, and the call from getBlock is always sending true. thus never entering this condition. not sure to understand expected logic.

AlfredoG87 (changes requested):

packages/relay/src/lib/eth.ts
```diff
@@ -80,6 +80,7 @@ export class EthImpl implements Eth {
   static blockLatest = 'latest';
   static blockEarliest = 'earliest';
   static blockPending = 'pending';
+  static fewBlocksGreater = 3;
```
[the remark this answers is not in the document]
Nana-EC: I was thinking a similar thing on the dynamic property aspect.
Each blocks at top performance is 2 secs so this gives us a 6 second window which is not unreasonable.
So far as the error also shows up in metrics as something we can react to I think we should be good to go with a value of `fewBlocksGreater = 5` to start.
This gives as 10 secs which is the worst E2E for transaction submission to query request path we might support.

packages/relay/src/lib/eth.ts
Nana-EC: Seems like `returnLatest` implies the blockTag is 'latest' in which case this flow will correctly ignore that consideration.
However where it's not i.e. `!returnLatest` it implies a blockTag had a value for a specific block to searhc for. In that case we won't to check it's within bounds and fail fast if it's not.
Is that correct @ebadiere 
If so this seems like a nice piece of logic to pull out into a method with method comments maybe `isBlockWithinBounds(blockNumberOrTag, requestId)`
Then this code become
```
if (!isBlockWithinBounds()) {
  return null;
}
```

Nana-EC (changes requested):
Nice.
Let's add a dynamic property and make the fail fast more readable.

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
ebadiere: Updated to 5, but I agree that making it a configurable property does seem to make sense.  I started with 3 because usually few is understood to be about 3.

packages/relay/src/lib/eth.ts
ebadiere: Done.

packages/relay/src/lib/eth.ts
Nana-EC: @ebadiere is this correct?

ebadiere:
Replaced by PR #1262

PR #1016, opened 2023-03-25
Enable `eth_getStorageAt` to return older state

### LINKED ISSUE #790: Support older contract state in `eth_getStorageAt`
With the refactoring performed in this [PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/783) we now support:
- Current state of a contract
- Previous state of a contract a contract, if we know the exact block number this state change has been made.

We are missing support of: 
- Previous state of a contract, without specifying the exact block number this change occurred. For example we pass a block number and return the state of the contract prior this block number.

This might require some changes to the mirror-node endpoint, for example support of timestamp.
Only way to do it directly in the relay is to search every block before the one we pass as parameter in search for this change. This is highly unoptimised and not a solution.

georgi-l95:
**Description**:
With the extension of state endpoint in the mirror-node, we can now fetch and return not only the current state, but older. This PR makes the necessary changes to enable this and removes the usage of the old endpoint as it is no longer needed.
Bump mirror-node version, because the needed change for this feature is in 0.77

**Related issue(s)**:

Fixes #790 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -513,10 +513,13 @@ export class MirrorNodeClient {
         return this.getContractResultsByAddress(address, contractResultsParams, limitOrderParams, requestId);
     }
 
-    public async getContractCurrentStateByAddressAndSlot(address: string, slot: string, requestId?: string) {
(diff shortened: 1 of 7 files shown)


Nana-EC (approved):
Woow lots or removed code. I like :)

PR #1011, opened 2023-03-24
TTL for WS Connections

AlfredoG87:
**Description**:
Adding TTL to websocket connections to avoid keeping connections for too long.

**Related issue(s)**:

Fixes #986 

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #1017, opened 2023-03-26
429-Response-are-not-being-closed-on-server: Decrements the counter a…

### LINKED ISSUE #1014: Connections Rejected with 429 Response are not being closed on the server
### Description

Connections Rejected with 429 Response are not being closed on the server, they stay open indefinitively or until they reach their TTL.

### Steps to reproduce

1. Create 10 connections using postman
2. Create an 11 connection using postman
3. 11 connection will fail with error code 429
4. Click 10 times on Connect, and all 10 times, connection will be rejected with code 429.
5. Close original 10 connections
6. Try to connect again, you cannot, still getting error code 429

### Additional context

_No response_

### Hedera network

other

### Version

unreleased

### Operating system

None

ebadiere:
Decrements connection counter and closes the websocket connection.

**Description**:
This PR modifies fixes the connection closing issue.  
* Decrements the connections counter and closes the websocket connection on the close.
* Leaves the verifyClient validation in.

Fixes #1014
- [* ] Tested (unit, integration, etc.). Enhanced the subscribe.spec.ts to cover the additional connections scenarios.


AlfredoG87 (dismissed):
Already tested it and the fix works as expected 👍 Thank you Eric for resolving this so soon.
just one nit.

packages/ws-server/src/webSocketServer.ts
```diff
@@ -120,6 +120,8 @@ app.ws.use((ctx) => {
     ctx.websocket.on('close', function () {
         relay.subs()?.unsubscribe(ctx.websocket);
         console.log('stopping client interval');
+        connectedClients--;
```
Nana-EC, line 123: Not sure if this change originates in this PR or in 1011 but I posed a questions around it here
https://github.com/hashgraph/hedera-json-rpc-relay/pull/1011/files#r1150967792

Nana-EC (changes requested):
Question on connection decrement

packages/ws-server/src/webSocketServer.ts
AlfredoG87: I copied from here, since my tests were failing due to the connections of the previous tests were not being closed...

AlfredoG87: leaving original question from Nana here:

Q: didn't we explore this earlier and find out that connectedClients = ctx.app.server._connections; captures increments and decrements?

If this is not the case we should add a comment saying decrements aren't handled.

Also should the decrement if applicable occur after the termination to avoid inconsistencies if termination doesn't occur.

packages/server/tests/acceptance/ws/subscribe.spec.ts
```diff
@@ -27,7 +27,8 @@ chai.use(solidity);
 import {Utils} from '../../helpers/utils';
 import {AliasAccount} from "../../clients/servicesClient";
 import {predefined} from '../../../../../packages/relay';
-const {ethers} = require('ethers');
+import { ethers } from "ethers";
+// const {ethers} = require('ethers');
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/server/tests/acceptance/ws/subscribe.spec.ts
```diff
@@ -36,8 +36,15 @@ const WS_RELAY_URL = `ws://localhost:${process.env.WEB_SOCKET_PORT}`;
 const establishConnection = async () => {
         const provider = await new ethers.providers.WebSocketProvider(WS_RELAY_URL);
         await provider.send('eth_chainId');
+        return provider;
 }; 
 
+const unsubscribeAndCloseConnections =async (provider: ethers.providers.WebSocketProvider, subId: string) => {
```
Nana-EC, line 42:
```suggestion
const unsubscribeAndCloseConnections = async (provider: ethers.providers.WebSocketProvider, subId: string) => {
```

packages/ws-server/src/webSocketServer.ts
ebadiere: We don't have access to the ctx.app.server._connections in the verifyClient call back, hence the need to track it in a class variable.  I added tests around the number of connections after opening and closing multiple connections.

Nana-EC: Okay please add a comment above the decrement to be clear why we decrement but never increment.

Nana-EC (commented):
Makes sense.
Minor suggestions then good to sign off

Ivo-Yankov (commented):
This fixes the bug, but I'm a bit concerned that currently it's hard to handle the error event on the client side and the error message is not very descriptive.

In this [PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1037) I added a limit on the maximum connections per IP and I used this approach:

```
ctx.websocket.close(IP_LIMIT_ERROR.code, IP_LIMIT_ERROR.message);
```

This way we emit a `close` event and send the `code` and `message` . The event can be handled by the client like so:

```
provider._websocket.on('close', (code, message) => {
    ...
});
```

packages/ws-server/src/webSocketServer.ts
ebadiere: Done.

PR #1023, opened 2023-04-03
Improve and add missing models for mirror node requests

georgi-l95:
**Description**:
Add missing models for all requests from the mirror-node. This drastically improves dev exp, because it lower chances for mistakes, improves project structure and design.
Using interfaces for models which does not require state changes and exist only in typescript during compilation.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/models/account.ts
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { Balance } from "./balance"
+import { Key } from "./key";
+
+export interface Account {
```
Nana-EC, line 24: For Mirror Node models I don't think we should add the dependency.
This will force us to have to maintain the model and changes.
Initially I thought eventually we'd push it back to the Mirror Node team so they expose an npm model package, however, I was reminded we did all the work around OpenAPI spec so that projects could use it for consideration like this.

I think we should explore something like https://github.com/swagger-api/swagger-js/blob/master/docs/usage/http-client-for-oas-operations.md that takes the Mirror Node Open AI Spec and generates model for us.
Then all we have to do is upgrade version to get the new details.
We can explore this or other model generators but we should only depend on the Open API Spec

packages/relay/src/index.ts
```diff
@@ -18,7 +18,7 @@
  *
  */
 
-import { Block, Log, Receipt, Transaction } from './lib/model';
+import { Block, Log, Receipt, Transaction } from './lib/models';
```
Nana-EC, line 21: I'd keep these models (i.e the JSON RPC response ones) and revert the mirror node ones for another PR per my other feedback

Nana-EC: Additionally, long term we'd want to consume the specs under https://github.com/ethereum/execution-apis/tree/main/src/schemas and generate the models this way and no longer needs these files.

Nana-EC (changes requested):
Intent is spot on but I think we should go a different route.
Let's keep the JSON RPC schema response models that you moved out but let's consume the Mirror Node Open API spec instead.
Move the Mirror Node improvements suggestion to a different future ticket that we can address in the future

packages/relay/src/lib/models/account.ts
georgi-l95: Gotcha, I've initially thought about mirror node team exposing a npm model package, but figured that this won't happen anytime soon. Will research the proposed solutions, closing this PR.

AlfredoG87: I've created a ticket issue for tracking this, also added some options that found. 
https://github.com/hashgraph/hedera-json-rpc-relay/issues/1416

PR #1029, opened 2023-04-04
Feature Flag for eth_subscribe multiple contracts support and acceptance tests

### LINKED ISSUE #926: Expand eth_subscribe logic to work with multiple contracts
Pass array of addresses and test out.

AlfredoG87:
**Description**:

- Since the current implementation of eth_subscribe was already accepting multiple contracts I only did the tests to validate it correct working.
- Also added a feature flag to disable the feature, and added tests to make sure it was working as expected and that the error received is the one expected.

**Related issue(s)**: 

Fixes #926

**Notes for reviewer**:

**Checklist**


packages/ws-server/src/webSocketServer.ts
```diff
@@ -99,7 +103,11 @@ app.ws.use((ctx) => {
             let subscriptionId;
 
             if (event === 'logs') {
-                subscriptionId = relay.subs()?.subscribe(ctx.websocket, event, filters);
+                if(!getMultipleAddressesEnabled() && Array.isArray(filters.address) && filters.address.length > 1) {
+                    response = predefined.INVALID_PARAMETER(4, 'Only one address is allowed');
```
Nana-EC, line 107: What's the `4`?

Nana-EC (commented):
LG, suggestions and questions

Ivo-Yankov:
Do we really need a feature flag for this? After this PR is merged and multiple addresses are supported I don't see a reason to ever block this feature.

Nana-EC:
> Collaborator

Controlled rollout and avoiding divided attention on development and testing.
I'd like us to be intentional about rolling it out hence the ask to feature gate

packages/ws-server/src/webSocketServer.ts
georgi-l95: Same question, also avoid using magic numbers if possible.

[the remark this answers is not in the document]
AlfredoG87: According to the Predefined and existing error `INVALID_PARAMETER` the first parameter is the index of the parameter that is `INVALID`, and given this example of the input, seems to me to be the `4` parameter.
```
{
    "jsonrpc":"2.0",
    "id": 2, 
    "method": "eth_subscribe", 
    "params": ["logs",
        {
        "address": ["0x6721d86A8405C8683dA3CAEBCbeb4c690EdAE963","0xb3611238Ecda2CB930d1C1A8d065743bfCbf7452"]
        }
    ]    
}
```

But now that I think about it, is a JSON, and as such, it could have the same parameters in different order and still should work... (also, if we were to count `0`, it would be the third parameter...
I will look up for the index of the "params" parameter and return that instead of the hardcoded `4`

AlfredoG87: According to the Predefined and existing error `INVALID_PARAMETER` the first parameter is the index of the parameter that is `INVALID`, and given this example of the input, seems to me to be the `4` parameter.
```
{
    "jsonrpc":"2.0",
    "id": 2, 
    "method": "eth_subscribe", 
    "params": ["logs",
        {
        "address": ["0x6721d86A8405C8683dA3CAEBCbeb4c690EdAE963","0xb3611238Ecda2CB930d1C1A8d065743bfCbf7452"]
        }
    ]    
}
```

But now that I think about it, is a JSON, and as such, it could have the same parameters in different order and still should work... (also, if we were to count `0`, it would be the third parameter...
I will look up for the index of the "params" parameter and return that instead of the hardcoded `4`

AlfredoG87: Changed my mind about the above comment, when realized that then Predefined error `INVALID_PARAMETER` also supports an string instead of a number for the "index" of the parameter that is invalid so, changed it to something more descriptive and leaved a comment on the code explaining it. `params[1].address`

                    // [1] is the second object in the expected array of parameters that contains the filters (address and topics) for the subscription

packages/server/tests/acceptance/ws/subscribe.spec.ts
```diff
@@ -229,6 +229,88 @@ describe('@web-socket Acceptance Tests', async function() {
...
+            await new Promise(resolve => setTimeout(resolve, 2000)); // wait for event to be received
+            expect("3: " + latestEventFromSubscription.params.result.address).to.be.eq("3: " + logContractSigner3.address.toLowerCase());
+            expect("3: " + latestEventFromSubscription.params.subscription).to.be.eq("3: " + subscriptionId);
+
+            // close the connection
+            webSocket.close();
+
+            // wait for the connections to be closed
+            await new Promise(resolve => setTimeout(resolve, 500));
+            process.env.WS_MULTIPLE_ADDRESSES_ENABLED = originalWsMultipleAddressesEnabledValue; // restore original value
```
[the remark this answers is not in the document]
AlfredoG87: ✅  Done!

packages/ws-server/src/webSocketServer.ts
```diff
@@ -104,7 +104,8 @@ app.ws.use((ctx) => {
 
             if (event === 'logs') {
                 if(!getMultipleAddressesEnabled() && Array.isArray(filters.address) && filters.address.length > 1) {
-                    response = predefined.INVALID_PARAMETER(4, 'Only one address is allowed');
+                    // [1] is the second object in the expected array of parameters that contains the filters (address and topics) for the subscription
+                    response = jsonResp(request.id, predefined.INVALID_PARAMETER("params[1].address", 'Only one contract address is allowed'), undefined);
```
Nana-EC, line 108: nit: use filters instead of `params[1]`
Q: Also do any of these params need validation before use?

Maybe a separate ticket/PR to ensure safety

AlfredoG87: ✅  Changed `params[1]` to `filters` completed 

Also, good catch on parameters sanity check, I have created this ticket to track parameter validations.
https://github.com/hashgraph/hedera-json-rpc-relay/issues/1045

Please add any other validations that you consider necessary besides the ones I have already included.

Ivo-Yankov (approved):

PR #1032, opened 2023-04-06
Test fix for release 0.21.0-rc1.  Updated test to use invalid account

ebadiere:
**Description**:
This PR updates the rcp_batch3.spec, "007 'data' from request body with wrong encoded parameter" to use an invalid address in the "wrong encoded parameter" as the old address on previewnet was getting used accumulating a balance.


lukelee-sl (approved):
LG.  

Nit: Maybe a comment might be useful?

Issue #995, opened 2023-03-17
Add IP based limits to the WS server


Ivo-Yankov:
We should limit the number of connections/subscriptions per IP address.

- [x] Track the number of connections and subscriptions from every IP
- [x] Add a configurable limitation. Default should be 10
- [x] Return a descriptive error message when a user tries to exceed the limitation
- [x] Add tests

PR #1037, opened 2023-04-06
Connection limits per IP for WebSocket server

Ivo-Yankov:
**Description**:
- Adds a configurable limit on the maximum connections from a single IP address.
- Adds a class for WebSocket errors. I suggest we follow the convention described in this [article](https://medium.com/weekly-webtips/the-essential-guide-on-the-web-socket-status-codes-97b54773f66d) regarding the error codes. E.g. our error codes should look like `4XXX`

**Related issue(s)**:

Addresses #995 

**Notes for reviewer**:

**Checklist**


.env.example
```diff
@@ -36,3 +36,4 @@ SERVER_PORT=
 WEB_SOCKET_PORT=
 CONNECTION_LIMIT=
 WS_MAX_CONNECTION_TTL=
+CONNECTION_LIMIT_PER_IP=
```
Nana-EC, line 39: nit: let's differentiate the servers
```suggestion
WS_CONNECTION_LIMIT_PER_IP=
```

packages/server/tests/acceptance/ws/subscribe.spec.ts
```diff
@@ -244,15 +248,76 @@ describe('@web-socket Acceptance Tests', async function() {
...
+            it('Does not allow more connections from the same IP than the specified limit', async function() {
+                const providers = [];
+                let closeEventHandled = false;
+
+                // Creates the maximum allowed connections
+                for (let i = 1; i < parseInt(process.env.CONNECTION_LIMIT_PER_IP); i++) {
+                    providers.push(await new ethers.providers.WebSocketProvider(WS_RELAY_URL));
+                }
+
+                await new Promise(resolve => setTimeout(resolve, 1000));
```
Nana-EC, line 294: I'm weary of waits in test code but not sure how else to achieve this now

packages/server/tests/acceptance/ws/subscribe.spec.ts
```diff
@@ -244,15 +248,76 @@ describe('@web-socket Acceptance Tests', async function() {
...
+
+                // Repeat the following several times to make sure the internal counters are consistently correct
+                for (let i = 0; i < 3; i++) {
+                    expect(server._connections).to.equal(parseInt(process.env.CONNECTION_LIMIT_PER_IP));
+
+                    // The next connection should be closed by the server
+                    const provider = await new ethers.providers.WebSocketProvider(WS_RELAY_URL);
+
+                    provider._websocket.on('close', (code, message) => {
+                        closeEventHandled = true;
```
Nana-EC, line 304: `closeEventHandled` is never reset to false.
How would a failure when `i > 0` occur?

packages/ws-server/src/webSocketServer.ts
```diff
@@ -63,10 +68,53 @@ function getMaxConnectionTTL() {
     return parseInt(process.env.WS_MAX_CONNECTION_TTL || '300000');
 }
 
-app.ws.use((ctx) => {
+// function to get maximum allowed connections from a single IP
+function getLimitPerIp() {
+    return parseInt(process.env.CONNECTION_LIMIT_PER_IP || '10');
```
Nana-EC, line 73: This should be from a constant so you're not looking up and converting each time.
If you're not exposing publicly not sure that a getter is needed

packages/ws-server/src/webSocketServer.ts
```diff
@@ -63,10 +68,53 @@ function getMaxConnectionTTL() {
...
+}
+
+async function handleConnectionClose(ctx) {
+    if (ctx.websocket.ipCounted) {
+        const {ip} = ctx.request;
+        clientIps[ip]--;
+        if (clientIps[ip] === 0) delete clientIps[ip];
+    }
+    connectedClients--;
```
Nana-EC, line 82: Q: should the decrement be moved to after the actual unsubscribe is done?
What happens if unsubscribe has a failure, does it still unsubscribe but report the failure or could there be a case where the subscription or connection exists but we decremented?

packages/ws-server/src/webSocketServer.ts
```diff
@@ -63,10 +68,53 @@ function getMaxConnectionTTL() {
...
+        clientIps[ip]++;
+    }
+    ctx.websocket.ipCounted = true;
+
+    // Limit checks
+
+    // Limit connections from a single IP address
+    if (clientIps[ip] && clientIps[ip] > getLimitPerIp()) {
+        logger.info(`Maximum allowed connections from a single IP (${clientIps[ip]}) exceeded for address ${ip}`);
+        ctx.websocket.close(IP_LIMIT_ERROR.code, IP_LIMIT_ERROR.message);
```
Nana-EC, line 114: we should have metrics when this occurs

packages/ws-server/src/webSocketServer.ts
```diff
@@ -63,10 +68,53 @@ function getMaxConnectionTTL() {
...
 }
 
-app.ws.use((ctx) => {
+// function to get maximum allowed connections from a single IP
+function getLimitPerIp() {
+    return parseInt(process.env.CONNECTION_LIMIT_PER_IP || '10');
+}
+
+async function handleConnectionClose(ctx) {
+    if (ctx.websocket.ipCounted) {
```
Nana-EC, line 77: Q: Should we use a similar approach tot he relay where we pulled out the rate limiting logic into a separate class instead of lumping it in the main server class?
I think this would be cleaner and easier to maintain.

Nana-EC (commented):
Looking good, some questions

packages/ws-server/src/webSocketServer.ts
```diff
@@ -42,27 +44,57 @@ const logger = mainLogger.child({ name: 'rpc-server' });
...
+            return done(false, 429, 'Connection limit exceeded');
         }
+        done(true);
+    }
 });
 
 const LOGGER_PREFIX = 'WebSocket:';
 
 const CHAIN_ID = relay.eth().chainId();
+const IP_LIMIT_ERROR = WebSocketError.CONNECTION_IP_LIMIT_EXCEEDED;
```
ebadiere, line 62: I agree that this approach is cleaner for handling the connection limits.  I would like to wait until this is merged into main so I can pick it up in this [PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1017) and use it.

packages/ws-server/src/webSocketServer.ts
Ivo-Yankov: Good point, `unsubscribe` should be executed before all decrements.

packages/ws-server/src/webSocketServer.ts
Ivo-Yankov: There is an issue for adding metrics to the ws-server, with a bullet point of tracking the count of active connections by IP

Nana-EC (approved):

packages/ws-server/src/ConnectionLimiter.ts
```diff
@@ -0,0 +1,97 @@
...
+            if (this.clientIps[ip] === 0) delete this.clientIps[ip];
+        }
+        this.connectedClients--;
+    }
+
+    public applyLimits(ctx) {
+        const {ip} = ctx.request;
+
+        // Limit connections from a single IP address
+        const limitPerIp = parseInt(process.env.WS_CONNECTION_LIMIT_PER_IP || '10');
```
Nana-EC, line 70: nit: still think this should be a constant at the class level. Maybe even on instantiation of this file

Ivo-Yankov: I did it this way so that it's possible to change the value after initialization - this is used in the tests. I know it's not the best solution, but the alternative is to add setter/getter methods and I think that would add unnecessary complexity.

PR #1050, opened 2023-04-12
RPC API Batch 1 - Acceptance Test - Improvement

### LINKED ISSUE #1049: RPC API Batch 1 Test is intermittently failing on CI/CD
### Problem

image

### Solution

- Notice that the `fromBlock` parameter has a "**negative**" value, in this case is "0x-9" but it varies from test to test.
- Probably since blocks are generated every `X` amount of time, is that this error is intermittent, so when prior tests take longer to complete, the test does not fail. :/
- Change the Test parameters so it does not try to get logs from that far away.

### Alternatives

_No response_

AlfredoG87:
**Description**:
changed the hardcoded value of "blocksBehindLatest" from 40 to 10, since is enough for the purpose of the test and gives higher chances that even if the tests and the hardware used to run the tests improves in the future, the test will continue to work.

Update:
Only if currentBlock is bigger than 10, we fetch past 10 blocks, otherwise we fetch everything since the block 0.

**Related issue(s)**:

Fixes #1049 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance/rpc_batch1.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch1.spec.ts
@@ -286,7 +286,7 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
                 const currentBlock = Number(await relay.call('eth_blockNumber', [], requestId));
                 let blocksBehindLatest = 0;
                 if(currentBlock > 10) {
-                    blocksBehindLatest = Number(await relay.call('eth_blockNumber', [], requestId)) - 10;
+                    blocksBehindLatest = currentBlock - 10;
                 }
                 const logs = await relay.call('eth_getLogs', [{
                     'fromBlock': EthImpl.numberTo0x(blocksBehindLatest),


packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -281,7 +281,8 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
                 //for the purpose of the test, we are settings limit to 2, and fetching all.
                 //setting mirror node limit to 2 for this test only
                 process.env['MIRROR_NODE_LIMIT_PARAM'] = '2';
-                const blocksBehindLatest = Number(await relay.call('eth_blockNumber', [], requestId)) - 40;
+                // 10 is an arbitrary number and might be changed in the future for a smaller number in case test fails due to negative number in the fromBlock param
+                const blocksBehindLatest = Number(await relay.call('eth_blockNumber', [], requestId)) - 10;
```
Nana-EC, line 285: Is it worth making sure the block count is always above this arbitrary number to avoid a negative number as you noted?
```suggestion
                const blockOffset = 10;
                const latestBlockNum = await relay.call('eth_blockNumber', [], requestId))
                const minBlockNum = min(latestBlockNum, blockOffset); // not sure if a further offset of 1 is needed
                const blocksBehindLatest = minBlockNum - blockOffset;
```

AlfredoG87: I've changed it to check if CurrentBlock is bigger than 10, otherwise fetch from block 0 (the beginning).

Nana-EC (approved):

ebadiere (commented):
LGTM.

PR #1063, opened 2023-04-14
configurable ttl Cache for current blockNumber

### LINKED ISSUE #1057: Add cache support to eth_blockNumber
### Problem

Currently the `eth_blockNumber` queries the Mirror Node for the latest block.
Under high load a mirror node may limit RPS

### Solution

To reduce the number of calls needed and therefore the load on a mirror node a cache may be added to `eth_blockNumber` endpoint.

Mirror Node processes a block every 2 secs, so a cache of 500 or 1000ms may be a fair balance between optimization and data freshness.
This should be added as a configureable property

### Alternatives

_No response_

AlfredoG87:
**Description**:
Adding a configurable TTL cache for current blockNumber to eth.ts

**Related issue(s)**:

Fixes #1057

**Notes for reviewer**:

**Checklist**


docs/configuration.md
```diff
@@ -42,27 +42,28 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
-| `SDK_REQUEST_TIMEOUT`             | "10000"       | The complete timeout for running the SDK `execute()` method. This controls the GRPC channel timeout config when querying with network nodes.   |
-| `TIER_1_RATE_LIMIT`               | "100"         | Maximum restrictive request count limit used for expensive endpoints rate limiting.   |
-| `TIER_2_RATE_LIMIT`               | "800"         | Maximum moderate request count limit used for non expensive endpoints.   |
-| `TIER_3_RATE_LIMIT`               | "1600"        | Maximum relaxed request count limit used for static return endpoints.   |
+| Name                                  | Default       | Description                                                                                                                                                                                                                                                                                            |
```
david-bakin-sl, line 45: This is annoying.  This happened to me too.  Totally legit markdown table formatting made "nicer" - except not really! - by something - spotless perhaps?  Totally gratuitous! Totally unnecessary!

k6/.envexample
```diff
@@ -10,3 +10,4 @@ DEFAULT_GRACEFUL_STOP=
 FILTER_TEST=
 DEBUG_MODE=
 SIGNED_TXS=
+ETH_BLOCK_NUMBER_CACHE_TTL_MS=
```
david-bakin-sl, line 13: Excellent to put units at end! (`MS`)

david-bakin-sl (commented):
(Only a comment because still in draft. But looks good so far.)

docs/configuration.md
AlfredoG87: yeah, I was in conflict if was better to let the IDE do its thing and format the whole table, I think is nice the end result... but is completely useless for code review 😓 

If we get a +1 on this I might update PR to update only 1 line (addition) and leave the rest of the table as it was...

docs/configuration.md
```diff
@@ -42,27 +42,28 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
+| `DEFAULT_RATE_LIMIT`                  | "200"         | default fallback rate limit, if no other is configured.                                                                                                                                                                                                                                                |
+| `ETH_CALL_CACHE_TTL`                  | "200"         | Maximum time in ms to cache an eth_call response.                                                                                                                                                                                                                                                      |
+| `ETH_BLOCK_NUMBER_CACHE_TTL_MS`       | "1000"        | Time in ms to cache response from mirror node                                                                                                                                                                                                                                                          |
```
lukelee-sl, line 51: I'm wondering out loud what the appropriate number should be for this cache expiration.  I believe our blocks are created at 2 second intervals which means at 1 second ( default value I understand ) we could be missing out on the latest block for a whole second?

AlfredoG87: Yes, and that was why I agree that it must be configurable. Took the biggest alternative (in the [ticket](https://github.com/hashgraph/hedera-json-rpc-relay/issues/1057)) order to see if there are complaints, if so we can lower it to 500ms, but if not, 1000 would have the biggest (positive) impact on RPS limits.

lukelee-sl (approved):
LGTM

k6/.envexample
Nana-EC, line 13: nit: new line in next PR

Nana-EC (approved):

k6/.envexample
david-bakin-sl: there's an intellij setting somewhere for this ...

editor/general/on save/ensure every saved file ends with a line break

PR #1067, opened 2023-04-14
Added the abort controller to cancel requests.

### LINKED ISSUE #1065: Terminate the axios calls when a timeout is triggered calling the mirrornode
### Description

Connections may be held when timeouts occur when querying the mirrornode.  These connections should be terminated.

### Steps to reproduce

Running eth_feeHistory with the timeout set to a low number will trigger the timeout.

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.21.0

### Operating system

None

ebadiere:
**Description**:
Uses the axios controller abort to cancel requests on timeouts.

Fixes #1065

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -235,12 +235,18 @@ export class MirrorNodeClient {
     }
 
     handleError(error: any, path: string, effectiveStatusCode: number, method: REQUEST_METHODS, allowedErrorStatuses?: number[], requestId?: string) {
+        const mirrorError = new MirrorNodeClientError(error, effectiveStatusCode);
+        if(mirrorError.isTimeout()){
+            controller.abort();
+            return null;
+        }
+
         const requestIdPrefix = formatRequestIdMessage(requestId);
         if (allowedErrorStatuses && allowedErrorStatuses.length) {
-            if (error.message === `timeout of ${parseInt(process.env.MIRROR_NODE_TIMEOUT || '10000')}ms exceeded`) {
-                controller.abort();
-                return null;
-            }
+            // if (error.message === `timeout of ${parseInt(process.env.MIRROR_NODE_TIMEOUT || '10000')}ms exceeded`) {
+            //     controller.abort();
+            //     return null;
+            // }
             if (error.response && allowedErrorStatuses.indexOf(effectiveStatusCode) !== -1) {
                 this.logger.debug(`${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`);
                 return null;
@@ -248,8 +254,6 @@ export class MirrorNodeClient {
         }
 
         this.logger.error(new Error(error.message), `${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`);
-
-        const mirrorError = new MirrorNodeClientError(error, effectiveStatusCode);
 
         // we only need contract revert errors here as it's not the same as not supported
         if (mirrorError.isContractReverted() && !mirrorError.isNotSupported() && !mirrorError.isNotSupportedSystemContractOperaton()) {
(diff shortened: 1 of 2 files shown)


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -233,6 +237,10 @@ export class MirrorNodeClient {
     handleError(error: any, path: string, effectiveStatusCode: number, method: REQUEST_METHODS, allowedErrorStatuses?: number[], requestId?: string) {
         const requestIdPrefix = formatRequestIdMessage(requestId);
         if (allowedErrorStatuses && allowedErrorStatuses.length) {
+            if (error.message === `timeout of ${parseInt(process.env.MIRROR_NODE_TIMEOUT || '10000')}ms exceeded`) {
```
Nana-EC, line 240: I don't think the placement is right.
Aren't we not making it in here right now?

Would it safer to to it outside of the if
```
        if (effectiveStatusCode === MirrorNodeClientError.ErrorCodes.ECONNABORTED) {
            controller.abort();
        }
```

You could do this at the beginning of the handleError

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -118,7 +121,8 @@ export class MirrorNodeClient {
             headers: {
                 'Content-Type': 'application/json'
             },
-            timeout: 10 * 1000
+            signal: controller.signal,
```
Nana-EC, line 124: Should we be doing this per client or per request?
As in here in the create or in the get?

Nana-EC (changes requested):
Some clarifications.
Also did we figure out a way to count connections to help confirm?

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -1681,35 +1681,6 @@ describe('Eth calls using MirrorNode', async function () {
       expectLogData(res, defaultLogs.logs[3], defaultDetailedContractResults3);
     };
 
-    it('blockHash filter timeouts and throws the expected error', async function () {
```
Nana-EC, line 1684: We should still throw a timeout error.
If not we're not handling the timeout correctly.
What happens instead now?

lukelee-sl:
> Some clarifications. Also did we figure out a way to count connections to help confirm?

My reading is the as of node 0.12 max socket connections is set to infinite https://nodejs.org/en/blog/release/v0.12.0.  We might want to limit the number of connection used by the default globalAgent by setting `http.globalAgent.maxSockets = 100` or some number like that.

I believe Axios will use the connection pool of the global agent unless you give it a specific one to run with

Nana-EC:
> > Some clarifications. Also did we figure out a way to count connections to help confirm?
> 
> My reading is that as of node 0.12 max socket connections is set to infinite https://nodejs.org/en/blog/release/v0.12.0. We might want to limit the number of connection used by the default globalAgent by setting `http.globalAgent.maxSockets = 100` or some number like that.
> 
> I believe Axios will use the connection pool of the global agent unless you give it a specific one to run with

Nice catch. This might be what's happening

packages/relay/src/lib/clients/mirrorNodeClient.ts
ebadiere: Done.

packages/relay/src/lib/clients/mirrorNodeClient.ts
ebadiere: Refactored to implement in the config defined in the request method.

packages/relay/tests/lib/eth.spec.ts
ebadiere: Test restored.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -195,17 +198,20 @@ export class MirrorNodeClient {
         let ms;
         try {
             let response;
-            const headers = {
+
+            controller = new AbortController();
```
Nana-EC, line 202: Q: Checking my js node knowledge here but will concurrent runs ever result in 1 request that was aborted affecting a new request ?
If so would having the `controller1 variable be scoped to the function and not the class be safer?

Nana-EC (commented):
Looking good. 
1 question.

packages/relay/src/lib/clients/mirrorNodeClient.ts
ebadiere: Done

Nana-EC (approved):
LG

lukelee-sl (approved):

PR #1075, opened 2023-04-15
RPC API Batch 1 - Acceptance Test - Improvement (#1050)

Nana-EC:
---------

**Description**:
Cherry pick #1050 
* changed the hardcoded value of "blocksBehindLatest" from 40 to 10, since is enough for the purpose of the test and gives higher chances that even if the tests and the hardware used to run the tests improves in the future, the test will continue to work.

* adding comments to change

* further improve test to fetch from beginning in case currentBlock is less than 10

* avoid double call of eth_blockNumber

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


lukelee-sl (approved):

david-bakin-sl (approved):
LGTM. (clever workaround)

PR #1086, opened 2023-04-16
Optimize eth_getBalance queries

### LINKED ISSUE #1085: Optimize eth_getBalance for Mirror Node 0.78
### Problem

### Solution

### Alternatives

_No response_

Nana-EC:
**Description**:
`eth_getBalance` logic around queries is outdated and in some cases we're making calls that don't need to be called

New logic should be
- check for a cache balance entry. Return if exists but do not cache as this is for high load
- consider blockNumberOrTag
  - if valid blockNumberOrTag get the account balance from `api/v1/balances` which is now real time. Note, this endpoint only takes a Hedera AccountId so a conversion from address is needed. For long zero we can convert, for evmAddress we must consult Mirror Node for mapping
  - if no blockNumberOrTag call the `api/v1/accounts` endpoint for most recent balance. Not 404 may occur for recent creations so retry should be covered in client
- cache balance and return

To achieve the above
- Add a `getAccountIdFromAddress()` to the eth.ts to isolate address to HederaAccountId logic.
- Update `getBalance()` to utilize above flow and method
- Add a `convertLongZeroAddressToHederaAccountId()` to the formatter.ts

**Related issue(s)**:

Fixes #1085

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -19,7 +19,6 @@
  */
 
 import { AccountId } from '@hashgraph/sdk';
-// import ContractId from "@hashgraph/sdk/lib/contract/ContractId";
 import constants from "./lib/constants";
 
 const hashNumber = (num) => {
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -18,8 +18,8 @@
  *
  */
 
-import AccountId from "@hashgraph/sdk/lib/account/AccountId";
-import ContractId from "@hashgraph/sdk/lib/contract/ContractId";
+import { AccountId } from '@hashgraph/sdk';
+// import ContractId from "@hashgraph/sdk/lib/contract/ContractId";
 import constants from "./lib/constants";
 
 const hashNumber = (num) => {
@@ -75,4 +75,4 @@ const convertLongZeroAddressToHederaAccountId = (evmAddress: string): string | n
     return AccountId.fromSolidityAddress(evmAddress).toString();
 }
 
-export { hashNumber, formatRequestIdMessage, hexToASCII, decodeErrorMessage, formatTransactionId, convertLongZeroAddressToHederaAccountId };
+export { hashNumber, formatRequestIdMessage, hexToASCII, decodeErrorMessage, formatTransactionId, convertLongZeroAddressToHederaAccountId, AccountId };
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -18,6 +18,8 @@
  *
  */
 
+import AccountId from "@hashgraph/sdk/lib/account/AccountId";
+import ContractId from "@hashgraph/sdk/lib/contract/ContractId";
 import constants from "./lib/constants";
 
 const hashNumber = (num) => {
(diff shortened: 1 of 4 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -580,21 +582,20 @@ export class EthImpl implements Eth {
...
-      return cachedBalance;
-    }
-
     let blockNumber = null;
     let balanceFound = false;
     let weibars: BigInt = BigInt(0);
     const mirrorAccount = await this.mirrorNodeClient.getAccount(account, requestId);
 
     try {
+      // check balance cache first for high load scenarios
```
Nana-EC, line 591: Move cache check to the top after `getBalance(account=` log

packages/relay/src/lib/eth.ts
```diff
@@ -580,21 +582,20 @@ export class EthImpl implements Eth {
...
+      const cacheKey = `${constants.CACHE_KEY.ETH_GET_BALANCE}-${account}-${blockNumberOrTag}`;
+      const cachedBalance = this.cache.get(cacheKey);
+      if (cachedBalance) {
+        this.logger.trace(`${requestIdPrefix} returning cached value ${cacheKey}:${JSON.stringify(cachedBalance)}`);
+        return cachedBalance;
+      }
+
       if (!EthImpl.blockTagIsLatestOrPending(blockNumberOrTag)) {
         const block = await this.getHistoricalBlockResponse(blockNumberOrTag, true, requestId);
```
Nana-EC, line 600: This is a dupe of the blockNumber() call before and will go get the new latest except it doesn't consider a cache.
Additionally there's some buggy logic in there for earliest tag also.
If we're going to restore the logic for historical calculations we should still do a perf check and optimize it

packages/relay/src/lib/eth.ts
```diff
@@ -1542,6 +1543,30 @@ export class EthImpl implements Eth {
     return logs;
   }
 
+  private async getAccountIdFromAddress(address: string, requestIdPrefix?: string): Promise<string | null> {
```
Nana-EC, line 1546: Doesn't seem with the revert that this is used anymore

Nana-EC (commented):
Let's revisit the approach here. I'm not sure the current form after revert has all the performance gains we could have.

packages/relay/src/formatters.ts
```diff
@@ -18,6 +18,8 @@
  *
  */
 
+import { AccountId } from '@hashgraph/sdk';
+// import ContractId from "@hashgraph/sdk/lib/contract/ContractId";
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/relay/src/lib/eth.ts
ebadiere: Done.

packages/relay/src/lib/eth.ts
ebadiere: Removed.

packages/relay/src/lib/eth.ts
ebadiere: Moved the cache up and removed the dup so the cache is checked.

Nana-EC: 1. line 600 - this is the first call that will call the blocks API with `blockNumberOrTag`
2. line 606 - we call block for latest just so we can check if it's not the same as `blockNumberOrTag`. I forget why we need the latest at all. Don't we just want to know the balance at `blockNumberOrTag`. One consideration is if `blockNumberOrTag` is not latest or pending or earliest then we know its a valid number. In which case we should just use the block timestamp and call the balances API for the account and get that timestamp. Use that timestamp and compare it's block number to `blockNumberOrTag`. 

If the diff is over 2 sec and under 15 mins then do the calculation of balance by getting the crypto transfers between balance timestamp and the block time. Otherwise return that balance

Nana-EC: Current thinking but let's see

Nana-EC (commented):
Current thoughts

packages/relay/src/lib/eth.ts
ebadiere: Agreed and as discussed, this will be address in a following issue and PR.

Nana-EC:
further options will be opened with https://github.com/hashgraph/hedera-mirror-node/issues/5863

PR #1088, opened 2023-04-17
Optimize eth_feeHistory API calls

### LINKED ISSUE #1087: Optimize eth_feeHistory
### Problem

CUrrently the eth_feeHistory calls for, blocks more often than it should

### Solution

Update the `eth_feeHistory` to utilize timestamps more so than block numbers for reuse and to consider whether the call to latest is necessary or not. Also consider caching improvements.

There 3 bocks always needed, newestBlock, and newestBlock - blockCount. If newestBlock is latestBlock, then we don't need to call getBlock
1. get blockNumber by checking if newestBlock is latestBlock. If not convert hex to number
2. call getBlock with newestBlock
3. get newestBlock - blockCount block
4. (optional) call getBlock with 'latest'
5. pass timestamps (start, end and latest) to getFeeHistory.
6. Update getFeeHistory to use this.getFeeByTimestamp

### Alternatives

_No response_

Nana-EC:
**Description**:
Optimize eth_feeHistory API calls

Restructure logic to pull out perf savings where network fees and blocks are in the optimal range

- Add `getBlocksInTimestampRange()` to mirrorNodeClient to allow getting a list of blocks instead of calling multiple times
- Add `CACHE_TTL.ONE_DAY` for use with network fees
- Add `getAndCacheLatestNetworkFeeChangeBlock()` to `eth.ts` to allow for networkFee and block retrieval on first call
- Add `getFeeByTimestamp()` to allow for fee retrieval without having to call for block again
- Add `getRepeatedFeeHistory()` to create a feeHistory object that is full of repeated cached values

**Related issue(s)**:

Fixes #1087 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -177,29 +177,58 @@ export class EthImpl implements Eth {
     this.logger.trace(`${requestIdPrefix} feeHistory(blockCount=${blockCount}, newestBlock=${newestBlock}, rewardPercentiles=${rewardPercentiles})`);
 
     try {
-      const latestBlockNumber = await this.translateBlockTag(EthImpl.blockLatest, requestId);
-      const newestBlockNumber = (newestBlock == EthImpl.blockLatest || newestBlock == EthImpl.blockPending)
-        ? latestBlockNumber
-        : await this.translateBlockTag(newestBlock, requestId);
-
-      if (newestBlockNumber > latestBlockNumber) {
-        return predefined.REQUEST_BEYOND_HEAD_BLOCK(newestBlockNumber, latestBlockNumber);
-      }
-
-      blockCount = blockCount > maxResults ? maxResults : blockCount;
-
-      if (blockCount <= 0) {
+      const resolvedBlockCount = blockCount > maxResults ? maxResults : blockCount;
+      if (resolvedBlockCount <= 0) {
         return EthImpl.feeHistoryZeroBlockCountResponse;
       }
 
       const cacheKey = `${constants.CACHE_KEY.FEE_HISTORY}_${blockCount}_${newestBlock}_${rewardPercentiles?.join('')}`;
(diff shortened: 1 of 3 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -177,29 +177,58 @@ export class EthImpl implements Eth {
...
+        this.logger.trace(`${requestIdPrefix} returning cached value ${cacheKey}:${JSON.stringify(feeHistory)}`);
+        return feeHistory;
       }
 
-      blockCount = blockCount > maxResults ? maxResults : blockCount;
+      // check for latest network fees. Ensure cache is populated
+      const lastFeeChangeBlockCacheKey = `${constants.CACHE_KEY.FEE_HISTORY}_${constants.CACHE_KEY.ETH_BLOCK_NUMBER}}`;
+      let lastFeeChangeBlock = this.cache.get(lastFeeChangeBlockCacheKey);
+      if (!lastFeeChangeBlock) {
+        lastFeeChangeBlock = await this.getAndCacheLatestNetworkFeeChangeBlock(requestId);
```
Nana-EC, line 196: I had another idea here.

With the cache block of the last fee change we also have a timestamp to help inform what `latest` string refers to.
If we cache one more block, the block maxResults after the blcok that set the fee, then we can do the following trick.
When you encounter `latest` check the current timestamp, if it's above the (feeCHangeBlock + maxResults).timestamp then again there's no need to call the Mirror Node.
We just create the repeated History.

In this way blocks API need only be called once every hour at most based on activity. After that only if a numeric tag is provided for `newestBlock` whose value - blockCount is less that the `feeCHangeBlock` would we need to call blocks.

Nana-EC:
Closing as with the current fixedFees strategy such an optimization is unnecessary. Can circle back to this in future sprints when needed

PR #1095, opened 2023-04-18
refactor eth_feeHistory to use a fixed fee instead of fetching actual…

### LINKED ISSUE #1087: Optimize eth_feeHistory

AlfredoG87:
… fees, since Hedera almost always has the same fees

**Description**:
refactor eth_feeHistory to use a fixed fee instead of fetching actual fees, since Hedera almost always has the same fees

**Related issue(s)**:

Fixes #1087 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -191,14 +199,25 @@ export class EthImpl implements Eth {
...
-
-      const cacheKey = `${constants.CACHE_KEY.FEE_HISTORY}_${blockCount}_${newestBlock}_${rewardPercentiles?.join('')}`;
-      let feeHistory: object | undefined = this.cache.get(cacheKey);
-      if (!feeHistory) {
-        feeHistory = await this.getFeeHistory(blockCount, newestBlockNumber, latestBlockNumber, rewardPercentiles, requestId);
-        if (newestBlock != EthImpl.blockLatest && newestBlock != EthImpl.blockPending) {
-          this.logger.trace(`${requestIdPrefix} caching ${cacheKey}:${JSON.stringify(feeHistory)} for ${constants.CACHE_TTL.ONE_HOUR} ms`);
-          this.cache.set(cacheKey, feeHistory);
+      let feeHistory: object | undefined;
+      if(this.getEthFeeHistoryFixedFee()) {
```
Nana-EC, line 203: The challenge with this is that `this.translateBlockTag(EthImpl.blockLatest, requestId)` makes a call to get the latest block whereas in this case you are trying to avoid that since today no cases in Mainnet will have a latest block before the fee was changed.

packages/relay/src/lib/eth.ts
```diff
@@ -191,14 +199,25 @@ export class EthImpl implements Eth {
...
+        if(newestBlockNumber <= 0){
```
Nana-EC, line 204: Same here. The fact that you have `newestBlockNumber` means in some cases you made an extra call even though you don't need it in many cases

Nana-EC: Also when will `newestBlockNumber` ever be less than 0?

Nana-EC (changes requested):
Looks clean but I'm not sure we got all the gains we could get without being too complex.
Though you cache the fee for a long time the relay would still  make block calls that it doesn't have to.

In the other approach in addition to caching the fee we would also cache the block at which fees were set.
This allowed us to avoid making any blocks calls if all numbers were above the cached block.

packages/relay/src/lib/eth.ts
Nana-EC: So i forgot the `this.translateBlockTag` internally calls another method which utilizes the `eth_blockNumber`.
It also doesn't do a call on number. 
So we should be good, here.

packages/relay/src/lib/eth.ts
Nana-EC: Correct, I updated my comment above. 
`newestBlockNumber` would only have seen a mirror block call if the block is a number or its `latest` and in both cases it doesn't exist in the cache.

I think we should always attempt a call for `newestBlockNumber` with consideration for  caches.
No need to check if `latest` > `newestBlockNumber` right now cause honestly by the time a response is returned there's a new block with a fee, so we can assume this to be true in your PR

packages/relay/src/lib/eth.ts
AlfredoG87: the additional next block fee in the `feeHistory.baseFeePerGas` should always be there regardless of the newestBlock input is `latest` or a `number`.
We do need to fetch the actual `latest` block number when receiving `latest` due to the response `oldestBlock` that should be `latest`-`blockCount`.

packages/relay/src/lib/eth.ts
AlfredoG87: I've improved the condition that checks for negative values, in reality the validation needs to be for the `oldestBlock` since they could input `newestBlock` positive, and oldest would still return a negative value due to the `blockCount` parameter. so I moved the validation around a bit and tests continued to pass as expected. 👍

packages/relay/src/lib/eth.ts
```diff
@@ -177,28 +185,50 @@ export class EthImpl implements Eth {
...
-      let feeHistory: object | undefined = this.cache.get(cacheKey);
-      if (!feeHistory) {
-        feeHistory = await this.getFeeHistory(blockCount, newestBlockNumber, latestBlockNumber, rewardPercentiles, requestId);
-        if (newestBlock != EthImpl.blockLatest && newestBlock != EthImpl.blockPending) {
-          this.logger.trace(`${requestIdPrefix} caching ${cacheKey}:${JSON.stringify(feeHistory)} for ${constants.CACHE_TTL.ONE_HOUR} ms`);
-          this.cache.set(cacheKey, feeHistory);
+      if(this.getEthFeeHistoryFixedFee()) {
+
+        let oldestBlock = newestBlockNumber - blockCount + 1;
+        if(oldestBlock <= 0) {
```
Nana-EC, line 215: I feel like we should just error out in this case.
Also this will only happen on new envs and more likely local envs. Even then by the time users start hitting it i'm sure 10 blocks have been made.

packages/relay/src/lib/eth.ts
```diff
@@ -222,6 +252,25 @@ export class EthImpl implements Eth {
...
+  private getRepeatedFeeHistory(blockCount: number, oldestBlockNumber: number, rewardPercentiles: Array<number> | null, fee: string) {
+    const shouldIncludeRewards = Array.isArray(rewardPercentiles) && rewardPercentiles.length > 0;
+
+    const feeHistory = {
+      baseFeePerGas: Array(blockCount).fill(fee),
+      gasUsedRatio: Array(blockCount).fill(EthImpl.defaultGasUsedRatio),
+      oldestBlock: EthImpl.numberTo0x(oldestBlockNumber),
+    };
+
+    // next fee
```
Nana-EC, line 264:
```suggestion
    // next fee. Due to high block production rate and low fee change rate we add the next fee since by the time a user utilizes the response there will be a next block likely with the same fee
```

Nana-EC (commented):
Looking good. One comment around an exception and a nit

packages/relay/src/lib/eth.ts
AlfredoG87: I leave it there cause there were 1 Unit Test that had the test case.

Nana-EC: Ahh this is the equivalent of `const oldestBlockNumber = Math.max(0, newestBlockNumber - blockCount + 1);`
I see. Okay.
Can reuse that code if you like

AlfredoG87: Opened another [ticket](https://github.com/hashgraph/hedera-json-rpc-relay/issues/1099) for taking care of this in the future.

PR #1103, opened 2023-04-19
Improve eth_transactionCount perf

### LINKED ISSUE #1038: There's no valid scenario right now to query the consensus node with the blockNum.
### Description

There's no valid scenario right now to query the consensus node with the blockNum.  The Consensus node only supports current account info.  There's no concept of past account info.  If there's a blockNum we should hit the Mirror Node with the timestamp on the account query. If that fails the transaction should fail.

### Steps to reproduce

1. Running a value transfer.
2. Batch3 test "Returns revert reason in receipt for payable methods" consistently triggers an error in previewnet when the code to be replaced, the consensus query code, is run.

### Additional context

_No response_

### Hedera network

previewnet

### Version

v0.21.0

### Operating system

None

### LINKED ISSUE #1102: Optimize eth_transactionCount REST API calls
### Problem

 eth_transactionCount occasionally makes more REST API calls that needed

### Solution

Optimize eth_transactionCount REST API calls

### Alternatives

_No response_

Nana-EC:
**Description**:
There were opportunities to improve the `eth_getTransactionCount` layout and perf

Ultimately 
- Add consideration for all block paths
- Add consideration for partial mirror nodes on earliest block
- Improve logic to cache entity existence 
- Cache nonce for historical calls
- Extract nonce value from past contract results to calculate the appropriate historical nonce value

Expanded version
- Update `MirrorNideClent.ts`
  - Rename account endpoint to clarify getAccountList vs getAccount endpoint
  - Add `ETHEREUM_TRANSACTION_TYPE` to support this transaction type querying
  - Update `getPaginatedResults` to break when a pageMax is reached. This is to ensure we don't page excessively
  - Add `getAccountLatestEthereumTransactionsByTimestamp()` to query for an accounts latest `EthereumTransaction` type transactions
  - Add `isValidContract()` which only cares for entity and caches the presence for a longer period
  - Add `getContractId()` which only cares for entity id and caches the value for a longer period
  - Add `getEarliestBlock()` which uses an optimized query to get the first block
- Update `eth.ts`
  - Organize `getTransactionCount()` to manage the 5 scenarios possible. 
    - no block number - get latest nonce from mirror node
    - genesis block number - return 0 for both accounts and contracts as they would not be present
    - earliest block number - return 0 for both accounts and contracts as they would not be present, if partial mirror node then return error
    - historical block number - return 0 on contract since we need HIP 729. For accounts count the number of `EthereumTransaction` occurrences since block and latest and subtract from latest nonce
    - invalid block number
  - Move cache ttl logic to constants and pull in env variables at top of class
  - Cache `earliest` labels for a day
  - Add `getEthereumNonceFromConsensusNode()` to handle consensus node nonce call logic
  - Add `getAccountLatestEthereumNonce()` to handle mirror node nonce call logic
  - Add `getAccountNonceByTransactionCount()` to handle historical nonce calculation, with concept of too large a range
  - Add `getEthereumTransactionTypeCountInRange()` to handle paginated call for ethereumTransaction types
- Add `UNKNOWN_BLOCK` per execution-api specs
- Add `UNSUPPORTED_OPERATION` to capture unsupported operations
- Fix broken tests

**Related issue(s)**:

Fixes #1102
Fixes #1038 

**Notes for reviewer**:
- Fixed Dapp tests, took the same changes into a smaller PR https://github.com/hashgraph/hedera-json-rpc-relay/pull/1369 and would like to merge that before this

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -882,6 +898,61 @@ export class EthImpl implements Eth {
...
+    return await this.getEthereumTransactionTypeCount(address, endTimestamp, requestId);
+  }
+
+  private async getEthereumTransactionTypeCount(address: string, timestamp: any, requestId: string | undefined) {
+    const ethereumTransactions = await this.mirrorNodeClient.getAccountEthereumTransactionsByTimestampFirstPage(address, timestamp, requestId);
+    if (ethereumTransactions == null || ethereumTransactions.transactions == null || ethereumTransactions.transactions.length === 0) {
+      return 0;
+    }
+
+    return ethereumTransactions.transactions.filter((tx) => tx.name === 'ETHEREUMTRANSACTION').length;
```
Nana-EC, line 953: Had moved the type filter into Mirror Node client so should be able to do
```suggestion
    return ethereumTransactions.transactions.length;
```

packages/relay/src/lib/eth.ts
```diff
@@ -1542,6 +1613,16 @@ export class EthImpl implements Eth {
     return logs;
   }
 
+  private async getAccountLatestEthereumNonce(address: string, requestId?: string) {
```
Nana-EC, line 1616: Appropriately named. Potentially though this doesn't work when account is a contract since contract nonce is likely not externalized yet

packages/relay/src/lib/eth.ts
```diff
@@ -850,26 +851,41 @@ export class EthImpl implements Eth {
...
+      if (nonceCount) {
+        this.logger.trace(`${requestIdPrefix} returning cached value ${cacheKey}:${JSON.stringify(nonceCount)}`);
+        return nonceCount;
       }
-      else if (result?.type === constants.TYPE_CONTRACT) {
-        return EthImpl.numberTo0x(1);
+
+      if (blockNumOrTag) {
+        const blockNum = Number(blockNumOrTag);
+        if (blockNum === 0) {
```
Nana-EC, line 866: Note previewnet and testnet have blocknumber = 1
Need to add consideration for this

Nana-EC:
Waiting on https://github.com/hashgraph/hedera-mirror-node/pull/6034 as this will simplify this PR by not needing to count `EthereumTransaction` instances for an account

AlfredoG87 (approved):
LGTM 👍

ebadiere (approved):
LGTM.  Can we get this merged into main?  I'd like to pull this into my work for the getBlocks, issue #1357.  I could use the "ethGetTransactionCountMaxBlockRange" that you left and is not getting used.

lukelee-sl (dismissed):

packages/relay/src/lib/eth.ts
```diff
@@ -99,6 +100,8 @@ export class EthImpl implements Eth {
   private readonly ethGetBalanceCacheTtlMs = Number.parseInt(process.env.ETH_GET_BALANCE_CACHE_TTL_MS ?? constants.ETH_GET_BALANCE_CACHE_TTL_MS_DEFAULT.toString());
   private readonly maxBlockRange = Number.parseInt(process.env.MAX_BLOCK_RANGE ?? constants.MAX_BLOCK_RANGE.toString());
   private readonly contractCallGasLimit = Number.parseInt(process.env.CONTRACT_CALL_GAS_LIMIT ?? constants.CONTRACT_CALL_GAS_LIMIT.toString());
+  private readonly ethGetTransactionCountCacheTtl = Number.parseInt(process.env.ETH_GET_TRANSACTION_COUNT_CACHE_TTL ?? constants.ETH_GET_TRANSACTION_COUNT_CACHE_TTL.toString());
+  private readonly ethGetTransactionCountMaxBlockRange = Number(process.env.ETH_GET_TRANSACTION_COUNT_MAX_BLOCK_RANGE ?? constants.ETH_GET_TRANSACTION_COUNT_MAX_BLOCK_RANGE.toString());
```
Nana-EC, line 104: remove as not used

packages/relay/src/lib/eth.ts
```diff
@@ -1850,6 +1849,91 @@ export class EthImpl implements Eth {
...
+      return EthImpl.oneHex;
+    }
+
+    // get the transaction result for the latest transaction
+    const transactionResult = await this.mirrorNodeClient.getContractResult(ethereumTransactions.transactions[0].transaction_id, requestId);
+    if (transactionResult == null) {
+      throw predefined.RESOURCE_NOT_FOUND(`Failed to retrieve contract results for transaction ${ethereumTransactions.transactions[0].transaction_id}`);
+    }
+
+    return EthImpl.numberTo0x(transactionResult.nonce + 1); // nonce is 0 indexed
```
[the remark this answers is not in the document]
Nana-EC: @lukelee-sl thanks, also I checked and mirror node API always returns a 0 for accounts with no nonce. I believe the db default is actually 0.

AlfredoG87 (dismissed):

dapp-example/tests/e2e/specs/index.spec.js
```diff
@@ -131,7 +142,7 @@ describe('Test Core Hedera User Scenarios', function() {
     }).timeout(180000);
 
     it('Transfer HBARs', function() {
-      cy.get('#sendHbarsToField').clear().type('0x54C51b7637BF6fE9709e1e0EBc8b2Ca6a24b0f0A');
+      cy.get('#sendHbarsToField').clear().type(randomHollowAccountAddress);
```
Ivo-Yankov, line 145: `randomHollowAccountAddress` is undefined here

ebadiere (commented):
What's the policy on the Copyright date?  I thought it should be updated as we updated the files.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -356,6 +360,20 @@ export class MirrorNodeClient {
             requestId);
     }
 
+    public async getAccountLatestEthereumTransactionsByTimestamp(idOrAliasOrEvmAddress: string, timestampTo: string, numberOfTransactions: number = 1, requestId?: string) {
```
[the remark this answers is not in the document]
AlfredoG87: even so, I actually like it when parameters as explicitly defined in strongly typed languages.

dapp-example/tests/e2e/specs/index.spec.js
Nana-EC: Fixed

packages/relay/src/lib/eth.ts
```diff
@@ -1850,6 +1848,101 @@ export class EthImpl implements Eth {
...
+   */
+  private async getAcccountNonceFromContractResult(address: string, blockNum: any, requestId: string | undefined) {
+    // get block timestamp for blockNum
+    const block = await this.mirrorNodeClient.getBlock(blockNum, requestId); // consider caching error responses
+    if (block == null) {
+      throw predefined.UNKNOWN_BLOCK;
+    }  
+
+    // get the latest 2 ethereum transactions for the account
+    let ethereumTransactions = await this.mirrorNodeClient.getAccountLatestEthereumTransactionsByTimestamp(address, block.timestamp.to, 2, requestId);
```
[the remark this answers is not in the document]
Nana-EC: Updated

Nana-EC:
> What's the policy on the Copyright date? I thought it should be updated as we updated the files.

We should update the files yearly to have the end year be the current year. We can do it in a future PR

AlfredoG87 (dismissed):
LGTM, 💯 

SonarCloud Analysis is failing due to duplicate code on `tests` should we disable checks for duplicate code on tests?

ebadiere (approved):
LG.

PR #1105, opened 2023-04-19
Add Contract Call Query Retry Logic

### LINKED ISSUE #1097: Add retry logic to sdkClient submitContractCallQuery
### Problem

Occasionally calls to `eth_call` fail with a timeout issue.
In some cases it might be valuable to retry the submission

### Solution

Add a `submitContractCallQueryWithRetry` which utilizes [submitContractCallQuery](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L253) with a configureable retry.
Default to 3.

Note we should explore if the transaction needs to be recreated or if the same transaction can just be resubmitted

### Alternatives

_No response_

georgi-l95:
**Description**:
Adds method for contract call query with retry logic for timeout exceeded error from SDK.

**Related issue(s)**:

Fixes #1097 

**Notes for reviewer**:

**Checklist**


lukelee-sl (approved):

AlfredoG87 (approved):
LGTM

PR #1115, opened 2023-04-21
Removed the unnecessary call to the mirrornode for the account transa…

### LINKED ISSUE #1085: Optimize eth_getBalance for Mirror Node 0.78

ebadiere:
**Description**:
Removed unnecessary call to the mirrornode to retrieve account transactions.  It removes calls to the mirrornode when considering transaction level transfers when the block requested in the eth_getBalance call is within the last 15 minutes.
The first call to the mirrornode for the account now returns 100 transactions instead of 20.  If there are still transactions that need to be considered, the old pagination functionality is used to fetch them.

### Background
The current functionality of getBalance retrieves the balance from the account balance in the mirrornode.  This balance is updated every 15 minutes.  If the block queried falls within the last 15 minutes, then the transactions endpoint is called to get a list of transactions from the block queried timestamp.to property, to the the current timestamp.  This makes sense from a hedera perspective but does not honor the block queried. The block is a two second time slice in which transactions for that account may have occurred.   This simple fact confuses the optimization and should be resolved before moving forward.

By considering the now 100 transactions available in the initial call to the mirror node for the account information we can remove most calls for the transactions that fall in the fifteen minute window, however this means we need to have a clear understanding of the block timestamps we are querying the mirror node for, and a strategy for paginating the account transactions in the case where more than 100 transaction may have occurred for that account in the last fifteen minutes.

The latest solutions pushed up is "a" solution, but not a clean solution in that it:

1. considers the old timestamp query for transactions to the mirror node that does not honor the blocks.
2. reuses the transactions endpoint for any necessary transaction pagination.

The tests all pass, but this solution smells.  It s a good starting point to illustrate what needs to be done, but the proper solution would address the two points mentioned above.

```

Fixes #1085

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -80,8 +80,6 @@ export class MirrorNodeClient {
 
     private static CONTRACT_RESULT_LOGS_PROPERTY = 'logs';
     private static CONTRACT_STATE_PROPERTY = 'state';
-    private static PAGE_LIMIT = 100;
-
 
 
     private static ORDER = {
@@ -324,7 +322,7 @@ export class MirrorNodeClient {
     }
 
     public async getAccountPageLimit(idOrAliasOrEvmAddress: string, requestId?: string) {
-        return this.get(`${MirrorNodeClient.GET_ACCOUNTS_ENDPOINT}${idOrAliasOrEvmAddress}?limit=${MirrorNodeClient.PAGE_LIMIT}`,
+        return this.get(`${MirrorNodeClient.GET_ACCOUNTS_ENDPOINT}${idOrAliasOrEvmAddress}?limit=${constants.MIRROR_NODE_QUERY_LIMIT}`,
             MirrorNodeClient.GET_ACCOUNTS_ENDPOINT,
             [400, 404],
             requestId);
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -80,6 +80,7 @@ export class MirrorNodeClient {
 
     private static CONTRACT_RESULT_LOGS_PROPERTY = 'logs';
     private static CONTRACT_STATE_PROPERTY = 'state';
+    private static PAGE_LIMIT = 100;
 
 
 
@@ -317,6 +318,13 @@ export class MirrorNodeClient {
 
     public async getAccount(idOrAliasOrEvmAddress: string, requestId?: string) {
         return this.get(`${MirrorNodeClient.GET_ACCOUNTS_ENDPOINT}${idOrAliasOrEvmAddress}`,
+            MirrorNodeClient.GET_ACCOUNTS_ENDPOINT,
+            [400, 404],
+            requestId);
+    }
+
+    public async getAccountPageLimit(idOrAliasOrEvmAddress: string, requestId?: string) {
+        return this.get(`${MirrorNodeClient.GET_ACCOUNTS_ENDPOINT}${idOrAliasOrEvmAddress}?limit=${MirrorNodeClient.PAGE_LIMIT}`,
             MirrorNodeClient.GET_ACCOUNTS_ENDPOINT,
             [400, 404],
             requestId);
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -642,7 +642,7 @@ export class EthImpl implements Eth {
     let blockNumber = null;
     let balanceFound = false;
     let weibars: BigInt = BigInt(0);
-    const mirrorAccount = await this.mirrorNodeClient.getAccount(account, requestId);
+    const mirrorAccount = await this.mirrorNodeClient.getAccountPageLimit(account, requestId);
 
     try {
       if (!EthImpl.blockTagIsLatestOrPending(blockNumberOrTag)) {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: packages/relay/src/lib/eth.ts [hunk 1]
```diff
@@ -669,12 +669,9 @@ export class EthImpl implements Eth {
                   currentTimestamp = mirrorAccount.balance.timestamp;
                 }
 
-                let transactionsInTimeWindow = await this.mirrorNodeClient.getTransactionsForAccount(
-                  mirrorAccount.account,
-                  block.timestamp.to,
-                  currentTimestamp,
-                  requestId
-                );
+                const transactionsInTimeWindow: any = mirrorAccount.transactions.filter((tx: any) => {
```
Nana-EC, line 672: Nice idea.
Challenge here like it was before is that this is a best guess. 
If the number of transactions is greater than the page limit we may be missing credits and debits and therefore missing paging logic.

In this PR i think you should update `this.mirrorNodeClient.getAccount` or add a `this.mirrorNodeClient.getAccountWithPageLimit()` that makes the call and sets the limit=100 (the page limit)
If not you'll only get 25

Nana-EC: Can add a ticket for the paging aspect to address later

Nana-EC (commented):
Nice catch.
I'd say increase the page limit and we should be good for this fix

packages/relay/src/lib/eth.ts
ebadiere: Updated.

ebadiere: Agreed.  Basically the pagination handles this issue.  My latest update checks for the need to fetch the next page and handles it if needed.  The latest mirrornode API returns 100 transactions per call on the account endpoint, and a brief look at mainnet suggests that the 100 transactions within the 15 minutes window, causing the need to load the next page, would be more of an exceptional case rather than the norm, so I would say this is an optimization.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -322,6 +323,13 @@ export class MirrorNodeClient {
             requestId);
     }
 
+    public async getAccountPageLimit(idOrAliasOrEvmAddress: string, requestId?: string) {
+        return this.get(`${MirrorNodeClient.GET_ACCOUNTS_ENDPOINT}${idOrAliasOrEvmAddress}?limit=${MirrorNodeClient.PAGE_LIMIT}`,
```
Nana-EC, line 327: This already exists with `constants.MIRROR_NODE_QUERY_LIMIT`
```suggestion
        return this.get(`${MirrorNodeClient.GET_ACCOUNTS_ENDPOINT}${idOrAliasOrEvmAddress}?limit=${constants.MIRROR_NODE_QUERY_LIMIT}`,
```

packages/relay/src/lib/eth.ts
```diff
@@ -669,12 +669,26 @@ export class EthImpl implements Eth {
...
-                  block.timestamp.to,
-                  currentTimestamp,
-                  requestId
-                );
+                // Need to check if there are any transactions before the block.timestamp.to in the current account set returned from the inital
+                // call to getAccountPageLimit.  If there are we may need to paginate.
+                let lastTransactionOnPageTimestamp;
+                if(mirrorAccount.links.next !== null) {
+                  // Get the end of the page of transactions timestamp
+                  lastTransactionOnPageTimestamp = mirrorAccount.links.next.substring(mirrorAccount.links.next.indexOf("gte:") + 4, mirrorAccount.links.next.indexOf("&timestamp"));
```
Nana-EC, line 677: Q: Is this safe way to parse the string out?
What if the params order are changed on the mirror node side, we'd be broken even though the functionality didn't change.
Would it be better to split the params on `&`, take the time stamp params and split that on appropriately?

packages/relay/src/lib/eth.ts
```diff
@@ -669,12 +669,26 @@ export class EthImpl implements Eth {
...
+                }
+                let transactionsInTimeWindow: any = [];
+                if((typeof lastTransactionOnPageTimestamp !== "undefined") && (mirrorAccount.transactions[mirrorAccount.transactions.length -1].consensus_timestamp >= lastTransactionOnPageTimestamp)) {
```
Nana-EC, line 680: Q: what's this testing or looking for and what's the significance?
Also unclear how they connect to the `block.timestamp.to` and `currentTimestamp` boundaries when true

Nana-EC (commented):
Looking good.
One not on the limit constant and some clarity requested on logic to be clear.
Also a potential string parsing suggesion

packages/relay/src/lib/clients/mirrorNodeClient.ts
ebadiere: Updated

packages/relay/src/lib/eth.ts
ebadiere: Updated to safer parsing with logging of unintended results.

packages/relay/src/lib/eth.ts
ebadiere: No longer necessary with the improved parsing described above.  It was trying to avoid a null or undefined related error before testing the value.

Nana-EC: The code is still there. I'm not clear here what this if is doing.
Are you just checking to see if there are more transaction since the next link has a value?

packages/relay/src/lib/eth.ts
```diff
@@ -669,12 +669,38 @@ export class EthImpl implements Eth {
...
+                  const timestampParameters = params.getAll('timestamp'); 
+                  lastTransactionOnPageTimestamp = timestampParameters[0].split(':')[1];
+                  if((lastTransactionOnPageTimestamp === null) || (lastTransactionOnPageTimestamp === undefined)) {
+                    this.logger.debug(`${requestIdPrefix} Unable to find expected beginning (gte:) timestamp in account next page link ${mirrorAccount.links.next}), returning 0x0 balance`);
+                    return EthImpl.zeroHex;
+                  }
+                }
+
+                let transactionsInTimeWindow: any = [];
+                if((typeof lastTransactionOnPageTimestamp !== "undefined") && (mirrorAccount.transactions[mirrorAccount.transactions.length -1].consensus_timestamp >= lastTransactionOnPageTimestamp)) {
```
Nana-EC, line 692: SO this is an example of the last transaction and a valid next link
```
{"bytes":null,"charged_tx_fee":169304,"consensus_timestamp":"1682378101.517897720","entity_id":null,"max_fee":"100000000","memo_base64":"RGV2T3BzIFN5bnRoZXRpYyAyLjAgVGVzdGluZw==","name":"CRYPTOTRANSFER","node":"0.0.8","nonce":0,"parent_consensus_timestamp":null,"result":"SUCCESS","scheduled":false,"staking_reward_transfers":[],"transaction_hash":"1iSh8lSUcEXcdsydXm+9CCq5cUodnTKUR4JfCb2BMU5YF6kaU3QintjF/m91l2iP","transaction_id":"0.0.90-1682378091-167906093","transfers":[{"account":"0.0.2","amount":1,"is_approval":false},{"account":"0.0.8","amount":6866,"is_approval":false},{"account":"0.0.90","amount":-169305,"is_approval":false},{"account":"0.0.98","amount":162438,"is_approval":false}],"valid_duration_seconds":"120","valid_start_timestamp":"1682378091.167906093"}],"links":{"next":"/api/v1/accounts/0.0.2?timestamp=lt:1682378101.517897720"}}
```

By definition the last transaction timestamp and the next timestamp value should be equal, so maybe you only have to check for a valid next.

However in this case it's still not clear to me what's being checked for.

What does the current time stamp refer to in this flow and how come this query doesn't use the transaction next timestamp.
Is there a timestamp gap here?

Nana-EC (commented):
Still unclear of the logic here and what time ranges are being considered and why

Nana-EC (approved):
LG.
Improvements to follow up in 0.23 to finish it off. We'l get some wins with this. Thanks

AlfredoG87 (approved):
LGTM 👍

PR #1122, opened 2023-04-24
Expand eth_call mirror node consensus retry scenarios

### LINKED ISSUE #1121: Expand eth_call retry allowance on Mirror Node failure
### Problem

Currently the `eth_call` flow to the mirror node will only retry if the error caught is of the Not supported type.
Whiles this is the final desired approach, whiles `eth_call` support on the mirror node experiences a staged out enablement the relay should retry on all unsuccessful cases.

### Solution

Update retry logic to retry on all failures to from the mirror node until full coverage is confirmed outside of non supported scenarios

### Alternatives

_No response_

Nana-EC:
**Description**:
Expand eth_call mirror node consensus retry scenarios

- Move consensusNode call in catch out to apply to all error cases
- Split mirror node call case into separate method with catch including consensus call
- Add `toNullableBigNumber` to manage value conversion
- Add `getCappedBlockGasLimit` to manage block gas limit max

**Related issue(s)**:

Fixes #1121

**Notes for reviewer**:

**Checklist**


Nana-EC (commented):
Thanks for the additions @georgi-l95

AlfredoG87:
@georgi-l95 Why it was needed for this change to include these subgraphs example files?
image

AlfredoG87:
> @georgi-l95 Why it was needed for this change to include these subgraphs example files? image

I see you have updated the PR, nvm 👍

georgi-l95:
> @georgi-l95 Why it was needed for this change to include these subgraphs example files? image

Removed them, at first after adding them they appeared to have fixed the subgraph failing tests. But now after I saw that other PRs are failing as well, I think that it's something beyond the changes in this PR. ( maybe runners are having some problems? )

AlfredoG87 (approved):
LGTM 👍

PR #1132, opened 2023-04-25
Add `eth_call` workaround for mirror-node empty response

### LINKED ISSUE #1130: Add eth_call workaround for mirror-node empty response
Add eth_call workaround for mirror-node empty response, which looks like this:
```
{
    "result": "0x"
}
```
This means that there is no result for some reason, but call to consensus node returns correct result.

georgi-l95:
**Description**:
This PR adds a temporary workaround for empty response from mirror-node `eth_call`. If `result: 0x` is returned, we fallback to consensus-node.

**Related issue(s)**:

Fixes #1130 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1055,7 +1055,10 @@ export class EthImpl implements Eth {
 
       const contractCallResponse = await this.mirrorNodeClient.postContractCall(callData, requestId);
       if (contractCallResponse && contractCallResponse.result) {
-        return contractCallResponse.result == EthImpl.emptyHex ? await this.callConsensusNode(call, gas, requestId) : EthImpl.prepend0x(contractCallResponse.result)
+        if (contractCallResponse.result == EthImpl.emptyHex) {
+          throw new MirrorNodeClientError({ message: "Empty Response" }, MirrorNodeClientError.statusCodes.NO_CONTENT);
+        }
+        return EthImpl.prepend0x(contractCallResponse.result);
       }
       return EthImpl.emptyHex;
     } catch (e: any) {
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1054,7 +1054,10 @@ export class EthImpl implements Eth {
       }
 
       const contractCallResponse = await this.mirrorNodeClient.postContractCall(callData, requestId);
-      return contractCallResponse && contractCallResponse.result ? EthImpl.prepend0x(contractCallResponse.result) : EthImpl.emptyHex;
+      if (contractCallResponse && contractCallResponse.result) {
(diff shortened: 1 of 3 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1054,7 +1054,10 @@ export class EthImpl implements Eth {
       }
 
       const contractCallResponse = await this.mirrorNodeClient.postContractCall(callData, requestId);
-      return contractCallResponse && contractCallResponse.result ? EthImpl.prepend0x(contractCallResponse.result) : EthImpl.emptyHex;
+      if (contractCallResponse && contractCallResponse.result) {
+        return contractCallResponse.result == EthImpl.emptyHex ? await this.callConsensusNode(call, gas, requestId) : EthImpl.prepend0x(contractCallResponse.result)
```
Nana-EC, line 1058: Make it cleaner with a check for `EthImpl.emptyHex` that throws a `MirrorNodeClientError`.

is there a way to determine the type of case by the call being make so that not all `0x`'s are retried on?

Nana-EC (changes requested):
I think we should separate but also consider if we can filter the condition

packages/relay/src/lib/eth.ts
Nana-EC: Also add a trace log to say this issue occured and we're retrying with consensus node so we can track it

Nana-EC: I get that we can't tell from the Mirror Node side now. So I think just the extra log and we're good.

Nana-EC: Actually @georgi-l95 I take that back we should be able to check the byte code value by calling `api/v1/contracts/(call.to}` and checking the runtime bytecode.
That should let us know we're in this specific scenario and retry.

I'm trying to avoid retrying on all `0x` responses cause many could be valid

georgi-l95: Yeah, i've added this check. Now if runtime_bytecode is 0x, we throw error and proceed with consensus.

packages/relay/src/lib/eth.ts
```diff
@@ -1042,9 +1042,14 @@ export class EthImpl implements Eth {
     }
   }
 
-  async callMirrorNode(call: any, gas: number, value: string | null, requestId?: string): Promise<string | JsonRpcError> {
+  async callMirrorNode(call: any, to: any, gas: number, value: string | null, requestId?: string): Promise<string | JsonRpcError> {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     try {
+      if (to?.type === constants.TYPE_CONTRACT && to?.entity.runtime_bytecode === EthImpl.emptyHex) {
```
Nana-EC, line 1048: Nice. I like this. We get to even skip the POST call.

Nana-EC (approved):
LGTM

PR #1145, opened 2023-04-28
refactor: use REQUEST_ID_STRING from constants instead of replicating

bguiz:
**Description**:

This PR modifies ... in order to support ...
- refactor: use `REQUEST_ID_STRING` from `constants` instead of replicating
- test: added test for `JsonRpcError` constructor

**Related issue(s)**:

- N/A

**Notes for reviewer**:

- Should be better for code coverage, as `JsonRpcError` was previously only tested directly

**Checklist**

- [ ] Documented (Code comments, README, etc.) --> N/A

### DIFF
--- a/packages/relay/tests/lib/errors/JsonRpcError.spec.ts
+++ b/packages/relay/tests/lib/errors/JsonRpcError.spec.ts
@@ -1,0 +1,36 @@
+import { expect } from 'chai';
+import { JsonRpcError } from '../../../src/lib/errors/JsonRpcError';
+
+describe('Errors', () => {
+    describe('JsonRpcError', () => {
+        it('Constructs correctly without request ID', () => {
+            const err = new JsonRpcError({
+                name: 'TestError',
+                code: -32999,
+                message: 'test error: foo',
+                data: 'some data'
+            });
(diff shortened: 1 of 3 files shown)


packages/relay/tests/lib/errors/JsonRpcError.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { expect } from 'chai';
```
Nana-EC, line 1: Please add the license header to the top here for OSS compliance.
You'll see it in all the other `.ts` files and should be 2023

Nana-EC (commented):
Nice, just missing a license header

packages/relay/tests/lib/errors/JsonRpcError.spec.ts
bguiz: ah, I used `packages/relay/tests/lib/formatters.spec.ts` as a reference, and it was missing a header ... assumed that it wasn't needed for test files - will add one to both now

Nana-EC: Thanks

Nana-EC (approved):

PR #1158, opened 2023-04-29
Add resolve type consideration for tokens

### LINKED ISSUE #1156: Add evm address consideration to MirrorNodeClient resolveEntityType
### Problem

Currently the `MirrorNodeClient.resolveEntityType()` checks for a matching entity for an address by calling both Mirror Node `api/v1/accounts` and `api/v1/tokens`.
However, tokens are always long zero Hedera addresses not high value EVM address

### Solution

Add logic to check for non long zero addresses. 
If valid no need to add the token consideration as the account requests will find a match if the entity is valid

### Alternatives

_No response_

Nana-EC:
**Description**:
Handle valid evm contract address considerations as part of entity resolution

- Add path to MirrorNodeClient retry log
- Conditionally add `this.getTokenById()` to `resolveEntityType()` only if address is of long zero type
- Move requestId to add of `resolveEntityType()` parameters
- Added test for to verify token is only called for long zero address

**Related issue(s)**:

Fixes #1156 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -655,7 +655,7 @@ export class MirrorNodeClient {
      * @param entityIdentifier the address of the contract
      * @param requestId the request id
      * @param searchableTypes the types to search for
-     * @returns enitty object or null if not found
+     * @returns entity object or null if not found
      */
     public async resolveEntityType(
       entityIdentifier: string,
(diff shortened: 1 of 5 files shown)


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -652,10 +652,17 @@ export class MirrorNodeClient {
         }
     }
 
+    /**
+     * Get the contract results for a given address
+     * @param entityIdentifier the address of the contract
+     * @param requestId the request id
+     * @param searchableTypes the types to search for
+     * @returns enitty object or null if not found
```
lukelee-sl, line 660: nit: typo

lukelee-sl (dismissed):
LGTM

ebadiere (commented):
LG

AlfredoG87 (approved):
LGTM 👍 
This will further help with the optimization efforts!! 💯

packages/relay/src/lib/clients/mirrorNodeClient.ts
Nana-EC: Fixed

PR #1163, opened 2023-05-01
Axios Retries against mirror node refactored

### LINKED ISSUE #1154: Improve axios retry condition logic
### Problem

Currently the `mirrorNodeClient.ts` use teh `MirrorNodeClientError.retryErrorCodes` in the `retryCondition`, however this result sin all paths retrying regardless of the `allowedErrorStatuses` passed in.

### Solution

Update the `retryCondition` to use a `MirrorNodeClientError.retryErrorCodes` that is a map of paths to array of retry codes instead of a single array.
In this way each path exposed to mirror node can have it's set statically set once.

With this `allowedErrorStatuses` can be removed from `request` and some clean up may be done thus reducing the number of retries.

### Alternatives

_No response_

AlfredoG87:
**Description**:
refactored how the allowedErrorStatuses work, since they were not actually working as it was intended, so this will improve the number of retries we perform against mirror node

**Related issue(s)**:

Fixes #1154 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -83,20 +83,20 @@ export class MirrorNodeClient {
 
     static acceptedErrorStatusesResponsePerRequestPathMap: Map<string, Array<number>> = new Map([
         [MirrorNodeClient.GET_ACCOUNTS_ENDPOINT, [400, 404]],
-        [MirrorNodeClient.GET_BALANCE_ENDPOINT, [400, 404]],
+        [MirrorNodeClient.GET_BALANCE_ENDPOINT, [400]],
         [MirrorNodeClient.GET_BLOCK_ENDPOINT, [400, 404]],
-        [MirrorNodeClient.GET_BLOCKS_ENDPOINT, [400, 404]],
+        [MirrorNodeClient.GET_BLOCKS_ENDPOINT, [400]],
         [MirrorNodeClient.GET_CONTRACT_ENDPOINT, [400, 404]],
         [MirrorNodeClient.GET_CONTRACT_RESULTS_BY_ADDRESS_ENDPOINT, [206, 400, 404]],
         [MirrorNodeClient.GET_CONTRACT_RESULTS_DETAILS_BY_CONTRACT_ID_ENDPOINT, [400]],
         [MirrorNodeClient.GET_CONTRACT_RESULT_ENDPOINT, [400, 404]],
-        [MirrorNodeClient.GET_CONTRACT_RESULT_LOGS_ENDPOINT, [400, 404]],
+        [MirrorNodeClient.GET_CONTRACT_RESULT_LOGS_ENDPOINT, [400]],
         [MirrorNodeClient.GET_CONTRACT_RESULT_LOGS_BY_ADDRESS_ENDPOINT, [400, 404]],
-        [MirrorNodeClient.GET_CONTRACT_RESULTS_ENDPOINT, [400, 404]],
+        [MirrorNodeClient.GET_CONTRACT_RESULTS_ENDPOINT, [400]],
         [MirrorNodeClient.GET_NETWORK_EXCHANGERATE_ENDPOINT, [400, 404]],
         [MirrorNodeClient.GET_NETWORK_FEES_ENDPOINT, [400, 404]],
-        [MirrorNodeClient.GET_TOKENS_ENDPOINT, [400, 404]],
-        [MirrorNodeClient.GET_TRANSACTIONS_ENDPOINT, [400, 404]],
+        [MirrorNodeClient.GET_TOKENS_ENDPOINT, [400]],
+        [MirrorNodeClient.GET_TRANSACTIONS_ENDPOINT, [400]],
         [MirrorNodeClient.CONTRACT_CALL_ENDPOINT, []],
         [MirrorNodeClient.GET_STATE_ENDPOINT, [400, 404]]
     ]);
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -83,20 +83,20 @@ export class MirrorNodeClient {
 
     static acceptedErrorStatusesResponsePerRequestPathMap: Map<string, Array<number>> = new Map([
         [MirrorNodeClient.GET_ACCOUNTS_ENDPOINT, [400, 404]],
-        [MirrorNodeClient.GET_BALANCE_ENDPOINT, [400]],
+        [MirrorNodeClient.GET_BALANCE_ENDPOINT, [400, 404]],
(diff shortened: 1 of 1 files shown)


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -81,6 +81,25 @@ export class MirrorNodeClient {
     private static CONTRACT_RESULT_LOGS_PROPERTY = 'logs';
     private static CONTRACT_STATE_PROPERTY = 'state';
 
+    static acceptedErrorStatusesResponsePerRequestPathMap: Map<string, Array<number>> = new Map([
+        [MirrorNodeClient.GET_ACCOUNTS_ENDPOINT, [400, 404]],
```
steven-sheehy, line 85: Why would you ever retry a 400 for any mirror node endpoint? It's invalid input

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -81,6 +81,25 @@ export class MirrorNodeClient {
...
+        [MirrorNodeClient.GET_BALANCE_ENDPOINT, [400, 404]],
+        [MirrorNodeClient.GET_BLOCK_ENDPOINT, [400, 404]],
+        [MirrorNodeClient.GET_BLOCKS_ENDPOINT, [400, 404]],
```
steven-sheehy, line 88: We will never return a 404 for any list endpoint.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -81,6 +81,25 @@ export class MirrorNodeClient {
...
+        [MirrorNodeClient.GET_CONTRACT_ENDPOINT, [400, 404]],
+        [MirrorNodeClient.GET_CONTRACT_RESULTS_BY_ADDRESS_ENDPOINT, [206, 400, 404]],
+        [MirrorNodeClient.GET_CONTRACT_RESULTS_DETAILS_BY_CONTRACT_ID_ENDPOINT, [400]],
+        [MirrorNodeClient.GET_CONTRACT_RESULT_ENDPOINT, [400, 404]],
+        [MirrorNodeClient.GET_CONTRACT_RESULT_LOGS_ENDPOINT, [400, 404]],
```
steven-sheehy, line 93: How are you coming up with these codes? You should be using the [OpenAPI spec file](https://github.com/hashgraph/hedera-mirror-node/blob/main/hedera-mirror-rest/api/v1/openapi.yml). Bonus points if you can replace this map with a dynamic call to the openapi yaml from the corresponding mirror node. We expose this yaml directly.

steven-sheehy (changes requested):

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: actually this refactor is precisely to stop doing retries if we receive any of the status codes on that list.

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: what you mean by `list` endpoint?

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: Right now I am not improving on the reality of the mirror node, I think that is ideal, to be honest, but right now am focusing on just refactoring the existent code and those codes were already there as acceptedErrorCodes for these endpoints, in the meaning that we will not retry if we get any of those errors. 
Due to how it is coded, fetching it from the yaml is not feasible at the time, since these `Constants` are being used to replace the `{Placeholders}` and then use them as path for the requests, in that case I need them to match exactly in the client in order to verify against this list. so for the time being I think they will stay hardcoded.

packages/relay/src/lib/clients/mirrorNodeClient.ts
Nana-EC: Many of the API endpoints return lists vs a single item e.g. `api/v1/accounts` return a list but `api/v1/accounts/{accountId}` will return a single object.
The former would return an empty list on no data but the latter would return a 404.
We should only look 404's on the non list endpoints.

Granted some business logic down the line may hint to a retry if a list endpoint returned nothin when we expect some data e.g. calling for transactions before mirror node processed the block for an accounts 1st transaction

packages/relay/src/lib/clients/mirrorNodeClient.ts
Nana-EC: We've discussed the creating a client from the Mirror Node OpenAPI spec a few times. 
Let's open a ticket to address this in a future sprint.
Right thing to do.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -255,29 +274,28 @@ export class MirrorNodeClient {
...
+    handleError(error: any, path: string, pathLabel: string, effectiveStatusCode: number, method: REQUEST_METHODS, requestId?: string) {
         const mirrorError = new MirrorNodeClientError(error, effectiveStatusCode);   
         const requestIdPrefix = formatRequestIdMessage(requestId);
-        if (allowedErrorStatuses && allowedErrorStatuses.length) {
-            if (error.response && allowedErrorStatuses.indexOf(effectiveStatusCode) !== -1) {
-                this.logger.debug(`${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`);
-                return null;
-            }
+        const acceptedErrorResponses = MirrorNodeClient.acceptedErrorStatusesResponsePerRequestPathMap.get(pathLabel);
+        if (error.response && acceptedErrorResponses && acceptedErrorResponses.indexOf(effectiveStatusCode) !== -1) {
```
steven-sheehy, line 296: Shouldn't there be some default fallback like before in case you use some new endpoint but forget to update this map? Or is there a test that ensures this map has every possible endpoint invoked within the `mirrorNodeClient`?

steven-sheehy (approved):

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: Good Question, Right now the fallback is throwing out an error. but due to the nature of this list is accepting an error code as a `Accepted` response. I think that throwing an error is the right fallback.

AlfredoG87: Yes I can open up a ticket for that. 👍

Nana-EC: Yeah, it's just a programmatic check for our future selves :)

Nana-EC (approved):
LG

ebadiere (commented):
LGTM

PR #1164, opened 2023-05-02
Reapplying refactor/enhancement in a fresh branch off of main since t…

### LINKED ISSUE #1085: Optimize eth_getBalance for Mirror Node 0.78

ebadiere:
**Description**:

This PR optimizes the getBalance method by removing unnecessary calls to the mirror node. 

**Related issue(s)**:

Fixes #1085 

**Notes for reviewer**:
Adds tests around the pagination.

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -110,6 +110,7 @@ export class MirrorNodeClient {
     private mirrorResponseHistogram;
 
     private readonly cache;
+    static readonly EVM_ADDRESS_REGEX: RegExp = /\/accounts\/([\d\.]+)/;    
 
     protected createAxiosClient(
         baseUrl: string
@@ -327,6 +328,50 @@ export class MirrorNodeClient {
             [400, 404],
             requestId);
     }
+    /*******************************************************************************
+     * To be used to make paginated calls for the account information when the 
+     * transaction count exceeds the constant MIRROR_NODE_QUERY_LIMIT.
+     *******************************************************************************/
+    public async getAccountPaginated(url: string, requestId?: string) {
+        const queryParamObject = {};
+        const accountId = this.extractAccountIdFromUrl(url, requestId);
+        const params = new URLSearchParams(url.split('?')[1]);
+        
+        this.setQueryParam(queryParamObject, 'limit', constants.MIRROR_NODE_QUERY_LIMIT);
+        this.setQueryParam(queryParamObject, 'timestamp', params.get('timestamp'));
+        const queryParams = this.getQueryParams(queryParamObject);
+
+        return this.getPaginatedResults(
+            `${MirrorNodeClient.GET_ACCOUNTS_ENDPOINT}${accountId}${queryParams}`,
+            MirrorNodeClient.GET_ACCOUNTS_ENDPOINT,
+            'transactions',
+            [400, 404],
+            requestId
+        );
+    }
+
+    public extractAccountIdFromUrl(url: string, requestId?: string): string | null {
+        const substringStartIndex = url.indexOf("/accounts/") + "/accounts/".length;
+        if (url.startsWith("0x", substringStartIndex)) {
+            // evm addresss
+            const regex = /\/accounts\/(0x[a-fA-F0-9]{40})/;
+            const match = url.match(regex);
+            const accountId = match ? match[1] : null;
+            if (!accountId) {
+                this.logger.error(`${formatRequestIdMessage(requestId)} Unable to extract evm address from url ${url}`);
+            }
+            return accountId;
+        } else {
+            // account id 
+            const match = url.match(MirrorNodeClient.EVM_ADDRESS_REGEX);
+            const accountId = match ? match[1] : null;
+            if (!accountId) {
+                this.logger.error(`${formatRequestIdMessage(requestId)} Unable to extract account ID from url ${url}`);
+            }
+            return accountId;
+        }
+    }
+
 
     public async getTransactionsForAccount(accountId: string, timestampFrom: string, timestampTo: string, requestId?: string) {
         const queryParamObject = {};
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -35,6 +35,10 @@ import crypto from 'crypto';
 const LRU = require('lru-cache');
 const _ = require('lodash');
 const createHash = require('keccak');
+interface LatestBlockNumberTimestamp {
+  blockNumber: string;
+  timeStampTo: string;
+}
 
 /**
  * Implementation of the "eth_" methods from the Ethereum JSON-RPC API.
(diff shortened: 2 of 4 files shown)


.github/workflows/dapp.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   setup-local-hedera:
     name: Dapp Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 35
+    timeout-minutes: 65
```
Nana-EC, line 14: Still concerned about this bump. I'd like this to not be necessary, but if it is we should be clear on what's going on

Nana-EC, line 14: Still concerned about this bump. I'd like this to not be necessary, but if it is we should be clear on what's going on

Nana-EC (commented):
LG. Just waiting on the timeout increase removal or explanation

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -327,6 +328,50 @@ export class MirrorNodeClient {
...
+        
+        this.setQueryParam(queryParamObject, 'limit', constants.MIRROR_NODE_QUERY_LIMIT);
+        this.setQueryParam(queryParamObject, 'timestamp', params.get('timestamp'));
+        const queryParams = this.getQueryParams(queryParamObject);
+
+        return this.getPaginatedResults(
+            `${MirrorNodeClient.GET_ACCOUNTS_ENDPOINT}${accountId}${queryParams}`,
+            MirrorNodeClient.GET_ACCOUNTS_ENDPOINT,
+            'transactions',
+            [400, 404],
```
AlfredoG87, line 348: I missed this on [PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1163) 
But we no longer need to send the acceptedResponseStatusCodes to the client.

It would be nice if you can remove that param from here and from the `getPaginatedResults` so it does not raise confusion.

on the details of `getPaginatedResults` you will notice that is not doing anything that passing that parameter to itself without actually using it.

AlfredoG87 (changes requested):

Nana-EC (approved):

packages/relay/src/lib/eth.ts
```diff
@@ -624,15 +657,31 @@ export class EthImpl implements Eth {
     const latestBlockTolerance = 1;
     this.logger.trace(`${requestIdPrefix} getBalance(account=${account}, blockNumberOrTag=${blockNumberOrTag})`);
 
+    let latestBlock: LatestBlockNumberTimestamp | null | undefined;
     // this check is required, because some tools like Metamask pass for parameter latest block, with a number (ex 0x30ea)
     // tolerance is needed, because there is a small delay between requesting latest block from blockNumber and passing it here
     if (!EthImpl.blockTagIsLatestOrPending(blockNumberOrTag)) {
-      const latestBlock = await this.blockNumber(requestId);
-      const blockDiff = Number(latestBlock) - Number(blockNumberOrTag);
+      const cacheKey = `${constants.CACHE_KEY.ETH_BLOCK_NUMBER}`;
```
lukelee-sl, line 664: are we caching a single block value?  It seems to me that the cache key is a constant `'eth_block_number'`.  Do we not want to add the block number to the cache key somehow?

packages/relay/src/lib/eth.ts
```diff
@@ -624,15 +657,31 @@ export class EthImpl implements Eth {
...
+      const blockNumberCached = this.cache.get(cacheKey);
+      
+      if(blockNumberCached) {
+        this.logger.trace(`${requestIdPrefix} returning cached value ${cacheKey}:${JSON.stringify(blockNumberCached)}`);
+        latestBlock = { blockNumber: blockNumberCached, timeStampTo: '0' };
+
+      } else {
+        latestBlock = await this.blockNumberTimestamp(requestId);
```
lukelee-sl, line 672: do we not want to add to the cache here?  Maybe, I'm misunderstanding what is happening.

packages/relay/src/lib/eth.ts
```diff
@@ -624,15 +657,31 @@ export class EthImpl implements Eth {
...
+      const blockDiff = Number(latestBlock.blockNumber) - Number(blockNumberOrTag);
 
       if (blockDiff <= latestBlockTolerance) {
         blockNumberOrTag = EthImpl.blockLatest;
       }
+
+      // If ever we get the latest block from cache, and blockNumberOrTag is not latest, then we need to get the block timestamp
+      // This should rarely happen.
+      if((blockNumberOrTag !== EthImpl.blockLatest) && (latestBlock.timeStampTo === "0")) {
+        latestBlock = await this.blockNumberTimestamp(requestId);
```
lukelee-sl, line 683: do we want to add to the cache here even though this may not happen very often?

.github/workflows/dapp.yml
ebadiere: Not sure how this is still here as on my box it's set to 35.  I simple added a comment to the setting to make a difference to check in and push up.

packages/relay/src/lib/eth.ts
ebadiere: Yes, we're only caching a single block value with a time to live set to 1000 ms.  This snippet was extracted from the original blockNumber method that's been working up to now.  The problem was the blockNumber method which is used in many places only returns the block number and later in the getBalance process the latest block timestamp is needed which caused another call to the mirror node.  In this new getBalance only version of the method we also return the latest block timestamp saving a call to the mirrornode.

packages/relay/src/lib/eth.ts
ebadiere: We could but it would require its own separate cache key, and value.  The old getBlock method also sets the cache and is called in many places, so by the time metamask connects and calls getBalance for the first time, the cache has a value in it.  Using a unique cache for getBalance would mean that it gets populated on the first call to getBalance only which I am afraid would cause the behavior I was seeing before this fix.  Metamask passed in the latest block using the block number and not the tag, so the small delay in look up caused an inconsistent result.  It defeated the purpose of the blocktolerance setting.

packages/relay/src/lib/eth.ts
ebadiere: Well block times are every two seconds and the cached value has a time to live set to one, so I think it makes sense to use the cache most used, hence updated most often, and if that cached value is not the latest then make a call to the mirrornode.

Nana-EC (approved):

packages/relay/src/lib/eth.ts
lukelee-sl: So we are using the cache only for its ttl feature.  Makes sense

lukelee-sl (approved):
LGTM

packages/relay/src/lib/clients/mirrorNodeClient.ts
ebadiere: Updated.

PR #1170, opened 2023-05-02
Edit eth_getCode cache

### LINKED ISSUE #1107: eth_getCode cache
### Description

We are caching the contract's bytecode which is acceptable in most cases because the bytecode is immutable and can not be changed. The problem appears when the user calls `selfdestruct` the the bytecode must be deleted, and if the end-user calls `eth_getCode` again, the cached response will be returned, which is not accurate and not corresponding to the state.

### Steps to reproduce

- Deploy a contract with `selfdestruct` via hardhat example
- Get the contract bytecode via `ethers.provider.getCode(<contractAddress>)` which will call `eth_getCode`
- Call the selfdestruct function
- Get the contract bytecode again - the returned value should be `0x` because the contract has been deleted

PS: to make sure that `selfdestruct` is called, you can transfer a hbars to the contract beforehand and check the balances before/after the destruction

### Solution
- [ ] Remove the cache from `eth_getCode`
- [ ] Add test

### Additional context

_No response_

### Hedera network

other

### Version

latest

### Operating system

Linux

natanasow:
**Description**:

The proposed solution is [here](https://github.com/hashgraph/hedera-json-rpc-relay/issues/1107#issuecomment-1521662616).

Currently, the bytecode is fetched via a [SDK call](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1170/files#diff-dd8c7990f7764960d05dcb4c1b5ae7eee48b425b1b83d6dcc7e7ccc442330ef0R799), once [the issue](https://github.com/hashgraph/hedera-services/issues/6248) in services is resolved, the SDK call must be replaced with a mirror node call for `deleted` property.

**Related issue(s)**:

Fixes #1107

**Notes for reviewer**:

**Checklist**


Nana-EC:
Rebase to get the `eth_estimateGas` test fixes

packages/relay/src/lib/eth.ts
```diff
@@ -782,10 +783,15 @@ export class EthImpl implements Eth {
...
           return EthImpl.redirectBytecodeAddressReplace(address);
-        }
-        else if (result?.type === constants.TYPE_CONTRACT) {
+        } else if (result?.type === constants.TYPE_CONTRACT) {
           if (result?.entity.runtime_bytecode !== EthImpl.emptyHex) {
-            return result?.entity.runtime_bytecode;
+            const prohibitedOpcodes = ['CALLCODE', 'DELEGATECALL', 'SELFDESTRUCT', 'SUICIDE'];
+            const opcodes = asm.disassemble(result?.entity.runtime_bytecode);
+            const hasProhibitedOpcode = opcodes.filter(opcode => prohibitedOpcodes.indexOf(opcode.opcode.mnemonic) > -1).length > 0;
+            if (!hasProhibitedOpcode) {
```
Nana-EC, line 791: Should we set a max size to the contracts we'd cache or would that defeat the purpose since caching those might be beneficial?

packages/server/package.json
```diff
@@ -6,6 +6,7 @@
     "keywords": [],
     "author": "Hedera Smart Contracts Team",
     "dependencies": {
+        "@ethersproject/asm": "^5.7.0",
```
Nana-EC, line 9: Q: why is this needed on the server side package?

packages/relay/src/lib/eth.ts
Nana-EC, line 791: Add a check for `runtime_bytecode === '0x'` as right now without complete traceability migration some contracts may report this but have valid code and for the mean time should still call the consensus node

Nana-EC (changes requested):
Solid changes, left some comments

packages/relay/src/lib/eth.ts
natanasow: Caching a large bytecode is beneficial and good for us. It's better to not set a max size.

packages/server/package.json
natanasow: Leftover from tests. Removed.

packages/relay/src/lib/eth.ts
natanasow: It's handled [here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1170/files/25d48c94c373db3528cf48500b3c8e46bfafee18#diff-dd8c7990f7764960d05dcb4c1b5ae7eee48b425b1b83d6dcc7e7ccc442330ef0R787), right :thinking: We don't need an additional check.

Nana-EC: Okay. Let's watch out for this though in case we need to revisit

Nana-EC: Ah good point. Okay no-op for this comment then :)

Nana-EC (approved):

PR #1181, opened 2023-05-04
Add resolve type consideration for tokens (#1158)

ebadiere:
Handle valid evm contract address considerations as part of entity resolution

- Add path to MirrorNodeClient retry log
- Conditionally add `this.getTokenById()` to `resolveEntityType()` only if address is of long zero type
- Move requestId to add of `resolveEntityType()` parameters
- Added test for to verify token is only called for long zero address

**Checklist**

### DIFF
--- a/.github/workflows/dapp.yml
+++ b/.github/workflows/dapp.yml
@@ -11,7 +11,7 @@ jobs:
   setup-local-hedera:
     name: Dapp Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 35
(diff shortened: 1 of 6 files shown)


Nana-EC (approved):

Nana-EC (changes requested):
Can close

Nana-EC (changes requested):
Dupe of https://github.com/hashgraph/hedera-json-rpc-relay/pull/1184 can be closed right?

PR #1194, opened 2023-05-06
change so we only retry on 404s if  a retry is desired

### LINKED ISSUE #1183: eth_call does not have any error status code defined
### Description

When doing calls to eth_call on mirror node, there is no accepted error codes defined.  (see [here](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/mirrorNodeClient.ts#L100))
hence Axios retrys are re-trying many times unnecessary.

### Steps to reproduce

Add the accepted error codes to the map, so it can fail faster based on the API spec of the mirror node found here:
https://testnet.mirrornode.hedera.com/api/v1/docs/#/contracts/contractsCall

### Additional context

_No response_

### Hedera network

other

### Version

0.23

### Operating system

macOS

AlfredoG87:
**Description**:
Only retries on `404` 
Also,  using the following ENV properties we should be able to change the values to stop doing retries altogether.
My recommendation is to try to set them as follows:
```
MIRROR_NODE_RETRIES = 0
MIRROR_NODE_RETRY_CODES=[]
```

and if we see there is a strong need for a single retry

```
MIRROR_NODE_RETRIES = 1
MIRROR_NODE_RETRY_DELAY = 2000
MIRROR_NODE_RETRY_CODES=[404]
```

**Related issue(s)**:

Fixes #1183

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/errors/MirrorNodeClientError.ts
```diff
@@ -23,7 +23,7 @@ export class MirrorNodeClientError extends Error {
     public statusCode: number;
     public errorMessage?: string;
 
-    static retryErrorCodes: Array<number> = [400, 404, 408, 425, 500]
+    static retryErrorCodes: Array<number> = [404]; // we should only retry on 404 errors
```
Nana-EC, line 26: This is currently only used in `MirroNodeClient.ts`.
Move it there, make it nonstatic and set it via an env config.
Then it's easy to configure whiles we explore the right UX for users.

Default can be [404] if the env variable is empty. If not parse it and set the array appropriately

Nana-EC (changes requested):

packages/relay/src/lib/errors/MirrorNodeClientError.ts
AlfredoG87: ✅  Done. 
I have added documentation for that property and also changed my recommended values to test initially as:
```
MIRROR_NODE_RETRIES = 0
MIRROR_NODE_RETRY_CODES=[]
```

Nana-EC (approved):

PR #1200, opened 2023-05-08
Removed Assertion.expectedError.

### LINKED ISSUE #1199: Remove the Assertions.expectedError() in the test relayClient callFailing method.
### Description

The Assertions.expectedError() is expecting true to equal false instead of catching the expected error.  It's causing the batch3 payment method tests to fail.

### Steps to reproduce

1. Run the acceptance:api_batch3 tests in either development or previewnet.
2. See the following tests fail:
Payable method revertWithNothing throws an error
Payable method revertWithString throws an error
Payable method revertWithCustomError throws an error
Payable method revertWithPanic throws an error

### Additional context

_No response_

### Hedera network

previewnet

### Version

v0.23.0

### Operating system

None

ebadiere:
**Description**:
Removed Assertion.expectedError in the callFailing method.  Doing so allows the proper error to be caught and the test to pass.

**Related issue(s)**:

Fixes #1199

**Checklist**

### DIFF
--- a/packages/server/tests/clients/relayClient.ts
+++ b/packages/server/tests/clients/relayClient.ts
@@ -59,7 +59,6 @@ export default class RelayClient {
         try {
             const res = await this.call(methodName, params, requestId);
             this.logger.trace(`${requestIdPrefix} [POST] to relay '${methodName}' with params [${params}] returned ${JSON.stringify(res)}`);
-            Assertions.expectedError();
         } catch (err) {
             Assertions.jsonRpcError(err, expectedRpcError);
         }


packages/server/tests/clients/relayClient.ts
```diff
@@ -59,7 +59,6 @@ export default class RelayClient {
         try {
             const res = await this.call(methodName, params, requestId);
             this.logger.trace(`${requestIdPrefix} [POST] to relay '${methodName}' with params [${params}] returned ${JSON.stringify(res)}`);
-            Assertions.expectedError();
```
Nana-EC, line 62: By removing this you're implying the method succeeds when it shouldn't no?

If the test is getting here doesn't that mean the call was not failing?
In which case the test is operating correctly and failing then the expectation is not met i.e. there's no revert error.

Nana-EC (changes requested):
Unclear why this is the right change.
If the call is succeeding when it should fail that's a bug, unless the actual eth method changed it's logic to now longer throw.

packages/server/tests/clients/relayClient.ts
ebadiere: Agreed.  Upon a second look I see that the test is not asserting the expected JsonRpcError.  I'll close this PR.

PR #1216, opened 2023-05-09
resolveEntityType error  handling  improvement  inside promises 

### LINKED ISSUE #1215: Correct handling of `504` timeouts from mirror node
### Description

`504` timeouts from mirror node are not being handled and logged correctly

### Steps to reproduce

Create a mock for   the mirror node that returns `504` instead  of the actual,  expected response.

### Additional context

_No response_

### Hedera network

other

### Version

v0.23.0

### Operating system

Other

AlfredoG87:
**Description**:
resolveEntityType uses promises, if the promises  returned an  exception,  it  was  not  being bubbled/propagated, but  instead  it  was  crashing  the  app,  the  most  common  case  is  when  the  mirror  node  returns  a  504  (timeout).  now  its  returning  null

**Related issue(s)**:

Fixes #1215 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -728,7 +728,7 @@ export class MirrorNodeClient {
         }));
 
         if (searchableTypes.find(t => t === constants.TYPE_CONTRACT)) {
-            const contract = await this.getContract(entityIdentifier, requestId);
+            const contract = await this.getContract(entityIdentifier, requestId).catch(() =>  {return null;});
             if (contract) {
                 const response = {
                     type: constants.TYPE_CONTRACT,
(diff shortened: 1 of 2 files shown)


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -728,7 +728,7 @@ export class MirrorNodeClient {
         }));
 
         if (searchableTypes.find(t => t === constants.TYPE_CONTRACT)) {
-            const contract = await this.getContract(entityIdentifier, requestId);
+            const contract = await this.getContract(entityIdentifier, requestId).catch(() =>  {return null;});
```
Nana-EC, line 731: This approach works, but do we think on promise catch for `resolveEntityType()` in `eth.ts` would satisfy all the cases?

Nana-EC: Another option, outside of this might be to rewrite this function utilizing more of a promise approach then we'd have a single catch.
This is good for now though

Nana-EC (approved):
LG.

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: Yeah, I thought and  tried bubbling  the  error and  it  works,  but  the issue  there is that when there are multiple  promises, then the  first  one  bubbles up  and  propagates  until  reaching  the  `server.ts` `logAndHandleResponse` returning  a  more  descriptive  error  message  to the user, but if  the  2nd  promise  were to throw exception as well,  the second one was being unhandled and thus also crashing,  that is why I decided to do this approach, also considering that we might need to circle back and do a refactor of the whole method.

ebadiere (approved):
LG

Issue #1222, opened 2023-05-10
Intermittent INVALID_TRANSACTION_BODY failures on eth_call


Nana-EC:
### Description

The relay is occasionally experience errors on consensus node calls with the error
`transaction <relayAccountId>@<validStartNs> failed precheck with status INVALID_TRANSACTION_BODY`.

The problem relates to contract queries and transactions, specifically the response for `ContractCallQuery`, `AccountInfoQuery` and `EthereumTransaction`. This impacts `eth_call`, `eth_getTransactionCount` submissions and some `eth_sendRawTransaction` calls.

### Steps to reproduce

Occasional observed on `eth_call`, `eth_getTransactionCount` or `eth_sendRawTransaction` calls

### Hedera network

other

### Version

v0.22.0

### Operating system

None

AlfredoG87:
@Nana-EC, I think this has been completely mitigated with fixes already implemented. I would recommend to close this ticket.

Nana-EC:
Indeed

PR #1235, opened 2023-05-10
HotFix to  re-create the sdk client every X amount of request on runt…

AlfredoG87:
**Description**:
Instead of using a singleton long live sdkClient, the EthImpl  will  create a new instance every X (configurable) amount of request.

**Related issue(s)**: 1222

Addresses #1222

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -56,6 +56,7 @@ import HbarLimit from '../hbarlimiter';
 import constants from './../constants';
 import { SDKClientError } from './../errors/SDKClientError';
 import { JsonRpcError, predefined } from './../errors/JsonRpcError';
+import {RelayImpl} from "../relay";
 
 const _ = require('lodash');
 
@@ -64,15 +65,9 @@ export class SDKClient {
     static queryMode = 'QUERY';
     static recordMode = 'RECORD';
 
-    private static chainIds = {
-        mainnet: 0x127,
-        testnet: 0x128,
-        previewnet: 0x129,
-    };
-
     static initClient(logger: Logger, hederaNetwork: string, type: string | null = null): Client {
         let client: Client;
-        if (hederaNetwork in SDKClient.chainIds) {
+        if (hederaNetwork in RelayImpl.chainIds) {
             client = Client.forName(hederaNetwork);
         } else {
             client = Client.forNetwork(JSON.parse(hederaNetwork));
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -46,7 +46,7 @@ import {
     EthereumTransactionData,
     PrecheckStatusError,
     TransactionRecordQuery,
-    Hbar,
+    Hbar, PrivateKey,
 } from '@hashgraph/sdk';
 import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
 import { Logger } from "pino";
@@ -63,6 +63,57 @@ export class SDKClient {
     static transactionMode = 'TRANSACTION';
     static queryMode = 'QUERY';
     static recordMode = 'RECORD';
+
+    private static chainIds = {
+        mainnet: 0x127,
+        testnet: 0x128,
+        previewnet: 0x129,
+    };
+
+    static initClient(logger: Logger, hederaNetwork: string, type: string | null = null): Client {
+        let client: Client;
+        if (hederaNetwork in SDKClient.chainIds) {
+            client = Client.forName(hederaNetwork);
+        } else {
+            client = Client.forNetwork(JSON.parse(hederaNetwork));
+        }
+
+        if (type === 'eth_sendRawTransaction') {
+            if (
+                process.env.OPERATOR_ID_ETH_SENDRAWTRANSACTION &&
+                process.env.OPERATOR_KEY_ETH_SENDRAWTRANSACTION
+            ) {
+                client = client.setOperator(
+                    AccountId.fromString(
+                        process.env.OPERATOR_ID_ETH_SENDRAWTRANSACTION
+                    ),
+                    PrivateKey.fromString(
+                        process.env.OPERATOR_KEY_ETH_SENDRAWTRANSACTION
+                    )
+                );
+            } else {
+                logger.warn(`Invalid 'ETH_SENDRAWTRANSACTION' env variables provided`);
+            }
+        } else {
+            if (process.env.OPERATOR_ID_MAIN && process.env.OPERATOR_KEY_MAIN) {
+                client = client.setOperator(
+                    AccountId.fromString(process.env.OPERATOR_ID_MAIN.trim()),
+                    PrivateKey.fromString(process.env.OPERATOR_KEY_MAIN)
+                );
+            } else {
+                logger.warn(`Invalid 'OPERATOR' env variables provided`);
+            }
+        }
+
+        client.setTransportSecurity(process.env.CLIENT_TRANSPORT_SECURITY === 'true' || false);
+        client.setRequestTimeout(parseInt(process.env.SDK_REQUEST_TIMEOUT || '10000'));
+
+        logger.info(`SDK client successfully configured to ${JSON.stringify(hederaNetwork)} for account ${client.operatorAccountId} with request timeout value: ${process.env.SDK_REQUEST_TIMEOUT}`);
+
+        return client;
+    }
+
+
     /**
      * The client to use for connecting to the main consensus network. The account
      * associated with this client will pay for all operations on the main network.
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -166,8 +166,6 @@ export class EthImpl implements Eth {
 
   /* Temporary code duplication for a temporary HotFix workaround while the definitive  fix is implemented
   * Please remove initClient workaround, or better yet, do not let this be merged on MAIN, short-lived fix intended only for 0.23.1 version */
-
-
   private requestsPerSdkClient = 0;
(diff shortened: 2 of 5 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -157,9 +160,40 @@ export class EthImpl implements Eth {
...
   }
 
+  /* Temporary code duplication for a temporary HotFix workaround while the definitive  fix is implemented
+  * Please remove initClient workaround, or better yet, do not let this be merged on MAIN, short-lived fix intended only for 0.23.1 version */
+
+
+  private requestsPerSdkClient = 0;
+  private maxRequestsPerSdkClient = process.env.ETH_CALL_MAX_REQUEST_PER_SDK_INSTANCE || 50;
+  private hederaNetwork: string = (process.env.HEDERA_NETWORK || '{}').toLowerCase();
+  private ethCallSdkClient: SDKClient | undefined;
```
Nana-EC, line 174: Looks like you ended up doing it for all cases. Just name it
```suggestion
  private ethSdkClient: SDKClient | undefined;
```

packages/relay/src/lib/eth.ts
```diff
@@ -149,6 +151,7 @@ export class EthImpl implements Eth {
     mirrorNodeClient: MirrorNodeClient,
     logger: Logger,
     chain: string,
+    register: Registry,
     cache?
   ) {
     this.sdkClient = nodeClient;
```
david-bakin-sl, line 157: I suggest _renaming_ this field.  That will ensure (I think? this is TS not JS so yes?) that _all_ uses of the sdk client are now going through `getSdkClient()` instead of accessing it directly.  Or not?

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -110,7 +110,7 @@ describe('Eth calls using MirrorNode', async function () {
     process.env.ETH_FEE_HISTORY_FIXED = 'false';
 
     // @ts-ignore
-    ethImpl = new EthImpl(sdkClientStub, mirrorNodeInstance, logger, '0x12a', cache);
+    ethImpl = new EthImpl(sdkClientStub, mirrorNodeInstance, logger, '0x12a', registry, cache);
```
Nana-EC, line 113: Q: what happened here? Was registry just missing or is this related to the change?

Nana-EC (commented):
Looking good, 1 nit suggestion and a question

david-bakin-sl (commented):
One comment based on my limited knowledge of TS.  And I remark: due to this being a GC language it is ok to replace the sdk client just by creating a new one.  Any uses of it that are in-flight will work fine with the instance of the sdk client they're already holding (which will get cleaned up whenever the last such use is finished).

packages/relay/tests/lib/eth.spec.ts
AlfredoG87: This is related to  the change, I had  to inject Registry for metrics in order to  be  able to instantiate the SDKClient on-demand.

packages/relay/src/lib/eth.ts
AlfredoG87: good catch, yeah, the name of the variable was defined when it was only going to be used by `eth_call` indeed. 👍 

✅  done!

lukelee-sl (commented):
Does the sdkClient in precheck.ts need to be updated to the new instance?

packages/relay/src/lib/eth.ts
Nana-EC: Is @david-bakin-sl comment more so instead of renaming a variable should we create a new instances so that existing the function removes it from use and allows it to be garbage collected?

packages/relay/src/lib/eth.ts
```diff
@@ -157,9 +160,38 @@ export class EthImpl implements Eth {
...
+    this.registry = register;
     if (!cache) this.cache = new LRU(this.options);
   }
 
+  /* Temporary code duplication for a temporary HotFix workaround while the definitive  fix is implemented
+  * Please remove initClient workaround, or better yet, do not let this be merged on MAIN, short-lived fix intended only for 0.23.1 version */
+  private requestsPerSdkClient = 0;
+  private maxRequestsPerSdkClient = process.env.ETH_CALL_MAX_REQUEST_PER_SDK_INSTANCE || 50;
+  private hederaNetwork: string = (process.env.HEDERA_NETWORK || '{}').toLowerCase();
+  private ethSdkClient: SDKClient | undefined;
```
Nana-EC, line 172: Actually, do we need this?
Maybe this is what @david-bakin-sl was referring to.
If `getSdkCLient()` is a true getter then the variable should be a local one and return in the function

Nana-EC (commented):
Question on variable assignment vs getter

packages/relay/src/lib/eth.ts
david-bakin-sl: I was reacting to [line 185](https://github.com/hashgraph/hedera-json-rpc-relay/blob/40b1bcc13d15a907f6bb7372e2fba470ffd930c7/packages/relay/src/lib/eth.ts#L185) where it looks like there's an instance variable holding the current SDK client and whenever it times out (call limit exceeded) you create a new sdk client and stomp over the old one.  But of course the old one still exists as long as anyone references it.  So calls _in flight_ still work with the old client they they're holding.  (I guess that's how the framework works, don't know - in other languages mostly a webservice call being handled has local state that disappears only when the webservice call is completed, and _outgoing_ clients would be held in that local state.)  _New_ calls coming in get the new client.  There might be a race condition where two calls come in at nearly the same time and so two new clients are created - one will win (be stored into sdkClient last) and be used for future calls, the other one will only be used for one call then thrown away.  This probably doesn't matter at all.

The point is: I think this code is fine as it is.

And to the injected dependency thing - well, have it inject to the _renamed_ variable's name, right?  Alternatively - just rename it _long enough to do a compile_ and check that it isn't used directly anywhere and let us know you confirmed that.

AlfredoG87: Thank you, good idea to rename it so the reviewer is confident there is not used anywhere where he does not see it explicitly on the changes.

I have rename  it and thanks for the insight, very interesting about the way the references work.

packages/relay/src/lib/eth.ts
AlfredoG87: We need it to live outside since we want to reuse the instance for X amount of request.
David was referencing to rename it so we can be sure as reviewer that is no longer being used outside where we want to keep on using it. (only for unit tests) so it should be visible the changes only there and on the constructor when is being injected.

AlfredoG87:
> Does the sdkClient in precheck.ts need to be updated to the new instance?

Good catch @lukelee-sl...
I think the `precheck.ts` would need to be refactored as well to instantiate its own sdkClient just as `eth.ts` was.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -56,13 +56,59 @@ import HbarLimit from '../hbarlimiter';
...
+                        process.env.OPERATOR_KEY_ETH_SENDRAWTRANSACTION
+                    )
+                );
+            } else {
+                logger.warn(`Invalid 'ETH_SENDRAWTRANSACTION' env variables provided`);
+            }
+        } else {
+            if (process.env.OPERATOR_ID_MAIN && process.env.OPERATOR_KEY_MAIN) {
+                client = client.setOperator(
+                    AccountId.fromString(process.env.OPERATOR_ID_MAIN.trim()),
```
david-bakin-sl, line 95: Of all these things you're pulling out of `env` why is this one the only one that needs a `trim()`?  (Just curious.)

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -3609,7 +3609,7 @@ describe('Eth', async function () {
   let ethImpl: EthImpl;
   this.beforeAll(() => {
     // @ts-ignore
-    ethImpl = new EthImpl(null, mirrorNodeInstance, logger);
+    ethImpl = new EthImpl(null, mirrorNodeInstance, logger, '0x12a', new Registry());
```
david-bakin-sl, line 3612: The other instantiation of `EthImpl` in this file (L113) supplies a cache argument; this one doesn't.  Why does one need the cache and the other not?  (Does the cache help the test up there?)

Nana-EC:
> > Does the sdkClient in precheck.ts need to be updated to the new instance?
> 
> 
> 
> Good catch @lukelee-sl...
> 
> I think the `precheck.ts` would need to be refactored as well to instantiate its own sdkClient just as `eth.ts` was.
> 
> 
> 
> @Nana-EC  should we also include it as part of  this PR? 
> 
> The only place where is used is [here](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/precheck.ts#L153) 
> 
> and is a failover in case there are issues getting the balance from the mirrorNode, and after this, it  will be the only place where a  long lived instance "could" be used, so it might  not be an issue...
> 
> 
> 
> Am saying this since I think that the refactor would take me longer to do and I don't think to be able to do it today, I could do it tomorrow morning.
> 
> 

It's okay leave for another PR.
This was supposed to be a partial attempt to alleviate pressure.

Nana-EC (approved):

packages/relay/tests/lib/eth.spec.ts
AlfredoG87: The  cache is always used, is just that if you don't inject the cache, the constructor is going to instantiate one instance only for this class.

image

packages/relay/src/lib/clients/sdkClient.ts
AlfredoG87: Good question, and the honest answer is I don't know. these code looks new, but is in fact "moved" from one class to another. so I only kept it as I found it  😅

PR #1241, opened 2023-05-12
Add SDK Client Service

### LINKED ISSUE #1197: Add logic to renew sdk client instance
### Problem

The relay utilizes the js SDK client in a  long lived manner submitting multiple queries and transactions over time.
This has performance consideration over time

### Solution

Add logic to renew the instance of the client every so often to ensure freshness.

Consider both timed options e.g. every 2hrs or a counter of transactions e.g. every 100k.
Both if considered separately or combined should be dynamic configurations that can be set

Notably this may require some coordination between `relay.ts` (where it's created), `eth.ts` (where it's used) and `sdkClient.ts` (where it's exposed)

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR adds new client service, which takes care of initialisation of the needed SDK clients. This service utilizes new feature which takes care of refreshing the SDK client (reinitialising) after X number of transaction, Y time passed or Z number of errors occured. Reinitialisation happens when one of those are hit. 
To enable this feature, we need to set at least one of the variables. (by default if we don't set them, they'll be equal to 0)
To disable this feature, set **all** variables to 0 or don't set them at all.

All those properties are configurable via environmental variables.

Moved the HBAR Limiter to be global and one instance for all new instances of the clients. Added more tests for this behaviour.

**Related issue(s)**:

Fixes #1197 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/services/clientService/index.ts
```diff
@@ -0,0 +1,276 @@
...
+
+  /**
+   * The logger used for logging all output from this class.
+   * @private
+   */
+  private logger: Logger;
+  private readonly register: Registry;
+  private clientResetCounter: Counter;
+
+  private chainIds = {
```
Nana-EC, line 55: Move to constants or make public static

packages/relay/src/lib/services/clientService/index.ts
```diff
@@ -0,0 +1,276 @@
...
+import { Registry, Counter } from 'prom-client';
+import { SDKClient } from '../../clients/sdkClient';
+
+export default class ClientService {
+  private transactionCount: number;
+  private errorCount: number;
+  private resetDuration: number;
+  private shouldReset: boolean;
+
+  private isEnabled: boolean;
```
Nana-EC, line 34: Q: What does this enable or disable? The whole SDK usage or just the batch reinitializations?

packages/relay/src/lib/services/clientService/index.ts
```diff
@@ -0,0 +1,276 @@
...
+  /**
+   *  Decrement transaction counter. If 0 is reached, reset the client. Check also if resetDuration has been reached and reset the client, if yes.
+   */
+  private decrementTransactionCounter() {
+    if (this.transactionCount == 0) {
+      return;
+    }
+
+    this.transactionCount--;
+    if (this.transactionCount <= 0) {
```
Nana-EC, line 107: Q: Is this line possible if the first line of the method prevent decrement in the 0 case?

packages/relay/src/lib/services/clientService/index.ts
```diff
@@ -0,0 +1,276 @@
...
+      help: 'Relay Client Service',
+      registers: [register],
+      labelNames: ['transactions', 'duration', 'errors'],
+    });
+  }
+
+  /**
+   *  Decrement transaction counter. If 0 is reached, reset the client. Check also if resetDuration has been reached and reset the client, if yes.
+   */
+  private decrementTransactionCounter() {
```
Nana-EC, line 101: Q: Does this method need a check for isEnabled?

packages/relay/src/lib/services/clientService/index.ts
```diff
@@ -0,0 +1,276 @@
...
+
+  /**
+   * Reset the main client, SDK Client and reset all counters.
+   */
+  private resetClient() {
+    this.clientResetCounter
+      .labels(this.transactionCount.toString(), this.resetDuration.toString(), this.errorCount.toString())
+      .inc(1);
+
+    const hederaNetwork: string = (process.env.HEDERA_NETWORK || '{}').toLowerCase();
```
Nana-EC, line 144: Network will never change per instance so we should cache this early once and resume here. No need to pull from env variable again

packages/relay/src/lib/services/clientService/index.ts
```diff
@@ -0,0 +1,276 @@
...
+    this.clientMain = this.initClient(this.logger, hederaNetwork);
+    this.client = this.initSDKClient(this.logger, this.register);
+
+    this.resetCounters();
+  }
+
+  /**
+   * Reset all counters with predefined configuration.
+   */
+  private resetCounters() {
```
Nana-EC, line 154: This method should deal with process envs. We should extract those all at once at the beginning

packages/relay/src/lib/services/clientService/index.ts
```diff
@@ -0,0 +1,276 @@
...
+   */
+  public getMainClient() {
+    return this.clientMain;
+  }
+
+  /**
+   * Return configured sdk client
+   * @returns SDK Client
+   */
+  public getSDKClient(): SDKClient {
```
Nana-EC, line 231: Should have a more informative function name.
Thinking on it myself but something suggests the client is managed in instances.

packages/relay/src/lib/services/clientService/index.ts
georgi-l95: Just batch reinitialisation. Maybe I should rename it to something like `isReinitEnabled` or something similar ?

packages/relay/src/lib/services/clientService/index.ts
georgi-l95: I wrote first this and later the upper part. I can remove the equal sign, but the less sign is needed.

packages/relay/src/lib/services/clientService/index.ts
georgi-l95: You mean it shouldn't deal with process env ?

packages/relay/src/lib/services/clientService/index.ts
georgi-l95: You're right, although there is a option to disable this whole reinitialisation feature and it's a valid name then. I'm open to suggestions

packages/relay/src/lib/services/clientService/index.ts
georgi-l95: No need, because it's a private method and we use it only when before returning the sdk client in another function and on the top of that function we already check isEnabled.

packages/relay/src/lib/services/clientService/index.ts
Nana-EC: Yeah shouldn't :)

packages/relay/src/lib/services/clientService/index.ts
Nana-EC: Hmm good point.
Okay how about this, leave this as `getSDKClient()` and change the other to `getSingletonSDKClient()`.
SO maybe the emphasis is on the fact that the other is the instance that is long lived across the application.

packages/relay/src/lib/eth.ts
```diff
@@ -146,17 +146,17 @@ export class EthImpl implements Eth {
...
   constructor(
-    nodeClient: SDKClient,
+    clientSevice: ClientService,
     mirrorNodeClient: MirrorNodeClient,
     logger: Logger,
     chain: string,
     cache?
   ) {
-    this.sdkClient = nodeClient;
+    this.clientService = clientSevice;
```
AlfredoG87, line 155: nit: typo `clientSevice` should be `clientService`

packages/relay/src/lib/services/clientService/index.ts
```diff
@@ -0,0 +1,276 @@
...
+  /**
+   * Reset the main client, SDK Client and reset all counters.
+   */
+  private resetClient() {
+    this.clientResetCounter
+      .labels(this.transactionCount.toString(), this.resetDuration.toString(), this.errorCount.toString())
+      .inc(1);
+
+    const hederaNetwork: string = (process.env.HEDERA_NETWORK || '{}').toLowerCase();
+    this.clientMain = this.initClient(this.logger, hederaNetwork);
```
AlfredoG87, line 145: I think is the same case `this.clientMain` I think we should not reInit it every time we reset the sdkClient, the `Client` only has configuration about the network, and it won't change. we should not `re-init` the client.

docs/configuration.md
```diff
@@ -42,8 +42,11 @@ The following table lists the available properties along with their default valu
...
+| `CLIENT_TRANSACTION_RESET`            | "50"          | Number of transaction executions, until client reinitialization.                                                                                                                                                                                                                                       |
+| `CLIENT_DURATION_RESET`               | "3600000"     | Time until client reinitialization. (ms)                                                                                                                                                                                                                                                               |
+| `CLIENT_ERROR_RESET`                  | "100"         | Number of error to be reached, until client reinitialization.                                                                                                                                                                                                                                          |
```
AlfredoG87, line 49: I am wondering why would we need to tolerate `100` errors before re-init. I would set this to `1`, since once we start getting errors, we should recycle/reset the sdkClient instance, and if we were to set it to `1`. why would we need this property and  logic  around  it on the first place?

packages/relay/src/lib/services/clientService/index.ts
AlfredoG87: and  also  refactor `relay.eth`  to  use the constant.
https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/relay.ts#L35

AlfredoG87 (commented):
LG, leave  some nits and a question.

AlfredoG87:
Should this PR  also address or Fix this ticket?
https://github.com/hashgraph/hedera-json-rpc-relay/issues/1222

docs/configuration.md
georgi-l95: All those values are just suggestions from me, I should've wrote in the description that I will appreciate any suggestions for more appropriate values. Also yeah 100 might be a lot, but error could also be `UNKNOWN` or `CONTRACT REVERT` or something else. Or should I just limit this check only for `INVALID_TRANSACTION_BODY` ?

georgi-l95:
> Should this PR also address or Fix this ticket? #1222

I think more like address and workaround this issue. But I'm suggesting we still search and try to fix the root cause of this error

docs/configuration.md
AlfredoG87: My suggestion: we might want to keep the `CLIENT_ERROR_RESET` property for all other errors but leave a hardcoded check exclusively for `INVALID_TRANSACTION_BODY` and as soon as we identify `1` occurrence of this, refresh the sdkClient instance.

Why? over the past 2 days our initial workaround has been live, I've noticed that there are some occurrences of `INVALID_TRANSACTION_BODY` and when they come, they come in group, right after another, presumably until the next refresh happens:


image

georgi-l95: Okay that make sense. Will do it. Thank you for the suggestion

Nana-EC: Going with this approach of reset on 1, I'd say instead of a number, the config value should be a list of HAPI response types, if the error is a match then reset. In this way should find another error type that requires a reinitialize we can add it to the list

packages/relay/src/lib/services/clientService/index.ts
georgi-l95: Added to constants.

packages/relay/src/lib/services/clientService/index.ts
georgi-l95: Renamed.

packages/relay/src/lib/services/clientService/index.ts
georgi-l95: Removed.

packages/relay/src/lib/services/clientService/index.ts
georgi-l95: Done.

packages/relay/src/lib/services/clientService/index.ts
georgi-l95: Renamed.

packages/relay/src/lib/eth.ts
georgi-l95: Fixed.

packages/relay/src/lib/services/clientService/index.ts
```diff
@@ -0,0 +1,276 @@
...
+ */
+
+import dotenv from 'dotenv';
+import findConfig from 'find-config';
+import { AccountId, Client, PrivateKey } from '@hashgraph/sdk';
+import { Logger } from 'pino';
+import { Registry, Counter } from 'prom-client';
+import { SDKClient } from '../../clients/sdkClient';
+
+export default class ClientService {
```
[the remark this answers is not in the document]
georgi-l95: Renamed.

.env.example
```diff
@@ -43,3 +43,6 @@ ETH_BLOCK_NUMBER_CACHE_TTL_MS=
 ETH_GET_BALANCE_CACHE_TTL_MS=
 ETH_FEE_HISTORY_FIXED=
 WEB_SOCKET_METRICS_PORT=
+HAPI_CLIENT_TRANSACTION_RESET=
+HAPI_LIENT_DURATION_RESET=
```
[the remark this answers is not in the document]
georgi-l95: Fixed.

packages/relay/tests/test.env
```diff
@@ -5,4 +5,7 @@ MIRROR_NODE_LIMIT_PARAM = 100
 HEDERA_NETWORK={"127.0.0.1:50211":"0.0.3"}
 OPERATOR_ID_MAIN=0.0.2
 OPERATOR_KEY_MAIN=302e020100300506032b65700422042091132178e72057a1d7528025956fe39b0b847f200ab59b2fdd367017f3087137
-WS_POLLING_INTERVAL=500
\ No newline at end of file
+WS_POLLING_INTERVAL=500
+HAPI_CLIENT_TRANSACTION_RESET= 50
+HAPI_CLIENT_DURATION_RESET= 3600000
+HAPI_CLIENT_ERROR_RESET= [50]
```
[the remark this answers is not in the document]
georgi-l95: Added.

packages/server/tests/localAcceptance.env
```diff
@@ -39,3 +39,6 @@ WEB_SOCKET_PORT=8546
 CONNECTION_LIMIT=10
 WS_MULTIPLE_ADDRESSES_ENABLED=false
 WS_MAX_CONNECTION_TTL=300000
+HAPI_CLIENT_TRANSACTION_RESET= 50
+HAPI_CLIENT_DURATION_RESET= 3600000
+HAPI_CLIENT_ERROR_RESET= [50]
```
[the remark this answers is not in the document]
georgi-l95: Added.

AlfredoG87 (changes requested):
We  found an  issue   regarding  `HBarLimiter`  on version  `0.23.1` that needed a fix,  basically a  fix  for  the  fix  that  this  PR  addresses as  well,  so  the  issue I believe would  happen here  too.  
Leaving here  the  PR  for  context  and  details.
https://github.com/hashgraph/hedera-json-rpc-relay/pull/1249

packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  /**
+   * Configure SDK Client from main client
+   * @param {Logger} logger
+   * @param {Registry} register
+   * @returns SDK Client
+   */
+  private initSDKClient(logger: Logger, register: Registry): SDKClient {
+    return new SDKClient(this.clientMain, logger.child({ name: `consensus-node` }), register);
```
Nana-EC, line 167: You'll wanna consider the recent but we just fixed https://github.com/hashgraph/hedera-json-rpc-relay/pull/1249

Nana-EC (commented):
Looks good, just need to adopt fix for 0.23.2 - https://github.com/hashgraph/hedera-json-rpc-relay/pull/1249

AlfredoG87 (changes requested):
LGTM,  I  see  you moved the  HBar  limiter  as  a global  intance  of the  hapi  client and  pass   it  as  a  parameter  when creating  the sdkClient,  also reverted to re-init the main client, where the issue is happening.

The only thing that you did not include is the code around the metrics, since everytime we refresh the sdkClient the metrics are going to be reseted due to the 
`register.removeSingleMetric(metricHistogramCost);`
But we cannot remove that line otherwise you will get an error telling you that that metric already exists in the registry.
Is there a reason you did not? you think is not needed to treat the metric counters as singletons? 
https://github.com/hashgraph/hedera-json-rpc-relay/pull/1249/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR161

georgi-l95:
@AlfredoG87 Not intentional, just simply forgot about this metric, will fix it and request review again. Thank you for noticing.

packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    if (this.transactionCount === 0 && this.errorCodes.length === 0 && this.isTimeResetDisabled) {
+      this.isReinitEnabled = false;
+    }
+    this.shouldReset = false;
+
+    this.register = register;
+    const metricCounterName = 'rpc_relay_client_service';
+    register.removeSingleMetric(metricCounterName);
+    this.clientResetCounter = new Counter({
```
Nana-EC, line 102: Q: should this metric also be a constructor param to work around the metric reset Fredy noted?

packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -0,0 +1,284 @@
...
+   */
+  private hbarLimiter: HbarLimit;
+  private readonly register: Registry;
+  private clientResetCounter: Counter;
+
+  /**
+   * @param {Logger} logger
+   * @param {Registry} register
+   */
+  constructor(logger: Logger, register: Registry, hbarLimiter: HbarLimit) {
```
Nana-EC, line 72: Maybe?
```suggestion
  constructor(logger: Logger, clientResetCounter: Counter, hbarLimiter: HbarLimit) {
```

packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -0,0 +1,284 @@
...
+      registers: [register],
+      labelNames: ['transactions', 'duration', 'errors'],
+    });
+  }
+
+  /**
+   *  Decrement transaction counter. If 0 is reached, reset the client. Check also if resetDuration has been reached and reset the client, if yes.
+   */
+  private decrementTransactionCounter() {
+    if (this.transactionCount == 0) {
```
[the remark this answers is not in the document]
georgi-l95: My logic here is that if transactionCount is zero, then either it's time to reset and there is no need to continue with the rest of the code in this method or whole feature is disabled, so there's again no need to continue with the method. While if we use only `this.isReinitEnabled` only one of those cases is covered.

packages/relay/src/lib/services/hapiService/hapiService.ts
georgi-l95: No need, I've created this metric only to keep track of the SDK reinitializations. It's not used in the SDK client class.

packages/relay/src/lib/services/hapiService/hapiService.ts
georgi-l95: I think we need the Registry contrainer, because it's one instances passed around in the whole app, coming from relay class. This counter is something i've added. And it's not present in the relay class, where this object is initialized and relay class doesn't need to know about it.

packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -0,0 +1,361 @@
...
+        registers: [register],
+        async collect() {
+            // Invoked when the registry collects its metrics' values.
+            // Allows for updated account balance tracking
+            try {
+                const accountBalance = await (new AccountBalanceQuery()
+                    .setAccountId(clientMain.operatorAccountId!))
+                    .execute(clientMain);
+                this.labels({ 'accountId': clientMain.operatorAccountId!.toString() })
+                    .set(accountBalance.hbars.toTinybars().toNumber());
```
AlfredoG87, line 354: Since we are refactoring this I would like  to  suggest to fetch accountBalance of the  operator from the mirror node? since getting  it from the the consensus node costs money?
Also,  the sdkClient does  not even use this  `Gauge` so no need  to  pass it along with the other metrics... not even sure, if could be moved to the `Relay.ts` itself so to not bring mirror node dependencies to `hapiService.ts` at all.

georgi-l95: Hmm, where is this gauge used? I'm not sure if `Relay.ts` also shouldn't know and take care about metrics, especially in the constructor. 
Agree that we should fetch balance from mirror node, if possible and not burden this service with mirror node details.
Since we have quite a lot of metrics now, I'm thinking about a metrics service, which takes care of all metric creations and assignments, what do you think?

georgi-l95: I've moved it over in the relay class, but only temporary, until I'm ready with the metrics service class.

AlfredoG87 (approved):
LGTM 👍 
Thank you!

Nana-EC (approved):
LGTM

PR #1242, opened 2023-05-12
feat(ci): Release the helm chart to gh pages

josedev-union:
This PR is to automate the helm chart release. Now the latest version is `v0.23.0` but the chart latest version is `0.17.1` which was released manually.
Changed the chart path to `charts/hedera-json-prc-relay`.

Before the merge, need to check the followings;
- A GitHub branch called gh-pages to store the published charts.
- In repo Settings/Pages, change the Source Branch to gh-pages.

**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


.github/workflows/release-production.yml
```diff
@@ -64,3 +64,28 @@ jobs:
...
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v3
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.5.0
+        with:
+          charts_dir: charts/
```
Nana-EC, line 89: With your rename of the main charts folder I don't think this would be needed

Nana-EC (commented):
Thanks @josedev-union for the contribution.

We have an existing effort which like you pointed out has been manual and we've wanted to automate it so appreciate the push here.
Give us a bit to make sure this agrees with the automated plan we have and if we can move forward with your suggestion as is.

Thanks

.github/workflows/release-production.yml
josedev-union: @Nana-EC 
Yes, the renaming of the charts folder is not necessary. But need to keep chart manifests in one more deeper path, i.e. `helm-chart/xx/` directory to use chart releaser action recommended by helm https://helm.sh/docs/howto/chart_releaser_action/
https://github.com/helm/chart-releaser-action/blob/main/cr.sh#L268

If we cannot change the chart path depth, it means we cannot use this action and rather need to use low level script.

But i suggest to have one more depth for chart path and use this upstream action.
What is your opinion?

josedev-union: @Nana-EC sorry, i was confused a bit.
Yes, definetly `charts/` is the default value so no need to specify explicitly
I will remove that

Nana-EC (commented):
Lg.
For some reason your last change broke it.
Maybe you can put `--charts=charts/hedera-json-rpc-relay` back.
Might be since charts will contain multiple future charts so the clarification is needed.

Also you commits seems to be unverified.

josedev-union:
> Lg. For some reason your last change broke it. Maybe you can put `--charts=charts/hedera-json-rpc-relay` back. Might be since charts will contain multiple future charts so the clarification is needed.
> 
> Also you commits seems to be unverified. Please see the contributing guide https://github.com/hashgraph/hedera-json-rpc-relay#contributing

I think that is related to changed chart detection. 
If `--charts` is specified, changed chart detection is disabled and the specified chart is used for the ct action.
I added a condition to check chart change in lint job.
Also `--all` also disables changed chart detection so it will install all charts alywas.

@Nana-EC btw, i accidently force-pushed without any changes because of missing gpg sign key on my local.

PR #1243, opened 2023-05-12
feat(chart): Support an existing secret and configmap

### LINKED ISSUE #1257: Add support for existing secret and configmap in Helm chart
### Problem

The current helm chart does not support existing external secret and configmaps

### Solution

Add support for existing secret and configmap in Helm chart

### Alternatives

_No response_

josedev-union:
**Description**:
This PR is to support the existing secret or configmap for the relay configuration.
The typical usecase is when the sensitive data is stored in the vault software and the k8s secret is generated by external-secret operator.

**Related issue(s)**:

Fixes #1257 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/helm-chart/templates/deployment.yaml
+++ b/helm-chart/templates/deployment.yaml
@@ -38,7 +38,10 @@ spec:
         - configMapRef:
             name: {{ include "json-rpc-relay.fullname" . }}
         - secretRef:
-            name: {{ include "json-rpc-relay.fullname" . }}  
+            name: {{ include "json-rpc-relay.fullname" . }}
+        {{- with .Values.extraEnvFrom }}
+          {{- tpl (toYaml .) $ | nindent 8 }}
+        {{- end }}
         ports:
           - containerPort: {{ .Values.ports.containerPort }}
             name: {{ .Values.ports.name  }}
--- a/helm-chart/templates/secret.yaml
+++ b/helm-chart/templates/secret.yaml
@@ -10,4 +10,3 @@ stringData:
   OPERATOR_KEY_MAIN: {{ .Values.config.OPERATOR_KEY_MAIN | quote }}
   OPERATOR_ID_ETH_SENDRAWTRANSACTION: {{ .Values.config.OPERATOR_ID_ETH_SENDRAWTRANSACTION | default (printf "%q" "") }}
   OPERATOR_KEY_ETH_SENDRAWTRANSACTION: {{ .Values.config.OPERATOR_KEY_ETH_SENDRAWTRANSACTION | default (printf "%q" "") }}  
-  
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -46,6 +46,11 @@ config:
   TIER_3_RATE_LIMIT: 400
   WEB_SOCKET_PORT: 8546
 
+# -- Extra environment variables from existing secrets or configmaps
+extraEnvFrom: []
+# - secretRef:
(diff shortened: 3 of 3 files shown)


Nana-EC (approved):

helm-chart/templates/deployment.yaml
```diff
@@ -38,7 +38,11 @@ spec:
         - configMapRef:
             name: {{ include "json-rpc-relay.fullname" . }}
         - secretRef:
+        {{- if .Values.existingSecret }}
+            name: {{ .Values.existingSecret }}
```
steven-sheehy, line 42: DRY:

```suggestion
        {{- with .Values.existingSecret }}
            name: {{ . }}
```

steven-sheehy, line 42: It should probably be templated so you can embed the release name in it to support multiple deploys in a single cluster.

helm-chart/templates/deployment.yaml
```diff
@@ -38,7 +38,11 @@ spec:
...
+        {{- else }}
             name: {{ include "json-rpc-relay.fullname" . }}  
+        {{- end }}
```
steven-sheehy, line 45: Is this really the best approach to take? We use a more simple, but customizable approach on the mirror node where we have a single [envFrom](https://github.com/hashgraph/hedera-mirror-node/blob/37a98829aa97872421435a3c7d6c92df603766de/charts/hedera-mirror-web3/values.yaml#L42) that users can add custom secrets or configmaps with templating support. So relay could do the same and have the existing secret and configmap moved to values.yaml and allow you to override it for your purposes.

Nana-EC (approved):

helm-chart/templates/deployment.yaml
```diff
@@ -38,7 +38,10 @@ spec:
         - configMapRef:
             name: {{ include "json-rpc-relay.fullname" . }}
         - secretRef:
-            name: {{ include "json-rpc-relay.fullname" . }}  
+            name: {{ include "json-rpc-relay.fullname" . }}
+        {{- with .Values.extraEnvFrom }}
```
Nana-EC, line 42: nit: envFrom would be simpler to conform with Mirror Node actually
```suggestion
        {{- with .Values.envFrom }}
```

Nana-EC, line 42: Also not sure if you need the with but could also simplify with below. Again from mirror node
```suggestion
        envFrom: {{ tpl (toYaml .Values.envFrom) . | nindent 12 }}
```

Nana-EC (changes requested):
LG, nit on naming to make the charts more consistent across the org

helm-chart/templates/deployment.yaml
josedev-union: Well, to make values identical across org charts as much as possible  is a good point. Just my concern is that `envFrom` is strictly used to replace the `envFrom` of a container spec, but we have 2 items for `envFrom` in default and users will just add more items using this value. So i think `extra` would be more reasonable and meaningful.
In other charts, we can also use `envFrom` and `extraEnvFrom` according to their manifest status.

Nana-EC: Makes sense. Thanks for clarifying

Nana-EC: Based on above I guess this also doesn't apply.

Nana-EC (approved):

PR #1245, opened 2023-05-13
Add clientCache to centralize cache

### LINKED ISSUE #1116: Refactor LRU caching logic
### Problem

The relay utilizes an LRU to manage request responses.
However the logic is distributed amongst all the methods in `eth.ts` and could find better use in place like `MirrorNodeClient`.
Each method created the key structure, does a get and set and null check

### Solution

Instead of every method handling logic

- Create a shared Cache class than had multiple method to handle the get, set and validation logic
- The `relay.ts` can create the cache and pass it to the needed classes.

### Alternatives

_No response_

Nana-EC:
**Description**:
Centralize caching logic for client calls

- Add a `clientCache class that encapsulates LRU implementation
- Update mirrornode and sdkclient calls
- create instance in relay and pass it along
- add metrics to capture cache hits to observe

**Related issue(s)**:

Fixes #1116 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -134,6 +134,7 @@ export class MirrorNodeClient {
 
     private mirrorResponseHistogram;
     private operatorAccountGauge;
+    private operatorAccount;
 
     private readonly cache: ClientCache;
     static readonly EVM_ADDRESS_REGEX: RegExp = /\/accounts\/([\d\.]+)/;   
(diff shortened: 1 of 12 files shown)


ebadiere (commented):
LG.

AlfredoG87 (dismissed):
LGTM

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -239,8 +242,26 @@ export class MirrorNodeClient {
...
+        register.removeSingleMetric(metricGaugeName);
+        this.operatorAccountGauge = new Gauge({
+            name: metricGaugeName,
+            help: 'Relay operator balance gauge',
+            labelNames: ['mode', 'type', 'accountId'],
+            registers: [register],
+            async collect() {
+                operatororBalanceCollect();
+            },
+        });
```
AlfredoG87, line 261: Just one question though. 

Why are we having this here? and what is the diff with the one on `relay.ts` also on the one on the relay after getting the acount there is a check `?` before using the value, since it could lead to issues.
```const accountBalance = account.balance?.balance;```

https://github.com/hashgraph/hedera-json-rpc-relay/blob/1116-lru-refactor/packages/relay/src/lib/relay.ts#L101-L122

Nana-EC: I think the goal was to centralize mirrorNodeClient usage. In this case i should have removed it from `relay.ts`
I think i'll leave it there

AlfredoG87 (approved):

PR #1252, opened 2023-05-16
Fix SDK reinitialization 

georgi-l95:
**Description**:
Fix SDK reinitialization. In our case SDK Client class is just a wrapper for Client class.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -455,6 +455,10 @@ export class SDKClient {
             if (sdkClientError.isGrpcTimeout()) {
                 throw predefined.REQUEST_TIMEOUT;
             }
+
+            if (sdkClientError.isInvalidTransactionBody()) {
```
Nana-EC, line 459: Add a `TRACE_LOG_INVALID_TRANSACTION_BODY` env variable that can be set to true but it defaulted to false.
If true then log

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -455,6 +455,10 @@ export class SDKClient {
...
+                this.logger.error(`${requestIdPrefix} Query object: ${query} Client object: ${client}`);
```
Nana-EC, line 460: This will only end up logging `Query object: [object Object] Client object: [object Object]`.
Further more due to the recursive nature of `client` it would actually fail.
Instead do
```suggestion
                this.logger.error(`${requestIdPrefix} Query object:  ${JSON.stringify(query)}`);
                this.logger.error(`${requestIdPrefix} Client object: ${JSON.stringify({...client, _timer: undefined})}`);
```

I tested this locally with success. Please confirm

Nana-EC (changes requested):
Fix looks good.
Log needs to be tweaked

packages/relay/src/lib/clients/sdkClient.ts
georgi-l95: Me too, works like this.

Nana-EC (approved):
LG

PR #1253, opened 2023-05-16
feat(ci): Release the helm chart to gh pages

josedev-union:
This PR is to automate the helm chart release. Now the latest version is `v0.23.0` but the chart latest version is `0.17.1` which was released manually.
Changed the chart path to `charts/hedera-json-prc-relay`.

Before the merge, need to check the followings;
- A GitHub branch called gh-pages to store the published charts.
- In repo Settings/Pages, change the Source Branch to gh-pages.

**Description**:
- Add a `helm-chart-release` job to the github `releaseproduction.yml` workflow that would automate chart creation on releases
- Move current chart to `charts/hedera-json-rpc-relay`
- Update github action `chart.yml` workflow to use the `--all` filter to manage new layout

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


ct.yaml
isavov, line 1: This file and `.github\ct.yaml` are duplicates

.github/workflows/charts.yml
```diff
@@ -18,7 +18,11 @@ jobs:
         uses: helm/chart-testing-action@v2.4.0
 
       - name: Run lint
-        run: ct lint --config .github/ct.yaml --charts=helm-chart
+        run: |
```
Nana-EC, line 21: Did we try the following and that didn't work?
```suggestion
        run: ct lint --config .github/ct.yaml --all
```

Nana-EC (commented):
LG, just waiting on linting fix

Nana-EC:
> Before the merge, need to check the followings;
> 
> A GitHub branch called gh-pages to store the published charts

@josedev-union is the branch name configurable? Since we do already have a branch for deploying charts to.

josedev-union:
> > Before the merge, need to check the followings;
> > A GitHub branch called gh-pages to store the published charts
> 
> @josedev-union is the branch name configurable? Since we do already have a branch for deploying charts to.

yes, i think so
https://github.com/helm/chart-releaser/tree/3060e0ea24dd7403f3bebd7a126d342629a74f4a#create-the-repository-index-from-github-releases
We can set `--pages-branch` for `cr by using config file or env var `CR_PAGES_BRANCH`. And also need to configure in settings>pages

Nana-EC:
> > > Before the merge, need to check the followings;
> > > A GitHub branch called gh-pages to store the published charts
> > 
> > 
> > @josedev-union is the branch name configurable? Since we do already have a branch for deploying charts to.
> 
> yes, i think so https://github.com/helm/chart-releaser/tree/3060e0ea24dd7403f3bebd7a126d342629a74f4a#create-the-repository-index-from-github-releases We can set `--pages-branch` for `cr by using config file or env var `CR_PAGES_BRANCH`. And also need to configure in settings>pages

Sounds good. We want to move to `gh-pages` but might not do it immediately and I don't want to block your PR.
Might need one more manual push and then enable this once it goes in

Nana-EC (commented):
LGTM, just waiting on getting the default branch and other items that were manual setup and then we can merge this. 
Thanks @josedev-union

.github/workflows/release-production.yml
```diff
@@ -64,3 +64,26 @@ jobs:
           name: ${{ env.PACKAGE_NAME }}-v${{env.TAG}}
           path: ./*.tgz
           if-no-files-found: error
+
+  helm-chart-release:
+    runs-on: ubuntu-latest
```
Nana-EC, line 69: Could you add a condition to make the job not run until we complete moving branches.
I want you PR to be able to go in without you getting more conflicts with other changes.
The multi chart structure in particular is a very good contribution I don't want to hold up.
```suggestion
    if: false
    runs-on: ubuntu-latest
```

Nana-EC (commented):
@josedev-union one additional to make the job not run just yet, but let's get everything else in.
Thanks

Nana-EC:
@josedev-union I think you need to rebase as you have some merge conflicts. Once you do that and CI passes I can approve and merge this

Nana-EC (approved):
LGTM

PR #1255, opened 2023-05-16
Fix flaky acceptance tests

### LINKED ISSUE #1224: Fix Flaky acceptance tests
### Description

Currently we have acceptance tests in the relay, but they fail occasionally. We need to make sure that tests are stable and we can run them against CI, testnet and previewnet.

### Steps to reproduce

1. Run tests against testnet/previewnet.

### Additional context

_No response_

### Hedera network

testnet, previewnet, other

### Version

latest

### Operating system

None

Ivo-Yankov:
- Makes acceptance tests more consistent
- Fixes `INVALID_FULL_PREFIX_SIGNATURE_FOR_PRECOMPILE`
- enables tests to run successfully against testnet and previewnet
- as a side effect increases the total HBAR cost of running all acceptacne tests
- bumps the local node version
- bumps the versions of images used in acceptance tests to `0.38.0` for services and `0.81.0-rc1` for mirror node

**Related issue(s)**:

Fixes #1224 

**Notes for reviewer**:

**Checklist**


packages/server/tests/clients/servicesClient.ts
```diff
@@ -264,13 +254,66 @@ export default class ServicesClient {
...
             servicesClient,
             privateKey,
             wallet
         );
+    }
+
+    // Creates an account that has 2 keys - ECDSA and a contractId. This is required for calling contract methods that create HTS tokens.
+    // The contractId should be the id of the contract.
+    // The account should be created after the contract has been deployed.
+    async createAccountWithContractIdKey(contractIdString: string, initialBalance = 10, provider = null, requestId?: string) {
```
Nana-EC, line 269: I think the better approach would be to have the token.key hold the contract ID
e.g. to authorize Contract A to freeze/unfreeze Token T, then T.freezeKey should be set to ContractID A at token creation.

Nana-EC (changes requested):
Nice improvements.
I would prefer the ContractID on the Token instead. If that will be a large lift then let's consider an immediate follow up PR to resolve that, but if doable let's update this PR with that approach.

packages/server/tests/localAcceptance.env
```diff
@@ -15,30 +15,32 @@ IS_TESTNET=false
...
-HBAR_RATE_LIMIT_DURATION = 80000
-RATE_LIMIT_DISABLED = false
-DEV_MODE = false
-MIRROR_NODE_RETRIES = 3
-MIRROR_NODE_RETRY_DELAY = 250
-MIRROR_NODE_RETRIES_DEVMODE = 5
-MIRROR_NODE_RETRY_DELAY_DEVMODE = 200
-GAS_PRICE_TINY_BAR_BUFFER = 10000000000
-MIRROR_NODE_LIMIT_PARAM = 2
+DEFAULT_RATE_LIMIT=200
```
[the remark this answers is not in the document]
Ivo-Yankov: Many of those env vars are marked as mandatory in the code. We should probably address this in a separate PR

packages/server/tests/clients/servicesClient.ts
Nana-EC: Agreed in follow up PR to also utilize T.freezeKey = ContractID for increased coverage

Nana-EC (approved):

ebadiere (commented):
LG

AlfredoG87 (approved):
LG

ebadiere (approved):

Nana-EC (approved):

ebadiere (commented):
LG

ebadiere (approved):

PR #1262, opened 2023-05-17
936 Added a max block number range.

### LINKED ISSUE #936: Add a max block number range check

ebadiere:
Signed-off-by: ebadiere <ebadiere@gmail.com>

Added a max block number range.

**Related issue(s)**:

Fixes #936


packages/relay/tests/lib/eth.spec.ts
```diff
@@ -651,7 +670,7 @@ describe('Eth calls using MirrorNode', async function () {
   });
 
 
-  it('"eth_blockNumber" should return the latest block number using cache', async function () {
+  it.only('"eth_blockNumber" should return the latest block number using cache', async function () {
```
AlfredoG87, line 673: Why this only?

AlfredoG87 (dismissed):
LG, Only leaved a question.

packages/relay/tests/lib/eth.spec.ts
ebadiere: Ugh!  I removed it but didn't stage the change.  Thanks for catching this.

AlfredoG87 (approved):

packages/relay/src/lib/eth.ts
```diff
@@ -1574,27 +1578,35 @@ export class EthImpl implements Eth {
...
   private async getHistoricalBlockResponse(blockNumberOrTag?: string | null, returnLatest?: boolean, requestId?: string | undefined): Promise<any | null> {
-    let blockResponse: any;
-    // Determine if the latest block should be returned and if not then just return null
+    const maxBlockRange = Number(process.env.MAX_BLOCK_RANGE) || 5;
     if (!returnLatest && EthImpl.blockTagIsLatestOrPending(blockNumberOrTag)) {
       return null;
     }
-
+  
+    if (blockNumberOrTag != null && blockNumberOrTag.length < 32 && !EthImpl.blockTagIsEarliestOrPending(blockNumberOrTag)) {
```
Nana-EC, line 1586: I think we've introduced dupklication here.
If `blockNumberOrTag` === `latest` will you not end up making the following calls twice? Here and then on line 1595
```
const latestBlockResponse = await this.mirrorNodeClient.getLatestBlock(requestId);
      const latestBlock = latestBlockResponse.blocks[0];
```

Nana-EC (commented):
Missed my feedback before merging but there might be a performance regression here

packages/relay/src/lib/eth.ts
ebadiere: You're right.  The fix is easy since we're only interested in actual block numbers, so we can add a !EthImpl.blockTagIsLatestOrPending(blockNumberOrTag) to that condition.

PR #1264, opened 2023-05-17
configurable eth_call accepted errors

AlfredoG87:
**Description**:
make accepted error codes from mirror node for eth_call configurable and added initial default ones

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
Whenever the mirror node returns an acceptedErrorCode  the  mirrorNode  request  will return null, and in response  we will  return empty hex.

image

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -304,6 +304,9 @@ export class MirrorNodeClient {
         const acceptedErrorResponses = MirrorNodeClient.acceptedErrorStatusesResponsePerRequestPathMap.get(pathLabel);
         if (error.response && acceptedErrorResponses && acceptedErrorResponses.indexOf(effectiveStatusCode) !== -1) {
             this.logger.debug(`${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`);
+            if(pathLabel  === MirrorNodeClient.CONTRACT_CALL_ENDPOINT) {
+                this.logger.warn(`${requestIdPrefix} [${method}] ${path} Error details: (StatusText: '${error.response.statusText}' Data: '${JSON.stringify(error.response.data)}'"')`);
+            }
             return null;
         }
 
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -97,7 +97,7 @@ export class MirrorNodeClient {
         [MirrorNodeClient.GET_NETWORK_FEES_ENDPOINT, [400, 404]],
         [MirrorNodeClient.GET_TOKENS_ENDPOINT, [400, 404]],
         [MirrorNodeClient.GET_TRANSACTIONS_ENDPOINT, [400, 404]],
-        [MirrorNodeClient.CONTRACT_CALL_ENDPOINT, []],
+        [MirrorNodeClient.CONTRACT_CALL_ENDPOINT, [400,404,415,429,500]],
         [MirrorNodeClient.GET_STATE_ENDPOINT, [400, 404]]
     ]);
 
@@ -228,6 +228,13 @@ export class MirrorNodeClient {
 
         this.logger.info(`Mirror Node client successfully configured to REST url: ${this.restUrl} and Web3 url: ${this.web3Url} `);
         this.cache = new LRU({ max: constants.CACHE_MAX, ttl: constants.CACHE_TTL.ONE_HOUR });
+
+        // set  up eth call  accepted error codes.
+        if(process.env.ETH_CALL_ACCEPTED_ERRORS) {
+            MirrorNodeClient.acceptedErrorStatusesResponsePerRequestPathMap
+                .set(MirrorNodeClient.CONTRACT_CALL_ENDPOINT, JSON.parse(process.env.ETH_CALL_ACCEPTED_ERRORS));
+        }
+
     }
 
     private buildUrl(baseUrl: string) {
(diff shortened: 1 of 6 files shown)


packages/relay/tests/lib/eth.spec.ts
```diff
@@ -3390,17 +3390,17 @@ describe('Eth calls using MirrorNode', async function () {
         }
       });
 
-      sdkClientStub.submitContractCallQueryWithRetry.returns({
+      /*sdkClientStub.submitContractCallQueryWithRetry.returns({
```
Nana-EC, line 3393: Did you mean to remove these comments?

packages/server/tests/acceptance/htsPrecompile/precompileCalls.spec.ts
```diff
@@ -738,12 +738,14 @@ describe('@precompile-calls Tests for eth_call with HTS', async function () {
...
 
-            await relay.callFailing(
+            /*await relay.callFailing(
                 'eth_call',
                 [callData, 'latest'],
                 predefined.CONTRACT_REVERT(),
                 requestId
-            );
+            );*/
+            const res = await relay.call('eth_call', [callData, 'latest'], requestId);
```
ebadiere, line 747: So these are meant to be enabled again I assume, correct?

ebadiere (commented):
One question around the commented out test relay calls.

packages/relay/tests/lib/eth.spec.ts
Nana-EC: Also when you don't expect a fallback to consensus not would be good to ensure this is not called

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -297,6 +304,9 @@ export class MirrorNodeClient {
         const acceptedErrorResponses = MirrorNodeClient.acceptedErrorStatusesResponsePerRequestPathMap.get(pathLabel);
         if (error.response && acceptedErrorResponses && acceptedErrorResponses.indexOf(effectiveStatusCode) !== -1) {
             this.logger.debug(`${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`);
+            if(pathLabel  === MirrorNodeClient.CONTRACT_CALL_ENDPOINT) {
```
Nana-EC, line 307: I think you should add a new set of tests for `handleError()`
This can confirm for `contract/call` that the codes we allow return null and the ones we don't throw the error

Nana-EC (changes requested):
Add some simple `handleErrror()` tests and remove the comments.

packages/server/tests/acceptance/htsPrecompile/precompileCalls.spec.ts
AlfredoG87: I will remove  the  comments.

packages/relay/src/lib/clients/mirrorNodeClient.ts
Nana-EC: Maybe
```ts
describe('handleError', async() => {

    const CONTRACT_CALL_ENDPOINT = 'contracts/call';
    const nullResponseCodes = [400,404,415,429,500];
    const errorRepsonseCodes = [501, 503];

    for (const code of nullResponseCodes) {
      it(`returns null when ${code} is returned`, async () => {
        const result = await mirrorNodeInstance.handleError(new Error('test error'), CONTRACT_CALL_ENDPOINT, code, 'POST');
        expect(result).to.equal(null);
      });
    }

    for (const code of errorRepsonseCodes) {
      it(`throws an error when ${code} is returned`, async () => {
        try {
          await mirrorNodeInstance.handleError(new Error('test error'), CONTRACT_CALL_ENDPOINT, code, 'POST');
          expect.fail('should have thrown an error');
        } catch (e) {
          expect(e.message).to.equal('test error');
        }
      });
    }
  });
```

packages/relay/tests/lib/eth.spec.ts
AlfredoG87: ✅  done

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: Thank for  the  starting point,  good suggestion!  👍 

✅   done

packages/relay/tests/lib/mirrorNodeClient.spec.ts
```diff
@@ -48,13 +50,44 @@ describe('MirrorNodeClient', async function () {
       },
       timeout: 20 * 1000
     });
+    //  process.env.ETH_CALL_ACCEPTED_ERRORS =
```
Nana-EC: nit: missed but leave unless you have to push another commit for something else

packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -348,7 +348,9 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
                             data: '0x3ec4de3800000000000000000000000067d8d32e9bf1a9968a5ff53b87d777aa8ebbee69'
                         };
 
-                        await relay.callFailing('eth_call', [callData, 'latest'], predefined.CONTRACT_REVERT(), requestId);
+                        // await relay.callFailing('eth_call', [callData, 'latest'], predefined.CONTRACT_REVERT(), requestId);
```
Nana-EC: nit: leftover

packages/server/tests/acceptance/htsPrecompile/precompileCalls.spec.ts
AlfredoG87: ✅   comments  removed

Nana-EC (approved):
left over comments, but looks good

packages/server/tests/acceptance/rpc_batch3.spec.ts
Nana-EC: Seems to be removed

PR #1269, opened 2023-05-18
Return correct error for rate limit

georgi-l95:
**Description**:
This PR modifies current behaviour to return correct error for code 429 and contract revert.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1174,7 +1174,11 @@ export class EthImpl implements Eth {
       
       if (e instanceof MirrorNodeClientError) {
         if (e.isRateLimit()) {
-          return predefined.IP_RATE_LIMIT_EXCEEDED(e.errorMessage || "eth_call");
+          return predefined.IP_RATE_LIMIT_EXCEEDED(e.errorMessage || `Rate limit exceeded on ${EthImpl.ethCall}`);
+        }
+
+        if (e.isContractReverted()) {
+          return predefined.CONTRACT_REVERT(e.errorMessage);
         }
 
         // Temporary workaround until mirror node web3 module implements the support of precompiles
--- a/packages/relay/tests/helpers.ts
+++ b/packages/relay/tests/helpers.ts
@@ -176,6 +176,18 @@ const mockData = {
                 }
             ]
         }
+    },
+
+    contractReverted: {
+        "_status": {
+            "messages": [
+                {
+                    "message": "CONTRACT_REVERT_EXECUTED",
+                    "detail": "",
+                    "data": "0x"
+                }
+            ]
+        }
     }
 };
 
--- a/packages/relay/tests/lib/eth.spec.ts
+++ b/packages/relay/tests/lib/eth.spec.ts
@@ -3331,6 +3331,21 @@ describe('Eth calls using MirrorNode', async function () {
         expect(result.name).to.eq("IP Rate limit exceeded");
     });
 
+    it('eth_call with all fields but mirrorNode throws 400', async function () {
+      const callData = {
+        ...defaultCallData,
+        "from": accountAddress1,
+        "to": contractAddress2,
+        "data": contractCallData,
+        "gas": maxGasLimit
+      };
+      web3Mock.onPost('contracts/call', {...callData, estimate: false}).reply(400, mockData.contractReverted);
+      const result = await ethImpl.call(callData, 'latest');
+      expect(result).to.be.not.null;
+      expect(result.code).to.eq(-32008);
+      expect(result.name).to.eq("Contract revert executed");
+    });
+
     it('eth_call with all fields but mirrorNode throws 504 (timeout) on pre-check', async function () {
 
       const timeoutAddress = "0x00000000000000000000000000000000000004e2";


packages/relay/src/lib/eth.ts
```diff
@@ -1168,18 +1168,26 @@ export class EthImpl implements Eth {
...
-      this.logger.error(e, `${requestIdPrefix} Failed to successfully submit eth_call`);
       if (e instanceof JsonRpcError) {
         return e;
       }
+      
+      if (e instanceof MirrorNodeClientError) {
+        if (e.isRateLimit()) {
+          return predefined.IP_RATE_LIMIT_EXCEEDED(e.errorMessage || "eth_call");
+        }
+
```
AlfredoG87, line 1179: are we missing the `400` error check here?

packages/relay/src/lib/errors/MirrorNodeClientError.ts
```diff
@@ -31,6 +31,7 @@ export class MirrorNodeClientError extends Error {
 
     static statusCodes = {
       NOT_FOUND: 404,
+      TOO_MANY_REQUESTS: 429,
       NO_CONTENT: 204
```
AlfredoG87, line 35: also, should we add `400` error for `CONTRACT_REVERT_EXECUTED`

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -97,7 +97,7 @@ export class MirrorNodeClient {
         [MirrorNodeClient.GET_NETWORK_FEES_ENDPOINT, [400, 404]],
         [MirrorNodeClient.GET_TOKENS_ENDPOINT, [400, 404]],
         [MirrorNodeClient.GET_TRANSACTIONS_ENDPOINT, [400, 404]],
-        [MirrorNodeClient.CONTRACT_CALL_ENDPOINT, [400,404,415,429,500]],
+        [MirrorNodeClient.CONTRACT_CALL_ENDPOINT, [404, 415, 500]],
```
AlfredoG87, line 100: just note that this is default values and currently is being override by env variable:
`ETH_CALL_ACCEPTED_ERRORS`

packages/relay/tests/lib/mirrorNodeClient.spec.ts
```diff
@@ -60,7 +60,7 @@ describe('MirrorNodeClient', async function () {
   describe('handleError', async() => {
 
     const CONTRACT_CALL_ENDPOINT = 'contracts/call';
-    const nullResponseCodes = [400,404,415,429,500];
+    const nullResponseCodes = [404,415,500];
```
Nana-EC, line 63: Q: should you be adding 400 and 429 to `errorRepsonseCodes`?

Nana-EC (changes requested):
Looks good, just test on http code error needs to be updated

packages/relay/tests/lib/mirrorNodeClient.spec.ts
AlfredoG87: ✅  Done. Added them.

packages/relay/src/lib/errors/MirrorNodeClientError.ts
Nana-EC: No since it only applies to the contracts/call endpoint. 400 status code on other mirror node calls just means an error right.

packages/relay/src/lib/eth.ts
Nana-EC: I think it's working because that check is already in `MirrorNodeClient.handleError()`
Might be cleaner to have it here but maybe for 0.24 to minimize moving things around.
However, if it's causing a gap or bug then yeah we can move or add a check here.
Adding a check would be harmless and address the concern

packages/relay/src/lib/eth.ts
```diff
@@ -1168,18 +1168,26 @@ export class EthImpl implements Eth {
...
-        return await this.callConsensusNode(call, gas, requestId);
-      } 
-      this.logger.error(e, `${requestIdPrefix} Failed to successfully submit eth_call`);
       if (e instanceof JsonRpcError) {
         return e;
       }
+      
+      if (e instanceof MirrorNodeClientError) {
+        if (e.isRateLimit()) {
+          return predefined.IP_RATE_LIMIT_EXCEEDED(e.errorMessage || "eth_call");
```
Nana-EC, line 1177:
```suggestion
          return predefined.IP_RATE_LIMIT_EXCEEDED(e.errorMessage || `Rate limit exceeded on ${EthImpl.ethCall});
```

AlfredoG87 (dismissed):
LGTM

packages/relay/src/lib/eth.ts
Nana-EC: Actually @AlfredoG87 i think you're right. We need to expose the contract reverted.
So there should be this check below

```ts
        if (e.isContractReverted()) {
          return predefined.CONTRACT_REVERT(e.errorMessage);
        }
```

Nana-EC (changes requested):
Looks good, just 2 suggestions on message and contract revert

packages/relay/src/lib/eth.ts
AlfredoG87: ✅  Done,  also  added  UT to validate is working  correctly

packages/relay/src/lib/eth.ts
AlfredoG87: ✅ Done  👍  good  suggestion.

Nana-EC (approved):
LGTM

AlfredoG87 (approved):
LG 😅

PR #1270, opened 2023-05-18
936 block number range 2

### LINKED ISSUE #936: Add a max block number range check

ebadiere:
Added check for latest block to avoid regression, multiple calls to the mirrornode.

Fixes #936 


packages/relay/src/lib/eth.ts
```diff
@@ -1583,7 +1583,7 @@ export class EthImpl implements Eth {
       return null;
     }
   
-    if (blockNumberOrTag != null && blockNumberOrTag.length < 32 && !EthImpl.blockTagIsEarliestOrPending(blockNumberOrTag)) {
+    if (blockNumberOrTag != null && blockNumberOrTag.length < 32 && !EthImpl.blockTagIsEarliestOrPending(blockNumberOrTag) && !EthImpl.blockTagIsLatestOrPending(blockNumberOrTag)) {
```
Nana-EC, line 1586: Should be a test to confirm behaviour here.
The fact that there's no failing tests means there's missing test coverage

Nana-EC (changes requested):
Nice, but missing test

packages/relay/src/lib/eth.ts
ebadiere: Done.

ebadiere:
Addressed in another PR.

PR #1271, opened 2023-05-18
Paginate eth_getBlockByHash

### LINKED ISSUE #1261: Missing Data in eth_getBlockByHash
### Description

When making a `eth_getBlockByHash` request and specifying a block hash with >100 transactions, only 100 are returned

### Steps to reproduce

1. Start the hedera-json-rpc-relay server running against the Hedera testnet
2. Issue a POST request to http://localhost:7546 with the following data:
```json
{
    "jsonrpc": "2.0",
    "method": "eth_getBlockByHash",
    "params": [
        "0x3e3ee4b61360009ad85be5f942c52fcfb87daa2b63aa3b567e923bda2d45b3d4",
        false
    ],
    "id": 1
}
```

This references block 4625259 which contains 1308 transactions according to <https://testnet.mirrornode.hedera.com/api/v1/blocks/4625259>.

### Additional context

Under the hood, `eth_getBlockByHash` makes mirror node calls to `https://testnet.mirrornode.hedera.com/api/v1/contracts/results?timestamp=gte:${from}&timestamp=lte:${to}&limit=100&order=asc` which returns the following pointer:
```
{
  results: [/* transaction data */],
  links: {
    next: 'https://testnet.mirrornode.hedera.com/api/v1/contracts/results/...'
  }
}
```

Increasing the limit value >100 has no effect on results.

Oddly enough when looping through that pointer, only 664 results are retrieved for this block (4625259)

### Hedera network

mainnet, testnet

### Version

0.23

### Operating system

Linux

bmino:
**Description**:

This PR paginates eth_getBlockByHash requests in order to return all transactions within the specified range

**Related issue(s)**:

Fixes #1261

**Notes for reviewer**:
One test is failing for me locally which I haven't tracked down why:
```
1) RPC Server
        should execute "eth_getTransactionByHash with missing transaction"
    Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1519,7 +1519,7 @@ export class EthImpl implements Eth {
     const maxGasLimit = constants.BLOCK_GAS_LIMIT;
     const gasUsed = blockResponse.gas_used;
 
-    if (contractResults.length === 0) {
+    if (contractResults == null) {
       // contract result not found
       return null;
     }
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1519,7 +1519,7 @@ export class EthImpl implements Eth {
     const maxGasLimit = constants.BLOCK_GAS_LIMIT;
     const gasUsed = blockResponse.gas_used;
 
-    if (contractResults?.results == null) {
+    if (contractResults == null) {
       // contract result not found
       return null;
     }
@@ -1529,7 +1529,7 @@ export class EthImpl implements Eth {
     const transactionObjects: Transaction[] = [];
     const transactionHashes: string[] = [];
 
-    for (const result of contractResults.results) {
+    for (const result of contractResults) {
       // depending on stage of contract execution revert the result.to value may be null
       if (!_.isNil(result.to)) {
         if(showDetails) {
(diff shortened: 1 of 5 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1519,7 +1519,7 @@ export class EthImpl implements Eth {
     const maxGasLimit = constants.BLOCK_GAS_LIMIT;
     const gasUsed = blockResponse.gas_used;
 
-    if (contractResults?.results == null) {
+    if (contractResults == null) {
```
Nana-EC, line 1522: When is this ever true with the use of pagination?
Since the result is an array at worst it'll be empty. So you should also check for emptiness

Nana-EC (changes requested):
Nice addition.
One suggestion where we can improve the flow for empty cases

packages/relay/src/lib/eth.ts
bmino: good eye. the results response from pagination defaults to `[]`. gonna change this to check array length

bmino: looking at this further and an empty array for contractResults is totally okay for blocks that don't have ContractCall, ContractCreate, or EthereumTransaction transactions

Nana-EC: Correct, however, you're utilizing the pagination logic which starts with `[]` and concats into it any results found.
So if no blocks are found the `const contractResults = await this.mirrorNodeClient.getContractResults()` will now return an empty array not a null response

Nana-EC: @bmino I still think this is an issue. A simple addition should do the trick
```suggestion
    if (contractResults == null || contractResults.length === 0) {
```

Nana-EC: I think I see the issue here. Logically this is a fail safe. A found may have no contract results, but we'd still want to send back the block response with no transactions in it, in which case we should return null on an empty set of contract results.
Will circle back to the null check but that's out of scope for your PR

Nana-EC (approved):
LG

georgi-l95 (approved):
LGTM

PR #1280, opened 2023-05-19
Refactor `eth_estimateGas`

georgi-l95:
**Description**:
This PR is part of the effort to stabilise DAPP tests.

Future-proof refactor of `eth_estimateGas` with mirror-node calls at top.
Flow is:
1. Get gas estimation from the mirror-node
    a. If there is a result, return it only if it's not a contract create( due to bug in the mirror-node in which it returns 21k, not enough)
2.  If there is a error:
   a. Continue with the old approach. Check if it's a transfer, if it is, check value and if the account exists and return correct gas.
   b. If it's a contract call / contract create return correct gas.

This flow is mostly tested on previewnet.
Also added some more unit tests for this behaviour.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -436,42 +436,54 @@ export class EthImpl implements Eth {
...
-          return EthImpl.gasTxBaseCost;
+    
+    let gas = EthImpl.gasTxBaseCost;
+    try {
+      const contractCallResponse = await this.mirrorNodeClient.postContractCall({
+        ...transaction,
+        estimate: true
+      }, requestId);
+      if (contractCallResponse?.result) {
+        // Workaround unitl mirror-node bugfix applied, currently mirror-node retunrs 21k for contract creation, which is wrong
```
Nana-EC, line 447:
```suggestion
        // Workaround until mirror-node bugfix applied, currently mirror-node returns 21k for contract creation, which is wrong
```

Nana-EC (approved):

Nana-EC (approved):

AlfredoG87 (approved):
LGTM 👍

PR #1285, opened 2023-05-22
Stabilize Dapp Tests

georgi-l95:
**Description**:
This PR is part of the effort to stabilize dapp example tests.
- Tweaks some values to be able to work with testnet, previewnet and local-node.
- Fix bug where sometimes, if transaction has failed it cannot extract reason.
- Prepare HRC facade associate instead of the current ( we can enable it when we bump local-node version, to do that we need to finish flaky acceptance tests task)

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


Nana-EC (approved):
LG.
@georgi-l95 Btw were there some wait times that can be reduced? When the tests are waiting it seems it might wait for a while. Might be better to wait less so that failed suite runs fail faster for a quicker manual or automated retry.

Also we should be using at least 0.38 images for testing where the IHRC associate is supported so we should be able to test that now

georgi-l95:
@Nana-EC I'm preparing another PR about dapp tests and will check there if there are any wait time and if it's possible to reduce them, I agree that tests are taking some time and would be good to reduce it.
About the 0.38 image, I agree that we should bump it and use the HRC facade, but sadly a lot of other tests are failing due to latest security measures. Ivo is working on those problems as well as on the flaky acceptance tests, they both have the same issues. When flaky acceptance test PR is ready, we should be good to go to bump the images and use the HRC facade.

PR #1286, opened 2023-05-22
Handle null transaction on contract revert.

### LINKED ISSUE #1268: Cannot read properties of null (reading 'transactions')
### Description

When a contract is reverted, in some cases the mirrornode seems to return an empty transactions array causing the following error:

```hedera-json-rpc-relay     |     err: {
hedera-json-rpc-relay     |       "type": "TypeError",
hedera-json-rpc-relay     |       "message": "Cannot read properties of null (reading 'transactions')",
hedera-json-rpc-relay     |       "stack":
hedera-json-rpc-relay     |           TypeError: Cannot read properties of null (reading 'transactions')
hedera-json-rpc-relay     |               at EthImpl.<anonymous> (/home/node/app/packages/relay/dist/lib/eth.js:970:28)
hedera-json-rpc-relay     |               at Generator.next (<anonymous>)
hedera-json-rpc-relay     |               at fulfilled (/home/node/app/packages/relay/dist/lib/eth.js:24:58)
hedera-json-rpc-relay     |               at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
hedera-json-rpc-relay     |     }```

### Steps to reproduce

Not obvious, however a unit test can be set up to trigger the above.

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.23.2

### Operating system

None

### PR DESCRIPTION
**Pull request by ebadiere**
Handle null transaction on contract revert in eth_sendRawTransaction.

**Related issue(s)**:

Fixes #1268 

### CODE REVIEW

File: packages/relay/tests/lib/eth.spec.ts [hunk 1]
```diff
@@ -3809,6 +3827,30 @@ describe('Eth calls using MirrorNode', async function () {
...
+        gasPrice,
+        gasLimit: maxGasLimitHex,
+      };
+    
+      sdkClientStub.getAccountBalanceInTinyBar.returns(ethers.BigNumber.from('1000000000000000000000'));    
+      const signed = await signTransaction(transaction);
+      const id = uuid();
+    
+      restMock.onGet('network/fees').reply(200, defaultNetworkFees);
+      restMock.onGet('transactions/0.0.902-1684375868-230217103').reply(200, null);
```
Nana-EC, line 3848: Based on your fix it seems the actual issue is a null transaction. In that case this call would actually be returning a 404 with a contract revert message

Nana-EC (changes requested):
Fix seems right but tests seem off.
I think we should be checking for a not found transaction

packages/relay/tests/lib/eth.spec.ts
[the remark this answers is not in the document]
ebadiere: This is a good catch.  In both cleaning up the test and dealing with merge conflicts I removed a necessary piece of the test setup.  The fact that this test passed illustrates how difficult it is to test.  My testing involved stepping through the code and watching the updated condition be evaluated correctly with no logging.   Once an error is caught, the hash is generated based on the transaction buffer.  An error can be thrown for many reasons.  The test passed due to the wrong error being thrown.  The only way this test could be narrowed in scope would be to add checking for the logging response, and in this case there would be none, so the test would check that the "Transaction failed with result:" was not logged.  Perhaps a small refactor for testability should be considered?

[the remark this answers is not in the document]
Nana-EC: Per the issue you're fixing the right test would be to confirm as you've done that when the consensus node returns a contract revert that the function doesn't fail when getting the details from the mirror node.

You're right on narrowing the test conditions, it shows that `extractContractRevertReason()` is probably not best situated in `eth.ts`, probably should be in  another class of util nature that would allow for testing that function specifically.
Maybe move it to `MirrorNodeClient.ts` as a `getContractRevertReasonFromTransaction`, then you can have 2 test verify a 200 and a 404 case in the `mirrorNodeClient.spec.ts`.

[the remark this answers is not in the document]
ebadiere: Agreed, and updated.

ebadiere: Should be addressed in the updates and three new tests.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -669,6 +670,40 @@ export class MirrorNodeClient {
...
+    public async getContractRevertReasonFromTransaction(e: any, requestId: string | undefined, requestIdPrefix: string): Promise<any | undefined> {
+        if (e instanceof SDKClientError && e.isContractRevertExecuted()) {
+            const transactionId = e.message.match(constants.TRANSACTION_ID_REGEX);
+            if (transactionId) {
+              const tx = await this.getTransactionById(transactionId[0], undefined, requestId);
+
+              if(tx === null){
+
+                this.logger.error(`${requestIdPrefix} Transaction failed with null result: ${MirrorNodeClientError.statusCodes.NOT_FOUND}`);
+                return MirrorNodeClientError.statusCodes.NOT_FOUND;
```
Nana-EC, line 689: Return null here.
`NOT_FOUND` would be if we just made an HTTP request call but for this function we want the message so it it's invalid return null

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -669,6 +670,40 @@ export class MirrorNodeClient {
...
+
+              } else if(tx.length === 0){
+
+                this.logger.error(`${requestIdPrefix} Transaction failed with empty result: ${MirrorNodeClientError.statusCodes.NO_CONTENT}`);
+                return MirrorNodeClientError.statusCodes.NO_CONTENT;
```
Nana-EC, line 694: return null

Nana-EC (changes requested):
Change i on the right path, i think it should return null in all invalid cases as the log is sufficient

packages/relay/src/lib/clients/mirrorNodeClient.ts
ebadiere: Done

packages/relay/src/lib/clients/mirrorNodeClient.ts
ebadiere: Done

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -3809,6 +3827,37 @@ describe('Eth calls using MirrorNode', async function () {
...
+          status: { _code: 33 },
+          message: 'Error: receipt for transaction 0.0.902@1684375868.230217103 contained error status',
+        })
+      );
+        
+      const signed = await signTransaction(transaction);
+      const id = uuid();
+    
+      restMock.onGet('network/fees').reply(200, defaultNetworkFees);
+      restMock.onGet('transactions/0.0.902-1684375868-230217103').reply(200, null);
```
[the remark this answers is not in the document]
ebadiere: Done.

Nana-EC (approved):

PR #1288, opened 2023-05-23
Handle undefined `block_number`

### LINKED ISSUE #1277: Error on EthImpl.numberTo0x() use on contract results block num
### Description

It seems for some contract results the relay is throwing errors converting the `contractResult.block_number` to a hex.
```
[POST] eth_getTransactionByHash: -32603 274 ms
    err: {
      "type": "TypeError",
      "message": "Cannot read properties of undefined (reading 'toString')",
      "stack":
          TypeError: Cannot read properties of undefined (reading 'toString')
              at EthImpl.numberTo0x (/Users/user/code/hedera/hedera-json-rpc-relay/packages/relay/dist/lib/eth.js:1129:41)
              at EthImpl.<anonymous> (/Users/user/code/hedera/hedera-json-rpc-relay/packages/relay/dist/lib/eth.js:1056:43)
              at fulfilled (/Users/user/code/hedera/hedera-json-rpc-relay/packages/relay/dist/lib/eth.js:24:58)
    }
```

Seems to map back to https://github.com/hashgraph/hedera-json-rpc-relay/blob/release/0.23/packages/relay/src/lib/eth.ts#L1312

### Steps to reproduce

This likely occurs for failed transactions of a particular  kind

1. Query `eth_getTransactionByHash` for a transaction with a null or undefined block_number 

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.23.0.5

### Operating system

None

georgi-l95:
**Description**:
This PR adds additional logic to handle rare cases of undefined `block_number`.

There are certain rare cases, where response from mirror-node does not have transaction_index or block_number properties. That can happen for example when contracts/results api is called before all the data is saved in the DB. A single re-fetch should fix this. But in case this is not enough, we can now check if `block_number` is `undefined` and return `null`. This is okay for all ethereum tools, as in ethereum `block_number == undefined` means that transaction is pending. So even if we return null, we are not braking any usecases.

**Related issue(s)**:

Fixes #1277 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1322,6 +1322,10 @@ export class EthImpl implements Eth {
       return null;
     }
 
+    if (!contractResult.block_number || !contractResult.transaction_index) {
+      this.logger.warn(`${requestIdPrefix} getTransactionByHash(hash=${hash}) mirror-node returned status 200 with missing properties in contract_results - block_number==${contractResult.block_number} and transaction_index==${contractResult.transaction_index}`);
+    }
+
     let fromAddress;
     if (contractResult.from) {
       fromAddress = contractResult.from.substring(0, 42);


Nana-EC (dismissed):
LG
Fix is solid, just unclear on how an undefined block number could happen.
Might need a follow up investigation

Nana-EC (approved):

Nana-EC:
> LG Fix is solid, just unclear on how an undefined block number could happen. Might need a follow up investigation

@georgi-l95 did you ever follow up here? @natanasow hit this recently.
Was there an issue or follow up on MN side?

Issue #822, opened 2023-01-19
Add `blockParam` to endpoints according to EIP-1898


georgi-l95:
According to [EIP-1898](https://eips.ethereum.org/EIPS/eip-1898) we should make the following endpoints accept block parameter ( tag, block number or object with block hash or block number ):

- [x] eth_getBalance
- [x] eth_getStorageAt
- [ ] eth_getTransactionCount
- [ ] eth_getCode
- [ ] eth_call
- [ ] eth_getProof

We currently only support blockParam on `eth_call`. Infura and Alchemy as well on support blockParam on the same endpoint.
Research and expand the rest of the endpoints, if needed.
```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/1856
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/1857
```

Nana-EC:
Converted list to tasks.

SUpported endpoints have the tag added.
However not all support historical cases as Mirror Node may not yet support it

georgi-l95:
Resolved for all relevant endpoints.Thus closing this issue.

PR #1297, opened 2023-05-26
Validate gas limit and block param on eth_call

### LINKED ISSUE #1056: Limit ContractCallQuery requests for eth_calls
### Problem

Currently, it appears that the relay does a `ContractCallQuery` request to a consensus node for `eth_call` requests under the following conditions:

### `ETH_CALL_DEFAULT_TO_CONSENSUS_NODE` is `true` AND:

1. gas specified is >15m; however, caps the gas at 15m
2. or `blockNumber`/`tag` is NOT `latest`

### `ETH_CALL_DEFAULT_TO_CONSENSUS_NODE` is `false` and the mirror node POST call fails(for example due to rate limits, downtime, etc and hence a `ContractCallQuery` request is used as a fallback) AND:

1.  gas specified is >15m; however, caps the gas at 15m
2. or `blockNumber`/`tag` is NOT `latest`

### Solution

The relay should not ever make `ContractCallQuery` requests under the following 2 conditions:
1. if gas specified > ContractCallQuery gas limit(currently 15m). The mirror node POST call supports a max gas of 120m; if users want to allow for the possibility of a fallback `ContractCallQuery` request in the case that the mirror node call fails then they should specify a gas <=15m.
2. or if `blockNumber`/`tag` is NOT `latest` since the consensus node only has the ability to execute `ContractCallQuery` requests at the `latest` block, so executing a `ContractCallQuery` if a prior block is specified for the `eth_call` request may result in incorrect data being returned if the result at `latest` does not equal the result at the specified `blockNumber`/`tag`.

### Alternatives

_No response_

Nana-EC:
**Description**:

**Related issue(s)**:

Fixes #1056 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/constants.ts
+++ b/packages/relay/src/lib/constants.ts
@@ -91,6 +91,9 @@ export default {
         mainnet: 0x127,
         testnet: 0x128,
         previewnet: 0x129,
-    }
+    },
+
+    // block ranges
+    MAX_BLOCK_RANGE: 5,
 };
 
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -41,20 +41,20 @@ export const predefined = {
     message: `execution reverted: ${decodeErrorMessage(errorMessage)}`,
     data: errorMessage
   }),
-  'GAS_LIMIT_TOO_HIGH': new JsonRpcError({
+  'GAS_LIMIT_TOO_HIGH': (gasLimit, maxGas) => new JsonRpcError({
     name: 'gasLimit too high',
     code: -32005,
-    message: 'Transaction gas limit exceeds block gas limit'
+    message: `Transaction gas limit '${gasLimit}' exceeds block gas limit '${maxGas}'`
   }),
-  'GAS_LIMIT_TOO_LOW': new JsonRpcError({
+  'GAS_LIMIT_TOO_LOW': (gasLimit, requiredGas) => new JsonRpcError({
     name: 'gasLimit too low',
     code: -32003,
-    message: 'Intrinsic gas exceeds gas limit'
+    message: `Transaction gas limit provided '${gasLimit}' is insufficient of intrinsic gas required '${requiredGas}'`
   }),
-  'GAS_PRICE_TOO_LOW': new JsonRpcError({
+  'GAS_PRICE_TOO_LOW': (gasPrice, minGasPrice) => new JsonRpcError({
     name: 'Gas price too low',
     code: -32009,
-    message: 'Gas price below configured minimum gas price'
+    message: `Gas price '${gasPrice}' is below configured minimum gas price '${minGasPrice}'`
   }),
   'HBAR_RATE_LIMIT_EXCEEDED': new JsonRpcError({
     name: 'HBAR Rate limit exceeded',
@@ -106,10 +106,10 @@ export const predefined = {
     code: -32602,
     message: `Missing value for required parameter ${index}`
   }),
-  'NONCE_TOO_LOW': new JsonRpcError({
+  'NONCE_TOO_LOW': (nonce, currentNonce) => new JsonRpcError({
     name: 'Nonce too low',
     code: 32001,
-    message: 'Nonce too low'
+    message: `Nonce too low. Provided nonce: ${currentNonce}, current nonce: ${nonce}`
   }),
   'NO_MINING_WORK': new JsonRpcError({
     name: 'No mining work',
@@ -211,5 +211,10 @@ export const predefined = {
     name: 'Exceeded maximum allowed subscriptions',
     code: -32608,
     message: 'Exceeded maximum allowed subscriptions'
-  })
+  }),
+  'UNSUPPORTED_HISTORICAL_EXECUTION': (blockId: string) => new JsonRpcError({
+    name: 'Unsupported historical block request',
+    code: -32000,
+    message: `Unsupported histoircal block identifier encounterd: ${blockId}`
+  }),
 };
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -95,6 +95,7 @@ export class EthImpl implements Eth {
   private readonly ethCallCacheTtl = Number.parseInt(process.env.ETH_CALL_CACHE_TTL ?? constants.ETH_CALL_CACHE_TTL_DEFAULT.toString());
   private readonly ethBlockNumberCacheTtlMs = Number.parseInt(process.env.ETH_BLOCK_NUMBER_CACHE_TTL_MS ?? constants.ETH_BLOCK_NUMBER_CACHE_TTL_MS_DEFAULT.toString());
   private readonly ethGetBalanceCacheTtlMs = Number.parseInt(process.env.ETH_GET_BALANCE_CACHE_TTL_MS ?? constants.ETH_GET_BALANCE_CACHE_TTL_MS_DEFAULT.toString());
+  private readonly maxBlockRange = Number(process.env.MAX_BLOCK_RANGE ?? constants.MAX_BLOCK_RANGE.toString());
 
   /**
    * Configurable options used when initializing the cache.
@@ -1146,7 +1147,7 @@ export class EthImpl implements Eth {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     this.logger.trace(`${requestIdPrefix} call(hash=${JSON.stringify(call)}, blockParam=${blockParam})`, call, blockParam);
 
-    const to = await this.performCallChecks(call, requestId);
+    const to = await this.performCallChecks(call, blockParam, requestId);
 
     // Get a reasonable value for "gas" if it is not specified.
     let gas = this.getCappedBlockGasLimit(call.gas, requestId);
@@ -1271,7 +1272,7 @@ export class EthImpl implements Eth {
    * @param call
    * @param requestId
    */
-  async performCallChecks(call: any, requestId?: string) {
+  async performCallChecks(call: any, blockParam: string | null, requestId?: string) {
     // The "to" address must always be 42 chars.
     if (!call.to || call.to.length != 42) {
       throw predefined.INVALID_CONTRACT_ADDRESS(call.to);
@@ -1284,6 +1285,32 @@ export class EthImpl implements Eth {
         throw predefined.NON_EXISTING_ACCOUNT(call.from);
       }
     }
+
+    // verify gas is withing the allowed range
+    if (call.gas && call.gas > constants.BLOCK_GAS_LIMIT) {
+      throw predefined.GAS_LIMIT_TOO_HIGH(call.gas, constants.BLOCK_GAS_LIMIT);
+    }
+
+    // If "BlockParam" is distinct from blank, we check is a valid block number
+    if(blockParam) {
+      const blockNum = Number(blockParam);
+      if (blockNum === 0 || blockNum === 1 || EthImpl.blockTagIsEarliest(blockParam)) {
+        throw predefined.UNSUPPORTED_HISTORICAL_EXECUTION(blockParam);
+      }
+
+      if (!isNaN(blockNum)) {
+        const block = await this.mirrorNodeClient.getLatestBlock(requestId);
+        if(!block) {
+          throw predefined.RESOURCE_NOT_FOUND(`invalid block: ${blockParam}`);
+        }
+
+        const trailingBlockCount = block.number - blockNum;
+        if(trailingBlockCount > this.maxBlockRange) {
+          this.logger.warn(`${formatRequestIdMessage(requestId)} referenced block '${blockParam}' trails latest by ${trailingBlockCount}, max trailing count is ${this.maxBlockRange}. Throwable UNSUPPORTED_HISTORICAL_EXECUTION scenario.`);
+        }
+      }      
+    }
+
     // Check "To" is a valid Contract or HTS Address
     const toEntityType = await this.mirrorNodeClient.resolveEntityType(call.to, [constants.TYPE_TOKEN, constants.TYPE_CONTRACT], requestId);
     if(!(toEntityType?.type === constants.TYPE_CONTRACT || toEntityType?.type === constants.TYPE_TOKEN)) {
@@ -1489,8 +1516,12 @@ export class EthImpl implements Eth {
     return tag == null || tag === EthImpl.blockLatest || tag === EthImpl.blockPending;
   };
 
+  private static blockTagIsEarliest = (tag) => {
+    return tag === EthImpl.blockEarliest;
+  };
+
   private static blockTagIsEarliestOrPending = (tag) => {
-    return tag == null || tag === EthImpl.blockEarliest || tag === EthImpl.blockPending;
+    return tag == null || EthImpl.blockTagIsEarliest(tag) || tag === EthImpl.blockPending;
   };   
 
   /**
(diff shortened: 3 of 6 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1284,6 +1285,32 @@ export class EthImpl implements Eth {
...
     }
+
+    // verify gas is withing the allowed range
+    if (call.gas && call.gas > constants.BLOCK_GAS_LIMIT) {
+      throw predefined.GAS_LIMIT_TOO_HIGH(call.gas, constants.BLOCK_GAS_LIMIT);
+    }
+
+    // If "BlockParam" is distinct from blank, we check is a valid block number
+    if(blockParam) {
+      const blockNum = Number(blockParam);
```
mshakeg, line 1296: Could `blockParam` also be `{blockHash}`? [EIP-1898](https://eips.ethereum.org/EIPS/eip-1898) allows for `{blockHash}` to be passed instead of the `blockNumber` for various methods including `eth_call`.

So the type of `blockParam` should actually be extended to:
```ts
type BlockParam = {
  blockHash: string
} | string | null;
```

packages/relay/src/lib/constants.ts
```diff
@@ -91,6 +91,9 @@ export default {
         mainnet: 0x127,
         testnet: 0x128,
         previewnet: 0x129,
-    }
+    },
+
+    // block ranges
+    MAX_BLOCK_RANGE: 5,
```
mshakeg, line 97: Is this value hardcoded on the mirror node? Ideally, it should be configurable, in which case the relay should make a request at startup to determine the mirror node's config.

packages/relay/src/lib/eth.ts
```diff
@@ -1605,7 +1636,6 @@ export class EthImpl implements Eth {
    * @param returnLatest
    */
   private async getHistoricalBlockResponse(blockNumberOrTag?: string | null, returnLatest?: boolean, requestId?: string | undefined): Promise<any | null> {
```
mshakeg, line 1638: `blockNumberOrTag` is possibly the same type as `blockParam` i.e.:
```
type BlockParam = {
  blockHash: string
} | string | null;
```

This issue likely affects all of the methods listed in [EIP-1898](https://eips.ethereum.org/EIPS/eip-1898).

packages/relay/src/lib/eth.ts
```diff
@@ -1614,7 +1644,7 @@ export class EthImpl implements Eth {
     if (blockNumberOrTag != null && blockNumberOrTag.length < 32 && !isNaN(blockNumber)) {
       const latestBlockResponse = await this.mirrorNodeClient.getLatestBlock(requestId);
       const latestBlock = latestBlockResponse.blocks[0];
-      if (Number(blockNumberOrTag) > latestBlock.number + maxBlockRange) {
+      if (Number(blockNumberOrTag) > latestBlock.number + this.maxBlockRange) {
```
mshakeg, line 1647: Why not use the identical `blockNumber` instead of `Number(blockNumberOrTag)` here?

packages/relay/src/lib/eth.ts
```diff
@@ -1284,6 +1285,32 @@ export class EthImpl implements Eth {
         throw predefined.NON_EXISTING_ACCOUNT(call.from);
       }
     }
+
+    // verify gas is withing the allowed range
+    if (call.gas && call.gas > constants.BLOCK_GAS_LIMIT) {
```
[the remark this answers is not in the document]
Nana-EC: Great confirmation. Was wondering if this should be a separate config.
Although i think it should be 15 m all around though.

packages/relay/src/lib/eth.ts
Nana-EC: Yes it could. Good point.
We have goals for this with https://github.com/hashgraph/hedera-json-rpc-relay/issues/822
In this implementation I'm covering our existing logic of just numbers.
Let me explore how to comment this as the above ticket would address this consideration

packages/relay/src/lib/constants.ts
Nana-EC: No. This is a relay concept to optimize considerations of doing best effort.
Given the mirror node blocks come in every 2 seconds there's a concept that a call response for a bock even 1 block back could be wrong.
This is to say on average it's likely not the case but we'll make consideration for up to `MAX_BLOCK_RANGE` behind the latest block exposed by the mirror node.

packages/relay/src/lib/eth.ts
Nana-EC: Correct. Captured in https://github.com/hashgraph/hedera-json-rpc-relay/issues/822

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
mshakeg: @Nana-EC I don't think the mirror node should be fixed to 15m, maybe use that as a default but imo it should be configurable by the relay provider as popular providers(Alchemy and Infura) typically have a gas limit of 50m or 100m.

packages/relay/src/lib/constants.ts
mshakeg: @Nana-EC hmm, ok, the default MAX_BLOCK_RANGE of 5 equals ~10s which may not be an issue for most cases. 

However until this `MAX_BLOCK_RANGE` is deprecated i.e. until the mirror node supports archive data requests(historical execution requests) the following issues may arise:

1. if a subgraph is deployed and attempts to index an event and part of that event handler entails an `eth_call` which specified the block of the event which exceeds the MAX_BLOCK_RANGE then this will result in `UNSUPPORTED_HISTORICAL_EXECUTION` and so the subgraph will be stuck forever or at least until the mirror node now supports historical execution requests and the relay logic has been update accordingly.
2. if an `eth_call` specifies a block that is not the `latest` but within the `MAX_BLOCK_RANGE` and the result of the `eth_call` is a function of block specific data(e.g. `block.timestamp`) then this will return an invalid result. Additionally, if state changes occurred after the specified block but before `latest`(and still within `MAX_BLOCK_RANGE`) and the `eth_call` is in this case a function of that (changed)state then this will also return invalid data.

Nana-EC: Yep. The 1st case is why we're trying to make a best effort and the 2nd case is why the range is small.

Due to fast block times (every 2 secs) anyone making a call to `eth_blockNumber()` that then uses the response to call `eth_call` would essentially being making a historical call if they took longer that 2 secs.
For most DApps or tools this may not be the case but there will be scenarios where you are a few blocks off. 
Which is why I added the log message to get more data.

@mshakeg what are you suggesting in place?
Currently users are always getting the latest. Is it better to get the latest even when you request a historical, or fail on historical requests that are likely wrong

Nana-EC: Actually whiles I was writing this I was reminded of another idea I had a while back.
Right now we're guess based off of arbitrary time with block number as a proxy.

One approach could be instead to follow up with an `api/v1/contracts/{contractIdOrAddress}/state` and use the `timestamp` field to potentially determine if there was a state change since `block.timeStamp.to`.
If no then the contracts hasn't seen any state changes and returning the current state in the mean time is okay, if so then we can fail with the  `UNSUPPORTED_HISTORICAL_EXECUTION` to highlight there's been a state change.

Will think on this and see if it's possible in a follow up PR.
For now still need to explore whether to leave existing behaviour or adopt the `UNSUPPORTED_HISTORICAL_EXECUTION` which is more disruptive.
A graph node that's indexing my not care the value is the most recent but an intended historical call flow may.

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -211,5 +211,10 @@ export const predefined = {
     name: 'Exceeded maximum allowed subscriptions',
     code: -32608,
     message: 'Exceeded maximum allowed subscriptions'
-  })
+  }),
+  'UNSUPPORTED_HISTORICAL_EXECUTION': (blockId: string) => new JsonRpcError({
+    name: 'Unsupported historical block request',
+    code: -32000,
+    message: `Unsupported histoircal block identifier encounterd: ${blockId}`
```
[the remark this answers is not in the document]
Nana-EC: Resolved

packages/relay/src/lib/eth.ts
```diff
@@ -95,6 +95,7 @@ export class EthImpl implements Eth {
   private readonly ethCallCacheTtl = Number.parseInt(process.env.ETH_CALL_CACHE_TTL ?? constants.ETH_CALL_CACHE_TTL_DEFAULT.toString());
   private readonly ethBlockNumberCacheTtlMs = Number.parseInt(process.env.ETH_BLOCK_NUMBER_CACHE_TTL_MS ?? constants.ETH_BLOCK_NUMBER_CACHE_TTL_MS_DEFAULT.toString());
   private readonly ethGetBalanceCacheTtlMs = Number.parseInt(process.env.ETH_GET_BALANCE_CACHE_TTL_MS ?? constants.ETH_GET_BALANCE_CACHE_TTL_MS_DEFAULT.toString());
+  private readonly maxBlockRange = Number(process.env.MAX_BLOCK_RANGE ?? constants.MAX_BLOCK_RANGE.toString());
```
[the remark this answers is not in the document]

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
Nana-EC: I see. We're likely to both start there and build out as scenarios require.
What's the value in cases above 15m when the network limit is 15m?

Notably I did add a `CONTRACT_CALL_GAS_LIMIT` config

packages/relay/src/lib/eth.ts

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
mshakeg: @Nana-EC one usecase that I can think of is multicalls, e.g. say you want to make a 100 eth_calls where each of those calls costs ~1m gas you can instead combine all 100 into a single multicall eth_call and a `CONTRACT_CALL_GAS_LIMIT` of 100m should be sufficient.

packages/relay/src/lib/constants.ts
mshakeg: @Nana-EC hmm, how far away is historical execution support?

---
> Is it better to get the latest even when you request a historical, or fail on historical requests that are likely wrong

For my use case, there are potentially 2 types of `eth_call`, one that depends on block specific data(such as `block.timestamp`) and for this it isn't a major issue if it uses a timestamp slightly different(within ~10s at the `MAX_BLOCK_RANGE` of 5). However, the other type of eth_call is a function on storage/state and so if state changes occurred after the block specified and before or at `latest`(but within `MAX_BLOCK_RANGE`) then this may result in incorrect data being returned corrupting the subgraph's data or even breaking it.

So to account for this type of eth_call I'd recommend NOT using `MAX_BLOCK_RANGE`, but instead determining if any state changes have occurred since the requested block using the `timestamp` in `/api/v1/contracts/{contractIdOrAddress}/state` as you suggested.

And so I think leave the existing behaviour as `UNSUPPORTED_HISTORICAL_EXECUTION` is indeed quite disruptive.

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
Nana-EC: Understood.
Unfortunately, such a scenario would still hit the 15m gas limit per the current design.
We could revisit that limit on the mirror node later and match the relay accordingly.
For now I would advise conducting multi calls that adhere to this limit to avoid getting error responses.

packages/relay/src/lib/constants.ts
Nana-EC: Captured this in https://github.com/hashgraph/hedera-json-rpc-relay/issues/1299

ebadiere (commented):
LG

Nana-EC:
> @Nana-EC is this not super ceded by #1299 especially since there's a high likelihood of these changes halting subgraphs

Per the scenarios we highlighted earlier I removed the exception on blocks that are recent historical i.e. in the past but close to latest.
The only time `UNSUPPORTED_HISTORICAL_EXECUTION` is thrown for `eth_call` is if block is 0, 1 or `earliest`.
Are you saying one of these 3 scenarios would halt a subgraph?

mshakeg:
@Nana-EC oh didn't notice that change, in that case it shouldn't be an issue. 

Though new graph node deployments should specify a `GRAPH_ETHEREUM_GENESIS_BLOCK_NUMBER` greater than 1 and preferably closest to latest.

Nana-EC:
> @Nana-EC oh didn't notice that change, in that case it shouldn't be an issue.
> 
> Though new graph node deployments should specify a `GRAPH_ETHEREUM_GENESIS_BLOCK_NUMBER` greater than 1 and preferably closest to latest.

Even then should that matter? This is just for `eth_call`, so a graph node should still sync from block 0.
For `eth_getBlockByNumber` or `eth_getBlockByHash` it won't matter.
Also given the block time speed it's very unlikely any user will be able to submit any transaction at the beginning of a new network.
Let me know if there's still a graph concern here.

mshakeg:
@Nana-EC it shouldn't be an issue deploying graph nodes at the genesis "block", it's more to do with the fact that at least for the time being graph nodes won't be able to support subgraphs that require eth_calls where the block is not in range of `latest`. This issue will likely also affect existing subgraph deployments as there's bound to be a period of more than `5*2-ETHEREUM_POLLING_INTERVAL` where eth_calls fail(due to rate limiting for example).

mshakeg:
Hey @Nana-EC seems my last comment was missed. Could you please clarify:

> which issue tracks support for historical execution on the mirror node?

Nana-EC:
Hey @mshakeg we don't yet have tickets for historical execution. W're focused on the first 2 user stories of the HIP to unblock most scenarios.
Historical flows will be in follow up.

mshakeg:
Hey @Nana-EC my concerns seems to have come true, these changes(included in v0.25.0) have halted my mainnet subgraph at block [48999101](https://hashscan.io/mainnet/block/48999101), presumably around the time the relay was upgraded.

Nana-EC:
Hey @mshakeg which issue specifically are you hitting?
We spoke about `UNSUPPORTED_HISTORICAL_EXECUTION` but that shouldn't apply to your ongoing execution.

Can you open a bug with more details so we can investigate?

mshakeg:
@Nana-EC the issue is:

> This issue will likely also affect existing subgraph deployments as there's bound to be a period of more than (5*2)e3 - ETHEREUM_POLLING_INTERVAL where eth_calls fail(due to rate limiting for example) and even if this never occurs if ETHEREUM_POLLING_INTERVAL>(5*2)e3 then subgraphs on that graph node will very likely halt.

It's been quite a while and so the only proper way for the subgraph to recover properly is for the historical execution support.

Nana-EC:
@mshakeg if my understanding is clear you're not sayin there's a regression based on this PRs changes, rather you're saying the fact the the relay has yet to support historical blocks  on `eth_call` is what's blocking your instance correct?
If we could move this out of this ticket and into a separate ticket giving more details of your flow that would be good for tracking.

We have to finish `latest` block support and design and schedule historical support before implementing that.

mshakeg:
@Nana-EC while this isn't a regression, I would have preferred if #1299 was implemented over these changes in the short term as even though eth_calls(that specify a block) may have still occasionally returned incorrect data(especially when an eth_call depends on block specific data e.g. `block.timestamp`) my subgraph would still be mostly working and I could always correct any minor issues by re-indexing once historical execution is supported.

Wouldn't it be a bit redundant to create an issue considering hashgraph/hedera-mirror-node/issues/6164 exists and once implemented will resolve this issue?

Nana-EC:
Thanks clarifying. 
If https://github.com/hashgraph/hedera-mirror-node/issues/6164 captures the context then no need to create anything. Thanks

We'll explore scheduling #1299 .

PR #1314, opened 2023-05-31
Migrate from ethers v5 to v6

### LINKED ISSUE #1136: Migrate from ethers v5 to v6
### Problem

Bump the [ethersjs](https://www.npmjs.com/package/ethers) version to the latest

### Solution

Breaking changes that have to be fixed in the relay:
- `BigNumber` is replaced by `BigInt`
- contracts have become ES6 proxies `contract.populateTransaction.foo(addr)` -> `contract.foo.populateTransaction(addr)`
- utils transaction methods are included in Transaction object `utils.parseTransaction()` -> `Transaction.from(txBytes)`
- utilities has been changed, the key ones (`ethers.constants.AddressZero` -> `ethers.ZeroAddress`, `hexlify()` -> `toBeHex()`, `ethers.utils.hexValue()` -> `ethers.toQuantity()`, `ethers.utils.arrayify()` -> `ethers.getBytes()`)

The full migration guide can be found here https://docs.ethers.org/v6/migrating/

### Alternatives

_No response_

natanasow:
**Description**:

There is a new major ethersjs version with many breaking changes. We're trying to bump the version but the RLP encoder/decoder has been modified and v5 rlp encoded tx can not be decoded in v6 which causes SDK errors. Blocked till the SDK doesn't migrate to ethersjs v6. Another showstopper is the error handling and transaction reverting in ethers v6 which has been entirely changed.

A part of the migration guide can be found [here](https://docs.ethers.org/v6/migrating/), but it doesn't include RLP encoder/decoder and error handling :confused: .

I tried to bump the version only in the relay package (the server package runs ethers v5) and run the acceptance tests against the relay that running ethers v6 and they were all green. So if we're pushed to bump the version to v6, as a workaround we can bump only the relay's ethers version, and postponed the bumping of acceptance tests ethers version.

**Related issue(s)**:

Fixes #1136 

Child good-to-have issue: https://github.com/hashgraph/hedera-json-rpc-relay/issues/1617

**Blocked by**:
- https://github.com/hashgraph/hedera-services/issues/7670
- https://github.com/hashgraph/hedera-local-node/pull/376 - in `hedera-services` has been added a new `config.txt` property, and we have to edit the `config.txt` in the local node, then release a new version that will be used by the `ci` otherwise if we just bump the `services` and `mirror-node` versions in acceptance test's `index.spec.ts`, the node will not start due to `config.txt` error`

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -481,7 +479,7 @@ export class EthImpl implements Eth {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     this.logger.trace(`${requestIdPrefix} gasPrice()`);
     try {
-      let gasPrice: number | undefined = this.cache.get(constants.CACHE_KEY.GAS_PRICE);
+      let gasPrice: any = this.cache.get(constants.CACHE_KEY.GAS_PRICE);
```
ebadiere, line 483: Is type safety not a concern here anymore?  getFreeWeibars still returns a number, or am I missing something?

packages/relay/tsconfig.json
```diff
@@ -14,7 +14,8 @@
         "forceConsistentCasingInFileNames": true,
         "noImplicitAny": false,
         "declaration": true,
-        "strict": true
+        "strict": true,
+        "skipLibCheck": true
```
ebadiere, line 18: Why is this set to true now?

packages/relay/src/lib/eth.ts
natanasow: There were a lot of changes, and finally, it's ready for review again :upside_down_face:.

natanasow:
> Should we also update the ethers.js version in the dapp-example?

Yeah, but let's do it in another task/PR.

packages/server/tests/acceptance/htsPrecompile/tokenCreate.spec.ts
```diff
@@ -638,45 +637,32 @@ describe('@tokencreate HTS Precompile Token Create Acceptance Tests', async func
...
       }];
 
-      try {
-        const txXfer = await mainContract.cryptoTransferPublic(tokenTransferList);
-        expect((await txXfer.wait()).events.filter(e => e.event === Constants.HTS_CONTRACT_EVENTS.ResponseCode)[0].args.responseCode).to.equal(TX_SUCCESS_CODE);
-      } catch (error: any) {
-        expect(error.code).to.equal(Constants.CALL_EXCEPTION);
-        expect(error.reason).to.equal("transaction failed");
-      }
+      await Assertions.expectRevert(mainContract.cryptoTransferPublic(tokenTransferList), Constants.CALL_EXCEPTION);
```
Nana-EC, line 640: Q: is the reason no longer important to check or is it always the same?

packages/server/tests/acceptance/htsPrecompile/tokenManagement.spec.ts
```diff
@@ -327,40 +324,40 @@ describe('@tokenmanagement HTS Precompile Token Management Acceptance Tests', as
...
+      const token = Utils.convertEthersResultIntoStringsArray(tokenInfoBefore);
+      token[7] = [];
 
       setUpdatedValues(token);
 
       const txUpdate = await mainContractOwner.updateTokenInfoPublic(NftHTSTokenContractAddress, token, Constants.GAS.LIMIT_1_000_000);
-      expect((await txUpdate.wait()).events.filter(e => e.event === Constants.HTS_CONTRACT_EVENTS.ResponseCode)[0].args.responseCode).to.be.equal(TX_SUCCESS_CODE);
+      expect((await txUpdate.wait()).logs.filter(e => e.fragment.name === Constants.HTS_CONTRACT_EVENTS.ResponseCode)[0].args.responseCode).to.be.equal(TX_SUCCESS_CODE);
+
+      await new Promise(r => setTimeout(r, 2000));
```
Nana-EC, line 357: Q: why do we need a timeout?
Should the wait on the transaction and it's receipt be enough?

packages/server/tests/acceptance/rateLimiter.spec.ts
```diff
@@ -60,11 +58,18 @@ describe('@ratelimiter Rate Limiters Acceptance Tests', function () {
                     await new Promise(r => setTimeout(r, 1));
                 }
             }
-            const error = predefined.IP_RATE_LIMIT_EXCEEDED(testConstants.ETH_ENDPOINTS.ETH_CHAIN_ID);
 
-            await expect(sendMultipleRequests()).to.eventually.be.rejected.and.satisfy((err) => {
-                return [error.code, error.name, error.message].every(substring => err.body.includes(substring));
-            });
+            // TODO: adapt to the new ethers v6 JsonRpcProvider error handling
```
Nana-EC, line 62: Q: Is this still applicable?
If so please open a ticket to capture this effort next.

packages/server/tests/acceptance/ws/subscribe.spec.ts
```diff
@@ -559,9 +559,12 @@ describe('@web-socket Acceptance Tests', async function() {
             const invalidTopic = '0x9999999999999999999999999999999999999999999999999999999999999999';
 
             const testFilters = [
-                {}, // all contract logs
+                // currently, there is no way to gossip about all contract logs
```
Nana-EC, line 562: Q: WHat does this mean for the relay and the tests?
Is there a follow up here?

packages/server/tests/clients/relayClient.ts
```diff
@@ -61,10 +64,11 @@ export default class RelayClient {
             this.logger.trace(`${requestIdPrefix} [POST] to relay '${methodName}' with params [${params}] returned ${JSON.stringify(res)}`);
             Assertions.expectedError();
         } catch (err) {
-            if (expectedRpcError.name == "Internal error"){
-                expectedRpcError = predefined.INTERNAL_ERROR(err.message);
-            }
-            Assertions.jsonRpcError(err, expectedRpcError);
+            // TODO: adapt to the new ethers v6 JsonRpcProvider error handling
```
Nana-EC, line 67: Usually not a fan of TODO's in code and also leaving these around when we can fix them now.
However, I do see this is a large PR so agree with fixing it in the next PR.
Let's have 1 ticket to capture all the TODO's of "adapting to the new ethers v6 JsonRpcProvider error handling"

Nana-EC (changes requested):
Looking good.
Glad to see we're unblocked.
I left some comments and questions

packages/server/tests/acceptance/htsPrecompile/tokenCreate.spec.ts
natanasow: I tried to remove all of the copy-paste hardcoded checks and to add centralized asserts (some kind of mini refactoring).

packages/server/tests/acceptance/htsPrecompile/tokenManagement.spec.ts
natanasow: The receipt should be enough, will remove it, I guess it's a leftover from testing.

packages/server/tests/acceptance/rateLimiter.spec.ts
natanasow: I already did. It's in the description [here](https://github.com/hashgraph/hedera-json-rpc-relay/issues/1617) but with no labels :confused:.

packages/server/tests/clients/relayClient.ts
natanasow: [The ticket.](https://github.com/hashgraph/hedera-json-rpc-relay/issues/1617)

Nana-EC: Okay, let's remove the comments then since we have the ticket.

Nana-EC (approved):

PR #1321, opened 2023-06-02
Helm chart deployment for eth_subscribe server

### LINKED ISSUE #1296: New helm chart deployment for eth_subscribe server
### Problem

Currently the helm chart doesn't support deployment of the websocket server

### Solution

- Add a helm chart for deployment of websocket server

### Alternatives

_No response_

isavov:
**Description**:
This PR adds a new helm chart in order to support deployment of the eth_subscribe/websocket interface for the relay.
* Add new helm chart

**Related issue(s)**:

Fixes #1296 

**Notes for reviewer**:

**Checklist**


charts/hedera-json-rpc-relay-websocket/templates/configmap.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "json-rpc-relay-websocket.fullname" . }}
+  labels:
+    app:  {{ template "json-rpc-relay-websocket.name" . }}
+    {{ include "json-rpc-relay-websocket.labels" . | nindent 4 }}
+data:
```
Nana-EC, line 8: A lot of configs below are not utilized in the websocket server package.
Can you reduce this to only the ones utilized by the websocket server as the default configuration

charts/hedera-json-rpc-relay-websocket/templates/secret.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
Nana-EC, line 1: Q: Is this resource needed?
The websocket server streams from the mirror node only, so the below SDK config secrets should not be needed

charts/hedera-json-rpc-relay-websocket/values.yaml
```diff
@@ -0,0 +1,157 @@
...
+affinity: {}
+
+autoscaling:
+  enabled: false
+  minReplicas: 2
+  maxReplicas: 8
+  targetCPUUtilizationPercentage: 80
+  # targetMemoryUtilizationPercentage: 80
+
+config:
```
Nana-EC, line 11: SImilar to configmap, this should only include configs for the webserver socket

Nana-EC (changes requested):
Nice addition.
Please simplify to cover the websocket server only for this chart

Nana-EC (changes requested):
Nice addition.
Please simplify to cover the websocket server only for this chart

isavov:
After digging deeper into the used and unused ENV variables I've limited them to the absolutely necessary.
In the process I've discovered one unnecessary variable in the `hedera-json-rpc-relay` helm chart and removed it in this PR, as the fix was minor and doesn't need a separate PR

@Nana-EC What should the default value for `WS_MULTIPLE_ADDRESSES_ENABLED` be? I've set it to `false` for now, but I'm not sure if that would be best.

Nana-EC:
> After digging deeper into the used and unused ENV variables I've limited them to the absolutely necessary.
> 
> In the process I've discovered one unnecessary variable in the `hedera-json-rpc-relay` helm chart and removed it in this PR, as the fix was minor and doesn't need a separate PR
> 
> 
> 
> @Nana-EC What should the default value for `WS_MULTIPLE_ADDRESSES_ENABLED` be? I've set it to `false` for now, but I'm not sure if that would be best.

False is the right value to start with. We'll build confidence and when ready switch to true.

Nana-EC (dismissed):
LGTM

AlfredoG87 (changes requested):
LGTM,
just 1 request.

Could you include the new helm chart on the bump-version script, similar to how it is the prior one?

https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/scripts/.bump-version.js#L67-L85

AlfredoG87 (changes requested):
LGTM,
just 1 request.

Could you include the new helm chart on the bump-version script, similar to how it is the prior one?

https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/scripts/.bump-version.js#L67-L85

ebadiere (commented):
What does this mean?  Will I not be able to run tests against the ws server in the local node?

charts/hedera-json-rpc-relay-websocket/ci/default-values.yaml
```diff
@@ -0,0 +1,9 @@
+# Default values for json-rpc-relay in CI. Configured to testnet
```
Nana-EC, line 1: In a follow up PR we should add a test pod under this charts template to verify it on chart install in CI

Nana-EC:
> What does this mean? Will I not be able to run tests against the ws server in the local node?

Shouldn't affect local runs, just the chart being deployed in production flow.

AlfredoG87 (approved):
LGTM

ebadiere (dismissed):
LG

AlfredoG87 (approved):
LGTM

isavov:
> LGTM, just 1 request.
> 
> Could you include the new helm chart on the bump-version script, similar to how it is the prior one?
> 
> https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/scripts/.bump-version.js#L67-L85

Thanks for including it in the bump script, I completely missed that 🙌

charts/hedera-json-rpc-relay-websocket/ci/default-values.yaml
Nana-EC: @isavov making sure you saw this one to add it to any follow up PRs

PR #1334, opened 2023-06-06
Fixed test to use correct contract to estimate gas.  Gas is estimated for running the ping method which belongs to the basic contract, not the mirror contract.

### LINKED ISSUE #1317: The batch2 spec test @release should execute "eth_estimateGas" for contract call uses the wrong contractId causing the wrong gas estimate to be returned.
### Description

The following snippet fixes it, but should be set up properly in the beforeAll function of the test suite:

```
           it.only('@release should execute "eth_estimateGas" for contract call', async function () {
                const basicContract = await servicesNode.deployContract(basicContractJson);
                const expectedRes = `0x${(400000).toString(16)}`;
                const res = await relay.call(RelayCalls.ETH_ENDPOINTS.ETH_ESTIMATE_GAS, [{
                    to: basicContract.evm_address,
                    data: BASIC_CONTRACT_PING_CALL_DATA
                }], requestId);
                expect(res).to.contain('0x');
                // add alternative for the assert, because we get race condition in CI and constants is not initialized sometimes.
                expect(res).to.equal(EthImpl.defaultTxGas || expectedRes);
            });
```

### Steps to reproduce

1. Run the current test again the local node or any environment.
2. Observe the failure.

### Additional context

_No response_

### Hedera network

previewnet

### Version

v0.25.0

### Operating system

None

ebadiere:
**Related issue(s)**:

Fixes #1317 

### DIFF
--- a/packages/server/tests/acceptance/rpc_batch2.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch2.spec.ts
@@ -130,7 +130,7 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
                 data: BASIC_CONTRACT_PING_CALL_DATA
             }], requestId);
             expect(estimatedGas).to.contain('0x');
-            expect(estimatedGas).to.equal(EthImpl.defaultTxGas || expectedRes);
+            expect(estimatedGas).to.equal(expectedRes);
         });
 
         it('@release should execute "eth_estimateGas" for existing account', async function() {
--- a/packages/server/tests/acceptance/rpc_batch2.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch2.spec.ts
@@ -109,6 +109,13 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
     });
 
     describe('eth_estimateGas', async function () {
+        let basicContract;
+
+        this.beforeAll(async () => {
+            basicContract = await servicesNode.deployContract(basicContractJson);
+        });
+
+
         it('@release should execute "eth_estimateGas"', async function () {
             const res = await relay.call(RelayCalls.ETH_ENDPOINTS.ETH_ESTIMATE_GAS, [{}], requestId);
             expect(res).to.contain('0x');
@@ -119,11 +126,11 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
         it('@release should execute "eth_estimateGas" for contract call', async function() {
             const expectedRes = `0x${(400000).toString(16)}`;
             const estimatedGas = await relay.call(RelayCalls.ETH_ENDPOINTS.ETH_ESTIMATE_GAS, [{
-                to: mirrorContract.evm_address,
+                to: basicContract.evm_address,
                 data: BASIC_CONTRACT_PING_CALL_DATA
             }], requestId);
             expect(estimatedGas).to.contain('0x');
-            expect(Number(estimatedGas)).to.be.gt(0);
+            expect(estimatedGas).to.equal(EthImpl.defaultTxGas || expectedRes);
         });
 
         it('@release should execute "eth_estimateGas" for existing account', async function() {


packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -119,11 +126,11 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
         it('@release should execute "eth_estimateGas" for contract call', async function() {
             const expectedRes = `0x${(400000).toString(16)}`;
             const estimatedGas = await relay.call(RelayCalls.ETH_ENDPOINTS.ETH_ESTIMATE_GAS, [{
-                to: mirrorContract.evm_address,
+                to: basicContract.evm_address,
```
Nana-EC, line 129: Q: Why are we changing the contract from `mirrorContract` to `basicContract`?

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -119,11 +126,11 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
...
                 data: BASIC_CONTRACT_PING_CALL_DATA
             }], requestId);
             expect(estimatedGas).to.contain('0x');
-            expect(Number(estimatedGas)).to.be.gt(0);
+            expect(estimatedGas).to.equal(EthImpl.defaultTxGas || expectedRes);
```
Nana-EC, line 133: Q: Why  are we expecting 2 different values.
If the test passes or fails for one of them we won't know why.
Can we break this out to be very clear when the test expects one over the other. 
Maybe add a comment also.

Nana-EC (changes requested):
Not clear on some of the changes.
Can you update the PR description and maybe add some comments if useful based on responses to questions.

packages/server/tests/acceptance/rpc_batch2.spec.ts
Ivo-Yankov: Both `EthImpl.defaultTxGas` and `expectedRes` seem to have the same value, maybe one of two can be removed.

Ivo-Yankov (commented):
Added a nit comment, otherwize LGTM

packages/server/tests/acceptance/rpc_batch2.spec.ts
ebadiere: Yes, they are the same.  That was added in a previous refactor, but I'll remove one.

packages/server/tests/acceptance/rpc_batch2.spec.ts
ebadiere: The data used to estimate the gas is the encoded ping function of the basic contract, not the mirror contract.  It's represented by the literal below BASIC_CONTRACT_PING_CALL_DATA.  When the basic contract is used the expected estimated gas is returned.

ebadiere: Updated the PR description.   Here is the contents of the basic contract: 
```
//SPDX-License-Identifier: Unlicense
pragma solidity ^0.8.0;

contract Basic {

    constructor() {}

    function ping() public pure returns (int) {
        return 1;
    }

    function destroy() public {
        selfdestruct(payable(msg.sender));
    }

    receive() external payable {}
}
```
The encoded ping function:
```
const BASIC_CONTRACT_PING_CALL_DATA = '0x5c36b186';
```
Reference: https://www.moesif.com/solidity-abi-hex-decoder/decode


So the test should use the basic contract not the mirrornode contract:
```
            const estimatedGas = await relay.call(RelayCalls.ETH_ENDPOINTS.ETH_ESTIMATE_GAS, [{
                to: basicContract.evm_address,
                data: BASIC_CONTRACT_PING_CALL_DATA
            }], requestId);
```

Nana-EC (approved):

PR #1335, opened 2023-06-07
Bump Synpress Version

georgi-l95:
**Description**:
This PR bumps all testing tools versions to the latest available. This will resolve some vulnerability issues.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


dapp-example/tests/e2e/specs/index.spec.js
```diff
@@ -85,6 +85,12 @@ describe('Test Core Hedera User Scenarios', function() {
   });
 
   function mandatoryTests() {
+
+    beforeEach(() => {
```
Nana-EC, line 89: Q: What's this doing and how come it's needed?

Nana-EC (commented):
LG.
Curious of the connect account addition. What's the context

dapp-example/tests/e2e/specs/index.spec.js
georgi-l95: Yes, for some reason it clears the context of the page after each test with the latest versions of the plugin. So we need to visit the page before each tests.
There are like hundreds of changes from version 3.0.5 to 3.7.1 and most of them are breaking, so I'm not sure where this behaviour was introduced.

Nana-EC: Okay any comments would be valuable. Not sure of the operation but seems like it's saying before every tests navigate to home and if there's a "Connect Account" item then click it

Nana-EC (approved):

PR #1340, opened 2023-06-09
Optimize mirror node get accounts call with limit

### LINKED ISSUE #1339: Optimize Mirror Node getAccountCall with limit 1
### Problem

The Mirror node getAccount calls `api/v1/accounts/{address}` also retrieves an accounts transactions.
This may be adding to call latency and is not needed in most cases

### Solution

Update calls to set a limit 1 to reduce mirror node work in pulling transactions

### Alternatives

_No response_

Nana-EC:
**Description**:
Mirror node calls for account has the unintended effect of getting the page limit of tranactions

- Updated get account call to use `order=desc` and `limit = 1`
- Updated spect ts
- Add missing mocks in some spec tests

**Related issue(s)**:

Fixes #1339 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18791,7 +18791,7 @@
         },
         "packages/relay": {
             "name": "@hashgraph/json-rpc-relay",
-            "version": "0.25.0-SNAPSHOT",
+            "version": "0.26.0-SNAPSHOT",
             "dependencies": {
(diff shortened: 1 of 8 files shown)


georgi-l95 (approved):
LG

AlfredoG87 (approved):
LGTM

PR #1345, opened 2023-06-09
Refactor eth_sendRawTransaction to use mirror node for tx hash

### LINKED ISSUE #1333: Update eth_sendRawTransaction to utilize Mirror Node on transaction record calls
### Problem

Currently the relay primarily calls the consensus node to retrieve the `TransactionRecord` to retrieve the `ethereumHash` to return.
However, this is not performant for the relay balance as it can be retrieved from the Mirror Node

### Solution

Call the `api/v1/contracts/results/{transactionIdOrHash}` and return the `hash` as this is the value

Note, `TransactionRecord` was also used to get the fee charged.
This can be retrieve from the `api/v1/contracts/results/{transactionId}` using the `charged_tx_fee` value

### Alternatives

_No response_

### LINKED ISSUE #1336: eth_sendRawTransaction is returning hash when receiving wrong_nonce (bigger than expected) instead of error
### Description

Currently if sending a transaction using a wrong nonce, there are 2 possibilities:
1. Nonce is too low, we are returning the appropriate message
2. Nonce is too big (bigger than the expected current nonce). We are returning a computed hash for the transaction, but that hash does not exist, since the SDK returned `wrong_nonce` 

### Steps to reproduce

 you can use this transaction on `previewnet` to replicate the issue.
 
 ```
0xf8ed818b85f7f6f25a008301c3e89410df38c7381cf2154a0fc3fb5455ab60b0483b0b80b884a4136862000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000304772656574696e67732066726f6d204175746f6d617465642054657374204e756d626572203133392c2048656c6c6f2100000000000000000000000000000000820276a0029a12dad96a95df97926ad0e044092ed7faf02948a16022818a0814523592cea07cedecff37a0ffd57ef9653580b118d79350c240f8e3923505e395764f3d59d5
 ```

```
curl --location 'https://previewnet.hashio.io/api' \
--data '{
    "method": "eth_sendRawTransaction",
    "params": [
        "0xf8ed818b85f7f6f25a008301c3e89410df38c7381cf2154a0fc3fb5455ab60b0483b0b80b884a4136862000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000304772656574696e67732066726f6d204175746f6d617465642054657374204e756d626572203133392c2048656c6c6f2100000000000000000000000000000000820276a0029a12dad96a95df97926ad0e044092ed7faf02948a16022818a0814523592cea07cedecff37a0ffd57ef9653580b118d79350c240f8e3923505e395764f3d59d5"
    ],
    "id": 120,
    "jsonrpc": "2.0"
}'
```

### Additional context

This transactions are also logging an error that the transaction record was not found (since the sdk did not processed the transaction due to wrong nonce, but still the relay is returning the `computed hash`

image

image

There are 2 ways to avoid this.

1. To verify before submitting the transaction to the SDK if the nonce is too high.
2. To handle the response from the SDK `WRONG_NONCE` and return an appropriate error message to the user.

The logic in question is happening here: https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1102C5-L1127

### Hedera network

mainnet, testnet, previewnet

### Version

v0.25.0

### Operating system

None

Ivo-Yankov:
**Description**:

- Refactors `eth_sendRawTransaction`. It now uses the mirror node to get the transaction hash. 
- As a result `eth_sendRawTransaction` no longer spends the operator's Hbars. An obsolete rate limitter test was removed.
- Adds a new method to the MirrorNodeClient class - `repeatedRequest`. It wraps any of the other class methods and in the case of 404 it repeats the request. So if for example the `getContractResult` is run and after all of the axios-retry iterations it still returns 404, then the whole method execution will be repeated. This basically extends the retry logic without increasing the maximum retries for the whole relay.
- Returns a formatted `WRONG_NONCE` error if it is thrown in Consensus during precheck. Note that this is different than the precheck performed in the relay and this case can only happen when several asynchronous transactions with the same nonce have been submitted at the same time. The relay precheck will pass but the consensus one will fail and as a result no ContractResult will be created in the Mirror node. This failed transaction cannot be read with `eth_getTransactionReceipt`.

**Related issue(s)**:

Fixes #1333 
Fixes #1336 

**Notes for reviewer**:

**Checklist**


Nana-EC:
@Ivo-Yankov and @AlfredoG87 
I think a better approach would be

1. Before any checks calculate the hash and check for the hash in a cache or from mirror node  `api/v1/contract/results/{hash}` for a valid transaction
2. If found return a new error `DUPLICATE_TRANSACTION` as the transaction already exists on the ledger
3. On successful transactions after submission to consensus node but before record retrieval cache the hash to support the above check

I don't we should be treating `WRONG_NONCE` in a special way as it's just a transaction that is acceptable response. What we should be doing is resubmitting a transaction the network has already come to consensus on

Nana-EC:
> @Ivo-Yankov and @AlfredoG87 I think a better approach would be
> 
> 1. Before any checks calculate the hash and check for the hash in a cache or from mirror node  `api/v1/contract/results/{hash}` for a valid transaction
> 2. If found return a new error `DUPLICATE_TRANSACTION` as the transaction already exists on the ledger
> 3. On successful transactions after submission to consensus node but before record retrieval via mirror node cache the hash to support the above check
> 
> I don't we should be treating `WRONG_NONCE` in a special way as it's just a transaction that is acceptable response. What we should be doing is resubmitting a transaction the network has already come to consensus on

@Ivo-Yankov you and @AlfredoG87 were right here. `WRONG_NONCE` should not return a hash. It seems to not return a ContractResult and therefore future looks ups for transaction from the mirror node would fail.

We should add a isWrongNonce to the sdkCLientError, if it's true then throw the error instead of returning the hash

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -911,4 +911,20 @@ export class MirrorNodeClient {
...
+        let result;
+        for (let i = 0; i < repeatCount; i++) {
+            result = await this[methodName](...args, requestId);
+            if (result) {
+                break;
+            }
+        }
+        return result;
+    }
 }
```
AlfredoG87, line 930: I like this method, I would like if we could disable the axios default retries. 
why? some logic flows are indeed expecting a `404` code as part of their conditions, and in those cases the axios default retry is adding a 3s latency unneeded before returning the expected and correct `404` code.
Also I think that the retries should be intended on certain cases and codes in specific logic flows and not for all methods as is currently with the axios interceptor implementation.
So I think this method takes in that direcction and once we use it with intention on the needed places, we can just set the amount of retries to 0 and effectively disabled it.

AlfredoG87: Another thing, it might not be needed right now, but I think the method should have a try..catch and identify `404`.
Right now, only some methods might `throw` an exception instead of returning `null` when receiving a `404` from the mirrorNode, although I believe that the `mirrorNodeClient` should always return the `404` and let the consuming method to handle the `404` as it sees fit.

packages/relay/src/lib/eth.ts
```diff
@@ -1106,57 +1107,66 @@ export class EthImpl implements Eth {
...
-   * @param transaction
-   */
-  async sendRawTransaction(transaction: string, requestId?: string): Promise<string | JsonRpcError> {
+  async parseRawTxAndPrecheck(transaction: string, requestId?: string){
     const requestIdPrefix = formatRequestIdMessage(requestId);
-    this.counter.labels(EthImpl.ethSendRawTransaction, transaction.substring(0,10)).inc();
-
     let interactingEntity = '';
     let originatingAddress = '';
+    let parsedTx: EthersTransaction;
```
Nana-EC, line 1114: Q: why move it up? It's not consumed outside of the try?

packages/relay/src/lib/eth.ts
```diff
@@ -1106,57 +1107,66 @@ export class EthImpl implements Eth {
...
+   *
+   * @param transaction
+   */
+  async sendRawTransaction(transaction: string, requestId?: string): Promise<string | JsonRpcError> {
+    const requestIdPrefix = formatRequestIdMessage(requestId);
+    this.counter.labels(EthImpl.ethSendRawTransaction, transaction.substring(0,10)).inc();
 
+    const parsedTx = await this.parseRawTxAndPrecheck(transaction, requestId);
     const transactionBuffer = Buffer.from(EthImpl.prune0x(transaction), 'hex');
+    const computedHash = EthImpl.prepend0x(createHash('keccak256').update(transactionBuffer).digest('hex'));
```
Nana-EC, line 1141: Move this to the last catch since that's the only place it seems to be used. We should avoid calculating this if there are flows that won't require the calculation

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -911,4 +911,20 @@ export class MirrorNodeClient {
...
+
+    /**
+     * This method is intended to be used in cases when the default axios-retry settings do not provide
+     * enough time for the expected data to be propagated to the Mirror node.
+     * It provides a way to have an extended retry logic only in specific places
+     */
+    public async repeatedRequest(methodName: string, args: any[], repeatCount: number, requestId?: string) {
+        let result;
+        for (let i = 0; i < repeatCount; i++) {
+            result = await this[methodName](...args, requestId);
```
Nana-EC, line 923: Q: shouldn't there be some kind of backoff before retry?
Should you be waiting `MIRROR_NODE_RETRY_DELAY` ms?

Nana-EC (commented):
Non blocking questions and usggestions

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: yeah, I thought the same.

packages/relay/src/lib/eth.ts
```diff
@@ -1106,57 +1107,66 @@ export class EthImpl implements Eth {
...
+      const formattedId = formatTransactionId(txId);
+      const  record = await this.mirrorNodeClient.repeatedRequest('getContractResult', [formattedId], 3, requestId);
+      if (!record) {
+        this.logger.warn(`${requestIdPrefix} No record retrieved`);
+        const tx = await this.mirrorNodeClient.getTransactionById(txId, 0, requestId);
+        if (tx.transactions?.length) {
+          const result = tx.transactions[0].result;
+          if (result === 'WRONG_NONCE') {
+            const accountInfo = await this.mirrorNodeClient.getAccount(parsedTx.from!, requestId)
+            throw predefined.NONCE_TOO_LOW(parsedTx.nonce, accountInfo.ethereumNonce);
```
AlfredoG87, line 1158: In case the NONCE is bigger than expected we receive `WRONG_NONCE`, should we return an appropiate error, instead of NONCE_TOO_LOW?

Ivo-Yankov: Good point, I added a `NONCE_TOO_HIGH` error

packages/relay/src/lib/clients/mirrorNodeClient.ts
Ivo-Yankov: Added backoff

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -911,4 +911,24 @@ export class MirrorNodeClient {
...
+            if (result) {
+                break;
+            }
+
+            // Backoff before repeating request
+            const mirrorNodeRetryDelay = parseInt(process.env.MIRROR_NODE_RETRY_DELAY || '250');
```
Nana-EC, line 929: nit: could pull this out to a const so it's calculated once

packages/relay/src/lib/eth.ts
```diff
@@ -1109,69 +1110,96 @@ export class EthImpl implements Eth {
...
+    const transactionBuffer = Buffer.from(EthImpl.prune0x(transaction), 'hex');
+
+    let txSubmitted = false;
+    try {
+      const contractExecuteResponse = await this.hapiService.getSDKClient().submitEthereumTransaction(transactionBuffer, EthImpl.ethSendRawTransaction, requestId);
+      txSubmitted = true;
+      // Wait for the record from the execution.
+      let txId = contractExecuteResponse.transactionId.toString();
+      const formattedId = formatTransactionId(txId);
+      const  record = await this.mirrorNodeClient.repeatedRequest('getContractResult', [formattedId], 3, requestId);
```
Nana-EC, line 1176: nit: number of retries should be configureable item

Nana-EC, line 1176: nit: `'getContractResult'` should be a const noted in MirrorNodeClient that matches the API name it exposes

packages/relay/src/lib/eth.ts
```diff
@@ -1109,69 +1110,96 @@ export class EthImpl implements Eth {
...
+      if (!record) {
+        this.logger.warn(`${requestIdPrefix} No record retrieved`);
+        const tx = await this.mirrorNodeClient.getTransactionById(txId, 0, requestId);
+        if (tx.transactions?.length) {
+          const result = tx.transactions[0].result;
+          if (result === 'WRONG_NONCE') {
```
Nana-EC, line 1182: `'WRONG_NONCE'` should also be a constant

packages/relay/src/lib/eth.ts
```diff
@@ -1109,69 +1110,96 @@ export class EthImpl implements Eth {
...
+            const accountInfo = await this.mirrorNodeClient.getAccount(parsedTx.from!, requestId)
+            const accountNonce = accountInfo.ethereum_nonce;
+            if (parsedTx.nonce > accountNonce) {
+              throw predefined.NONCE_TOO_HIGH(parsedTx.nonce, accountNonce);
+            }
+
+            throw predefined.NONCE_TOO_LOW(parsedTx.nonce, accountNonce);
+          }
+        }
+        throw predefined.INTERNAL_ERROR();
```
Nana-EC, line 1192: nit: add a message in the error to note there was no matching transactions.

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -111,6 +111,11 @@ export const predefined = {
     code: 32001,
     message: `Nonce too low. Provided nonce: ${nonce}, current nonce: ${currentNonce}`
   }),
+  'NONCE_TOO_HIGH': (nonce, currentNonce) => new JsonRpcError({
```
Nana-EC, line 114: Could also repurpose `INCORRECT_NONCE` error in this file if desired. 
Optionally, we can leave as is and later just remove `INCORRECT_NONCE` as it's not used.

Nana-EC (dismissed):
LG, lots of non blocking items that i'd like addressed in 0.27 but aren't necessary to hold back pushing this into 0.26

packages/relay/src/lib/errors/JsonRpcError.ts
AlfredoG87: so instead of using `NONCE_TOO_HIGH` or `NONCE_TOO_LOW`, we just use the `INCORRECT_NONCE` right?

packages/relay/src/lib/eth.ts
AlfredoG87: ✅  done

packages/relay/src/lib/eth.ts
AlfredoG87: instead of declaring a const, I am referencing the name of the enum defined at the response protobuf. using the hashgraph lib
https://github.com/hashgraph/hedera-protobufs/blob/c1de1a199d11e6ace89679af9e4500cb92eb841c/services/response_code.proto#L152

Let me know your thoughts.

packages/relay/src/lib/eth.ts
AlfredoG87: `getContractResult` is a method name, that will corresponds to `contracts/results/` so its  value should match the name of the method to work. and setting the Const name as contractsResults o `CONTRACTS_RESULTS_METHOD_NAME` does not seem to describe the purpose of the parameter.

Although I agree on referencing instead of hardcoding the value, so I have got the name from the function that is being called itself.

let me know your thoughts.

AlfredoG87: ✅ Done

packages/relay/src/lib/errors/JsonRpcError.ts
AlfredoG87: I think that remove `INCORRECT_NONCE` would be better, so the errors are more explicitly how is right now.
✅ Done

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: ✅  Done!

Nana-EC (approved):

packages/relay/src/lib/eth.ts
AlfredoG87: I nevertheless  created its own constants were we can put the values we use of the transactionEnum response status.

AlfredoG87 (approved):

PR #1347, opened 2023-06-09
Remove `from` check from `eth_call` prechecks

### LINKED ISSUE #1346: Remove unnecessary mirror node call when calling `eth_call`
### Problem

Now with the archive node, we are using only the mirror-node for `eth_call` and since there are already existing checks for the from field, we don't need to check it with a mirror-node call.
Previously we needed to check to avoid unnecessary consensus node, if the from field doesn't exist. But now we can make calls to mirror node with wrong from field without any issues as in the evm it doesn't matter if it exists.

### Solution

Remove the from check in the `eth_call` prechecks.

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR removes the from check when we call `eth_call`. More context: 

Now with the archive node, we are using only the mirror-node for `eth_call` and since there are already existing checks for the from field, we don't need to check it with a mirror-node call.
Previously we needed to check to avoid unnecessary consensus node, if the from field doesn't exist. But now we can make calls to mirror node with wrong from field without any issues as in the `evm` it doesn't matter if it exists.

**Related issue(s)**:

Fixes #1346 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1189,7 +1189,7 @@ export class EthImpl implements Eth {
       this.ethExecutionsCounter.labels(EthImpl.ethCall, call.data.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH)).inc();
 
     
-    const to = await this.performCallChecks(call, blockParam, requestId);
+    await this.performCallChecks(call, blockParam, requestId);
 
     // Get a reasonable value for "gas" if it is not specified.
     const gas = this.getCappedBlockGasLimit(call.gas, requestId);
@@ -1200,7 +1200,7 @@ export class EthImpl implements Eth {
       if ((process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE === 'undefined') || (process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE == 'false')) {
         //temporary workaround until precompiles are implemented in Mirror node evm module
         // Execute the call and get the response
-        return await this.callMirrorNode(call, to, gas, value, requestId);
+        return await this.callMirrorNode(call, gas, value, requestId);
       }
       
       return await this.callConsensusNode(call, gas, requestId);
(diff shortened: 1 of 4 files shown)


ebadiere (dismissed):
LGTM.

packages/relay/src/lib/eth.ts
```diff
@@ -1281,14 +1281,6 @@ export class EthImpl implements Eth {
       throw predefined.INVALID_CONTRACT_ADDRESS(call.to);
     }
 
-    // If "From" is distinct from blank, we check is a valid account
```
Nana-EC, line 1284: Q: Do you think we should totally remove it?
I think we should only remove it when the call has no value i.e. it's not for a transfer. In that case it doesn't matter. If the call has value then we should still check `from` fail fast and save us and mirror node the overhead of an unnecessary call.

Nana-EC (changes requested):
Agreed with the direction i think we should minimize the check to value transferring scenarios only and not totally remove.

packages/relay/src/lib/eth.ts
georgi-l95: We could check the from address, if value is filled. But what do you think about removing all checks, `to` address field as well? Web3 module is currently doing all those check either way when we do the request and what we are doing it the relay is double work. So basically in all scenarios (we have not existing `from` or `to` or whatever) we are doing only one mirror-node request and that's it, compared to now -> in the best case 2 calls, in worst case 3 calls.
While keeping in mind, that we still have filed validations ( you can't pass incorrect address in `to`/`from` field, because it needs to be 20 bytes).

Nana-EC: Yeah i thin we should conditionally check the `from` field only if it's a `value` transfer case.

Nana-EC: Ah i think i'm seeing it now. I forgot the validations require a Mirror Node call.
okay, if we're calling the mirror node just to verify whether to call the mirror node when it's gonna make the same check then I agree let's not do double work in either `from` or `to` case.

Maybe move the check only to the fall back consensus node case

AlfredoG87: I agree that we don't need to call 3 times the mirror node.

AlfredoG87: I have moved the validation to `callConsensusNode`

packages/relay/src/lib/eth.ts
```diff
@@ -1270,7 +1270,15 @@ export class EthImpl implements Eth {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     // Execute the call and get the response
     this.logger.debug(`${requestIdPrefix} Making eth_call on contract ${call.to} with gas ${gas} and call data "${call.data}" from "${call.from}" using consensus-node.`, call.to, gas, call.data, call.from);
-    
+
+    // If "From" is distinct from blank, we check is a valid account
+    if (call.from) {
+      const fromEntityType = await this.mirrorNodeClient.resolveEntityType(call.from, [constants.TYPE_ACCOUNT], requestId);
```
Nana-EC, line 1276: nit: for a future PR we should just substitute this for a simple getAccount call

Nana-EC (dismissed):
LG, thanks team

packages/relay/src/lib/eth.ts
AlfredoG87: agreed.

Nana-EC (approved):

ebadiere (approved):
LG.

PR #1351, opened 2023-06-10
Reorganize latency calculation logic

### LINKED ISSUE #1348: Fix request latency metric calculate
### Problem

Currently the latency metric for E2E requests is spread out and hard to follow.
This may contribute to the missing latency chunk of the flow making it smaller than it should be

### Solution

Revisit and fix metric logic

### Alternatives

_No response_

Nana-EC:
**Description**:
Simplify relay E2E latency calculation

- move E2E log into single handler
- centralize happy path in server.ts
- update error flows in index.ts for clear messages

**Related issue(s)**:

Fixes #1348 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/src/koaJsonRpc/index.ts
+++ b/packages/server/src/koaJsonRpc/index.ts
@@ -50,6 +50,8 @@ const IP_RATE_LIMIT_EXCEEDED = "IP RATE LIMIT EXCEEDED";
 const JSON_RPC_ERROR = "JSON RPC ERROR"
 const METHOD_NOT_FOUND = "METHOD NOT FOUND";
 
+const responseSuccessStatusCode = '200';
+
 export default class KoaJsonRpc {
   private registry: any;
   private registryTotal: any;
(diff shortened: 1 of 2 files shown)


packages/server/src/koaJsonRpc/index.ts
```diff
@@ -111,8 +113,8 @@ export default class KoaJsonRpc {
         return;
       }
 
+      ctx.state.methodName = body.method;
       const methodName = body.method;
-      const messagePrefix = `${requestIdPrefix} [POST] ${methodName}:`;
```
ebadiere, line 115: Removing this line causes the requestIdPrefix on line 98 above to no longer be needed.  It can also be removed.

packages/server/src/server.ts
```diff
@@ -202,10 +209,6 @@ const logAndHandleResponse = async (methodName: any, methodParams: any, methodFu
     }
     return response;
   } catch (e: any) {
-    ms = Date.now() - app.getStartTimestamp();
-    methodResponseHistogram.labels(methodName, responseInternalErrorCode).observe(ms);
```
ebadiere, line 206: Removing these calls to getStartTimestamp no longer requires the getStartTimestamp method in the koaJsonRpc, so it can be removed as well.

Nana-EC: Removed, thanks

packages/server/src/koaJsonRpc/index.ts
Nana-EC: Good catch. Removed

ebadiere (approved):
LG.

PR #1361, opened 2023-06-14
1357 - Added max transaction count that can be returned on an eth_get…

### LINKED ISSUE #1357: 502s are returned on some accounts with thousands of transactions in a block.  An error should be returned.
### Description

502s are returned with some requests against testnet.  The release and K6 performance tests may trigger these.  If so we'll need to update the accounts/addresses used.

### Steps to reproduce

The following curl request will reproduce it.

```
curl -X POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":"2","method":"eth_getBlockByHash","params":["0xfd601bc88c27e104f52cc35c2397c5bda69fd47b808d0ec59f0987283df62b5d", true]}'  https://testnet.hashio.io/api
```

```
<html><head>
<meta http-equiv="content-type" content="text/html;charset=utf-8">
<title>502 Server Error</title>
</head>
<body text=#000000 bgcolor=#ffffff>
<h1>Error: Server Error</h1>
<h2>The server encountered a temporary error and could not complete your request.<p>Please try again in 30 seconds.</h2>
<h2></h2>
</body></html>
```
Timing details:

```
Establish Connection: 0.209954s
TTFB: 0.237018s
Total: 30.305170s
0.01s user 0.01s system 0% cpu 30.210 total
```

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.25.0

### Operating system

None

ebadiere:
Added transaction count max to getBlock when transaction objects are set to true, to return the transactions in the block.

**Related issue(s)**:

Fixes #1357 


docs/configuration.md
```diff
@@ -53,8 +53,12 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
-| `ETH_GET_LOGS_BLOCK_RANGE_LIMIT`      | "1000"        | The maximum block number range to consider during an eth_getLogs call.                                                                                                                                                                                                                                 |
+| `ETH_CALL_DEFAULT_TO_CONSENSUS_NODE ` | "false"       | Flag to set if eth_call logic should first query the mirror node.                                                                                                             
+| `ETH_GET_LOGS_BLOCK_RANGE_LIMIT`      | "1000"        | The maximum block number range to consider during an eth_getLogs call.                                                                                  
+
+| `ETH_GET_TRANSACTION_COUNT_MAX_BLOCK_RANGE`      | "1000"   | The maximum number of transactions to return when running eth_getBlockByHash with transaction objects set to true call.
```
Nana-EC, line 59:
```suggestion
| `ETH_GET_TRANSACTION_COUNT_MAX_BLOCK_RANGE`      | "1000"   | The maximum number of transactions to return when running eth_getBlockByHash or eth_getBlockByNumber with transaction objects set to true call.
```

packages/relay/src/lib/eth.ts
```diff
@@ -1618,6 +1619,11 @@ export class EthImpl implements Eth {
       // depending on stage of contract execution revert the result.to value may be null
       if (!_.isNil(result.to)) {
         if(showDetails) {
+          // check the size of the block before querying for transaction details
+          if (blockResponse.count >= this.ethGetTransactionCountMaxBlockRange) {
```
Nana-EC, line 1623: Not sure this is the right check. 
If we're just checking the blockSize then this could go much higher in this method to save on API calls and logic.
`blockResponse.count` is the number of transaction overall in the block, however, on Hedera that also includes non-EVM transactions.

What we should be doing is checking the number of EVM transaction (`contractResults`) to see if it's larger than `this.ethGetTransactionCountMaxBlockRange`.

docs/configuration.md
Nana-EC, line 59: 1000 might be too high. Maybe 100 to start with
```suggestion
| `ETH_GET_TRANSACTION_COUNT_MAX_BLOCK_RANGE`      | "100"   | The maximum number of transactions to return when running eth_getBlockByHash with transaction objects set to true call.
```

Nana-EC (changes requested):
Need to update the check logic.
Additionally would need a test in `eth.spec.ts` to confirm the logic

docs/configuration.md
ebadiere: Done.

packages/relay/src/lib/eth.ts
ebadiere: Done.

Nana-EC (changes requested):
Looks good, you're still missing a test in `eth.spec.ts` to verify this

Nana-EC (commented):
I see you added the tests.
However, I think there's a bad merge here. There's lots of extra files that don't seem related to your changes. I presume they are already checked in changes but it's not clear.
You should rebase

ebadiere:
When I rebased I got 
```
 [rejected]        1357-thousands-of-txns-in-a-block -> 1357-thousands-of-txns-in-a-block (non-fast-forward)
error: failed to push some refs to 'https://github.com/hashgraph/hedera-json-rpc-relay.git'
```
I created a new branch with the changes.  

https://github.com/hashgraph/hedera-json-rpc-relay/pull/1372

PR #1366, opened 2023-06-15
Remove AcceptedErrorCodes from MirrorNodeClient - GET_ACCOUNTS_BY_ID_ENDPOINT

### LINKED ISSUE #1276: Reduce accepted error codes on mirror node contracts/call
### Problem

Currently the relay will accept error code in `acceptedErrorStatusesResponsePerRequestPathMap[MirrorNodeClient.CONTRACT_CALL_ENDPOINT]` and not throw an error.
This may be problematic as it would return a `0x` value instead of throwing to clients.

### Solution

Verify that `0x` is not an appropriate value to return on any of these codes.
Assuming so reduce the array of codes to an empty array `[]`

In this way all error from the mirror node would result in errors to the user.

Also remove the `0x` return in `callMirrorNode()` if not applicable

Endpoints to refactor:

- [x] GET_ACCOUNTS_BY_ID_ENDPOINT
- [x] GET_BALANCE_ENDPOINT
- [x] GET_BLOCK_ENDPOINT
- [x] GET_BLOCKS_ENDPOINT
- [x] GET_CONTRACT_ENDPOINT
- [x] GET_CONTRACT_RESULTS_BY_ADDRESS_ENDPOINT
- [x] GET_CONTRACT_RESULTS_DETAILS_BY_CONTRACT_ID_ENDPOINT
- [x] GET_CONTRACT_RESULT_ENDPOINT
- [x] GET_CONTRACT_RESULT_LOGS_ENDPOINT
- [x] GET_CONTRACT_RESULT_LOGS_BY_ADDRESS_ENDPOINT
- [x] GET_CONTRACT_RESULTS_ENDPOINT
- [x] GET_NETWORK_EXCHANGERATE_ENDPOINT
- [x] GET_NETWORK_FEES_ENDPOINT
- [x] GET_TOKENS_ENDPOINT
- [x] GET_TRANSACTIONS_ENDPOINT
- [x] CONTRACT_CALL_ENDPOINT
- [x] GET_STATE_ENDPOINT

### Alternatives

_No response_

AlfredoG87:
**Description**:
Refactored code around MirrorNodeClient calls to GET_ACCOUNTS_BY_ID_ENDPOINT so it does not return null, but instead throws the corresponding error and the methods that consume that endpoint handle the errors as they see fit.
**Related issue(s)**:
#1276 
Fixes 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -496,7 +496,7 @@ export class EthImpl implements Eth {
           const accountCacheKey = `${constants.CACHE_KEY.ACCOUNT}_${transaction.to}`;
           let toAccount: object | null = this.cache.get(accountCacheKey);
           if (!toAccount) {
-            toAccount = await this.mirrorNodeClient.getAccount(transaction.to, requestId);
+            toAccount = await this.getAccountOrNull(transaction.to, requestId);
           }
 
           // when account exists return default base gas, otherwise return the minimum amount of gas to create an account entity
@@ -520,6 +520,31 @@ export class EthImpl implements Eth {
 
     return gas;
   }
+
+  /**
+   * Gets the account data using the mirror node and handles the potential 404 error and returns null instead.
+   * */
+    async getAccountOrNull(address: string, requestId?: string) {
+      let account;
+      try {
+        account = await this.mirrorNodeClient.getAccount(address, requestId);
+      } catch (error: any) {
+        if(error instanceof MirrorNodeClientError) {
+          if(error.statusCode == 404){
+            return null;
+
+          } else if(error.statusCode == 400){
+            this.logger.debug(`${formatRequestIdMessage(requestId)} Got Invalid Parameter when trying to fetch account from mirror node: ${JSON.stringify(error)}`);
+            throw predefined.INVALID_PARAMETER(address, `Invalid 'address' field in transaction param. Address must be a valid 20 bytes hex string`);
+          }
+        } else {
+          this.logger.error(`${formatRequestIdMessage(requestId)} Unexpected error raised while fetching account from mirror-node: ${JSON.stringify(error)}`);
+          throw error;
+        }
+      }
+
+      return account;
+    }
 
   /**
    * Gets the current gas price of the network.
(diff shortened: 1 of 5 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -493,7 +493,15 @@ export class EthImpl implements Eth {
           const accountCacheKey = `${constants.CACHE_KEY.ACCOUNT}_${transaction.to}`;
           let toAccount: object | null = this.cache.get(accountCacheKey);
           if (!toAccount) {
-            toAccount = await this.mirrorNodeClient.getAccount(transaction.to, requestId);
+            try {
```
Nana-EC, line 496: Extract to a shared method doing the logic and returning the account, null or throwing an error

packages/relay/src/lib/eth.ts
```diff
@@ -493,7 +493,15 @@ export class EthImpl implements Eth {
...
+              toAccount = await this.mirrorNodeClient.getAccount(transaction.to, requestId);
+            } catch (error: any) {
+              if(error instanceof MirrorNodeClientError && error.statusCode == 404) {
+                toAccount = null;
+              } else {
+                throw error;
```
Nana-EC, line 502: Q: Should we log and throw `predefined.RESOURCE_NOT_FOUND` instead like in precheck?

Nana-EC (changes requested):

packages/relay/src/lib/eth.ts
AlfredoG87: Not sure, since when is not found (404), we will return null instead.
So the only other option is an unexpected error or a (400 - Invalid Parameter [see docs](https://testnet.mirrornode.hedera.com/api/v1/docs/#/accounts/getAccountByIdOrAliasOrEvmAddress) )

I think that for Invalid Parameter we could also return Null or throw a related error, even thought right now we are returning null, I think we should throw an INVALID_PARAMETER error.

But for other errors, since we don't know what "unexpected error" would be, I think we should throw instead the same error we receive.

That said, I will log the error in both cases nevertheless.

Nana-EC (dismissed):

Nana-EC (approved):

Nana-EC:
@AlfredoG87 should rebase this so we can get it in

AlfredoG87:
closed in favor of this one: https://github.com/hashgraph/hedera-json-rpc-relay/pull/1649

PR #1367, opened 2023-06-16
Removed a call to the mirror node.

### LINKED ISSUE #1330: Update Mirror Node getBalances calls to utilize evm address
### Problem

Initially the `api/v1/balances` endpoint required the accountId to retrieve details, which required an additional query to go from evmAddress -> accountId

### Solution

With Mirror Node 0.81 implementation of https://github.com/hashgraph/hedera-mirror-node/pull/5979 the relay can remove the extra step to go from evmAddress -> accountId

### Alternatives

_No response_

ebadiere:
**Description**:
Flipped logic to check if queried block timeDiff is greater than interval.  This allows calling the mirror node for the account info only when necessary.  The evm address can now be used to get balance at timestamp from the mirror node.

**Related issue(s)**:

Fixes #1330


AlfredoG87 (dismissed):
LGTM

packages/relay/src/lib/eth.ts
```diff
@@ -757,8 +757,8 @@ export class EthImpl implements Eth {
 
     let blockNumber = null;
     let balanceFound = false;
-    let weibars: BigInt = BigInt(0);
-    const mirrorAccount = await this.mirrorNodeClient.getAccountPageLimit(account, requestId);
```
Nana-EC, line 761: Aren't there any mocking changes that need to be done with this no longer being called?

Nana-EC (dismissed):

packages/relay/src/lib/eth.ts
AlfredoG87: I think is called but only on certain cases. when the block is within the last 15 minutes.

AlfredoG87 (approved):

PR #1376, opened 2023-06-20
Ignore test files in sonar cloud checks

Ivo-Yankov:
**Description**:

Something seems to have changed in the default SonarCloud configuration and every PR has a lot of code smells and code duplication warnings that cause the SonarCloud check to fail. All of the warnings are related to very old code and I think that the test files used to be ignored by default. This PR brings back the previous default behaviour.

- Adds a config file for SonarCloud
- Ignores all test files from SonarCloud checks

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,0 +1,10 @@
+sonar.sources=.
+sonar.exclusions=**/*spec.ts
+sonar.exclusions=**/*spec.js
+
+sonar.tests=.
+sonar.test.inclusions=**/*spec.ts
+sonar.test.inclusions=**/*spec.js
+
+sonar.cpd.exclusions=**/*spec.ts
+sonar.cpd.exclusions=**/*spec.js
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,10 +1,4 @@
 sonar.sources=.
 sonar.exclusions=**/*spec.ts
-sonar.exclusions=**/*spec.js
-
-sonar.tests=.
-sonar.test.inclusions=**/*spec.ts
-sonar.test.inclusions=**/*spec.js
 
 sonar.cpd.exclusions=**/*spec.ts
-sonar.cpd.exclusions=**/*spec.js
--- a/packages/relay/tests/lib/eth.spec.ts
+++ b/packages/relay/tests/lib/eth.spec.ts
@@ -92,8 +92,6 @@ let clientServiceStub;
 let cache;
 let mirrorNodeCache;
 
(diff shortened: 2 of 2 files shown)


AlfredoG87 (approved):
LGTM

Nana-EC (changes requested):
Q: Why ignore them vs fix them? 
Granted it's annoying at first but maybe we should just address the sonar complaints and maintain the improvements as they increase readability and have caught some code smells that should be in the code

Ivo-Yankov:
> Q: Why ignore them vs fix them? Granted it's annoying at first but maybe we should just address the sonar complaints and maintain the improvements as they increase readability and have caught some code smells that should be in the code

I agree that we should fix them, but maybe it's better to do that at a planned time and in a dedicated PR - this way the current PRs won't be blocked and we won't encounter a lot of conflicts from everyone trying to fix the same code smells on their own. 

Also a lot of the test suites require major refactoring at this point - for example `eth.spec.ts` has grown to almost 5000 lines of code.

Nana-EC:
> > Q: Why ignore them vs fix them? Granted it's annoying at first but maybe we should just address the sonar complaints and maintain the improvements as they increase readability and have caught some code smells that should be in the code
> 
> I agree that we should fix them, but maybe it's better to do that at a planned time and in a dedicated PR - this way the current PRs won't be blocked and we won't encounter a lot of conflicts from everyone trying to fix the same code smells on their own.
> 
> Also a lot of the test suites require major refactoring at this point - for example `eth.spec.ts` has grown to almost 5000 lines of code.

I don't see any PRs being blocked by sonar right now. My concern is removing the checks will push this into tech debt territory that never gets addressed.
I think a few edits here and there as we see the sonar warnings will push up to get it done over time

Ivo-Yankov:
> > > Q: Why ignore them vs fix them? Granted it's annoying at first but maybe we should just address the sonar complaints and maintain the improvements as they increase readability and have caught some code smells that should be in the code
> > 
> > 
> > I agree that we should fix them, but maybe it's better to do that at a planned time and in a dedicated PR - this way the current PRs won't be blocked and we won't encounter a lot of conflicts from everyone trying to fix the same code smells on their own.
> > Also a lot of the test suites require major refactoring at this point - for example `eth.spec.ts` has grown to almost 5000 lines of code.
> 
> I don't see any PRs being blocked by sonar right now. My concern is removing the checks will push this into tech debt territory that never gets addressed. I think a few edits here and there as we see the sonar warnings will push up to get it done over time

Fair point, I'm closing the PR

PR #1380, opened 2023-06-20
Adding bigger buckets to JSON RPC method statusCode latency

### LINKED ISSUE #1378: Adding buckets to response of relay metrics
### Problem

The biggest bucket for the metric `rpc_relay_method_response` is 10,000 ms, so only 10 seconds... even if is useful, I think we should include a couple more buckets in order to be able to see how many requests are above 20s, 30s, 40s, 50s and 60s.

Considering that we have found several timeouts after 30s in `testnet` and we are not able to know how many, when, and what methods are being affected in clear numbers.

### Solution

Add new buckets to metric `rpc_relay_method_response` for 20s, 30s, 40s, 50s and 60s.

### Alternatives

_No response_

AlfredoG87:
**Description**:
Adding bigger buckets to JSON RPC method statusCode latency so we are aware of timeouts surpassing 30 seconds and above.

**Related issue(s)**:

Fixes #1378 

**Notes for reviewer**:
image

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -231,7 +231,7 @@ export class MirrorNodeClient {
             help: 'Mirror node response method statusCode latency histogram',
             labelNames: ['method', 'statusCode'],
             registers: [register],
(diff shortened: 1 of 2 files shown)


ebadiere (approved):
LG.

PR #1381, opened 2023-06-21
Simplify this down to one container and the built-in wget

rustyShacklefurd:
**Description**:
This PR modifies the newman cronjob test
* Removes init container, volume and volume mounts
* Utilized a single container to run this test

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
Running in integration as `cronjob.integration-json-rpc-relay-test`.  The integration cronjob was edited to test the updated jobspec and then a manual job was created using ` k create job --from=cronjob/integration-json-rpc-relay-test -n integration $NAME`

**Checklist**

### DIFF
--- a/charts/hedera-json-rpc-relay/templates/cronjob.yaml
+++ b/charts/hedera-json-rpc-relay/templates/cronjob.yaml
@@ -59,17 +59,8 @@ spec:
               image: "{{ .Values.cronjob.image.repository }}:{{ .Values.cronjob.image.tag }}"
               imagePullPolicy: {{ .Values.cronjob.image.pullPolicy }}
               command: 
-                - newman
-                - run
-                - postman.json
-                - --env-var 
-                - baseUrl=http://{{ include "json-rpc-relay.fullname" . }}:{{ .Values.service.port }}
-              volumeMounts:
-              - name: config-json
-                mountPath: "/"
-                readOnly: true
-          volumes:
-            - name: config-json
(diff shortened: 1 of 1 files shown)


charts/hedera-json-rpc-relay/templates/cronjob.yaml
```diff
@@ -59,17 +59,8 @@ spec:
...
-              volumeMounts:
-              - name: config-json
-                mountPath: "/"
-                readOnly: true
-          volumes:
-            - name: config-json
-              emptyDir: {}                          
+                - /bin/sh
+                - -c
+                - wget https://raw.githubusercontent.com/hashgraph/hedera-json-rpc-relay/main/charts/hedera-json-rpc-relay/postman.json ; newman run postman.json --env-var baseUrl=http://{{ include "json-rpc-relay.fullname" . }}:{{ .Values.service.port }}
```
Nana-EC, line 64: Q: Is the result of dropping the volumes why we have to do `wget` for the `postman.json` file even though it's already in the chart dir?

rustyShacklefurd: The postman.json is not in this container anywhere so it has to be pulled from the source.  This would be preferred over having postman.json existing in the container at boot because it allows for the postman test to be changed/updated without a need to update the job or image to get those updates.

Nana-EC (approved):

PR #1384, opened 2023-06-22
Batch parallel contract result requests for detailed block info

### LINKED ISSUE #1383: Parallelize block contracts/{address}/results/{timestamp} calls
### Problem

Currently the relay makes successive `contracts/{address}/results/{timestamp}` calls to the Mirror Node when returning detailed block information.
This can take some time and is inefficients

### Solution

Parallelize block contracts/{address}/results/{timestamp} calls

### Alternatives

_No response_

Nana-EC:
**Description**:
Batch parallel contract results calls for getBlockBy calls
- Add `ETH_GET_BLOCK_BY_RESULTS_BATCH_SIZE` env to set batch size
- Update `ETH_GET_TRANSACTION_COUNT_MAX_BLOCK_RANGE` to an increased size as parallel calls increase support
- Replace sequential contract results logic when `showDetails` is true with parallel logic
- Added a new spec test file that can eventually house the eth_getBlockBy tests to reduce the size of `eth.spec.ts`

**Related issue(s)**:

Fixes #1383 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -1647,22 +1648,23 @@ export class EthImpl implements Eth {
...
+    }
 
-          const transaction = await this.getTransactionFromContractResult(result.to, result.timestamp, requestId);
-          if (transaction !== null) {
-            transactionObjects.push(transaction);
+    for (let i = 0; i < contractResults.length; i+= this.ethGetBlockByResultsBatchSize) {
+      if (showDetails) {
+        Promise.all(contractResults.slice(i, i + this.ethGetBlockByResultsBatchSize).map(async result => {
+          // depending on stage of contract execution revert the result.to value may be null
+          if (!_.isNil(result.to)) {
```
AlfredoG87, line 1659:
```suggestion
          if (result.to != null) {
```

This should be equivalent, does not uses `_` and is easier to understand.

packages/relay/src/lib/eth.ts
```diff
@@ -1647,22 +1648,23 @@ export class EthImpl implements Eth {
...
+            const transaction = await this.getTransactionFromContractResult(result.to, result.timestamp, requestId);
```
AlfredoG87, line 1660: do we have a tests were we validate behaviour if we get exception from this method while inside the promise?
`504` or `404` exceptions could come, but any other unexpected exception too.

Nana-EC: Incoming

packages/relay/src/lib/eth.ts
Nana-EC: Existing code but will update

packages/relay/src/lib/eth.ts
Nana-EC: Added a timeout test

AlfredoG87 (approved):
LGTM

SonarCloud is detecting some duplicated lines on a test, I actually think that some duplicated lines in tests is better, so tests are more independent from one another and also is more readable.

Issue #1385, opened 2023-06-22
Explore optionality of the id field in RPC requests


Nana-EC:
### Problem

The JSON RPC 2.0 mandates requests have a body with json in teh form
```json
{"jsonrpc":"2.0","id":"2","method":"methodName","params": null|array }
```

However, some well known wallets don't send the `id` in the body

### Solution

Investigate and explore support for optionality of the `id` field.

Nana-EC:
Making id field optional would break the adherence to the spec.
PR 1386 is the minimum needed to support testing by devs using tools that don't provide this id.
No further work will be done on this.

PR #1386, opened 2023-06-22
Add optional flag for request id

Nana-EC:
**Description**:
Add optional flag for request id
- Update config docs
- Add `REQUEST_ID_IS_OPTIONAL` env config consideration
- Add logic to allow missing field if config is set and to set id as 0 to avoid failing response checks

**Related issue(s)**:

Touches on #1385 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/src/koaJsonRpc/index.ts
+++ b/packages/server/src/koaJsonRpc/index.ts
@@ -64,7 +64,7 @@ export default class KoaJsonRpc {
   private koaApp: Koa<Koa.DefaultState, Koa.DefaultContext>;
   private requestId: string;
   private logger: Logger;
-  private startTimestamp!: number;
+  private readonly requestIdIsOptional = process.env.REQUEST_ID_IS_OPTIONAL == 'true';
 
   constructor(logger: Logger, register: Registry, opts?) {
     this.koaApp = new Koa();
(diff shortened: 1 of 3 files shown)


packages/server/src/koaJsonRpc/index.ts
```diff
@@ -122,10 +122,17 @@ export default class KoaJsonRpc {
...
+        {          
+          // If the request is invalid, we still want to return a valid JSON-RPC response, default id to 0
+          body.id = '0';
+        } else {
+          ctx.body = jsonResp(body.id || null, new InvalidRequest(), undefined);
+          ctx.status = 400;
+          ctx.state.status = `${ctx.status} (${INVALID_REQUEST})`;
+          return;
+        }
       }
```
AlfredoG87, line 136: I would recommend extracting validation for `id` into another method for simplicity, and separate it from `POST` and other fields validation for clarity, just cause otherwise the lo added is out of context s and double evaluation of !hasOwnProperty(body, 'id') might not be needed.

Nana-EC: Agreed. Pulled it out into `hasInvalidReqestId()`

ebadiere (commented):
Agree with @AlfredoG87 suggestion.  Otherwise, LG.

AlfredoG87 (approved):

PR #1413, opened 2023-06-29
Update default env and some misc fixes

### LINKED ISSUE #1246: Add default values to .env.example
### Problem

Currently, it's difficult to know which values should be set in the .env file because the example does not provide default values. This sometimes results in misconfiguration, causing the JSON RPC relay to not work properly.

### Solution

To make running JSON-RPC easier, we should set default values in the .env file for developers so that they can easily get the server up and running.

### Alternatives

_No response_

Nana-EC:
**Description**:
A couple of useful misc fixes sitting on my local
- Update my `.env.example` to simplify contents 
- Update README.md helm install section to fix location of umbrella chart ref
- Update `postman.json` files to query `eth_getTransctionCount`
- Update `sdkClient.ts`, `mirrorNodeClient.ts` and `eth.ts` use of `cache.set` to provide `requestId` for better logging
- Update `server.spec.ts` to add cors header content checks

**Related issue(s)**:

Fixes #1246 

**Notes for reviewer**:

**Checklist**


.env.example
AlfredoG87, line 1: I like that on the example we don't put all available configs, that is for the docs.
and leave here only the ones that are **required**

AlfredoG87 (approved):
LGTM

PR #1244, opened 2023-05-13
Improve logging across classes

Nana-EC:
**Description**:
Improve logging for easier troubleshooting

- Exchange `requestId` for `requestIdPrefix` to reduce number of calls to `formatRequestIdMessage()`
- Set logger name for poller and subImpl loggers
- add request id log to ws-server, distinguishing connection id and request id

**Related issue(s)**:

Fixes #1218 

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #1424, opened 2023-06-30
1398 execution reverted should be returned to provider

### LINKED ISSUE #1398: `execution reverted` should be returned to the provider as a ProviderError by the JSON RPC Relay
### Description

A contract with the require statement below will revert the contract transaction if the condition is not met and return the message:
```
        require(
            (numRecipients > 1) &&
                (numRecipients <= MAX_COUNT),
            "Invalid number of recipients"
        );
```
The following test passes on the hardhat node, foundry anvil node, alchemy sepolia endpoint, and infura sepolia endpoint:
```
  it("should trigger the 'Invalid number of recipients'", async function () {
    const balanceBefore = await erc20.balanceOf(signers[1].address);
    console.log(`Balance before: ${balanceBefore}`);
    await erc20.approve(multiSendAddress, 66);
    console.log(`Approved`);
    const recipients = [signers[1].address];

    await expect(multiSend.multiSendFungibleTokenPush(erc20, 33, recipients)).to.be.revertedWith("Invalid number of recipients");  
  });  
```
 When run against the relay it fails with the following:
```
  1) MultiSend
       should trigger the 'Invalid number of recipients':
     AssertionError: Expected transaction to be reverted with reason 'Invalid number of recipients', but it didn't revert
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Context.<anonymous> (test/MultiSend.ts:133:5)
```

### Steps to reproduce

Use the snippets above.

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.25.0

### Operating system

macOS

ebadiere:
Returns the JsonRpcError in the eth_estimateGas.

**Related issue(s)**:

Fixes #1398


packages/relay/tests/lib/eth.spec.ts
```diff
@@ -3033,6 +3033,21 @@ describe('Eth calls using MirrorNode', async function () {
     expect(gas).to.equal(EthImpl.numberTo0x(defaultGasOverride));
   });
 
+  it('eth_estimateGas with contract revert returns the JsonRpcError', async function () {
+    const transaction = {
+      from: "0x05fba803be258049a27b820088bab1cad2058871",
+      data: "0x60806040523480156200001157600080fd5b50604051620019f4380380620019f48339818101604052810190620000379190620001fa565b818181600390816200004a9190620004ca565b5080600490816200005c9190620004ca565b5050505050620005b1565b6000604051905090565b600080fd5b600080fd5b600080fd5b600080fd5b6000601f19601f8301169050919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052604160045260246000fd5b620000d08262000085565b810181811067ffffffffffffffff82111715620000f257620000f162000096565b5b80604052505"
+    };
+    const id = uuid();
+    web3Mock.onPost('contracts/call', {...transaction, estimate: true}).reply(400, predefined.CONTRACT_REVERT);
```
Nana-EC, line 3042: make this a mirror node error with a custom message that you can check for later

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -3033,6 +3033,21 @@ describe('Eth calls using MirrorNode', async function () {
...
+    const contractRevert= new JsonRpcError( {name: 'Contract revert executed', code: -32008, message: 'execution reverted: '});
```
Nana-EC, line 3043:
```suggestion
    const contractRevert= predefined.CONTRACT_REVERT(<customMirrorerrorMessage>);
```

packages/relay/src/lib/eth.ts
```diff
@@ -509,6 +509,10 @@ export class EthImpl implements Eth {
         // Handle Contract Call or Contract Create
         gas = this.defaultGas;
       }
+
+      if (e instanceof JsonRpcError) {
```
Nana-EC, line 513: Move to first line and adjust to something like this. I think it should cover both scenarios where we shouldn't fallback. Basically throw if we get a valid revert message from Mirror Node
```suggestion
      if (e instanceof JsonRpcError && e.mesage !== '') {
```

Nana-EC (changes requested):
Provided some logic and test suggestions

packages/relay/tests/lib/eth.spec.ts
ebadiere: Done.

packages/relay/src/lib/eth.ts
ebadiere: Done.

packages/relay/tests/lib/eth.spec.ts
ebadiere: Done, however, the 400 error response is used to instantiate a MirrorNodeClientError that discards the JsonRpcError.data field, so the data field in the error cannot be verified.

packages/relay/src/lib/eth.ts
```diff
@@ -483,6 +484,11 @@ export class EthImpl implements Eth {
       }
     } catch (e: any) {
       this.logger.error(`${requestIdPrefix} Error raised while fetching estimateGas from mirror-node: ${JSON.stringify(e)}`);
+      
+      // execute reverted: with no message comes from a contract deployment
+      if ((e.statusCode != 501) && (e instanceof JsonRpcError) && (e.message === EthImpl.executionReverted)) {
```
Nana-EC, line 489: I don't think this will work, the error here if it's a contract revert is not an HTTPError. 
`e.statusCode != 501` will be true for other cases and even when the mirror node returns 400 it won't make it here as the same error. 
`JSONRpcError` doesn't even have a `statusCode` field

Look at [MirrorNodeClient.handleError()](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/mirrorNodeClient.ts#L335), it throws a `predefined.CONTRACT_REVERT(mirrorError.errorMessage)`.
Which means here the error is a new error type. I think the suggestion i provided previously is closer to what you need.

```suggestion
      if ((e.statusCode == MirrorNodeClientError.ErrorCodes) && (e instanceof JsonRpcError) && (e.message === EthImpl.executionReverted)) {
```

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -3033,6 +3033,43 @@ describe('Eth calls using MirrorNode', async function () {
     expect(gas).to.equal(EthImpl.numberTo0x(defaultGasOverride));
   });
 
+  it('eth_estimateGas with contract revert returns the JsonRpcError', async function () {
+    const transaction = {
+      from: "0x05fba803be258049a27b820088bab1cad2058871",
+      data: "0x60806040523480156200001157600080fd5b50604051620019f4380380620019f48339818101604052810190620000379190620001fa565b818181600390816200004a9190620004ca565b5080600490816200005c9190620004ca565b5050505050620005b1565b6000604051905090565b600080fd5b600080fd5b600080fd5b600080fd5b6000601f19601f8301169050919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052604160045260246000fd5b620000d08262000085565b810181811067ffffffffffffffff82111715620000f257620000f162000096565b5b80604052505"
+    };
+    const id = uuid();
+    web3Mock.onPost('contracts/call', {...transaction, estimate: true}).reply(400, predefined.CONTRACT_REVERT('solidity require failed!'));
```
Nana-EC, line 3042: The mirror node would never return a JSON RPC Error in this format 
```suggestion
    web3Mock.onPost('contracts/call', {...transaction, estimate: true}).reply(400, {
      "_status": {
        "messages": [
          {
            "data": "0x3000",
            "detail": "Generic detailed error message",
            "message": "solidity require failed!"
          }
        ]
      }
    });
```

Nana-EC (commented):
Logic is still a bit off hence the CI failures. Left comments

packages/relay/tests/lib/eth.spec.ts
Nana-EC: Correct. 
Except the MirrorNodeClientError then causes a `predefined.CONTRACT_REVERT` to be thrown in `MirrorNodeCLient.ts` `handleError()`

ebadiere: Done.

packages/relay/tests/lib/eth.spec.ts
ebadiere: Done.

packages/relay/src/lib/eth.ts
```diff
@@ -506,6 +507,10 @@ export class EthImpl implements Eth {
           return predefined.INVALID_PARAMETER(0, `Invalid 'value' field in transaction param. Value must be greater than 0`);
         }
       } else {
+        // execute reverted: with no message comes from a contract deployment
+        if ((e.statusCode != 501) && (e instanceof JsonRpcError) && (e.message === EthImpl.executionReverted)) {
```
Nana-EC, line 511: Similar to previous comments I think it'd be best to return valid revert messages for now or else fallback to what we did before

```suggestion
        if (e instanceof JsonRpcError && e.message !== EthImpl.executionReverted)) {
```

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -3033,6 +3033,55 @@ describe('Eth calls using MirrorNode', async function () {
...
+          }
+        ]
+      }
+    });
+
+    const contractRevert = predefined.CONTRACT_REVERT("solidity require failed!");
+    const result: any = await ethImpl.estimateGas(transaction, id);
+
+    expect(result.code).to.equal(contractRevert.code);
+    expect(result.message).to.equal(contractRevert.message);
```
Nana-EC, line 3058: This will never be true as you've mocked the mirror node response with a message that differs from the expected `contractRevert` error message.
If you expect them to be equal then the message should be the same

packages/relay/src/lib/eth.ts
Nana-EC: None of our `JsonRpcError` objects have a `statusCode` field so `(e.statusCode != 501) && (e instanceof JsonRpcError)` shouldn't be true in the case of `predefined.CONTRACT_REVERT

packages/relay/tests/lib/eth.spec.ts
Nana-EC: Ah i see it now. The error message is in the data

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -3033,6 +3033,55 @@ describe('Eth calls using MirrorNode', async function () {
...
+    const transaction = {
+      from: "0x05fba803be258049a27b820088bab1cad2058871",
+      data: "0x60806040523480156200001157600080fd5b50604051620019f4380380620019f48339818101604052810190620000379190620001fa565b818181600390816200004a9190620004ca565b5080600490816200005c9190620004ca565b5050505050620005b1565b6000604051905090565b600080fd5b600080fd5b600080fd5b600080fd5b6000601f19601f8301169050919050565b7f4e487b7100000000000000000000000000000000000000000000000000000000600052604160045260246000fd5b620000d08262000085565b810181811067ffffffffffffffff82111715620000f257620000f162000096565b5b80604052505"
+    };
+    const id = uuid();
+    web3Mock.onPost('contracts/call', {...transaction, estimate: true}).reply(501, {
+      "_status": {
+        "messages": [
+          {
+            "data": "0x3000",
```
Nana-EC, line 3073: nit: AN unimplemented case wouldn't return a valid gas amount

packages/relay/src/lib/eth.ts
Nana-EC: It's the opposite `e.statusCode != 501)` will alays be true since in the `predefined.CONTRACT_REVERT` case  `e.statusCode` will be `undefined`

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -3033,6 +3033,55 @@ describe('Eth calls using MirrorNode', async function () {
     expect(gas).to.equal(EthImpl.numberTo0x(defaultGasOverride));
   });
 
+  it('eth_estimateGas with contract revert returns the JsonRpcError', async function () {
```
Nana-EC, line 3036: You'd need 1 more test to confirm your fix. One where a 400 returns the `'execution reverted: '` message and you confirm that it doesn't throw but instead returns a static value

Nana-EC (changes requested):
Missing test for given logic

packages/relay/tests/lib/eth.spec.ts
ebadiere: The MirrorNodeClientError constructor parses the incoming message into a message and data and the errorMessage is set using the data.  The errorMessage is returned and later used to set the predefined.CONTRACT_REVERT.

Nana-EC: That flow is actually problematic because the mirror node error message is actually in the message part of the response. I've opened #1429 which should be addressed first before your PR. I have a feeling that is contributing to the empty errors

packages/relay/src/lib/eth.ts
ebadiere: Agreed, but 501 was discussed.  Updated to return defaultGas on a 501.

packages/relay/tests/lib/eth.spec.ts
ebadiere: Removed.

packages/relay/tests/lib/eth.spec.ts
ebadiere: Done.

packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -251,10 +251,15 @@ describe('@erc20 Acceptance Tests', async function () {
                                     });
 
                                     it('reverts', async function () {
+                                     try {
                                         await Assertions.expectRevert(
                                             contract.connect(spenderWallet).transferFrom(tokenOwner, to, amount),
                                             Constants.CALL_EXCEPTION
                                         );
+                                     }catch(e){
+                                        expect(e.code).to.be.equal('UNPREDICTABLE_GAS_LIMIT');
```
Nana-EC, line 260: Q: why are we catching `UNPREDICTABLE_GAS_LIMIT` in multiple places?
That doesn't seen to be the aim of the tests. We can always add another test cases for this specific case

packages/relay/src/lib/eth.ts
```diff
@@ -517,6 +519,10 @@ export class EthImpl implements Eth {
           return predefined.INVALID_PARAMETER(0, `Invalid 'value' field in transaction param. Value must be greater than 0`);
         }
       } else {
+        // execute reverted: with no message comes from a contract deployment
+        if (e.stack.substring(0, 31) === EthImpl.errorContractReverted) {
```
Nana-EC, line 523: Q: @ebadiere do you still see errors with no messages once you rebased?
I'm wondering if the check for no message is still necessary.

packages/relay/src/lib/eth.ts
```diff
@@ -86,6 +87,7 @@ export class EthImpl implements Eth {
   static ethFeeHistory = 'eth_feeHistory';
   static ethGetTransactionCount = 'eth_getTransactionCount';
   static ethSendRawTransaction = 'eth_sendRawTransaction';
+  static executionReverted = 'execution reverted: ';
```
Nana-EC, line 90: If you need this add it in the JSONRPCError class and call from here so the string is centralized?

packages/relay/src/lib/eth.ts
```diff
@@ -1262,15 +1268,16 @@ export class EthImpl implements Eth {
         }
 
         if (e.isContractReverted()) {
-          this.logger.trace(`${requestIdPrefix} mirror node eth_call request encoutered contract revert. message: ${e.message}, details: ${e.detail}, data: ${e.data}`);
```
Nana-EC, line 1265: Q: Why remove the e.message log?

packages/relay/src/lib/eth.ts
```diff
@@ -1262,15 +1268,16 @@ export class EthImpl implements Eth {
...
-          return predefined.CONTRACT_REVERT(e.detail || e.message, e.data);
+          this.logger.trace(`${requestIdPrefix} mirror node eth_call request encoutered contract revert. details: ${e.detail}, data: ${e.data}`);
+          return predefined.CONTRACT_REVERT(e.message + e.detail, e.data);
```
Nana-EC, line 1272: Q: why this change?
Is this in the case both message and detail are populated?

Nana-EC (commented):
A few questions.
Also why is the package-lock.json updated here?

packages/relay/src/lib/eth.ts
ebadiere: Merged in the latest mirror node error handling fix and updated the condition to:
```
        if (e.message === EthImpl.errorContractReverted) {
          return predefined.CONTRACT_REVERT(e.detail, e.data);
        }      
```

packages/relay/src/lib/eth.ts
ebadiere: Updated condition with mirror node fix.

packages/server/tests/acceptance/erc20.spec.ts
ebadiere: The UNPREDICTABLE_GAS_LIMIT is an error that ethers.js wraps around the error returned from the relayClient.  The JsonRpc error returned by the relay is handled properly in a standard hardhat project.  The test was updated to extract the JSON RPC Relay error and validate it.

packages/relay/src/lib/eth.ts
ebadiere: Removed

packages/relay/src/lib/eth.ts
ebadiere: You restored it.

packages/relay/src/lib/eth.ts
ebadiere: No, and this has been updated.

Nana-EC (commented):
Nice improvements. Minor suggestions left.
Will circle back for the test extraction review

packages/relay/src/lib/eth.ts
```diff
@@ -517,6 +518,10 @@ export class EthImpl implements Eth {
           return predefined.INVALID_PARAMETER(0, `Invalid 'value' field in transaction param. Value must be greater than 0`);
         }
       } else {
+        // execute reverted: with no message comes from a contract deployment
+        if (e.message === EthImpl.errorContractReverted) {
```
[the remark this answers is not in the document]
Nana-EC: > A 400 response can also include the following:

That's fine. If there were invalid params shouldn't we still return an error with those details? It's not a contract revert but it is a 400 error from the Mirror Node.
I believe we want to return the error with details for whatever reason that the Mirror Node returns except unsupported, which is a 501 case.
Is there a scenario still in the 400 case where we should not return the error?

In the current logic will all contract reverts be caught with this message check?
What if the mirror node changes the message text, this logic would break right?

> The e.isContractReverted() is used in the eth_call, via callMirrorNode, so that would have to be tested. eth_estimateGas simply calls the mirrorNode using the mirroNodeClient.

callMirrorNode calls `this.mirrorNodeClient.postContractCall`, which is the same call the estiamteGas is making via the  mirroNodeClient. It's the same mirror node endpoint in both cases

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "root",
-    "lockfileVersion": 2,
+    "lockfileVersion": 3,
```
Nana-EC, line 3: Q: what's the cause of this?

packages/relay/src/lib/eth.ts
```diff
@@ -524,6 +524,12 @@ export class EthImpl implements Eth {
           return predefined.INVALID_PARAMETER(0, `Invalid 'value' field in transaction param. Value must be greater than 0`);
         }
       } else {
+        // The size limit of the encoded contract posted to the mirror node can cause contract deployment transactions to fail with a 400 response code.
+        // The contract is actually deployed on the consensus node, so the contract will work.  In these cases, we don't want to return a 
+        // CONTRTACT_REVERT error.
+        if (e.isContractReverted() && e.detail && e.data && e.message !== MirrorNodeClientError.messages.INVALID_HEX && process.env.ALLOW_INVALID_HEX_RESPONSE) {
```
Nana-EC, line 530: Q: why do we need to check for `e.detail && e.data`?

packages/relay/src/lib/eth.ts
```diff
@@ -1269,8 +1275,8 @@ export class EthImpl implements Eth {
         }
 
         if (e.isContractReverted()) {
-          this.logger.trace(`${requestIdPrefix} mirror node eth_call request encoutered contract revert. message: ${e.message}, details: ${e.detail}, data: ${e.data}`);
-          return predefined.CONTRACT_REVERT(e.detail || e.message, e.data);
+          this.logger.trace(`${requestIdPrefix} mirror node eth_call request encoutered contract revert. details: ${e.message}, ${e.detail}, data: ${e.data}`);
```
Nana-EC, line 1278:
```suggestion
          this.logger.trace(`${requestIdPrefix} mirror node eth_call request encountered contract revert. message: ${e.message}, details: ${e.detail}, data: ${e.data}`);
```

Nana-EC (approved):
Unblocking, left some suggestions.

package-lock.json
ebadiere: I was using npm version 9 which sets this to 3, to allow backwards compatibility.  I downgraded to npm v8.  The readme only lists a version number for node.  Perhaps we want to fill out the rest of the tool versioning?  Maybe in the configuration.md?

packages/relay/src/lib/eth.ts
ebadiere: I wanted to lock in all of the conditions that would describe the invalid hex error.  It is the exceptional case now.  So far we've treated 400 responses as CONTRACT_REVERT.

packages/relay/src/lib/eth.ts
ebadiere: Done.

packages/relay/src/lib/eth.ts
AlfredoG87: Should not be something like this preferable:
```suggestion
          return predefined.CONTRACT_REVERT(`message: ${e.message}, detail: ${e.detail}`, e.data);
```
or
```suggestion
          return predefined.CONTRACT_REVERT(`${e.message} : ${e.detail}`, e.data);
```

packages/relay/src/lib/eth.ts
```diff
@@ -1269,8 +1275,8 @@ export class EthImpl implements Eth {
         }
 
         if (e.isContractReverted()) {
-          this.logger.trace(`${requestIdPrefix} mirror node eth_call request encoutered contract revert. message: ${e.message}, details: ${e.detail}, data: ${e.data}`);
-          return predefined.CONTRACT_REVERT(e.detail || e.message, e.data);
+          this.logger.trace(`${requestIdPrefix} mirror node eth_call request encountered contract revert. message: ${e.message}, details: ${e.detail}, data: ${e.data}`);
+          return predefined.CONTRACT_REVERT(e.message + e.detail, e.data);
```
AlfredoG87, line 1279: Should not be something like this preferable:
```suggestion
          return predefined.CONTRACT_REVERT(`message: ${e.message}, detail: ${e.detail}`, e.data);
```
or
```suggestion
          return predefined.CONTRACT_REVERT(`${e.message} : ${e.detail}`, e.data);
```

packages/relay/tests/helpers.ts
AlfredoG87, line 1: we dont have a test with scenario
```
{
    "_status": {
        "messages": [
            {
                "message": "data field invalid hexadecimal string",
                "detail": "",
                "data": ""
            },
            {
                "message": "to field must not be empty",
                "detail": "",
                "data": ""
            }
        ]
    }
}
```

packages/relay/src/lib/eth.ts
AlfredoG87: I also not sure why we need to check for `e.detail && e.data`

ebadiere: Removed.

packages/relay/src/lib/eth.ts
ebadiere: Updated.

packages/relay/tests/helpers.ts
ebadiere: So I was a little bit off on that response from the mirrornode.  When "estimate": true is included in the post request then the following response is returned:
```
{
    "_status": {
        "messages": [
            {
                "message": "data field invalid hexadecimal string",
                "detail": "",
                "data": ""
            }
        ]
    }
}
```

Test was updated to cover it.

packages/relay/src/lib/eth.ts
ebadiere: Actually the test 'SDK returns a precheck error' expects only the 'detail' in the error.  Updated to conform to the test.

package-lock.json
Nana-EC: Yeah good idea. We should set an npm version and document it. Please open a ticket for that

packages/relay/tests/helpers.ts
```diff
@@ -402,6 +402,18 @@ const mockData = {
         }
     },
 
+    notSuported: {
```
Nana-EC, line 405: This already exist in the object and is a duplicate

package-lock.json
```diff
@@ -9673,6 +9673,7 @@
             "resolved": "https://registry.npmjs.org/bufferutil/-/bufferutil-4.0.5.tgz",
             "integrity": "sha512-HTm14iMQKK2FjFLRTM5lAVcyaUzOnqbPtesFIvREgXpJHdQm8bWS+GkQgIkfaBYRHuCnea7w8UVNfwiAQhlr9A==",
             "dev": true,
+            "hasInstallScript": true,
```
Nana-EC, line 9676: nit: rebase as I don't think these changes are needed.

packages/relay/tests/helpers.ts
ebadiere: Removed.

package-lock.json
ebadiere: I am not sure how this got marked as added.  Git says it was added by Nikolay 5 months ago.
Screenshot 2023-07-11 at 9 28 20 PM

Nana-EC: Strange. The `hasInstallScript` shouldn't be needed but the snapshot updates should already be in teh repo with [this PR ](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1244/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519)

packages/relay/tests/helpers.ts
Nana-EC: It's still there. I don't think you pushed your removal

packages/relay/src/lib/eth.ts
```diff
@@ -524,6 +524,12 @@ export class EthImpl implements Eth {
           return predefined.INVALID_PARAMETER(0, `Invalid 'value' field in transaction param. Value must be greater than 0`);
         }
       } else {
+        // The size limit of the encoded contract posted to the mirror node can cause contract deployment transactions to fail with a 400 response code.
+        // The contract is actually deployed on the consensus node, so the contract will work.  In these cases, we don't want to return a 
+        // CONTRTACT_REVERT error.
+        if (e.isContractReverted() && e.message !== MirrorNodeClientError.messages.INVALID_HEX && process.env.ALLOW_INVALID_HEX_RESPONSE) {
```
Nana-EC, line 530: This will pass so far as `process.env.ALLOW_INVALID_HEX_RESPONSE` has a value vs making sure it's true.
Also i think as it is it's operating more as a `ESTIMATE_GAS_THROWS` vs what you have it named as. If the value is false you won't throw at all.

I think the below is more appropriate.
`private readonly estimateGasThrows = process.env. ESTIMATE_GAS_THROWS ? process.env. ESTIMATE_GAS_THROWS === "true" : true;`
```suggestion
        if (estimateGasThrows && e.isContractReverted() && e.message !== MirrorNodeClientError.messages.INVALID_HEX) {
```

Nana-EC: This would also need a test to make sure we can toggle appropriately.

Nana-EC (commented):
One final improvement in addition to the package-lock.json revert

packages/relay/src/lib/eth.ts
ebadiere: Updated.

packages/relay/src/lib/eth.ts
```diff
@@ -126,6 +127,7 @@ export class EthImpl implements Eth {
     parseNumericEnvVar('ETH_GET_BLOCK_BY_RESULTS_BATCH_SIZE', 'ETH_GET_BLOCK_BY_RESULTS_BATCH_SIZE');
   private readonly MirrorNodeGetContractResultRetries =
     parseNumericEnvVar('MIRROR_NODE_GET_CONTRACT_RESULTS_RETRIES', 'MIRROR_NODE_GET_CONTRACT_RESULTS_DEFAULT_RETRIES');
+  private readonly estimateGasThrows = process.env. ESTIMATE_GAS_THROWS ? process.env. ESTIMATE_GAS_THROWS === "true" : true;
```
Nana-EC, line 130:
```suggestion
  private readonly estimateGasThrows = process.env.ESTIMATE_GAS_THROWS ? process.env. ESTIMATE_GAS_THROWS === "true" : true;
```

packages/relay/src/lib/eth.ts
```diff
@@ -524,6 +526,12 @@ export class EthImpl implements Eth {
           return predefined.INVALID_PARAMETER(0, `Invalid 'value' field in transaction param. Value must be greater than 0`);
         }
       } else {
+        // The size limit of the encoded contract posted to the mirror node can cause contract deployment transactions to fail with a 400 response code.
+        // The contract is actually deployed on the consensus node, so the contract will work.  In these cases, we don't want to return a 
+        // CONTRTACT_REVERT error.
+        if (this.estimateGasThrows && e.isContractReverted() && e.message !== MirrorNodeClientError.messages.INVALID_HEX) {
```
Nana-EC, line 532: Missing test for when `this.estimateGasThrows` is `false` to confirm the endpoint doesn't throw a revert

docs/configuration.md
```diff
@@ -45,6 +45,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
 
 | Name                                  | Default       | Description                                                                                                                                                                                                                                                                                            |
 | ------------------------------------- | ------------- |--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `ALLOW_INVALID_HEX_RESPONSE`          | "true"        | Account for the benign 400 response with message 'data field invalid hexadecimal string' coming back from the mirror node when deploying a contract.
```
Nana-EC, line 48: Need to update this
`ESTIMATE_GAS_THROWS` => Something to describe that this decided is `eth_estimateGas` throws a revert error on execution reverts.

Nana-EC: Feel free to rename if `ESTIMATE_GAS_THROWS` isn't descriptive enough and needs to be `ESTIMATE_GAS_EXPOSES_REVERT` or something else instead.

ebadiere: Updated.

packages/relay/src/lib/eth.ts
ebadiere: Added test.

packages/relay/src/lib/eth.ts
ebadiere: Fixed.

docs/configuration.md
```diff
@@ -45,6 +45,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
 
 | Name                                  | Default       | Description                                                                                                                                                                                                                                                                                            |
 | ------------------------------------- | ------------- |--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `ESTIMATE_GAS_THROWS`                 | "true"        | eth_estimateGase throws an error if the encoded contract is bigger than 12288 bytes when posting to the mirror node.  Set this to true to ignore.
```
Nana-EC, line 48: Wrong description.
```suggestion
| `ESTIMATE_GAS_THROWS`                 | "true"        | Flag to enable or disable whether eth_estimateGas should throw an error on contract revert 400 HTTP responses from the mirror node.
```

package-lock.json
```diff
@@ -81,9 +81,9 @@
             }
         },
         "node_modules/@babel/core": {
-            "version": "7.22.7",
-            "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.22.7.tgz",
-            "integrity": "sha512-exABdCVjEk8+IFJW0gOK6+cou8VKMXfbkLGeK5Xdsa5MsuQmem1SsnnZ+6avm2gRhZ4M7UgAnE6YoAzVg9P/pw==",
+            "version": "7.22.8",
```
Nana-EC, line 84: Revert the changes in this file. 
I think your rebase might have gone wrong here

Nana-EC (commented):
Logic is good now.
Just the config description and the `package-lock.json`.

docs/configuration.md
ebadiere: Done.

package-lock.json
ebadiere: Reverted it, and then ran ```npm install``` and checked it in.

georgi-l95:
@ebadiere We should go to fallback logic, if error from mirror-node is 400 and msg is empty. That means that there is probably a NPE in the mirror-node due to bug. It's temporary until all precompiles are implemented in the mirror-node and all bugs are cleared.

PR #1432, opened 2023-07-01
Fix clientCache metric collection

### LINKED ISSUE #1431: ClientCache metrics should be a gauge not a counter
### Description

Metric collection fo the centralied client cache is failing on collection
```
  TypeError: this.cacheKeyCounter.set is not a function
  at cacheSizeCollect (.../hedera-json-rpc-relay/packages/relay/dist/lib/clients/clientCache.js:57:34)
  at Counter.<anonymous> (.../hedera-json-rpc-relay/packages/relay/dist/lib/clients/clientCache.js:68:21)
  at .../hedera-json-rpc-relay/packages/relay/dist/lib/clients/clientCache.js:27:71
  at new Promise (<anonymous>)
  at __awaiter (.../hedera-json-rpc-relay/packages/relay/dist/lib/clients/clientCache.js:23:12)
  at Counter.collect (.../hedera-json-rpc-relay/packages/relay/dist/lib/clients/clientCache.js:67:24)
  at Counter.get (.../hedera-json-rpc-relay/node_modules/prom-client/lib/counter.js:54:19)
  at Registry.getMetricAsPrometheusString (.../hedera-json-rpc-relay/node_modules/prom-client/lib/registry.js:26:29)
  at Registry.metrics (.../hedera-json-rpc-relay/node_modules/prom-client/lib/registry.js:70:23)
```

This is since the metric object is using Gauge logic to do a `.set()` for the cache size even though it's a COunter.
Instead the metric should actually be a Gauge since the size can go up and down

### Steps to reproduce

1. Deploy 0.26.1
2. trigger metrics collection

### Additional context

_No response_

### Hedera network

previewnet

### Version

0.26.1

### Operating system

None

Nana-EC:
**Description**:
Fix clientCache metric collection
- Convert metric from Counter to Gauge since values can also decrease
- Add tests in `server.spec.ts` to ensure metrics and health endpoints are functional

**Related issue(s)**:

Fixes #1431 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/clientCache.ts
+++ b/packages/relay/src/lib/clients/clientCache.ts
@@ -81,13 +81,9 @@ export class ClientCache {
     }
 
     public get(key: string, callingMethod?: string, requestIdPrefix?: string): any {
-        let value = this.cache.get(key);
+        const value = this.cache.get(key);
         if (value) {
-            if (callingMethod) {
-                this.cacheKeyGauge.labels(key, callingMethod).inc(1);
-            } else {
-                this.cacheKeyGauge.labels(key, '').inc(1);
-            }
+            this.cacheKeyGauge.labels(key, callingMethod || '').inc(1);
 
             this.logger.trace(`${requestIdPrefix} returning cached value ${key}:${JSON.stringify(value)} on ${callingMethod} call`);
             return value;
--- a/packages/relay/src/lib/clients/clientCache.ts
+++ b/packages/relay/src/lib/clients/clientCache.ts
@@ -19,7 +19,7 @@
  */
 
 import { Logger } from "pino";
-import { Counter, Registry } from "prom-client";
+import { Gauge, Registry } from "prom-client";
 import constants from "../constants";
 
 const LRU = require('lru-cache');
@@ -69,9 +69,9 @@ export class ClientCache {
 
         const metricCounterName = 'rpc_relay_cache';
         register.removeSingleMetric(metricCounterName);
-        this.cacheKeyCounter = new Counter({
+        this.cacheKeyCounter = new Gauge({
             name: metricCounterName,
-            help: 'Relay cache counter',
+            help: 'Relay cache gauge',
             labelNames: ['key', 'method'],
             registers: [register],
             async collect() {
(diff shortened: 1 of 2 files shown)


packages/relay/src/lib/clients/clientCache.ts
```diff
@@ -84,9 +84,9 @@ export class ClientCache {
         let value = this.cache.get(key);
```
AlfredoG87, line 84: This could be a const

packages/relay/src/lib/clients/clientCache.ts
```diff
@@ -84,9 +84,9 @@ export class ClientCache {
...
         if (value) {
             if (callingMethod) {
-                this.cacheKeyCounter.labels(key, callingMethod).inc(1);
+                this.cacheKeyGauge.labels(key, callingMethod).inc(1);
             } else {
-                this.cacheKeyCounter.labels(key, null).inc(1);
+                this.cacheKeyGauge.labels(key, '').inc(1);
             }
```
AlfredoG87, line 90: Here Gauge is being used like a counter, no big issue, but might not be ideal.
Also, this could be simplified for readability to:

```
this.cacheKeyGauge.labels(key, callingMethod || '').inc(1);

```

AlfredoG87 (commented):
LGTM, just a couple nits.

packages/relay/src/lib/clients/clientCache.ts
Nana-EC: Yeah it's a supported method of Gauge.
Let me know why it would not be ideal?

Nana-EC: Adopted the simplification suggestion

AlfredoG87: you are right, I looked and indeed is a supported method, I guess there is no problem.
Just that when I imagine a gauge I think on something that goes up and down and has some thresholds. like memory, cpu, temperature, active connections, etc... but to be honest, I don't see an issue either. 👍

AlfredoG87 (dismissed):
LGTM

georgi-l95 (dismissed):
LG

AlfredoG87 (approved):
LGTM

PR #1434, opened 2023-07-03
Fixes try catch tests and remove unecessary asseritons

konstantinabl:
**Description**:
Currently in the acceptance tests try catch logic is used, however we do not check whether an error is actually thrown and we enter the catch. Therefore, this PR changes the logic so we take the maximum of the chai library in order to make the tests effective and intuitive

**What is included**

1. Added chai as promised library for testing on promises
2. Replaced try catch with suitable chai methods, which detect if an error is not thrown
3. Added missing semicolons
4. Replaced some uses of let with const
5. Removed the unnecessary use of Assertions

### DIFF
--- a/dapp-example/tests/e2e/specs/index.spec.js
+++ b/dapp-example/tests/e2e/specs/index.spec.js
@@ -35,6 +35,9 @@ describe('Test Core Hedera User Scenarios', function() {
 
       // deploy the contract
       cy.get('#btnDeployContract').should('not.be.disabled').click();
+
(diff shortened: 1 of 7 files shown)


konstantinabl:
Added temporary fix in dapp-example/tests/e2e/specs/index.spec.js due to issue in synpress library:
https://github.com/Synthetixio/synpress/issues/795

Ivo-Yankov (approved):
LGTM

Nana-EC (approved):
LG

PR #1446, opened 2023-07-05
Improve `eth_sendRawTransaction` latency by simplifying precheck

### LINKED ISSUE #1443: Improve `eth_sendRawTransaction` latency
### Problem

As part of the effort to improve HashIO performance, we need to look at the `eth_sendRawTransaction` method and determine where we can it's performance.

### Solution

We can do this by reducing the number of requests towards the mirror-node or making them asynchronous them. Also we need to look at the file creation function, can we improve it.
After and before the changes compare using the grafana dashboard.

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR reduces `eth_sendRawTransaction` latency by reducing the number of requests and simplifying precheck.
Now we make 1 less request to the mirror node and 1 less request to consensus node. This reduces both hbar cost and latency. Performance test was done and metrics we collected by prometheus, by average transactions now take 10% less.

Also remove the FileAppendTransaction getReceipt, as transaction in hedera are executed in the order they are received, so we know that the next one will be executed after this one, no need to wait.

Also remove the getReceipt for file
**Related issue(s)**:

Fixes #1443 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -550,16 +550,14 @@ export class SDKClient {
         );
 
         if (fileId && callData.length > 4096) {
-            await (
-                await new FileAppendTransaction()
-                    .setFileId(fileId)
-                    .setContents(
-                        hexedCallData.substring(4096, hexedCallData.length)
-                    )
-                    .setChunkSize(4096)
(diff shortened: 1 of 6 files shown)


Ivo-Yankov (dismissed):
Looks good

Nana-EC (approved):
Nice one.

PR #1469, opened 2023-07-08
Improve handling of boolean cache retrieval and metric labels

### LINKED ISSUE #1468: ClientCache doesn't handle boolean retrieval well
### Description

The ClientCache has a `get` method.
The check for a valid cache value uses `if (value)`
For most cases this confirm non undefined values, however fails to consider when the value is `false`

### Steps to reproduce

1. Cause a cache set
2. Cause a cache get

Logs will highlight the code tries to set against because it believes the cache for the given key is empty

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.26.3

### Operating system

None

Nana-EC:
**Description**:
Fix boolean `false` issue on cache
- Expand metrics to manage set scenarios also
- Ensure all metrics have calling method
- Update cacheClient to manage where the cached value is false and not treat it as undefined

**Related issue(s)**:

Fixes #1468 

**Notes for reviewer**:

**Checklist**


packages/relay/tests/lib/clients/clientCache.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      it('verify cache ttl nature', async function () {
+        const customClientCache = new ClientCache(logger.child({ name: `cache` }), registry);
+        const key = 'key';
+        customClientCache.set(key, 'value', callingMethod, 100); // set ttl to 1 ms
+        await new Promise(r => setTimeout(r, 500)); // wait for ttl to expire
+        const cacheValue = customClientCache.get(key, callingMethod);
+        expect(cacheValue).to.be.null;
+      });
+    });
+});
```
AlfredoG87, line 138: EOF Blank line.

AlfredoG87 (approved):
Overall it looks good to me.
Just 1 nit, and a question/suggestion:

I try to avoid creating labels for metrics that cannot be aggregated or that every instance will be a different value, in this case is `key`.

I raised a similar concern when we added the `interactingEntity` in the past.

My rationale is that they add little value and add a lot of compute and storage load, not sure for how long are we storing the metrics, but including `key` in this context will make it grow exponentially faster that if we don't and not sure we really get a valuable information from including it. other than for those keys that really don't change that much, but for short lived TTL caches, including the key might not really yield that much valuable information and it does add a ton of storage and compute load.

Let me know your thoughts.

ebadiere (approved):
Other than the EOF Blank line mentioned by @AlfredoG87 , LG.

AlfredoG87 (dismissed):
LGTM

AlfredoG87 (approved):

PR #1496, opened 2023-07-10
Optimize eth_getTransactionCount

### LINKED ISSUE #1444: Improve `eth_getTransactionCount` latency
### Problem

As part of the effort to improve HashIO performance, we need to look at the `eth_getTransactionCount` method and determine where we can improve it's performance.

### Solution

We can do this by reducing the number of requests towards the mirror-node or making them asynchronous them.
After and before the changes compare using the grafana dashboard.

### Alternatives

_No response_

Ivo-Yankov:
**Description**:

Reduces the execution time of `eth_getTransactionCount` by reversing the order of getting data from the Mirror node. First the account is retrieved and after that it is checked if the address is a contract. The `isValidContract()` method is called with a param that specifies that no retries should be made on 404 status. On average this should reduce the execution time by around 1000-1200 ms.

**Related issue(s)**:

Fixes #1444 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -1932,16 +1932,29 @@ export class EthImpl implements Eth {
   }
 
   private async getAccountLatestEthereumNonce(address: string, requestId?: string) {
-    // check if address is a valid contract then get ethereumNonce from consensus node until HIP 729 is implemented
-    const validContract = await this.mirrorNodeClient.isValidContract(address, requestId);
-    if (validContract)  {
+    const cachedLabel = `${constants.CACHE_KEY.RESOLVE_ENTITY_TYPE}_${address}`;
```
Nana-EC, line 1935: Isn't this approach essentially extracting the logic out?
`this.mirrorNodeClient.isValidContract` has caching logic.

In which scenario do we reduce query attempts to the Mirror Node?
`constants.CACHE_KEY.RESOLVE_ENTITY_TYPE` currently caches for 1hr.
`isValidContract` is caching for 1 day and was supposed to be an improvement.

packages/relay/src/lib/eth.ts
```diff
@@ -1932,16 +1932,29 @@ export class EthImpl implements Eth {
   }
 
   private async getAccountLatestEthereumNonce(address: string, requestId?: string) {
```
Nana-EC, line 1934: The goal here was to wait for the HIP 729 implementation and then replace `isValidContract`.
Since the Mirror Node also manages accounts and contracts on the `api/v1/accounts` endpoint this whole method may just get simplified to the single `getAccount` call.

I think rather another consideration in the mean time is to
1. update mirrorNodeClient.getContract to populate the same cache that isValidContract gets.
2. [Stretch] Add an isValidAccount() and update mirrorNodeClient.getAccount to populate the same cache that isValidAccount gets. This was the intention but wasn't needed when improving `eth_getTransactionCount()`

Both these optimization should help reduce the number of calls to resolveEntityType

Nana-EC (changes requested):
Not sure of the changes.
The goal of isValid...() was to reduce the need to keep calling the more complex resolveEntityType which should be reserved for cases where the entity type could be account, contract or token.

In this case we know it's either contract or account.
I think we just need some additional cache stitching to tie together the flows and see when the HIP 729 improvement will come in to simplify this even further

Ivo-Yankov:
I reworked the initial solution.

packages/relay/src/lib/eth.ts
```diff
@@ -1897,17 +1897,15 @@ export class EthImpl implements Eth {
...
-    } 
-    
-    // get latest ethereumNonce from mirror node account API
-    const mirrorAccount = await this.mirrorNodeClient.getAccount(address, requestIdPrefix);
-    if (mirrorAccount?.ethereum_nonce) {
-      return EthImpl.numberTo0x(mirrorAccount.ethereum_nonce);
+  private async getAccountLatestEthereumNonce(address: string, requestId?: string) {
+    const accountData = await this.mirrorNodeClient.getAccount(address, requestId);
+    if (accountData) {
+      const contract = await this.mirrorNodeClient.isValidContract(address, requestId, 0);
```
Nana-EC, line 1903: Q: should this be outside of the if statement and before the  return 0?

Nana-EC (changes requested):
I like the retry addition.
I think the logic needs updating to call to not call the contract endpoint if you confirm it's an account.

packages/relay/src/lib/eth.ts
```diff
@@ -1897,17 +1897,24 @@ export class EthImpl implements Eth {
...
-    
-    // get latest ethereumNonce from mirror node account API
-    const mirrorAccount = await this.mirrorNodeClient.getAccount(address, requestIdPrefix);
-    if (mirrorAccount?.ethereum_nonce) {
-      return EthImpl.numberTo0x(mirrorAccount.ethereum_nonce);
+    }
+
+    const accountData = await this.mirrorNodeClient.getAccount(address, requestId);
+    if (accountData) {
+      if (!accountData.ethereum_nonce) {  // if nonce > 0 then the entity cannot be a contract
```
Nana-EC, line 1910: Q: just to confirm this works as expected when nonce is 0. Can add a test to be sure.

Nana-EC (approved):
LG.
One Q and test suggestion if applicable

packages/relay/src/lib/eth.ts
Nana-EC: Confirmed offline

Nana-EC:
@Ivo-Yankov looks like you need to rebase.

Nana-EC (approved):
LG

ebadiere (approved):
LG

Issue #1402, opened 2023-06-28
Unify the methods used for sending transactions in relay acceptance tests


konstantinabl:
### Problem

Currently, in the hedera-json-rpc-relay acceptance tests we use different methods for sending transactions to the relay, which may lead to confusion and questions

### Solution

We should agree on the methods used for the tests and use it everywhere

PR #1497, opened 2023-07-10
Changes way of sending raw transaction instead of relay.call with rel…

konstantinabl:
**Description**:
This PR unifies the way raw transactions are sent. Instead of using relay.call, we use directly relay.sendRawTransaction

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/1402


packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -70,6 +70,14 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
     let blockNumberAtStartOfTests = 0;
     let mirrorAccount0AtStartOfTests;
 
+    const signAndSendTransaction = async (transaction, accounts, requestId) => {
```
Nana-EC, line 73: nit: naming update suggestion
```suggestion
    const signSendAndConfirmTransaction = async (transaction, accounts, requestId) => {
```

Nana-EC (approved):
LG, I like the update.
One minor suggested naming update.

Nana-EC (approved):
LGTM

Nana-EC (approved):

PR #1505, opened 2023-07-12
Improve logging across classes (#1244)

### LINKED ISSUE #1468: ClientCache doesn't handle boolean retrieval well

ebadiere:
* Improved eth logging

* Improve logging in server side

* Cleaned up rebase

* Fix missing replaces

* Revert package-lock change

* minor string improvements in mirrrNodeClient

* Fix get balance tests

* Swap in isValid check and setup wait

* Add log message of failure

* Fix request id on web socket messages

* fix getBalance cache

---------

**Description**:

**Related issue(s)**:

Fixes #1468

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19990,7 +19990,7 @@
         },
         "packages/relay": {
             "name": "@hashgraph/json-rpc-relay",
-            "version": "0.27.0-rc2",
+            "version": "0.28.0-SNAPSHOT",
             "dependencies": {
                 "@ethersproject/asm": "^5.7.0",
                 "@hashgraph/sdk": "^2.29.0",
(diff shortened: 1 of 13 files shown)


Nana-EC:
Not needed in 0.27

Issue #1139, opened 2023-04-27
Relay returns 502 on contract deploys


mshakeg:
### Description

When deploying contracts to the mainnet using the hashio relay I've frequently experienced the relay returning the following response:
```
<html><head>
<meta http-equiv="content-type" content="text/html;charset=utf-8">
<title>502 Server Error</title>
</head>
<body text=#000000 bgcolor=#ffffff>
<h1>Error: Server Error</h1>
<h2>The server encountered a temporary error and could not complete your request.<p>Please try again in 30 seconds.</h2>
<h2></h2>
</body></html>
```

It seems like a timeout error, however, the contract does in fact deploy and reflect on hashio about 20s after the above response is returned.

### Steps to reproduce

Try deploy relatively large contracts(that take longer to deploy) to mainnet.

### Hedera network

mainnet

### Version

relay/0.22.1

### Operating system

None

Nana-EC:
@mshakeg we're investigating

Nana-EC:
@mshakeg we've made some improvements across the board and hope to have resolved this.
Can you confirm if you're still seeing this and if it's only in mainnet instance

mshakeg:
@Nana-EC I don't need to deploy any contracts to mainnet atm.

Can the ops team not ascertain whether this is still an issue based on recent (large) contract deploys via the hashio relay?

mshakeg:
This issue now appears to also be affecting testnet :(

mshakeg:
Hey @AlfredoG87 has any progress been made on this issue? The issue seems to have been exacerbated since at least on testnet as the contract does not even deploy(I've been trying to deploy this contract for about a day now). I don't have an issue deploying the same contract to previewnet.

Nana-EC:
@mshakeg we're making progress on the investigation.
We'll circle back once we have more details and a direction. Thanks

AlfredoG87:
Hi @mshakeg,

same as issue [1323](https://github.com/hashgraph/hedera-json-rpc-relay/issues/1323) we will be closing this issue, but please let us know if the issue persists on `mainnet` after we have completed our upgrade of mainnet to `v0.28.0` that should happen mid next week.

PR #1519, opened 2023-07-14
Using better-lookup to improve dns resolution performance for each outgoing requests

### LINKED ISSUE #1507: Implement a Cached DNS Lookup for MirrorNode Axios Client
### Problem

Under heavy load and in Cloud environments, the relay is experiencing long latency times when calling the mirror node.

I believe that it could be due to how nodejs handles concurrent  `outgoing` requests

https://httptoolkit.com/blog/configuring-nodejs-dns/

### Solution

This would be fixed by implementing a cached dns lookup.

luckily there is already a fix out there for axios clients:

https://www.npmjs.com/package/axios-cached-dns-resolve

### Alternatives

Change the host to the actual IP instead of the FQDN

Is not possible to access the production mirror nodes by IP.

AlfredoG87:
**Description**:
Using **better-lookup** to improve dns resolution performance for each outgoing requests

**Related issue(s)**: #1123, #1139 

Fixes #1507 

**Notes for reviewer**:

This is an alternative to PR #1508, that was a more popular option, but it is not compatible with CommonJS

**About this package:**
https://www.npmjs.com/package/better-lookup
https://github.com/ayonli/better-lookup

**Checklist**


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -173,6 +174,25 @@ export class MirrorNodeClient {
...
+
+        const httpsAgent = new https.Agent({
+            keepAlive: mirrorNodeHttpKeepAlive,
+            keepAliveMsecs: mirrorNodeHttpKeepAliveMsecs,
+            maxSockets: mirrorNodeHttpMaxSockets,
+            maxTotalSockets: mirrorNodeHttpMaxTotalSockets,
+            timeout: mirrorNodeHttpSocketTimeout,
+        });
+
+        betterLookupInstall(httpAgent);
```
Nana-EC, line 193: Q: should we wrap this in a feature flag?

Nana-EC (commented):
LG. 1 Q and likely recommendation

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: Yeah, I agree we can wrap it in a feature flag, I recommend that by default the value of the flag is set to true, since we are expecting this to be the performant option.

AlfredoG87: ✅  Done

Nana-EC (approved):

ebadiere (approved):
LG.

PR #1534, opened 2023-07-18
Returned bytecode is incorrect after selfdestruct and recreation via create2

### LINKED ISSUE #1533: Returned bytecode is incorrect after selfdestruct and recreation via create2
### Description

The relay caches non-existing contracts or deleted ones.

### Steps to reproduce

- Create a contract via create2
- selfdestruct it
- Call `.getCode` // will cache the response because the consensus will throw CONTRACT_DELETED
- Re-deploy the contract via create2
 - Call `.getCode` // will return the cached response with empty bytecode which is incorrect

### Additional context

_No response_

### Hedera network

other

### Version

latest

### Operating system

None

natanasow:
**Description**:

The relay caches non-existing contracts or deleted ones, which is a problem when we redeploy contracts via create2.

**Related issue(s)**:

Fixes #1533 

**Notes for reviewer**:

**Checklist**


Nana-EC (approved):
LGTM

PR #1535, opened 2023-07-18
Update eth_getTransactionCount to return contract nonce

Ivo-Yankov:
**Description**:
- Simplifies the existing `getTransactionCount` logic. There is no longer the need to check the entity type - both contracts and accounts have a nonce field and it is always the same.
- Bumps the versions of services and mirror node images in the acceptance tests.
- Depends on #1549

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


AlfredoG87 (dismissed):
LGTM 👍

packages/relay/src/lib/eth.ts
```diff
@@ -1905,22 +1905,8 @@ export class EthImpl implements Eth {
...
     const accountData = await this.mirrorNodeClient.getAccount(address, requestId);
     if (accountData) {
-      if (!accountData.ethereum_nonce) {  // if nonce > 0 then the entity cannot be a contract
-        const contract = await this.mirrorNodeClient.isValidContract(address, requestId, 0);
-        if (contract) {
-          return EthImpl.oneHex;
-        }
-      }
-
       return EthImpl.numberTo0x(accountData.ethereum_nonce);
```
Nana-EC, line 1910: Actually we now have to handle the null case.
Historical contracts that didn't track the nonce will return a null on this.
In this case you can be assured it's a contract and you should send back `EthImpl.oneHex`
e.g. https://previewnet.mirrornode.hedera.com/api/v1/accounts/0.0.1036

Nana-EC:
```suggestion
      // with HIP 729 ethereum_nonce should always be 0+ and null. Historical contracts may have a null value as the nonce was not tracked, return default EVM compliant 0x1 in this case
      return accountData.ethereum_nonce ? EthImpl.numberTo0x(accountData.ethereum_nonce) : EthImpl.oneHex;
```

Nana-EC (changes requested):
Realized we need to handle and test for null response on historical contracts

Nana-EC (dismissed):

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -350,7 +350,16 @@ export class MirrorNodeClient {
     handleError(error: any, path: string, pathLabel: string, effectiveStatusCode: number, method: REQUEST_METHODS, requestIdPrefix?: string) {
         const mirrorError = new MirrorNodeClientError(error, effectiveStatusCode);
         const acceptedErrorResponses = MirrorNodeClient.acceptedErrorStatusesResponsePerRequestPathMap.get(pathLabel);
-        if (error.response && acceptedErrorResponses && acceptedErrorResponses.indexOf(effectiveStatusCode) !== -1) {
+
+        const acceptError = (
+            // Handles a FAIL_INVALID error thrown in the case of a non-existing contract in /contracts/call
```
Nana-EC, line 355: This should be managed at the `eth_x` method level

Nana-EC: `handleError` should really only manage fundamental retries and then throw a valid `MirrorNOdeClientError`.
Eth methods should decide for themselves how to handle the errors.

Nana-EC (commented):
Looking good. Test fixes but also the mirrorNoceClient.handleError change I think should be revert and the management logic pushed to the eth classes

Nana-EC (dismissed):

ebadiere (approved):
LG.

Nana-EC (approved):

PR #1536, opened 2023-07-18
subscriptions_enabled=true is necessary to enable the websockets func…

rustyShacklefurd:
This should always be true when deploying the websockets

**Description**:
This PR modifies the `SUBSCRIPTIONS_ENABLED` key/value in the websockets configmap.   This value needs to be true to enable the websocket functionality.


AlfredoG87 (approved):
LGTM

Nana-EC (approved):

PR #1537, opened 2023-07-18
refactor getContractByteCode to set cost based on the sdk assessment

### LINKED ISSUE #1523: Make setMaxQueryPayment configurable
### Problem

Currently, the JSON RPC relay uses the Hedera JavaScript SDK to query smart contract bytecodes from the consensus node 
https://github.com/hashgraph/hedera-json-rpc-relay/blob/3c2fe6784aeb973f4686c7ce32bd1dba1e25e6ac/packages/relay/src/lib/clients/sdkClient.ts#L138
The current default max query payment is 1 hbar, and it's not configurable from the relay. Therefore, it can cause some queries for big smart contracts to fail because the max query payment is set too low.

### Solution

- Make `setMaxPaymentQuery` configurable from the relay environment variable.

### Alternatives

_No response_

AlfredoG87:
**Description**:
refactor getContractByteCode to set cost based on the sdk assessment using getCost method

**Related issue(s)**:

Fixes #1523 

**Notes for reviewer**:
Did not include a test case, since if every tests we already have continues to work as expected, there should not be an issue, also, testing this overlaps with testing the sdk itself, which is outside the relay and is already tested by the sdk test suites

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -136,8 +136,12 @@ export class SDKClient {
     }
 
     async getContractByteCode(shard: number | Long, realm: number | Long, address: string, callerName: string, requestId?: string): Promise<Uint8Array> {
+        const contractByteCodeQuery = new ContractByteCodeQuery().setContractId(ContractId.fromEvmAddress(shard, realm, address));
+        const cost = await contractByteCodeQuery
+            .getCost(this.clientMain);
+
         return this.executeQuery(new ContractByteCodeQuery()
-            .setContractId(ContractId.fromEvmAddress(shard, realm, address)), this.clientMain, callerName, address, requestId);
+            .setQueryPayment(cost), this.clientMain, callerName, address, requestId);
     }
 
     async getContractBalance(contract: string, callerName: string, requestId?: string): Promise<AccountBalance> {
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -135,13 +135,18 @@ export class SDKClient {
             .setAccountId(AccountId.fromString(address)), this.clientMain, callerName, address, requestId);
     }
 
-    async getContractByteCode(shard: number | Long, realm: number | Long, address: string, callerName: string, requestId?: string): Promise<Uint8Array> {
-        const contractByteCodeQuery = new ContractByteCodeQuery().setContractId(ContractId.fromEvmAddress(shard, realm, address));
-        const cost = await contractByteCodeQuery
-            .getCost(this.clientMain);
+    async getContractByteCode(shard: number | Long, realm: number | Long, address: string, callerName: string, hasProhibitedOpCodes: boolean, requestId?: string): Promise<Uint8Array> {
+        if(hasProhibitedOpCodes) {
+            return this.executeQuery(new ContractByteCodeQuery()
+                .setContractId(ContractId.fromEvmAddress(shard, realm, address)), this.clientMain, callerName, address, requestId);
+        } else {
+            const contractByteCodeQuery = new ContractByteCodeQuery().setContractId(ContractId.fromEvmAddress(shard, realm, address));
+            const cost = await contractByteCodeQuery
+                .getCost(this.clientMain);
 
-        return this.executeQuery(new ContractByteCodeQuery()
-            .setQueryPayment(cost), this.clientMain, callerName, address, requestId);
+            return this.executeQuery(new ContractByteCodeQuery()
+                .setQueryPayment(cost), this.clientMain, callerName, address, requestId);
+        }
     }
 
     async getContractBalance(contract: string, callerName: string, requestId?: string): Promise<AccountBalance> {
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -135,18 +135,13 @@ export class SDKClient {
             .setAccountId(AccountId.fromString(address)), this.clientMain, callerName, address, requestId);
     }
 
-    async getContractByteCode(shard: number | Long, realm: number | Long, address: string, callerName: string, hasProhibitedOpCodes: boolean, requestId?: string): Promise<Uint8Array> {
-        if(hasProhibitedOpCodes) {
-            return this.executeQuery(new ContractByteCodeQuery()
-                .setContractId(ContractId.fromEvmAddress(shard, realm, address)), this.clientMain, callerName, address, requestId);
-        } else {
-            const contractByteCodeQuery = new ContractByteCodeQuery().setContractId(ContractId.fromEvmAddress(shard, realm, address));
-            const cost = await contractByteCodeQuery
-                .getCost(this.clientMain);
+    async getContractByteCode(shard: number | Long, realm: number | Long, address: string, callerName: string, requestId?: string): Promise<Uint8Array> {
+        const contractByteCodeQuery = new ContractByteCodeQuery().setContractId(ContractId.fromEvmAddress(shard, realm, address));
+        const cost = await contractByteCodeQuery
+            .getCost(this.clientMain);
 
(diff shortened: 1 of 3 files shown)


Nana-EC (approved):

packages/relay/src/lib/constants.ts
```diff
@@ -18,6 +18,8 @@
   *
   */
 
+import {PrecheckStatusError} from "@hashgraph/sdk";
```
Nana-EC, line 21: Q: where is this used?

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -135,9 +135,18 @@ export class SDKClient {
             .setAccountId(AccountId.fromString(address)), this.clientMain, callerName, address, requestId);
     }
 
-    async getContractByteCode(shard: number | Long, realm: number | Long, address: string, callerName: string, requestId?: string): Promise<Uint8Array> {
-        return this.executeQuery(new ContractByteCodeQuery()
-            .setContractId(ContractId.fromEvmAddress(shard, realm, address)), this.clientMain, callerName, address, requestId);
+    async getContractByteCode(shard: number | Long, realm: number | Long, address: string, callerName: string, hasProhibitedOpCodes: boolean, requestId?: string): Promise<Uint8Array> {
+        if(hasProhibitedOpCodes) {
```
Nana-EC, line 139: Q: why the difference based on `hasProhibitedOpCodes`?
Shouldn't we query for cost regardless?

It should be possible to have a cost over the default even for prohibited contents no?

Nana-EC (changes requested):
Good catch.
A couple of comments. Also please update PR description and add a test the sdkClient as there's now a conditional path

packages/relay/src/lib/eth.ts
```diff
@@ -902,6 +903,16 @@ export class EthImpl implements Eth {
 
         this.hapiService.decrementErrorCounter(e.statusCode);
         this.logger.error(e, `${requestIdPrefix} Error raised during getCode for address ${address}, err code: ${e.statusCode}`);
+      } if(e instanceof PrecheckStatusError) {
+        if(e.status._code === constants.PRECHECK_STATUS_ERROR_STATUS_CODES.INVALID_CONTRACT_ID ||
+            e.status._code === constants.PRECHECK_STATUS_ERROR_STATUS_CODES.CONTRACT_DELETED ) {
+          this.logger.debug(`${requestIdPrefix} Unable to find code for contract ${address} in block "${blockNumber}", returning 0x0, err code: ${e.message}`);
+          return EthImpl.emptyHex;
```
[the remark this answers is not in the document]
AlfredoG87: I would say I meant `emptyHex` since is the same error as when the type is `e instanceof SDKClientError` but for `e instanceof PrecheckStatusError`

packages/relay/src/lib/constants.ts
AlfredoG87: in the catch, for new condition `if(e instanceof PrecheckStatusError) {`

packages/relay/src/lib/clients/sdkClient.ts
AlfredoG87: It should be possible, but it is not, when I was debugging I found that the cost is calculated, but when doing `executeQuery` the behaviour of the sdk is different, my guess is that the sdk under the hood is doing something different when cost is provided.

packages/relay/src/lib/constants.ts
Nana-EC: There's no catch in this file.

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC: What's the difference?
What does the SDK return in the case of executeQuery when `hasProhibitedOpCOdes` is true?

Also maybe `hasProhibitedOpCOdes` is indicative of something else.
It also seems like we're assuming a precCheck error will occur if there are prohibitedOpCOdes, but I don't think that's assured. I think it's only applicable if `SELFDESTRUCT` has been executed on a contract but it may not have yet.

packages/relay/src/lib/constants.ts
AlfredoG87: ohhh, I see it now, hahaha, removing it now. not sure how it got there, thanks for noticing!! 👍

packages/relay/src/lib/clients/sdkClient.ts
AlfredoG87: when we do `executeQuery` with cost it throws errors

```
transaction 0.0.2@1689870961.095051289 failed precheck with status INVALID_CONTRACT_ID

transaction 0.0.2@1689871011.865656621 failed precheck with status CONTRACT_DELETED
```

As oppossed when we it has prohibited OpCodes but we don't define the cost, it returns the expected `ByteCode`

I did the path based on opCodes since that is where the difference lays, but maybe we should change the tests, and instead of expecting the ByteCodes, we should expect `0x` ?

packages/relay/src/lib/constants.ts
AlfredoG87: ✅  Done

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC: That's the challenge we can't make assumptions of the bytecode for some contracts so we need to make the call and see what we get.
A contract may have the prohibited opCodes but it may have not been executed and could be a valid contract with valid non 0x bytecode

Nana-EC: The the prohibited opCodes were only ever to dictate whether we could cache the bytecode or not because you might have valid bytecode at time T1 but after a SELFDESTRUCT is executed it would now be 0x on the mirror node or one of those precheck errors on the consensus node.

This flow is only happening because some `SELFDESTRUCT` cases were not migrated to mirror node so we could get valid bytecode back from the mirror node but the contract was deleted hence the call to the consensus node. In the backup flow we can't make any assumptions of what it could be

AlfredoG87: Okay, I found the bug and removed the conditional path, good thing that we were able to find it 👍 
Thank you!

packages/relay/src/lib/eth.ts
```diff
@@ -902,6 +902,16 @@ export class EthImpl implements Eth {
 
         this.hapiService.decrementErrorCounter(e.statusCode);
         this.logger.error(e, `${requestIdPrefix} Error raised during getCode for address ${address}, err code: ${e.statusCode}`);
+      } if(e instanceof PrecheckStatusError) {
```
Nana-EC, line 905: nit: should there be a new line here?
Is this at the same level of the SDKCLientError check?

Nana-EC: Or is this an `else if`?

Nana-EC (dismissed):
LG.
Nit on indentation and else if but unblocking otherwise

packages/relay/src/lib/eth.ts
AlfredoG87: hmmm, yeah, it would be either SDKCLientError or PrecheckStatusError... so even if it works `as is` the `else if` would be better.

Nana-EC (approved):
LGTM

packages/relay/src/lib/eth.ts
```diff
@@ -881,7 +882,7 @@ export class EthImpl implements Eth {
           if (result?.entity.runtime_bytecode !== EthImpl.emptyHex) {
             const prohibitedOpcodes = ['CALLCODE', 'DELEGATECALL', 'SELFDESTRUCT', 'SUICIDE'];
             const opcodes = asm.disassemble(result?.entity.runtime_bytecode);
```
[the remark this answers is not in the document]
Nana-EC: @AlfredoG87 Please open a ticket so we don't miss this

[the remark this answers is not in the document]
AlfredoG87: https://github.com/hashgraph/hedera-json-rpc-relay/issues/1550

PR #1547, opened 2023-07-20
Cherry pick 1518: construct synthetic contract result for HTS

AlfredoG87:
**Description**:
Cherry pick 1518: construct synthetic contract result for HTS

**Related issue(s)**:
#1527


ebadiere (dismissed):

packages/relay/src/lib/eth.ts
```diff
@@ -1646,6 +1678,7 @@ export class EthImpl implements Eth {
     }
 
     await this.batchGetAndPopulateContractResults(contractResults, showDetails, transactionObjects, transactionHashes, requestIdPrefix);
+    await this.filterAndPopulateSyntheticContractResults(showDetails, logs, transactionObjects, transactionHashes, requestIdPrefix);
```
Nana-EC, line 1681: Let's feature gate this if we're cherry picking.
Default to false though

Nana-EC (changes requested):
Feature gate with default off

packages/relay/src/lib/eth.ts
AlfredoG87: should the feature gate be only on the `0.28` branch or should I include it in main too?

Also, do you think we should cherry pick it for `0.28` or is it better to wait for `0.29` to include it?

AlfredoG87: ✅  Done

Nana-EC (approved):
LG

Issue #1377, opened 2023-06-20
fix: unable to parse some contract transaction reverted, resulting in "TypeError: Cannot read properties of null (reading 'transactions')"


bguiz:
### Description

- Be able to parse the error message, instead of "unknown error"

### Steps to reproduce

1. Submit this transaction: https://hashscan.io/testnet/transaction/1687274355.056813003
2. Observe that transaction is reverted
3. But there is no error message on Hashscan
4. Observe in browser error: `Unknown error invoking RPC` - full message copied below
5. Observe RPC relay error: `TypeError: Cannot read properties of null (reading 'transactions')` - logs copied below

### Additional context

Error message in browser (from Metamask):

```text
MetaMask - RPC Error: [ethjs-query] while formatting outputs from RPC '{"value":{"code":-32603,"data":{"jsonrpc":"2.0","id":6696677332751,"error":{"code":-32603,"name":"Internal error","message":"[Request ID: 66de5563-d3e5-47d2-9d3d-444f557ace76] Unknown error invoking RPC"}}}}'
```

Relevant part of log output from RPC Relay:

```text
[2023-06-20 15:19:16.579 +0000] DEBUG (consensus-node/91353 on admins-MacBook-Pro.local): [Request ID: 66de5563-d3e5-47d2-9d3d-444f557ace76] 0.0.3996280@1687274343.363097580 eth_sendRawTransaction EthereumTransaction record status: CONTRACT_REVERT_EXECUTED (33), cost: 1.81207359 ℏ
[2023-06-20 15:19:17.045 +0000] TRACE (mirror-node/91353 on admins-MacBook-Pro.local): [Request ID: 66de5563-d3e5-47d2-9d3d-444f557ace76] Retry delay 250 ms
[2023-06-20 15:19:17.529 +0000] TRACE (mirror-node/91353 on admins-MacBook-Pro.local): [Request ID: 66de5563-d3e5-47d2-9d3d-444f557ace76] Retry delay 500 ms
[2023-06-20 15:19:18.110 +0000] INFO (consensus-node/91353 on admins-MacBook-Pro.local):  0.0.3996280@1687274346.358351484 eth_sendRawTransaction FileContentsQuery status: SUCCESS (22), cost: 36 tℏ
[...]

PR #1553, opened 2023-07-21
Fix cannot read properties of null reading 'transactions' (#1377)

quiet-node:
**Description**: 

**Related issue(s)**: #1377

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1151,7 +1151,13 @@ export class EthImpl implements Eth {
       txSubmitted = true;
       // Wait for the record from the execution.
       let txId = contractExecuteResponse.transactionId.toString();
-      const formattedId = formatTransactionId(txId);
+      const formattedId = formatTransactionIdWithoutQueryParams(txId);
+      
+      // handle formattedId beng null
+      if (!formattedId) {
+        throw predefined.INTERNAL_ERROR(`Invalid transactionID.`);
+      }
+
       const  record = await this.mirrorNodeClient.repeatedRequest(this.mirrorNodeClient.getContractResult.name, [formattedId], this.MirrorNodeGetContractResultRetries, requestIdPrefix);
       if (!record) {
         this.logger.warn(`${requestIdPrefix} No record retrieved`);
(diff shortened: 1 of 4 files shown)


Nana-EC (dismissed):
LG, just waiting on that acceptance test

packages/relay/src/lib/eth.ts
```diff
@@ -1161,12 +1161,19 @@ export class EthImpl implements Eth {
       txSubmitted = true;
       // Wait for the record from the execution.
       let txId = contractExecuteResponse.transactionId.toString();
-      const formattedId = formatTransactionId(txId);
+      const formattedId = formatTransactionIdWithoutQueryParams(txId);
+      
+      // handle formattedId beng null
+      if (!formattedId) {
+        throw predefined.INTERNAL_ERROR(`Invalid transactionID.`);
```
Nana-EC, line 1168: Let's add a log here and generalize the error so as to
```suggestion        
        throw predefined.INTERNAL_ERROR(`Invalid transactionID: ${txId}`);
```

Nana-EC (approved):
LG.
Added one suggestion

packages/relay/src/lib/eth.ts
quiet-node: Heard!

Nana-EC (dismissed):

packages/relay/src/lib/eth.ts
```diff
@@ -1161,12 +1161,19 @@ export class EthImpl implements Eth {
       txSubmitted = true;
       // Wait for the record from the execution.
       let txId = contractExecuteResponse.transactionId.toString();
-      const formattedId = formatTransactionId(txId);
+      const formattedId = formatTransactionIdWithoutQueryParams(txId);
+      
+      // handle formattedId beng null
```
AlfredoG87, line 1166: nit: typo `being?`

AlfredoG87 (approved):
LGTM 👍

packages/relay/src/lib/eth.ts
quiet-node: Good catch thanks for that 😅

AlfredoG87 (approved):
LGTM 👍

Issue #1501, opened 2023-07-11
Implement changes to the mirror-node API `accounts/{accountIdOrEvmAddress}`


georgi-l95:
### Problem

Currently the /api/v1/accounts/{accountIdOrEvmAddress} endpoint always returns the transactions of an account in addition to the account details. In many cases these are unnecessary and users are better served by calling /api/v1/transactions?account.id={accountId}. Still in some cases account transactions are valuable but when not applicable they increase the response size and the mirror node REST API work unnecessarily.

### Solution

Implement the changes introduced in the mirror-node with https://github.com/hashgraph/hedera-mirror-node/issues/6297

PR #1563, opened 2023-07-25
Call mirror node accounts api without requesting transactions list

Ivo-Yankov:
**Description**:
Implements a recent change in the mirror node accounts API, that enables to retrieve an account without a list of transactions. This should provide a performance increase since it reduces the DB queries made in the mirror node and thus reducing the time to retrieve account data.

Fixes: #1501


Nana-EC (approved):
Nice.
Haha 1 liner changes with such heavy test fallout due to all the API mocks.
At least there's good coverage 😄

PR #1582, opened 2023-07-28
Added typeof prefix in conditional.

### LINKED ISSUE #1581: Set ETH_CALL_DEFAULT_TO_CONSENSUS_NODE to evaluate as a string.
### Description

process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE === 'undefined' is evaluating as false when undefined.  It needs to be prefixed with typeof.

### Steps to reproduce

1. Run some tests against the relay w/out the ETH_CALL_DEFAULT_TO_CONSENSUS_NODE defined.  
2. See that it queries the consensus node.

### Additional context

_No response_

### Hedera network

other

### Version

v0.29.0

### Operating system

macOS

ebadiere:
Adds `typeof` in string property conditional check.

Fixes #1581 

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1220,7 +1220,7 @@ export class EthImpl implements Eth {
     
     try {
       // ETH_CALL_DEFAULT_TO_CONSENSUS_NODE = false enables the use of Mirror node
-      if ((typeof process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE === 'undefined') || (process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE == 'false')) {
+      if ((String(process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE) === 'undefined') || (process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE == 'false')) {
         //temporary workaround until precompiles are implemented in Mirror node evm module
         // Execute the call and get the response
         return await this.callMirrorNode(call, gas, value, requestIdPrefix);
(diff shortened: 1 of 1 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1220,7 +1220,7 @@ export class EthImpl implements Eth {
     
     try {
       // ETH_CALL_DEFAULT_TO_CONSENSUS_NODE = false enables the use of Mirror node
-      if ((process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE === 'undefined') || (process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE == 'false')) {
+      if ((String(process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE) === 'undefined') || (process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE == 'false')) {
```
lukelee-sl, line 1223: looks fine but can't we just test for undefined without converting to a string?

lukelee-sl (approved):
LG

packages/relay/src/lib/eth.ts
ebadiere: It's a serious gotcha.  The mocha tests pass, but when stepping through it with the debugger you can see that when process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE is undefined the condition is not honored, and the js === strict equality operator checking if (process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE) === 'undefined' will not resolve to true.  It needs to be cast to a string which is the same type.

Nana-EC: @ebadiere, @lukelee-sl's question was spot on. The fix here should have been to remove the string format of the `'undefined'` and make it just `undefined`.


The comparison would always fail because the item on the left is not a string and the item on the right is always a string.

We should revert this change as we've added an unnecessary conversion for every single `eth_call` instance

PR #1585, opened 2023-07-31
Design Document for Filter API

### LINKED ISSUE #1484: Add documentation for filter API
### Problem

As a developer I'd like to be able to use the filter API

### Solution

Add design documentation for filter API, what is, how it works. Can be influenced by the `eth_subscribe` doc.

### Alternatives

_No response_

georgi-l95:
**Description**:
Adds new design doc for introduction of Filter API, alongside couple of new methods - `eth_newFilter`, `eth_uninstallFilter`, `eth_getFilterChanges` and `eth_getFilterLogs`

**Related issue(s)**:

Fixes #1484 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/design/eth-filter.md
+++ b/docs/design/eth-filter.md
@@ -1,0 +1,133 @@
+# Filter API design
+
+## Purpose
+
+The JSON-RPC Relay currently doesn't support monitoring events using filter API as is common in many other relays and is expected by many web3 tools, which makes polling for event logs, new blocks and transactions hard for developers.
+
+## Goals
+
+1. Introduce new methods:
+   1. `eth_newFilter`
+   2. `eth_uninstallFilter`
+   3. `eth_getFilterChanges`
+   4. `eth_getFilterLogs`
+
+## Non-Goals
+
+1. Decrease the load on the subscription, as this is an alternative.
+2. Increase the supported functionality of the relay.
+
+## Architecture
+
+New filter API methods can be added and handled by extending `eth` class. Adding several new methods is needed and for saving the filter IDs it needs to utilize the already available cache.
+
+### Filter Types
+
+|      Filter Type       |                                                    Description                                                     |                           Support                           |
+| :--------------------: | :----------------------------------------------------------------------------------------------------------------: | :---------------------------------------------------------: |
+|          logs          |                               Returns all logs matching filter with given filter ID.                               |                       Yes, initially                        |
+|       newBlocks        |                            Returns all new blocks matching filter with given filter ID.                            |                 No, maybe in future release                 |
+| newPendingTransactions | Returns transaction hashes that are sent to the network and marked as `pending`, depending on the given filter ID. | No, as Hedera does not have pending transactions on a node. |
+
+### Initiating a filtration
+
+#### Request
+
+```javascript
+{"jsonrpc":"2.0","id": 1, "method": "eth_newFilter", "params": [fromBlock, toBlock, address, topics]}
+```
+
+#### Response
+
+```javascript
+{"jsonrpc":"2.0","id": 1, "result": FILTER_ID}
+```
+
+### Getting Filter Changes
+
+     The `getFilterLogs` method returns logs that match a specified topic filter and are included in newly added blocks.
+
+#### Parameters
+
+An existing
+
+#### Request
+
+```javascript
+{"jsonrpc":"2.0","id": 1, "method": "eth_getFilterLogs", "params": ["0x123"]}
+```
+
+#### Response
+
+```javascript
+{
+    "jsonrpc": "2.0",
+    "id": 73,
+    "result": [{
+        "address": "0xb5a5f22694352c15b00323844ad545abb2b11028",
+        "blockHash": "0x99e8663c7b6d8bba3c7627a17d774238eae3e793dee30008debb2699666657de",
+        "blockNumber": "0x5d12ab",
+        "data": "0x0000000000000000000000000000000000000000000000a247d7a2955b61d000",
+        "logIndex": "0x0",
+        "removed": false,
+        "topics": ["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef", "0x000000000000000000000000bdc0afe57b8e9468aa95396da2ab2063e595f37e", "0x0000000000000000000000007503e090dc2b64a88f034fb45e247cbd82b8741e"],
+        "transactionHash": "0xa74c2432c9cf7dbb875a385a2411fd8f13ca9ec12216864b1a1ead3c99de99cd",
+        "transactionIndex": "0x3"
+    }]
+}
+```
+
+## Error Codes
+
+| Error Codes |   Error message   |                              Solution                               |
+| :---------: | :---------------: | :-----------------------------------------------------------------: |
+|    32000    | Filter not found. | Occurs when user attempts receive logs using non-existing filterID. |
+
+## Rate Limits
+
+1. All filters expire after 5 minutes of inactivity (no queries).
+
+## Metric Capturing
+
+Capture metrics for the following:
+
+1. Log every call to all filter API method, as a total amount and broken down by IP.
+2. The duration of active filters.
+3. The total amount of requests to the mirror node per filter.
+4. The amount of requests to the mirror node per filter that have returned non-null data.
+
+## Tests
+
+The following test cases should be covered but additional tests would be welcome.
+
+1. Overall functionality of creating and uninstalling new filters.
+2. Receiving requested information depending on filterID using `eth_getFilterChanges` and `eth_getFilterLogs`.
+3. Deleting of filter due to inactivity.
+4. E2E test using popular libraries (`ethers.js WebSocketProvider`).
+
+## Non-Functional Requirements
+
+Users should be required to renew their filters in the case of an inactivity. If for some reason the relay is restarted it will lose all cached filters.
+
+## Deployment
+
+Filter API will run alongside the already available HTTP server.
+
+## Answered Questions
+
+1. How will we implement the `getFilterChanges` and `getFilterLogs` flow?
+2. How will the relay poll the mirror node?
+3. What kind of limits will be implemented?
+
+## Tasks (in suggested order):
+
+#### Milestone 1
+
+1. Finalize design document
+
+#### Milestone 2
+
+1. Introduce `еth_newFilter` and add acceptance and unit tests.
+2. Introduce `eth_uninstallFilter` and add acceptance and unit tests.
+3. Introduce `eth_getFilterChanges` and add acceptance and unit tests.
+4. Introduce `eth_getFilterLogs` and add acceptance and unit tests.


docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+{"jsonrpc":"2.0","id": 1, "result": FILTER_ID}
+```
+
+### Getting Filter Changes
+
+     The `getFilterLogs` method returns logs that match a specified topic filter and are included in newly added blocks.
+
+#### Parameters
+
+An existing
```
Ivo-Yankov, line 52:
```suggestion
An existing FILTER_ID
```

georgi-l95: Thanks, added.

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Architecture
+
+New filter API methods can be added and handled by extending `eth` class. Adding several new methods is needed and for saving the filter IDs it needs to utilize the already available cache.
+
+### Filter Types
+
+|      Filter Type       |                                                    Description                                                     |                           Support                           |
+| :--------------------: | :----------------------------------------------------------------------------------------------------------------: | :---------------------------------------------------------: |
+|          logs          |                               Returns all logs matching filter with given filter ID.                               |                       Yes, initially                        |
```
Nana-EC, line 28: Can you speak to the time frame of the logs. I believe it's all the logs since the last query to the filter correct?

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+|       newBlocks        |                            Returns all new blocks matching filter with given filter ID.                            |                 No, maybe in future release                 |
+| newPendingTransactions | Returns transaction hashes that are sent to the network and marked as `pending`, depending on the given filter ID. | No, as Hedera does not have pending transactions on a node. |
+
+### Initiating a filtration
+
+#### Request
+
+```javascript
+{"jsonrpc":"2.0","id": 1, "method": "eth_newFilter", "params": [fromBlock, toBlock, address, topics]}
```
Nana-EC, line 37: Q's
1. Can a `toBlock` be in the future?
2. If a `toBlock` is in the past is this call intended to give you all the logs in one go?
3. Are there limits to the amount of logs we should return in either case

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+#### Response
+
+```javascript
+{"jsonrpc":"2.0","id": 1, "result": FILTER_ID}
```
Nana-EC, line 43: Q: what format will the FILTER_ID take? Is it an integer, a hex string ...

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+        "transactionIndex": "0x3"
+    }]
+}
+```
+
+## Error Codes
+
+| Error Codes |   Error message   |                              Solution                               |
+| :---------: | :---------------: | :-----------------------------------------------------------------: |
+|    32000    | Filter not found. | Occurs when user attempts receive logs using non-existing filterID. |
```
Nana-EC, line 84: Q: 
1. are there any error codes around range?
2. What happens if during a request Mirror Node fails to give us logs, how do we intend to behave and respond?

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Rate Limits
+
+1. All filters expire after 5 minutes of inactivity (no queries).
```
Nana-EC, line 88: This should be configureable. Suggest a config label

Nana-EC, line 88: Q: What about filter count limits? Is there a max number of filters we should support? If so this should be configureable also.

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Metric Capturing
+
+Capture metrics for the following:
+
+1. Log every call to all filter API method, as a total amount and broken down by IP.
```
Nana-EC, line 94: Q's
1. What purpose is IP serving?
3. What about number of filters

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+2. The duration of active filters.
+3. The total amount of requests to the mirror node per filter.
+4. The amount of requests to the mirror node per filter that have returned non-null data.
+
+## Tests
+
+The following test cases should be covered but additional tests would be welcome.
+
+1. Overall functionality of creating and uninstalling new filters.
+2. Receiving requested information depending on filterID using `eth_getFilterChanges` and `eth_getFilterLogs`.
```
Nana-EC, line 104: Under this we'd want
1. Case where no logs applicable
2. Case where there are logs within range
3. Case where there's a logs out of range situation

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+3. Deleting of filter due to inactivity.
+4. E2E test using popular libraries (`ethers.js WebSocketProvider`).
+
+## Non-Functional Requirements
+
+Users should be required to renew their filters in the case of an inactivity. If for some reason the relay is restarted it will lose all cached filters.
```
Nana-EC, line 110: Q: How would this work?
Do they not just call `eth_newFilter` again?

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Deployment
+
+Filter API will run alongside the already available HTTP server.
+
+## Answered Questions
+
+1. How will we implement the `getFilterChanges` and `getFilterLogs` flow?
```
Nana-EC, line 118: I'm not sure 1 & 2 are fully answered.
As part of the design doc I'd like to get a better sense of what classes are involved, both existing and new ones.
These changes should be called out in summary

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
...
+## Architecture
+
+New filter API methods can be added and handled by extending `eth` class. Adding several new methods is needed and for saving the filter IDs it needs to utilize the already available cache.
+
+### Filter Types
+
+|      Filter Type       |                                                    Description                                                     |                           Support                           |
+| :--------------------: | :----------------------------------------------------------------------------------------------------------------: | :---------------------------------------------------------: |
+|          logs          |                               Returns all logs matching filter with given filter ID.                               |                       Yes, initially                        |
+|       newBlocks        |                            Returns all new blocks matching filter with given filter ID.                            |                 No, maybe in future release                 |
```
Nana-EC, line 29: Q: Do we know how often this is utilized in other RPCs?

Nana-EC (changes requested):
Great start.
Lots of questions and suggestions 😄

georgi-l95:
> Thank you @georgi-l95. Some general thoughts. I know this isn't a HIP and I'm not sure where exactly we draw the line between HIP and Design Doc, but I think it will be useful for anyone reviewing this document to have a bit more information:
> 
> 1. Several example user stories that explain _how_ the Filter API is used and _why_ it's useful. Code snippet examples will help clarify things as well.
> 2. Exactly what params each endpoint supports and what they signify. For example, where are the Filter Types used if the `eth_newFilter` method doesn't accept a `type` parameter?
> 3. What are the answers to the `Answered Questions` listed below?
> 4. If I understand correctly, once a user creates a Filter, that Filter is "living" on the Relay side and the user can reference it by its Filter ID. As @Ivo-Yankov pointed out in a separate channel, we should briefly describe the current mainnet Relay high-level architecture (how many Relay servers are operational, are they load-balanced, do they share a cache) and design the filters to be "shareable" between different Relay instances if possible, we should even try to be future-proof about this. Is this the first stateful feature we're implementing on the Relay side?

Thank you for the review.
1. I've added user stories to the HIP. Should I add them also here ? 
2. Added needed params for each method and how they affect each other.
3. Reworked some of the answered questions. Basically these questions are answered in the doc itself.
4. Yes, for us to be able to support this new functionally properly, we'd need to add shared cache across instances. I think eth_subscribe is our full stateful method.

docs/design/eth-filter.md
```diff
@@ -0,0 +1,133 @@
+# Filter API design
+
+## Purpose
+
+The JSON-RPC Relay currently doesn't support monitoring events using filter API as is common in many other relays and is expected by many web3 tools, which makes polling for event logs, new blocks and transactions hard for developers.
```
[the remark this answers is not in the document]
georgi-l95: I've added the ones that 100% use it.

docs/design/eth-filter.md
georgi-l95: Hm, we should follow `eth_getLogs` limitations. `eth_getFilterChanges` will give you all new logs since last poll and `eth_getFilterLogs` gives you all logs for the provided filters.

docs/design/eth-filter.md
georgi-l95: 1. Should be able to set in future.
2. Yes, if they don't exceed the predefined limit. (similar to the one we use in `eth_getLogs`)
3. I think 2 answer this.

docs/design/eth-filter.md
georgi-l95: Sorry, added, it should be hex.

docs/design/eth-filter.md
georgi-l95: 1. I've copied the error codes that other providers are using.
2. Good questions, we should either our defined rpc error ,or empty array(this is what other rpc providers are returning if there are no logs to return)

docs/design/eth-filter.md
georgi-l95: Added.

georgi-l95: Good questions, not sure if we should start with limit or wait and see how many people are using it and if there is performance impact.

docs/design/eth-filter.md
georgi-l95: Added.

docs/design/eth-filter.md
georgi-l95: Yes, they need to call eth_newFilter again with their filter params. Added to the doc.

docs/design/eth-filter.md
georgi-l95: Yes, you are right. I've reworked most of the doc.

docs/design/eth-filter.md
georgi-l95: Didn't find any information regarding this.

docs/design/eth-filter.md
Nana-EC: Start. Harder to add after and limit. So we should set a limit like we did for `eth_subscribe` and have errors for that.

docs/design/eth-filter.md
Nana-EC: I don't think you've answered the flow or interfaces much.
This isn't new logic i guess since it's similar to `eth_getLogs`

docs/design/eth-filter.md
AlfredoG87: Yeah I agree that we should remove unique values labels for metrics since they drastically increase cardinality and costs.

docs/design/eth-filter.md
```diff
@@ -0,0 +1,201 @@
...
+## Metric Capturing
+
+Capture metrics for the following:
+
+1. Log every call to all filter API method, as a total amount.
+2. The number of active filters.
+3. The duration of active filters.
+4. The total amount of requests to the mirror node per filter.
+5. The amount of requests to the mirror node per filter that have returned non-null data.
+
```
AlfredoG87, line 162: 6. Amount of filters that were uninstalled by Time Limit vs Explicitly by the user.

docs/design/eth-filter.md
```diff
@@ -0,0 +1,201 @@
...
+2. Receiving requested information depending on filterID using `eth_getFilterChanges` and `eth_getFilterLogs`.
+3. Deleting of filter due to inactivity.
+4. Case where logs are within limit range.
+5. Case where logs are not within limit range.
+6. Case where no logs are available.
+7. E2E test using popular libraries (`ethers.js WebSocketProvider`).
+
+## Non-Functional Requirements
+
+Users should be required to renew their filters, by calling again `eth_newFilter`, in the case of an inactivity. If for some reason the relay is restarted it will lose all cached filters.
```
AlfredoG87, line 177: should time limits restart every time a filter is used within `eth_getFilterChanges` or `eth_getFilterLogs`, since is sign of activity?

docs/design/eth-filter.md
```diff
@@ -0,0 +1,201 @@
...
+The following test cases should be covered but additional tests would be welcome.
+
+1. Overall functionality of creating and uninstalling new filters.
+2. Receiving requested information depending on filterID using `eth_getFilterChanges` and `eth_getFilterLogs`.
+3. Deleting of filter due to inactivity.
+4. Case where logs are within limit range.
+5. Case where logs are not within limit range.
+6. Case where no logs are available.
+7. E2E test using popular libraries (`ethers.js WebSocketProvider`).
+
```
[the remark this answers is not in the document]
georgi-l95: Yes, added those test cases, thank you.

docs/design/eth-filter.md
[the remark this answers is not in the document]
georgi-l95: Yes, you're right, in the case of distributed cache, in normal scenarios it will only be deleted by expired TTL.

docs/design/eth-filter.md
```diff
@@ -0,0 +1,201 @@
...
+## Non-Functional Requirements
+
+Users should be required to renew their filters, by calling again `eth_newFilter`, in the case of an inactivity. If for some reason the relay is restarted it will lose all cached filters.
+
+## Deployment
+
+Filter API will run alongside the already available HTTP server.
```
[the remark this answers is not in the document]
georgi-l95: Yep, will run on the same http server.

docs/design/eth-filter.md
georgi-l95: Thanks, added.

docs/design/eth-filter.md
georgi-l95: Yes, filter should be deleted only if it's not used in the last 5 minutes (for example) or whatever our configuration of this property is.

docs/design/eth-filter.md
georgi-l95: Gotcha, removed.

docs/design/eth-filter.md
georgi-l95: Okay, added.

Nana-EC (approved):
LG, updates should be made as implementation alters items

PR #1590, opened 2023-08-01
Implement `eth_uninstallFilter`

### LINKED ISSUE #1486: Implement `eth_uninstallFilter`
### Problem

As a developer I'd like to be able to call `eth_uninstallFilter` to uninstall already existing filter.

Uninstalls a filter with given id. Should always be called when watch is no longer needed. Additionally, Filters timeout when they aren’t requested with `eth_getFilterChanges` for a period of time.

### Solution

**Parameters**
-

- params (array of strings) - The filter id -> example ['0x1']

**Returns**
- 

- result (boolean) - Returns true if the filter was successfully uninstalled, otherwise false.

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR adds implementation for eth_uninstallFilter, unit and acceptance tests.
Also adds delete method to the ClientCache.

eth_uninstallFilter method is responsible for deleting existing filters from the cache. If filter exists and it's deleted it will return true to the user, else returns false.

**Related issue(s)**:

Fixes #1486 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/clients/clientCache.ts
```diff
@@ -101,6 +101,12 @@ export class ClientCache {
         this.cacheKeyGauge.labels('', ClientCache.setCacheLabel, callingMethod || '').inc(1);
     }
 
+    public delete(key: string, callingMethod: string, requestIdPrefix?: string): void {
+        this.logger.trace(`${requestIdPrefix} delete cache for ${key}`);
+        this.cache.delete(key);
+        this.cacheKeyGauge.labels('', ClientCache.setCacheLabel, callingMethod || '').inc(1);
```
Nana-EC, line 107: This isn't a set. You can add a new deleteCacheLabel or better yet remove the metric increment

packages/relay/tests/lib/clients/clientCache.spec.ts
```diff
@@ -87,6 +87,15 @@ describe('Simple ClientCache', async function () {
         const cacheValue = clientCache.get(key, callingMethod);
         expect(cacheValue).to.be.equal(expectedValue);
       });
+
+      it('delete a valid object', async function () {
+        const key = 'key';
+        const expectedValue = { key: 'value' };
+        clientCache.set(key, expectedValue, callingMethod);
+        clientCache.delete(key, callingMethod);
```
Nana-EC, line 95: Add a check before this line to make sure value is non null

packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -780,6 +780,24 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
             expect(nonceAfter).to.be.equal('0x3');
 
         });
+    });
+
+    describe('Filter API Test Suite', () => {
+        const nonExstingFilter = "0x111222331";
+
+        //Fix when newFilter method is added.
```
Nana-EC, line 788:
```suggestion
        // Fix when newFilter method is added.
```

Nana-EC (changes requested):
LG, some minor suggestions

packages/relay/src/lib/services/ethService/ethFilterService/index.ts
```diff
@@ -0,0 +1,54 @@
+import { Logger } from 'pino';
```
Nana-EC, line 1: nit: missing license header

packages/relay/src/lib/services/ethService/ethFilterService/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.cache = clientCache;
+  }
+
+  public async uninstallFilter(filterId: string, requestIdPrefix?: string | undefined): Promise<boolean> {
+    this.logger.trace(`${requestIdPrefix} uninstallFilter(${filterId})`);
+
+    const cacheKey = `${constants.CACHE_KEY.FILTERID}_${filterId}`;
+    const filter = this.cache.get(cacheKey, this.ethUninstallFilter, requestIdPrefix);
+    
+    if(filter) {
```
Nana-EC, line 47: nit:
```suggestion
    if (filter) {
```

packages/relay/src/lib/services/ethService/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./ethFilterService";
```
Nana-EC, line 1: nit: license

packages/relay/tests/lib/eth/filter.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+const registry = new Registry();
+
+let restMock: MockAdapter, web3Mock: MockAdapter;
+let mirrorNodeInstance: MirrorNodeClient;
+let filterService: FilterService;
+let clientCache: ClientCache;
+let mirrorNodeCache: ClientCache;
+
+
+describe('Filter API Test Suite', async function () {
```
Nana-EC, line 43: nit: something to think on and doesn't have to be fixed now but we should consider if we want the test structure to follow the src structure

Nana-EC (dismissed):
LG.
Unblocking, left some nits

Nana-EC (commented):
Also need to update the README under `/docs` for API and configurations

Nana-EC (approved):

PR #1595, opened 2023-08-02
Add eth_newFilter

### LINKED ISSUE #1485: Implement `eth_newFilter`
### Problem

As a developer I'd like to be able to add a new filter, which I can later uninstall or check for changes.

Creates a filter object, based on filter options, to notify when the state changes (logs). To check if the state has changed, call `eth_getFilterChanges`. 

### Solution

Add new endpoint for creating filters which returns the id of the newly created filter.

**Parameters**
-

A transaction response object which contains the following filter information:
- fromBlock (string) - The block number as a string in hexadecimal format or tags.
- toBlock (string) - The block number as a string in hexadecimal format or tags.
- address (string) - The contract address or a list of addresses from which logs should originate.
- topics (string) - An array of DATA topics and also, the topics are order-dependent.

**Returns**
- 

- result - It returns a filter id to be used when calling eth_getFilterChanges

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
- Adds a new method: `eth_newFilter`, which returns a `filterId` and stores in cache the filter parameters for later usage by other methods.
- Adds a feature flag for all filter API related methods: `FILTER_API_ENABLED`

**Related issue(s)**:

Fixes #1485  

**Checklist**


packages/relay/src/formatters.ts
```diff
@@ -19,11 +19,16 @@
  */
 
 import constants from "./lib/constants";
+import crypto from "crypto";
 
 const hashNumber = (num) => {
   return '0x' + num.toString(16);
 };
 
+const generateRandomHex = (bytes = 16) => {
```
Nana-EC, line 28: nit: 
```suggestion
const generateRandomHex = (bytesLength = 16) => {
```

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -237,4 +237,9 @@ export const predefined = {
     code: -39012,
     message: 'Unknown block'
   }),
+  'INVALID_BLOCK_RANGE': new JsonRpcError({
```
Nana-EC, line 240: If the range is too large or invalid. Might be good to add a custom message to distinguish when say toBlock is before fromBlock vs when the range is too large

packages/relay/src/index.ts
```diff
@@ -133,4 +133,6 @@ export interface Eth {
   syncing(requestId?: string): Promise<boolean>;
 
   accounts(requestId?: string): Array<any>;
+
+  newFilter(fromBlock: string, toBlock: string, address?: string, topics?: any[], requestIdPrefix?: string):  Promise<string | JsonRpcError>;
```
Nana-EC, line 137: Need to update the API README under `/docs`

Nana-EC (changes requested):
Looking good.
Can you refactor the layout to match the suggested reorg similar to https://github.com/hashgraph/hedera-json-rpc-relay/pull/1590
Left some additional suggestions

packages/relay/tests/lib/eth.spec.ts
```diff
@@ -4392,6 +4393,93 @@ describe('Eth calls using MirrorNode', async function () {
...
+        expect(cachedFilter.lastQueried).to.be.null;
+      });
+
+      it('validates fromBlock and toBlock', async function() {
+        // fromBlock is larger than toBlock
+        await RelayAssertions.assertRejection(predefined.INVALID_BLOCK_RANGE, ethImpl.newFilter, true, ethImpl, [blockNumberHexes[1500], blockNumberHexes[1400]]);
+        await RelayAssertions.assertRejection(predefined.INVALID_BLOCK_RANGE, ethImpl.newFilter, true, ethImpl, ['latest', blockNumberHexes[1400]]);
+
+        // block range is too large
+        await RelayAssertions.assertRejection(predefined.RANGE_TOO_LARGE(1000), ethImpl.newFilter, true, ethImpl, [blockNumberHexes[5], blockNumberHexes[2000]]);
```
Ivo-Yankov, line 4475: This tests the case where the block range is too large

packages/relay/src/lib/errors/JsonRpcError.ts
Ivo-Yankov: This case is already covered and there is a test for it: https://github.com/hashgraph/hedera-json-rpc-relay/pull/1595#discussion_r1284222480

Ivo-Yankov:
> Looking good. Can you refactor the layout to match the suggested reorg similar to #1590 Left some additional suggestions

I'll wait until #1590 is merged and then I'll refactor the layout

Ivo-Yankov:
I refactored the code to follow the new structure and created a new `CommonService` class that will be used for logic that will be shared between other services. Further improvements to the structure, like creating a `BaseService` class, or better separation of the `common` methods can be made in future PRs when we focus on the refactoring of the relay.

Nana-EC (approved):
LG, let's see if we can note the alpha state of the API

PR #1618, opened 2023-08-09
Add eth_newBlockFilter

### LINKED ISSUE #1488: Implement `eth_newBlockFilter`
### Problem

As a developer I'd like to be able to call `eth_newBlockFilter`, to create a filter in the node, to notify when a new block arrives.

### Solution

**Parameters**
-

This method does not accept any parameters

**Returns**
- 

- filter id - It returns a filter id to be used when calling eth_getFilterChanges

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
Adds a new method: `eth_newBlockFilter`, which is used to create a filter that returns the latest block hashes since the last calling of `eth_getFilterChanges`.

**Related issue(s)**:

Fixes #1488 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/services/ethService/ethFilterService/index.ts
```diff
@@ -118,6 +118,19 @@ export class FilterService implements IFilterService {
     }
   }
 
+  async newBlockFilter(requestIdPrefix?: string): Promise<string | JsonRpcError> {
+    this.logger.trace(`${requestIdPrefix} newBlockFilter()`);
+    try {
+      FilterService.requireFiltersEnabled();
+      return this.createFilter(constants.FILTER.TYPE.NEW_BLOCK, {
```
Nana-EC, line 125: Q: do handle if a matching filter already exists or would that end up extending the TTL of the existing filter?

Nana-EC (approved):
LG
One question that can be addressed in follow up

packages/relay/src/lib/services/ethService/ethFilterService/index.ts
Ivo-Yankov: Calling `eth_newFilter` or `eth_newBlockFilter` always creates a new filter with a unique `filterId`. Every filter has its own cached `lastQueried` value, which is going to be used as a `fromBlock` when `eth_getFilterChanges` is called for that filter. So when 2 filters are created with `eth_newBlockFilter`, they will return different results depending on the last time they were called.

The TTL is going to be extended on `eth_getFilterChanges` and `eth_filterLogs`

Nana-EC (approved):

AlfredoG87 (approved):
LGTM 👍

PR #1619, opened 2023-08-09
Add unsupported eth_newPendingTransactionFilter

### LINKED ISSUE #1490: Implement `eth_newPendingTransactionFilter`
### Problem

As a developer I'd like to be able to call eth_newPendingTransactionFilter and create a filter for new pending transactions.

⚠️ However due to the reason that we are fetching this information from the mirror-node and transactions there are already processed, we cannot implement this filter.

### Solution

**Parameters**
-

This method does not accept any parameters

**Returns**
- 

- NOT IMPLEMENTED

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
Adds the `eth_newPendingTransactionFilter` method which always returns an `UNSUPPORTED` json rpc error

**Related issue(s)**:

Fixes #1490 

**Notes for reviewer**:

**Checklist**


Nana-EC (approved):
LG

Issue #1475, opened 2023-07-08
[Design] Explore continuous automated RPC API schema verification


Nana-EC:
### Problem

The relay implements the Ethereum JSON RPC API 2.0 spec.
Occasionally changes are made externally or internally that may create differences and regressions in schema support.
The schema is clearly noted at execution-apis and can be found here https://github.com/ethereum/execution-apis

### Solution

Add a script, which scout the execution-apis repo for the tests (can explore https://github.com/octokit/rest.js). Create a list from the [folder names](https://github.com/ethereum/execution-apis/tree/main/tests) and go through each test, by sending a request to the relay using the schema described in the .io file. Add this to a workflow task, so that it should run at every pull request and report the result.

### Alternatives
Explore https://github.com/ethereum/hive/tree/master/simulators/ethereum/rpc-compat which does this, but for clients.

### Exploration
```[tasklist]
### Exploration tasks
- [x] eth_getBalance
- [x] eth_getBlockByHash
- [x] eth_getTransactionReceipt
- [x] eth_getTransactionByHash
- [x] eth_getTransactionByBlockHashAndIndex
- [x] eth_getTransactionByBlockNumberAndIndex
```

```[tasklist]
### Schema gap tasks
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/1614
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/1620
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/1647
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/1655
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/1656
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/1657
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/1654
```

PR #1621, opened 2023-08-09
Script executing execution api tests

konstantinabl:
**Description**:

The goal of this PR is to introduce a script which runs whenever a PR is created and tests if we conform to the JSON RPC Specification
**Related issue(s)**:

Fixes #
https://github.com/hashgraph/hedera-json-rpc-relay/issues/1475
**Notes for reviewer**:

**Checklist**


natanasow (approved):
LG, some good-to-have fixes and we're ready :rocket:

scripts/check-ethereum-api-schema.js
```diff
@@ -0,0 +1,156 @@
...
+        throw error;
+      }
+}
+
+function checkResponse(actualReponse, expectedResponse) {
+    const actualResponseKeys = extractKeys(actualReponse);
+    const expectedResponseKeys = extractKeys(expectedResponse);
+    const missingKeys = expectedResponseKeys.filter(key => !actualResponseKeys.includes(key));
+
+    const areEqual = actualResponseKeys.length === expectedResponseKeys.length && actualResponseKeys.every(item => expectedResponseKeys.includes(item));
```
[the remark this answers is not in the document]
konstantinabl: Yes, we do not need it

konstantinabl:
TBD: the github token we will use to query the Octokit API

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+  if (request.method === 'eth_getTransactionByBlockHashAndIndex') {
+    request.params[0] = legacyTransactionAndBlockHash.blockHash;
+    request.params[1] = legacyTransactionAndBlockHash.transactionIndex;
+  }
+  if (request.method === 'eth_getTransactionByBlockNumberAndIndex') {
+    request.params[0] = legacyTransactionAndBlockHash.blockNumber;
+    request.params[1] = legacyTransactionAndBlockHash.transactionIndex;
+  }
+  if (request.method === 'eth_sendRawTransaction') {
+    const transactionCount = await getTransactionCount();
```
natanasow, line 156: There are some redundant variables like this one.

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+};
```
natanasow, line 61: Maybe we should add contract deployments via eip 2930 and eip1559 as well?

natanasow (commented):
LG, some suggestions TBD

.github/workflows/acceptance-public.yml
```diff
@@ -47,6 +47,15 @@ jobs:
       operator_id: ${{ inputs.operator_id }}
       operator_key: ${{ inputs.operator_key }}
 
+  api_conformity:
```
Nana-EC, line 50:
```suggestion
  rpc_api_schema_conformity:
```
or
```suggestion
  rpc_api_schema_validation:
```
or
```suggestion
  rpc_api_schema_check:
```

packages/server/package.json
```diff
@@ -38,7 +38,8 @@
         "shelljs": "^0.8.5",
         "ts-mocha": "^9.0.2",
         "ts-node": "^10.8.1",
-        "typescript": "^4.5.5"
+        "typescript": "^4.5.5",
+        "execution-apis": "git://github.com/ethereum/execution-apis.git"
```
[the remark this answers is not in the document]
Nana-EC: Agreed. I think we want to be in control of the versions and take updates as they come along. If not it could break us unnecessarily

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -0,0 +1,271 @@
...
...82601f1061019557805160ff19168380011785556101c2565b828001600101855582156101c2579182015b828111156101c25782518255916020019190600101906101a7565b506101ce9291506101d2565b5090565b5b808211156101ce57600081556001016101d3565b634e487b7160e01b600052604160045260246000fd5b60006020828403121561020f57600080fd5b813567ffffffffffffffff8082111561022757600080fd5b818401915084601f83011261023b57600080fd5b81358181111561024d5761024d6101e7565b604051601f8201601f19908116603f01168101908382118183101715610275576102756101e7565b8160405282815287602084870101111561028e57600080fd5b826020860160208301376000928101602001929092525095945050505050565b600060208083528351808285015260005b818110156102db578581018301518582016040015282016102bf565b818111156102ed576000604083870101525b50601f01601f1916929092016040019392505050565b600181811c9082168061031757607f821691505b6020821081141561033857634e487b7160e01b600052602260045260246000fd5b5091905056fea2646970667358221220d450959bd13a5c79ab8546a400f6af65e1c3d24b6877b871e663861bbf17234664736f6c63430008090033000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000076e696b6f6c617900000000000000000000000000000000000000000000000000',
```
Nana-EC, line 60: Can we pick the bytecode up from the built contract instead of hard coding?
Just add a folder and the .sol and .json files under packages/server/tests/contracts

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+        keys = keys.concat(extractKeys(obj[key], newKey));
+      }
+    }
+  }
+
+  return keys;
+}
+
+function formatTransactionByHashAndReceiptRequests(fileName, request) {
+  switch (fileName) {
```
Nana-EC, line 202: WHat are these file name and where are they from?
Maybe add comment but also if possible make them constants

Nana-EC (changes requested):
Nice work.
Some suggestions

packages/server/tests/acceptance/conformityTests.spec.ts
konstantinabl: eslint does not show me this error, so could you please state all the redundant variables that your ide warns you for?

packages/server/tests/acceptance/conformityTests.spec.ts
konstantinabl: the tests here are replicated from the test in the execution apis repo
so i just execute them
we can create additional tests, but i think they should be a separate issue?

Nana-EC: Yeah additional test can be captured in a separate ticket.
It's unclear that we're replicating tests from the package.
Maybe some more comments would help

packages/server/tests/acceptance/conformityTests.spec.ts
konstantinabl: They are taken on the go, so cannot make the constants
This is the fileName of each file we process from the execution api tests

packages/server/tests/acceptance/conformityTests.spec.ts
konstantinabl: https://github.com/hashgraph/hedera-json-rpc-relay/issues/1730
Created an issue for this. Also added some comments to clarify what we are doing here

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -0,0 +1,295 @@
...
+
+async function checkRequestBody(fileName, request) {
+  /**
+   * Modifies a request object for compatability with our relay.
+   *
+   * @param {string} fileName - The name of the file associated with the request.
+   * @param {Object} request - The request object to be modified.
+   * @returns {Object} - The modified request object.
+   */
+  if (
```
Nana-EC, line 148: nit: The check values for `request.params[0]` seem random, can you assign then to a constant with an informative variable name so the test is more understandable.

Also might be better to group the `eth_sendRawTransaction` considerations

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -0,0 +1,295 @@
...
+    request = formatTransactionByHashAndReceiptRequests(fileName, request);
+  }
+  return request;
+}
+
+function checkResponseFormat(fileName, actualReponse, expectedResponse) {
+  const actualResponseKeys = extractKeys(actualReponse);
+  const expectedResponseKeys = extractKeys(expectedResponse);
+  const missingKeys = expectedResponseKeys.filter((key) => !actualResponseKeys.includes(key));
+  if ((fileName === 'get-dynamic-fee.io' || fileName === 'get-access-list.io') && missingKeys[0] === 'result.v') {
```
Nana-EC, line 184: nit: set file names as constants.

packages/server/tests/acceptance/conformityTests.spec.ts
Nana-EC: I'm not clear why the case values of the fileName can't be constants.
For instance `get-dynamic-fee.io` and `get-access-list.io` are used twice in this file.
It'd make it more manageable to have a section at the top of all the fileNames.

Nana-EC (dismissed):
Approving.
Left some non blocking comment which I would like addressed but can be done in a follow up separate PR if desired

packages/server/tests/acceptance/conformityTests.spec.ts
konstantinabl: Sorry, now I got what you meant :) Hope its okay now

Nana-EC (approved):

PR #1624, opened 2023-08-10
Implement eth_getFilterLogs

### LINKED ISSUE #1489: Implement `eth_getFilterLogs`
### Problem

As a developer I'd like to be able to call `eth_getFilterLogs` to get an array of all logs matching filter with given id.

ℹ️ Can compute the same results with an eth_getLogs call.

### Solution

⚠️This method only works for filters created with `eth_newFilter` not for filters created using `eth_newBlockFilter` or `eth_newPendingTransactionFilter`, which will return "filter not found".

**Parameters**
-

- params (array of strings) - The filter id that is returned from `eth_newFilter`, `eth_newBlockFilter` or `eth_newPendingTransactionFilter` -> example ['0x1']

**Returns**
- 

- address - An address from which this log originated
- topics - An array of zero to four 32 Bytes DATA of indexed log arguments.
- data - It contains one or more 32 Bytes non-indexed arguments of the log
- blockNumber - The block number where this log was in. null when its a pending log
- transactionHash - The hash of the transactions this log was created from. null when its a pending log
- transactionIndex - The integer of the transaction's index position that the log was created from. null when it's a pending log
- blockHash - The hash of the block where this log was in. null when its a pending log
- logIndex - The integer of the log index position in the block. null when it's a pending log
- removed - It is true when the log was removed due to a chain reorganization, and false if it's a valid log

### Alternatives

_No response_

natanasow:
### Problem

As a developer I'd like to be able to call `eth_getFilterLogs` to get an array of all logs matching filter with given id.

ℹ️ Can compute the same results with an eth_getLogs call.

### Solution

⚠️This method only works for filters created with `eth_newFilter` not for filters created using `eth_newBlockFilter` or `eth_newPendingTransactionFilter`, which will return "filter not found".

**Parameters**
-

- params (array of strings) - The filter id that is returned from `eth_newFilter`, `eth_newBlockFilter` or `eth_newPendingTransactionFilter` -> example ['0x1']

**Returns**
- 

- address - An address from which this log originated
- topics - An array of zero to four 32 Bytes DATA of indexed log arguments.
- data - It contains one or more 32 Bytes non-indexed arguments of the log
- blockNumber - The block number where this log was in. null when its a pending log
- transactionHash - The hash of the transactions this log was created from. null when its a pending log
- transactionIndex - The integer of the transaction's index position that the log was created from. null when it's a pending log
- blockHash - The hash of the block where this log was in. null when its a pending log
- logIndex - The integer of the log index position in the block. null when it's a pending log
- removed - It is true when the log was removed due to a chain reorganization, and false if it's a valid log

### Alternatives

_No response_

**Related issue(s)**:

Fixes #1489 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/services/ethService/ethCommonService/ICommonService.ts
```diff
@@ -31,4 +32,14 @@ export interface ICommonService {
   getLatestBlockNumber(requestIdPrefix?: string): Promise<string>;
 
   genericErrorHandler(error: any, logMessage?: string): JsonRpcError;
+
+  validateBlockHashAndAddTimestampToParams(params: any, blockHash: string, requestIdPrefix?: string): Promise<boolean>;
```
Nana-EC, line 36: Hmm, i'm not sure all of these need to be interface methods. If it's in here I'm guessing the goal is more than one Eth service will utilize it.
Notably in the long run eth.ts should be split across all the eth services.
We might see an abstract spin out and this interface reduce but this isn't a blocker right now as we juggle the two designs

Nana-EC (approved):
LG

AlfredoG87 (commented):
Question: does this would support synthetic (HTS) logs?

natanasow:
> Question: does this would support synthetic (HTS) logs?

Yes. Under the hood, `eth_getFilterLogs` is using the `eth_getLogs` logic. As long as synthetic HTS logs are supported on `eth_getLogs`, they will be included on `eth_getFilterLogs` as well.

Nana-EC (approved):

PR #1626, opened 2023-08-10
Add eth_getFilterChanges

### LINKED ISSUE #1487: Implement `eth_getFilterChanges`
### Problem

As a developer I'd like to be able to call `eth_getFilterChanges`, to get the changes depending on my filter.

Polling method for a filter, which returns an array of logs which occurred since last poll.

### Solution

**Parameters**
-

- params (array of strings) - The filter id that is returned from `eth_newFilter`, `eth_newBlockFilter` or `eth_newPendingTransactionFilter` -> example ['0x1']

**Returns**
- 

An array of one of the following, depending on the filter type, or empty if there were no changes since the last poll:

**`eth_newBlockFilter`**
- blockHash - The 32 byte hash of a block that meets your filter requirements

**`eth_newPendingTransactionFilter `**
- NOT GOING TO BE IMPLEMENTED

**`eth_newFilter `**
- address - An address from which this log originated
- topics - An array of zero to four 32 Bytes DATA of indexed log arguments.
- data - It contains one or more 32 Bytes non-indexed arguments of the log
- blockNumber - The block number where this log was in. null when its a pending log
- transactionHash - The hash of the transactions this log was created from. null when its a pending log
- transactionIndex - The integer of the transaction's index position that the log was created from. null when it's a pending log
- blockHash - The hash of the block where this log was in. null when its a pending log
- logIndex - The integer of the log index position in the block. null when it's a pending log
- removed - It is true when the log was removed due to a chain reorganization, and false if it's a valid log

### Alternatives

_No response_

Ivo-Yankov:
# Problem

As a developer I'd like to be able to call `eth_getFilterChanges`, to get the changes depending on my filter.

Polling method for a filter, which returns an array of logs which occurred since last poll.

# Solution

**Parameters**
-

- params (array of strings) - The filter id that is returned from `eth_newFilter`, `eth_newBlockFilter` or `eth_newPendingTransactionFilter` -> example ['0x1']

**Returns**
- 

An array of one of the following, depending on the filter type, or empty if there were no changes since the last poll:

**`eth_newBlockFilter`**
- blockHash - The 32 byte hash of a block that meets your filter requirements

**`eth_newPendingTransactionFilter `**
- NOT GOING TO BE IMPLEMENTED

**`eth_newFilter `**
- address - An address from which this log originated
- topics - An array of zero to four 32 Bytes DATA of indexed log arguments.
- data - It contains one or more 32 Bytes non-indexed arguments of the log
- blockNumber - The block number where this log was in. null when its a pending log
- transactionHash - The hash of the transactions this log was created from. null when its a pending log
- transactionIndex - The integer of the transaction's index position that the log was created from. null when it's a pending log
- blockHash - The hash of the block where this log was in. null when its a pending log
- logIndex - The integer of the log index position in the block. null when it's a pending log
- removed - It is true when the log was removed due to a chain reorganization, and false if it's a valid log

## Related issue(s):

Fixes #1487 

## Checklist


packages/relay/src/lib/services/ethService/ethFilterService/index.ts
```diff
@@ -137,4 +142,47 @@ export class FilterService implements IFilterService {
...
   }
+
+  public async getFilterChanges(filterId: string, requestIdPrefix?: string): Promise<string[] | Log[] | JsonRpcError> {
+    this.logger.trace(`${requestIdPrefix} getFilterChanges(${filterId})`);
+    FilterService.requireFiltersEnabled();
+
+    const cacheKey = `${constants.CACHE_KEY.FILTERID}_${filterId}`;
+    const filter = this.cache.get(cacheKey, this.ethGetFilterChanges, requestIdPrefix);
+
+    if (!filter || this.supportedTypes.indexOf(filter.type) === -1) {
```
AlfredoG87, line 153: > this.supportedTypes.indexOf(filter.type) === -1

should not this be validated when the filter is saved into the cache?

Ivo-Yankov: It's not possible to create a filter with an unsupported type, but I think it's good to have this check just in case

AlfredoG87: I see your point, but not sure to agree. if is really not needed it only makes the code harder to read (readability impact). and a very small unneeded computation cost.

I would leave this comment here open in case someone else has his opinion on this.

For me is just a Nit and would not matter too much if we have it or not.

docs/openrpc.json
```diff
@@ -689,6 +689,40 @@
             },
             "tags": [{ "name": "alpha API"}]
         },
+        {
+            "name": "eth_getFilterChanges",
+            "summary": "Gets the latest results since the last request for a provided filter according to its type",
```
Nana-EC, line 694: Can we add the alpha label to all the filter APIs for now on this file

packages/relay/src/lib/services/ethService/ethFilterService/index.ts
```diff
@@ -167,11 +172,56 @@ export class FilterService implements IFilterService {
...
+      ], undefined, {
+        order: 'asc'
+      });
+
+      latestBlockNumber = Number(
+        result?.blocks?.length ? result.blocks[result.blocks.length - 1].number : await this.common.getLatestBlockNumber(requestIdPrefix)
+      );
+
+      result = result?.blocks?.map(r => r.hash) || [];
+    }
```
Nana-EC, line 216: add an else that handles if filter.type is none of the supported. Also add test case to confirm

Nana-EC (changes requested):
Looking good.
Some suggestions and questions

docs/openrpc.json
Nana-EC: Thanks for adding it here.
If you can add it on all filter methods that would be great also

packages/relay/tests/lib/services/eth/filter.spec.ts
```diff
@@ -326,4 +344,103 @@ describe('Filter API Test Suite', async function () {
...
+    restMock.onGet(`contracts/results/logs?timestamp=gte:${customBlock.timestamp.from}&timestamp=lte:${customBlock.timestamp.to}&limit=100&order=asc`).reply(200, filteredLogs);
+    restMock.onGet('blocks/1').reply(200, { ...defaultBlock, block_number: 1 });
+
+    const filterId = await filterService.newFilter('0x1');
+    const logs = await filterService.getFilterChanges(filterId);
+    expect(logs).to.not.be.empty;
+    logs.every(log => expect(Number(log.blockNumber)).to.equal(9));
+  });
+
+  it('should return valid list of logs', async function() {
```
Nana-EC, line 437: nit: update test message as this differs from above and is for empty case

packages/relay/src/lib/services/ethService/ethFilterService/index.ts
Nana-EC: I don't see a UT case for this

Nana-EC (changes requested):
Almost there 2 test nits.
Also an API doc request if possible in this PR or follow up

docs/openrpc.json
natanasow: All of the other methods have that label, it was missing only here.

Nana-EC (dismissed):

Nana-EC (approved):

AlfredoG87 (approved):
LGTM 👍

PR #1630, opened 2023-08-10
Sub-Graph Example improvements after synthetic transactions support

### LINKED ISSUE #1628: Sub-Graph Example Enhancements
### Problem

While doing another round of tests after the inclusion of indexing for both `htsft` and `htsnft` tokens found some inconsistencies on the documentation and some minor issues that need fixing:

1. Link with query to graph-ql playground, is using htss instead of htsfts, that in turn shows up as error: 
```
"Type `Query` has no field `htss`"
```

2. Documentation on deploying contracts has missing steps:
  2.1. deployHTS
  2.2. deployHTSNFT
  2.3. mintNFTHTS
  2.4. transferHTSFT

3. When re-executing hardhat task `transferHTSFT` for 2nd time we get the following error: `TOKEN_ALREADY_ASSOCIATED_TO_ACCOUNT`, So the solution is to verify for association and then do the transfer.

4. Documentation is out of date with this paragraph:

```
Note: Currently this example needs to be executed against older relay (v0.10.0) and mirror-node (v0.67.0-rc1) versions, until all fixes have been released.

```

### Solution

1. Fix the Documentation link with the correct one.
2. Add the needed steps on the documentation
3. Fix the task `transferHTSFT` so it verifies for already associated token.
4. Remove the outdated text and add the relevant versions needed for the hedera-local-node and other pre-requisites.

### Alternatives

_No response_

AlfredoG87:
**Description**:
Changes to Documentation:
1. added steps for HTS FT and NFT Tokens
2. Removed outdated relay and mirror node versions and changed it for a recommended local-node version
3. Updated GraphQL query URL to fix issue

Improvements to Task:
1. Avoid error when running `transferHtsFT` task for second time, error of account already associated.
2. Changed initial mint amount from `10` to `1000` to allow for much more transfer interactions of HTS FT Token.

**Related issue(s)**:

Fixes #1628 

**Notes for reviewer**:

**Checklist**


AlfredoG87:
SonarCloud Analysis is failing on code duplication on a expected.json file used in tests. I think a exception is valid in this case.

tools/subgraph-example/scripts/hts-transfer.ts
```diff
@@ -35,17 +35,24 @@ export async function transferHtsFT(receiver: string, hre: any) {
...
-
+        await (
+            await associateTx.signWithSigner(wallet)
+        ).executeWithSigner(wallet)
+    ).getReceiptWithSigner(wallet);
+  } catch (error) {
+    if(error instanceof ReceiptStatusError && error.status._code === 194) {
+      console.log("Token already associated with account")
+    } else {
+      console.log(error);
```
Nana-EC, line 53: Should this throw or is not throwing and not doing any extra logic preferred?

tools/subgraph-example/test/subgraph.ts
```diff
@@ -40,35 +40,36 @@ describe("Subgraph", () => {
       const result = await getData(GRAVATAR_QUERY);
       const gravatars = result.data.gravatars;
 
-      expect(isEqual(gravatars, expected.gravatar.initial)).to.be.true;
+      expect(JSON.stringify(gravatars)).to.equal(JSON.stringify(expected.gravatar.initial));
```
Nana-EC, line 43: Q: why the change here and in the other cases?

Nana-EC (commented):
LG
2 Questions

tools/subgraph-example/test/subgraph.ts
AlfredoG87: I had a hard time figuring out what were the totality of the offended changes to the expected vs current.
Once I changed it to this approach I was able to debug much more faster the failing tests, then changed them all for consistency and future work on these tests are more maintanable.

tools/subgraph-example/scripts/hts-transfer.ts
AlfredoG87: Its almost the same.
1. Association fails for any other reason than `Already Associated` (code `194`) 
     a. but it was already associated, is preferable to continue, so the graph-example will work as expected.
     b. was not already associated, will fail on the next step when attempting to perform the transfer, probably with error, account is not associated.

tools/subgraph-example/test/subgraph.ts
Nana-EC: Ahh I see. I've encountered a similar issue in the past.
Good fix

Nana-EC (approved):
LG

georgi-l95 (approved):
Nice improvements. Should ignore duplication error IMO.

PR #1637, opened 2023-08-12
Fix Wrong Intrinsic gas cost in `eth_sendRawTransaction` precheck calculation

### LINKED ISSUE #1631: Wrong Intrinsic Gas Cost in `eth_sendRawTransaction` precheck
### Description

Wrong intrinsic gas is calculated by `eth_sendRawTransaction` precheck in all cases. For example deploying [this](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/tools/hardhat-example/contracts/Greeter.sol) greeter contract takes around 79-80k gas, but relay calculates that it needs at least 160k. This results in error throw in the relay even before the transaction reaches consensus node.

This bug didn't surface until now, because we've used hardcoded values for estimateGas, now that we have precise calculation of estimate gas from the mirror-node it appears that relay calculations were wrong.

### Steps to reproduce

1. Deploy contract using image from [this](https://github.com/hashgraph/hedera-mirror-node/pull/6633) PR, which now calculates correct estimateGas

### Additional context

_No response_

### Hedera network

other

### Version

main

### Operating system

None

georgi-l95:
**Description**:
Wrong intrinsic gas is calculated by eth_sendRawTransaction precheck in all cases. For example deploying [this](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/tools/hardhat-example/contracts/Greeter.sol) greeter contract takes around 79-80k gas, but relay calculates that it needs at least 160k. This results in error throw in the relay even before the transaction reaches consensus node.

This bug didn't surface until now, because we've used hardcoded values for estimateGas, now that we have precise calculation of estimate gas from the mirror-node it appears that relay calculations were wrong.
Calculation done according to Hedera [docs](https://docs.hedera.com/hedera/core-concepts/smart-contracts/gas-and-fees#gas-fees)

Removed estimateGas workaround for the mirror-node bug, as it's fixed there.

⚠️ Should merge and release this after https://github.com/hashgraph/hedera-mirror-node/pull/6633 is ready and deployed on public network.
**Related issue(s)**:

Fixes #1631 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/precheck.ts
+++ b/packages/relay/src/lib/precheck.ts
@@ -201,7 +201,7 @@ export class Precheck {
     const gasLimit = Number(tx.gasLimit);
     const failBaseLog = 'Failed gasLimit precheck for sendRawTransaction(transaction=%s).';
 
-    const intrinsicGasCost = Precheck.transactionIntrinsicGasCost(tx.data, tx.to!);
+    const intrinsicGasCost = Precheck.transactionIntrinsicGasCost(tx.data);
 
     if (gasLimit > constants.BLOCK_GAS_LIMIT) {
       this.logger.trace(
@@ -223,29 +223,26 @@ export class Precheck {
   }
 
   /**
-   * Calculates the intrinsic gas cost based on the number of empty bytes and whether the transaction is CONTRACT_CREATE
+   * Calculates the intrinsic gas cost based on the number of bytes in the data field
    * @param data
-   * @param to
    * @private
    */
-  private static transactionIntrinsicGasCost(data: string, to: string | undefined) {
-    const isCreate = to == undefined || to.length == 0;
+  private static transactionIntrinsicGasCost(data: string) {
+    let trimmedData = data.replace('0x', '');
 
     let zeros = 0;
-
-    const dataBytes = Buffer.from(data, 'hex');
-
-    for (const c of dataBytes) {
-      if (c == 0) {
+    let nonZeros = 0;
+    for (let index = 0; index < trimmedData.length; index += 2) {
+      const bytes = trimmedData[index] + trimmedData[index + 1];
+      if (bytes === '00') {
         zeros++;
+      } else {
+        nonZeros++;
       }
     }
 
-    const nonZeros = data.replace('0x', '').length - zeros;
-    const cost =
-      constants.TX_BASE_COST +
-      constants.TX_DATA_ZERO_COST * zeros +
-      constants.ISTANBUL_TX_DATA_NON_ZERO_COST * nonZeros;
-    return isCreate ? cost + constants.TX_CREATE_EXTRA : cost;
+    return (
+      constants.TX_BASE_COST + constants.TX_DATA_ZERO_COST * zeros + constants.ISTANBUL_TX_DATA_NON_ZERO_COST * nonZeros
+    );
   }
 }
--- a/packages/relay/tests/lib/precheck.spec.ts
+++ b/packages/relay/tests/lib/precheck.spec.ts
@@ -37,41 +37,39 @@ const logger = pino();
 const limitOrderPostFix = '?order=desc&limit=1';
 
 describe('Precheck', async function () {
+  const txWithMatchingChainId =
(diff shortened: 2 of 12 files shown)


packages/relay/src/lib/precheck.ts
```diff
@@ -186,26 +186,24 @@ export class Precheck {
...
+            zeros++;
+        } else {
+            nonZeros++;
+        }
     }
 
-    const nonZeros = data.replace('0x', '').length - zeros;
-    const cost = constants.TX_BASE_COST + constants.TX_DATA_ZERO_COST * zeros + constants.ISTANBUL_TX_DATA_NON_ZERO_COST * nonZeros;
-    return isCreate ? cost + constants.TX_CREATE_EXTRA : cost;
+    return constants.TX_BASE_COST + constants.TX_DATA_ZERO_COST * zeros + constants.ISTANBUL_TX_DATA_NON_ZERO_COST * nonZeros;
```
AlfredoG87, line 207: One question, what is the `IntrinsicGasCost`? and why is it calculated with this formula? 
I thought that accurate GasCost was possible only using contracts/call?

georgi-l95: This is like the minimum amiunt of gas needed to execute the transaction, based on the data provided and base cost for every transaction. The real and more accurate gas estimation is done by the web3 module. Here we are just making sure that the provided gasLimit is covering the bare minimum for this tx. Calucalation is done based on the info from Hedera docs, but basicaly for every empty byte we multiply by 4, and for every non empty we multiply by 16 and add all this to the base cost (21000).

AlfredoG87: Thanks for the explanation, it makes sense 👍

Nana-EC:
@georgi-l95 0.87.0 us in mainent now so you should be able to rebase

packages/relay/tests/lib/precheck.spec.ts
```diff
@@ -427,4 +427,37 @@ describe('Precheck', async function () {
...
+    it('should be able to able to calculate transfer', function () {
+      const intrinsicGasCost = Precheck.transactionIntrinsicGasCost(transfer);
+      expect(intrinsicGasCost).to.be.equal(constants.TX_BASE_COST);
+    });
+
+    it('should be able to calculate for odd length tx', function () {
+      const intrinsicGasCost = Precheck.transactionIntrinsicGasCost(invalidTx);
+      expect(intrinsicGasCost).to.be.greaterThan(constants.TX_BASE_COST);
+    });
+  });
```
Nana-EC, line 462: Can you add some more fine tuned examples that confirm exact numbers for a a small and medium set of contract bytes e.g.
- empty contract (smallest permissable)
- simple contract with a singleslot that is set on constructor
- you can also use existing test contracts in the repo

Doesn't have to be exhaustive but 1 or 2 to know the number calculation work for the intrinsic gas calculation on a create and not on create scenario

Nana-EC (changes requested):
Nice, some suggestions

packages/relay/src/lib/precheck.ts
```diff
@@ -223,29 +223,26 @@ export class Precheck {
...
       }
     }
 
-    const nonZeros = data.replace('0x', '').length - zeros;
-    const cost =
-      constants.TX_BASE_COST +
-      constants.TX_DATA_ZERO_COST * zeros +
-      constants.ISTANBUL_TX_DATA_NON_ZERO_COST * nonZeros;
-    return isCreate ? cost + constants.TX_CREATE_EXTRA : cost;
+    return (
```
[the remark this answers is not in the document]
Nana-EC: Synced offline. Adding here would make us inconsistent with the mirror node

packages/relay/tests/lib/precheck.spec.ts
Nana-EC: @georgi-l95 can address this in follow up

Nana-EC (approved):
LG
Minor comments and tests requested. Please address in follow up PR

PR #1640, opened 2023-08-15
Fix eth_getBalance to accept hash

### LINKED ISSUE #1614: eth_getBalance method does not conform to Ethereum JSON-RPC schema
### Description

While working on https://github.com/hashgraph/hedera-json-rpc-relay/issues/1475 I found that our eth_getBalance doesn't conform to the Ethereum JSON-RPC Specification - https://ethereum.github.io/execution-apis/api-documentation/ . 
In the docs you can see that eth_getBalance should accept a block hash, but in our case it only accepts blockNumber/Tag. Therefore, if we want to match the Ethereum sche we need to add the possibility for blockHash.

### Steps to reproduce

1. Run the relay
2. Send request for eth_getBalance with second parameter a blockHash

### Additional context

_No response_

### Hedera network

other

### Version

0.29.0

### Operating system

None

konstantinabl:
**Description**:

This PR modifies the validations used for eth_getBalance endpoint as well as the getBalance method itself in order to allow the user to get balance by block hash.

- Add possibility for multiple validations with | e.g not only blockNumber or blockHash, but both at the same time
- Add change so blockDiff can be calculated correctly even if blockHash is passed to getBalance
- Add unit and acceptance tests
- Fix tests after the change

**Related issue(s)**:

Fixes #1614 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -734,14 +734,14 @@ export class EthImpl implements Eth {
    * @param account
    * @param blockNumberOrTag
    */
-  async getBalance(account: string, blockNumberOrTag: string | null, requestIdPrefix?: string) {
+  async getBalance(account: string, blockNumberOrTagOrHash: string | null, requestIdPrefix?: string) {
     const latestBlockTolerance = 1;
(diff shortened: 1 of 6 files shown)


Nana-EC (changes requested):
LG. Question on if we need a test for when the block is not latest.

Nana-EC (approved):

PR #1642, opened 2023-08-15
Add Distributed Cache Design Document

### LINKED ISSUE #1641: Add design documentation for distributed cache
### Problem

Relay does not seem to support distributed cache, which will help functionality like filter API to be fully utilized.

### Solution

Add design documentation for distributed cache

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR adds design documentation around implementation of distributed cache using Redis. Explains architecture details using diagram, as well as answering questions around what are the goals and more.

**Related issue(s)**:

Fixes #1641 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/design/distributed-cache.md
+++ b/docs/design/distributed-cache.md
@@ -1,0 +1,110 @@
+# Distributed cache design
+
+## Purpose
+
+The purpose of a distributed cache is to improve the performance and scalability of JSON-RPC Relay by storing frequently accessed data in memory, making it readily available for quick retrieval by all running instances. This will help functionality like Filter API to work properly as it's heavily dependent on cache.
+
+## Goals
+
+1. **Scalability**: Distributed caches can be used to distribute the load across multiple instances of an application. As the application scales, the cache can help maintain consistent performance by providing a shared data store that all instances can access.
+2. **Improved Performance**: By sharing cache accross multiple JSON-RPC Relay instances, correct data will always be available for methods to access upon requests.
+3. **Reduced Load on Mirror-Node**: By caching data and serving it directly from the cache, the load on the underlying data sources (such as Mirror-Node) is reduced. This can help prevent performance bottlenecks and overloading of these resources.
+
+## Architecture
+
+For distributed cache JSON-RPC Relay will be using Redis and in order to establish connection and manage, it needs to utilize `node-redis` package. All cache managment will be done in the respective cache client, either for the internal `lru-cache` or shared redis, which methods will be defined by an interface. Using dependency injection pattern with dependency inversion all services that require cache will receive only abstraction in the constructor. This will allow for flexible usage of the cache without additional coupling, depending on the service need.
+Important details is that, if an operator does not want to use Redis, it does not need to, because all services that use shared cache will automatically switch to using internal `lru-cache`.
+
+### Diagram
+
+```mermaid
+flowchart TD
+    subgraph Relay Instance 1
+        subgraph Services Instance 1
+        A{{Filter Service}}
+        B{{Websocket Service}}
+        C{{History Service}}
+        end
+            subgraph Internal Cache Instance 1
+            D{{lru-cache}}
+            end
+    end
+    subgraph Distributed Cache
+    I{{Redis}}
+    end
+    subgraph Relay Instance 2
+        subgraph Services Instance 2
+        E{{Filter Service}}
+        F{{Websocket Service}}
+        G{{History Service}}
+        end
+            subgraph Internal Cache Instance 2
+            H{{lru-cache}}
+            end
+    end
+    A == Redis is enabled ==> I
+    E == Redis is enabled ==> I
+    A == Redis is disabled ==> D
+    E == Redis is disabled ==> H
+    B ==> D
+    C ==> D
+    F ==> H
+    G ==> H
+```
+
+### Interface
+
+```javascript
+interface ICacheClient {
+  get(key: string, callingMethod: string, requestIdPrefix?: string): any;
+  set(key: string, value: any, callingMethod: string, ttl?: number, requestIdPrefix?: string): void;
+  delete(key: string, callingMethod: string, requestIdPrefix?: string): void;
+  purgeStale(): void;
+  clear(): void;
+}
+```
+
+## Limits
+
+1. Should distributed cache be enabled. Env. variable `REDIS_ENABLED`.
+
+## Metric Capturing
+
+Capture metrics for the following:
+
+1. Log every service what kind of cache it's using.
+2. For each service log calls to `get`, `set`, `delete`.
+
+## Tests
+
+The following test cases should be covered but additional tests would be welcome.
+
+1. Connecting to the redis server is sucesfull.
+2. Fallback to `lru-cache` if redis is not available or respective feature flags are set.
+3. Cases where we set a cache.
+4. Cases where we get a cache.
+5. Cases where we try to get expired or deleted cache.
+6. Cases where we delete a cache.
+7. Cases where we try to delete non existing cache.
+
+## Deployment
+
+The Redis cache will need to run in a separate docker container alongside JSON-RPC Relay instances. A feature flag should be set and url specify in order for the relay instances to make connection with the shared cache.
+Environment variable are needed for enabling distributed cache - `REDIS_ENABLED` and for specifying cache address - `REDIS_URL`.
+
+## Answered Questions
+
+1. What is the purpose of adding distributed cache ?
+2. How will Redis be implemented to coexist with the current `lru-cache` ?
+3. What is the whole workflow of cache usage ?
+
+## Tasks (in suggested order):
+
+#### Milestone 1
+
+1. Finalize design document
+
+#### Milestone 2
+
+1. Implement cacheClient class refactor to utilize interface and `node-redis` package.
+2. Add new env. variable to the corresponding files, like env. example, helm charts and more.


AlfredoG87 (commented):
Overall looks good left some minor suggestions and questions

Nana-EC:
Smh at the DApp Test failing on a doc update

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,110 @@
...
+## Purpose
+
+The purpose of a distributed cache is to improve the performance and scalability of JSON-RPC Relay by storing frequently accessed data in memory, making it readily available for quick retrieval by all running instances. This will help functionality like Filter API to work properly as it's heavily dependent on cache.
+
+## Goals
+
+1. **Scalability**: Distributed caches can be used to distribute the load across multiple instances of an application. As the application scales, the cache can help maintain consistent performance by providing a shared data store that all instances can access.
+2. **Improved Performance**: By sharing cache accross multiple JSON-RPC Relay instances, correct data will always be available for methods to access upon requests.
+3. **Reduced Load on Mirror-Node**: By caching data and serving it directly from the cache, the load on the underlying data sources (such as Mirror-Node) is reduced. This can help prevent performance bottlenecks and overloading of these resources.
+
```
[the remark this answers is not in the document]
Nana-EC: I think it's hinted in #2 . We can update it with reference to eth filter and eth subscribe calls.

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,110 @@
...
+## Architecture
+
+For distributed cache JSON-RPC Relay will be using Redis and in order to establish connection and manage, it needs to utilize `node-redis` package. All cache managment will be done in the respective cache client, either for the internal `lru-cache` or shared redis, which methods will be defined by an interface. Using dependency injection pattern with dependency inversion all services that require cache will receive only abstraction in the constructor. This will allow for flexible usage of the cache without additional coupling, depending on the service need.
```
Nana-EC, line 15: This part is not clear to me. It's hints to an either approach whereas I think an and approach is better,
Services should be able to use both a local and a shared cache at the same time. This would avoid needing to store some items in a shared node when it's really only useful locally. 
A relay instance should be able to continue to operate partially if access to a redis cache went down. So only shared resources should go to redis

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,110 @@
...
+Important details is that, if an operator does not want to use Redis, it does not need to, because all services that use shared cache will automatically switch to using internal `lru-cache`.
+
+### Diagram
+
+```mermaid
+flowchart TD
```
Nana-EC, line 21: This confirms my concern above. 
However, the diagram is appropriate for features that might utilize the shared cache. However, this shouldn't be the case for all.

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+### Interface
+
+```javascript
+interface ICacheClient {
+  get(key: string, callingMethod: string, requestIdPrefix?: string): any;
+  set(key: string, value: any, callingMethod: string, ttl?: number, requestIdPrefix?: string): void;
+  delete(key: string, callingMethod: string, requestIdPrefix?: string): void;
+  purgeStale(): void;
```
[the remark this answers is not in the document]
Nana-EC: Agreed, this might be a method for the LRUCache unless it's also exposed via Redis API

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,110 @@
...
+Capture metrics for the following:
+
+1. Log every service what kind of cache it's using.
+2. For each service log calls to `get`, `set`, `delete`.
+
+## Tests
+
+The following test cases should be covered but additional tests would be welcome.
+
+1. Connecting to the redis server is sucesfull.
```
Nana-EC, line 82:
```suggestion
1. Connecting to the redis server is successful.
```

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,110 @@
...
+2. Fallback to `lru-cache` if redis is not available or respective feature flags are set.
+3. Cases where we set a cache.
+4. Cases where we get a cache.
+5. Cases where we try to get expired or deleted cache.
+6. Cases where we delete a cache.
+7. Cases where we try to delete non existing cache.
```
Nana-EC, line 88: Would need to be an acceptance test with 
- a call to relay instance 1 which sets a cache value
- a future query to instance 2 which relies on the previous cache 
- assertion that instance 2 provides appropriate answer since it accessed the cache

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,110 @@
...
+2. How will Redis be implemented to coexist with the current `lru-cache` ?
+3. What is the whole workflow of cache usage ?
+
+## Tasks (in suggested order):
+
+#### Milestone 1
+
+1. Finalize design document
+
+#### Milestone 2
```
Nana-EC, line 107: Add a Milestone 3 which reviews existing LRU cache usage and explores which should be move to shared cache

Nana-EC, line 107: Update Milestone 2 to not it will focus on using the shared cache only for eth filter and potentially eth subscribe

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,110 @@
...
+
+```javascript
+interface ICacheClient {
+  get(key: string, callingMethod: string, requestIdPrefix?: string): any;
+  set(key: string, value: any, callingMethod: string, ttl?: number, requestIdPrefix?: string): void;
+  delete(key: string, callingMethod: string, requestIdPrefix?: string): void;
+  purgeStale(): void;
+  clear(): void;
+}
+```
```
Nana-EC, line 65: There's no mention of the `LocalLRUCache` and `RedisCache` classes that will need to be added that separate the usage.

Nana-EC (changes requested):
Nice start, some suggestions and requests

AlfredoG87:
There should be a metric label that indicates the type of usage of the cache, that could be either `cache` or  `state-storage`.
Or 
May also be solved by using the current `CACHE_KEY` (that can also be used as cache key label) having a convention for example: `FILTERID = 'filterId'` --> `FILTERID = 'state/filterId'`  and used as the suggested `keyLabel` metric label.

The purpose is to be able to differentiate in the metrics what usage is purely for caching and for state sharing features.

docs/design/distributed-cache.md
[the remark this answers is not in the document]
georgi-l95: You're right, this is lru specific, should not be in the interface.

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Limits
+
+1. Should distributed cache be enabled. Env. variable `REDIS_ENABLED`.
+
+## Metric Capturing
+
+Capture metrics for the following:
+
+1. Log every service what kind of cache it's using.
```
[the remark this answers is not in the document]
georgi-l95: Thanks for this suggestion, added.

docs/design/distributed-cache.md
georgi-l95: Updated the logic.

docs/design/distributed-cache.md
georgi-l95: Updated the logic and added additional diagram

docs/design/distributed-cache.md
georgi-l95: Fixed.

docs/design/distributed-cache.md
georgi-l95: Added.

docs/design/distributed-cache.md
georgi-l95: Added.

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,110 @@
...
+2. For each service log calls to `get`, `set`, `delete`.
+
```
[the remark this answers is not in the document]
AlfredoG87: This is related to the comment at the end, I thought that the `cacheType` only tells us if is lru(local) or redis (distributed). 

But I think it would be useful to know when the cache is being used, for state storage or caching purposes.
The rationale being that caching would not disrupt the features if gets deleted/cleared, while the state storage function would (filterId or synthetic transactions). 

So not only would be a nice tracking label, but even if the key could have a prefix would make easier to delete/clear any or all caches, without affecting features that need to share state between pods.

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,247 @@
...
+     * @private
+     */
+    private readonly sharedCache;
+
+    public constructor() {
+        const internalCache = new LocalLRUCache();
+        const sharedCache = new RedisCache();
+    }
+
+    get(key: string, callingMethod: string, shared: boolean = false, requestIdPrefix?: string) {
```
Nana-EC, line 123: nit: optional param should be last

docs/design/distributed-cache.md
```diff
@@ -0,0 +1,247 @@
...
+        // In case of error does NOT fallback to shared cache.
+        // Clearing from shared cache depends on REDIS_ENABLED env. variable
+    }
+}
+```
+
+### Flowchart
+
+```mermaid
+flowchart TD
```
Nana-EC, line 151: Modification here is that the `Cache Client Service` sits inside each relay instance. Based on the need for share the call will either be serviced in instance by the `LocalLRUCache` or it will go out of the instance to the `RedisCache`.
Just clarifying as the digram may mislead to think the LocalLRUCache is somehow not also in the relay instance.

Nana-EC (commented):
Looking good.
One nit and then digram update and I'm good.
Otherwise, I'm agreed on approach

docs/design/distributed-cache.md
georgi-l95: Ah, yes, thank you. Fixed it.

docs/design/distributed-cache.md
georgi-l95: You're right, fixed.

Nana-EC (approved):
LG.
We can keep this up to date later as implementation requires changes

AlfredoG87 (approved):
Nice  👍

PR #1649, opened 2023-08-18
Refactor Remove AcceptedErrorCodes from MirrorNodeClient - GET_ACCOUNTS_BY_ID_ENDPOINT

AlfredoG87:
**Description**:
Refactor methods that use mirror node endpoint `GET_ACCOUNTS_BY_ID_ENDPOINT`: **getAccount, getAccountLatestEthereumTransactionsByTimestamp, getAccountPageLimit** to stop accepting error codes 400 and 404, but instead to throw the error and moved the handling of the error to be specific of those methods and not as a general handler for all errors on the mirror node requests

**Related issue(s)**

Addresses #1276 

**Notes for reviewer**:
This replaces this other [PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1366) 

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -91,7 +91,7 @@ export class MirrorNodeClient {
     private readonly MIRROR_NODE_RETRY_DELAY = parseInt(process.env.MIRROR_NODE_RETRY_DELAY || '250');
 
     static acceptedErrorStatusesResponsePerRequestPathMap: Map<string, Array<number>> = new Map([
-        [MirrorNodeClient.GET_ACCOUNTS_BY_ID_ENDPOINT, [400, 404]],
+        [MirrorNodeClient.GET_ACCOUNTS_BY_ID_ENDPOINT, []],
         [MirrorNodeClient.GET_BALANCE_ENDPOINT, [400, 404]],
         [MirrorNodeClient.GET_BLOCK_ENDPOINT, [400, 404]],
         [MirrorNodeClient.GET_BLOCKS_ENDPOINT, [400, 404]],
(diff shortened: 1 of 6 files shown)


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -388,30 +388,66 @@ export class MirrorNodeClient {
...
             requestIdPrefix);
     }
 
-    public async getAccountLatestEthereumTransactionsByTimestamp(idOrAliasOrEvmAddress: string, timestampTo: string, numberOfTransactions: number = 1, requestIdPrefix?: string) {
+    /**
+     * Returns account or null if account does not exist
+     * @param {string} idOrAliasOrEvmAddress commonly an evm address
+     * @param {string} requestIdPrefix optional request id prefix
+     */
+    public async getAccountOrNull(idOrAliasOrEvmAddress: string, requestIdPrefix?: string) {
```
Nana-EC, line 417: Q: why the `OrNull` method overload?
Should services calling this handle the null?

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -388,30 +388,66 @@ export class MirrorNodeClient {
...
+        } catch (error: any) {
+            return this.handleAccountNotFound(error, idOrAliasOrEvmAddress, requestIdPrefix);
+        }
     }
 
-    public async getAccountPageLimit(idOrAliasOrEvmAddress: string, requestIdPrefix?: string) {
-        return this.get(`${MirrorNodeClient.GET_ACCOUNTS_BY_ID_ENDPOINT}${idOrAliasOrEvmAddress}?limit=${constants.MIRROR_NODE_QUERY_LIMIT}`,
-            MirrorNodeClient.GET_ACCOUNTS_BY_ID_ENDPOINT,
-            requestIdPrefix);
+    public async getAccountPageLimitOrNull(idOrAliasOrEvmAddress: string, requestIdPrefix?: string) {
```
Nana-EC, line 443: Q: why the `OrNull` method overload?
Should services calling this handle the null?

Nana-EC (changes requested):
Looking good.
Not sure why we need the overloads vs just adopting the logic and allowing callers to manage the null

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: That was my original approach, but then realized that this methods are used by many different services (eth.ts, precheck.ts and relay.ts) and that I had to handle basically the same logic in all 3 services, so I needed to share the `handleAccountNotFound` logic in all 3 classes, so Instead of copy pasting (or creating a new helper class for the mirrorNodeClient), and then  modify all the reference where the logic already expects a `null` to add a try catch block, did not seem like progress or improvement to the code.

So I came up with this other approach, where the method name keeps working as expected returning null, but with a more explicit intention due to the `OrNull` overload and only handling `404` errors, not `400` or any other exception. and I get that this is eerly similar to the current `acceptedErrorStatusCodes` collection approach, but on this case the use of that error handler is intentional and specific to these methods since in these logic flows getting a `404` most be handled and is already handled by expecting the result to be null. (this has the advantage of re-using current logic flow where null are expected, not needing to add redundant try catch blocks every reference of the method, and re-utilizing the same error handler code logic for NotFound accounts)

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: Same as above

AlfredoG87:
Closed in favor of https://github.com/hashgraph/hedera-json-rpc-relay/pull/1651

PR #1651, opened 2023-08-18
Removed 400 AcceptedErrorCodes from MirrorNodeClient for GET_ACCOUNTS_BY_ID_ENDPOINT path

AlfredoG87:
**Description**:
Removed 400 from acceptedErrorCodes for GET_ACCOUNTS_BY_ID_ENDPOINT path, added UT to verify that errors are being thrown for both 400 and 500 error status codes

**Related issue(s)**:  #1276 

Fixes #

**Notes for reviewer**:
This replaces [PR 1649](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1649)
And [PR 1366](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1366)

**Checklist**


Nana-EC (approved):
LG
If we don't already have on there should be a UT to check that on 404 for those account methods that the method doesn't through an error but returns null

AlfredoG87:
> LG If we don't already have on there should be a UT to check that on 404 for those account methods that the method doesn't through an error but returns null

Yes there are already some UTs that cover those scenarios

PR #1658, opened 2023-08-21
Add ping messages to websocket server

Ivo-Yankov:
**Description**:

Adds a pinging mechanism to the websocket server. The interval timing can be configured by a new env var `WS_PING_INTERVAL` (default is 1000ms). The message is a simple JSON string with no `result` and `id`:
```
{"jsonrpc": "2.0"}
```

A known issue is that clients using `ethers@^5.0` receive a warning message: `this should not happen`. This is resolved with `ethers@6.0`

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/ws-server/src/webSocketServer.ts
+++ b/packages/ws-server/src/webSocketServer.ts
@@ -46,6 +46,8 @@ const mainLogger = pino({
     }
 });
 
+const pingInterval = Number(process.env.WS_PING_INTERVAL || 1000);
+
 const logger = mainLogger.child({ name: 'rpc-ws-server' });
 const register = new Registry();
 const relay: Relay = new RelayImpl(logger, register);
@@ -213,6 +215,10 @@ app.ws.use(async (ctx) => {
 
         ctx.websocket.send(JSON.stringify(response));
     });
+
+    setInterval(async () => {
+        ctx.websocket.send(JSON.stringify(jsonResp(null, null, null)));
+    }, pingInterval);
 });
 
 const httpApp = (new KoaJsonRpc(logger, register)).getKoaApp();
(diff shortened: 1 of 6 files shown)


AlfredoG87 (dismissed):
LGTM 👍

AlfredoG87 (approved):
LGTM 👍

packages/ws-server/src/webSocketServer.ts
```diff
@@ -213,6 +215,10 @@ app.ws.use(async (ctx) => {
 
         ctx.websocket.send(JSON.stringify(response));
     });
+
```
AlfredoG87, line 218: should we make it a configurable property, if `pingInterval > 0`. ? 
Supposed we fixed it from the infrastructure side, or a different deployment setup does not need it.

AlfredoG87 (changes requested):
looks like `charts` test was broken.

packages/ws-server/src/webSocketServer.ts
```diff
@@ -213,6 +215,10 @@ app.ws.use(async (ctx) => {
...
+    setInterval(async () => {
```
Nana-EC, line 219: Q: How does affect subscriptions with messages?
Do they receive empty messages not conforming with log formats?

Nana-EC, line 219: Q: Does this extend the connection time of some subscribers. Will this keep their session alive vs closing after a while of no message activity?

Nana-EC (changes requested):
Nice.
Just some questions

packages/ws-server/src/webSocketServer.ts
Ivo-Yankov: Messages related to subscriptions need to have a `subscriptionId` property and messages that are in repsonse to a called api method have a `method` property. The ping message doesn't have either of those, so it doesn't get handled by the clients.

Ivo-Yankov: The current TTL period is not extended when a message is sent. Maybe we should change it to be based on the last subscription message?

Nana-EC: Yeah that sounds appreciate.
If ttl is 15 mins and the last message shows up at minute 14, ttl should get reset to 15 mins again.

Nana-EC (approved):

PR #1666, opened 2023-08-22
Add distributed cache using Redis

### LINKED ISSUE #1665: Refactor cacheClient to support distributed cache
### Problem

With the implementation of the shared cache, refactor of the cacheClient class will be needed.

### Solution

Add cacheService, which will utilize both internal and shared cache, depending on defined configuration.

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR refactors `clientCache`, by separating responsibilities between clientService, which takes care of deciding where should the cache be saved/get/deleted - internal or shared.

Added two new classes `localLRUCache` and `redisCache`, who are implementing `ICacheClient` interface, which will be responsible for set/get/delete from it's cache.

Whether redisCache should be enabled or disabled depends entirely on the configuration (REDIS_ENABLED and REDIS_URL). By default it's disabled and all methods/services using cache will use internal cache.

❗ No metrics and acceptance tests were added, because this PR got far too big. We can add them in separate PRs.

**Related issue(s)**:

Fixes #1665 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -0,0 +1,187 @@
...
+   * If the shared cache is enabled and an error occurs while accessing it,
+   * the internal cache is used as a fallback.
+   * @param {string} key - The key associated with the cached value.
+   * @param {string} callingMethod - The name of the method calling the cache.
+   * @param {string} requestIdPrefix - A prefix to include in log messages (optional).
+   * @param {boolean} shared - Whether to use the shared cache (optional, default: false).
+   * @returns {*} The cached value if found, otherwise null.
+   */
+  public async get(key: string, callingMethod: string, requestIdPrefix?: string, shared: boolean = false): Promise<any> {
+    if (shared && this.isSharedCacheEnabled) {
```
Nana-EC, line 102: First part is always true based on setup
```suggestion
    if (this.isSharedCacheEnabled) {
```

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -0,0 +1,187 @@
...
+      try {
+        return await this.sharedCache.get(key, callingMethod, requestIdPrefix);
+      } catch (error) {
+        const redisError = new RedisCacheError(error);
+        this.logger.error(`Error occured while getting the cache from Redis. Fallback to internal cache. Error is: ${redisError.fullError}`);
+      }
```
Nana-EC, line 108: The logic in this class slightly differs from the agreed design.
As shown here the current conception is "if it's not in the shared then check the local". 
That won't always hold, items should be allowed to exist in local only without first checking in shared, items in shared shouldn't all have to be in local.
Services should be able to use both caches as they see fit. As coded a service cannot dictate it's own population and consideration for local and shared.

You could expose a public getter for the 2 caches and allow services to call what they need e.g. `await this.cacheService.local().get(...)` or `await this.cacheService.shared().get(...)`

Nana-EC: If you want their to be 1 default then you can based on `this.isSharedCacheEnabled` but you'd still want to expose the 2 and only interact with 1 at a time.

packages/relay/src/lib/eth.ts
```diff
@@ -314,11 +315,11 @@ export class EthImpl implements Eth {
       } else { // once we finish testing and refining Fixed Fee method, we can remove this else block to clean up code
 
         const cacheKey = `${constants.CACHE_KEY.FEE_HISTORY}_${blockCount}_${newestBlock}_${rewardPercentiles?.join('')}`;
-        feeHistory = this.cache.get(cacheKey, EthImpl.ethFeeHistory, requestIdPrefix);
+        feeHistory = await this.cacheService.get(cacheKey, EthImpl.ethFeeHistory, requestIdPrefix);
```
Nana-EC, line 318: For instance if using the local cache you shouldn't need to require this await

packages/relay/src/lib/eth.ts
```diff
@@ -1860,7 +1861,7 @@ export class EthImpl implements Eth {
   static isArrayNonEmpty(input: any): boolean {
     return Array.isArray(input) && input.length > 0;
   }
-
+  0x297a74308b60a564d6d2a4921ee19c92505430857ea6737b1e6c2fb3ad62ddc571fbfb9d0531ec2babd2c7a118b86b0b
```
Nana-EC, line 1864: Bad merge, should remove

Nana-EC (changes requested):
Good start.
I think there's a few items to iron out on the cacheService side.
I've left comments

packages/relay/src/lib/services/cacheService/cacheService.ts
georgi-l95: Actually `shared` depends on whether the specific method wants to get/set/delete this data from shared cache or internal. isSharedCacheEnabled is true if redis is configured and enabled.

packages/relay/src/lib/services/cacheService/cacheService.ts
georgi-l95: Not sure i understand, but the logic is the same in the design doc -> "by default always go to internal cache, except if the method didn't explicitly asked for shared, by setting the shared variable to true" : 
1. based on shared variable, which is set to true by the method and if it wants to address the Redis or internal. 
2. If shared is true, go to shared cache. If redis is not available for some reason or it returns an error, then fallback to internal
3. If shared is false, go to internal

This way method can decide what cache it wants to use and cacheService takes care of using it and error handling.
By exposing getters and using them by methods (in filter service or subscribe service or somewhere else) the whole cache Service is not needed, because it's main reason is to hide implementation from external services (that way if we change something, we don't need to change it everywhere as long as we keep following the interface) and another main reason is the error handling, if something goes wrong with the shared cache to fallback and report.

packages/relay/src/lib/eth.ts
georgi-l95: True, something left from the rebase i guess. Will remove

packages/relay/src/lib/clients/cacheClient/redisCache.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.client.connect();
+
+    this.client.on('ready', function () {
+      logger.info('Connected to Redis server successfully!');
+    });
+
+    this.client.on('error', function (error) {
+      const redisError = new RedisCacheError(error);
+      if (redisError.isSocketClosed()) {
+        logger.error(`Error occured with Redis Connection. Error is: ${redisError.message}`);
```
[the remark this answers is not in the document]
georgi-l95: Changed.

packages/relay/src/lib/clients/cacheClient/redisCache.ts
```diff
@@ -0,0 +1,166 @@
...
+      } else {
+        logger.error(`${redisError.fullError}`);
```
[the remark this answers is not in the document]
georgi-l95: Added.

packages/relay/src/lib/clients/index.ts
```diff
@@ -18,6 +18,7 @@
  *
  */
 
-export * from './clientCache';
+export * from './cacheClient/localLRUCache';
```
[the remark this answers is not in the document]
georgi-l95: Changed.

packages/relay/src/lib/errors/RedisCacheError.ts
```diff
@@ -0,0 +1,47 @@
+
+/*-
+ *
+ * Hedera JSON RPC Relay
+ *
+ * Copyright (C) 2022 Hedera Hashgraph, LLC
```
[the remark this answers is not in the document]
georgi-l95: Changed.

packages/relay/src/lib/services/cacheService/cacheService.ts
Nana-EC: Ahh I mixed up `shared` with `this.sharedCache`. Thanks

packages/relay/src/lib/services/cacheService/cacheService.ts
Nana-EC: By exposing getters and using them by methods (in filter service or subscribe service or somewhere else) the whole cache service is not needed
Yes this is one of the items I'm trying to resolve with my suggestion.
I missed the `shared` param so i'm re-reviewing this and thinking through how it scales

Nana-EC: Resolving this as I'm on board now that i've re-reviewed

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -0,0 +1,187 @@
...
+      return this.internalCache.get(key, callingMethod, requestIdPrefix);
```
Nana-EC, line 109: I don't think this fallback should be there. To honour the intent of `shared` we shouldn't manage both. Allow callers to manage that logic
```suggestion
      return null;
```

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -0,0 +1,187 @@
...
+    shared: boolean = false
+  ): void {
+    if (shared && this.isSharedCacheEnabled) {
+      try {
+        return this.sharedCache.set(key, value, callingMethod, ttl, requestIdPrefix);
+      } catch (error) {
+        const redisError = new RedisCacheError(error);
+        this.logger.error(`Error occured while setting the cache to Redis. Fallback to internal cache. Error is: ${redisError.fullError}`);
+      }
+      this.internalCache.set(key, value, callingMethod, ttl, requestIdPrefix);
```
Nana-EC, line 141: DOn't think we should have this.
In the future if we want to keep the caches in sync we can explore that
```suggestion
```

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+
+  /**
+   * Clears the cache.
+   * If the shared cache is enabled and an error occurs while clearing it, just logs the error.
+   * Else the internal cache clearing is attempted.
+   * @param {boolean} shared - Whether to clear the shared cache (optional, default: false).
+   */
+  public clear(shared: boolean = false): void {
```
Nana-EC, line 175:
```suggestion
  public clear(requestIdPrefix?: string, shared: boolean = false): void {
```

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -0,0 +1,187 @@
...
+    if (shared && this.isSharedCacheEnabled) {
+      try {
+        return this.sharedCache.clear();
+      } catch (error) {
+        const redisError = new RedisCacheError(error);
+        this.logger.error(`Error occured while clearing Redis cache. Error is: ${redisError.fullError}`);
```
Nana-EC, line 181: All the logs in this class as missing the use of the `requestIdPrefix` values
```suggestion
        this.logger.error(`${requestIdPrefix} Error occurred while clearing Redis cache. Error is: ${redisError.fullError}`);
```

Nana-EC (changes requested):
Nice progress. 
Left some more comments

packages/relay/src/lib/clients/cache/redisCache.ts
```diff
@@ -0,0 +1,164 @@
...
+   * @param {string} [requestIdPrefix] - The optional request ID prefix.
+   * @returns {Promise<any | null>} The cached value or null if not found.
+   */
+  async get(key: string, callingMethod: string, requestIdPrefix?: string | undefined) {
+    const result = await this.client.get(key);
+    if (result) {
+      this.logger.trace(
+        `${requestIdPrefix} returning cached value ${key}:${JSON.stringify(result)} on ${callingMethod} call`
+      );
+      //add metrics
```
AlfredoG87, line 109:
```suggestion
      //TODO: add metrics
```

so is easier to spot that something is missing here

packages/relay/src/lib/clients/cache/redisCache.ts
```diff
@@ -0,0 +1,164 @@
...
+    requestIdPrefix?: string | undefined
+  ): Promise<void> {
+    const serializedValue = JSON.stringify(value);
+    const resolvedTtl = ttl ?? this.options.ttl;
+
+    await this.client.setEx(key, resolvedTtl, serializedValue);
+    this.logger.trace(
+      `${requestIdPrefix} caching ${key}: ${serializedValue} on ${callingMethod} for ${resolvedTtl} ms`
+    );
+    //add metrics
```
AlfredoG87, line 139:
```suggestion
    //TODO: add metrics
```

packages/relay/src/lib/clients/cache/redisCache.ts
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @param {string} key - The cache key.
+   * @param {string} callingMethod - The name of the calling method.
+   * @param {string} [requestIdPrefix] - The optional request ID prefix.
+   * @returns {Promise<void>} A Promise that resolves when the value is deleted from the cache.
+   */
+  async delete(key: string, callingMethod: string, requestIdPrefix?: string | undefined): Promise<void> {
+    await this.client.del(key);
+    this.logger.trace(`${requestIdPrefix} delete cache for ${key} on ${callingMethod} call`);
+    //add metrics
```
AlfredoG87, line 153:
```suggestion
    //TODO: add metrics
```

packages/server/tests/localAcceptance.env
```diff
@@ -3,6 +3,8 @@ OPERATOR_ID_MAIN=0.0.2
 OPERATOR_KEY_MAIN=302e020100300506032b65700422042091132178e72057a1d7528025956fe39b0b847f200ab59b2fdd367017f3087137
 CHAIN_ID=0x12a
 MIRROR_NODE_URL_WEB3=http://127.0.0.1:8545
+REDIS_ENABLED=true
```
AlfredoG87, line 6: does this needs redis service to be running on `hedera-local-node` to work properly?

AlfredoG87 (commented):
Looking good, just leaved some minor suggestions so we don't miss metrics in future PR

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -0,0 +1,187 @@
...
+   * Retrieves a cached value associated with the given key.
+   * If the shared cache is enabled and an error occurs while accessing it,
+   * the internal cache is used as a fallback.
+   * @param {string} key - The key associated with the cached value.
+   * @param {string} callingMethod - The name of the method calling the cache.
+   * @param {string} requestIdPrefix - A prefix to include in log messages (optional).
+   * @param {boolean} shared - Whether to use the shared cache (optional, default: false).
+   * @returns {*} The cached value if found, otherwise null.
+   */
+  public async get(key: string, callingMethod: string, requestIdPrefix?: string, shared: boolean = false): Promise<any> {
```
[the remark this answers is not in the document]
georgi-l95: Agree, done.

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -0,0 +1,187 @@
...
+    key: string,
+    value: any,
+    callingMethod: string,
+    ttl?: number,
+    requestIdPrefix?: string,
+    shared: boolean = false
+  ): void {
+    if (shared && this.isSharedCacheEnabled) {
+      try {
+        return this.sharedCache.set(key, value, callingMethod, ttl, requestIdPrefix);
```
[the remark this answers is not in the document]
georgi-l95: We can wait it out or just execute it and continue forward without waiting. It will be executed in the background.

packages/relay/src/lib/services/cacheService/cacheService.ts
georgi-l95: Added.

packages/relay/src/lib/services/cacheService/cacheService.ts
georgi-l95: Added and to other missing places as well.

packages/relay/src/lib/clients/cache/redisCache.ts
georgi-l95: Done.

packages/relay/src/lib/clients/cache/redisCache.ts
georgi-l95: Done.

packages/relay/src/lib/clients/cache/redisCache.ts
georgi-l95: Done.

docs/configuration.md
```diff
@@ -24,68 +24,71 @@ These properties are noted below and should be custom set per deployment.
...
+| `REDIS_ENABLED`                             | "true"                   | Enable usage of Redis as shared cache                                                                                                                                                                                                                                                                  |
+| `REDIS_URL`                                 | "redis://127.0.0.1:6379" | Sets the url for the Redis shared cache                                                                                                                                                                                                                                                                |
+| `REDIS_RECONNECT_DELAY`                     | "1000"                   | Sets the delay between reconnect retries from the Redis client                                                                                                                                                                                                                                         |
```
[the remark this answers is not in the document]
georgi-l95: Done.

charts/hedera-json-rpc-relay/value-test.yaml
```diff
@@ -129,6 +129,9 @@ config:
   HAPI_CLIENT_TRANSACTION_RESET: 50
   HAPI_CLIENT_DURATION_RESET: 3600000
   HAPI_CLIENT_ERROR_RESET: [50]
+  REDIS_ENABLED: false
+  REDIS_URL: ''
+  REDIS_RECONNECT_DELAY: 1000
```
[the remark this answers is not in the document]
georgi-l95: Done.

charts/hedera-json-rpc-relay-websocket/values.yaml
```diff
@@ -28,6 +28,9 @@ config:
   WS_MAX_CONNECTION_TTL: 300000
   WS_MULTIPLE_ADDRESSES_ENABLED: false
   WS_SUBSCRIPTION_LIMIT: 100
+  REDIS_ENABLED: false
+  REDIS_URL: ''
+  REDIS_RECONNECT_DELAY: 1000
```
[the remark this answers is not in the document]
georgi-l95: Done.

packages/server/tests/localAcceptance.env
georgi-l95: No, only requirement is REDIS_URL to be pointed to correct address. But for acceptance test we will use hedera-local-node redis container.

packages/relay/src/lib/services/cacheService/cacheService.ts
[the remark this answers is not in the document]
Nana-EC: Ohhh sneaky. Hmmm
Whiles optimal in most cases it might result in some double sets because a future process gets, find nothing and then sets.
Maybe add a note as to this intent and let watch how things unfold

charts/hedera-json-rpc-relay/values.yaml
```diff
@@ -62,7 +62,7 @@ cronjob:
     pullPolicy: IfNotPresent
     repository: postman/newman
     tag: 5.3.1-alpine
-  postman: '' # Custom postman.json in base64 encoding
+  postman: ""  # Custom postman.json in base64 encoding
```
AlfredoG87, line 65: what does this fix?

georgi-l95: old commit, can you check again? my linter edited those and i've reverted them to the original, it should be the same as main now

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -517,13 +517,13 @@ export class MirrorNodeClient {
         return `${constants.CACHE_KEY.GET_CONTRACT}.valid.${contractIdOrAddress}`;
     }
 
-    public getIsValidContractCache(contractIdOrAddress): any {
+    public async getIsValidContractCache(contractIdOrAddress): Promise<any> {
```
Nana-EC, line 520: `await` not necessary with `get`
```suggestion
    public getIsValidContractCache(contractIdOrAddress): any {
```

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -517,13 +517,13 @@ export class MirrorNodeClient {
...
         const cachedLabel = this.getIsValidContractCacheLabel(contractIdOrAddress);
-        return this.cache.get(cachedLabel, MirrorNodeClient.GET_CONTRACT_ENDPOINT);
+        return this.cacheService.get(cachedLabel, MirrorNodeClient.GET_CONTRACT_ENDPOINT);
     }
 
     public async isValidContract(contractIdOrAddress: string, requestIdPrefix?: string, retries?: number) {
-        const cachedResponse: any = this.getIsValidContractCache(contractIdOrAddress);
+        const cachedResponse: any = await this.getIsValidContractCache(contractIdOrAddress);
```
Nana-EC, line 526: Shouldn't be needed right?
```suggestion
        const cachedResponse: any = this.getIsValidContractCache(contractIdOrAddress);
```

packages/relay/src/lib/errors/RedisCacheError.ts
```diff
@@ -0,0 +1,47 @@
...
+     */
+    constructor(error: any) {
+        this.type = error.type;
+        this.message = error.message;
+        this.stack = error.stack;
+        this.fullError = error;
+    }
+
+    public isSocketClosed(): boolean {
+        return this.type === "SocketClosedUnexpectedlyError";
```
Nana-EC, line 45: nit: set text as constant

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -0,0 +1,213 @@
...
+   * @param {string} callingMethod - The name of the calling method.
+   * @param {string} [requestIdPrefix] - The optional request ID prefix.
+   * @param {boolean} [shared=false] - Whether to use the shared cache (optional, default: false).
+   * @returns {Promise<any>} A Promise that resolves with the cached value or null if not found.
+   */
+  public async getAsync(
+    key: string,
+    callingMethod: string,
+    requestIdPrefix?: string,
+    shared: boolean = false
```
Nana-EC, line 107: Is shared needed now?
localLRUCache will never require `async` so should be served by the `get()` call.
I'd simplify and remove the use of `shared` and the else

Nana-EC (changes requested):
Nice. Almost there.
Minor fixes left around `async` and  `shared`

packages/relay/src/lib/services/cacheService/cacheService.ts
georgi-l95: Removed it and the else. Thanks.

packages/relay/src/lib/errors/RedisCacheError.ts
georgi-l95: Done.

packages/relay/src/lib/clients/mirrorNodeClient.ts
georgi-l95: You're right, thanks.

packages/relay/src/lib/clients/mirrorNodeClient.ts
georgi-l95: Yes, removed the async and promise

Nana-EC (approved):
LG. Thanks

Issue #1648, opened 2023-08-18
eth_getBalance returns 500 error due to account not found


AlfredoG87:
### Description

Due to an increase of failed requests on both `previewnet` and `testnet` did some research and found the following:
1. Those failed requests were due to `eth_getBalance` method and the response status code was `500` (Server error)
2. The logic inside the `eth_getBalance` is trying to get account info, and getting `404` (null account)
3. trying to read null --> Error.

https://cloudlogging.app.goo.gl/zoCAEqJXRgNr4EWp6

### Steps to reproduce
```
curl --location 'https://testnet.hashio.io/api' \
--header 'Content-Type: application/json' \
--data '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "eth_getBalance",
    "params": [
        "0x00a338914aa6c7141241e8e61f613dc8f0baabbf",
        "0xdef41"
    ]
}'
```
### Additional context

https://cloudlogging.app.goo.gl/zoCAEqJXRgNr4EWp6

https://cloudlogging.app.goo.gl/WjWqwffPk3GRgMvWA

### Hedera network

testnet, previewnet

### Version

v0.29.1

### Operating system

macOS

Nana-EC:
We should handle not found accounts with 0 HBAR response as is consistent on the EVM

PR #1671, opened 2023-08-22
eth_getBalance not found account for past 15 minutes block in the past read null fix

### LINKED ISSUE #1670: eth_getBalance still failing with 500 status code on some scenarios
### Description

After the release of `0.30.0-rc1` the amount of failed request with `500` status seems to have decreased, and even if some cases were fixed, it seems like there are still other scenarios that are failing with a similar error

### Steps to reproduce

**address does not have to exist.
block number should be of the past 15 minutes, but not latest 2 blocks.**

**Run the following curl request:**

```
curl --location 'https://previewnet.hashio.io/api' \
--data '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "eth_getBalance",
    "params": [
        "0x0851072d7bb726305032eff23cb8fd22eb74c85b",
        "0xa1864"
    ]
}'
```

**Response:**
```
{
    "jsonrpc": "2.0",
    "id": "1",
    "error": {
        "code": -32603,
        "name": "Internal error",
        "message": "[Request ID: 1e96c0c0-575a-437b-927f-b7597aab6596] Error invoking RPC: Cannot read properties of null (reading 'balance')"
    }
}
```

Log for new case found:

https://cloudlogging.app.goo.gl/kqRfDGdSgkezzCkY6

### Additional context

_No response_

### Hedera network

previewnet

### Version

v0.30.0-rc1

### Operating system

None

AlfredoG87:
**Description**:
When trying to fetch the balance for accounts that does not exists and a block number in the past 15 minutes is provided, there was a null check missing that provoked a unexpected error `500` be returned instead of the expected `0x0` response.

This PR adds the check and missing unit test.

**Related issue(s)**: #1648

Fixes #1670 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -822,34 +822,36 @@ export class EthImpl implements Eth {
               let currentBalance = 0;
               let balanceFromTxs = 0;
               mirrorAccount = await this.mirrorNodeClient.getAccountPageLimit(account, requestIdPrefix);
-              if (mirrorAccount.balance) {
-                currentBalance = mirrorAccount.balance.balance;
-              }
-
-              // The balance in the account is real time, so we simply subtract the transactions to the block.timestamp.to to get a block relevant balance.
-              // needs to be updated below.
-              const nextPage: string = mirrorAccount.links.next;
-
-              if (nextPage) {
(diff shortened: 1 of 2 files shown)


ebadiere (approved):
LG.

Nana-EC (approved):

PR #1675, opened 2023-08-23
Support synthetic logs in eth_subscribe

### LINKED ISSUE #1602: Support synthetic logs in eth_subscribe
### Problem

HAPI transactions now generate log records. When attempting to subscribe for logs related to an hts token the ws server responds with:
```
    code: -32602,
    name: 'Invalid parameter',
    message: 'Invalid parameter filters.address: 0x..... is not a valid contract type or does not exists'
```

### Solution

Modify the address validation. Currently it only checks if it is a contract.

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
Fixes the address validation logic before creating a subscription through `eth_subscribe`. It now checks if the address is of an existing contract or token. This makes it possible to subscribe to `transfer` and `approve` events generated by HTS tokens.

**Related issue(s)**:

Fixes #1602

### DIFF
--- a/packages/ws-server/src/webSocketServer.ts
+++ b/packages/ws-server/src/webSocketServer.ts
@@ -90,9 +90,9 @@ function getMultipleAddressesEnabled() {
     return process.env.WS_MULTIPLE_ADDRESSES_ENABLED === 'true';
 }
 
-async function validateIsContractAddress(address, requestId) {
+async function validateIsContractOrTokenAddress(address, requestId) {
 
-    const isContractOrToken = await mirrorNodeClient.resolveEntityType(address, [constants.TYPE_CONTRACT, constants.TYPE_TOKEN], 'eth_subscribe', requestId)
+    const isContractOrToken = await mirrorNodeClient.resolveEntityType(address, [constants.TYPE_CONTRACT, constants.TYPE_TOKEN], constants.METHODS.ETH_SUBSCRIBE, requestId)
     if (!isContractOrToken) {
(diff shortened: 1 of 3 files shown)


packages/ws-server/src/webSocketServer.ts
```diff
@@ -90,9 +91,10 @@ function getMultipleAddressesEnabled() {
 }
 
 async function validateIsContractAddress(address, requestId) {
```
Nana-EC, line 93: nit: rename the method since it's doing more than it states

packages/ws-server/src/webSocketServer.ts
```diff
@@ -90,9 +91,10 @@ function getMultipleAddressesEnabled() {
...
-    const isContract = await mirrorNodeClient.isValidContract(address, requestId)
-    if (!isContract) {
-        throw new JsonRpcError(predefined.INVALID_PARAMETER(`filters.address`, `${address} is not a valid contract type or does not exists`), requestId);
+
+    const isContractOrToken = await mirrorNodeClient.resolveEntityType(address, [constants.TYPE_CONTRACT, constants.TYPE_TOKEN], 'eth_subscribe', requestId)
```
Nana-EC, line 95: Use a constant for `eth_subscribe`.

Nana-EC (changes requested):
Looking good.
Seems like the PR is trying to support subscribing to logs where the address is for an HTS token. Please update the PR with an appropriate description and see my other suggestions

AlfredoG87 (approved):
LGTM 👍

Nana-EC (approved):
LG

PR #1678, opened 2023-08-24
Refactor connection TTL logic in WS server

### LINKED ISSUE #1664: Refactor TTL logic for eth_subscribe
### Problem

Websocket connections are automatically closed after a TTL period. The timer should be reset after a message has been sent either from the client or the server.

Ivo-Yankov:
**Description**:
- Changes the way the automatic closing of a WS connection is achieved. It used to close after a set amount of time. With this PR the timer is reset whenever any message is sent to/from the client (excluding the keep-alive ping messages).
- The env var `WS_CONNECTION_TTL` is renamed to `WS_INACTIVITY_TTL` to  better describe its purpose.

**Related issue(s)**:

Fixes #1664 

**Notes for reviewer**:

**Checklist**


packages/relay/tests/lib/subscriptionController.spec.ts
```diff
@@ -29,19 +29,24 @@ import sinon from 'sinon';
 import dotenv from "dotenv";
 import path from 'path';
 import {Registry} from "prom-client";
+import ConnectionLimiter from "@hashgraph/json-rpc-ws-server/dist/ConnectionLimiter";
```
Nana-EC, line 32: Q: why is this file in relay vs in `ws-server` package again?
Shouldn't we move it to `ws-server`?

Nana-EC (approved):

packages/relay/tests/lib/subscriptionController.spec.ts
Ivo-Yankov: I was thinking that we want to keep the business logic in the `relay` package, and the `server` and `ws-server` are more related to routing. It does make sence to move it to `ws-server`, we can do that when we start with the refactoring of the project.

Nana-EC: @Ivo-Yankov  Okay please open a ticket so we don't miss it and we can plan it

Nana-EC (approved):

Issue #1672, opened 2023-08-23
Flaky Chart Test failing most of the time


AlfredoG87:
### Problem

When running CI Tests one test is failing most of the time and needs several retries in order to work.

Eventually the test passes.

"Charts / install"

<img width="339" alt="image" src="https://github.com/hashgraph/hedera-json-rpc-relay/assets/123040664/de82b84c-ead5-45f9-8f2e-ebb2bad531cc">

### Solution

Improve the Test so it always succeeds

PR #1682, opened 2023-08-25
Add e2e tests for hbar/token allowance

### LINKED ISSUE #1681: Add additional E2E tests for HBAR/Token allowances.
### Problem

We don't have full coverage of approvals/transfers and allowances for HBAR/Token when wrapped like ERC20.

### Solution

Add and expand current test cases so we can have more coverage and notice, if there are regressions.

### Alternatives

_No response_

natanasow:
**Description**:

There was a test but was commented on because the image with the fix wasn't ready. HBAR approval could be done only via hapi but we don't test hapi calls in acceptance tests, so no actions are needed.

PS: charts install fails due to removing pnpm from dockerfile, so it has to be added in packages/server as a dependency

**Related issue(s)**:

Fixes #1681 

**Blocked by**:
- https://github.com/hashgraph/hedera-mirror-node/pull/6778

**Notes for reviewer**:

**Checklist**


isavov (approved):

AlfredoG87 (approved):
LGTM

packages/server/package.json
```diff
@@ -18,7 +18,8 @@
         "mocha": "^9.2.2",
         "pino": "^7.11.0",
         "pino-pretty": "^7.6.1",
-        "uuid": "^3.3.2"
+        "uuid": "^3.3.2",
+        "pnpm": "^8.6.6"
```
AlfredoG87, line 22: just one question? 
why was pnpm needed here?

natanasow: It's fixing https://github.com/hashgraph/hedera-json-rpc-relay/issues/1672. The `pnpm` was removed from the DockerFile and the `charts install` ci could not be able to build `packages/server` and that's why it started to fail recently. Adding `pnpm` as a dependency in `packages/server` will make `charts install` ci to succeed.

AlfredoG87: Ohh, I see, thanks!! indeed the Test has passed on this branch. 👍 
Thank you!

Nana-EC (approved):

AlfredoG87:
@natanasow should we split this PR in 2? so we can move forward with the fix of #1672 ?

natanasow:
> @natanasow should we split this PR in 2? so we can move forward with the fix of #1672 ?

Yep, yesterday I found a bug in the latest mirror node image and this PR will be blocked for a while. Will open a separate PR only for #1672 fix :rocket:.

Nana-EC:
@natanasow you should rebase. I think Eric address the other items so it should just be the  pnpm part

natanasow:
> @natanasow you should rebase. I think Eric address the other items so it should just be the pnpm part

The `pnpm` had been already added in another PR I guess, so this PR could be closed because it doesn't change anything right now :confused: .

PR #1683, opened 2023-08-25
Add request-id as part of all response headers

### LINKED ISSUE #1629: Request-Id is only returned when errors occurr
### Problem

Sometimes a method is not explicitly failing, but just returning and odd or plain wrong value, it would be useful for users when reporting bugs or issues to devs, if they can provide the `RequestId` of the request. but those are only returned as part of the body, when there is an explicit error.

Example of current errors responses with request id:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32000,
        "name": "Block range too large",
        "message": "[Request ID: 0964ff77-d3f7-46c5-b3fd-9799c0e73323] Exceeded maximum block range: 1000"
    }
}
```

### Solution

Always return the `RequestId` for all request regardless of the outcome.
Return the RequestId as part of the `Headers`

### Alternatives

_No response_

AlfredoG87:
**Description**:
Add request-id as part of all response headers

**Related issue(s)**:

Fixes #1629 

**Notes for reviewer**:
image

**Checklist**

### DIFF
--- a/packages/server/src/koaJsonRpc/index.ts
+++ b/packages/server/src/koaJsonRpc/index.ts
@@ -97,6 +97,7 @@ export default class KoaJsonRpc {
       let body, result;
 
       this.requestId = ctx.state.reqId;
+      ctx.set('request-id', this.requestId);
       
       if (this.token) {
         const headerToken = ctx.get('authorization').split(' ').pop();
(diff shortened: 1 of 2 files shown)


Nana-EC (approved):
LG.

PR #1685, opened 2023-08-25
Turn allowance test back on now that fix is in mirror node.

ebadiere:
Turn allowance test back on.

Addresses# https://github.com/hashgraph/hedera-mirror-node/issues/6721 and https://github.com/hashgraph/hedera-mirror-node/issues/6775.  Depends on mirror node version 0.88.0 with the fixes to the issues listed.


AlfredoG87 (dismissed):
LGTM

PR #1695, opened 2023-08-29
Automatically formatting and cleaning code on commit

### LINKED ISSUE #559: Add automated prettier command
### Problem

Files continue to have inconsistent formatting since not all PRs apply the prettier logic.

### Solution

Automate the prettier command run.
Consider something like [husky precommit](https://prettier.io/docs/en/precommit.html)

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR adds automatically formatting and cleaning code on commit.
Code style is standard for typescript. All files changed prior commit will be formatted.

**Related issue(s)**:

Fixes #559 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,19 +2,21 @@
     "name": "root",
     "devDependencies": {
         "@types/chai-as-promised": "^7.1.5",
-        "@typescript-eslint/eslint-plugin": "^5.11.0",
-        "@typescript-eslint/parser": "^5.11.0",
+        "@typescript-eslint/eslint-plugin": "^6.5.0",
+        "@typescript-eslint/parser": "^6.5.0",
         "axios-mock-adapter": "^1.20.0",
         "chai-as-promised": "^7.1.1",
         "codecov": "^3.8.3",
-        "eslint": "^7.32.0",
+        "eslint": "^8.48.0",
         "eslint-config-airbnb-base": "^15.0.0",
         "eslint-config-prettier": "^8.3.0",
-        "eslint-plugin-import": "^2.25.4",
+        "eslint-plugin-import": "^2.28.1",
         "ethereum-waffle": "^4.0.7",
+        "husky": "^8.0.3",
+        "lint-staged": "^14.0.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -71,5 +75,21 @@
         "protobufjs": "^7.2.4",
         "semver": "^7.5.3",
         "nx": "16.5.5"
-    }
+    },
+    "lint-staged": {
+        "packages/**/src/**/*.ts": [
+          "prettier --write"
+        ],
+        "!(packages/**/src/dist/)*.ts": [
```
Nana-EC, line 83: We don't check dist in, is this just a nice to have for local troubleshooting view?

package.json
```diff
@@ -71,5 +75,21 @@
...
+            "prettier --write"
+        ],
+        "!(packages/**/src/node_modules/)*.ts": [
```
Nana-EC, line 86: Same as above, is this just a nice to have for local troubleshooting view?

package.json
Nana-EC, line 83: No such dir. Did you mean?
```suggestion
        "!(packages/**/dist/)*.ts": [
```

Nana-EC (commented):
Nice. Some questions

package.json
georgi-l95: Yes, I think we don't need to check and format anything under dist folder. Also do you think we should also include tests to be properly formated upon changes and commits ? 
```
        "packages/**/tests/**/*.ts": [
          "prettier --write"
        ],
```

package.json
georgi-l95: Yes, i believe so.

package.json
georgi-l95: Ah, yes you're right. I've typed the wrong dir.

Nana-EC: Yeah we should format test files also

georgi-l95: Added them as well.

Nana-EC (approved):

PR #1699, opened 2023-08-30
Enable shared cache in Filter API

### LINKED ISSUE #1650: Refactor `eth_filter` and `eth_subscribe` to use shared cache
### Problem

We've added distributed cache in the relay, now APIs can take advantage of it.

### Solution

Update and test `eth_filter` and `eth_subscribe` to use shared cache when available.

### Alternatives

_No response_

Ivo-Yankov:
**Description**:
Enables `redisCache` for caching filter ids and data. This does not affect other data used by the Filter API - for example results from the mirror node are still saved in the local cache.

**Related issue(s)**:

Fixes #1650 

**Notes for reviewer**:

**Checklist**


Nana-EC (dismissed):

ebadiere (dismissed):
LG.

Nana-EC (approved):

hexdigest:
This is a breaking change guys. If there wasn't a shared cache configured prior to deploying this update it breaks the filters functionality. The key is *always* written to the shared cache but then it's being read from whatever cache is default. No error is returned to the client except "Filter not found".

PR #1704, opened 2023-09-01
Apply formatting

georgi-l95:
**Description**:
This PR apply prettier formatting to all files, according to TS Coding Conventions. Now that we have auto format on commit, this is one time format to reduce unnecessary changes in new PRs.
Also adds 10 more seconds after graph node start, to allow enough time before creating subgraph.
Also attempt to fix chart install workflow and dapp tests.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


AlfredoG87 (dismissed):
I am afraid this will break and ask for rebase of every single outstanding PRs but nevertheless worth to have conventions implemented on the project 👍

georgi-l95:
@AlfredoG87 i've rebased and lost approval

.prettierrc
```diff
@@ -1,5 +1,5 @@
 {
-    "singleQuote": true,
+    "singleQuote": false,
```
Nana-EC, line 2: I prefer single actually 😆

Nana-EC (changes requested):
Thanks.
Blocking whiles I confirm some items

ebadiere (approved):
Mainly single quotes to double quotes, added semi-colons and some method call formatting.  LGTM.

.prettierrc
georgi-l95: Okay, done, switched to single :D

PR #1709, opened 2023-09-04
Add Documentation for Debug endpoint

### LINKED ISSUE #1479: Add documentation for debug endpoint.
### Problem

As a developer I'd like to be able to use debug endpoints.

### Solution

Add design documentation for debug endpoint, what is, how it works. Can be influenced by the `eth_subscribe` doc.

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR adds design documentation for the debug endpoint. What new methods are going to be introduced and how.

**Related issue(s)**:

Fixes #1479

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/design/debug-api.md
+++ b/docs/design/debug-api.md
@@ -1,0 +1,346 @@
+# Debug API design
+
+## Purpose
+
+The Debug API gives you access to several non-standard RPC methods, which will allow you to inspect, debug and set certain debugging flags during runtime. Also, it could help you replay all the transactions that have been executed prior and save considerable effort for your debugging process.
+
+## Goals
+
+1. **Debugging Capabilities**: Enable developers to retrieve detailed information about transactions, execution traces, and other debugging-related data.
+2. **Enhanced Insights**: Provide deeper insights into the behavior of smart contracts, transaction execution, and state changes on the Hedera network.
+3. **Ethereum Equivalent**: Makes the Hedera JSON-RPC Relay more equivalent to other providers used in the ethereum ecosystem, by providing the needed endpoints for easier debugging by developers.
+
+## Architecture
+
+New Debug API methods can be added and handled by adding debug service, which will be blueprinted by new `debug` interface added in the relay interface itself. Each method will be separate from the other and they won't overlap in terms of functionality.
+
+### Interface
+```javascript
+export interface Debug {
+  traceTransaction(transactionHash, { tracer, tracerConfig });
+
+  accountRange(blockNrOrHash, start, maxResults);
+
+  getModifiedAccountsByHash(startHash, endHash);
+
+  getModifiedAccountsByNumber(startNum, endNum);
+
+  verbosity(level);
+}
+```
+
+### Service class
+```javascript
+class DebugService implements Debug{
+  traceTransaction(transactionHash, { tracer, tracerConfig }) {
+    // Implementation of traceTranscation depending on passed params.
+  }
+
+  accountRange(blockNrOrHash, start, maxResults) {
+    // Implementation of accountRange depending on passed params.
+  }
+
+  getModifiedAccountsByHash(startHash, endHash){ 
+    // Implementation of getModifiedAccountsByHash depending on passed params.
+  }
+
+  getModifiedAccountsByNumber(startNum, endNum){
+    // Implementation of getModifiedAccountsByNumber depending on passed params.
+  }
+
+  verbosity(level){
+    // Implementation of verbosity depending on passed params.
+  }
+}
+```
+
+### Method description
+
+`debug_traceTransaction` - Attempts to run the transaction in the exact same manner as it was executed on the network.
+
+#### Parameters
+`transactionHash` - string - This is the hash of the transaction that we want to trace.
+`tracer` - string - to specify the type of tracer. In the beginning only `callTracer` will be accepted.
+`tracerConfig` - object - consists of one property inside called `onlyTopCall`, which is a boolean. 
+
+#### Example Request
+
+```JSON
+{"jsonrpc": "2.0",
+         "id": 1,
+         "method": "debug_traceTransaction",
+         "params": [
+          "0x8fc90a6c3ee3001cdcbbb685b4fbe67b1fa2bec575b15b0395fea5540d0901ae",
+          {
+               "tracer": "callTracer"
+          }
+         ]
+       }
+```
+#### Example Response
+```JSON
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": {
+    "type": "CALL",
+    "from": "0x5067c042e35881843f2b31dfc2db1f4f272ef48c",
+    "to": "0x3ee18b2214aff97000d974cf647e7c347e8fa585",
+    "value": "0x0",
+    "gas": "0x17459",
+    "gasUsed": "0x166cb",
+    "input": "0x0f5287b0000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb48000000000000000000000000000000000000000000000000000000001debea42000000000000000000000000000000000000000000000000000000000000000167c46aa713cfe47608dd1c16f8a0325208df084c3cbebf9f366ad0eafc2653e400000000000000000000000000000000000000000000000000000000001e8542000000000000000000000000000000000000000000000000000000006eca0000",
+    "output": "0x000000000000000000000000000000000000000000000000000000000001371e",
+    "calls": [
+      {
+        "type": "DELEGATECALL",
+        "from": "0x3ee18b2214aff97000d974cf647e7c347e8fa585",
+        "to": "0x76364611e457b1f97cd58ffc332ddc7561a193f6",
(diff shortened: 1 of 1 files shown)


docs/design/debug-api.md
```diff
@@ -0,0 +1,352 @@
...
+
+  getModifiedAccountsByNumber(startNum, endNum){
+    // Implementation of getModifiedAccountsByNumber depending on passed params.
+  }
+}
+```
+
+### Method description
+
+`debug_traceTransaction` - Attempts to run the transaction in the exact same manner as it was executed on the network.
```
Nana-EC, line 47: Please add a high level description of the logic to enable this in terms of other calls. That is what mirror node calls will the logic be making and what data formatting might be necessary

docs/design/debug-api.md
```diff
@@ -0,0 +1,352 @@
...
+      }
+    ]
+  }
+}
+```
+
+
+### Method description
+
+`debug_getModifiedAccountsByHash` - Returns all accounts that have changed between the two blocks specified.
```
Nana-EC, line 219: Please add a high level description of the logic to enable this in terms of other calls. That is what mirror node calls will the logic be making and what data formatting might be necessary.

docs/design/debug-api.md
```diff
@@ -0,0 +1,352 @@
...
+
+1. Finalize design document
+
+#### Milestone 2
+
+1. Implement new interfaces and classes.
+2. Implement `debug_traceTransaction`.
+3. Add needed acceptance tests for `debug_traceTransaction`.
+
+#### Milestone 3
```
Nana-EC, line 349: For the `deug_getModifiedAccounts...` I'm not clear on the amount of work and extra API calls needed. Once you expand let's see if Milestone 3 should follow Milestone 2 immediately or something for a later date

Nana-EC (changes requested):
Nice.
Just need a bit more information on calls to be made

docs/design/debug-api.md
```diff
@@ -0,0 +1,352 @@
...
+1. Overall functionality of the methods.
+2. Test different scenarios with all possible parameter combinations.
+3. Case where transaction is not found for `debug_traceTransaction`.
+4. Case where block hash is not found for `debug_getModifiedAccountsByHash`.
+5. Case where block number is not found for `debug_getModifiedAccountsByNumber`.
+
+
+## Deployment
+
+Debug API will run alongside the already available HTTP server.
```
AlfredoG87, line 323: does this means, it will be part of the same "Relay" module right?
I recommend that is disabled by default and that these methods are available only if enabled.
`DEBUG_API_ENABLED=false`

docs/design/debug-api.md
```diff
@@ -0,0 +1,352 @@
...
+
+## Answered Questions
+
+1. What is the purpose of Debug API and what kind of goals adding it fulfil ?
+2. What new interfaces and classes would it need ?
+3. What new endpoints will be exposed ?
+4. What kind of parameters each method accepts ?
+5. What kind of response is expected to return ?
+6. What is the expected request and response ?
+7. What limits should be introduced ?
```
AlfredoG87, line 333: what limits are being introduced? I did not found any limits sections or on each method.
I would recommend that the limits for each method should also be configurable.

AlfredoG87 (commented):
Overall looks good.

docs/design/debug-api.md
georgi-l95: Yes, they are part of the relay module.
Once exposed I'm not sure that we can just disabled them, more like throw an error if they are disabled like -> Method Disabled when you try to call debug_traceTransaction for example.

docs/design/debug-api.md
georgi-l95: Limits are in their own Limit section . Also yes, fully agree that all limits should be configurable.

docs/design/debug-api.md
AlfredoG87: Thanks, that is what I thought with respect to the `alongside` but wanted to confirm.

This is for `TRACE` API, but is similar for `DEBUG`

```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32600,
        "message": "trace_filter is not available on the Free tier - upgrade to Growth, Scale, or Enterprise for access. See available methods at https://docs.alchemy.com/alchemy/documentation/apis"
    }
}
```

I am thinking that maybe `DEBUG` methods end up being too expensive to offer them for Free for Node Operators, and they will want to be able to enable them or disable them on certain deployments (PAID vs FREE)

docs/design/debug-api.md
AlfredoG87: I see it now, maybe I missed it since is only 1 Limit currently. will leave any other comment there.

docs/design/debug-api.md
konstantinabl: Hi Nana, I have explored the possibility of implementing this method, but to my understanding currently it would not be possible. This method checks if an account has changed from block A to block B, meaning in the past, but currently in the mirror node we cannot check an accounts state in previous blocks.
Same goes to debug_getModifiedAccountsByNumber

docs/design/debug-api.md
konstantinabl: Hi Nana, I also explored the possibility of implementing debug_traceTransaction. Unfortunately, it appears that the actions endpoint in the mirrornode ->  /api/v1/contracts/results/{transactionIdOrHash}/actions does not support all opcodes and the information it returns is not enough to return an adequate trace for a given transaction. We would need change in the mirror node in order to make this work

AlfredoG87: If we cannot support `debug_traceTransaction` that means we cannot support `trace_filter` either?

docs/design/debug-api.md
Nana-EC: This is not a priority so should be done after the main debug transaction API is complete. 
In which case just remove this mention from the doc and when we get to it we can always update the design doc.

docs/design/debug-api.md
Nana-EC: There should be a config for this correct. If disabled then we should return a not supported error like we do with other methods.
`debug_traceTransaction` should not be expensive for the callTrace case but will be for opcode flow.
callTrace should be enabled by default and opcode disabled by default per our plan.

docs/design/debug-api.md
Nana-EC: Yes, this is in line with expectations.
Initially, we will only support the callTrace form of `debug_traceTransaction` which should only need the side car actions.
Once the opcode flow of [HIP 801](https://github.com/hashgraph/hedera-improvement-proposal/pull/801) is implemented we'll simply make the call to `/api/v1/contracts/results/{transactionIdOrHash}/opcodes`

I just want the design doc to call out that when
- `tracer = callTracer` then we call `/api/v1/contracts/results/{transactionIdOrHash}/actions` and format the response
- `tracer = opcodeLogger` then we call `/api/v1/contracts/results/{transactionIdOrHash}/opcodes` and format the response

Then it should briefly describe any intricate response formatting needed. If it's just basic number to hex etc there's no need to call that part out in detail

docs/design/debug-api.md
konstantinabl: Well, in here https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debuggetmodifiedaccountsbyhash
It is actually mentioned that a change is  "A change is defined as a difference in nonce, balance, code hash, or storage hash.." , so this complicates things a little bit, if we want to do it in the same way

Nana-EC: Yes, please remove but capture findings and suggestions in a separate ticket for future consideration.

docs/design/debug-api.md
```diff
@@ -0,0 +1,349 @@
...
+
+## Limits
+1. Trying to get modified accounts for more than ex. 100 blocks. Env. variable can be `DEBUG_MODIFIED_ACCOUNTS_LIMIT`
+
+
+## Metric Capturing
+
+Capture metrics for the following:
+
+1. Log every call to all filter API method, as a total amount.
```
Nana-EC, line 305:
```suggestion
1. Log every call to all filter debug method, as a total amount.
```

docs/design/debug-api.md
```diff
@@ -0,0 +1,349 @@
...
+2. Log every success or fail for each new API method, as a total amount.
+
+## Tests
+The following test cases should be covered but additional tests would be welcome.
+
+1. Overall functionality of the methods.
+2. Test different scenarios with all possible parameter combinations.
```
Nana-EC, line 312: Please be more specific.
Also no need to cover UT's here. Just high level acceptance tests to cover the happy paths and failures based in interface values and larger edge cases is sufficient.

docs/design/debug-api.md
konstantinabl: I will remove milestone 3 for now as discussed and create a ticket so we follow we keep track

Nana-EC (approved):

PR #1716, opened 2023-09-04
HBar Operator Fees Consumption is not being tracked on metrics for FileCreate and FileAppend Transactions

### LINKED ISSUE #1701: HBar Operator Fees Consumption is not being tracked on metrics for FileCreate and FileAppend Transactions
### Description

When creating a smart contract using the `JSON RPC` method `eth_sendRawTransaction` the bytecode needs to be uploaded. 
See here:
https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L537

These transactions are expensive and the costs needs to be tracked and showed on metrics, so we know where the operator budget is being spent and plan accordingly.

My recommendation would be to use the same metric that is being currently used for tracking `query` and `transaction` expenses: `rpc_relay_consensusnode_response_sum`

with labels:
 mode=transaction
and type: 
    FileCreateTransaction,
    FileAppendTransaction,

### Steps to reproduce

see logs and compare metrics:

https://cloudlogging.app.goo.gl/WM8a1vWVJZKkLMcS7

### Additional context

_No response_

### Hedera network

### Version

v0.29.1

### Operating system

None

natanasow:
**Description**:

When creating a smart contract using the JSON RPC method eth_sendRawTransaction the bytecode needs to be uploaded.
See here:
https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L537

These transactions are expensive and the costs needs to be tracked and showed on metrics, so we know where the operator budget is being spent and plan accordingly.

My recommendation would be to use the same metric that is being currently used for tracking query and transaction expenses: rpc_relay_consensusnode_response_sum

with labels:
mode=transaction
and type:
FileCreateTransaction,
FileAppendTransaction,

**Related issue(s)**:

Fixes #1701 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -19,37 +19,38 @@
  */
 
 import {
-    AccountBalance,
-    AccountBalanceQuery,
-    AccountId, AccountInfoQuery,
-    Client,
-    ContractByteCodeQuery,
-    ContractCallQuery,
-    ExchangeRates,
-    FeeSchedules,
-    FileContentsQuery,
-    ContractId,
-    ContractFunctionResult,
-    TransactionResponse,
-    AccountInfo,
-    HbarUnit,
-    TransactionId,
-    FeeComponents,
-    Query,
-    Transaction,
-    TransactionRecord,
-    Status,
-    FileCreateTransaction,
-    FileAppendTransaction,
-    FileInfoQuery,
-    EthereumTransaction,
-    EthereumTransactionData,
-    PrecheckStatusError,
-    TransactionRecordQuery,
-    Hbar,
+  AccountBalance,
+  AccountBalanceQuery,
+  AccountId,
+  AccountInfoQuery,
+  Client,
+  ContractByteCodeQuery,
+  ContractCallQuery,
+  ExchangeRates,
+  FeeSchedules,
+  FileContentsQuery,
+  ContractId,
+  ContractFunctionResult,
+  TransactionResponse,
+  AccountInfo,
+  HbarUnit,
+  TransactionId,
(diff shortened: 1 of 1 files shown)


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -61,523 +62,621 @@ const _ = require('lodash');
...
+
+    if (fileId && callData.length > 4096) {
+      const fileAppendTx = await new FileAppendTransaction()
+        .setFileId(fileId)
+        .setContents(hexedCallData.substring(4096, hexedCallData.length))
+        .setChunkSize(4096)
+        .setMaxChunks(this.maxChunks);
+      const fileAppendTxResponse = await fileAppendTx.execute(client);
+
+      const fileAppendRecord = await fileAppendTxResponse.getRecord(this.clientMain);
```
Nana-EC, line 658: I don't think this is sufficient.
If you see from https://github.com/hashgraph/hedera-sdk-js/blob/develop/src/file/FileAppendTransaction.js#L367 the SDK submits multiple individual transactions, however it only returns the 1st.
As such you'll inly get the response of the 1st.

I think we have at least 2 options
1. Perform the separate FileAppends ourselves
2. Query all the transaction records based on chunk size. Each transaction will be separated by a non second so you have predict al the transactionIds

Nana-EC (changes requested):
Potential for missed costs

packages/relay/src/lib/clients/sdkClient.ts
natanasow: Got it, fixed.

PR #1720, opened 2023-09-05
Applied test fixes.

### LINKED ISSUE #1514: When setting the ETH_CALL_DEFAULT_TO_CONSENSUS_NODE=false the allowance on ERC20 wrapped HTS tokens is not getting updated.
### Description

The test ```'@release when the spender has unlimited allowance``` , ```'decreases the spender allowance'``` fails with the following:
```
       Acceptance tests
         @erc20 Acceptance Tests
           HTS token
             should behave like erc20
               transfer from
                 when the token owner is not the zero address
                   when the recipient is not the zero address
                     @release when the spender has unlimited allowance
                       decreases the spender allowance:

      AssertionError: expected '10000' to equal '9999'
      + expected - actual

      -10000
      +9999
``` when the ETH_CALL_DEFAULT_TO_CONSENSUS_NODE=false.  

### Steps to reproduce

1. Run the release acceptance tests pointed to testnet.
```
npm run acceptancetest:release
```

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.27.0

### Operating system

macOS

### PR DESCRIPTION
**Pull request by ebadiere**
Fixes eth_estimateGas.  

Fixes #1514

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance/rpc_batch2.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch2.spec.ts
@@ -40,741 +40,952 @@ import Address from '../../tests/helpers/constants';
 import { numberTo0x } from '../../../../packages/relay/src/formatters';
 
 describe('@api-batch-2 RPC Server Acceptance Tests', function () {
-    this.timeout(240 * 1000); // 240 seconds
+  this.timeout(240 * 1000); // 240 seconds
 
-    const accounts: AliasAccount[] = [];
+  const accounts: AliasAccount[] = [];
 
-    // @ts-ignore
-    const { servicesNode, mirrorNode, relay, logger } = global;
+  // @ts-ignore
+  const { servicesNode, mirrorNode, relay, logger } = global;
 
-    // cached entities
-    let tokenId;
-    let contractId;
-    let contractExecuteTimestamp;
-    let mirrorContract;
-    let mirrorContractDetails;
-    let mirrorSecondaryAccount;
-    let requestId;
+  // cached entities
+  let tokenId;
+  let contractId;
+  let contractExecuteTimestamp;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/server/tests/acceptance/rpc_batch2.spec.ts [hunk 1]
```diff
@@ -40,741 +40,952 @@ import Address from '../../tests/helpers/constants';
...
 
-    this.beforeEach(async () => {
-        requestId = Utils.generateRequestId();
+    it('@release should execute "eth_estimateGas" for contract call', async function () {
+      const expectedRes = `0x${(400000).toString(16)}`;
+      const estimatedGas = await relay.call(
+        RelayCalls.ETH_ENDPOINTS.ETH_ESTIMATE_GAS,
+        [
+          {
+            from: accounts[0].address,
```
Nana-EC, line 146: Good addition with `from` but why remove the `to` when the `data` field is for a function?
The test description doesn't seem to suggest that it's intentionally making a call with a missing to.

Nana-EC (changes requested):
Missing `to` field on test

packages/server/tests/acceptance/rpc_batch2.spec.ts
ebadiere: Added the `to` back and the value returned is correct, however the following hardhat ethers call fails which maybe a separate issue:
```
 const estimatedGas = await ethers.provider.estimateGas({
            to: basicContract.address,
            data: BASIC_CONTRACT_PING_CALL_DATA,
            from: "0x67d8d32e9bf1a9968a5ff53b87d777aa8ebbee69"
        });
        
```

Nana-EC (approved):

PR #1729, opened 2023-09-08
Usage of Shared Cache for Synthetic transaction receipts

### LINKED ISSUE #1599: Blocks Synthetic transactions sometimes are missing.
### Description

When doing some tests on TheGraph longstanding node against mainnet, I decided to index an HTS token with fairly high activity, so I choose `SAUCER` Token of SaucerSwap.

Address: `0x00000000000000000000000000000000000b2ad5`
startBlock: `38629934`

I choose that starting block for the indexing since it was the first interaction found on `Hashscan` for that token. see [here](https://hashscan.io/mainnet/token/0.0.731861).

Indexing went good until suddenly stopped at block: number: `0x30e597a` or hash: `0xe0d8abbb5811629eac5c1f549b8013394a73c042f752bcf76f277f5b88a5b30c`

When seeing the status of the subgraph, I found a `fatalError` with a message something like this:

** I did not copy the original block error, but then there were many more blocks with same issue, so I started copying some for reference.
```
{
    "message": "failed to process trigger: block #51390342 (0x018f…6c4b), transaction a179e7dd235f25e3df57a6c9346a1b1d806f87a546bec51610d2836aeb954ad2: Found no transaction for event",
    "block": {
    "number": "51390342",
    "hash": "0x018f787384b920cadcd35800363c045898ecad63bacb952b531b367d846c6c4b"
    },
    "handler": null
}
```

But to my surprise, when I fetched the block using **postman** I did not encounter the issue, I was able to find said transaction (synthetic) among the transactions for that block.

So I found that there was a way to `delete` the cache for a given block hash or number using a troubleshooting tool for the `graph-node` called: `graphman`. This tool resides within the same instance where the graph-node is running, so a remote exec was needed to run it, and the command is something like this:

```
graphman --config config.toml chain check-blocks mainnet by-number 51442539
```

or by hash:

```
graphman --config config.toml chain check-blocks mainnet by-hash 0xe0d8abbb5811629eac5c1f549b8013394a73c042f752bcf76f277f5b88a5b30c
```

if inconsistencies are found a message like this will appear and the cache will be deleted forcing the graph-node to refetch the block and resolved the indexing and continuing to index the `subgraph`

```
root@graph-node-85b68c875c-lhm4n:/home# graphman --config config.toml chain check-blocks mainnet by-number 51453737
Aug 01 21:44:54.702 INFO Graph Node version: 0.28.2 (2022-10-11)
Aug 01 21:44:54.703 INFO Reading configuration file `config.toml`
Aug 01 21:44:54.711 INFO Creating transport, capabilities: archive, traces, url: https://mainnet.hashio.io/api, provider: mainnet
Aug 01 21:44:54.872 INFO Connecting to Postgres, weight: 1, conn_pool_size: 3, url: postgresql://graph-node:HIDDEN_PASSWORD@postgres:5432/graph-node, pool: main, shard: primary
Aug 01 21:44:54.881 INFO Pool successfully connected to Postgres, pool: main, shard: primary, component: Store
Aug 01 21:44:54.882 DEBG Using postgres host order [Main], shard: primary, component: Store
block 0xc95a…7c6a diverges from cache:
 {
   transactions: [
+    {
+      blockHash: "0xc95a992922553b0cccf92ee677426442f9a2410258eb8e67c41d86be5bee7c6a"
+      blockNumber: "0x3111f29"
+      from: "0x00000000000000000000000000000000000b2ad5"
+      gas: "0x61a80"
+      gasPrice: "0xfe"
+      hash: "0x1770c0cc82f54b5e400cf4af296ad079876a18ea28850b4b56dc712f75bdef44"
+      input: "0x0000000000000000"
+      maxFeePerGas: "0x0"
+      maxPriorityFeePerGas: "0x0"
+      nonce: "0x0"
+      r: "0x0"
+      s: "0x0"
+      to: "0x00000000000000000000000000000000000b2ad5"
+      transactionIndex: "0x16f"
+      type: "0x2"
+      v: "0x0"
+      value: "0x1234"
+    }
   ]
-  transactionsRoot: "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"
+  transactionsRoot: "0xc95a992922553b0cccf92ee677426442f9a2410258eb8e67c41d86be5bee7c6a"
 }

Deleting block 0xc95a…7c6a from cache.
Done.
```

And after a few minutes you can see that the indexing status of the subgraph is continuing and is no longer stuck.

So this happened around 20 times between blocks: `38629934` and `51532212` (start to end)

I documented the need to `check` the cached block for the following block instances, but there were a few more:

```
51442539
51434203
51403767
51400371
51276784
51278825
51300058
51349773
51370110
51375513
51390342
51393614
```

What I think is happening:

graph-node is requesting logs for a range of blocks for that given address.
logs re returned with reference to the given block.

block is fetched using `eth_getBlockByHash` or `eth_getBlockByNumber` and the expected transaction is not found there, hence failing.

### Steps to reproduce

1. Deploy a graph-node connected to RPC on Hedera Mainnet
2. Deploy a sub-graph with the above parameters
3. issue might appear on random blocks

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.28.0

### Operating system

Other

### LINKED ISSUE #1627: TheGraph Node Indexing Error
### Description

When indexing a sub-graph with synthetic transactions, from time to time the following logs appears on the `graph-node` console.

```
INFO Trying again after block polling failed: Receipt for tx 0x965fc6051ad8302ab908841457e0acbdcc897e2aaf0f9432c028ac2641e85367 unavailable, block was likely uncled (block hash = 0x2cd5b2c2f091d24d1204b1ec59996a20fe6f84bc16d4ceca27f94be4fee08217), provider: mainnet-rpc-0, component: BlockIngestor"

```

The logs are similar but change on `tx hash` and `block hash` however when I do the manual lookup of the transaction receipt: 

```

curl --location 'https://mainnet.hashio.io/api' \
--data '{
    "method": "eth_getTransactionReceipt",
    "params": [
       "0x965fc6051ad8302ab908841457e0acbdcc897e2aaf0f9432c028ac2641e85367"
    ],
    "id": 68,
    "jsonrpc": "2.0"
}'
```

or the block:

```
curl --location 'https://mainnet.hashio.io/api' \
--data '{
    "jsonrpc": "2.0",
    "method": "eth_getBlockByHash",
    "params": [
                "0x2cd5b2c2f091d24d1204b1ec59996a20fe6f84bc16d4ceca27f94be4fee08217",
                true
    ],
    "id": 1
}'
```

Everything seems as expected.

### Steps to reproduce

Run a long term sub-graph for a HTS Token that has movement, in this case I chose SAUCE Token.

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.29.0

### Operating system

None

AlfredoG87:
**Description**:
- Refactored getBlock and eth_getTransactionReceipt to use Shared Cache when working with synthetic receipts for HTS transactions.
- Added Unit Tests where needed.

**Related issue(s)**:

Fixes #1599
Fixes #1627

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1737,7 +1737,7 @@ export class EthImpl implements Eth {
     }
 
     const cacheKeySyntheticLog = `${constants.CACHE_KEY.SYNTHETIC_LOG_TRANSACTION_HASH}${hash}`;
-    const cachedLog = this.cacheService.getSharedWithFallback(
+    const cachedLog = await this.cacheService.getSharedWithFallback(
       cacheKeySyntheticLog,
       EthImpl.ethGetTransactionReceipt,
       requestIdPrefix,
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1737,7 +1737,11 @@ export class EthImpl implements Eth {
     }
 
     const cacheKeySyntheticLog = `${constants.CACHE_KEY.SYNTHETIC_LOG_TRANSACTION_HASH}${hash}`;
-    const cachedLog = this.cacheService.get(cacheKeySyntheticLog, EthImpl.ethGetTransactionReceipt, requestIdPrefix);
+    const cachedLog = this.cacheService.getSharedWithFallback(
+      cacheKeySyntheticLog,
+      EthImpl.ethGetTransactionReceipt,
+      requestIdPrefix,
+    );
     if (cachedLog) {
       const receipt: any = {
         blockHash: cachedLog.blockHash,
@@ -1992,7 +1996,14 @@ export class EthImpl implements Eth {
         transactionArray.push(transaction);
 
         const cacheKey = `${constants.CACHE_KEY.SYNTHETIC_LOG_TRANSACTION_HASH}${log.transactionHash}`;
-        this.cacheService.set(cacheKey, log, EthImpl.ethGetBlockByHash, this.syntheticLogCacheTtl, requestIdPrefix);
+        this.cacheService.set(
+          cacheKey,
+          log,
+          EthImpl.ethGetBlockByHash,
+          this.syntheticLogCacheTtl,
+          requestIdPrefix,
+          true,
+        );
       });
     } else {
       filteredLogs = logs.filter((log) => !transactionArray.includes(log.transactionHash));
@@ -2000,7 +2011,14 @@ export class EthImpl implements Eth {
         transactionArray.push(log.transactionHash);
 
         const cacheKey = `${constants.CACHE_KEY.SYNTHETIC_LOG_TRANSACTION_HASH}${log.transactionHash}`;
-        this.cacheService.set(cacheKey, log, EthImpl.ethGetBlockByHash, this.syntheticLogCacheTtl, requestIdPrefix);
+        this.cacheService.set(
+          cacheKey,
+          log,
+          EthImpl.ethGetBlockByHash,
+          this.syntheticLogCacheTtl,
+          requestIdPrefix,
+          true,
+        );
       });
 
       this.logger.trace(
--- a/packages/relay/src/lib/services/cacheService/cacheService.ts
+++ b/packages/relay/src/lib/services/cacheService/cacheService.ts
@@ -149,6 +149,21 @@ export class CacheService {
   }
 
   /**
+   * If SharedCacheEnabled will use shared, otherwise will fallback to internal cache.
+   * @param {string} key - The cache key.
+   * @param {string} callingMethod - The name of the calling method.
+   * @param {string} [requestIdPrefix] - The optional request ID prefix.
+   * @returns {Promise<any>} A Promise that resolves with the cached value or null if not found.
+   */
+  public getSharedWithFallback(key: string, callingMethod: string, requestIdPrefix?: string): any {
+    if (this.isSharedCacheEnabled) {
+      return this.getAsync(key, callingMethod, requestIdPrefix);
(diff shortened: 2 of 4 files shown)


Nana-EC (approved):

packages/relay/src/lib/eth.ts
```diff
@@ -1737,7 +1737,11 @@ export class EthImpl implements Eth {
     }
 
     const cacheKeySyntheticLog = `${constants.CACHE_KEY.SYNTHETIC_LOG_TRANSACTION_HASH}${hash}`;
-    const cachedLog = this.cacheService.get(cacheKeySyntheticLog, EthImpl.ethGetTransactionReceipt, requestIdPrefix);
+    const cachedLog = await this.cacheService.getSharedWithFallback(
```
georgi-l95, line 1740: This will never get any value from the fallback for the internal cache currently. Because when we use `set` with `shared == true`, and for some reason it fails to actually save the value to the cache, it does not go and saves it in the internal cache.

AlfredoG87: I was under the impression that it does.
https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/services/cacheService/cacheService.ts#L198

You are right for failures inside the try...catch.

But this is just for when the `sharedCache` is not enabled.


```
if (shared && this.isSharedCacheEnabled)
```

for the above condition, `shared` is true, but `this.isSharedCacheEnabled` is false. it will set it on the internal cache.

Issue #1736, opened 2023-09-12
eth_getTransactionReceipt `effectiveGasPrice` field right populate instead of `0x0`


AlfredoG87:
### Problem

when using method `eth_getTransactionReceipt` for synthetic transactions (HTS), Contract Create or Contract Call, the field `effectiveGasPrice` on the response is always set to false instead of the appropriate `gasPrice` at the time of the transaction.

### Solution

gasPrice is fixed at exchange rate and is updated each hour. is possible to see the `priceGas` at the time of the transaction using the mirror node:

**for tx hash:** `0xa8795e383e7ee5b138e61f5da7b0d85664e6feec3f2e5e24385f2ff265c401d3`

**timestamp is:** `1694561311.909203003`

**and gasPrice fee is:** `71` (this will need some conversion to hex and wei format)

```
https://previewnet.mirrornode.hedera.com/api/v1/contracts/results/0xa8795e383e7ee5b138e61f5da7b0d85664e6feec3f2e5e24385f2ff265c401d3?nonce=1

https://previewnet.mirrornode.hedera.com/api/v1/network/fees?order=desc&timestamp=1694561311.909203003

```

**Response:**

```
{
  "fees": [
    {
      "gas": 71,
      "transaction_type": "EthereumTransaction"
    },
    {
      "gas": 71,
      "transaction_type": "ContractCreate"
    },
    {
      "gas": 71,
      "transaction_type": "ContractCall"
    }
  ],
  "timestamp": "1692903147.842862003"
}
```

Nana-EC:
@stefan-stefanooov can this be closed?
If so please link to the right PR and close

PR #1743, opened 2023-09-14
Add actual Gas Price to synthetic transaction receipts

stefan-stefanooov:
**Description**:
This PR modifies **eth_getTransactionReceipt** function in order to add **effectiveGasPrice** for synthetic transactions (HTS), Contract Create or Contract Call.

**Related issue(s)**:  https://github.com/hashgraph/hedera-json-rpc-relay/issues/1736

**Notes for reviewer**:
**Hash** of a test transaction in **testnet**:
0xd624d50e109c51d3765c3ade221797b6003b5b0dd0063c49fad5add8f04ffcc6 

**Checklist**
image


georgi-l95 (dismissed):
LG

packages/relay/src/lib/eth.ts
```diff
@@ -1756,7 +1756,7 @@ export class EthImpl implements Eth {
         blockNumber: cachedLog.blockNumber,
         contractAddress: cachedLog.address,
         cumulativeGasUsed: EthImpl.zeroHex,
-        effectiveGasPrice: EthImpl.zeroHex,
+        effectiveGasPrice: await this.gasPrice(requestIdPrefix),
```
AlfredoG87, line 1759: My understanding is that this is the `current` gasPrice, when we should be returning the `gasPrice` at the time of the transaction.

package.json
```diff
@@ -94,10 +95,10 @@
         "!(packages/**/node_modules/)*.ts": [
             "prettier --write"
         ]
-      },
-      "husky": {
+    },
+    "husky": {
         "hooks": {
-          "pre-commit": "lint-staged"
+            "pre-commit": "lint-staged"
```
AlfredoG87, line 101: why are we adding husky with this PR?

georgi-l95: I think this is just indentation fixing from prettier, nothing new added.

Nana-EC (changes requested):
This isn't the approach we want.
As the ticket notes we need to pick up the gasPrice as the time of the transaction.
This is trickier than just taking the current gasPrice.
Happy to sync on this but we should take a different approach of put it back in the backlog and address other tickets

packages/relay/src/lib/eth.ts
```diff
@@ -1750,13 +1750,21 @@ export class EthImpl implements Eth {
       EthImpl.ethGetTransactionReceipt,
       requestIdPrefix,
     );
+
     if (cachedLog) {
+      const block = await this.getBlockByHash(cachedLog.blockHash, false);
+      const gasPriceForTimestamp = await this.getFeeWeibars(
```
Nana-EC, line 1756: Please add an acceptance test case under `rpc_batch1.spec.ts` that submits an HTS transfer transaction and verifies the transactionReceipt has the expected value

Nana-EC (changes requested):
Nice first addition, please add an acceptance test for an E2E consideration

packages/relay/src/lib/eth.ts
```diff
@@ -1750,13 +1750,21 @@ export class EthImpl implements Eth {
...
+        block?.timestamp,
+      );
+
       const receipt: any = {
         blockHash: cachedLog.blockHash,
         blockNumber: cachedLog.blockNumber,
         contractAddress: cachedLog.address,
         cumulativeGasUsed: EthImpl.zeroHex,
-        effectiveGasPrice: EthImpl.zeroHex,
+        effectiveGasPrice: numberTo0x(gasPriceForTimestamp),
```
Nana-EC, line 1767: If it's not too much scope creep please consider fixing this for the setting of `effectiveGasPrice` further down should it be null after https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1794

Then similar ask, add an acceptance tests that checks a submitted ContractCall that executes a transaction and you confirm the `effectiveGasPrice` is correct.
This one should be easier as you can use the details from [mirrorContractDetails.hash](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/server/tests/acceptance/rpc_batch1.spec.ts#L83) to make the query

Nana-EC (changes requested):
Option to fix it all in one go and cover non EthereumTransaction flows also

.env.example
```diff
@@ -4,4 +4,21 @@ MIRROR_NODE_URL=
 HEDERA_NETWORK=
 #mandatory env variables to support queries and transactions to Consensus Node e.g. eth_sendRaTransaction and fallback scenarios on calls to Mirror Node
 OPERATOR_ID_MAIN=<AccountID e.g. 0.1.2, can be obtained from portal.hedera.com, wallet or exchange of choice>
-OPERATOR_KEY_MAIN=<DER Encoded Private Key, can be obtained from portal.hedera.com, wallet or exchange of choice>
\ No newline at end of file
+OPERATOR_KEY_MAIN=<DER Encoded Private Key, can be obtained from portal.hedera.com, wallet or exchange of choice>
```
Nana-EC, line 7: Should revert this change

Nana-EC (changes requested):
Exciting change.
Please resolve DCO change and revert `.env` change so we can get this in :)

stefan-stefanooov:
@Nana-EC 
Hello :) 
I have added the two tests you asked for. 
Please have a look and comment on them.
Hopefully all your recommendations are addressed.

P.S.
(the previous push was just a test one. I am sorry for the spam)

ebadiere (approved):
There are still test failures.

ebadiere (changes requested):
There are still test failures.

packages/server/tests/localAcceptance.env
```diff
@@ -14,3 +14,4 @@ HAPI_CLIENT_TRANSACTION_RESET=50
 HAPI_CLIENT_DURATION_RESET=3600000
 HAPI_CLIENT_ERROR_RESET=[50]
 FILTER_API_ENABLED=true
+ETH_POPULATE_SYNTHETIC_CONTRACT_RESULTS=true
```
Nana-EC, line 17: nit: is this not the default? If so why is this needed?

Nana-EC (approved):

packages/server/tests/localAcceptance.env
stefan-stefanooov: I think the default is false. At least that is what is in `~/hedera-json-rpc-relay/docs/configuration.md` And since you asked. I saw in `eth.ts` something strange.

>   private readonly shouldPopulateSyntheticContractResults = process.env.ETH_POPULATE_SYNTHETIC_CONTRACT_RESULTS
    ? process.env.ETH_POPULATE_SYNTHETIC_CONTRACT_RESULTS === 'true'
    : false;
    
Do you think it's a mistake ?

georgi-l95: It's not a mistake, we've decided to keep it false by default and configure it depending on needs, until it's clear that it won't cause any troubles for normal ethereum receipts.

PR #1744, opened 2023-09-14
TheGraph environment variables recommended values for HashIO Relay

### LINKED ISSUE #700: List all graph node env variables for hedera
### Problem

Hedera isn't supported by The Graph's hosted service, so developers looking to deploy their subgraphs will have to setup their own graph node. The `subgraph-example` doesn't list the graph node's [env variables](https://github.com/graphprotocol/graph-node/blob/master/docs/environment-variables.md) optimised for hedera. It would be helpful to anyone looking to set up a graph node to use that list of env variables instead of using their own guesses or the default values.

For example `ETHEREUM_POLLING_INTERVAL` defaults to 500ms, however looking at [hashscan's blocks](https://hashscan.io/testnet/blocks) the block time seems to be about 2000ms and looking at [hedera metrics](https://hedera.com/#metrics) the time to finality is ~5.45s so perhaps a more suitable `ETHEREUM_POLLING_INTERVAL` is between 6_000 - 10_000 ms.

Another example would be the `GRAPH_ETHEREUM_BLOCK_INGESTOR_MAX_CONCURRENT_JSON_RPC_CALLS_FOR_TXN_RECEIPTS` variable which defaults to 1_000, however, should probably be based on the relay service limits which in the case of Arkhia's Enterprise tier should be 200.

### Solution

- List all the graph node's [environment variables](https://github.com/graphprotocol/graph-node/blob/master/docs/environment-variables.md) optimised for Hedera in the `subgraph-example`

### Alternatives

_No response_

AlfredoG87:
**Description**:
Overrides some GraphNode `env` configurations to optimize for HashIO Hedera JSON RPC Relay.

**Related issue(s)**:

Fixes #700

**Notes for reviewer**:

**Checklist**


Nana-EC (approved):
LG

PR #1745, opened 2023-09-14
Add `debug_traceTransaction` endpoint

### LINKED ISSUE #1477: Add base debug endpoint structure in server
### Problem

As a developer I'd like to be able to use debug endpoints.

### Solution

Add base structure for acceptance test for debug operations, as well as needed server changes to support this new endpoint.

### Alternatives

_No response_

### LINKED ISSUE #1478: Add base debug endpoint structure in relay
### Problem

As a developer I'd like to be able to use debug endpoints.

### Solution

Add new base class structure for debug, similar to eth. Also base structure for unit tests.

### Alternatives

_No response_

konstantinabl:
**Description**:

This PR adds debug_traceTransaction endpoint to the server, as well as DebugService in the relay to support the server endpoint.

* Adds new endpoint to server
* Adds new interface and service implementing it to relay
* Adds unit tests

**Related issue(s)**:
Fixes #1477 
Fixes #1478 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/services/debugService/IDebugService.ts
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+export interface IDebugService {
+  debug_traceTransaction: (
+    transactionHash: string,
+    tracer: string,
```
AlfredoG87, line 24: would not make sense to change `tracer` type, from string to enum?

with the following 2 options:
1. callTracer
2. prestateTracer

https://www.quicknode.com/docs/ethereum/debug_traceTransaction

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /**
+   * Checks if the Filter API is enabled
+   */
+  static requireDebugAPIEnabled(): void {
+    if (!process.env.DEBUG_API_ENABLED || process.env.DEBUG_API_ENABLED !== 'true') {
+      throw predefined.UNSUPPORTED_METHOD;
+    }
+  }
```
AlfredoG87, line 44: I believe is better to read the contents of the `ENV` variable once and have something like this:

Bit more readable and condition is already processed.
```js
 /**
   * Checks if the Filter API is enabled
   * @public
   */
  public static readonly isDebugAPIEnabled = process.env.DEBUG_API_ENABLED === 'true' || false;
  /**
   * Checks if the Filter API is enabled
   */
  static requireDebugAPIEnabled(): void {
    if (!this.isDebugAPIEnabled) {
      throw predefined.UNSUPPORTED_METHOD;
    }
  }
```

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,118 @@
...
+    requestIdPrefix?: string
+  ): Promise<any> {
+    this.logger.trace(`${requestIdPrefix} debug_traceTransaction(${transactionHash})`);
+    try {
+      DebugService.requireDebugAPIEnabled();
+      if (tracer.tracer === 'callTracer') {
+        return await this.callTracer(transactionHash, tracerConfig, requestIdPrefix);
+      } else if (tracer.tracer === 'opcodeLogger') {
+        throw Error('opcodeLogger is currently not supported');
+      }
```
AlfredoG87, line 59: do we support `prestateTracer` ?

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,118 @@
...
+      call_result: output,
+      error_message: error
+    } = transactionsResponse;
+    return {
+      type,
+      from,
+      to,
+      value: amount === 0 ? '0x0' : amount,
+      gas,
+      gasUsed,
```
AlfredoG87, line 110: I believe that `value`, `gas` and `gasUsed` need to be converted to HEX always, and not only when is `0`.

see mirror-node docs:
https://testnet.mirrornode.hedera.com/api/v1/docs/#/contracts/getContractActionsByTransactionIdOrHash

AlfredoG87 (changes requested):
Overall looks good, and thank you for going overr the linting issues, however is best to divide the PRs for linting and styles fix or improvements from the feature  so to have smaller, focused PRs that are easier to review.

Also, TestCases are missing, will circle back when added (I understand is still in draft 👍 )

packages/relay/src/lib/services/debugService/index.ts
konstantinabl: it has not been included in the design discussion, so I think no? The plan is only for callTracer and opcodeLogger

packages/relay/src/lib/services/debugService/index.ts
konstantinabl: You are right, I will fix it

konstantinabl:
> Overall looks good, and thank you for going overr the linting issues, however is best to divide the PRs for linting and styles fix or improvements from the feature so to have smaller, focused PRs that are easier to review.
> 
> Also, TestCases are missing, will circle back when added (I understand is still in draft 👍 )

@AlfredoG87 So should I remove them from the PR?

packages/relay/src/lib/services/debugService/IDebugService.ts
konstantinabl: Done :)

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,219 @@
+import type { Logger } from 'pino';
```
Nana-EC, line 1: missing license

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,219 @@
...
+    tracerConfig: object,
+    requestIdPrefix?: string,
+  ): Promise<any> {
+    this.logger.trace(`${requestIdPrefix} debug_traceTransaction(${transactionHash})`);
+    try {
+      DebugService.requireDebugAPIEnabled();
+      if (tracer === TracerType.CallTracer) {
+        return await this.callTracer(transactionHash, tracerConfig, requestIdPrefix);
+      } else if (tracer === TracerType.OpcodeLogger) {
+        throw Error('opcodeLogger is currently not supported');
```
Nana-EC, line 71: Use `predefined.UNSUPPORTED_METHOD`

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,219 @@
...
+            action.to,
+            [constants.TYPE_CONTRACT, constants.TYPE_TOKEN, constants.TYPE_ACCOUNT],
+            requestIdPrefix,
+          ),
+        ]);
+
+        // The actions endpoint does not return input and output for the calls so we get them from another endpoint
+        // The first one is excluded because we take its input and output from the contracts/results/{transactionIdOrHash} endpoint
+        const getContract =
+          index !== 0 ? await this.mirrorNodeClient.getContract(action.to, requestIdPrefix) : undefined;
```
Nana-EC, line 105: Q: should we be checking for CREATE option type to know if this is applicable?

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,219 @@
...
+  /**
+   * Returns an address' evm equivalence.
+   *
+   * @async
+   * @param {string} address - The address to be resolved.
+   * @param {[string]} types - The possible types of the address.
+   * @param {string} requestIdPrefix - The request prefix id.
+   * @returns {Promise<string>} The address returned as an EVM address.
+   */
+  async resolveAddress(
```
Nana-EC, line 130: Isn't this duplicated logic? I thought we had a resolveAddress somewhere else that we should reuse

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,219 @@
...
+        amount,
+        gas_limit: gas,
+        gas_used: gasUsed,
+        function_parameters: input,
+        call_result: output,
+        error_message: error,
+        result,
+      } = transactionsResponse;
+
+      const [fromToEvmAddress, toToEvmAddress] = await Promise.all([
```
Nana-EC, line 184: nit: move the `from, to -> fromEvmAddress, toEvmAddress` in to a shared method.
It's used here and above

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,219 @@
...
+          requestIdPrefix,
+        ),
+        this.resolveAddress(
+          to,
+          [constants.TYPE_CONTRACT, constants.TYPE_TOKEN, constants.TYPE_ACCOUNT],
+          requestIdPrefix,
+        ),
+      ]);
+
+      const value = amount === 0 ? '0x0' : numberTo0x(amount);
```
Nana-EC, line 197: nit: Use constants for `0x0` and `SUCCESS`. I think these are already defined in places

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,219 @@
...
+        to: toToEvmAddress,
+        value,
+        gas: numberTo0x(gas),
+        gasUsed: numberTo0x(gasUsed),
+        input,
+        output: result !== 'SUCCESS' ? error : output,
+        ...(result !== 'SUCCESS' && { error: errorResult }),
+        ...(result !== 'SUCCESS' && { revertReason: decodeErrorMessage(error) }),
+        // if we have two actions or more executed during the transaction the first one is returned in the top and the other ones in
+        // a calls array, although both are returned form the actions endpoint
```
Nana-EC, line 212:
```suggestion
        // a sub-calls array, although both are returned form the actions endpoint
```

Nana-EC, line 212: Q: What do you mean by `although both are returned form the actions endpoint`?

Nana-EC (changes requested):
Exciting
Left a few comments and questions

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  /**
+   * Checks if the Debug API is enabled
+   * @public
+   */
+  public static readonly isDebugAPIEnabled = process.env.DEBUG_API_ENABLED;
+
+  static requireDebugAPIEnabled(): void {
+    if (!process.env.DEBUG_API_ENABLED || process.env.DEBUG_API_ENABLED !== 'true') {
```
georgi-l95, line 40: Add this new env. variable in the `localAcceptance.env`, readme, docs and helm charts.

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,219 @@
...
+   * The logger used for logging all output from this class.
+   * @private
+   */
+  private readonly logger: Logger;
+
+  private readonly common: CommonService;
+
+  public readonly debugTraceTransaction = 'debug_traceTransaction';
+
+  constructor(mirrorNodeClient: MirrorNodeClient, logger: Logger, common: CommonService) {
```
georgi-l95, line 27: nit: JSDoc missing on constructor and class, explaining the class function.

packages/server/src/koaJsonRpc/lib/methodConfiguration.ts
```diff
@@ -171,4 +171,7 @@ export const methodConfiguration = {
   eth_getFilterLogs: {
     total: tier2rateLimit,
   },
+  debug_traceTransaction: {
+    total: tier2rateLimit,
```
georgi-l95, line 175: Probably this should be tier1 as it's more heavier task making several mirror-node requests.

georgi-l95 (changes requested):
Nice, left some nits

packages/relay/src/lib/services/debugService/index.ts
konstantinabl: Yes, you are correct actually :)

packages/relay/src/lib/services/debugService/index.ts
konstantinabl: Yes, it is and I know this is not the best solution, however the debugService is initialized in the EthImpl class, where the function resolveEvmAddress is and I cannot use it, it will cause circular dependency. I am open to suggestion on how to fix it. :)

packages/relay/src/lib/services/debugService/index.ts
konstantinabl: I have made it more clear now, but what i mean is that in the actions endpoint we return all calls executed during the transaction, however in our response in the sub-calls array we need to return all but the first one

packages/relay/src/lib/services/debugService/index.ts
Nana-EC: Okay, no biggie for now. Something to revisit

charts/hedera-json-rpc-relay/values.yaml
```diff
@@ -50,6 +50,8 @@ config:
   REDIS_ENABLED: false
   REDIS_URL: ''
   REDIS_RECONNECT_DELAY_MS: 1000
+  DEBUG_API_ENABLED: false
+  FILTER_API_ENABLED: false
```
Nana-EC, line 54: Why is filter false by default?

packages/server/tests/localAcceptance.env
```diff
@@ -15,3 +15,4 @@ HAPI_CLIENT_DURATION_RESET=3600000
 HAPI_CLIENT_ERROR_RESET=[50]
 FILTER_API_ENABLED=true
 ETH_POPULATE_SYNTHETIC_CONTRACT_RESULTS=true
+DEBUG_API_ENABLED=false
```
Nana-EC, line 18: Wouldn't we want this true, especially for acceptance test?

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -909,6 +918,13 @@ export class MirrorNodeClient {
     ).replace(MirrorNodeClient.TIMESTAMP_PLACEHOLDER, timestamp);
   }
 
+  private getContractResultsActionsByTransactionId(transactionId: string) {
```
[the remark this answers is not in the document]
Nana-EC: CHange wasn't pushed

Nana-EC (changes requested):
A few more improvements. Thanks

packages/server/tests/localAcceptance.env
konstantinabl: @Nana-EC  well, since the acceptance tests are not yet merged with this PR I thought at first we would want this as false
Should I set it to true here

charts/hedera-json-rpc-relay/values.yaml
konstantinabl: Shouldnt be :) fixed

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,261 @@
...
+    tracerConfig: object,
+    requestIdPrefix?: string,
+  ): Promise<any> {
+    this.logger.trace(`${requestIdPrefix} debug_traceTransaction(${transactionHash})`);
+    try {
+      DebugService.requireDebugAPIEnabled();
+      if (tracer === TracerType.CallTracer) {
+        return await this.callTracer(transactionHash, tracerConfig, requestIdPrefix);
+      } else if (tracer === TracerType.OpcodeLogger) {
+        throw predefined.UNSUPPORTED_METHOD;
```
Nana-EC, line 111: nit: please add a message to make it clear it's not supported for callTracer types

packages/server/tests/localAcceptance.env
Nana-EC: Yeah, set it to true for local flows.
Also theres a bad merge in here on lines 18, 20 and 22

Nana-EC (commented):
Looking good, 2 minor items and i'm good to sign off

packages/relay/src/lib/services/debugService/index.ts
konstantinabl: @Nana-EC You mean its not supported for opcode logger or?

Nana-EC: AH yes I meant opcode.
Basically as coded it'll be a general error message. 
I'm suggesting it might be good to add a Text highlighting it'd just this type (opcode) of tracing that isn't supported.

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,261 @@
...
+        const { resolvedFrom, resolvedTo } = await this.resolveMultipleAddresses(
+          action.from,
+          action.to,
+          requestIdPrefix,
+        );
+
+        // The actions endpoint does not return input and output for the calls so we get them from another endpoint
+        // The first one is excluded because we take its input and output from the contracts/results/{transactionIdOrHash} endpoint
+        const getContract =
+          index !== 0 && action.call_operation_type === 'CREATE'
```
AlfredoG87, line 138: would it be valuable to create an Enum with the possible values of this `action.call_operation_type` and compare against the Enum?

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,261 @@
...
+   */
+  async resolveAddress(
+    address: string,
+    types = [constants.TYPE_CONTRACT, constants.TYPE_TOKEN, constants.TYPE_ACCOUNT],
+    requestIdPrefix?: string,
+  ): Promise<string> {
+    const entity = await this.mirrorNodeClient.resolveEntityType(
+      address,
+      types,
+      'debug_traceTransaction',
```
AlfredoG87, line 173: could this be a constant similar to `eth.ts` endpoints/methods?  

https://github.com/hashgraph/hedera-json-rpc-relay/blob/5e38be160980685074b2574950e8e34f89bf0577/packages/relay/src/lib/eth.ts#L96C6-L110

AlfredoG87 (commented):
Overall goods look, just a couple nits.

packages/relay/src/lib/services/debugService/index.ts
konstantinabl: Well, posibly yes, I will add it

packages/relay/src/lib/services/debugService/index.ts
konstantinabl: Yes :)

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -265,4 +265,10 @@ export const predefined = {
     code: -32001,
     message: 'Filter not found',
   }),
+  UNSUPPORTED_TRACER: (type: string) =>
```
Nana-EC, line 268: No need to add this.
You could us `UNSUPPORTED_OPERATION`

packages/relay/src/lib/services/debugService/index.ts
```diff
@@ -0,0 +1,259 @@
...
+    tracerConfig: object,
+    requestIdPrefix?: string,
+  ): Promise<any> {
+    this.logger.trace(`${requestIdPrefix} debug_traceTransaction(${transactionHash})`);
+    try {
+      DebugService.requireDebugAPIEnabled();
+      if (tracer === TracerType.CallTracer) {
+        return await this.callTracer(transactionHash, tracerConfig, requestIdPrefix);
+      } else if (tracer === TracerType.OpcodeLogger) {
+        throw predefined.UNSUPPORTED_TRACER(TracerType.OpcodeLogger);
```
Nana-EC, line 109:
```suggestion
        throw predefined.UNSUPPORTED_OPERATION(`${TracerType.OpcodeLogger} is not supported on eth_debugTransaction`);
```

AlfredoG87 (approved):
LGTM 👍

georgi-l95 (approved):
Nice

PR #1748, opened 2023-09-18
Adds property to prettierrc to prevent trailing commas

konstantinabl:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
     "singleQuote": true,
     "semi": true,
-    "printWidth": 120
+    "printWidth": 120,
+    "trailingComma": "none"
 }


.prettierrc
```diff
@@ -1,5 +1,6 @@
 {
     "singleQuote": true,
     "semi": true,
-    "printWidth": 120
+    "printWidth": 120,
+    "trailingComma": "none"
```
Nana-EC, line 5: Q: why the change?
I prefer the trailing comma for the reason that most is that when it comes to code reviews it reduces the consideration of what's new.

Case in point your current change here required a comma on `printWidth` just to add `trailingComma`.
Hade there been a comma here this PR would have been a 1 liner.
When you apply this to multiple files and changes you see how PR reviews take a little lonegr.
It's certainly a styling item but I prefer it as s with the default `all` setting.
Is there a reason we're changing it?

Nana-EC (changes requested):
Let's discuss. I prefer the way it is now

konstantinabl:
Better to have trailing commas

PR #1756, opened 2023-09-20
Fix  Flaky Subgraph Acceptance Test - Adding a sleep to wait for the subgraph to index before running tests

### LINKED ISSUE #1755: Flaky Subgraph Acceptance Test
### Description

Since this [PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1744) `Subgraph Tests` have been flaky and a couple of retries are needed for it to pass most of the time.

### Steps to reproduce

Do any PR and see the CI Test result.

### Additional context

_No response_

### Hedera network

other

### Version

v0.33.0-SNAPSHOT

### Operating system

None

AlfredoG87:
**Description**:
Fixing Flacky SubGraph Test

**Related issue(s)**:

Fixes #1755 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/subgraph.yml
+++ b/.github/workflows/subgraph.yml
@@ -56,5 +56,9 @@ jobs:
       - name: Deploy subgraph
         run: npm run deploy-local -- --network local --version-label 0.0.1
 
+      - name: Sleep for 30 seconds
+        run: sleep 30s
+        shell: bash
+
       - name: Run the tests
         run: npx hardhat test
--- a/.github/workflows/subgraph.yml
+++ b/.github/workflows/subgraph.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Deploy subgraph
         run: npm run deploy-local -- --network local --version-label 0.0.1
 
-      - name: Sleep for 10 seconds
+      - name: Sleep for 10 seconds (allow time to index past events following new sub-graph deployment)
         run: sleep 10s
         shell: bash
 
--- a/.github/workflows/subgraph.yml
+++ b/.github/workflows/subgraph.yml
@@ -56,8 +56,8 @@ jobs:
       - name: Deploy subgraph
         run: npm run deploy-local -- --network local --version-label 0.0.1
 
-      - name: Sleep for 30 seconds
-        run: sleep 30s
+      - name: Sleep for 10 seconds
+        run: sleep 10s
         shell: bash
 
       - name: Run the tests


AlfredoG87:
It passed the test at first try:
image

ebadiere (commented):
Is there a way to confirm that the subgraph is up and running?  Perhaps a URL or liveness endpoint that can be hit?  I believe sleeping for 10 seconds will work for now, and may continue to work, but often as systems mature and grow they can take longer to start, and the assumption here is that 10 seconds is a constant real time expectation for the deploy to complete.  Perhaps it is?

AlfredoG87:
> Is there a way to confirm that the subgraph is up and running? Perhaps a URL or liveness endpoint that can be hit? I believe sleeping for 10 seconds will work for now, and may continue to work, but often as systems mature and grow they can take longer to start, and the assumption here is that 10 seconds is a constant real time expectation for the deploy to complete. Perhaps it is?

The Service is up and running, the wait is for giving the graph-node some time to index after the deployment of a new sub-graph, and yes, there is an API that can be checked is the indexing status api on port `8030` it uses graphql.

So even if possible, and I like the suggestion, I believe that is simpler and quicker to just wait 10s so to fix the flakiness of the tests sooner rather than later 😅

ebadiere (dismissed):

Nana-EC:
> > Is there a way to confirm that the subgraph is up and running? Perhaps a URL or liveness endpoint that can be hit? I believe sleeping for 10 seconds will work for now, and may continue to work, but often as systems mature and grow they can take longer to start, and the assumption here is that 10 seconds is a constant real time expectation for the deploy to complete. Perhaps it is?
> 
> The Service is up and running, the wait is for giving the graph-node some time to index after the deployment of a new sub-graph, and yes, there is an API that can be checked is the indexing status api on port `8030` it uses graphql.
> 
> So even if possible, and I like the suggestion, I believe that is simpler and quicker to just wait 10s so to fix the flakiness of the tests sooner rather than later 😅

I also like the suggestion. It speaks to readiness vs liveness.
Let's watch this and see if the API query becomes needed in follow up

.github/workflows/subgraph.yml
```diff
@@ -56,5 +56,9 @@ jobs:
       - name: Deploy subgraph
         run: npm run deploy-local -- --network local --version-label 0.0.1
 
+      - name: Sleep for 10 seconds
```
Nana-EC, line 59: nit: so we know why the sleep in code
```suggestion
      - name: Sleep for 10 seconds to allow time to index post new sub-graph deployment
```

Nana-EC (commented):
LG. 
Nit suggestion

.github/workflows/subgraph.yml
AlfredoG87: ✅  Done

Nana-EC (approved):

PR #1761, opened 2023-09-21
Stop treating CONTRACT_REVERT as ERROR

### LINKED ISSUE #1596: Logging Improvement - Reduce false ERROR level logs
### Problem

image

```


[2023-08-02 21:14:01.223 +0000] ERROR (mirror-node/62 on mainnet-hashio-6cb4dc7ccc-vt8nj): [Request ID: d90eadd6-75a7-49b8-bb99-d8527011b03f] [POST] contracts/call 400 status

[2023-08-02 21:14:01.223 +0000] ERROR (rpc-server/62 on mainnet-hashio-6cb4dc7ccc-vt8nj): [Request ID: d90eadd6-75a7-49b8-bb99-d8527011b03f] execution reverted: CONTRACT_REVERT_EXECUTED

[2023-08-02 21:14:01.223 +0000] INFO (rpc-server/62 on mainnet-hashio-6cb4dc7ccc-vt8nj): [Request ID: d90eadd6-75a7-49b8-bb99-d8527011b03f] [POST]: eth_call 400 (JSON RPC ERROR) 210 ms


```

### Solution

Change the 3 logs above.

1. `[POST] contracts/call 400 status` should not be log as `ERROR` since is an expected status, when a contract execution fails (REVERT), not that the call did not went as intended, should be `INFO` or lower level.
2. `execution reverted: CONTRACT_REVERT_EXECUTED` Same as above, should be `INFO` or lower level.
3. `eth_call 400 (JSON RPC ERROR) 210 ms` this is correctly logged with `INFO` level but having the word `ERROR` as part of the test makes it appear on Queries were a relay operator might be looking for actual errors.

### Alternatives

Not a big issue, not a big priority either.

### LINKED ISSUE #1616: Stop treating CONTRACT_REVERT_EXECUTED on `eth_call` as errors
### Problem

Currently the relay is treating `eth_call` `CONTRACT_REVERT_EXECUTED` as errors, when in fact they are an expected result of the logical execution of the data provided on the `eth_call` request data.

Is not the same encountering an error when trying to execute `eth_call` than en error while evaluating the result of a smart contract, so let's consider a smart contract that verifies for permissions, if the sender does not have such permissions is going to REVERT. but that is the actual and expected result of the `eth_call` request, hence is not a Relay error.

Nevertheless we are receiving 3 error logs on the relay for such instances and operators should not concern with those, they are creating noise in the logs and the metrics, as well as confusing developers and users of the relay.

**See the following example:**

[2023-08-08 22:03:55.883 +0000] ERROR (mirror-node/62 on mainnet-hashio-74647fcb96-rh76k): [Request ID: de46711b-c015-4934-9005-d5f5cde4f64b] [POST] contracts/call 400 status 

[2023-08-08 22:03:55.883 +0000] ERROR (rpc-server/62 on mainnet-hashio-74647fcb96-rh76k): [Request ID: de46711b-c015-4934-9005-d5f5cde4f64b] execution reverted: CONTRACT_REVERT_EXECUTED 
  
[2023-08-08 22:03:55.884 +0000] INFO (rpc-server/62 on mainnet-hashio-74647fcb96-rh76k): [Request ID: de46711b-c015-4934-9005-d5f5cde4f64b] [POST]: eth_call 400 (JSON RPC ERROR) 144 ms

### Solution

- Change the LOG LEVEL from ERROR to TRACE or DEBUG, even INFO is OK. 
- On the third log, is correctly logged as INFO, but on the text is `(JSON RPC ERROR)` remove that since there is no ERROR.
- Change the response we return to the client to the appropriate result and status code. most likely it will be `200` status code, verify expected result with Infura or Alchemy in order to see what should be the appropriate response. 

### Alternatives

_No response_

### LINKED ISSUE #1707: Mirror node client error when calling write function
### Description

Recently this error is happening when invoking a write smart contract function. We are using our own node (w/ commit cf48856a9957712d7cdced47e6812663aeaf94d7 latest master) but issue happens in prior version.

The error is very generic too to inspect.

### Steps to reproduce

1. Call a write smart contract function
2. Logs says error 400 from mirrorNodeClient.js

### Additional context

(from own node logs)
image

(client side)
Screen Shot 2023-09-03 at 9 05 27 PM

### Hedera network

testnet

### Version

latest master (cf48856a9957712d7cdced47e6812663aeaf94d7)

### Operating system

None

AlfredoG87:
**Description**:
Stopped treating CONTRACT_REVERT as ERROR, MirrorNode returns 400 status, but is a valid and expected response. returning 200 status code on the relay and logging accordingly to avoid false error positives

Return 200 instead of 400 on Relay HTTP Status Code Response

Removed Error Logs from:
1. mirrorNodeClient.ts
2. server.ts

**Related issue(s)**:

Fixes #1616
Fixes #1596
Fixes #1707 

**Notes for reviewer**:

**Example Response:**
image

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -383,17 +383,22 @@ export class MirrorNodeClient {
 
     if (error.response && acceptedErrorResponses && acceptedErrorResponses.indexOf(effectiveStatusCode) !== -1) {
       this.logger.debug(`${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`);
-      if (pathLabel === MirrorNodeClient.CONTRACT_CALL_ENDPOINT) {
-        this.logger.warn(
-          `${requestIdPrefix} [${method}] ${path} Error details: ( StatusCode: '${effectiveStatusCode}', StatusText: '${
-            error.response.statusText
-          }', Data: '${JSON.stringify(error.response.data)}')`,
-        );
-      }
       return null;
     }
 
-    this.logger.error(new Error(error.message), `${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`);
+    // Contract Call returns 400 for a CONTRACT_REVERT but is a valid response, expected and should not be logged as error:
+    if (pathLabel === MirrorNodeClient.CONTRACT_CALL_ENDPOINT && effectiveStatusCode === 400) {
+      this.logger.debug(
+        `${requestIdPrefix} [${method}] ${path} Contract Revert: ( StatusCode: '${effectiveStatusCode}', StatusText: '${
+          error.response.statusText
+        }', Detail: '${JSON.stringify(error.response.detail)}',Data: '${JSON.stringify(error.response.data)}')`,
+      );
+    } else {
+      this.logger.error(
+        new Error(error.message),
+        `${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`,
+      );
+    }
 
     throw mirrorError;
   }
--- a/packages/server/src/koaJsonRpc/index.ts
+++ b/packages/server/src/koaJsonRpc/index.ts
@@ -49,6 +49,7 @@ const INVALID_PARAMS_ERROR = 'INVALID PARAMS ERROR';
 const INVALID_REQUEST = 'INVALID REQUEST';
 const IP_RATE_LIMIT_EXCEEDED = 'IP RATE LIMIT EXCEEDED';
 const JSON_RPC_ERROR = 'JSON RPC ERROR';
+const CONTRACT_REVERT = 'CONTRACT REVERT';
 const METHOD_NOT_FOUND = 'METHOD NOT FOUND';
 const REQUEST_ID_HEADER_NAME = 'X-Request-Id';
 
@@ -172,8 +173,24 @@ export default class KoaJsonRpc {
 
       ctx.body = jsonResp(body.id, null, result);
       if (result instanceof JsonRpcError) {
-        ctx.status = result.code == -32603 ? 500 : 400;
-        ctx.state.status = `${ctx.status} (${JSON_RPC_ERROR})`;
+        // What Status code to return for JsonRpcError
+        switch (result.code) {
+          // INTERNAL_ERROR
+          case -32603:
+            ctx.status = 500;
+            ctx.state.status = `${ctx.status} (${JSON_RPC_ERROR})`;
+            break;
+          // CONTRACT_REVERT
+          case -32008:
+            ctx.status = 200;
+            ctx.state.status = `${ctx.status} (${CONTRACT_REVERT})`;
+            break;
+          // ANYTHING ELSE
+          default:
+            ctx.status = 400;
+            ctx.state.status = `${ctx.status} (${JSON_RPC_ERROR})`;
+            break;
+        }
       }
     };
   }
--- a/packages/server/src/server.ts
+++ b/packages/server/src/server.ts
@@ -210,7 +210,10 @@ const logAndHandleResponse = async (methodName: any, methodParams: any, methodFu
 
     const response = await methodFunction(requestIdPrefix);
     if (response instanceof JsonRpcError) {
-      logger.error(`${requestIdPrefix} ${response.message}`);
+      // log only if it is not a contract revert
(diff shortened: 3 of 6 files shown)


ebadiere (approved):
LG.

ebadiere (commented):
LG.

Nana-EC (commented):
Question on logging.
Also none of these changes cause the UTs to change?
Does that mean there was coverage missing we should add for `mirrorNodeCLient.ts`, `eth.ts` or https server `index.ts` or `server.ts`

packages/server/src/server.ts
```diff
@@ -210,7 +210,10 @@ const logAndHandleResponse = async (methodName: any, methodParams: any, methodFu
 
     const response = await methodFunction(requestIdPrefix);
     if (response instanceof JsonRpcError) {
-      logger.error(`${requestIdPrefix} ${response.message}`);
+      // log only if it is not a contract revert
+      if (!(response.code === predefined.CONTRACT_REVERT().code)) {
```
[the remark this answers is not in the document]
AlfredoG87: yes, this is preventing logs from `CONTRACT_REVERT` since they are not actual errors.
What we could do, is keep logging `CONTRACT_REVERT` responses here, but as Debug or Trace.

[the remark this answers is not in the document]
Nana-EC: debug trace sounds good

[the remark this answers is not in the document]
AlfredoG87: ✅  done!

AlfredoG87:
> Question on logging. Also none of these changes cause the UTs to change? Does that mean there was coverage missing we should add for `mirrorNodeCLient.ts`, `eth.ts` or https server `index.ts` or `server.ts`

🤔  yeah, I think that the change is actually on `server.ts` let me see it's UTs so we can add the scenario.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -383,17 +383,22 @@ export class MirrorNodeClient {
...
-          `${requestIdPrefix} [${method}] ${path} Error details: ( StatusCode: '${effectiveStatusCode}', StatusText: '${
-            error.response.statusText
-          }', Data: '${JSON.stringify(error.response.data)}')`,
-        );
-      }
       return null;
     }
 
-    this.logger.error(new Error(error.message), `${requestIdPrefix} [${method}] ${path} ${effectiveStatusCode} status`);
+    // Contract Call returns 400 for a CONTRACT_REVERT but is a valid response, expected and should not be logged as error:
```
Nana-EC, line 389: Q: Should we have additional test cases for handleError to confirm this change?

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -172,8 +173,24 @@ export default class KoaJsonRpc {
 
       ctx.body = jsonResp(body.id, null, result);
       if (result instanceof JsonRpcError) {
-        ctx.status = result.code == -32603 ? 500 : 400;
-        ctx.state.status = `${ctx.status} (${JSON_RPC_ERROR})`;
+        // What Status code to return for JsonRpcError
+        switch (result.code) {
```
Nana-EC, line 177: Q: Should we have additional test cases for status code or do we already have this covered?

Nana-EC (changes requested):
LG
Let's confirm if we're missing any test coverage here.

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: we already have coverage of this error handling.

packages/server/src/koaJsonRpc/index.ts
AlfredoG87: ✅  Done.

I first tried to do it through UnitTest for `server.ts` but since the artifact only exports the KOA app, and there is no way to inject dependencies, there was no way for me to mock them, and the current UT for that class is only testing the errors and validations that occur prior to the mirror node calls, and excludes exceptions that might come from the mirror node or relay, since is not possible to mock them or even to spy on them to stub the responses.

That said, I created an acceptance test that instead of using the relayImpl directly to perform the `eth_call` call, it does so using an http testClient that gets the whole http response with its corresponding status code that we can assert.

ebadiere (approved):
LG.

Nana-EC (approved):
LGTM

PR #1774, opened 2023-10-02
Add more tests and refine Intrinsic gas cost calculation

### LINKED ISSUE #1631: Wrong Intrinsic Gas Cost in `eth_sendRawTransaction` precheck

georgi-l95:
**Description**:
This PR is a follow up on https://github.com/hashgraph/hedera-json-rpc-relay/pull/1637
Adding some more tests as well as fixing the precheck test suite, because it was not running at all due to errors.

**Related issue(s)**:

Fixes #1631 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/tests/lib/precheck.spec.ts
+++ b/packages/relay/tests/lib/precheck.spec.ts
@@ -26,12 +26,13 @@ const registry = new Registry();
 import pino from 'pino';
 import { Precheck } from '../../src/lib/precheck';
 import { expectedError, mockData, signTransaction } from '../helpers';
-import { ClientCache, MirrorNodeClient } from '../../src/lib/clients';
+import { MirrorNodeClient } from '../../src/lib/clients';
 import axios from 'axios';
 import MockAdapter from 'axios-mock-adapter';
 import { ethers } from 'ethers';
 import constants from '../../src/lib/constants';
 import { predefined } from '../../src';
+import { CacheService } from '../../src/lib/services/cacheService/cacheService';
 const logger = pino();
 
 const limitOrderPostFix = '?order=desc&limit=1';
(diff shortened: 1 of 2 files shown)


packages/relay/tests/lib/precheck.spec.ts
```diff
@@ -429,33 +431,50 @@ describe('Precheck', async function () {
...
     const contractCall = '0xcfae3217';
     const transfer = '0x';
     const invalidTx = '0x60806040523480156200001157600080fd5b';
-    it('should be able to calculate contract create', function () {
-      const intrinsicGasCost = Precheck.transactionIntrinsicGasCost(contractCreate);
+    it('should be able to calculate small contract create', function () {
+      const mirrorNodeEstimation = 60364;
```
ebadiere, line 442: How did this number for the small contract create come about?

AlfredoG87 (approved):
Overall looks good.

packages/relay/tests/lib/precheck.spec.ts
georgi-l95: Created the contract with Remix, tried to deploy on previewnet and got the number from the mirror-node.

Nana-EC: @georgi-l95 you can add a really simple comment on top of the estimations here and below to show future developers what call to make to the mirror node to get the value of `mirrorNodeEstimation` in case it ever changes
```suggestion
      // curl -X POST --data '{"jsonrpc":"2.0","id":1,"method":"eth_call","params":[{"from":"0x...","data":<greeterContractCreate>},"latest"]}'
      const mirrorNodeEstimation = 60364;
```

Nana-EC (dismissed):
LG
Added a suggestion for a curl comment

Nana-EC (approved):

PR #1788, opened 2023-10-04
Productize ETH_POPULATE_SYNTHETIC_CONTRACT_RESULTS = true

### LINKED ISSUE #1787: Productize ETH_POPULATE_SYNTHETIC_CONTRACT_RESULTS = TRUE
### Problem

We had some reports from users and operators of the Relay that are not able to fetch the `synthetic` logs for HTS transactions, but we were able to do it on our production environments.

So we realized that they were using the recommended and default values for the relay and that the property config `ETH_POPULATE_SYNTHETIC_CONTRACT_RESULTS` was set as `false` as default.

### Solution

Change the default value of config `ETH_POPULATE_SYNTHETIC_CONTRACT_RESULTS` to `true`

### Alternatives

_No response_

AlfredoG87:
**Description**:
productising ETH_POPULATE_SYNTHETIC_CONTRACT_RESULTS = true, as it has been running on production environments from some time without issues.

**Related issue(s)**:

Fixes #1787 

**Notes for reviewer**:

**Checklist**


ebadiere (approved):
LG.

Issue #1416, opened 2023-06-30
Create a proper clientSDK for the mirror node using the openAPI Spec provided


AlfredoG87:
## 🆕🐥 First Timers Only

This issue is reserved for people who have never contributed to [Hedera](https://hedera.com) or any open source project in general.
We know that creating a pull request (PR) is a major barrier for new contributors.
The goal of this issue and all other issues labeled by [**'Good First Issue'**](https://github.com/issues?q=is%3Aopen+is%3Aissue+org%3Ahashgraph+archived%3Afalse+label%3A%22Good+First+Issue%22+) is to help you make your first contribution to Hedera.

## 👾 Description of the issue

Currently we are using the `REST` API of the mirror node without a proper clientSDK for the types, enums, errors, etc. that makes it cumbersome and flaky, there are no standards and code is not ideal. Also, makes checking the api more often for supported fields, objects, errors, and statuses types.

### Proposed Solution:

Generate a clientSDK for the mirror node using the openAPI spec provided for the mirror node 
https://testnet.mirrornode.hedera.com/api/v1/docs/openapi.yml

There are 2 options that are promising:

1. https://openapistack.co/docs/openapi-client-axios/intro/ 
2. Swagger Editor IO
<img width="1846" alt="image" src="https://github.com/hashgraph/hedera-json-rpc-relay/assets/123040664/affcf861-393b-4075-8fe8-ac9422f62aa8">

From the 2 options above, the preferred one if possible is the first one since no generated code needs to be updated with each release of the mirror node, types are created in runtime by the tool.

## 📋 Step by step guide to do a contribution

If you have never contributed to an open source project at GitHub, the following step-by-step guide will introduce you to the workflow. More information and concrete samples for shell commands for each step can be found in our [CONTRIBUTING.md](https://github.com/hashgraph/.github/blob/main/CONTRIBUTING.md) file.
A more detailed general documentation of the GitHub PR workflow can be found [here](https://github.com/firstcontributions/first-contributions/blob/master/README.md).

- [ ] **Claim this issue:** Comment below that you are interested in working on the issue
- [ ] **Wait for assignment:** A community member with the given rights will add you as an assignee of the issue
- [ ] **Fork the repository:** You can do that in GitHub (by simply clicking the 'fork' button).
- [ ] **Check out the forked repository**
- [ ] **Create a feature branch** for the issue. We do not have a hard naming definition for branches but it is best practice to prefix the branch name with the issue id.
[...]

PR #1798, opened 2023-10-06
Fix timestamp from hex to decimal when querying the mirror node;

Neurone:
**Description**:

This PR fixes an error in querying the mirror node with the wrong format for the timestamp parameter.

```sh
[GET] network/fees?timestamp=0x651f99d8 400 status
...
      "type": "Error",
      "message": "Request failed with status code 400",
...
      "message": "Invalid parameter: timestamp",
...
```

In particular, the problem is the relay takes the value from the result of the previous `eth_getBlockByHash` operation, but that value is in hex and it should be converted to decimal for the Mirror Node to be able to use it as a parameter.

**Related issue(s)**:

N/A

**Notes for reviewer**:

This is an example of the error I encountered during some tests:

```
[2023-10-06 05:44:21.868 +0000] TRACE (localLRUCache/48904 on mcfly.local): undefined returning cached value eth_getBlockByHash_undefined_false:{"timestamp":"0x651f99d8","difficulty":"0x0","extraData":"0x","gasLimit":"0xe4e1c0","baseFeePerGas":"0x199c82cc000","gasUsed":"0x138800","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","miner":"0x0000000000000000000000000000000000000000","mixHash":"0x0000000000000000000000000000000000000000000000000000000000000000","nonce":"0x0000000000000000","receiptsRoot":"0x0000000000000000000000000000000000000000000000000000000000000000","sha3Uncles":"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347","size":"0x1c17","stateRoot":"0x0000000000000000000000000000000000000000000000000000000000000000","totalDifficulty":"0x0","transactions":["0x5524e449184e92b48efcb6a10bbd534e091ac609b21a9f3de90e1ab8e7f77abf"],"transactionsRoot":"0x612e61273c89a84afbcc8743cfbbe354126d073a28adbdea274ea65c8718e889","uncles":[],"withdrawals":[],"withdrawalsRoot":"0x0000000000000000000000000000000000000000000000000000000000000000","number":"0x2d0895","hash":"0x612e61273c89a84afbcc8743cfbbe354126d073a28adbdea274ea65c8718e889","parentHash":"0xeca3cc7c24114a01e75fb195b052ddc807f63a7eab1db59e55fc719f148c4a97"} on eth_GetBlockByHash call
[2023-10-06 05:44:21.982 +0000] ERROR (mirror-node/48904 on mcfly.local): undefined [GET] network/fees?timestamp=0x651f99d8 400 status
    err: {
      "type": "Error",
      "message": "Request failed with status code 400",
      "stack":
          Error: Request failed with status code 400
              at MirrorNodeClient.handleError (/Users/giuseppebertone/workspace/hedera/hedera-json-rpc-relay/packages/relay/dist/lib/clients/mirrorNodeClient.js:235:31)
              at MirrorNodeClient.<anonymous> (/Users/giuseppebertone/workspace/hedera/hedera-json-rpc-relay/packages/relay/dist/lib/clients/mirrorNodeClient.js:206:22)
              at rejected (/Users/giuseppebertone/workspace/hedera/hedera-json-rpc-relay/packages/relay/dist/lib/clients/mirrorNodeClient.js:25:65)
    }
[2023-10-06 05:44:21.983 +0000] WARN (relay-eth/48904 on mcfly.local): undefined Mirror Node threw an error retrieving fees. Fallback to network
    err: {
      "message": "Invalid parameter: timestamp",
      "stack":
          Error: Invalid parameter: timestamp
              at MirrorNodeClient.handleError (/Users/giuseppebertone/workspace/hedera/hedera-json-rpc-relay/packages/relay/dist/lib/clients/mirrorNodeClient.js:224:29)
              at MirrorNodeClient.<anonymous> (/Users/giuseppebertone/workspace/hedera/hedera-json-rpc-relay/packages/relay/dist/lib/clients/mirrorNodeClient.js:206:22)
              at rejected (/Users/giuseppebertone/workspace/hedera/hedera-json-rpc-relay/packages/relay/dist/lib/clients/mirrorNodeClient.js:25:65)
      "statusCode": 400
    }
```

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1845,10 +1845,11 @@ export class EthImpl implements Eth {
 
   private async getCurrentGasPriceForBlock(blockHash: string, requestIdPrefix?: string): Promise<string> {
     const block = await this.getBlockByHash(blockHash, false);
+    const timestampDecimal = parseInt(block ? block.timestamp : '0', 16);
     const gasPriceForTimestamp = await this.getFeeWeibars(
       EthImpl.ethGetTransactionReceipt,
       requestIdPrefix,
-      block?.timestamp,
+      timestampDecimal > 0 ? timestampDecimal.toString() : '',
     );
 
     return numberTo0x(gasPriceForTimestamp);


Nana-EC (changes requested):
LG
You'll want to update the tests to pass and add another test or two to `eth.spec.ts` to ensure coverage for this.
Also please update PR description for a bit more clarity on the actual problem.
Basically "What the problem was and how this fixes it"

Neurone:
> LG You'll want to update the tests to pass and add another test or two to `eth.spec.ts` to ensure coverage for this. Also please update PR description for a bit more clarity on the actual problem. Basically "What the problem was and how this fixes it"

Hi Nana, the tests passed on my machine, and GitHub logs show no specific errors, even if it said one test failed.
I didn't change anything and I tried to push on the branch to restart the testing on GitHub again, and now it seems to work for the tests but not for the acceptance tests.

I'll try to add a couple of tests for that problem anyway.

Nana-EC:
> > LG You'll want to update the tests to pass and add another test or two to `eth.spec.ts` to ensure coverage for this. Also please update PR description for a bit more clarity on the actual problem. Basically "What the problem was and how this fixes it"
> 
> Hi Nana, the tests passed on my machine, and GitHub logs show no specific errors, even if it said one test failed. I didn't change anything and I tried to push on the branch to restart the testing on GitHub again, and now it seems to work for the tests but not for some acceptance tests (but they worked last time, so I think the problem is somewere else).
> 
> I'll try to add a couple of tests for that problem anyway.

Yeah seems like a CI failure not your code.
For coverage please look in `eth.spec.ts` under `describe('eth_getTransactionReceipt'`
You should look to add a variation of the `'valid receipt on match'` test to cover your scenario.
Thanks

AlfredoG87 (commented):
Looking good, thanks for the contributionn @Neurone.
Just add a Unit Test that covers this scenario and we are good.
Btw, I re-ran the failed tests on the CI and they passed the second time 👍

Neurone:
Hi @ Nana-EC @ AlfredoG87, I'm scratching my head a bit here to create a meaningful test case because the problem is just a wrong parameter format.

That part of the code is never called during tests because of the presence of mocks, so the bug never emerged. However, the issue relates to sending wrong parameters to the Mirror Node's REST APIs, so I don't consider that a unit test.

A specific unit test goes near to test that function, which is `valid receipt on cache match`, but again I should remove the mocks to test it properly, and it feels wrong proceeding that way.

If we want to test this issue, I would consider an additional integration test in the same place where we check all the parameters, formats, methods, etc., to interact with the Mirror Node.

I think I found in the code checks to test how the relay behaves for messages it receives (i.e., `validator.spec.ts`), but I cannot find where we check the parameters we send.

What do you think?

AlfredoG87:
@Neurone,

I took a look at the code and the test that @Nana-EC  mentioned on `eth.spec.ts` and you are right, the whole method `getCurrentGasPriceForBlock` gets mocked/stub, so the code inside never gets tested.

However we could move the mock a couple levels down. and mock: `getBlockByHash` and `getFeeWeibars` so the code on method `getCurrentGasPriceForBlock` can be tested.

Also, I like your approach of testing it on the `mirrorNodeClient.spec.ts` side. (`getNetworkFees`  method)
but for that we would need to add a check for the `timestamp` parameter that verifies that the format of the parameter is correct, otherwise `throw` a corresponding error.

AlfredoG87:
> > Hi @Nana-EC @AlfredoG87, I'm scratching my head a bit here to create a meaningful test case because the problem is just a wrong parameter format.
> > That part of the code is never called during tests because of the presence of mocks, so the bug never emerged. However, the issue relates to sending wrong parameters to the Mirror Node's REST APIs, so I don't consider that a unit test.
> > A specific unit test goes near to test that function, which is `valid receipt on cache match`, but again I should remove the mocks to test it properly, and it feels wrong proceeding that way.
> > If we want to test this issue, I would consider an additional integration test in the same place where we check all the parameters, formats, methods, etc., to interact with the Mirror Node.
> > In particular, we should check that the [`MirrorNodeClient.getNetworkFees` function](https://github.com/hashgraph/hedera-json-rpc-relay/blob/63953eecc600695a70dd565841719a74b9f66546/packages/relay/src/lib/clients/mirrorNodeClient.ts#L888) in `mirrorNodeClient.ts` uses a correct `timestamp` format, but I don't see checks for other parameters either:
> > ```
> >   public async getNetworkFees(timestamp?: string, order?: string, requestIdPrefix?: string) {
> >     const queryParamObject = {};
> >     this.setQueryParam(queryParamObject, 'timestamp', timestamp);
> >     this.setQueryParam(queryParamObject, 'order', order);
> >     const queryParams = this.getQueryParams(queryParamObject);
> >     return this.get(
> >       `${MirrorNodeClient.GET_NETWORK_FEES_ENDPOINT}${queryParams}`,
> >       MirrorNodeClient.GET_NETWORK_FEES_ENDPOINT,
> >       requestIdPrefix,
> >     );
> >   }
> > ```
> > 
> > 
> > I think I found in the code checks to test how the relay behaves for messages it receives (i.e., `validator.spec.ts`), but I cannot find where we check the parameters we send.
> > What do you think?
> 
> Ah I see. It's more a input issue to the mirror node call. Instead could you use the `restMock` object to verify that when `getNetworkFees` is called it's called with the valid input for your code improvement? Would this be the simplest test in `eth.spec.ts`? It would verify the flow and the concern here.
> 
> @AlfredoG87 i think we should create a ticket nonetheless to explore further validation of inputs for `mirroNodeClient` params

@Nana-EC I created ticket https://github.com/hashgraph/hedera-json-rpc-relay/issues/1818 for keeping track of it. but also, this relates to using a auto-generated client for the mirror node. https://github.com/hashgraph/hedera-json-rpc-relay/issues/1416

Neurone:
Cool! If you agree I suggest to merge this fix - so at least we don't have that error anymore - and to delay the development of specific test cases when we implement those for all parameters, and not just for this single use case.

AlfredoG87:
@Neurone I think is fair to modify the mock and instead of mocking the method `` to mock the methods inside that one.

I've made a quick draft on how I would go on to make the test for this:

```
      /*getCurrentGasPriceForBlockStub = sandbox
        .stub(ethImpl, <any>'getCurrentGasPriceForBlock')
        .resolves('0xad78ebc5ac620000');*/

      let getBlockByHash = sandbox.stub(ethImpl, <any>'getBlockByHash').resolves(defaultBlock);
      let getFeeWeibars = sandbox.stub(ethImpl, <any>'getFeeWeibars').resolves(`12500000000000000000`); // 0xad78ebc5ac620000 in decimal

      // set cache with synthetic log
      const cacheKeySyntheticLog1 = `${constants.CACHE_KEY.SYNTHETIC_LOG_TRANSACTION_HASH}${defaultDetailedContractResultByHash.hash}`;
      const cachedLog = new Log({
        address: defaultLogs1[0].address,
        blockHash: toHash32(defaultLogs1[0].block_hash),
        blockNumber: numberTo0x(defaultLogs1[0].block_number),
        data: defaultLogs1[0].data,
        logIndex: numberTo0x(defaultLogs1[0].index),
        removed: false,
        topics: defaultLogs1[0].topics,
        transactionHash: toHash32(defaultLogs1[0].transaction_hash),
        transactionIndex: nullableNumberTo0x(defaultLogs1[0].transaction_index),
      });

      cacheService.set(cacheKeySyntheticLog1, cachedLog, EthImpl.ethGetTransactionReceipt);

      // w no mirror node requests
      const receipt = await ethImpl.getTransactionReceipt(defaultTxHash);

      // Assert the matching reciept
      expect(receipt.blockHash).to.eq(cachedLog.blockHash);
      expect(receipt.blockNumber).to.eq(cachedLog.blockNumber);
      expect(receipt.contractAddress).to.eq(cachedLog.address);
      expect(receipt.cumulativeGasUsed).to.eq(EthImpl.zeroHex);
      expect(receipt.effectiveGasPrice).to.eq(defaultReceipt.effectiveGasPrice);
      expect(receipt.from).to.eq(EthImpl.zeroAddressHex);
      expect(receipt.gasUsed).to.eq(EthImpl.zeroHex);
      expect(receipt.logs).to.deep.eq([cachedLog]);
      expect(receipt.logsBloom).to.be.eq(EthImpl.emptyBloom);
      expect(receipt.status).to.eq(EthImpl.oneHex);
      expect(receipt.to).to.eq(cachedLog.address);
      expect(receipt.transactionHash).to.eq(cachedLog.transactionHash);
      expect(receipt.transactionIndex).to.eq(cachedLog.transactionIndex);
      expect(receipt.root).to.eq(EthImpl.zeroHex32Byte);
      // expect(getCurrentGasPriceForBlockStub.calledOnce).to.be.true;


      // verify getFeeWeibars was called with the correct format
      expect(getFeeWeibars.calledOnce).to.be.true;
      const expectedFormat = defaultBlock.timestamp; // here is necesary to set the correct format
      expect(getFeeWeibars.calledWith(`eth_GetTransactionReceipt`, undefined, expectedFormat)).to.be.true;

```

Neurone:
Hi @AlfredoG87 the problem is if we mock `getFeeWeibars` the offending code in `getNetworkFees` is never executed, so we are not testing it (i.e., we can change the code in the real `getFeeWeibars` or `getNetworkFees` completely, and the test will pass regardless). And even if we go lower, to `getNetworkFees`, our tests cannot verify we are sending the correct parameters because we don't have anything to test (a _parameter verifier_, a spec to rely on, etc.). And it's not an integration test, so we are not really calling a mirror node to check the return value and compare it with a error/non error response.

I think the correct solution is to go with #1818, but I'm not expert enough to develop a good starting point for it.

Neurone:
Closing in favor of #1836

Issue #1802, opened 2023-10-09
JsonRpcProvider failed to detect network and cannot start up


iron4548:
### Description

```typescript

import * as ethers from 'ethers'; //ethers.js 6.7.3

async function main() {
  try {
    const networkId = 'testnet';
    const provider = new ethers.JsonRpcProvider(`https://${networkId}.hashio.io/api`);
        
    const network = await provider.getNetwork();
    console.log(network);

    const blockNumber = await provider.getBlockNumber();
    console.log(blockNumber);

  }
  catch (err:unknown) {
    console.warn(err);
  }
}
main();
```

For console.log(network) I got:

`Network {#name: 'unknown', #chainId: 296n, #plugins: Map(0)}`

The chainId is correct for testnet.

However provider.getBlockNumber() resulted in an error:

```
"JsonRpcProvider failed to detect network and cannot start up"
"Error: server response 400 Bad Request"
```

Any idea why?

### Steps to reproduce

package.json

```json
{
  "name": "ethersTest",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "build": "tsc",
    "lint": "eslint 'src/**/*.{ts,tsx}'",
    "start": "node dist/index.js"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "@types/node": "^20.7.2",
    "@typescript-eslint/eslint-plugin": "^6.7.3",
    "@typescript-eslint/parser": "^6.7.3",
    "dotenv": "^16.3.1",
    "eslint": "^8.50.0",
    "typescript": "^5.2.2"
  },
  "dependencies": {
    "ethers": "^6.7.1"
  }
}
```

tsconfig.json
```json
{
  "compilerOptions": {
    "target": "ES2018",
    "module": "CommonJS",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "sourceMap":true,
    "forceConsistentCasingInFileNames": true,
    "outDir": "dist"
  },
  "paths": {
    "*": ["node_modules/*", "src/types/*"]
  },
  "include": ["src"]
}
```
Run `npm i`

Add to \src\index.ts

```typescript
import * as ethers from 'ethers'; //ethers.js 6.7.3

async function main() {
  try {
    const networkId = 'testnet';
    const provider = new ethers.JsonRpcProvider(`https://${networkId}.hashio.io/api`);
        
[...]

PR #1817, opened 2023-10-11
Improve info in case of tracing activities

Neurone:
**Description**:
While conducting debug activities you often enable trace log level, and you want to check the actual requests and response between client and server (i.e., #1802).

This PR:
- adds info about request and response bodies when trace log level is enabled.
- reduces the default log level to `info` instead of `trace`

**Related issue(s)**:

N/A

**Notes for reviewer**:

N/A

**Checklist**

### DIFF
--- a/packages/server/src/koaJsonRpc/index.ts
+++ b/packages/server/src/koaJsonRpc/index.ts
@@ -119,6 +119,8 @@ export default class KoaJsonRpc {
         return;
       }
 
+      this.logger.trace(`[Request ID: ${this.getRequestId()}] Request body ${JSON.stringify(body)}`);
+
       ctx.state.methodName = body.method;
       const methodName = body.method;
 
@@ -192,6 +194,7 @@ export default class KoaJsonRpc {
             break;
         }
       }
+      this.logger.trace(`[Request ID: ${this.getRequestId()}] Response body ${JSON.stringify(ctx.body)}`);
     };
   }
 
--- a/packages/server/src/server.ts
+++ b/packages/server/src/server.ts
@@ -30,7 +30,7 @@ import { formatRequestIdMessage } from './formatters';
 
 const mainLogger = pino({
   name: 'hedera-json-rpc-relay',
-  level: process.env.LOG_LEVEL || 'trace',
+  level: process.env.LOG_LEVEL || 'info',
   transport: {
     target: 'pino-pretty',
     options: {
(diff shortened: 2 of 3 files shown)


Nana-EC (changes requested):
Simple enough change.

@Neurone can you add an example of the logs emitted after a simple call to say `eth_getTransactionCount(..)`
One concern here is the expansion of logs and this will result in duplication of log details. Logging every request and response body can be very large.

All the important details should be logged down the line. Are there specific HTTP properties you're looking for that are missing?

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -119,6 +119,8 @@ export default class KoaJsonRpc {
         return;
       }
 
+      this.logger.trace(`[Request ID: ${this.getRequestId()}] Request body ${JSON.stringify(body)}`);
```
Nana-EC, line 122: This may result in repeated info in the logs as some methods log the contents of the body which are params.
In a few cases that's transaction bytes or a whole contract.
So we might need some coordination on that

ebadiere: @Neurone Is this still an issue?

Neurone: The code of the PR must be probably updated at this point, but yes, imho the issue this PR tries to solve is still there.

I think when as a developer I set TRACE as a log level, I want to see anything because I'm doing this for heavy debugging activities. The PR includes the answers and the responses bodies to the logs, and this info is essential for debugging purposes.

At the same time, I think the default log level should be INFO, not TRACE as we use now. In general I think using INFO as default is a good practice, and in this case it shows another good reason why. If we use TRACE as our default, when adding more data to the log it **seems** too much (as reported by Nana) but the problem for me it's not there, because when I enable trace I want to see everything, and duplicated data are not an issue at all and in some cases are very useful (because I can see nothing happened to those data during the business flow).

ebadiere: Thank you for your efforts on this PR and for sharing your insights on the logging levels. We appreciate your initiative to enhance our debugging process.

After discussing with the team, we've decided that it's best to keep the default log level at TRACE for now. Given the current stage of the project, having detailed logs by default is helpful for our ongoing development and debugging activities.

We agree that setting the default log level to INFO is a good practice for more mature projects. We'll certainly revisit this topic as the project evolves.

For the time being, I am planning to close this PR.  Let me know if you have any questions or concerns.

ebadiere:
PR #3133 Updates the configuration doc to list the default `LOG_LEVEL` for now.

PR #1797, opened 2023-10-06
Fix `withdrawalsRoot`

Neurone:
**Description**:
This PR fix the `withdrawalsRoot` value setting it to `0x0000000000000000000000000000000000000000000000000000000000000000`.

That parameter is not a `QUANTITY` but a `DATA` of 32 Bytes, so it cannot be `0x0` or some libraries, i.e., [ethers-rs](https://github.com/gakonst/ethers-rs) fail to parse the JSON result of operations related to blocks, like `eth_getBlockByNumber`, `eth_getBlockByHash`, etc.

**Related issue(s)**:

N/A

**Notes for reviewer**:

I discovered the problem while playing with ethers-rs. This is an example of the error, fixed after updating the `withdrawalsRoot` parameter:

```
[...]

PR #1834, opened 2023-10-16
Foundry Forge Create Contract Fix

### LINKED ISSUE #1800: Restore `v` value in Transaction model
### Problem

2930 and 1559 transactions introduced the `yParity` parameter in the Transaction model and removed the `v` parameter.

These modifications still struggle to spread across the Ethereum ecosystem, so a few months ago, even the official specifications [were updated to include the `v` parameter again](https://github.com/ethereum/execution-apis/pull/446) in the Transaction model.

That would still be an optional, deprecated parameter so it will be removed in the future, but this update gives more time to the ecosystem to upgrade libraries and clients.

As Hedera wants to be as compliant as possible with the Ethereum libraries and tools ecosystem, I suggest following the new JSON-RPC specs and reintroducing the `v` parameter.

The behavior of the `yParity` does not change.

**What if we don't apply this?** Strange things can happen, and sometimes really hard to debug.

For example, I experienced problems with [ethers-rs](https://github.com/gakonst/ethers-rs) yesterday while deploying some smart contracts.

That library does not yet recognize the `yParity` parameter, so it considers the absence of the `v` parameter a non-conformity. As a result, ethers-rs keeps asking multiple times for the same information to the JSON-RPC Relay, thinking the result is corrupted, and essentially bombing the relay until the `TIER_2_RATE_LIMIT` is reached and the `IP_RATE_LIMIT_EXCEEDED` error appears.

All happens silently under the hood, so the users are not aware of what's happening, and they can 1) think it's a Hedera problem and 2) be locked out of other legitim requests because that behaviour eroded silently their requests limit.

Please note this issue can be related to ongoing activities on #1730 

### Solution

As Hedera wants to be as compliant as possible with the Ethereum libraries and tools ecosystem, I suggest following the new JSON-RPC specs and reintroducing the `v` parameter.

The behavior of the `yParity` does not change.

### Alternatives

_No response_

### LINKED ISSUE #1824: Deploy Smart Contract using Foundry Fails with Deserialization Error: invalid length 1...
### Description
Back in January 2023, I successfully used `forge create` to deploy a smart contract on Hedera.
See closed issue #834 
Fast forward to October 2023 and I am running to the below error.

### What I did:
 When i run the following command
`forge create --rpc-url <your_rpc_url> --private-key <your_private_key> contracts/TodoList.sol:TodoList`
I received the following error:

```
Error: 
Deserialization Error: invalid length 1, expected a (both 0x-prefixed or not) hex string or byte array containing 32 bytes at line 1 column 1266. Response: {"timestamp":"0x65299eb0","difficulty":"0x0","extraData":"0x","gasLimit":"0xe4e1c0","baseFeePerGas":"0x1aebc97c400","gasUsed":"0x0","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","miner":"0x0000000000000000000000000000000000000000","mixHash":"0x0000000000000000000000000000000000000000000000000000000000000000","nonce":"0x0000000000000000","receiptsRoot":"0x0000000000000000000000000000000000000000000000000000000000000000","sha3Uncles":"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347","size":"0x59a","stateRoot":"0x0000000000000000000000000000000000000000000000000000000000000000","totalDifficulty":"0x0","transactions":[],"transactionsRoot":"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421","uncles":[],"withdrawals":[],"withdrawalsRoot":"0x0","number":"0x320a0f","hash":"0xb734b7645ec3a5f3e5d51a15b4da8235f56486994c13f6b213be4eaa4b779cc4","parentHash":"0x563b39bf894273da3cd678b9c2952abb3561bb42ecf7750df127f47e3702ba84"}
```

image

### What I expected to happen:
My Contract to be deployed successfully and a similar message to the one below:

```
[⠰] Compiling...
No files changed, compilation skipped
Deployer: 0x4DdC478A1F35Bbf266D2F5767AE399f4d6E83857
Deployed to: 0x47c81E6e3f2880a42149574f08DCaBffCa10e164
Transaction hash: 0xfec3c5a6d726a4f94a57011c7c5f6fe3f54dad70700652570cda626f58a81dca
```

### Steps to reproduce

[hedera_foundry_fork_testVendingMachine.zip](https://github.com/hashgraph/hedera-json-rpc-relay/files/10494681/hedera_foundry_fork_testVendingMachine.zip)
2. edit the .env file
4. run `forge create --rpc-url https://testnet.hashio.io/api --private-key <HEX_Encoded_Private_Key> contracts/TodoList.sol:TodoList`

### Additional context

_No response_

### Hedera network

testnet

### Version

latest

### Operating system

macOS

AlfredoG87:
**Description**:
For compatibility with foundry forge command, we are modifying the block response to have withdrawalsRoot as a long 32Bytes hex 0 and signed transactions to have both yparity and v values.

Changes to the unit tests that were affected

**Related issue(s)**: #1797 

Fixes #1824 
Fixes #1800 

**Notes for reviewer**:

**After the Fix:**
image

**Checklist**

### DIFF
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -1228,7 +1228,8 @@
                         "required": [
                             "yParity",
                             "r",
-                            "s"
+                            "s",
+                            "v"
                         ],
                         "properties": {
                             "yParity": {
@@ -1251,6 +1252,19 @@
                             "s": {
                                 "title": "s",
                                 "$ref": "#/components/schemas/uint"
+                            },
+                            "v": {
+                                "title": "v",
+                                "description": "For backwards compatibility",
+                                "oneOf": [
+                                    {
+                                        "$ref": "#/components/schemas/byte"
+                                    },
+                                    {
+                                        "$ref": "#/components/schemas/null"
+                                    }
+                                ]
+
                             }
                         }
                     }
@@ -1307,7 +1321,7 @@
                 ]
             },
             "TransactionSigned": {
-                "oneOf": [
+                "anyOf": [
                     {
                         "$ref": "#/components/schemas/Transaction1559Signed"
                     },
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -1255,7 +1255,7 @@
                             },
                             "v": {
                                 "title": "v",
-                                "description": "For backwards compatibility",
+                                "description": "For backwards compatibility, `v` is optionally provided as an alternative to `yParity`. This field is DEPRECATED and all use of it should migrate to `yParity`.",
                                 "oneOf": [
                                     {
                                         "$ref": "#/components/schemas/byte"
--- a/packages/relay/src/lib/model.ts
+++ b/packages/relay/src/lib/model.ts
@@ -164,7 +164,6 @@ export class Transaction2930 extends Transaction {
   public readonly yParity!: string | null;
 
   constructor(args: any) {
-    //const { v, ...parentArgs } = args;
     super(args);
     this.yParity = args.v;
     this.accessList = args.accessList;
--- a/packages/relay/src/lib/model.ts
+++ b/packages/relay/src/lib/model.ts
@@ -45,7 +45,7 @@ export class Block {
   public readonly transactionsRoot: string = '0x0';
   public readonly uncles: string[] = [];
   public readonly withdrawals: string[] = [];
-  public readonly withdrawalsRoot: string = '0x0';
+  public readonly withdrawalsRoot: string = '0x0000000000000000000000000000000000000000000000000000000000000000';
 
   constructor(args?: any) {
     if (args) {
@@ -164,9 +164,9 @@ export class Transaction2930 extends Transaction {
   public readonly yParity!: string | null;
 
   constructor(args: any) {
-    const { v, ...parentArgs } = args;
-    super(parentArgs);
-    this.yParity = v;
+    //const { v, ...parentArgs } = args;
+    super(args);
+    this.yParity = args.v;
     this.accessList = args.accessList;
   }
 }
--- a/packages/relay/tests/assertions.ts
+++ b/packages/relay/tests/assertions.ts
@@ -134,5 +134,6 @@ export default class RelayAssertions {
     expect(block.stateRoot).equal(EthImpl.zeroHex32Byte);
     expect(block.totalDifficulty).equal(EthImpl.zeroHex);
     expect(block.uncles).to.deep.equal([]);
+    expect(block.withdrawalsRoot).to.equal(EthImpl.zeroHex32Byte);
   };
 }
--- a/packages/relay/tests/lib/formatters.spec.ts
+++ b/packages/relay/tests/lib/formatters.spec.ts
@@ -212,7 +212,7 @@ describe('Formatters', () => {
       expect(formattedResult.transactionIndex).to.equal('0x9');
       expect(formattedResult.type).to.equal('0x2');
       expect(formattedResult.yParity).to.equal('0x1');
-      expect(formattedResult.v).to.equal(undefined);
+      expect(formattedResult.v).to.equal(`0x1`);
       expect(formattedResult.value).to.equal('0x0');
     });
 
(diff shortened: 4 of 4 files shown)


docs/openrpc.json
```diff
@@ -1251,6 +1252,19 @@
                             "s": {
                                 "title": "s",
                                 "$ref": "#/components/schemas/uint"
+                            },
+                            "v": {
+                                "title": "v",
+                                "description": "For backwards compatibility",
```
Neurone, line 1258: I suggest using the same description provided by the official Ethereum specs:

```text
For backwards compatibility, `v` is optionally provided as an alternative to `yParity`. This field is DEPRECATED and all use of it should migrate to `yParity`.
```

docs/openrpc.json
```diff
@@ -1251,6 +1252,19 @@
...
+                                "oneOf": [
```
Neurone, line 1259: I noticed the official doc uses `$ref: '#/components/schemas/uint'` both for `yParity` and `v`, while we are using `byte` or `null`. Is this intended?

AlfredoG87: both would work. I am following the existing schema that `yparity` already had, and I think is intentional:

```
"byte": {
                "title": "hex encoded byte",
                "type": "string",
                "pattern": "^0x([0-9,a-f,A-F]?){1,2}$"
            },

```

docs/openrpc.json
AlfredoG87: Good suggestion, Thanks!!

✅  Done

docs/openrpc.json
Neurone: Ok, I digged more and compiled the [`openrpc.json` file from the specs](https://github.com/InternetOfPeers/execution-apis/blob/35fb8b528482ce2b2ca92dad6547eae63f1bdcff/openrpc.json) and, weirdly enough, that `$ref: '#/components/schemas/uint'` is not even the final definition, but an intermediate YAML one.

In the final OpenRPC definition, all those `uint` are essentially [`string` with a pattern like this `^0x([1-9a-f]+[0-9a-f]*|0)$`.](https://github.com/InternetOfPeers/execution-apis/blob/35fb8b528482ce2b2ca92dad6547eae63f1bdcff/openrpc.json#L290)

For the future, we should consider to strictly adhering to the generated `openrpc.json` schema from the Ethereum spec, especially if we plan to build a client from the `openrpc.json` file, just to be sure we will be in sync with other clients.

Neurone (approved):

docs/openrpc.json
Nana-EC: @Neurone we do indeed adhere to the specs from the official Ethereum execution-apis repo.
That's what actually pushed us to remove `v` initially.

Also agreed the intermediate specs can be misleading

docs/openrpc.json
```diff
@@ -1307,7 +1321,7 @@
                 ]
             },
             "TransactionSigned": {
-                "oneOf": [
+                "anyOf": [
```
Nana-EC, line 1324: Q: why the change?

Nana-EC (commented):
LG
One question

docs/openrpc.json
Neurone: We should fix some parameters to remain in sync with those specs (i.e., the `yParity` definition is different, I assume due to subsequent updates). I created a dedicated enhancement request #1837

docs/openrpc.json
AlfredoG87: Oh good question and glad you asked.
It was failing, and I had no idea why 😅 until I realized that it was doing match with both , since `oneOf` was looking for either: `Transaction1559Signed` or `TransactionLegacySigned`. and since I added `v` to `Transaction1559Signed` now it was matching for both of them 😅  and hence failing.

in other words, now that we included `v` on `Transaction1559Signed` it now became a superset of `TransactionLegacySigned` and it was matching on both schemas, and since `oneOf` means that only can match `1`. and `anyOf` can match `1` or more.

Nana-EC: Interesting. Thanks for clarifying

Nana-EC (approved):
LG

PR #1849, opened 2023-10-18
Implement foundry example

### LINKED ISSUE #1172: Implement Precompiled Mock Contracts for Faster Local Development Experience
### Problem

Smart contract developers working with Hedera's precompiled contracts (HTS, Exchange Rate, and Prng) currently face challenges when trying to test their smart contracts in a local development environment using popular frameworks like Hardhat and Foundry. This is because these precompiled contracts are only available on Hedera networks, which slows down the development and testing process as even the hedera local node is significantly slower than the foundry and hardhat local nodes.

### Solution

Create mock contracts that mimic the behaviour of Hedera's precompiled contracts and publish them as either:

1. a separate package for easy integration into developers' projects or
2. under a new folder(say `hedera-precompiled-mocks`) in [hashgraph/hedera-smart-contracts/contracts](https://github.com/hashgraph/hedera-smart-contracts/tree/main/contracts)

This would enable a faster and more efficient development experience by allowing developers to test their smart contracts locally without the need to interact with a relatively slow hedera local node/network.

Develop Solidity smart contracts that implement the respective interfaces (`IHederaTokenService`, `IExchangeRate`, and `IPrng`) and mimic the behaviour of the precompiled/system contracts.

Provide documentation and examples for using these mock contracts in popular development frameworks such as Hardhat and Foundry. This could include guidance on deploying the mock contracts and updating the smart contract under test to interact with the mock contract addresses based on the network (main network vs. local development network); this is really only needed for frameworks that do NOT allow developers to specify a deployment address such as **Hardhat**, but not needed for frameworks that do such as **Foundry**.

## Hardhat example

Modify the `ExchangeRatePrecompile`, `PrngSystemContract` and `HederaTokenService` contracts to use the mock contract address when running on a local development network, and the actual precompile contract address when running on the Hedera networks (local, previewnet, testnet, and mainnet). For example the `HederaTokenService` contract can be modified as follows:

```solidity
pragma solidity ^0.8.0;

import "./IHederaTokenService.sol";

abstract contract HederaTokenService is IHederaTokenService {
    address constant actualPrecompileAddress = address(0x167);
    address constant mockAddress = address(0x1234567); // whatever the deployment address of the mock contract is on the hardhat local node
    // to avoid having to frequently change this address the mock contract can be deployed via a deterministic-deployment-proxy such as https://github.com/Arachnid/deterministic-deployment-proxy
    // however the 10mb limit may have to be lifted in a local environment as large test suites may cause the mock contract's storage to exceed this

    // Replace precompileAddress constant with a view function
    function precompileAddress() internal view returns (address) {
        if (block.chainid == 295 || block.chainid == 296 || block.chainid == 297 || block.chainid == 298) {
            return actualPrecompileAddress;
        } else {
            return mockAddress; // ensure that the hardhat local node chainId is not equal to any of the Hedera network chainIds
        }
    }

    function associateTokens(address account, address[] memory tokens) internal returns (int responseCode) {
        address target = precompileAddress(); // will cost a bit more gas

        (bool success, bytes memory result) = target.call(
            abi.encodeWithSelector(IHederaTokenService.associateTokens.selector,
            account, tokens));
        responseCode = success ? abi.decode(result, (int32)) : HederaResponseCodes.UNKNOWN;
    }

    // other functions ...
}
```

## Foundry example

In Foundry, developers can use the `vm.etch` cheat code to deploy the mock contract at the same address as the precompiled contract on the Hedera network, and hence does NOT require any modifications to the `ExchangeRatePrecompile`, `PrngSystemContract` and `HederaTokenService` contracts to dynamically select a precompile address based on the network.

```solidity
HtsPrecompileMock htsPrecompileMock = new HtsPrecompileMock();
bytes memory code = address(htsPrecompileMock).code;
vm.etch(htsPrecompileAddress, code);
```

### Alternatives

_No response_

mshakeg:
**Description**:

Replaces https://github.com/hashgraph/hedera-json-rpc-relay/pull/1198 for a cleaner review.

In addition to the changes described in #1198 the changes in this PR also demo how to use the system contract mock contracts in the `hardhat-example` project against a local hardhat node which supports the `hardhat_setCode` method.

**Related issue(s)**:

Fixes #1172

**Notes for reviewer**:

Requires https://github.com/hashgraph/hedera-smart-contracts/pull/503 ; once this PR is merged I'll update the `hardhat-example` dependency to point to the official `hedera-smart-contracts` instead of my fork of it.

Should I add `tools/` to the `.dockerignore`?

**Checklist**


tools/hardhat-example/contracts/SimpleVault.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
Nana-EC, line 1: Q: what's the goal with the `hardhat-example` update?
Tool examples in this repo are supposed to show simple coverage that the tool in this case hardhat works with the relay. It doesn't need to have additional contract testing.
If there's a gap in contract coverage that should go in the smart-contracts repo.
In this repo the additional tests should be around the relay API which would be the in the main `test` folder and not under a single tool.

Either ways this change would have to be pulled out into it's own PR as we try to keep PRs focused  for easier review and tracking.

.husky/commit-msg
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
Nana-EC, line 1: Thanks 😄

Nana-EC (changes requested):
Thanks.
We should aim to keep this PR targeting to `foundry-example` and simple.
Additional scope should be in either the hedera-smart-contracts repo or under general relay tests depending on what it's testing.

Nana-EC:
> 
> Should I add `tools/` to the `.dockerignore`?
> 
> **Checklist**
> 
> * [ ]  Documented (Code comments, README, etc.)
> * [ ]  Tested (unit, integration, etc.)

Yeah, it definitely should be part of the docker build process.

.gitmodules
```diff
@@ -0,0 +1,9 @@
+[submodule "tools/foundry-example/lib/forge-std"]
```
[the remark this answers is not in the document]
mshakeg: As far as I can tell, not without initializing `.git` in `./tools/foundry-example`, I'll do this in a separate PR to this branch which we can merge if that's better.

tools/foundry-example/.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
+name: test
```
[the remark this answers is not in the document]
mshakeg: I missed it, `forge init` created it.

.gitmodules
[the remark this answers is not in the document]
mshakeg: I think this is the cleanest solution as explained [here](https://github.com/mshakeg/hedera-json-rpc-relay/pull/28#issuecomment-1770497993)

tools/hardhat-example/contracts/SimpleVault.sol
mshakeg: `hardhat-example` changes are now on [this PR](https://github.com/mshakeg/hedera-json-rpc-relay/pull/27)

Nana-EC: > It's to simply demo how the system contract mocks can be used in a hardhat project:
> 
> 
> 
> 1. install [hashgraph/hedera-smart-contracts](https://github.com/hashgraph/hedera-smart-contracts) as a dependency
> 
> 2. In your contracts use the various Hedera system contract abstract contracts/interfaces as you normally would as demonstrated in the `SimpleVault` example contract
> 
> 3. In your JS/TS test scripts deploy the requisite Hedera system contract mocks to their respective pre-defined addresses using `hardhat_setCode`
> 
> 4. You can now FULLY test your Hedera specific smart contracts against a hardhat local node, whereas before you had to connect to a Hedera local node/network, which is way slower and doesn't have the many debugging methods of the hardhat local node.
> 
> 
> 
> ok, I'll create a separate PR titled "demo system contract mocks in hardhat example"

Okay. Great.
Initially there were concerns that system contracts would not work since a Hardhat local node does not have the Hedera Java logic that matches the contracts.
Are you saying this approach isn't affected by that.
If it's just a regular non system contract then hardhat and other tools would work fine with a Hedera relay. However, I'm curious about the above.

Thanks for breaking it out into a separate PR, will take a look.

.gitmodules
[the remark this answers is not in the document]
Nana-EC: I agree. 
Let's move it all into the self contained foundry example directory and put the right documentatioj there for people when they do a git clone.
The main relay root dir shouldn't be impacted by a single tool example.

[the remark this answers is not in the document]
mshakeg: @Nana-EC I'm confused. By "I think this is the cleanest solution" I meant the current solution that has the `.gitmodules` in the relay root dir, not in `foundry-example` as the latter complicates the `foundry-example` project setup/installation.

Do you still think I should go with the less clean solution?

tools/hardhat-example/contracts/SimpleVault.sol
mshakeg: > Initially there were concerns that system contracts would not work since a Hardhat local node does not have the Hedera Java logic that matches the contracts. Are you saying this approach isn't affected by that.

That's still an issue, but that's not what the addition to the `hardhat-example` does. Rather it's a demo of how developers could avoid having to use a Hedera local node(or network) in their hardhat projects by instead deploying the newly added **system contracts mocks** defined in `hedera-smart-contracts` to their local hardhat node and effectively mock the environment of a Hedera node.

Hope that clarifies it.

.gitmodules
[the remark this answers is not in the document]
Nana-EC: @mshakeg I see.
Yes, let's go with the less clean option then.
The relay code, checkout and build process should not be affected by the steps necessary to explore any of the tool examples.
All tool examples should be self contained in their directory

tools/foundry-example/README.md
```diff
@@ -0,0 +1,104 @@
+## Foundry
+
+## Attempt at Managing Submodules in `foundry-example` in the context of the `hedera-json-rpc-relay` repo
```
Nana-EC, line 3: Rename and move this to right after the `##Documentation`
```suggestion
## Git Submodules Management
```

tools/foundry-example/script/Counter.s.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: UNLICENSED
```
Nana-EC, line 1: nit: can you set contract licenses to Apache-2.0

Nana-EC (changes requested):
Solid changes and thanks for compartmentalizing them.
Just a few nits for improved documentation and license policy

tools/hardhat-example/contracts/SimpleVault.sol
Nana-EC: Sounds good, will watch out for PR submission here later if you want to bring that along to enhance the hardhat project example.
Thanks

mshakeg: will do, I was considering implementing the `hardhat-example` in TS and JS is a pain to develop in. I was also surprised that the `hedera-smart-contract` hardhat project doesn't use TS.

mshakeg:
@Nana-EC thanks for the review, are there any comments I missed?

Nana-EC (approved):
LGTM
Another great addition.
Thanks

Nana-EC:
> @Nana-EC thanks for the review, are there any comments I missed?

Nahh you got everything.
Just gonna wait on CI to pass and I'll merge

mshakeg:
@Nana-EC thanks, I've pushed a fix for the Charts workflow that failed.

mshakeg:
Hey @Nana-EC any idea why that `Dapp Tests` test case timed out? It might just be an intermittent issue so re-running that job should get it to pass.

Nana-EC:
> Hey @Nana-EC any idea why that `Dapp Tests` test case timed out? It might just be an intermittent issue so re-running that job should get it to pass.

@mshakeg  Passed on rerun.
Chart install is still broken on build images in relation to sub modules.

mshakeg:
@Nana-EC that's really strange. The Charts workflow goes past the `Build and push images` step on [my run](https://github.com/mshakeg/hedera-json-rpc-relay/actions/runs/6598661185/job/17926864608).

I suspect there may be caching going on, so I created a new PR https://github.com/hashgraph/hedera-json-rpc-relay/pull/1854

Hopefully it's no longer an issue.

PR #1854, opened 2023-10-22
Implement foundry example

### LINKED ISSUE #1172: Implement Precompiled Mock Contracts for Faster Local Development Experience

mshakeg:
**Description**:

Replaces https://github.com/hashgraph/hedera-json-rpc-relay/pull/1849 in an attempt to fix `Charts` workflow failure.

**Related issue(s)**:

Fixes #1172 

**Notes for reviewer**:

**Checklist**


mshakeg:
@Nana-EC I temporarily disabled the cache, if that fixes it then I'll revert back to using it.

mshakeg:
@Nana-EC well that's frustrating :(

I really don't know what else to try.

Nana-EC:
> @Nana-EC well that's frustrating :(
> 
> I really don't know what else to try.

Hmm, @mshakeg let me ask around. We're almost there.
@isavov any ideas?

tools/foundry-example/.gitmodules
isavov, line 1: the `build-push-action` checks out the repo independently when it builds images, so `.gitmodules` in subdirectories should be avoided to prevent issues with it, this should either be moved to the root of the git repo, or the dependency on submodules should be avoided entirely.

mshakeg: thanks for the review @isavov

`tools/` is in the `.dockerignore` so should it be an issue? Also why does [my run](https://github.com/mshakeg/hedera-json-rpc-relay/actions/runs/6598661185/job/17926864608) go past the `Build and push images` Charts step?

.github/workflows/charts.yml
```diff
@@ -50,8 +51,6 @@ jobs:
       - name: Build and push images
         uses: docker/build-push-action@v4
         with:
```
isavov, line 53:
```suggestion
        with:
          cache-from: type=gha
          cache-to: type=gha,mode=max
          context: .
```

isavov (changes requested):

tools/foundry-example/.gitmodules
isavov: @mshakeg It appears that the run you've linked to is from main, anyways I've discovered a workaround for the issue. The suggested changes use the locally checked out copy instead of trying to checkout the repo inside docker.

.github/workflows/charts.yml
mshakeg: Hey @isavov thanks for this, I made the change, let's hope it works.

isavov (approved):

.github/workflows/charts.yml
Nana-EC: Seems it does. Thanks @isavov

Nana-EC (approved):

PR #1872, opened 2023-11-03
Adding support for Batch Requests.

### LINKED ISSUE #1820: Support for Batch/Multi Request
### Problem

Ethers V6 by default is using batch requests against the JSON-RPC Providers.
But `Hedera-JSON-RPC-Relay` does not support it.

**Example Request of  a batch request:**

```bash
curl --location 'https://eth-mainnet.g.alchemy.com/v2/<TOKEN_API>' \
--data '[
    {
        "jsonrpc": "2.0",
        "id": 1,
        "method": "eth_chainId",
        "params": []
    },
    {
        "jsonrpc": "2.0",
        "id": 2,
        "method": "eth_blockNumber",
        "params": []
    },
    {
        "jsonrpc": "2.0",
        "method": "eth_getBlockByNumber",
        "params": [
            "0x117c97e",
            false
        ],
        "id": 3
    }
]'
```

**Response on HEDERA-JSON-RPC-RELAY**
```json
{
    "error": {
        "message": "Invalid Request",
        "code": -32600
    },
    "jsonrpc": "2.0",
    "id": null
}
```

**Example response: (against Alchemy Ethereum mainnet)**
```json
[
    {
        "jsonrpc": "2.0",
        "id": 1,
        "result": "0x1"
    },
    {
        "jsonrpc": "2.0",
        "id": 2,
        "result": "0x117cc49"
    },
    {
        "jsonrpc": "2.0",
        "id": 3,
        "result": {
            "number": "0x117c97e",
            "hash": "0x40e89a4ee5728677937af5044f4161bb04660012638e35b6171474e57621f035",
            "transactions": [
                "0xed75937b14532d76081edbba588e5467b72b6791898b89f2f83dcab761a3b7fb",
               ...
            ],
            "difficulty": "0x0",
            "extraData": "0xd883010c02846765746888676f312e32302e37856c696e7578",
            "gasLimit": "0x1c9c380",
            "gasUsed": "0x8819e7",
            "logsBloom": "0x64a00005c588d1c8c44b886cd090792a8881134936d2c0c00d590004b6008e08ab61748c820135c490093783242435ed7681924ca90c76e6474100120173f80240021c2d6043a92c68fb510ac2e88528a2831ce14d623c4c013d845391e982ac3ec0660bca02c90c2c88a051a82868170b41c4212e18c5001f0c4454b029062300509108b0054fc935c854a02120400e242024210d53143a2694004841922c36cb89915c8962673b9add44dc94a19e10b83444805020430750f0963cfc6a29e28a8314ea4745d54102014da812a0f820606e5a0296dd757c07c1002a0226e04351d6243c484082681805a1a261a0501825385188581800e01920085b920e2310",
            "miner": "0xcfe1b6d89057db4e5bf0dba527fa10ee919c3342",
            "mixHash": "0xad40f88a46b52b8b7a49f5b8405ca5c12e89b6bc13deef3cdfb36b6d6e275700",
            "nonce": "0x0000000000000000",
            "parentHash": "0xd7f625b2aedc59b4f955a655b7ebb2f053378186fd151dec8f11af9ade344456",
            "receiptsRoot": "0x15a8087062c2e038059f8c8c960f0a5ca8b0da833b92f26a221b572036741d15",
            "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
            "size": "0x1077c",
            "stateRoot": "0x8a11aa3bfe1fabf6fc3a768263013dacb931c361d36c22c1989dd3e39b9bcb52",
            "timestamp": "0x65283af7",
            "totalDifficulty": "0xc70d815d562d3cfa955",
            "transactionsRoot": "0x84884cf851b3d752d01f6778c01ae0484611fbdbf107cda794da66988a417dcc",
            "uncles": [],
            "baseFeePerGas": "0x257b62cf8",
            "withdrawalsRoot": "0x357e38c82008155d4cbf5d7b85fb667932a2b6d5d21a767777300a67dcf7d18d",
            "withdrawals": [...]
        }
    }
]
```

### Solution

Add support for batch request, the request can be of many of the same method/endpoint or even different ones.

### Alternatives

_No response_

AlfredoG87:
**Description**:
Adding support for Batch Requests.

Refactored some logic to make it cleaner and simpler. deleted some unused classes (isJson and RpcInvalidError).

Improved Method not found error message by including the name of the method that was not found.

Fixed tests after improvements and added Test Cases for Batch Requests functionality

**Related issue(s)**:

Fixes #1820 

**Notes for reviewer**:

Batch Requests will always return `200` Http status code, even if some or all of its individual requests have errors. (this is standard in JsonRpc Services, I tried with both `Alchemy` and `Infura`.

examples here:

### **Infura:**

**Batch request of 2, one correct eth_chainId and another incorrect due to missing params**
image

**Batch request of 3, one correct eth_chainId, a second one method not found eth_chainId2 and another incorrect due to missing params**

image

### **Alchemy:**

**Batch request of 3, one correct eth_chainId, a second one method not found eth_chainId2 and another incorrect due to missing params**
image

### **Relay after this changes:**
**Batch request of 3, one correct eth_chainId, a second one method not found eth_chainId2 and another incorrect due to missing params**
image

**When Batch Requests are disabled:**

image

```
{
    "error": {
        "code": -32001,
        "name": "Batch requests disabled",
        "message": "Batch requests are disabled"
    },
    "jsonrpc": "2.0",
    "id": null
}
```

****For Current single requests, everything stays as is currently working**

**Checklist**


AlfredoG87:
Charts / install  CI WF Test is failing due to a verification on the charts installed using the postman newman tests on main, and the test on main have the previous `method not found`. so this should be fixed once is merged into `main`.

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -119,80 +134,162 @@ export default class KoaJsonRpc {
...
+      );
+      ctx.status = 400;
+      ctx.state.status = `${ctx.status} (${INVALID_REQUEST})`;
+      return;
+    }
+
+    const response: any[] = [];
+    ctx.state.methodName = 'batch_requests';
+
+    // we do the requests in parallel to save time, but we need to keep track of the order of the responses (since the id might be optional)
```
Nana-EC, line 223: Q: do we need to ensure the result is in the same order submitted?

packages/server/src/server.ts
```diff
@@ -770,10 +770,22 @@ const rpcApp = app.rpcApp();
...
+  // single request
   if (ctx.body && ctx.body.result instanceof JsonRpcError) {
     ctx.body.error = { ...ctx.body.result };
     delete ctx.body.result;
   }
+
+  // batch request
+  if (ctx.body && ctx.body instanceof Array) {
+    ctx.body.forEach((item) => {
+      if (item.result instanceof JsonRpcError) {
```
Nana-EC, line 783: Add a comment for the logic here

Nana-EC (changes requested):
Nice changes. 
Could have been better broken out but not too bad.
A few suggestions to address

packages/server/src/koaJsonRpc/index.ts
AlfredoG87: The clients will and should use the `id`, but even if they don't results will come in the same order.  That said, the `body.map` will return an array of promises in the same order of the requests, so the order should be preserved.

packages/server/src/server.ts
AlfredoG87: Due to my refactor I am pretty sure sure that the whole of the KoaApp middleware is no longer needed, it used to be needed since there was a bug on the way the results/responses were built even when there were errors. but after the refactor and code improvements done on this ticket.  I don't think is needed.

So thanks for bringing this to my attention, I will remove the following code and nothing will break.
```
app.getKoaApp().use(async (ctx, next) => {
  await rpcApp(ctx, next);
  // Handle custom errors

  // single request
  if (ctx.body && ctx.body.result instanceof JsonRpcError) {
    ctx.body.error = { ...ctx.body.result };
    delete ctx.body.result;
  }

  // batch request
  if (ctx.body && ctx.body instanceof Array) {
    ctx.body.forEach((item) => {
      if (item.result instanceof JsonRpcError) {
        item.error = { ...item.result };
        delete item.result;
      }
    });
  }
});
```

AlfredoG87: Okay, the KoaApp is needed 😅 
But the Handle of custom errors is not, this was a workaround due to another bug that was fixed during the refactors on this PR.

Nana-EC: Okay let's minimize the refactor where possible and open separate tickets if necessary.
if we improve as we go along though that's great

packages/server/tests/integration/server.spec.ts
```diff
@@ -423,6 +423,215 @@ describe('RPC Server', async function () {
...
+      BaseTest.baseDefaultResponseChecks(response);
+
+      // verify response for each request
+      for (let i = 0; i < response.data.length; i++) {
+        expect(response.data[i].id).to.be.equal((i + 2).toString());
+        expect(response.data[i].result).to.be.equal('0x' + Number(process.env.CHAIN_ID).toString(16));
+      }
+    });
+
+    it('should execute "eth_chainId" and "eth_accounts" in batch request', async function () {
```
Nana-EC, line 490: Would be great to have an acceptance test and add it to `@release` that covers the standard flow of

1. chain id
2. account nonce
3. gas estimate
4. submit raw transaction
5. transaction receipt
6. contract state call

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -119,80 +134,162 @@ export default class KoaJsonRpc {
...
+    const promises = body.map((item: any) => this.getRequestResult(item, ctx.ip));
```
Nana-EC, line 224: Q: should `handleMultipleRequest()` be calling `handleSingleRequest()` so simplify and optimize code share?
If it's a large refactor let's open a ticket to follow up

Nana-EC (approved):
Looks good.
Can add acceptance test now or in follow up

packages/server/src/koaJsonRpc/index.ts
AlfredoG87: It can't I placed all the shared code on `getRequestResult` and basically the difference between `handleMultipleRequest` and `handleSingleRequest` is that the first executes a batch of calls to `getRequestResult` and then response `200` http status with the collection of responses, as opposed `handleSingleRequest` keeps current behavior intact and depending on the result of the request changes the `http status` code response accordingly to the switch that has after the `getRequestResult`. hope this makes sense.

packages/server/tests/integration/server.spec.ts
AlfredoG87: just created a new issue for follow up on this 👍

PR #1877, opened 2023-11-07
Design Documentation for Batch Requests

### LINKED ISSUE #1875: Add a Design Document for Batch Requests
### Problem

We are working on allowing batch requests on the relay, however it implies several changes and design decisions that need to be written in a design doc for future reference

### Solution

Add a design document to leave in written the design decisions that were taken for batch requests feature, how it works (parallel or sequential), the implications it has on rate limits, metrics and performance.
List of tasks that will be needed to complete the feature.

### Alternatives

_No response_

AlfredoG87:
**Description**:
Including Documentation for Batch Requests including the following topics:

Overview
Goals / Motivation
Configuration
Implementation and Usage
Errors
Metrics
Rate Limits
Tests
Tasks

**Related issue(s)**:

Fixes #1875

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/design/batch-request.md
+++ b/docs/design/batch-request.md
@@ -17,14 +17,14 @@ This is a standard JSON-RPC 2.0 feature. For more information, see the [JSON-RPC
 2. **BATCH_REQUESTS_MAX_SIZE:** If the batch request is enabled, this is the maximum number of requests allowed in a batch request, default value is `100`.
 3. **WS_BATCH_REQUESTS_ENABLED:** Flag to enable `batch requests` on the websocket, it can be `true` or `false`, default value is `false`
 4. **WS_BATCH_REQUESTS_MAX_SIZE:** If the batch request is enabled, this is the maximum number of requests allowed in a batch request, default value is `20`.
-
+5. **BATCH_DISALLOWED_METHODS:** A list of methods that are not allowed to be part of a batch request. default: (debug_traceTransaction,eth_getFilterLogs,eth_uninstallFilter,eth_newFilter,eth_newPendingTransactionFilter,eth_newBlockFilter)
 
 ## Implementation
 
 * Current single requests should stay the same as they are currently.
 
 * All batch requests should return a `200` response code, even if there are errors on the individual requests. except in case that fails the validations for the batch request as a whole or common validations.
-* Batch requests should be implemented on top of the current implementation, refector into common flow and common code as much as possible.
+* Batch requests should be implemented on top of the current implementation, refactored into common flows and common code reused much as possible.
 
 **Common Flow for both Batch and Single Requests:**
 1. Validate the request is a valid json
@@ -34,13 +34,13 @@ This is a standard JSON-RPC 2.0 feature. For more information, see the [JSON-RPC
 **Validations per Each individual request on a batch or as a single request:**
 1. Validate JSON-RPC 2.0 Specification
 2. Validate Method is supported
-3. Validate and increase the request count for the rate limit
+3. Validate Method is allowed for batch request
+4. Validate and increase the request count for the rate limit
 
 **Validations for the Batch Request as a whole:**
 1. Validate that the request is an array
 2. Validate Batch Request is enabled
 3. Validate Batch Request size is within the limit
-
 
 
 #### Batch Request examples
@@ -56,7 +56,7 @@ The request object is a JSON array with one or more transactions, a unique `id` 
         "params": []
     },
     {
-        "id": "1",
+        "id": 2,
         "jsonrpc": "2.0",
         "method": "eth_getBalance",
         "params": [
@@ -68,7 +68,7 @@ The request object is a JSON array with one or more transactions, a unique `id` 
         "jsonrpc": "2.0",
         "method": "eth_blockNumber",
         "params": [],
-        "id": 1
+        "id": 3
     }
 ]
 ```
@@ -87,17 +87,15 @@ The response object is a JSON array with each result from the request array. Eac
     {
         "result": "0x21e19e0c9bab2400000",
         "jsonrpc": "2.0",
-        "id": "1"
+        "id": 2
     },
     {
         "result": "0x147",
         "jsonrpc": "2.0",
-        "id": 1
+        "id": 3
     }
 ]
 ```
-
-
 
 ## Errors
 
@@ -123,6 +121,7 @@ This happens when there are errors on one or many individual requests items of t
 2. The request item does not conform to the JSON-RPC 2.0 specification (Invalid Request)
 3. Missing required parameters or wrong parameter type 
 4. The request item is valid but the result is an error, like an `eth_call` that returns a `revert` or `out of gas` error
+5. The request item method is not allowed for batch requests
 
 Example of a response with the above errors (in the same order):
 
@@ -161,6 +160,15 @@ Example of a response with the above errors (in the same order):
         },
         "jsonrpc": "2.0",
         "id": 4
+    }
+    {
+        "error":{
+            "code":-32007,
+            "name": "Method not allowed for batch requests",
+            "message": "Method eth_getFilterLogs is not permitted as part of batch requests"
+        },
+        "jsonrpc": "2.0",
+        "id": 5
     }
 ]
 ```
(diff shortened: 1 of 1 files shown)


docs/design/batch-request.md
```diff
@@ -0,0 +1,260 @@
...
+## Overview
+
+Batch requests consolidate multiple API calls within a single HTTP request. Clients bundle multiple requests into an array and receive an array of responses in return. The server may process these batched requests in any sequence, concurrently. Clients must then correlate the responses to the original requests using the unique ID of each object.
+
+This is a standard JSON-RPC 2.0 feature. For more information, see the [JSON-RPC 2.0 Specification - Batch](https://www.jsonrpc.org/specification#batch)
+
+## Goals / Motivation
+1. Adhere to the JSON-RPC 2.0 specification for batch requests
+2. Support Tooling that uses batch requests, such as EthersJs v6
+3. Support batch requests for all methods that are supported for single requests
```
Nana-EC, line 13: Note that certain methods aren't supported for batch e.g. debug and trace.
We should probably add a config that captures a list of methods not supported.

docs/design/batch-request.md
```diff
@@ -0,0 +1,260 @@
...
+3. **WS_BATCH_REQUESTS_ENABLED:** Flag to enable `batch requests` on the websocket, it can be `true` or `false`, default value is `false`
+4. **WS_BATCH_REQUESTS_MAX_SIZE:** If the batch request is enabled, this is the maximum number of requests allowed in a batch request, default value is `20`.
+
+
+## Implementation
+
+* Current single requests should stay the same as they are currently.
+
+* All batch requests should return a `200` response code, even if there are errors on the individual requests. except in case that fails the validations for the batch request as a whole or common validations.
+* Batch requests should be implemented on top of the current implementation, refector into common flow and common code as much as possible.
```
Nana-EC, line 27:
```suggestion
* Batch requests should be implemented on top of the current implementation, refactored into common flows and common code reused much as possible.
```

docs/design/batch-request.md
```diff
@@ -0,0 +1,260 @@
...
+
+**Common Flow for both Batch and Single Requests:**
+1. Validate the request is a valid json
+2. Validate the Method is supported (POST only)
```
Nana-EC, line 31: Note we do support some GETs on the relay regarding CORS etc.
Let's make sure we don't regress here
I think this will come down to the order of KOA middleware operations

docs/design/batch-request.md
```diff
@@ -0,0 +1,260 @@
...
+```json
+[
+    {
+        "jsonrpc": "2.0",
+        "id": 1,
+        "method": "eth_chainId",
+        "params": []
+    },
+    {
+        "id": "1",
```
Nana-EC, line 59: id is the same here and next. Did you want to increment them?

docs/design/batch-request.md
```diff
@@ -0,0 +1,260 @@
...
+[
+    {
+        "result": "0x12a",
+        "jsonrpc": "2.0",
+        "id": 1
+    },
+    {
+        "result": "0x21e19e0c9bab2400000",
+        "jsonrpc": "2.0",
+        "id": "1"
```
Nana-EC, line 90: set the ids to unique.
Actually is this something we need to validate as part of the batch logic also?

docs/design/batch-request.md
```diff
@@ -0,0 +1,260 @@
...
+]
+```
+
+
+
+## Errors
+
+There can be success responses and error responses in the same batch request.
+
+### Errors related to Batch Requests
```
Nana-EC, line 106: Q: should we add one for duplicated ids found in batch? Or might this be too restrictive?

docs/design/batch-request.md
```diff
@@ -0,0 +1,260 @@
...
+The suggested enhancement is to retain the existing metric unchanged while introducing an additional metric to track individual requests within a batch.
+
+The new metric will be called `rpc_relay_batch_request` and will capture the following labels:
+1. `method` (method name such as eth_call, eth_getBlockByNumber)
+2. `status` (instead of HTTP status code, it will be the status of the individual request, success or error, in case of sucess it will have the value `200` to keep it similar to the current metric, but in case of error it will have the RPC error code like `-32005` or `-32006`)
+
+The processing time in milliseconds (ms) will continue to be observed for this metric as well.
+
+This proposed approach offers several benefits. It preserves the integrity of our current metric system while enabling the capture of detailed performance metrics for each request within a batch. This data can then be showcased on a new dashboard designed to monitor method performance independently of request type—singular or batch. Concurrently, our existing dashboards will continue to provide comprehensive metrics for all single and batch requests at the whole HTTP request level.
```
Nana-EC, line 226: Q: does this mean the dashboard would need an aggregated view of both the metrics?

docs/design/batch-request.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Metrics Capture Enhancement
+Currently metrics are captured using a KoaApp middleware that captures the metric `rpc_relay_method_response` per each HTTP request capturing the amount of time it took to process the request in milliseconds (ms).
+
+The metric currently uses the following labels:
+
+1. `method` (method name such as eth_call, eth_getBlockByNumber or batch_request)
+2. `status` (HTTP status code)
+
+This approach works for both single and batch requests, however it does not capture the metrics for each individual request within the batch request.
```
Nana-EC, line 213: Q: Why does this not work. Could the metric not be captured in the batching logic so that there's no need for a new metric?
Is there no share request processing logic that would allow the metric to be captured regardless of if it was part of a single or batch request?

Nana-EC (changes requested):
Looking good.
Some question and suggestions

docs/design/batch-request.md
AlfredoG87: Yes, I noticed that on the documentation for `Alchemy` but did not found any exceptions (or way to configure that) on the documentation for ethersjs, but I agree that we should configure them via property. will add it to the design doc and add a follow up ticket for that.

docs/design/batch-request.md
AlfredoG87: Yes, but I will take a double look regardless 👍

docs/design/batch-request.md
AlfredoG87: it might be a good consideration, although Alchemy or infura let you send repeated ids and they both return the items in the same order, and since `id` might be optional in some cases. I did not add that validation.

docs/design/batch-request.md
AlfredoG87: Yes, but only if what we want to measure is the time it takes a `method` to reply, if we are measuring the time it takes an `http request` to reply, the same dashboards will work.

So My initial thought is to create a new dashboard and be more descriptive on the intended use for each one, in one we are looking for http latency (potencial timeouts) and in the other we are measuring the time it takes a method to complete. (they will overlap in the case of single request)

If you have other ideas am open to discussing them. 👍

docs/design/batch-request.md
AlfredoG87: yes, is possible to use the current metric, and only use `1` but then it won't be the same as what we are measuring right now (we are mostly using that metric to measure `http latency` to find potential `timeouts`. 

But come to think of it again, I think we could use it and just add a new label, isBatched, and we can filter out all the occurrences that come with this flag when doing a dashboard for `http latency` and not when measuring the time it takes a method to return...

I like this approach best 👍

docs/design/batch-request.md
AlfredoG87: https://github.com/hashgraph/hedera-json-rpc-relay/issues/1885

docs/design/batch-request.md
AlfredoG87: ✅  I was able to find another approach that might also work better and simpler 👍

docs/design/batch-request.md
AlfredoG87: ✅ I was able to find another approach that might also work better and simpler 👍

docs/design/batch-request.md
AlfredoG87: ✅ Done

docs/design/batch-request.md
AlfredoG87: ✅ I've confirmed not to break `GET` Endpoints as `/health/readiness`, /health/liveness, `/metrics` and OPTIONS

docs/design/batch-request.md
AlfredoG87: ✅  Done

docs/design/batch-request.md
AlfredoG87: ✅  added to Documentation and created a follow up ticket for this #1885

docs/design/batch-request.md
Nana-EC: Okay let's leave validation out, please note this in the design doc.

docs/design/batch-request.md
Nana-EC: No-op here, just document behaviour

ebadiere (approved):
LG.  Questions have been answered and acknowledged.

Nana-EC (approved):

PR #1889, opened 2023-11-09
1600 - estimate gas with precompiles

attest08:
Added 22 scenarios from the following test plan for eth_estimateGas with Precompiles:
https://www.notion.so/limechain/eth_estimateGas-with-Precompiles-json-rpc-relay-ef4b9112654044c5a4be81dd27b05e4b?pvs=4

### DIFF
--- a/packages/server/tests/acceptance/estimateGasPrecompile.spec.ts
+++ b/packages/server/tests/acceptance/estimateGasPrecompile.spec.ts
@@ -31,7 +31,6 @@ import ERC721MockJson from '../contracts/ERC721Mock.json';
 describe('EstimatePrecompileContract tests', function () {
   const signers: AliasAccount[] = [];
   const prefix = '0x';
-  const SERVER_ERROR = 'SERVER_ERROR';
   const CALL_EXCEPTION = 'CALL_EXCEPTION';
   let contract: ethers.Contract;
   let contractReceipt;
@@ -41,6 +40,7 @@ describe('EstimatePrecompileContract tests', function () {
   let nftAddress;
   let NftSerialNumber;
   let estimateContractSigner0;
+  let estimateContractSigner1;
   let estimateContract;
 
   const accounts: AliasAccount[] = [];
@@ -136,7 +136,7 @@ describe('EstimatePrecompileContract tests', function () {
     tokenAddress = await createFungibleToken();
     nftAddress = await createNft();
 
-    let estimateContractSigner1 = new ethers.Contract(
+    estimateContractSigner1 = new ethers.Contract(
       prefix + EstimatePrecompileContractAddress,
       EstimatePrecompileContractJson.abi,
       accounts[1].wallet,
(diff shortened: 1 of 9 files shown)


packages/server/tests/acceptance/estimateGasPrecompile.spec.ts
```diff
@@ -0,0 +1,473 @@
...
+    await contract.grantTokenKycExternal(tokenAddress, accounts[2].wallet.address, { gasLimit: 500_000 });
+    await contract.grantTokenKycExternal(tokenAddress, prefix + EstimatePrecompileContractAddress, {
+      gasLimit: 500_000,
+    });
+  });
+
+  const baseGasCheck = (estimatedGasValue, expectedValue: number) => {
+    // handle deviation of 20%
+    expect(Number(estimatedGasValue)).to.be.lessThan(expectedValue * 1.4);
+  };
```
yiliev0, line 192: Currently this checks if estimateGas value is less than 140% of the expectedValue. This should be tweaked to accept lower bound and upper bound params to set a certain deviation range, like 5%-20%. This way, we can precisely control how much `estimatedGasValue` can differ from `expectedValue`.

packages/server/tests/acceptance/estimateGasPrecompile.spec.ts
```diff
@@ -0,0 +1,473 @@
...
+    baseGasCheck(estimateGasResponse, 786166);
+  });
+
+  //EGP-003
+  it('should call estimateGas with dissociate token function without association for fungible token - negative', async function () {
+    const tx = await contract.dissociateTokenExternal.populateTransaction(
+      prefix + EstimatePrecompileContractAddress,
+      tokenAddress,
+    );
+    negativeScenarioVerification(tx, SERVER_ERROR);
```
yiliev0, line 214: Does `SERVER_ERROR` mean we get status code 500 from the API. If that's the case, this means the test is not doing something correctly, because we don't expect code 500 in a negative scenarios, only 400 that results in `CONTRACT_REVERT_EXECUTED`.
Same comment applies for the tests below.

packages/server/tests/acceptance/estimateGasPrecompile.spec.ts
```diff
@@ -0,0 +1,473 @@
...
+    );
+    await estimateContractSigner0.associateTokenExternal(
+      prefix + EstimatePrecompileContractAddress,
+      tokenAddress,
+      Constants.GAS.LIMIT_1_000_000,
+    );
+
+    const tx = await contract.dissociateTokenExternal.populateTransaction(estimateContractSigner0, tokenAddress);
+    const estimateGasResponse = await relay.call(RelayCalls.ETH_ENDPOINTS.ETH_ESTIMATE_GAS, [tx]);
+    baseGasCheck(estimateGasResponse, 785119);
```
nickeynikolovv, line 273: If the expected values change it will be hard to be maintained one by one, so you can use enum for them.

AlfredoG87 (approved):
LGTM

AlfredoG87:
@attest08 please review this PR, I approved, but will let you decide if and when to merge: https://github.com/hashgraph/hedera-json-rpc-relay/pull/1930/files

AlfredoG87 (approved):
LGTM

Issue #1878, opened 2023-11-07
Batch Requests Acceptance Tests


AlfredoG87:
Do Acceptance test to validate a Request Batch with the following methods:

**Batch 1:**

1. chainId
2. accountNonce
3. fee and gasPrice

**Single Request 2:**
Then Perform the SubmitRawTransaction individually that changes something on a given contract

**Batch 2:**
and then perform another batch request with the following requests:
1. AccountNonce *should be higher by 1
2. transaction receipt
3. a eth_call that validate the change in the contract.

*** Old ask --> 
```
Would be great to have an acceptance test and add it to `@release` that covers the standard flow of

1. chain id
5. account nonce
6. gas estimate
7. submit raw transaction
8. transaction receipt
9. contract state call

_Originally posted by @Nana-EC in https://github.com/hashgraph/hedera-json-rpc-relay/pull/1872#discussion_r1385617299_
```

PR #1891, opened 2023-11-09
Add initial e2e batch request test

acuarica:
**Description**:

This PR adds acceptance tests for the batch request feature.

**Related issue(s)**: #1878 

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance/rpc_batch3.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch3.spec.ts
@@ -1878,4 +1878,28 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
       });
     });
   });
+
+  describe('Batch Request Test Suite', async function () {
+    // make sure env is set to true BATCH_REQUESTS_ENABLED
+
+    it('@release Should return a batch of requests', async function () {
+      const payload = [
+        {
+          id: 1,
+          method: RelayCall.ETH_ENDPOINTS.ETH_CHAIN_ID,
+          params: [],
+        },
+        {
+          id: 2,
+          method: RelayCall.ETH_ENDPOINTS.ETH_GAS_PRICE,
+          params: [],
+        },
+      ];
+
+      const res = await relay.callBatch(payload);
+      expect(res).to.have.length(payload.length);
+      expect(res.filter((r) => r.id === 1)[0].result).to.be.equal(CHAIN_ID);
+      expect(res.filter((r) => r.id === 2)[0].result).to.be.equal('0xa54f4c3c00');
+    });
+  });
 });
--- a/packages/server/tests/clients/relayClient.ts
+++ b/packages/server/tests/clients/relayClient.ts
@@ -55,6 +55,16 @@ export default class RelayClient {
     return result;
   }
 
+  async callBatch(payload: { id: string; method: string; params: any[] }[]) {
+    const request = this.provider._getConnection();
+    request.setHeader('content-type', 'application/json');
+    request.body = JSON.stringify(payload.map((r) => ({ ...r, jsonrpc: '2.0' })));
+    const response = await request.send();
+    response.assertOk();
+
+    return response.bodyJson;
+  }
+
   /**
    * Calls the specified methodName with the provided params and asserts that it fails
    * @param methodName


packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -1878,4 +1878,103 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
       });
     });
   });
+
+  describe('Batch Request Test Suite', async function () {
+    // make sure env is set to true BATCH_REQUESTS_ENABLED
```
Nana-EC, line 1883: Instead of comment you can just set it.
Maybe add a before and after.
See the `describe('DEV_MODE = true', ...` example

packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -1878,4 +1878,103 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
...
+        const res = await relay.call(
+          RelayCalls.ETH_ENDPOINTS.ETH_GET_TRANSACTION_COUNT,
+          [account.address, 'latest'],
+          requestId,
+        );
+        expect(res).to.be.equal('0x1');
+      }
+
+      {
+        const payload = [
```
Nana-EC, line 1960: Can move this part into it's own separate tests - `it('....`
Unless there's a reason to keep it together

packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -1878,4 +1878,103 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
       });
     });
   });
+
+  describe('Batch Request Test Suite', async function () {
+    // make sure env is set to true BATCH_REQUESTS_ENABLED
+
+    it('@release Should return a batch of requests', async function () {
```
Nana-EC, line 1885: Leave this off for now since we haven't turned it on in production, otherwise `@release` tests against deployed environments will fail

Nana-EC (changes requested):
Nice start.
Some suggestions

packages/server/tests/acceptance/rpc_batch3.spec.ts
acuarica: The goal of the 3rd step is to ensure changes made in the 2nd step are visible.

packages/server/tests/acceptance/rpc_batch3.spec.ts
acuarica: Thanks for the heads up on this

packages/server/tests/acceptance/rpc_batch3.spec.ts
Nana-EC: No docs but good point we should have it.
Basically we have a lot of tests so we use these string test flags to allow for filtering.
So during release process we run a subset of all the acceptance tests using the `@release` tag

packages/server/tests/acceptance/rpc_batch3.spec.ts
Nana-EC: I see. No-op then

AlfredoG87 (approved):
LGTM 👍

packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -1879,8 +1879,14 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
...
 
-  describe('Batch Request Test Suite', async function () {
-    // make sure env is set to true BATCH_REQUESTS_ENABLED
+  describe('Batch Request Test Suite BATCH_REQUESTS_ENABLED = true', async function () {
+    before(async () => {
+      process.env.BATCH_REQUESTS_ENABLED = 'true';
+    });
+
+    after(async () => {
+      process.env.BATCH_REQUESTS_ENABLED = 'false';
```
AlfredoG87, line 1888: normally we store the current value on a const, and the we restore it.

AlfredoG87 (approved):
LGTM

Nana-EC (approved):
LG

PR #1902, opened 2023-11-13
Enhanced the release-production workflow to support conditional checks to publish and user entered tags.

ebadiere:
Docker relelase build image fix. 

**Related issue(s)**:  #1898

Fixes #

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,7 @@
 Dockerfile
-packages/relay/node_modules/
-packages/relay/tests/
-packages/server/node_modules/
-packages/server/tests/
-node_modules/
-tools/
+**/packages/relay/node_modules/
+**/packages/relay/tests/
+**/packages/server/node_modules/
+**/packages/server/tests/
+**/node_modules/
+**/tools/
--- a/.github/workflows/release-production.yml
+++ b/.github/workflows/release-production.yml
@@ -27,6 +27,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+        with:
+          submodules: true
         
       - name: Get tag
         run: echo "TAG=${{ github.event.inputs.image_tag }}" >> $GITHUB_ENV   
--- a/.github/workflows/release-production.yml
+++ b/.github/workflows/release-production.yml
@@ -61,7 +61,7 @@ jobs:
           platforms: linux/amd64, linux/arm64          
           file: ./Dockerfile
           load: true # This loads the image into the Docker environment
-          tags: my-image:${{ env.IMAGE_TAG }}  
+          tags: ${{ env.IMAGE_TAG }}  
 
       # Push the Docker image
       - name: Push Docker image
@@ -73,7 +73,7 @@ jobs:
           platforms: linux/amd64, linux/arm64
           file: ./Dockerfile
           push: true
-          tags: my-image:${{ env.IMAGE_TAG }}                
+          tags: ${{ env.IMAGE_TAG }}                
 
   npm-package-artifact:
     runs-on: ubuntu-latest
--- a/.github/workflows/release-production.yml
+++ b/.github/workflows/release-production.yml
@@ -38,7 +38,6 @@ jobs:
         with:
           cache-from: type=gha
           cache-to: type=gha,mode=max
-          context: .
           platforms: linux/amd64, linux/arm64
           push: true
           tags: ${{ env.REGISTRY }}/${{ github.repository }}:${{env.TAG}}
--- a/.github/workflows/release-production.yml
+++ b/.github/workflows/release-production.yml
@@ -1,10 +1,13 @@
 name: Release Production Environment
 
 on:
+  push:
+    tags: [ v* ]
+
   workflow_dispatch:
     inputs:
-      should_push:
-        description: 'Set to true to push the image after building'
+      publish:
+        description: 'Set to true to publish artifacts'
         required: true
         default: 'false'
       image_tag:
@@ -28,6 +31,13 @@ jobs:
       - name: Get tag
         run: echo "TAG=${GITHUB_REF#refs/*/v}" >> $GITHUB_ENV
 
+      - name: Set default tag
+        run: echo "IMAGE_TAG=${{ env.REGISTRY }}/${{ github.repository }}:${{ env.TAG }}" >> $GITHUB_ENV
+
+      - name: Override tag if custom tag is provided
+        if: github.event.inputs.image_tag != ''
+        run: echo "IMAGE_TAG=${{ github.event.inputs.image_tag }}" >> $GITHUB_ENV        
+
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v2
         with:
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -1,7 +1,7 @@
 Dockerfile
-packages/relay/node_modules/
-packages/relay/tests/
-packages/server/node_modules/
-packages/server/tests/
-node_modules/
-tools/
\ No newline at end of file
+**/packages/relay/node_modules/
```
Nana-EC, line 2: Which package dirs does this refer to?

Nana-EC (changes requested):
Unclear which dirs under the repo this refers to.
If the goal is when the process is run by an external process then please update the PR description to make this clear and if possible add a comment in the file

.dockerignore
ebadiere: I read this to mean the `node_modules` under the relay.  My approach was to modify what was already there to use the proper syntax.  The `packages/relay/node_modules/` was already listed in the .dockerignore file.

.github/workflows/release-production.yml
```diff
@@ -33,14 +41,25 @@ jobs:
...
-      - name: Build and push images
+      # Build the Docker image
+      - name: Build Docker image
+        id: docker_build
         uses: docker/build-push-action@v2
         with:
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
-          platforms: linux/amd64, linux/arm64
+          context: .
```
Nana-EC, line 49: Why these context changes

Nana-EC (changes requested):
dispatch should be in additional, also not clear on the additional changes

.github/workflows/release-production.yml
```diff
@@ -1,8 +1,16 @@
 name: Release Production Environment
 
 on:
-  push:
-    tags: [ v* ]
+  workflow_dispatch:
```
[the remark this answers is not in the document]
ebadiere: Added `push` back.

.github/workflows/release-production.yml
```diff
@@ -1,8 +1,16 @@
...
+    inputs:
+      should_push:
```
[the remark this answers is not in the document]
ebadiere: Updated.

.github/workflows/release-production.yml
```diff
@@ -1,8 +1,16 @@
...
+        description: 'Set to true to push the image after building'
```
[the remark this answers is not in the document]
ebadiere: Updated.

.github/workflows/release-production.yml
ebadiere: I missed the original context settings when refactoring.  They are restored.

.github/workflows/release-production.yml
```diff
@@ -33,14 +41,25 @@ jobs:
...
+      # Push the Docker image
+      - name: Push Docker image
+        if: success() && github.event.inputs.should_push == 'true'
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
           push: true
-          tags: ${{ env.REGISTRY }}/${{ github.repository }}:${{env.TAG}}
+          tags: my-image:${{ github.event.inputs.image_tag }}
```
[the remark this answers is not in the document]
ebadiere: I was thinking of the custom tag.  I updated it to now define the old default tag and only use the custom tag if it has been provided.

.github/workflows/release-production.yml
```diff
@@ -4,6 +4,17 @@ on:
   push:
     tags: [ v* ]
 
+  workflow_dispatch:
+    inputs:
+      publish:
```
Nana-EC, line 9: Input unused.
You should make it so only the build process runs if `publish` is false

.github/workflows/release-production.yml
```diff
@@ -33,14 +51,29 @@ jobs:
...
+      - name: Build Docker image
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          platforms: linux/amd64, linux/arm64          
+          file: ./Dockerfile
+          load: true # This loads the image into the Docker environment
+          tags: my-image:${{ env.IMAGE_TAG }}
```
Nana-EC: Fix tag value

.github/workflows/release-production.yml
```diff
@@ -33,14 +51,29 @@ jobs:
...
+        if: success() && github.event.inputs.should_push == 'true'
         uses: docker/build-push-action@v2
         with:
           cache-from: type=gha
           cache-to: type=gha,mode=max
           platforms: linux/amd64, linux/arm64
+          file: ./Dockerfile
           push: true
-          tags: ${{ env.REGISTRY }}/${{ github.repository }}:${{env.TAG}}
+          tags: my-image:${{ env.IMAGE_TAG }}
```
Nana-EC: Why tag change?

.github/workflows/release-production.yml
```diff
@@ -20,6 +31,13 @@ jobs:
       - name: Get tag
         run: echo "TAG=${GITHUB_REF#refs/*/v}" >> $GITHUB_ENV
 
+      - name: Set default tag
```
Nana-EC, line 34: This step isn't needed if it's a version tag. You should remove or else you'll break the release flow

Nana-EC (changes requested):
Still some changes left that break release flow

.github/workflows/release-production.yml
```diff
@@ -20,6 +31,13 @@ jobs:
...
+        run: echo "IMAGE_TAG=${{ env.REGISTRY }}/${{ github.repository }}:${{ env.TAG }}" >> $GITHUB_ENV
+
+      - name: Override tag if custom tag is provided
+        if: github.event.inputs.image_tag != ''
+        run: echo "IMAGE_TAG=${{ github.event.inputs.image_tag }}" >> $GITHUB_ENV
```
Nana-EC, line 39: Easier if you use the same env variable
```suggestion
        run: echo "TAG=${{ github.event.inputs.image_tag }}" >> $GITHUB_ENV        
```

Nana-EC (changes requested):

.github/workflows/release-production.yml
ebadiere: my-image removed.

.github/workflows/release-production.yml
ebadiere: my-image removed

.github/workflows/release-production.yml
ebadiere: Updated.

.github/workflows/release-production.yml
ebadiere: Updated.

.github/workflows/release-production.yml
ebadiere: Updated.

Nana-EC (changes requested):

PR #1911, opened 2023-11-20
Productization of Environment optimization for retries against mirror node calls.

### LINKED ISSUE #1907: eth_sendRawTransaction / "No matching record found for transaction id" Issue
### Description

eth_sendRawTransaction submits the transaction and waits for the result to be propagated to the mirror node to respond, sometimes the transaction is successful but the propagation takes longer than expected resulting in the following error:

```
No matching record found for transaction id
```

### Steps to reproduce

1. Run a Local Node.
2. Set the following configurable properties to the following values:
```
MIRROR_NODE_RETRIES=0
MIRROR_NODE_RETRY_DELAY=2000
```
3. Send any transaction that uses `eth_sendRawTransaction` transfer of HBAR could be, but better if is a smart contract interaction.

4. Error should happend

### Additional context

Currently the workaround is to either increase the `MIRROR_NODE_RETRIES` or `MIRROR_NODE_RETRY_DELAY`

In the first, we are affecting all the flows where we should not retry due to `404` responses from the mirror node.
And in the second we are making all `eth_sendRawTransactions` take a minimun time defined there.

Currently I have set it to `0` and `5000` but I would like it to be trying more often. every X amount of time. so it could be as little as 2s, but as much as is needed, without affecting other endpoints...

There is another solution, that would make `eth_sendRawTransaction` be more equivalent to EVM, that is returnin the hash of the transaction as soon as is avaiable and then let `metamask` poll for the response using `eth_getTransactionReceipt` until it becomes available.

Explore both options as part of the resolution of this ticket.

### Hedera network

mainnet, testnet, previewnet

### Version

v0.35.2

### Operating system

None

AlfredoG87:
**Description**:
Productization of Environment optimization for eth_getBalance, eth_sendRawTransaction so there is no retries by default on the mirrorNodeClient.ts and also increase the retries count on the eth_sendRawTransaction flow, so receipt record is found on mirror node without affecting other request

**Related issue(s)**:

Fixes #1907 

**Notes for reviewer**:
The new configuration parameters have been running on all SwirldsLabs production environments (previewnet, testnet and mainnet) for the last few days without issues, and a noticeable decrease in latency was observed for many methods, but specially for `eth_getBalance` also, `eth_sendRawTransaction` is faster and `100%` reliable, so far.

**Checklist**


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -166,13 +166,13 @@ export class MirrorNodeClient {
     const mirrorNodeHttpMaxTotalSockets = parseInt(process.env.MIRROR_NODE_HTTP_MAX_TOTAL_SOCKETS || '300');
     const mirrorNodeHttpSocketTimeout = parseInt(process.env.MIRROR_NODE_HTTP_SOCKET_TIMEOUT || '60000');
     const isDevMode = process.env.DEV_MODE && process.env.DEV_MODE === 'true';
-    const mirrorNodeRetries = parseInt(process.env.MIRROR_NODE_RETRIES || '3');
+    const mirrorNodeRetries = parseInt(process.env.MIRROR_NODE_RETRIES || '0'); // we are in the process of deprecation this "feature" (general retries regardless of the flow)
```
ebadiere, line 169: Nit. "we are in the process of deprecating this feature"

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -166,13 +166,13 @@ export class MirrorNodeClient {
...
     const mirrorNodeRetriesDevMode = parseInt(process.env.MIRROR_NODE_RETRIES_DEVMODE || '5');
-    const mirrorNodeRetryDelay = parseInt(process.env.MIRROR_NODE_RETRY_DELAY || '250');
+    const mirrorNodeRetryDelay = parseInt(process.env.MIRROR_NODE_RETRY_DELAY || '2000');
     const mirrorNodeRetryDelayDevMode = parseInt(process.env.MIRROR_NODE_RETRY_DELAY_DEVMODE || '200');
     const mirrorNodeRetryErrorCodes: Array<number> = process.env.MIRROR_NODE_RETRY_CODES
       ? JSON.parse(process.env.MIRROR_NODE_RETRY_CODES)
-      : [404]; // by default we should only retry on 404 errors
+      : []; // we are in the process of deprecation this "feature" (general retries regardless of the flow)
```
ebadiere, line 175: Same as above

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: ✅  Done

packages/relay/src/lib/clients/mirrorNodeClient.ts
AlfredoG87: ✅  Done

ebadiere (approved):
LG.

Issue #1827, opened 2023-10-16
Relay "getBalance" returns incorrect, cached response


stefan-stefanooov:
### Description

In local node setup. (development)
When requesting the balance form the relay we hit the following function [823](https://github.com/hashgraph/hedera-json-rpc-relay/blob/b4d80db9ab40c96f640ff7875b70e2d019deb688/packages/relay/src/lib/eth.ts#L837C49-L837C49)

We get the **balance** for an account (through the relay). This **creates** a **cache entry** with the current balance (0 for example)
If we make a **transfer** to that account, **wait** for the **receipt** and we get the **balance** for it **again**, the balance is **not** correct. It does not include the latest transaction. 

This happens only in development mode if we are executing these actions consecutavly.
Like in a test example: 
```
        // Get balance
        const recipientBalanceInitial = await ethers.provider.getBalance(recipientAddr)

        // Execute transaction
        const tx = await contract.transferTo(recipientAddr, TRANSFER_AMOUNT)
        await tx.wait()

        // Get balance
        const recipientBalanceFinal = await ethers.provider.getBalance(recipientAddr)
        const diff = recipientBalanceFinal.sub(recipientBalanceInitial)

        expect(weibarTotinybar(diff)).to.equal(TRANSFER_AMOUNT)
```

Possible solution to **disable** the cache for requests with parameter `blockNumberOrTagOrHash`  set to  "latest"/"pending"

### Steps to reproduce

1. Get account balance through the relay 
2. Transfer some value to the account
3. Wait for the transaction receipt
4. get the balance for that account again (as soon as the receipt arrives)

The balance does not show the latest transaction

### Hedera network

other

### Version

latest

### Operating system

macOS

PR #1924, opened 2023-11-24
Removing cache for getBalance in dev mode and 'blockNumberOrTagOrHash' is explicit

stefan-stefanooov:
Removing cache for dev mode if blockNumberOrTagOrHash is explicitly 'latest' or 'pending'

**Related issue(s)**:
[Relay "getBalance" returns incorrect, cached response#1827](https://github.com/hashgraph/hedera-json-rpc-relay/issues/1827)

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -260,6 +260,10 @@ export class EthImpl implements Eth {
     this.debugServiceImpl = new DebugService(mirrorNodeClient, logger, this.common);
   }
 
+  private shouldUseCacheForBalance(tag: string | null): boolean {
+    return !CommonService.blockTagIsLatestOrPendingStrict(tag) && !CommonService.isDevMode;
+  }
+
   private initEthExecutionCounter(register: Registry) {
     const metricCounterName = 'rpc_relay_eth_executions';
     register.removeSingleMetric(metricCounterName);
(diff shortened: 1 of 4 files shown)


packages/relay/src/lib/services/ethService/ethCommonService/index.ts
```diff
@@ -61,6 +61,7 @@ export class CommonService implements ICommonService {
   static blockLatest = 'latest';
   static blockEarliest = 'earliest';
   static blockPending = 'pending';
+  static isDevMode = !!(process.env.DEV_MODE && process.env.DEV_MODE === 'true');
```
AlfredoG87, line 64: Why the added complexity on this assignment?

```suggestion
  static isDevMode = process.env.DEV_MODE === 'true';
```

packages/relay/src/lib/eth.ts
```diff
@@ -260,6 +260,10 @@ export class EthImpl implements Eth {
     this.debugServiceImpl = new DebugService(mirrorNodeClient, logger, this.common);
   }
 
+  private shouldUseCacheForBalance(tag: string | null): boolean {
+    return !CommonService.blockTagIsLatestOrPendingStrict(tag) && !CommonService.isDevMode;
+  }
```
AlfredoG87, line 265: Add a comment as below to improve readability.

```suggestion
  private shouldUseCacheForBalance(tag: string | null): boolean {
    // should only cache balance when is Not latest or pending and is not in dev mode
    return !CommonService.blockTagIsLatestOrPendingStrict(tag) && !CommonService.isDevMode;
  }
```

AlfredoG87 (commented):
Looks good, just 2 nits.

PR #1942, opened 2023-11-30
feat: add PR title check workflow

georgi-l95:
**Description**:
This PR adds workflow, which is responsible for checking the title of the PR and if it's compliant with the Conventional Commits standard

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/flow-pr-title-check.yml
+++ b/.github/workflows/flow-pr-title-check.yml
@@ -1,0 +1,45 @@
+##
+# Copyright (C) 2023 Hedera Hashgraph, LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


isavov (approved):

AlfredoG87 (changes requested):
Could we add a note on `README.md` with what is that standard?

Maybe under the `Contributing` section?

georgi-l95:
@AlfredoG87 Agree with you, but I think this should be included in the contribution guide, more so because all PRs in all repos will need to comply with this rule. Not sure, if we should clutter the readme here.

AlfredoG87 (approved):

Nana-EC:
> > @AlfredoG87 Agree with you, but I think this should be included in the contribution guide, more so because all PRs in all repos will need to comply with this rule. Not sure, if we should clutter the readme here.
> 
> yeah, I agree with you that this should be on contributing guide. do we have one? 😅

Yeah you can open a ticket here https://github.com/hashgraph/.github/blob/main/CONTRIBUTING.md

.github/workflows/flow-pr-title-check.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  title-check:
+    name: Title Check
+    runs-on: ubuntu-latest
+    permissions:
+      statuses: write
+    steps:
+      - name: Check PR Title
+        uses: aslafy-z/conventional-pr-title-action@v3
```
Nana-EC, line 43: Q: does this support things like checking if labels are assigned and milestones set etc?

Nana-EC (approved):

PR #1954, opened 2023-12-05
Changes validation so endpoint accepts block hash

### LINKED ISSUE #1857: Add `blockParam` to `eth_getCode`
### Problem

Add `blockParam` to `eth_getCode` according to [EIP-1898](https://eips.ethereum.org/EIPS/eip-1898).

### Solution

Extend parameters, validations, tests and all related.

### Alternatives

_No response_

konstantinabl:
**Description**:

**Related issue(s)**:

Fixes #1857 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance/rpc_batch3.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch3.spec.ts
@@ -1845,12 +1845,11 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
           DeployerContractJson.bytecode,
           testAccount.wallet,
         );
-        const deployerContract = await factory.deploy(Helper.GAS.LIMIT_15_000_000);
+        const deployerContract = await factory.deploy();
         await deployerContract.waitForDeployment();
 
         // get contract details
         const mirrorContract = await mirrorNode.get(`/contracts/${deployerContract.target}`, Utils.generateRequestId());
-
         const defaultGasPrice = numberTo0x(Assertions.defaultGasPrice);
         const defaultGasLimit = numberTo0x(3_000_000);
         const defaultTransaction = {
@@ -1873,9 +1872,7 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
           maxPriorityFeePerGas: gasPrice,
           maxFeePerGas: gasPrice,
         });
-
         transactionHash = await relay.sendRawTransaction(signedTx, requestId);
-
         await mirrorNode.get(`/contracts/results/${transactionHash}`, requestId);
 
         const res = await relay.call(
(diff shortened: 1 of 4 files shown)


natanasow (changes requested):
LGTM, one nit :rocket:

stefan-stefanooov (dismissed):

natanasow (dismissed):
LGTM

georgi-l95 (dismissed):
LG

packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -1893,6 +1893,7 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
 
     it('Should return a batch of requests', async function () {
       const testAccount = await servicesNode.createAliasAccount(100, relay.provider, Utils.generateRequestId());
+      await new Promise((r) => setTimeout(r, 5000));
```
Nana-EC, line 1896: Not a fan of sleeps, what's this for?

Nana-EC: Do you need to poll for the existence of the account instead?

Nana-EC (commented):
Validation update makes sene, would recommend changing the sleep

packages/server/tests/acceptance/rpc_batch3.spec.ts
AlfredoG87: we've added a bunch of those delays across the testing suites to ensure enough time is allowed for the mirror node to pick up the changes, I belive a polling mechanism is a better approach.

Also not a big fan of `sleeps` but is better to wait a `5s` delay and have the tests succeed, rather that waste around `10m` of testing due to the lack of.

If we go the route of doing a polling, instead of wait, it should be worked as a helper tool that can be reutilized across the testing suites, and still have a `MAX` amount of wait while polling, and then a follow-up PR for implementing it and refactoring all the acceptance testing suites that currently use the `sleep` approach.

AlfredoG87 (approved):

AlfredoG87 (dismissed):
LGTM.

packages/server/tests/acceptance/rpc_batch3.spec.ts
konstantinabl: We can create a separate task to implement polling as Alfredo suggested. So if you agree, we can close this PR and work separately on this.

ebadiere: I agree that the polling mechanism should be in another task, as it will increase the scope of this change and it's unrelated.

ebadiere (approved):
Opened a separate issue to implement the polling: https://github.com/hashgraph/hedera-json-rpc-relay/issues/1986

AlfredoG87 (approved):

ebadiere (approved):
LG.

PR #1968, opened 2023-12-08
fix: deprecated DEV_MODE

### LINKED ISSUE #1964: Deprecate DEV_MODE
### Problem

At the moment we have a Dev mode in the relay, which was used before estimateGas was implemented properly, to return the revert reason, when contract calls revert. Currently estimateGas is working correctly, which means that we don't need dev mode anymore.

### Solution

Remove all dev mode related functionality. 

### Alternatives

_No response_

ebadiere:
**Related issue(s)**:

Fixes #1964


packages/relay/src/lib/eth.ts
```diff
@@ -1747,15 +1747,6 @@ export class EthImpl implements Eth {
     const fromAddress = await this.resolveEvmAddress(contractResult.from, requestIdPrefix, [constants.TYPE_ACCOUNT]);
     const toAddress = await this.resolveEvmAddress(contractResult.to, requestIdPrefix);
 
-    if (
```
Nana-EC, line 1750: Q: Where the test being removed or added to affirm the change being made?
In a sense shouldn't an unset or `false` dev mode act exactly the same as this code being removed?

Nana-EC (changes requested):
Simple fix but clarifying implications

PR #1978, opened 2023-12-13
feat: Add option to configure file chunk size.

### LINKED ISSUE #1977: Add an option to configure the file append chunk size for the HAPI FileAppendTransaction.
### Problem

The current file append chunk size is hard coded to 4096 bytes.  Increasing this may improve performance and reduce costs.

### Solution

Add a configurable property to adjust this value.

### Alternatives

_No response_

ebadiere:
Adds an option to configure the file append chunk size.
**Related issue(s)**:

Fixes #1977 

**Checklist**


AlfredoG87 (dismissed):
LGTM.
Just one question?
How did we calculate the current ideal/default values of 20 and 5120 respectively?
I see 20, was already there but the chunk size is bigger by 1024


I found [this](https://docs.hedera.com/hedera/sdks-and-apis/sdks/transactions): 

Based on the text below I guess we should be fine.
> Transactions have a 6,144 bytes size limit. This includes the signatures on the transaction. The estimated single signature size is about 80-100 bytes.

Nana-EC (dismissed):

AlfredoG87 (approved):

AlfredoG87 (approved):

PR #1979, opened 2023-12-13
fix: Removed the contract revert reason from the eth_getTransactionByHash.

### LINKED ISSUE #1964: Deprecate DEV_MODE

ebadiere:
Removes the contract revert reason from the eth_getTransactionByHash.

**Related issue(s)**:

Fixes #1964


AlfredoG87 (approved):
LGTM, are there associated tests that also need to be removed?

AlfredoG87 (dismissed):
LGTM

AlfredoG87 (approved):
LGTM

Issue #2013, opened 2024-01-05
Add polling to the relay tests in order to stabilize them.


ebadiere:
### Description

There is a lag between when a transaction is executed on the consensus node vs when it is available in the mirror node.  To overcome this a polling mechanism was added in the smart contracts tests.  Apply a similar solution for the relay tests.

### Steps to reproduce

1. Run the `acceptancetest:api_batch3` tests against the local node.
2. Notice the number of test failures.
Most of these pass when the test run is slowed down.  

### Hedera network

other

### Version

v0.39.0

### Operating system

macOS

PR #2028, opened 2024-01-09
Extracted eth_call to helper to support polling.

ebadiere:
**Description**:
Adds polling to avoid using hardcoded delays.

**Related issue(s)**: #2013 

Fixes #


packages/server/tests/helpers/utils.ts
```diff
@@ -150,4 +152,19 @@ export class Utils {
     }
     return res.toString();
   };
+
+  static ethCall = async (
+    relay: RelayClient,
+    callData: { from: string; to: any; gas: string; data: string },
+    requestId: string,
```
Nana-EC, line 159: Add block as a param, right now you've defaulted to `latest` with no way to modify

packages/server/tests/helpers/utils.ts
```diff
@@ -150,4 +152,19 @@ export class Utils {
     }
     return res.toString();
   };
+
+  static ethCall = async (
```
Nana-EC, line 156: Update name to make clear there's a retry e.g ethCallWRetries or something.
We want to make sure people use this method only when they want retries

Nana-EC (commented):

packages/server/tests/helpers/utils.ts
ebadiere: Done.

packages/server/tests/helpers/utils.ts
ebadiere: Done.

packages/server/tests/helpers/utils.ts
```diff
@@ -150,4 +153,39 @@ export class Utils {
...
+      await new Promise((r) => setTimeout(r, 2000));
+      res = await relay.call(RelayCall.ETH_ENDPOINTS.ETH_CALL, [callData, blockNumber], requestId);
+      numberOfCalls++;
+    }
+    return res;
+  };
+
+  // Handles the odd case where the contract is loading in the factory but for some reason
+  // ethers is not able to deploy it.  The ethers deploy returns a 4001 error.
+  static deployContract = async (contractFactory: ContractFactory): Promise<ethers.Contract> => {
```
Nana-EC, line 175: This method will indefinitely retry under the right circumstances.
You should limit the number if retries and update the method name appropriately.

Nana-EC (changes requested):

packages/server/tests/helpers/utils.ts
ebadiere: Fixed.

ebadiere:
Rebased too many changes.  Cherry-picked relevant commits into a new PR: https://github.com/hashgraph/hedera-json-rpc-relay/pull/2031

PR #2031, opened 2024-01-11
test: 2013 add polling to the relay stabilize

### LINKED ISSUE #2013: Add polling to the relay tests in order to stabilize them.

ebadiere:
Added polling to stabilize tests.

**Related issue(s)**:

Fixes #2013


Nana-EC (dismissed):

AlfredoG87 (approved):
LGTM

PR #2033, opened 2024-01-11
fix:  Added 20% deviation in gasPricing.

### LINKED ISSUE #2010: Tests fail during the business day with `Assert transactionReceipt: 'effectiveGasPrice' should equal mirrorNode response`
### Description

One of the failing tests is `@release should execute "eth_getTransactionReceipt" for hash of London transaction` which calls `Assertions.transactionReceipt(res, mirrorResult, currentPrice);` which eventually checks the effect gas price with the response from the mirror node.  Other tests use this code and seem to fail mainly during the US business day.

### Steps to reproduce

1. Point the tests to testnet.  This seems to only happen on testnet.
2. Run the release tests, `npm run acceptancetest:release`.  This will take some time to complete.  You can isolate the `@release should execute "eth_getTransactionReceipt" for hash of London transaction` and use that test to investigate.  
3. These are flaky tests, so they will pass at times.  You may need to run the whole suite at different times of the day or just the single test.  You may need to experiment a bit here.

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.39.0

### Operating system

macOS

ebadiere:
**Description**:
This adds a 20% deviation in the gas price.  Gas prices on testnet vary during the business day.

**Related issue(s)**:

Fixes #2010

**Checklist**


packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -718,7 +719,9 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
           requestId,
         );
 
-        expect(receiptFromRelay.effectiveGasPrice).to.equal(prepend0x(currentPrice.toString(16)));
+        // handle deviation of 20% in gas price
+        expect(receiptFromRelay.effectiveGasPrice).to.be.lessThan(Number(prepend0x(currentPrice)) * 1.2);
```
Nana-EC, line 723: Instead of 20% with hardcoded 1.2 and 0.8 why not add a test config that allows us to set the % deviation as a decimal

Nana-EC (commented):
Can do in the form of a test config for easier management.
Additionally not the config in test documentation where appropriate

packages/server/tests/acceptance/rpc_batch1.spec.ts
ebadiere: The effectiveGasContract.spec.ts has the % deviation hardcoded.  I'll make the update.

docs/configuration.md
```diff
@@ -167,6 +167,8 @@ Unless you need to set a non-default value, it is recommended to only populate o
 | `E2E_RELAY_HOST` | ""      | Remote relay url to point to.                                |
 | `DEV_MODE`       | "false" | Flag if relay should operate in developer optimization mode. |
 | `TEST_WS_SERVER` | "false" | Flag config for enable or disable the WS server tests.       |
+| `LOWER_GAS_PRICE_DEVIATION` | 0.80 | Lower value to use when comparing gas prices in the rpc-batch1.spec.ts |
```
Nana-EC, line 170: Q: should we instead make it one single config that gives it a +/- feature. So in this case the deviation is 20% i.e. 0.2

Nana-EC, line 170: Also let's use a `TEST_` prefix to not this is test config not product config

docs/configuration.md
```diff
@@ -207,8 +209,12 @@ CLIENT_TRANSACTION_RESET= 400000
...
 
 > **_NOTE 2:_**: Read more about `DEV_MODE` [here](./dev-mode.md)
+
+> **_NOTE 3:_**: Hedera transaction fees are set in fiat(USD) but paid using the gas price in HBAR.  The LOWER_GAS_PRICE_DEVIATION and UPPER_GAS_PRICE_DEVIATION allow for a range of gas prices reflecting the current market rates, when testing the current gas price in the acceptance tests.  Read more about transaction fees [here](https://hedera.com/fees)
```
Nana-EC, line 220: Let's add a few lines in here that better explain how this connection can cause the gas price to change on the network.

Nana-EC (changes requested):
Some improvement suggestions

docs/configuration.md
ebadiere: Agreed. Done.

ebadiere: Done.

docs/configuration.md
ebadiere: Done.

AlfredoG87 (dismissed):
LGTM 👍

packages/server/tests/testnetAcceptance.env
```diff
@@ -14,3 +14,4 @@ HAPI_CLIENT_ERROR_RESET=[50]
 SUBSCRIPTIONS_ENABLED=true
 FILTER_API_ENABLED=true
 DEBUG_API_ENABLED=true
+TEST_GAS_PRICE_DEVIATION=0.2
```
AlfredoG87, line 17: nit: blank line at EOF

AlfredoG87 (approved):

PR #2042, opened 2024-01-16
fix: nonce too low precheck relaxed to allow a buffer

### LINKED ISSUE #2043: When Mirror-node out of sync with account `nonce`, unable to set it below the expected by the mirror node
### Description

I found a case where I was unable to send HBar or perform any actions using Metamask for a given account on `previewnet` due to the error :

```
"[Request ID: 2e952947-ffad-408b-aed9-35f3ed152001] Nonce too low. Provided nonce: 15, current nonce: 15"
```

When I tried to increase it also got error:
```
"[Request ID: 3ec591b4-9396-49f4-a03f-06c415a7cc6a] Nonce too high. Provided nonce: 16, current nonce: 15"
```

Using the SDK I found that the correct Nonce was `14`, however when setting it manually to `14` got the following error:
```
Nonce too low. Provided nonce: 14, current nonce: 15
```

Upon further research, realized that this was due to a `PRECHECK` validation and was not reaching the EVM on Hedera Services, hence I had a blocked account on previewnet.

### Steps to reproduce

This is hard to reproduce, since for an account to be out of sync with the `mirror-node` some edge cases need to happen, however this is the previewnet account where I found the issue, but since I worked on a fix, is now resolved. 😅 

0.0.1028

### Additional context

_No response_

### Hedera network

previewnet

### Version

v0.39.0

### Operating system

Other

AlfredoG87:
**Description**:
Added a configurable property called `NONCE_PRECHECK_BUFFER` defaulted to `1`, so when precheck verifies for `nonce` too low, it allows a buffer or range to be out of sync, in case there is an issue with the mirror node reported nonce and the user is setting it manually after looking it up using the sdk.

**Related issue(s)**:

Fixes #2043

**Notes for reviewer**:

**Checklist**


ebadiere (approved):
LG.

PR #2048, opened 2024-01-17
fix: Removed prechecks for existing accounts. (#2046)

### LINKED ISSUE #2039: Remove pre-checks for non-existing entities
### Problem

In order to run full ethereum equivalence, we should be able to call non-existing addresses/contracts and these calls should result in SUCCESS with gas-charged as expected (21k base + N for the data field).

### Solution

Remove these pre-checks for non-existing entities:
- for the `from` param - [here](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1587)
- for the `to` param - [here](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1599)

### Alternatives

_No response_

ebadiere:
* fix: Removed prechecks for existing accounts.

* fix: Updated copyrights

* fix: Updated copyright.

**Related issue(s)**:

Fixes #2039 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -2,7 +2,7 @@
  *
  * Hedera JSON RPC Relay
  *
- * Copyright (C) 2023 Hedera Hashgraph, LLC
+ * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


AlfredoG87 (approved):
LG

PR #2051, opened 2024-01-19
fix: remove unneeded applying of gas on

### LINKED ISSUE #2050: Not providing gas on `eth_call` fails the call sometimes
### Description

When trying to make `eth_call`, but not providing a gas, this results in automatically assigned from the relay around 400k gas, which is incorrect in most cases. Even if the mirror node receives a call without specified gas, it estimates and add the correct one.

### Steps to reproduce

1. Make `eth_call` without specifying gas for a call, which requiers more than 400k gas.

### Additional context

_No response_

### Hedera network

### Version

main

### Operating system

None

georgi-l95:
**Description**:
This PR modifies the current eth_call behavior when no gas is provided in the call object.
We are currently setting incorectly a default gas value, when non is provided. Which is not correct, when the call needs more than the provided. 
It's also not needed, because mirror-node is already doing this internally, if non is provided.

**Related issue(s)**:

Fixes #2050 

**Notes for reviewer**:
Example case:

So let’s say you have a contract with method which does some action, and this is estimated to consume 1 milion gas (for example).
If you do eth_call and in the transaction object you specify gas, we will redirect this gas to the mirror node and your call will succeed or fail depending, whether this gas is higher or lower than 1 milion.
If you don’t specify gas, in the relay we assign you predefined (in the env. variables) hardcoded gas, which by default is i think 400k. So this will always fail your call.
It’s not needed for us to do this in the relay, because when this request arrive in the mirror node and it doesn’t have gas in the payload, mirror node will automatically estimate and assign. 

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1458,7 +1458,6 @@ export class EthImpl implements Eth {
     this.logger.trace(
       `${requestIdPrefix} call({to=${call.to}, from=${call.from}, value=${call.value}, gas=${call.gas}, ...}, blockParam=${blockParam})`,
     );
-
     if (call.data?.length >= constants.FUNCTION_SELECTOR_CHAR_LENGTH)
       this.ethExecutionsCounter
         .labels(EthImpl.ethCall, call.data.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH))
@@ -1495,7 +1494,7 @@ export class EthImpl implements Eth {
 
   async callMirrorNode(
     call: any,
-    gas: number,
+    gas: number | null,
     value: string | null,
     requestIdPrefix?: string,
   ): Promise<string | JsonRpcError> {
@@ -1510,8 +1509,8 @@ export class EthImpl implements Eth {
       );
       callData = {
         ...call,
-        gas,
-        value,
+        ...(gas !== null ? { gas } : {}), // Add gas only if it's not null
+        ...(value !== null ? { value } : {}),
         estimate: false,
       };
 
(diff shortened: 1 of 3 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1566,8 +1565,12 @@ export class EthImpl implements Eth {
    * @param gas
    * @param requestIdPrefix
    */
-  async callConsensusNode(call: any, gas: number, requestIdPrefix?: string): Promise<string | JsonRpcError> {
+  async callConsensusNode(call: any, gas: number | null, requestIdPrefix?: string): Promise<string | JsonRpcError> {
     // Execute the call and get the response
+    if (!gas) {
+      gas = Number.parseInt(this.defaultGas);
```
AlfredoG87, line 1571: So we do need to set gas for calls to consensus node, but not mirror node?

AlfredoG87 (approved):
LG 👍

AlfredoG87 (changes requested):
Should we add a test case to this case? When Gas is not supplied.

AlfredoG87:
> Should we add a test case to this case? When Gas is not supplied.

I see that we already have one, but then maybe we can assert that the mirror node also receives null?

packages/relay/src/lib/eth.ts
georgi-l95: Not sure if it's needed. But we ultimatly have to remove the fallback to consensus at one point, so it doesn't really matter.

AlfredoG87: agreed 👍

AlfredoG87 (approved):
LGTM

PR #2056, opened 2024-01-23
docs: add verification instructions in Hardhat example

### LINKED ISSUE #2054: Include verification setup in Hardhat example
### Problem

The Hardhat example does not provide instructions nor setup to verify the contract. It would be useful to devs to have an end-to-end solution to build, test, deploy and **verify** their smart contracts on Hedera.

### Solution

Include verification instructions in the Hardhat example.

### Alternatives

_No response_

acuarica:
Fixes 
- #2054.

> The Hardhat example does not provide instructions nor setup to verify the contract. It would be useful to devs to have an end-to-end solution to build, test, deploy and verify their smart contracts on Hedera.

**Description**:

This PR adds instructions to setup the verification of deployed smart contracts in a custom Sourcify instance. To do so, it includes a new `env` key to set up the user private key on _testnet_ and updates Hardhat dependencies. More specifically, `ethers.js`

There are breaking changes when migrating `ethers.js` _v5_ to _v6_ (this is because we had to update `hardhat-toolbox` for the custom verification to work).

- `getBalance` needs to be invoked from the `provider`
- Use `target` instead of `address` to get the address of a contract
- Use `METHOD.staticCall` instead of `callStatic.METHOD`
- Use `getCreateAddress` instead of `getContractAddress`
- Use `logs` instead of `events` to get the emitted events in a receipt
- Use native BigInt instead of BigNumber

Some of these changes can be found in the official `ethers.js` docs https://docs.ethers.org/v6/migrating/.

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/tools/hardhat-example/.env.example
+++ b/tools/hardhat-example/.env.example
@@ -4,3 +4,5 @@ RECEIVER_PRIVATE_KEY=0x2e1d968b041d84dd120a5860cee60cd83f9374ef527ca86996317ada3
 RELAY_ENDPOINT='http://localhost:7546'
 
 MNEMONIC="here is where your twelve words mnemonic should be put my friend"
+
+TESTNET_OPERATOR_PRIVATE_KEY='<if you want to deploy on testnet go to https://portal.hedera.com/ to setup your private key>'
--- a/tools/hardhat-example/.env.example
+++ b/tools/hardhat-example/.env.example
@@ -5,4 +5,5 @@ RELAY_ENDPOINT='http://localhost:7546'
 
 MNEMONIC="here is where your twelve words mnemonic should be put my friend"
 
-TESTNET_OPERATOR_PRIVATE_KEY='<if you want to deploy on testnet go to https://portal.hedera.com/ to setup your private key>'
+# If you want to deploy on testnet go to https://portal.hedera.com/ to setup your private key
+TESTNET_OPERATOR_PRIVATE_KEY=0x0000000000000000000000000000000000000000000000000000000000000000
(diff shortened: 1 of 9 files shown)


tools/hardhat-example/.env.example
```diff
@@ -4,3 +4,5 @@ RECEIVER_PRIVATE_KEY=0x2e1d968b041d84dd120a5860cee60cd83f9374ef527ca86996317ada3
 RELAY_ENDPOINT='http://localhost:7546'
 
 MNEMONIC="here is where your twelve words mnemonic should be put my friend"
+
+TESTNET_OPERATOR_PRIVATE_KEY='<if you want to deploy on testnet go to https://portal.hedera.com/ to setup your private key>'
```
AlfredoG87, line 8: I like this 👍

AlfredoG87 (approved):
LGTM

AlfredoG87:
Looks like there is an issue with the key, it is expecting 32 bytes:

>  Invalid account: #0 for network: testnet - private key too long, expected 32 bytes

tools/hardhat-example/.env.example
AlfredoG87: Althought I think this is breaking the CI

Nana-EC: Q: why is this needed vs just changing your `OPERATOR_PRIVATE_KEY`
The current value assumes local node per the comment but if you changed networks it makes sense to change value no?

acuarica: Yes and no. We could indeed use the same key. But having different ones allows the user to deploy to different networks at the same time without changing the configuration or the env file.

acuarica:
Hardhat example CI job has been fixed and all other CI jobs now pass, pls take a look.

Nana-EC (approved):

PR #2057, opened 2024-01-23
test: Improve Dapp-example Acceptance Test completion speed.

AlfredoG87:
**Description**:
Lowered the time it needs to wait for `eth_getTransactionCount` so the right nonce gets populated from 10 to 2 seconds, considering that the avg request takes around 200ms, I think 2 seconds is enough time for propagation and also timeout.

I have tested this agains both `previewnet` and `testnet` acceptance tests while deploying `0.40` version and testing it.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Before: 7minutes 44 seconds**
image

**After: 3 minutes, 6 seconds**
image

**Checklist**


AlfredoG87:
In local node as well as testnet found some issues, closing this.

PR #2083, opened 2024-02-06
fix: Added mirror node retries.

ebadiere:
Adding mirror node retries.

**Related issue(s)**: #2080

Fixes #


packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -1012,6 +1012,8 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
       requestId = Utils.generateRequestId();
       reverterContract = await servicesNode.deployContract(reverterContractJson);
       // Wait for creation to propagate
+      await new Promise((r) => setTimeout(r, 7000));
```
Nana-EC, line 1015: Not a fan of timeouts, also at this point no value in changing a release branch with this logic.

Nana-EC (changes requested):
Not sure there's value in this PR.
The issues encountered weren't feature release blocking and timeouts aren't a great solution.
I'd say just cancel the PR and take any improvements in the `main` branch at this point

PR #2097, opened 2024-02-12
chore: update local node

georgi-l95:
**Description**:
This PR updates local-node to latest version and bumps all used images as well.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/helpers/prerequisite.ts
+++ b/packages/server/tests/helpers/prerequisite.ts
@@ -11,16 +11,11 @@ const RELAY_URL = process.env.E2E_RELAY_HOST || LOCAL_RELAY_URL;
 
 (function () {
   if (USE_LOCAL_NODE) {
-    const networkNodeImage = '0.47.0-alpha.4';
-    const mirrorNodeImage = '0.98.0-rc1';
-
-    console.log(`Docker container versions, services: ${networkNodeImage}, mirror: ${mirrorNodeImage}`);
-
     console.log('Installing local node...');
     shell.exec(`npm install @hashgraph/hedera-local -g`);
 
     console.log('Starting local node...');
-    shell.exec(`hedera start -d --network-tag=${networkNodeImage} --mirror-tag=${mirrorNodeImage} --verbose=trace`);
+    shell.exec(`hedera start -d --verbose=trace`);
     console.log('Hedera Hashgraph local node env started');
   }
 
--- a/packages/server/tests/helpers/prerequisite.ts
+++ b/packages/server/tests/helpers/prerequisite.ts
@@ -11,19 +11,16 @@ const RELAY_URL = process.env.E2E_RELAY_HOST || LOCAL_RELAY_URL;
 
 (function () {
   if (USE_LOCAL_NODE) {
-    process.env['NETWORK_NODE_IMAGE_TAG'] = '0.45.0-alpha.0';
-    process.env['HAVEGED_IMAGE_TAG'] = '0.45.0-alpha.0';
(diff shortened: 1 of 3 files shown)


AlfredoG87 (dismissed):
Nice, do you know if this new versions fix the `nonce out of sync` issues we were experiencing?

packages/server/tests/helpers/prerequisite.ts
```diff
@@ -11,19 +11,16 @@ const RELAY_URL = process.env.E2E_RELAY_HOST || LOCAL_RELAY_URL;
 
 (function () {
   if (USE_LOCAL_NODE) {
-    process.env['NETWORK_NODE_IMAGE_TAG'] = '0.45.0-alpha.0';
-    process.env['HAVEGED_IMAGE_TAG'] = '0.45.0-alpha.0';
-    process.env['MIRROR_IMAGE_TAG'] = '0.92.0';
+    const networkNodeImage = '0.47.0-alpha.4';
```
Nana-EC, line 14: Q: why no longer process env variables.
Doing it as is means we can't override without code changes

Nana-EC (changes requested):
Question about hard coding

georgi-l95:
@AlfredoG87 Yep, this new version have the needed changes.

packages/server/tests/helpers/prerequisite.ts
georgi-l95: Removed them, local node will be up to date with the latest images always, so no need here to do anything.

packages/server/tests/helpers/prerequisite.ts
```diff
@@ -11,19 +11,11 @@ const RELAY_URL = process.env.E2E_RELAY_HOST || LOCAL_RELAY_URL;
...
 (function () {
   if (USE_LOCAL_NODE) {
-    process.env['NETWORK_NODE_IMAGE_TAG'] = '0.45.0-alpha.0';
-    process.env['HAVEGED_IMAGE_TAG'] = '0.45.0-alpha.0';
-    process.env['MIRROR_IMAGE_TAG'] = '0.92.0';
-
-    console.log(
-      `Docker container versions, services: ${process.env['NETWORK_NODE_IMAGE_TAG']}, mirror: ${process.env['MIRROR_IMAGE_TAG']}`,
-    );
-
```
AlfredoG87, line 21: I think there are pros and cons of having the latest images always. we did have a regression lately due to that fact, the good thing is that it helped us to find issues faster and sooner, the bad is that it blocked the CI pipeline for a couple days.

@Nana-EC @ebadiere, What are your thoughts on this?

I personally prefer to get latests the sooner so we catch potencial issues earlier.

AlfredoG87 (commented):
Leave a comment to start a discussion.

packages/server/tests/helpers/prerequisite.ts
Nana-EC: Same thoughts.
Latest is great but it takes control away from us.
If there's a way to have a back up config whereby if local node had a version different from what we want to test that would be good.
I think there's such a flow

georgi-l95: If it's blocked we can just add wanted tag versions in the yml file, until issue is resolved.

Nana-EC: I think let's get this in as noted.
@ebadiere is working on the Configurable Manual Tests that can help with concerns of running targeted tests on specific versions

AlfredoG87 (approved):

Nana-EC (approved):

PR #2102, opened 2024-02-14
fix: Issue 2086, turning back on tests.

### LINKED ISSUE #2086: Both network node and mirror node images need to have HIP-844 changes in place to handle wallet nonce values properly.
### Description

Currently the batch api3 test `should be able to debug a failing CREATE transaction of type Legacy with call depth and onlyTopCall false` explicitly fails a transaction causing subsequent transactions to fail with `"Nonce too low. Provided nonce: 6, current nonce: 6"`. This test will be disabled until the network node updates are available in the local node.

### Steps to reproduce

1. Push up a PR to trigger CI.
2. Notice the api batch3 tests don't pass, blocking the merge.

### Additional context

_No response_

### Hedera network

other

### Version

v0.39.0

### Operating system

macOS

ebadiere:
Turn back on tests.

**Related issue(s)**:

Fixes #2086

### DIFF
--- a/packages/server/tests/acceptance/rpc_batch3.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch3.spec.ts
@@ -1172,9 +1172,7 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
           expect(resultDebug.calls).to.have.lengthOf(1);
         });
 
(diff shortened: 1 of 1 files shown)


AlfredoG87 (approved):
LGTM

PR #2123, opened 2024-02-20
fix: Enables getStorageAt to accept blockHashes

### LINKED ISSUE #2107: `eth_getStorageAt` doesn't support block hashes
### Problem

`eth_getStorageAt` currently supports only block number or alias as parameter, it should support block hash as well

### Solution

- Allow the use of `eth_getStorageAt` with a block hash as parameter

### Alternatives

_No response_

konstantinabl:
**Description**:

This pull request introduces a modification to the relay validator, enabling it to accommodate block hashes within the context of the getStorageAt endpoint. The primary focus of this update is to align the functionality with the underlying logic, which inherently supports block hashes.
Additionally, new tests were added, including adjustments to existing tests to ensure consistency.

**Related issue(s)**:

Fixes #2107

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -439,7 +439,7 @@
                     "name": "Block",
                     "required": false,
                     "schema": {
-                        "$ref": "#/components/schemas/BlockNumberOrTag"
+                        "$ref": "#/components/schemas/BlockNumberOrTagOrHash"
                     }
                 }
             ],
@@ -1441,6 +1441,23 @@
                     {
                         "title": "Block tag",
                         "$ref": "#/components/schemas/BlockTag"
+                    }
+                ]
+            },
+            "BlockNumberOrTagOrHash": {
+                "title": "Block number or tag",
+                "oneOf": [
+                    {
+                        "title": "Block number",
+                        "$ref": "#/components/schemas/uint"
+                    },
+                    {
+                        "title": "Block tag",
+                        "$ref": "#/components/schemas/BlockTag"
+                    },
+                    {
+                        "title": "Block hash",
+                        "$ref": "#/components/schemas/hash32"
                     }
                 ]
             },
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -793,20 +793,20 @@ export class EthImpl implements Eth {
   async getStorageAt(
     address: string,
     slot: string,
-    blockNumberOrTag?: string | null,
+    blockNumberOrTagOrHash?: string | null,
     requestIdPrefix?: string,
   ): Promise<string> {
     this.logger.trace(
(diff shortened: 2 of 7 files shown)


georgi-l95 (approved):
LG 👍

stefan-stefanooov (approved):

docs/openrpc.json
```diff
@@ -1444,6 +1444,23 @@
...
+                        "$ref": "#/components/schemas/uint"
+                    },
+                    {
+                        "title": "Block tag",
+                        "$ref": "#/components/schemas/BlockTag"
+                    },
+                    {
+                        "title": "Block hash",
+                        "$ref": "#/components/schemas/hash32"
+                    }
```
AlfredoG87, line 1461: According to the [EIP 1898](https://eips.ethereum.org/EIPS/eip-1898) blockParam can either be a String or an Object. 
If is a String, it can either be a BlockTag or Number. 
ex: "earliest", "latest", "0x1"
If is an Object, it can contain a BlockNumber or a BlockHash Field. 
```json
{
   "blockHash": "0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3"
}

or

{ 
    "blockNumber":"0x1"
}
```

I don't think that it supports a blockHash in case of receiving a single string as is implied on this PR.

AlfredoG87 (changes requested):
LG, however I think the schema is not entirely correct

docs/openrpc.json
konstantinabl: @AlfredoG87 You are right about this. According to this EIP hash could be an object or string. Currently, we support blockParam for eth_call only, in other relays like Alchemy they also support blockHash as an object on eth_call and eth_estimateGas, but not on other endpoints. 
Not sure if we should introduce blockParam for all endpoints or jsut to follow alchemy and do it only for estimateGas
@Nana-EC What do you think?

Nana-EC: @konstantinabl I'd say keep it simple and targeted for now.
So in this PR focus on eth_getStorageAt issue and resolve it.
@AlfredoG87 I think you added the logic to handle string or object, we should make sure that is in a shared place so all endpoints can use it easily as need be.
It looks like the [EIP 1898](https://github.com/hashgraph/hedera-json-rpc-relay/issues/822) support wasn't completely resolved as we thought.
We can add it to each endpoint as we need so we're intentional about it.
Can someone re-open that ticket and update the Epic to capture all the applicable endpoints and the remaining work so be done (string or object support).
We can tackle the work in a future sprint.

AlfredoG87 (approved):
LGTM

docs/openrpc.json
konstantinabl: @Nana-EC thanks, im merging it then :)

Issue #2117, opened 2024-02-16
Enhance the `eth_subscribe` implementation to support `newHeads`


ebadiere:
### Problem

The relay's `eth_subscribe` implementation currently does not support `newHeads`.

### Solution

Enhance the `eth_subscribe` implementation to support the `newHeads`

- [x] [Enhance](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2147) the webSocketServer to handle `newHeads`.
- [x] [Enhance](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2148) the `poller.ts` to handle the `newHeads` call.
- [x] [Add `wscat`/`WebSocket`](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2149) level tests to cover the `newHeads` call.
- [x] [Enhance](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2181) the webSocketServer to support necessary JSON RPC calls.
- [x] [Add](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2150) Hardhat tests to cover polling for new blocks.
- [x] [Add](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2151) ethers.js tests.
- [x] [Add ](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2168)metrics tracking polls.

Nana-EC:
@ebadiere is there anything left over here or can this be closed?

PR #2145, opened 2024-02-23
feat: Design update to support polling for new blocks, newHeads.

ebadiere:
Design document update.

**Related issue(s)**: #2117

Fixes #


docs/configuration.md
```diff
@@ -113,6 +113,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
 | `WS_CONNECTION_LIMIT_PER_IP`    | "10"     | Maximum amount of connections from a single IP address                                                                        |
 | `WS_MULTIPLE_ADDRESSES_ENABLED` | "false"  | If enabled eth_subscribe will allow subscription to multiple contract address.                                                |
 | `WS_CACHE_TTL`                  | "20000"  | The time to live for cached entries.                                                                                          |
+| `WS_NEW_HEADS_ENABLED`.         | "true".  | Enables subscriptions for the lates blocks, `newHeads`.                                                                       |
```
Nana-EC, line 116: Default should be false until we've tested

docs/design/eth-subscribe.md
```diff
@@ -200,6 +206,15 @@ if (success) {
 
 3. Create a scheduled job that loops through all saved subscriptions and calls the Mirror node Rest API and sends the response to all `subscribers`. It should be possible to configure the time interval. A throttling mechanism can be implemetented that limits the consecutive calls to the Mirror node, and at every interval only X subscriptions with the oldest `lastUpdated` should be polled.
 4. Whenever a connection is closed all corresponding subscriptions should be deleted from memory.
+   
+### Poll for Blocks, `newHeads` Implementation 
+The `newHeads` implementation simply builds on the existing mechanism for polling for logs.  The differences with the polling for logs is as follows:
+1. Add a condition check for `newHeads`, `validateSubscribeEthNewHeads`, and a `newHeads` subscription to the `webSocketServer.ts`.
+2. Add a call to the relay subscribe with the socket, event, and filter. 
+3. Add a call to the mirror node in the `poller.ts` to handle the `newHeads` event by calling `eth_getBlockByNumber` for `latest`.
```
Nana-EC, line 214: Is the logic here poll mirror node for latest block number and then repeat.
How do we ensure there's no missed block?

Nana-EC: Looking at logs the issue I'm getting at may be the same for both.
Let's see

Nana-EC (commented):
Please break out PR.
I see a config doc update then a new sever point etc.
Let's make PRs targeted and easy to indecently review, test and merge

docs/configuration.md
ebadiere: Done.

docs/design/eth-subscribe.md
ebadiere: Initial runs didn't show missing or duplicate blocks, and there is a planned test to cover that case, however I could see the behavior changing under load.  Load testing is also planned.  Is there an issue with missing or duplicated logs now?

packages/server/tests/localAcceptance.env
```diff
@@ -20,3 +20,4 @@ DEBUG_API_ENABLED=true
 SEND_RAW_TRANSACTION_SIZE_LIMIT=131072
 BATCH_REQUESTS_ENABLED=true
 TEST_GAS_PRICE_DEVIATION=0.2
+WS_NEW_HEADS_ENABLED=true
```
AlfredoG87, line 23: missing blank line EOF

packages/relay/src/lib/poller.ts
```diff
@@ -76,6 +76,12 @@ export class Poller {
           this.logger.error(`${LOGGER_PREFIX} Polling for unsupported event: ${event}. Tag: ${poll.tag}`);
         }
 
+        if (event === 'newHeads' && process.env.WS_NEW_HEADS_ENABLED === 'true') {
```
AlfredoG87, line 79: is there a test case when `WS_NEW_HEADS_ENABLED=false`?

packages/ws-server/src/webSocketServer.ts
```diff
@@ -175,71 +176,84 @@ app.ws.use(async (ctx) => {
...
+                  event,
+                ));
+              }
+              break;
+            case constants.SUBSCRIBE_EVENTS.NEW_PENDING_TRANSACTIONS:
+              response = jsonResp(request.id, predefined.UNSUPPORTED_METHOD, undefined);
+              break;
+
+            default:
+              response = jsonResp(request.id, predefined.UNSUPPORTED_METHOD, undefined);
```
AlfredoG87, line 224: is hard to read due to the nested switch block, would be possible to do something like:
```
switch(method) {
  case: ETH_SUBSCRIBE: 
    response = handleEthSubscribe(params...)
  case: ETH_UNSUBSCRIBE:
    response = handleEthUnsubscribe(params)
   ...
 }
```

}

AlfredoG87 (approved):
Looking good, leaved some nits.

AlfredoG87 (changes requested):
meant to request changes 😅.
leaved some nits on previous review, but would love to see the refactor and method extraction of the nested switch if possible.

also, seems like tests are missing, but naturally since is still a draft.

what happens or what should happen when you are already subscribed to NEW_HEADS and then you attempt to subscribe **again** ? should we do a test case for this too?

packages/ws-server/src/webSocketServer.ts
```diff
@@ -175,71 +176,84 @@ app.ws.use(async (ctx) => {
...
+              if (process.env.WS_NEW_HEADS_ENABLED === 'true') {
+                ({ response, subscriptionId } = subscribeToNewHeads(
+                  filters,
+                  response,
+                  request,
+                  subscriptionId,
+                  ctx,
+                  event,
+                ));
+              }
```
AlfredoG87, line 217: in case the feature is disabled, should we return an error response similar to `NEW_PENDING_TRANSACTIONS` such as `Unsupported method`?

this would need a test case.

ebadiere: Done. Good catch.

packages/server/tests/localAcceptance.env
ebadiere: Done.

packages/relay/src/lib/poller.ts
ebadiere: Added test

ebadiere:
> meant to request changes 😅. leaved some nits on previous review, but would love to see the refactor and method extraction of the nested switch if possible.
> 
> also, seems like tests are missing, but naturally since is still a draft.
> 
> what happens or what should happen when you are already subscribed to NEW_HEADS and then you attempt to subscribe **again** ? should we do a test case for this too?

Added test "Try setting up multiple subscriptions for newHeads."

PR #2160, opened 2024-02-29
feat: adding polling for new heads.

ebadiere:
Added polling for latest block, `newHeads`

**Related issue(s)**: #2117 

Fixes #


package.json
```diff
@@ -72,7 +73,7 @@
     "keyv-file": "^0.3.0",
     "koa-cors": "^0.0.16",
     "koa-websocket": "^7.0.0",
-    "lerna": "^8.1.1",
+    "lerna": "^8.0.2",
```
Nana-EC, line 76: Q: why are we bumping down?

packages/server/tests/localAcceptance.env
```diff
@@ -20,3 +20,4 @@ DEBUG_API_ENABLED=true
 SEND_RAW_TRANSACTION_SIZE_LIMIT=131072
 BATCH_REQUESTS_ENABLED=true
 TEST_GAS_PRICE_DEVIATION=0.2
+WS_NEW_HEADS_ENABLED=true
```
Nana-EC, line 23: default to false for this sprint

packages/ws-server/src/webSocketServer.ts
```diff
@@ -175,71 +176,81 @@ app.ws.use(async (ctx) => {
...
-              `${connectionIdPrefix} ${requestIdPrefix} Encountered error on ${
-                ctx.websocket.id
-              }, method: ${method}, params: ${JSON.stringify(params)}`,
-            );
-            response = jsonResp(request.id, error, undefined);
-            ctx.websocket.send(JSON.stringify(response));
-            return;
-          }
+    // Early return if subscription limit is exceeded
+    if (method === constants.METHODS.ETH_SUBSCRIBE && !limiter.validateSubscriptionLimit(ctx)) {
```
Nana-EC, line 180: need a test for this

Nana-EC (commented):
A few items.
Also please update the PR with a description

packages/server/tests/localAcceptance.env
ebadiere: Done.

packages/ws-server/src/webSocketServer.ts
ebadiere: There was an existing test to cover the subscription limit. I just added one for subscription limit and `newHeads`.

AlfredoG87 (approved):
LGTM 👍

PR #2169, opened 2024-03-04
2149 add websocket tests for newheads

ebadiere:
Added websocket test for block content, fixed the JSON response and the design document.

**Related issue(s)**: #2117 

Fixes #


packages/server/tests/acceptance/ws/subscribeNewHeads.spec.ts
```diff
@@ -150,4 +120,74 @@ describe('@web-socket Acceptance Tests', async function () {
...
+            expect(response.params.result).to.have.property('receiptsRoot');
+            expect(response.params.result).to.have.property('sha3Uncles');
+            expect(response.params.result).to.have.property('stateRoot');
+            expect(response.params.result).to.have.property('timestamp');
+            expect(response.params.result).to.have.property('transactionsRoot');
+            expect(response.params.result).to.have.property('hash');
+            expect(response.params).to.have.property('subscription');
+          }
+
+          if (responseCounter > 1) {
```
Nana-EC, line 177: To be clear this will mean the test will only ever confirm 2 responses.
The 1st to get the subscription ID and the 2nd to get the 1st block. You'll never confirm more than 1 blocks details.

You should definitely be testing for 0,1 and n numbers of blocks cases.

ebadiere: Yes, it's a subscription so I need to limit the number of responses I test and focus on the content of those responses.

ebadiere:
Addressed in PR #2225

PR #2180, opened 2024-03-08
fix: trimmed preceding zeros for gas value (#2174)

### LINKED ISSUE #2174: Leading zeros for hex causes incompatibility with common utility packages
### Problem

The JSON RPC responses sometimes return gas values in Hex with leading zeros making it hard to use common utility packages such as https://pkg.go.dev/github.com/ethereum/go-ethereum/common/hexutil 
For example response from `eth_estimateGas` method

```
{
    "result": "0x0000000000006050",
    "jsonrpc": "2.0",
    "id": 1
}
```

### Solution

- Trim the leading zeros

### Alternatives

_No response_

quiet-node:
**Description**:
- added a quick helper method to convert the long zero gas value to a decimal value and revert it back to a hexadecimal value. This ensures that the gas value remains intact during the trimming process.
- fixed corresponded unit test

**Related issue(s)**:

Fixes #2174

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -181,6 +181,10 @@ const prepend0x = (input: string): string => {
   return input.startsWith(EMPTY_HEX) ? input : EMPTY_HEX + input;
 };
 
+const trimPrecedingZeros = (input: string) => {
+  return parseInt(input, 16).toString(16);
+};
+
 const numberTo0x = (input: number | BigNumber | bigint): string => {
   return EMPTY_HEX + input.toString(16);
 };
@@ -251,6 +255,7 @@ export {
   toNullableBigNumber,
   toNullIfEmptyHex,
   generateRandomHex,
+  trimPrecedingZeros,
   weibarHexToTinyBarInt,
   stringToHex,
   toHexString,
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -575,7 +575,7 @@ export class EthImpl implements Eth {
         requestIdPrefix,
       );
 
-      if (contractCallResponse && contractCallResponse.result) {
+      if (contractCallResponse?.result) {
         gas = prepend0x(trimPrecedingZeros(contractCallResponse.result));
       }
     } catch (e: any) {
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -39,6 +39,7 @@ import {
   toHash32,
   toNullableBigNumber,
   weibarHexToTinyBarInt,
+  trimPrecedingZeros,
 } from '../formatters';
 import crypto from 'crypto';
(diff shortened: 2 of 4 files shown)


packages/relay/src/formatters.ts
```diff
@@ -181,6 +181,10 @@ const prepend0x = (input: string): string => {
   return input.startsWith(EMPTY_HEX) ? input : EMPTY_HEX + input;
 };
 
+const trimPrecedingZeros = (input: string) => {
```
Nana-EC, line 184: Add tests for this method.
You should consider multiple inputs valid and invalid

Nana-EC (changes requested):
Missing UTs

packages/relay/src/lib/eth.ts
```diff
@@ -573,8 +574,9 @@ export class EthImpl implements Eth {
         },
         requestIdPrefix,
       );
-      if (contractCallResponse?.result) {
-        gas = prepend0x(contractCallResponse.result);
+
+      if (contractCallResponse && contractCallResponse.result) {
```
acuarica, line 578: Out of curiosity, how does the old condition differ from this one?

acuarica (approved):
lg

packages/relay/src/formatters.ts
quiet-node: Pushed in the UTs!

packages/relay/src/lib/eth.ts
quiet-node: Ah yeah so the "?." operator allows for accessing properties of an object that maybe null or undefined without causing error. On the other hand, the `&&` condition firstly checks if contractCallResponse exists or not, and then access the result property. If contractCallResponse is null or undefined it will not attempt to access to the result property. Huh now that it's broken down it seems like they both eventually serve the same purpose in this context.

packages/relay/tests/lib/formatters.spec.ts
```diff
@@ -262,6 +263,19 @@ describe('Formatters', () => {
...
   });
 
+  describe('trimPrecedingZeros', () => {
+    it('should trim all the unnecessary preceding 0s in a hex value', () => {
+      expect('0x' + trimPrecedingZeros('0x00000000603')).to.eq('0x603');
+      expect('0x' + trimPrecedingZeros('0x00000300042')).to.eq('0x300042');
+      expect('0x' + trimPrecedingZeros('0x00012000000')).to.eq('0x12000000');
+    });
+
+    it('should return NaN if inputs are invalid number', () => {
```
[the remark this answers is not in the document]
quiet-node: I added a case for `''`. This function accepts only a param of type string, so `null` and `undefined` would not be applicable.

AlfredoG87 (dismissed):
LGTM 👍

Nana-EC (approved):

AlfredoG87 (approved):

Issue #2188, opened 2024-03-13
Add logging to help identify memory consumption


ebadiere:
### Problem

Pods were running into memory issues.  

### Solution

Add logging to facilitate the tracing of memory issues around blocks with many transactions.

PR #2189, opened 2024-03-13
feat: Added logging around memory usage.

ebadiere:
Added logging to trace memory usage.
**Related issue(s)**: #2188

Fixes #


AlfredoG87 (commented):
should we put this measurements inside a feature flag so we don't always do this, just we turn it on whenever we need it for troubleshooting purposes but without impacting performance?

PR #2222, opened 2024-03-20
feat: Add Pipeline Set

### LINKED ISSUE #2210: Add support for setting multiple K/V in cache
### Problem

Currently we don't support setting multiple K/V in the cache, which may result in slow performance when we try to set more items at once.

### Solution

Implement multiSet and pipelineSet from redis, which allows for setting multiple K/V at once. We should make their usage configurable to test it out and collect metrics.

### Alternatives

_No response_

georgi-l95:
**Description**:
This PR aims to add another way to set multiple K/V pairs in the shared cache (in our case Redis). You can switch between mSet and pipeline using the .env variable MULTI_SET. By default mSet is set to be used, if MULTI_SET is false - pipeline will be used. Main advantage of pipeline is abillity to use TTL.

**Related issue(s)**:

Fixes #2210 

**Notes for reviewer**:

**Checklist**


AlfredoG87 (approved):
LGTM 👍

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -217,35 +225,45 @@ export class CacheService {
...
         const redisError = new RedisCacheError(error);
         this.logger.error(
           `${requestIdPrefix} Error occurred while setting the cache to Redis. Fallback to internal cache. Error is: ${redisError.fullError}`,
         );
-        // Fallback to internal cache
-        this.internalCache.multiSet(entries, callingMethod, requestIdPrefix);
-        this.cacheMethodsCounter.labels(callingMethod, CacheService.cacheTypes.LRU, CacheService.methods.MSET).inc(1);
+        // Fallback to internal cache, but use pipeline, because of it's TTL support
+        this.internalCache.pipelineSet(entries, callingMethod, ttl, requestIdPrefix);
+        this.cacheMethodsCounter.labels(callingMethod, CacheService.cacheTypes.LRU, CacheService.methods.SET).inc(1);
```
Nana-EC, line 261: It's still a multi set thought right. Why the change to the metric label here and below?
If you change it you'll give the impression the SET was called instead

packages/relay/tests/lib/services/cacheService/cacheService.spec.ts
```diff
@@ -158,7 +159,40 @@ describe('CacheService Test Suite', async function () {
...
+        .returns(entries['key3']);
+
+      for (const [key, value] of Object.entries(entries)) {
+        const valueFromCache = cacheService.getSharedWithFallback(key, callingMethod, undefined);
+        expect(valueFromCache).eq(value);
+      }
+    });
+
+    it('should be able to set using pipelineSet and get them separately using internal cache', async function () {
+      process.env.MULTI_SET = 'false';
```
Nana-EC, line 179: The only difference between the two tests is the value of `MULTI_SET`
How can we confirm whether pipeline was used or not?

packages/relay/src/lib/clients/cache/redisCache.ts
```diff
@@ -160,6 +160,38 @@ export class RedisCache implements ICacheClient {
...
+   */
+  async pipelineSet(
+    keyValuePairs: Record<string, any>,
+    callingMethod: string,
+    ttl?: number | undefined,
+    requestIdPrefix?: string,
+  ): Promise<void> {
+    const resolvedTtl = (ttl ?? this.options.ttl) / 1000; // convert to seconds
+
+    const pipeline = this.client.multi();
```
Nana-EC, line 180: Q: are there any memory impacts of this get?
Does it initiate a new one each time or should we consider getting one and setting as a class variable that is used in method?

Also do we need to clear the pipeline after or does the API do it for you?

packages/relay/src/lib/clients/cache/redisCache.ts
```diff
@@ -160,6 +160,38 @@ export class RedisCache implements ICacheClient {
...
+
+    for (const [key, value] of Object.entries(keyValuePairs)) {
+      const serializedValue = JSON.stringify(value);
+      pipeline.setEx(key, resolvedTtl, serializedValue);
+    }
+
+    // Execute pipeline operation
+    await pipeline.execAsPipeline();
```
Nana-EC, line 188: Q: Are there any pipeline configurations that should be exposed an configureable?

Nana-EC (changes requested):
Looking good.
SOme questions

packages/relay/src/lib/services/cacheService/cacheService.ts
georgi-l95: Actually, if it fallbacks to internal cache, it will use `set`, because internal cache does not support `multi set`.

packages/relay/tests/lib/services/cacheService/cacheService.spec.ts
georgi-l95: Only thing I can think of, if we expose getter for this flag -> `MULTI_SET` in the cache service. And then call and check this flag, depending on the case. If it's false, it will always use pipeline.

packages/relay/src/lib/clients/cache/redisCache.ts
georgi-l95: I believe closing the pipeline is handled by redis implementation itself, we don't need to do anything. New one is opened upon new request. 
Also I don't think that we can use it as class variable, because when execution is completed, the pipeline is closed and until you execute the request it will just keep storing your commands (set,get,etc) and not sending them over. 
Here is comment from the go-redis implementation, but should be the same for node-redis. https://github.com/redis/go-redis/discussions/2583#discussioncomment-5823938

packages/relay/src/lib/clients/cache/redisCache.ts
georgi-l95: Didn't find any https://github.com/redis/node-redis/blob/dbf8f59a47573e6a1c75b78e566af8c493015d5d/packages/client/lib/cluster/multi-command.ts#L130

Nana-EC (approved):
Unblocking.
Everything looks good but would like a follow up way to confirm pipeline logic vs multiset.
Right now they are indistinguishable from tests

packages/relay/tests/lib/services/cacheService/cacheService.spec.ts
Nana-EC: You could confirm via log mock.
Example from Logans recent PR https://github.com/hashgraph/hedera-json-rpc-relay/pull/2243/files#diff-62cee0fdf114bfe63346e4562cdf07ee2cec616ad8ba07784347d8b49fa09d2fR226

georgi-l95: Thanks, opened issue to track it. https://github.com/hashgraph/hedera-json-rpc-relay/issues/2267

PR #2236, opened 2024-03-21
feat: remove block range limit in `eth_getLogs` when filtering with a single address

### LINKED ISSUE #2175: Issues with `eth_getlogs` API
### Description

I have a sample transaction ([here](https://hashscan.io/testnet/contract/0.0.3665157?p=1&k=1709721331.243569998)) that has a log, but cannot be queried through the `eth_getLogs` api

And using `fromBlock` & `toBlock` seems to have no effect at all & cannot also query with the `blockHash` as per [EIP-234](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-234.md)

### Steps to reproduce

```
{
	"jsonrpc":"2.0",
	"method":"eth_getLogs",
	"params":[{
		"topics":[
			"0x87f16aa184eca14ea45e132328a5effbb79b9f921657bd03d83608f26d76f3cf"
		],
        "address":"0x0ed544cdac078ba3bc354b2458c04e358e758c9a"
      
	}],
	"id":1
}
```

### Additional context

_No response_

### Hedera network

testnet

### Version

v2.0

### Operating system

None

acuarica:
**Description**:

This PR removes the block range limit, _e.g._, 1000, when the filter in the `eth_getLogs` contains a single address (the optional `address` param might be either a string or an addresses array).

This change allows consumers to query all the log entries of the given address. For now, we permit only a single address, but in the future we might relax this constraint as well.

The block range validation is also used in `eth_newFilter`, but it should not be affected by this change.

**Related issue(s)**:

Fixes #2175 

**Notes for reviewer**:

It also fixes function type signatures, _i.e., use `string[]` instead of `[string]`.

**Checklist**

### DIFF
--- a/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
+++ b/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
@@ -268,7 +268,7 @@ export class CommonService implements ICommonService {
     }
   }
 
-  public async getLogsByAddress(address: string | [string], params: any, requestIdPrefix) {
+  public async getLogsByAddress(address: string | string[], params: any, requestIdPrefix) {
     const addresses = Array.isArray(address) ? address : [address];
     const logPromises = addresses.map((addr) =>
       this.mirrorNodeClient.getContractResultsLogsByAddress(addr, params, undefined, requestIdPrefix),
@@ -283,7 +283,7 @@ export class CommonService implements ICommonService {
     return logs;
   }
 
-  public async getLogsWithParams(address: string | [string] | null, params, requestIdPrefix?: string): Promise<Log[]> {
+  public async getLogsWithParams(address: string | string[] | null, params, requestIdPrefix?: string): Promise<Log[]> {
     const EMPTY_RESPONSE = [];
 
     let logResults;
@@ -321,7 +321,7 @@ export class CommonService implements ICommonService {
     blockHash: string | null,
     fromBlock: string | 'latest',
     toBlock: string | 'latest',
-    address: string | [string] | null,
+    address: string | string[] | null,
     topics: any[] | null,
     requestIdPrefix?: string,
   ): Promise<Log[]> {
--- a/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
+++ b/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
@@ -74,6 +74,10 @@ export class CommonService implements ICommonService {
     'ETH_BLOCK_NUMBER_CACHE_TTL_MS_DEFAULT',
   );
 
+  private getLogsBlockRangeLimit() {
+    return parseNumericEnvVar('ETH_GET_LOGS_BLOCK_RANGE_LIMIT', 'DEFAULT_ETH_GET_LOGS_BLOCK_RANGE_LIMIT');
+  }
+
   constructor(mirrorNodeClient: MirrorNodeClient, logger: Logger, cacheService: CacheService) {
     this.mirrorNodeClient = mirrorNodeClient;
     this.logger = logger;
@@ -139,10 +143,9 @@ export class CommonService implements ICommonService {
         return false;
       }
 
-      const blockRangeLimit =
-        process.env.TEST === 'true'
-          ? constants.DEFAULT_ETH_GET_LOGS_BLOCK_RANGE_LIMIT
-          : Number(process.env.ETH_GET_LOGS_BLOCK_RANGE_LIMIT);
+      const blockRangeLimit = this.getLogsBlockRangeLimit();
+      // Increasing it to more then one address may degrade mirror node performance
+      // when addresses contains many log events.
       const isSingleAddress = Array.isArray(address)
         ? address.length === 1
         : typeof address === 'string' && address !== '';
--- a/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
+++ b/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
@@ -143,8 +143,10 @@ export class CommonService implements ICommonService {
         process.env.TEST === 'true'
           ? constants.DEFAULT_ETH_GET_LOGS_BLOCK_RANGE_LIMIT
           : Number(process.env.ETH_GET_LOGS_BLOCK_RANGE_LIMIT);
-      const isSingleAddress = Array.isArray(address) ? address.length === 1 : address !== '';
-      if (!isSingleAddress || toBlockNum - fromBlockNum > blockRangeLimit) {
+      const isSingleAddress = Array.isArray(address)
+        ? address.length === 1
+        : typeof address === 'string' && address !== '';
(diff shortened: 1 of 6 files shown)


packages/relay/tests/lib/eth/eth_getLogs.spec.ts
```diff
@@ -281,6 +281,46 @@ describe('@ethGetLogs using MirrorNode', async function () {
     expectLogData3(result[2]);
   });
 
+  [CONTRACT_ADDRESS_1, [CONTRACT_ADDRESS_1]].forEach((address) => {
+    console.log(address);
```
Nana-EC, line 285: nit: remove the console log

Nana-EC (approved):
LG
1 nit

packages/relay/tests/lib/eth/eth_getLogs.spec.ts
acuarica: Good catch!

Nana-EC (approved):

acuarica:
@ebadiere @AlfredoG87 I would appreciate your input here if you have some time, thanks!

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
```diff
@@ -141,7 +137,16 @@ export class CommonService implements ICommonService {
 
       if (fromBlockNum > toBlockNum) {
         return false;
-      } else if (toBlockNum - fromBlockNum > blockRangeLimit) {
+      }
+
+      const blockRangeLimit =
+        process.env.TEST === 'true'
```
AlfredoG87, line 143: `process.env.TEST` what is this property doing?

packages/relay/tests/lib/eth/eth_getLogs.spec.ts
```diff
@@ -281,6 +281,45 @@ describe('@ethGetLogs using MirrorNode', async function () {
     expectLogData3(result[2]);
   });
 
+  [CONTRACT_ADDRESS_1, [CONTRACT_ADDRESS_1]].forEach((address) => {
+    it(`single address filter \`${JSON.stringify(address)}\` and large block range`, async function () {
```
AlfredoG87, line 285: not a fan of using a dynamic name for the test case, I understand the intent but I think is the same effect if the loop is inside the test. That would help readability and maintenability. And in the case the test fail on the CI, to find it faster on the codebase

I could suggest the name: `should filter logs by single address or array of addresses with a large block range` instead?

packages/relay/tests/lib/eth/eth_getLogs.spec.ts
```diff
@@ -439,27 +478,29 @@ describe('@ethGetLogs using MirrorNode', async function () {
...
-    };
-
-    restMock.onGet(BLOCKS_LIMIT_ORDER_URL).reply(200, { blocks: [blockBeyondMaximumRange] });
-    restMock.onGet('blocks/1').reply(200, fromBlock);
-    restMock.onGet('blocks/1003').reply(200, toBlock);
-
-    await ethGetLogsFailing(ethImpl, [null, '0x1', '0x3eb', null, null], (error) => {
-      expect(error.message).to.equal('Exceeded maximum block range: 1000');
+  [null, [], [CONTRACT_ADDRESS_1, CONTRACT_ADDRESS_2]].forEach((address) => {
+    it(`when block range is too large with address \`${JSON.stringify(address)}\``, async function () {
```
AlfredoG87, line 482: I see that we are trying to create many tests out of a collection, but the same could be achieved as part of the same test. and would be cleaner in my opinion.

Also the name of the test should be more descriptive and say what is expecting to happen.

like: `should fail when block range is too large for various addresses`

```
it('should fail when block range is too large for various addresses', async function () {
  const addresses = [null, [], [CONTRACT_ADDRESS_1, CONTRACT_ADDRESS_2]];
  for (const address of addresses) {
    const fromBlock = {
      ...DEFAULT_BLOCK,
      number: 1,
    };
    const toBlock = {
      ...DEFAULT_BLOCK,
      number: 1003,
    };

    const blockBeyondMaximumRange = {
      ...DEFAULT_BLOCK,
      number: 1007,
    };

    // Mocking the HTTP responses for the blocks limit order URL and for specific block numbers
    restMock.onGet(BLOCKS_LIMIT_ORDER_URL).reply(200, { blocks: [blockBeyondMaximumRange] });
    restMock.onGet('blocks/1').reply(200, fromBlock);
    restMock.onGet('blocks/1003').reply(200, toBlock);

    // Testing the failure scenario for the ethGetLogs function with various addresses
    await ethGetLogsFailing(ethImpl, [null, '0x1', '0x3eb', address, null], (error) => {
      expect(error.message).to.equal('Exceeded maximum block range: 1000', `Failed for address: ${JSON.stringify(address)}`);
    });
  }
});

```

AlfredoG87 (changes requested):
Overall looks good, leaved some comments around testing and a question on `process.env.TEST` that I don't understand, if is a new feature flag it should be documented and use a constant at the top of the class as other feature flags do, and should have a more descriptive name than `TEST`

2 other questions:

1. should we really allow for `unlimited` amount of blocks? could it become too much too handle, maybe having a different limit altogether? like maybe, 100,000 blocks or something. let me know your rationale here.
Have we tested limits on the mirror node for timestamp of query? (`contracts/${MirrorNodeClient.ADDRESS_PLACEHOLDER}/results/logs)

2. On the tests, what happens if we increase the default limit of `1000` to lets say `2000`, it would fail?

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
acuarica: This was already used in `eth_getLogs`. As discussed, this was removed.

packages/relay/tests/lib/eth/eth_getLogs.spec.ts
acuarica: As discussed with @AlfredoG87, we left the dynamic test cases, but improve the test title to match `it should` format.

packages/relay/tests/lib/eth/eth_getLogs.spec.ts
acuarica: As discussed with Fready, we left the dynamic test cases, but improve the test title to match `it should` format.

acuarica:
> 1. should we really allow for `unlimited` amount of blocks? could it become too much too handle, maybe having a different limit altogether? like maybe, 100,000 blocks or something. let me know your rationale here.
>    Have we tested limits on the mirror node for timestamp of query? (`contracts/${MirrorNodeClient.ADDRESS_PLACEHOLDER}/results/logs)

The mirror node supports the query without timestamp. The `openrpc.json` was updated to reflect pagination limits in f341324e5a9ac865dee9f3ba3685843b62443a2b.

AlfredoG87 (approved):

PR #2243, opened 2024-03-22
feat: added deleteFile() method to delete files from the network

### LINKED ISSUE #2190: Issue a FileDelete transaction to remove File used in large contract creation flows
### Problem

Currently when a contract bytecode is greater than 5120 bytes we utilize HFS to temporarily hold the contract byte code until contract creation.
However, we don't clean up the file after use

### Solution

Delete the file entity after the operation. Regardless of success or failure we should remove it.

### Alternatives

_No response_

quiet-node:
**Description**:
This PR adds a private function,`deleteFile()` within the `sdkClient` class. This function attempts to remove the `fileId` file that is generated during `CONTRACT_CREATE` transactions if the contract's bytecode is greater than 5KBs. `deleteFile()` is included in the `finally` block of the `sdkClient.executeTransaction()` function. This ensures that the created file is deleted regardless of whether the transaction succeeds or fails.

This PR also adds new UT coverage for `deleteFile()`

Fixes #2190

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -726,9 +726,7 @@ export class SDKClient {
       // Create fileDeleteTx
       const fileDeleteTx = await new FileDeleteTransaction()
         .setFileId(fileId)
-        .setMaxTransactionFee(
-          Hbar.fromTinybars(Math.floor((await this.getTinyBarGasFee('FileDelete')) * constants.BLOCK_GAS_LIMIT)),
-        )
+        .setMaxTransactionFee(new Hbar(2))
         .freezeWith(client);
 
       // execute fileDeleteTx
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -48,6 +48,8 @@ import {
   PrecheckStatusError,
   TransactionRecordQuery,
   Hbar,
+  FileId,
+  FileDeleteTransaction,
 } from '@hashgraph/sdk';
 import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
 import { Logger } from 'pino';
@@ -192,7 +194,6 @@ export class SDKClient {
 
   async getFeeSchedule(callerName: string, requestId?: string): Promise<FeeSchedules> {
     const feeSchedulesFileBytes = await this.getFileIdBytes(constants.FEE_SCHEDULE_FILE_ID, callerName, requestId);
-
     return FeeSchedules.fromBytes(feeSchedulesFileBytes);
   }
 
@@ -515,6 +516,20 @@ export class SDKClient {
         throw predefined.HBAR_RATE_LIMIT_EXCEEDED;
       }
       throw sdkClientError;
+    } finally {
+      /**
+       *  For transactions of type CONTRACT_CREATE, if the contract's bytecode (calldata) exceeds 5120 bytes, HFS is employed to temporarily store the bytecode on the network.
+       *  After transaction execution, whether successful or not, any entity associated with the 'fileId' should be removed from the Hedera network.
+       */
+      if (transaction['_callDataFileId']) {
+        await this.deleteFile(
+          this.clientMain,
+          transaction['_callDataFileId'] as FileId,
+          requestId,
+          callerName,
+          interactingEntity,
+        );
+      }
     }
   };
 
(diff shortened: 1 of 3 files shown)


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -515,6 +516,20 @@ export class SDKClient {
         throw predefined.HBAR_RATE_LIMIT_EXCEEDED;
       }
       throw sdkClientError;
+    } finally {
+      /**
+       *  For transactions of type CONTRACT_CREATE, if the contract's bytecode (calldata) exceeds 5120 bytes, HFS is employed to temporarily store the bytecode on the network.
+       *  After transaction execution, whether successful or not, any entity associated with the 'fileId' should be removed from the Hedera network.
+       */
+      if (transaction['_callDataFileId']) {
```
georgi-l95, line 524: If for some reason the sdk changes this property name at some point, we would have to change again. Ideal for me would be to keep track of the transacion type in the relay and not rely on specific implementation somewhere else. But this might never happen, so this is also good 👍

georgi-l95 (approved):
LG, solid update. Thank you, slight concern, which we can totally ignore now.

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC, line 524: Q: why not check for the transaction type and then just use [callDataFileId()](https://github.com/hashgraph/hedera-sdk-js/blob/develop/src/EthereumTransaction.js#L188) and check for a valid entityId
This would also ensure we catch if the sdk ever deprecates and breaks us

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -515,6 +516,20 @@ export class SDKClient {
...
+        await this.deleteFile(
+          this.clientMain,
+          transaction['_callDataFileId'] as FileId,
```
Nana-EC, line 527: Use `transaction.callDataFileId()`

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -692,4 +707,61 @@ export class SDKClient {
...
+    interactingEntity?: string,
+  ) => {
+    // format request ID msg
+    const requestIdPrefix = formatRequestIdMessage(requestId);
+
+    try {
+      // Create fileDeleteTx
+      const fileDeleteTx = await new FileDeleteTransaction()
+        .setFileId(fileId)
+        .setMaxTransactionFee(
```
Nana-EC, line 733: Q: what's this calculation here?
We should have to calculate it since it's a transaction initiated by us.
Maybe just set a reasonable limit based on testing.
Also see https://testnet.mirrornode.hedera.com/api/v1/docs/#/network/getNetworkFees, FileDelete fees is not supported by mirror node

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -692,4 +707,61 @@ export class SDKClient {
...
+        interactingEntity,
+      );
+
+      // ensure the file is deleted
+      const receipt = deleteFileRecord.receipt;
+      const fileInfo = await new FileInfoQuery().setFileId(fileId).execute(client);
+      if (receipt.status === Status.Success && fileInfo.isDeleted) {
+        this.logger.trace(`${requestIdPrefix} Deleted file with fileId: ${fileId}`);
+      } else {
+        throw new SDKClientError({}, `${requestIdPrefix} Fail to delete file with fileId: ${fileId} `);
```
Nana-EC, line 761: Don't throw.
Let's leave this as best effort

Nana-EC: Same for the catch. Make it no op so that we don't give the wrong response due to cleanup attempts.
Warning logs should be fine

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -157,4 +170,53 @@ describe('SdkClient', async function () {
...
+      // mock captureMetrics
+      sinon.stub(sdkClient, 'captureMetrics').callsFake(() => {});
+    });
+
+    it('should execute deleteFile', async () => {
+      // prepare fileId
+      const signedTx = await signTransaction(transaction);
+      const transactionBuffer = Buffer.from(signedTx.substring(2), 'hex');
+      const ethereumTransactionData: EthereumTransactionData = EthereumTransactionData.fromBytes(transactionBuffer);
+      const fileId = await sdkClient.createFile(ethereumTransactionData.callData, client, requestId, callerName, '');
```
Nana-EC, line 198: Is fileId a valid entity here?
I don't see any stubbing to show that the response is properly mocked.

Nana-EC (changes requested):
Good start.
Some suggestions

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: I was looking around in the EthereumTransaction and Transaction extension classes but they don't seem like they carry transaction type around. 

However, @georgi-l95 suggested passing the `fileId` (generated by FileCreateTransaction()) down the path and this makes very good sense. We know for sure that if `fileId !== null` => there's a file to delete so we don't have to worry about the SDK every changing the property and breaks us later.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: New changes will use `fileId` from parameters

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: I have revert back to a max amount of 2 HBARs for this method.

packages/relay/tests/lib/sdkClient.spec.ts
quiet-node: At first I thought the same as I also didn't see any stub on this method. Turns out this spec class initialized a real client and hooked it to the current network in the environment. That's why the fileId is succesfully created. I also added some assertions to make sure fileId is a valid entity.

Nana-EC: That's odd, all these tests should be based on mocks.
Whereas the acceptance test should be hitting a real consensus node.

Maybe config got updated as some tests are still mocks in this class

Nana-EC (dismissed):
Unblocking, please confirm the spec tests as it should use mocks.
Kindly add an acceptance tests that confirms E2E against an actual consensus node

quiet-node:
> Unblocking, please confirm the spec tests as it should use mocks. Kindly add an acceptance tests that confirms E2E against an actual consensus node

Thanks @Nana-EC yes you're right just pushed in an acceptance test to confirm the deleteFile() function. 

About the spec tests I'll create a ticket about modifying the SDKClient spec class to use mocks for testing. I also noticed that there are many of features of the SDKClient class don't get test coverage. Maybe the ticket will add more UTs to cover other functions as well.

Nana-EC (approved):
LGTM

PR #2254, opened 2024-03-28
feat: reset account nonce script

natanasow:
**Description**:

Simple hardhat project that demonstrates how an end-user should be able to delete their account via SDK calls. The deletion process requires a receiver account that will be used to transfer the remaining funds to, and an operator account that should execute the transaction.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


Nana-EC (changes requested):
Blocking to prevent merge but PR was needed to allow review

reset-account-nonce-script/.env.example
```diff
@@ -0,0 +1,6 @@
+DELETABLE_ACCOUNT_PK=    # hex encoded
+
+OPERATOR_ID=             # format 0.0.9999
```
Nana-EC, line 3: Add a comment for EVM devs to show how from their Operator ID from their evm address
Actually better yet i don't think we should use this. 
Let the evm user provide their address and let the code pick up the Operator ID

reset-account-nonce-script/.env.example
```diff
@@ -0,0 +1,6 @@
...
+OPERATOR_PK=             # hex encoded
+
+RECEIVER_ID=             # format 0.0.9999 - the id of the account to transfer the remaining funds to
```
Nana-EC, line 6: Same thing, let them provide an evm address

reset-account-nonce-script/README.md
```diff
@@ -0,0 +1,23 @@
+# Account deletion hardhat script
+
+This project demonstrates how to delete an already existing ECDSA account.
```
Nana-EC, line 3: Note that the perspective is from an EVM dev and not a Hedera dev.
Also provide context to say one application of this might be the reseting of an accounts nonce to allow flows that require initial nonces such as deterministic contract address cases

reset-account-nonce-script/hardhat.config.js
```diff
@@ -0,0 +1,51 @@
+/*-
+ *
+ * Hedera JSON RPC Relay
+ *
+ * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
```
Nana-EC, line 5:
```suggestion
 * Copyright (C) 2024 Hedera Hashgraph, LLC
```

reset-account-nonce-script/scripts/deleteAccount.js
```diff
@@ -0,0 +1,57 @@
+/*-
+ *
+ * Hedera JSON RPC Relay
+ *
+ * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
```
Nana-EC, line 5:
```suggestion
 * Copyright (C) 2024 Hedera Hashgraph, LLC
```

reset-account-nonce-script/scripts/deleteAccount.js
```diff
@@ -0,0 +1,57 @@
...
+
+  console.log(`Selected network: ${NETWORK}`);
+
+  const deletableWallet = new hre.ethers.Wallet(process.env.DELETABLE_ACCOUNT_PK);
+
+  const preDeletionAccount = await (await fetch(`${MIRROR_NODE_URL}/api/v1/accounts/${deletableWallet.address}`)).json();
+  console.log(`address: ${preDeletionAccount.evm_address} deleted: ${preDeletionAccount.deleted} balance: ${preDeletionAccount.balance.balance}`);
+
+  const client = Client.forNetwork(NETWORK)
+    .setOperator(process.env.OPERATOR_ID, PrivateKey.fromStringECDSA(process.env.OPERATOR_PK));
```
Nana-EC, line 36: Here's where I'd say use a value calculated from the evm address instead of asking them to provide an ID.

Basically call the mirror node and get the value

Nana-EC (commented):
Nice work.
Functionally sound, added some suggestions to make it more easy for evm devs

reset-account-nonce-script/.env.example
natanasow: I got it and fixed the script to be used from an EVM dev perspective, not a Hedera one. Actually, the operator ID or address is not needed because it can be derived by the operator's private key and the query the mirror node its account ID.

reset-account-nonce-script/.env.example
natanasow: Done.

reset-account-nonce-script/README.md
natanasow: Fixed.

reset-account-nonce-script/.env.example
```diff
@@ -0,0 +1,3 @@
+DELETABLE_ACCOUNT_PK= # hex encoded
+OPERATOR_PK=          # hex encoded
+RECEIVER_ADDRESS=     # hex encoded - the account to transfer the remaining funds to
```
AlfredoG87, line 3: Blank line missing at EOF

Nana-EC:
> Looks good, just one nit of Blank line missing at EOF.
> 
> Besides that, not sure if the project `reset-account-nonce-script` should be at root `/`, maybe inside `/tools/` or maybe a new folder called `/examples/`
> 
> @ebadiere @Nana-EC, what do you think about the right location for this example project?

We're planning to make this a protected branch for use as need be since it's a more power suer flow.
If we need to move it to `main` we can.
This PR is only for review purposes but won't be merged

reset-account-nonce-script/scripts/deleteAccount.js
AlfredoG87: Thanks for the clarification 👍

reset-account-nonce-script/scripts/deleteAccount.js
```diff
@@ -0,0 +1,60 @@
...
+  const transaction = new AccountDeleteTransaction()
+    .setAccountId(preDeletionAccount.account)
+    .setTransferAccountId(receiver.account);
+
+  const txResponse = await (await transaction.freezeWith(client).sign(PrivateKey.fromStringECDSA(process.env.DELETABLE_ACCOUNT_PK))).execute(client);
+  const receipt = await txResponse.getReceipt(client);
+
+  console.log('deletion status: ' + receipt.status + ' (SUCCESS = 22)');
+  console.log('waiting for mirror node data population...');
+  await new Promise(r => setTimeout(r, 5000));
```
Nana-EC, line 50: nit: bump to 10 secs since you're not poling, just in case

reset-account-nonce-script/scripts/deleteAccount.js
```diff
@@ -0,0 +1,60 @@
...
+
+  const afterDeletionAccount = await (await fetch(`${MIRROR_NODE_URL}/api/v1/accounts/${deletableWallet.address}`)).json();
+  console.log(`mirror node query after deletion: ${afterDeletionAccount?._status?.messages[0]?.message} (expected "Not found")`);
```
Nana-EC, line 53: Log the full mirror node url query so user can re-run on their own as a self-service step before reaching out for assistance.

reset-account-nonce-script/scripts/deleteAccount.js
```diff
@@ -0,0 +1,60 @@
...
+  process.exit(1);
```
Nana-EC, line 54: Shouldn't this be conditional on if the account is deleted or not?
I forget now but doesn't the mirror node return 200 even on deleted account with deleted set to true?

Nana-EC (commented):
Nice fixes for EVM perspective.
Some comments on the final stage of the script to handle latency but also confirm deletion status

reset-account-nonce-script/.env.example
natanasow: Fixed.

reset-account-nonce-script/scripts/deleteAccount.js
natanasow: Done.

reset-account-nonce-script/scripts/deleteAccount.js
natanasow: Done.

reset-account-nonce-script/scripts/deleteAccount.js
natanasow: The mirror node returns 404 for non-existing or deleted accounts. [That account](https://testnet.mirrornode.hedera.com/api/v1/accounts/0xF9747f7FeAd8e30a4EfabCA412c4FA8343Bd506f) was auto-created and deleted afterward.

Nana-EC (commented):
LGTM
Can close the PR and future update will need a PR to the protected branch

PR #2259, opened 2024-03-28
feat: added  parameter check on block param for getcode()

### LINKED ISSUE #2130: getCode unit tests missing test for block number and tag parameters
### Description

Currently, in the eth_getCode.spec.ts file we do not have tests covering the second parameter that can be passed - namely glock number and block tag.

### Steps to reproduce

1. Check the eth_getCode.spec.ts

### Additional context

_No response_

### Hedera network

other

### Version

0.41

### Operating system

None

quiet-node:
**Description**:
The `blockParam` parameter for the `getCode()` method is currently validated exclusively by the [validateParam()](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/server/src/validator/utils.ts#L4) method at the server-level. However, it is necessary to extend this validation to the relay-level as well.

This PR adds a parameter validation step for the `blockParam` parameter in the `getCode()` method. It also includes UTs to ensure coverage of the new validation check.

**Related issue(s)**:

Fixes #2130

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1015,8 +1015,7 @@ export class EthImpl implements Eth {
    */
   async getCode(address: string, blockNumber: string | null, requestIdPrefix?: string) {
     if (!EthImpl.isBlockParamValid(blockNumber)) {
-      throw predefined.INVALID_PARAMETER(
-        1, // param index
+      throw predefined.UNKNOWN_BLOCK(
         `The value passed is not a valid blockHash/blockNumber/blockTag value: ${blockNumber}`,
       );
     }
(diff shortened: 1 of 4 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1014,6 +1014,13 @@ export class EthImpl implements Eth {
    * @param blockNumber
    */
   async getCode(address: string, blockNumber: string | null, requestIdPrefix?: string) {
+    if (!EthImpl.isBlockParamValid(blockNumber)) {
+      throw predefined.INVALID_PARAMETER(
```
Nana-EC, line 1018: Can use predefined.UNKNOWN_BLOCK.
Feel free to update with message

Nana-EC (commented):

packages/relay/src/lib/eth.ts
quiet-node: Right UNKNOWN_BLOCK would make more sense in this case! Thanks for pointing it out!

Nana-EC (approved):

PR #2296, opened 2024-04-05
feat: Implemented getCode.

### LINKED ISSUE #2276: Implement `getCode(address, blockTag)` :in the websocket server
### Problem

When using websockets some JSON RPC methods are not available. `getCode(address, blockTag)` is one of them.

### Solution

Enhance the outer switch, case, statement in the webSocketServer.ts to call the `getCode(address, blockTag)` method in the Relay.

### Alternatives

_No response_

ebadiere:
Implemented `eth_getCode` over websockets.

**Related issue(s)**:

Fixes #2276


packages/ws-server/src/webSocketServer.ts
```diff
@@ -169,6 +169,16 @@ app.ws.use(async (ctx) => {
             connectionIdPrefix,
           );
           break;
+        case WS_CONSTANTS.METHODS.ETH_GET_CODE:
+          relayResponse = await relay.eth().getCode(params[0], params[1]);
```
quiet-node, line 173: Why didn't you use the `response` variable init at line 131? Using the same `response`you wouldn't have to send the result back to the client again yourself but down below (after the switch block) it already send back to the client. 

```suggestion
const contractCode = await relay.eth().getCode(params[0], params[1]);
response = jsonResp(request.id, null, contractCode)
```

quiet-node (changes requested):

packages/ws-server/src/webSocketServer.ts
ebadiere: If I do that it falls into the if statement on line 194, which will not format the response correctly, or at least match the ethereum RPC's response:
```
    if (response) {
      ctx.websocket.send(JSON.stringify(response));
    }
```

quiet-node (approved):

quiet-node (changes requested):

packages/ws-server/src/webSocketServer.ts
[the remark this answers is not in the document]
ebadiere: Fixed.

packages/ws-server/src/controllers/eth_getCode.ts
```diff
@@ -0,0 +1,85 @@
...
+ *
+ */
+
+import { sendToClient } from '../utils/utils';
+import { Relay } from '@hashgraph/json-rpc-relay';
+import { validateParamsLength } from '../utils/validators';
+import { handleSendingTransactionRequests } from './helpers';
+
+/**
+ * Handles the "eth_getTransactionByHash" method request by retrieving transaction details from the Hedera network.
```
quiet-node, line 27: You might want to update the doc comment

quiet-node (approved):
LG one nit

packages/ws-server/src/controllers/eth_getCode.ts
ebadiere: Done.

quiet-node (approved):
LGTM

PR #2305, opened 2024-04-09
fix: fixes unit test heap memory issue

### LINKED ISSUE #2300: Unit Tests are failing with Fatal Error: Memory Leak
Unit test are failing with the following error: `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`
Last commit where test have been succesfully running is from Feb 15: cc943f5a0ad9a1e9151d6d7b13270c164e225b5d

konstantinabl:
**Description**:

This PR addresses a memory issue occurring during unit tests execution.
In order to do this the openrpc.json file had to have a recursion, which led to exhaustion of memory, removed.

Additionally, the following changes has been made:
* Tests that failed due to the memory issue disrupting the test workflow have been fixed. This includes removing irrelevant tests, such as the Gas tests in eth_call.spec.
* Introduces Reds in memory server, which fixes all the tests, that were using RedisCache and were throwing ECONNREFUSED
* The tsconfig file has been updated to include the sourceMap property, enabling debugging directly within TypeScript files.

**Related issue(s)**:

Fixes #2300 

**Checklist**


packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -174,56 +174,4 @@ describe('SdkClient', async function () {
       sinon.assert.calledOnce(convertGasPriceToTinyBarsStub);
     });
   });
-
-  describe('deleteFile', () => {
```
georgi-l95, line 178: Are we deleting them entirely ?

packages/relay/tests/lib/eth/eth_call.spec.ts
```diff
@@ -200,97 +201,6 @@ describe('@ethCall Eth Call spec', async function () {
       process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE = initialEthCallConesneusFF;
     });
 
-    it('gas exceeds limit', async function () {
```
georgi-l95, line 203: Are we deleting them entirely ?

docs/openrpc.json
```diff
@@ -2058,7 +2105,7 @@
                     "calls": {
                         "type": "array",
                         "items": {
-                            "$ref": "#/components/schemas/callframe"
```
georgi-l95, line 2061: Nice catch

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -92,7 +92,7 @@ export class CacheService {
 
     this.internalCache = new LocalLRUCache(logger.child({ name: 'localLRUCache' }), register);
     this.sharedCache = this.internalCache;
-    this.isSharedCacheEnabled = process.env.TEST === 'true' ? false : this.isRedisEnabled();
+    this.isSharedCacheEnabled = process.env.TRUE === 'true' ? false : this.isRedisEnabled();
```
georgi-l95, line 95: Is the name of the env. variable correct ?

georgi-l95 (commented):
LG, thank you. Will try to circle back.

packages/relay/tests/lib/eth/eth_call.spec.ts
```diff
@@ -135,6 +135,7 @@ describe('@ethCall Eth Call spec', async function () {
     });
 
     it('eth_call with incorrect `to` field length', async function () {
+      web3Mock.onPost('contracts/call').reply(200);
```
victor-yanev, line 138: QQ: Why are we mocking this here?

packages/relay/tests/lib/eth/eth_estimateGas.spec.ts
```diff
@@ -165,6 +165,11 @@ describe('@ethEstimateGas Estimate Gas spec', async function () {
   });
 
   it('should eth_estimateGas transfer to existing account', async function () {
+    const callData = {
+      to: RECEIVER_ADDRESS,
+      value: 10, //in tinybars
+    };
+    mockContractCall(callData, true, 501, { errorMessage: '', statusCode: 501 });
```
victor-yanev, line 172: Same here and for the other tests below, why do we need to do this mock contract call with 501? Isn't this a positive scenario?

packages/relay/tests/lib/eth/eth_getBlockByNumber.spec.ts
```diff
@@ -399,30 +412,30 @@ describe('@ethGetBlockByNumber using MirrorNode', async function () {
     it('eth_getBlockByNumber with finalized tag', async function () {
       const result = await ethImpl.getBlockByNumber('finalized', false);
       // check that we only made the expected number of requests with the expected urls
-      expect(restMock.history.get.length).equal(4);
+      expect(restMock.history.get.length).equal(12);
```
victor-yanev, line 415: Where is this number coming from, can't we get it dynamically instead of hardcoding it? Same for other occurrences below.

victor-yanev (changes requested):
Nice changes overall, I like the `RedisInMemoryServer` addition for the tests. Just a few concerns:

packages/relay/tests/lib/eth/eth_getBlockByNumber.spec.ts
georgi-l95: Good point, if we can't get it dynamically, we should set them in a variable, instead of using magic number, for better readabillity.

packages/relay/src/lib/services/cacheService/cacheService.ts
konstantinabl: Noope

packages/relay/tests/lib/eth/eth_call.spec.ts
konstantinabl: I made a mistake. Some of the tests show this error
 
Screenshot 2024-04-19 at 12 39 53
However, I fixed it in the wrong test, in reality the one below this one is missing this mock. Will take a look and fix it

packages/relay/tests/lib/eth/eth_call.spec.ts
konstantinabl: @georgi-l95 
I think it makes sense to delete them:
* gas exceeds limit - we would no longer exceed the limit, since we cap it to the limit ourselves
* block 0, block 1, block earliest, block hash - these are also not relevant, since we have changed eth_call to work with historical block as well as block hashes

packages/relay/tests/lib/sdkClient.spec.ts
konstantinabl: Not sure, we should actually now that I have double checked. Will take a second look

packages/relay/tests/lib/eth/eth_estimateGas.spec.ts
konstantinabl: The problem here is that tests have not been refactored for the new changes in estimate gas. We used to be able to get estimation from mirror-node only for read-only smart contract calls and if the call was a transfer for example, we would return the baseCost. Thus, this test is written in this way. However, now we can also estimate transfers and the test and is no more relevant and needs to be refactored. 
Good catch : )

packages/relay/tests/lib/eth/eth_getBlockByNumber.spec.ts
konstantinabl: I don't think we can get it dynamically, so I added it to a constant. In general, this can also be refactored and the test can be made more specific, but I think it is out of this PR's scope

packages/relay/src/lib/clients/cache/ICacheClient.ts
```diff
@@ -22,12 +22,13 @@ export interface ICacheClient {
   get(key: string, callingMethod: string, requestIdPrefix?: string): any;
```
victor-yanev, line 22:
```suggestion
  get(key: string, callingMethod: string, requestIdPrefix?: string): Promise<any> | any;
```

packages/relay/src/lib/clients/cache/ICacheClient.ts
```diff
@@ -22,12 +22,13 @@ export interface ICacheClient {
...
   set(key: string, value: any, callingMethod: string, ttl?: number, requestIdPrefix?: string): void;
```
victor-yanev, line 23:
```suggestion
  set(key: string, value: any, callingMethod: string, ttl?: number, requestIdPrefix?: string): Promise<void> | void;
```

packages/relay/src/lib/clients/cache/ICacheClient.ts
```diff
@@ -22,12 +22,13 @@ export interface ICacheClient {
...
   delete(key: string, callingMethod: string, requestIdPrefix?: string): void;
```
victor-yanev, line 24: Same here

packages/relay/src/lib/clients/cache/ICacheClient.ts
```diff
@@ -22,12 +22,13 @@ export interface ICacheClient {
...
-  clear(): void;
+  clear(): Promise<void> | void;
   multiSet(keyValuePairs: Record<string, any>, callingMethod: string, requestIdPrefix?: string): void;
```
victor-yanev, line 26: And here

packages/relay/src/lib/clients/cache/ICacheClient.ts
```diff
@@ -22,12 +22,13 @@ export interface ICacheClient {
...
   pipelineSet(
     keyValuePairs: Record<string, any>,
     callingMethod: string,
     ttl?: number | undefined,
     requestIdPrefix?: string,
   ): void;
```
victor-yanev, line 32: And here

packages/relay/tests/lib/eth/eth_getBlockByNumber.spec.ts
```diff
@@ -363,16 +376,17 @@ describe('@ethGetBlockByNumber using MirrorNode', async function () {
 
     it('eth_getBlockByNumber with latest tag', async function () {
       const result = await ethImpl.getBlockByNumber('latest', false);
+      console.log('In latest taaag test', restMock.history.get.length);
```
victor-yanev, line 379: remove this console.log☝️

packages/relay/tests/lib/eth/eth_getBlockByNumber.spec.ts
```diff
@@ -399,30 +413,30 @@ describe('@ethGetBlockByNumber using MirrorNode', async function () {
...
+      expect(restMock.history.get.length).equal(TOTAL_GET_CALLS_EXECUTED);
       expect(restMock.history.get[0].url).equal(BLOCKS_LIMIT_ORDER_URL);
       expect(restMock.history.get[1].url).equal(
         'contracts/results?timestamp=gte:1651560386.060890949&timestamp=lte:1651560389.060890949&limit=100&order=asc',
       );
       expect(restMock.history.get[2].url).equal(
         'contracts/results/logs?timestamp=gte:1651560386.060890949&timestamp=lte:1651560389.060890949&limit=100&order=asc',
       );
-      expect(restMock.history.get[3].url).equal('network/fees');
+      expect(restMock.history.get[11].url).equal('network/fees');
```
victor-yanev, line 424: I guess we can replace `11` with `TOTAL_GET_CALLS_EXECUTED - 1` - same for other occurrences of `11`.

packages/relay/tests/lib/ethGetBlockBy.spec.ts
```diff
@@ -262,8 +266,16 @@ describe('eth_getBlockBy', async function () {
   this.timeout(10000);
   let ethImpl: EthImpl;
   const sandbox = sinon.createSandbox();
-
-  this.beforeAll(() => {
+  let multiSetSpy;
```
victor-yanev, line 269:
```suggestion
  let multiSetSpy: SinonSpy<[keyValuePairs: Record<string, any>, callingMethod: string, requestIdPrefix?: string], Promise<void> | void>;
```

packages/relay/tests/lib/subscriptionController.spec.ts
```diff
@@ -252,5 +260,7 @@ describe('subscriptionController', async function () {
     const subId2 = subscriptionController.subscribe(wsConnection, tag1.event);
 
     expect(subId).to.be.eq(subId2);
+    subscriptionController.unsubscribe(wsConnection, subId);
+    subscriptionController.unsubscribe(wsConnection, subId2);
```
victor-yanev, line 264: Can we move all `subscribe` and `unsubscribe` calls in `beforeEach` and `afterEach`?

packages/relay/tests/redisInMemoryServer.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  async getHost(): Promise<string> {
+    return this.host;
+  }
+
+  getInstanceInfo(): false | RedisInstanceDataT {
+    return this.inMemoryRedisServer.getInstanceInfo();
+  }
+
+  async stop() {
```
victor-yanev, line 65:
```suggestion
  async stop(): Promise<void> {
```

packages/relay/tests/redisInMemoryServer.ts
```diff
@@ -0,0 +1,74 @@
...
+   * @private
+   */
+  private portToRunOn: number;
+
+  constructor(logger: Logger, portToRunOn: number) {
+    this.logger = logger;
+    this.portToRunOn = portToRunOn;
+  }
+
+  async start() {
```
victor-yanev, line 42:
```suggestion
  async start(): Promise<void> {
```

packages/relay/tests/lib/subscriptionController.spec.ts
konstantinabl: Well, they are not all the same, so Im not sure whats the point of moving them all?
What can be moved is the creation of the MockWsConnection instantiation, but in the different tests we subscribe to different tags/events etc. :?

victor-yanev: Sorry, you are right. I thought we could extract them somehow in order to ensure that we always call unsubscribe on all subscribed events/tags at the end of the test. Let's keep it as is.

victor-yanev (dismissed):

packages/relay/src/lib/clients/cache/ICacheClient.ts
victor-yanev: Well, it makes sense for both implementations to be using Promises, if they are going to follow a common interface. We can add async modifier to the `LocalLRUCache` methods and use Promises as return types in the interface. In this way both classes will conform to the interface and it wouldn't have any effect on the actual implementation.

victor-yanev (approved):
LGTM

PR #2331, opened 2024-04-11
feat: Updated to deploy the contract using the websocket provider.

### LINKED ISSUE #2301: Our `maxPriorityFeePerGas` response of `0x0` does not play well with ethers.js deploy transaction via websocket provider.
### Description

When using the websocket provider to deploy transactions using ethers.js `maxPriorityFeePerGas` gets called and causes the deployment process to error out.  Zero padding the `0x0` value does not fix the issue.  Running a deployment against another EVM RPC will return a hex value in the `maxPriorityFeePerGas` response.  Hardcoding that value in our RPC response will allow the ethers.js driven deployment to complete.  

### Steps to reproduce

A test like the following can reproduce the issue and be run against any other EVM RPC to verify:
```
it("should deploy a contract using sendTransaction", async function () {
    const wsUrl = "ws://localhost:8546";
    
    // Create a new provider connected to the Ganache WebSocket URL
    const provider = new ethers.WebSocketProvider(wsUrl);

    // Use the provider to get signers
    const signer = await provider.getSigner();

    // Deploy the contract
    const SimpleContract = await ethers.getContractFactory("SimpleContract", signer);
    const simpleContract = await SimpleContract.deploy();
    await simpleContract.deploymentTransaction();

  });
  ```

### Additional context

_No response_

### Hedera network

testnet, other

### Version

v0.45.0

### Operating system

macOS

ebadiere:
Could not reproduce.  Updated test to use ethers.js websocketProvider to deploy a contract.

**Related issue(s)**:

Fixes #2301

### DIFF
--- a/packages/server/tests/acceptance/ws/estimateGas.spec.ts
+++ b/packages/server/tests/acceptance/ws/estimateGas.spec.ts
@@ -23,32 +23,39 @@ import WebSocket from 'ws';
 import { expect } from 'chai';
 import basicContractJson from '../../contracts/Basic.json';
 import { AliasAccount } from '../../clients/servicesClient';
-import { Contract, ethers, JsonRpcProvider, WebSocketProvider } from 'ethers';
+import { ethers, WebSocketProvider } from 'ethers';
 
 describe('@release @web-socket eth_estimateGas', async function () {
   // @ts-ignore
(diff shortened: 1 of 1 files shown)


ebadiere:
Replaced by #2332

PR #2356, opened 2024-04-16
feat: Added prometheus gauges.

### LINKED ISSUE #2353: Implement resource utilization metrics in the `webSocketServer.ts
### Problem

We need to be aware of any spike in resource consumption by the `webSocketServer.ts`.  To help with this prometheus gauges should be implemented to monitor CPU, memory, and network usage.

### Solution

Thoughts in pseudo code below:
```
const { Gauge } = require('prom-client');
const os = require('os');

// Define gauges for resource utilization metrics
const cpuUsageGauge = new Gauge({
    name: 'websocket_cpu_usage_percentage',
    help: 'CPU usage percentage of the WebSocket server',
});

const memoryUsageGauge = new Gauge({
    name: 'websocket_memory_usage_bytes',
    help: 'Memory usage of the WebSocket server in bytes',
});

const networkUsageGauge = new Gauge({
    name: 'websocket_network_usage_bytes',
    help: 'Network usage of the WebSocket server in bytes',
});

// Function to update resource utilization metrics
function updateResourceUtilizationMetrics() {
    // Update CPU usage metric
    const cpuUsage = process.cpuUsage();
    const totalCpuTime = cpuUsage.user + cpuUsage.system;
    const cpuUsagePercentage = (totalCpuTime / os.cpus().length) * 100;
    cpuUsageGauge.set(cpuUsagePercentage);

    // Update memory usage metric
    const memoryUsage = process.memoryUsage().rss;
    memoryUsageGauge.set(memoryUsage);

    // Update network usage metric (example: using network interface 'eth0')
    const networkUsage = os.networkInterfaces()['eth0'].reduce((acc, iface) => acc + iface.rxBytes + iface.txBytes, 0);
    networkUsageGauge.set(networkUsage);
}

// Update resource utilization metrics periodically (e.g., every 5 seconds)
setInterval(updateResourceUtilizationMetrics, 5000);

app.ws.use(async (ctx) => {
    // Other WebSocket processing code...

...........
});
```

### Alternatives

_No response_

ebadiere:
Added prometheus gauges to cover resources, CPU, memory, and network.

Fixes #2353


quiet-node (approved):
LGTM

AlfredoG87 (changes requested):
Leave a change, but also, are these metrics not better to be collected using other tools. I am no sure how the relay does it, but we don't have them exposed like this, it seems to me that the relay does it with a configuration from the infrastructure. we should ask Brad or Matt

packages/ws-server/src/webSocketServer.ts
```diff
@@ -87,6 +89,42 @@ const methodsCounterByIp = generateMethodsCounterById(register, {
...
+  memoryUsageGauge.set(memoryUsage);
+
+  const networkUsage =
+    os.networkInterfaces().eth0?.reduce((acc, iface) => acc + (iface as any).tx_bytes + (iface as any).rx_bytes, 0) ??
+    0;
+  networkUsageGauge.set(networkUsage);
+};
+
+// Update resource utilization metrics periodically (e.g., every 5 seconds)
+setInterval(updateResourceUtilizationMetrics, 5000);
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/ws-server/src/webSocketServer.ts
```diff
@@ -87,6 +91,51 @@ const methodsCounterByIp = generateMethodsCounterById(register, {
...
+};
+
+const cpuUsageGauge = generateGauge(
+  register,
+  {
+    name: WS_CONSTANTS.cpuUsageGauge.name,
+    help: WS_CONSTANTS.cpuUsageGauge.help,
+    labelNames: WS_CONSTANTS.cpuUsageGauge.labelNames,
+  },
+  updateResourceUtilizationMetrics,
```
AlfredoG87, line 116: all 3 gauges are using the same method. that would be inneficient since it would be called 3 times when it only needs to be done once, the specific for each gauge.

also, the set of the value is something like the provided example:

using the `this` keyword, setting the appropiate `labels` and then `set` the gauge value.

```      
    this.labels({ accountId: clientMain.operatorAccountId?.toString() }).set(accountBalance);
```

packages/ws-server/src/webSocketServer.ts
```diff
@@ -87,6 +91,51 @@ const methodsCounterByIp = generateMethodsCounterById(register, {
...
+);
+
+const memoryUsageGauge = generateGauge(
+  register,
+  {
+    name: WS_CONSTANTS.memoryUsageGauge.name,
+    help: WS_CONSTANTS.memoryUsageGauge.help,
+    labelNames: WS_CONSTANTS.memoryUsageGauge.labelNames,
+  },
+  updateResourceUtilizationMetrics,
```
AlfredoG87, line 126: same as before

packages/ws-server/src/webSocketServer.ts
```diff
@@ -87,6 +91,51 @@ const methodsCounterByIp = generateMethodsCounterById(register, {
...
+);
+
+const networkUsageGauge = generateGauge(
+  register,
+  {
+    name: WS_CONSTANTS.networkUsageGauge.name,
+    help: WS_CONSTANTS.networkUsageGauge.help,
+    labelNames: WS_CONSTANTS.networkUsageGauge.labelNames,
+  },
+  updateResourceUtilizationMetrics,
```
AlfredoG87, line 136: same as before

AlfredoG87 (changes requested):

PR #2372, opened 2024-04-18
fix: updated MIRROR_NODE_RETRY_DELAY to 1800ms (#2371)

### LINKED ISSUE #2371: Increase MIRROR_NODE_RETRY_DELAY value to expand the delayed duration between repeated retries to the mirror nodes
### Problem

Currently, the default value for MIRROR_NODE_RETRY_DELAY is set to 250. However, this setting many times triggers a 'No matching record' error, indicating that the mirror node requires additional time to update records.

### Solution

Increase the default value for MIRROR_NODE_RETRY_DELAY

### Alternatives

_No response_

quiet-node:
**Description**:
this PR updated the MIRROR_NODE_RETRY_DELAY to 1800ms to extend the delayed duration between repeated retries to the mirror node

Initially it was discussed to be 2000ms, but 2000ms seems a little bit long so eth_getLogs seems to skip blocks. In anoother way, instead of getting the logs in the consecutive block, with delay being 2000ms, eth_getLogs will jump to the second next block in the specified range. 

**Related issue(s)**:

Fixes #2371 

**Notes for reviewer**:

**Checklist**


AlfredoG87 (approved):
LGTM

PR #2393, opened 2024-04-23
feat: support deterministic proxy contract deployment transaction (#2287)

### LINKED ISSUE #2287: Foundry Create2 Support
### Problem

Foundry `CREATE2` requires a predeployed deterministic deployment proxy to be available at `0x4e59b44847b379578588920ca78fbf26c0b4956c`.

The proxy's repo is here: https://github.com/Arachnid/deterministic-deployment-proxy
The way to deploy it is with a Pre EIP 155 transaction, which are now supported on Hedera.
Trying to send the transaction that creates it results in an error of low gas price
```shell
cast publish --rpc-url  https://testnet.hashio.io/api 0xf8a58085174876e800830186a08080b853604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf31ba02222222222222222222222222222222222222222222222222222222222222222a02222222222222222222222222222222222222222222222222222222222222222
```

Results in an error:
```
Error: 
(code: -32009, message: [Request ID: fbda5b21-041b-4e61-b4bf-c6a814304aac] Gas price '100000000000' is below configured minimum gas price '830000000000', data: None)
```

Increasing the Gas price would result in the address of the proxy changing, so it's not possible.

### Solution

Find a workaround to deploy the proxy to Hedera, to increase compability with Foundry and EVM equivalence.

### Alternatives

-

quiet-node:
**Description**:
This PR adds support for the deterministic proxy contract deployment transaction by incorporating a utility function capable of identifying whether a signed transaction corresponds to the deterministic proxy contract deployment transaction. Upon recognition of such a transaction, it will skip the gasPrice precheck and proceed with submission to the network. In addition to this feature, unit tests, integration and e2e tests have been included to ensure the correct behavior.

*Notice: Currently, the `hedera-local` package is at its most recent version and exclusively supports the `NETWORK_NODE_IMAGE_TAG` up to `0.49.0-alpha.3`. However, the `deterministic proxy contract deployment transaction` feature is supported in the `services` at tag `0.49.0-alpha.5`. Consequently, ABI Batch 1 and Websocket Batch 2 CI tasks are configured with `networkTag` set to `0.49.1`, which is the latest official services version encompassing `0.49.0-alpha.5`. Nonetheless, in the near future, when `hedera-local` updates the `NETWORK_NODE_IMAGE_TAG` to a version supporting the `deterministic deployment transaction`, these configurations will be deprecated and removed. [This ticket](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2402) is established to monitor this transition.

**Related issue(s)**:

Fixes #2287 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/constants.ts
```diff
@@ -175,4 +175,10 @@ export default {
     NEW_HEADS: 'newHeads',
     NEW_PENDING_TRANSACTIONS: 'newPendingTransactions',
   },
+
+  // @source: Foundry related constants below can be found at https://github.com/Arachnid/deterministic-deployment-proxy?tab=readme-ov-file#latest-outputs
```
lukelee-sl, line 179: I think we want to remove references to "Foundry" as this can be relevant to others.

packages/relay/tests/lib/precheck.spec.ts
```diff
@@ -247,6 +247,31 @@ describe('Precheck', async function () {
...
+      expect(Precheck.isDeterministicDeploymentTransaction(parsedTxWithNonMatchingChainId)).to.be.false;
+    });
+
+    it('should pass for gas price if the transaction is the deterministic deployment transaction', async function () {
+      const parsedDeterministicDeploymentTransaction = ethers.Transaction.from(
+        constants.DETERMINISTIC_DEPLOYER_TRANSACTION,
+      );
+      const result = precheck.gasPrice(
+        parsedDeterministicDeploymentTransaction,
+        100 * constants.TINYBAR_TO_WEIBAR_COEF,
```
lukelee-sl, line 270: would this test be more meaningful of the gasPrice were a low number like 0 or 1 tinyBar?

lukelee-sl (commented):
Looks good.  Just a comment or two

packages/relay/tests/lib/precheck.spec.ts
quiet-node: Hmm my understanding for this case is this. The `100 * constants.TINYBAR_TO_WEIBAR_COEF` (100 hbars) arg right there is the minimum gasPrice to compared against the `parsedDeterministicDeploymentTransaction.gasPrice` (10 hbars). 

The comparing condition is
```
const passes = txGasPrice >= minGasPrice || Precheck.isDeterministicDeploymentTransaction(tx);
```
in which, `txGasPrice = parsedDeterministicDeploymentTransaction.gasPrice` (10 hbars), and `minGasPrice = 100 * constants.TINYBAR_TO_WEIBAR_COEF` (100 hbars).

So if the minGasPrice is set to 0 or 1, then the condition `txGasPrice >= minGasPrice` (10 >= 1) will be true, and regardless the outcome of the `Precheck.isDeterministicDeploymentTransaction(tx);` this `passes` will be true. 

But if the minGasPrice is set to be greater than 10, then the condition `txGasPrice >= minGasPrice` (10 >= 100) will be false -> now this condition will rely on the outcome of the `Precheck.isDeterministicDeploymentTransaction(tx);` function -> and because it's the deterministic deployment the outcome will be true -> the `passes` will be true.

More, isn't the whole purpose of this feature is to bypass deterministic deployment when its gasPrice (10hbar) is too low compared to  network's gasPrice (minimum is 71 hbars for localnet)?

lukelee-sl: Ah makes sense.  I was incorrect in the way that the parameter was used.

lukelee-sl (approved):
LGTM

PR #2400, opened 2024-04-23
docs: include block range limit in `eth_getLogs` spec

### LINKED ISSUE #2399: Update the Open RPC `eth_getLogs`'s spec to include block range limits when `topics` is present
### Problem

The `eth_getLogs` RPC method has a block range limitation when `topics` param is specified. This limitation comes from the Mirror Node when the query specifies `topics` param.

For example, an RPC call such as

```json
{
    "id": "1686746113447",
    "jsonrpc": "2.0",
    "method": "eth_getLogs",
    "params": [
       {
            "fromBlock":"0x0",
            "toBlock":"latest",
            "address":"0xB9cf2eFcBeCdF96E6A7E46AECd79A784B41Bcf6B",
            "topics":["0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",null,null,"0xd4dac3496a84f55ef2f0d4833a8d70bdf5b19078bb8817a2643b0893b999de06"]
       }
   ]
}
```

fails with the following message `[Request ID: 53703a7f-1951-45f3-a709-df29ef508150] Unknown error invoking RPC`. However, the underlying Mirror Node error is `Cannot search topics without a valid timestamp range: Timestamp range by the lower and upper bounds must be positive and within 7d`.

This means that when `topics` are present, the block range must be within ~`302,400` blocks, the equivalent to 7 days.

### Solution

Update the Open RPC specs `docs/openrpc.json` for `eth_getLogs` to include the 7 days (~ `302,400` blocks) range limitation when `topics` are specified.

### Alternatives

_No response_

acuarica:
**Description**:

This PR fixes #2399.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -402,7 +402,7 @@
         {
             "name": "eth_getLogs",
             "summary": "Returns an array of all logs matching a given filter object.",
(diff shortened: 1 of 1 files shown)


quiet-node:
I broke the websocket CI task down in this PR https://github.com/hashgraph/hedera-json-rpc-relay/pull/2401 so it might help to rebase this on to main

Nana-EC (approved):

PR #2409, opened 2024-04-25
fix: estimate_gas doesn't reflect changes in mirror-node

### LINKED ISSUE #2379: estimate_gas doesn't reflect changes in mirror-node
### Description

The mirror node endpoint contracts/call used to return estimation only for read-only smart contract calls. However, this has changed and now we can also get estimations for transfers, account creations etc. The changes are not reflected in the estimateGas method

### Steps to reproduce

1. Call eth_estimateGas with a transaction transferring value between two accounts
2. You will not get the base cost of 21000 (0x5208)

### Refactoring suggestions

We can keep the try/catch in our code and implement a retry mechanism. In case the mirror node is not available, we can retry the call for N number of attempts. 
If there is still no response we can do the following:
1. For contract deploy the intrinsicGasCost method, from the precheck file, can be used
2. For contract call return some hardcoded value (TBD)
3. For anything else the base cost - 21000

### Hedera network

other

### Version

0.46.0

### Operating system

None

victor-yanev:
**Description**:
This PR updates the fallback calculations of `estimateGas` (in cases where the request to mirror node fails):
- for contract create tx we reuse the `Precheck.transactionIntrinsicGasCost` method
- for contract call tx we have a new constant `TX_CONTRACT_CALL_AVERAGE_GAS: 500_000` which we use instead of using `TX_DEFAULT_GAS: 400_000`
  - Note: `500_000` is an informed guess which adds a small buffer on top of the default gas of `400_000`
  - Couldn't come up with an exact value which might be "correct" most of the times and the `400_000` which is set for `TX_DEFAULT_GAS` is definitely not enough for contract calls and in some scenarios the method would return insufficient amount of gas as an estimation.
  - We can watch out for related bugs and update this to a different value later if we find that it's not sufficient again or if it's giving too large of an estimation

**Related issue(s)**:

Fixes #2379

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -553,81 +556,106 @@ export class EthImpl implements Eth {
...
-        if (
-          this.estimateGasThrows &&
-          e.isContractReverted() &&
-          e.message !== MirrorNodeClientError.messages.INVALID_HEX
-        ) {
-          return predefined.CONTRACT_REVERT(e.detail, e.data);
-        }
-        // Handle Contract Call or Contract Create
-        gas = this.defaultGas;
+  private async predefinedGasForTransaction(
```
konstantinabl, line 599: Add JSDoc here

packages/relay/src/lib/eth.ts
```diff
@@ -553,81 +556,106 @@ export class EthImpl implements Eth {
...
+      this.logger.warn(`${requestIdPrefix} Returning predefined gas for contract call: ${this.contractCallAverageGas}`);
+      return this.contractCallAverageGas;
+    } else {
+      this.logger.warn(`${requestIdPrefix} Returning predefined gas for unknown transaction: ${this.defaultGas}`);
+      return this.defaultGas;
     }
-    return gas;
+  }
+
+  private async getAccount(address: string, requestIdPrefix?: string) {
```
konstantinabl, line 651: JSDoc also here

konstantinabl:
If possible fix some of the SonarCloud issues

packages/relay/src/lib/precheck.ts
```diff
@@ -190,6 +189,7 @@ export class Precheck {
       throw predefined.RESOURCE_NOT_FOUND(`tx.from '${tx.from}'.`);
     }
 
+    let tinybars: bigint;
```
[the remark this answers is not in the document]
victor-yanev: Moved the declaration of the variable closer to where it is defined and used.

It's easier to follow through the code this way if you are reading it in a top-down manner like a book.

ebadiere (commented):
Please resolve the merge issue.

georgi-l95 (approved):

Nana-EC (approved):
LG

PR #2410, opened 2024-04-25
Makes ICacheClient return type promise and changes LRUCache accordingly

konstantinabl:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


.gitignore
```diff
@@ -9,4 +9,4 @@ tsconfig.tsbuildinfo
 **/**/coverage/
 **/**/.nyc_output/
 .DS_Store
-*.xml
+test-results.*.xml
```
victor-yanev, line 12: QQ: Why was this changed? Were there any `.xml` files which were wrongly ignored?

packages/relay/src/lib/clients/cache/ICacheClient.ts
```diff
@@ -19,16 +19,21 @@
  */
 
 export interface ICacheClient {
-  get(key: string, callingMethod: string, requestIdPrefix?: string): any;
-  set(key: string, value: any, callingMethod: string, ttl?: number, requestIdPrefix?: string): void;
-  delete(key: string, callingMethod: string, requestIdPrefix?: string): void;
-  clear(): void;
-  multiSet(keyValuePairs: Record<string, any>, callingMethod: string, requestIdPrefix?: string): void;
+  get(key: string, callingMethod: string, requestIdPrefix?: string | undefined): Promise<any>;
```
victor-yanev, line 22: The presence of `?` already signifies that `requestIdPrefix` could be `undefined`.

packages/relay/src/lib/clients/cache/ICacheClient.ts
```diff
@@ -19,16 +19,21 @@
...
+  set(
+    key: string,
+    value: any,
+    callingMethod: string,
+    ttl?: number | undefined,
+    requestIdPrefix?: string | undefined,
+  ): Promise<void>;
```
victor-yanev, line 29: Same here

packages/relay/src/lib/clients/cache/ICacheClient.ts
```diff
@@ -19,16 +19,21 @@
...
+  multiSet(keyValuePairs: Record<string, any>, callingMethod: string, requestIdPrefix?: string): Promise<void>;
   pipelineSet(
     keyValuePairs: Record<string, any>,
     callingMethod: string,
     ttl?: number | undefined,
     requestIdPrefix?: string,
-  ): void;
-  disconnect(): Promise<void> | void;
+  ): Promise<void>;
+  delete(key: string, callingMethod: string, requestIdPrefix?: string | undefined): Promise<void>;
```
victor-yanev, line 37: Same here

packages/relay/src/lib/clients/cache/localLRUCache.ts
```diff
@@ -123,10 +125,19 @@ export class LocalLRUCache implements ICacheClient {
...
+    callingMethod: string,
+    ttl?: number,
+    requestIdPrefix?: string,
+  ): Promise<void> {
+    return new Promise((resolve) => {
+      const resolvedTtl = ttl ?? this.options.ttl;
+      this.logger.trace(`${requestIdPrefix} caching ${key}:${JSON.stringify(value)} for ${resolvedTtl} ms`);
+      this.cache.set(key, value, { ttl: resolvedTtl });
+      resolve();
+    });
```
victor-yanev, line 140: Simply adding `async` to the method signature already makes this method return `Promise<void>` instead of `void`. 

When you don't explicitly return a `Promise` in an async method, the value you return will automatically be wrapped in a `Promise`. 

Thus, those changes are unneeded and we can revert this (same for other methods in this class where we haven't added anything additional to the behaviour of the methods).

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -91,12 +91,13 @@ export class CacheService {
     this.register = register;
 
     this.internalCache = new LocalLRUCache(logger.child({ name: 'localLRUCache' }), register);
```
victor-yanev, line 93: QQ: Is this `LocalLRUCache` object ever going to be used if `this.isSharedCacheEnabled === true`? 

We can instantiate it like this to avoid wasting heap space:

```diff
-    this.internalCache = new LocalLRUCache(logger.child({ name: 'localLRUCache' }), register);
     ...
     if (this.isSharedCacheEnabled) {
       this.sharedCache = new RedisCache(logger.child({ name: 'redisCache' }), register);
     } else {
-      this.sharedCache = this.internalCache;
+      this.sharedCache = new LocalLRUCache(logger.child({ name: 'localLRUCache' }), register);
     }
```

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -168,11 +171,11 @@ export class CacheService {
    * @param {string} [requestIdPrefix] - The optional request ID prefix.
    * @returns {Promise<any>} A Promise that resolves with the cached value or null if not found.
    */
-  public getSharedWithFallback(key: string, callingMethod: string, requestIdPrefix?: string): any {
```
victor-yanev, line 171: QQ: Do we need to specify such details to the user of this class? I don't think the person using this class will care about implementation details, he wants to just retrieve something from the cache.

I think that's It would be simpler if this was named just `get` or `getAsync`. Especially, since we only provide this one as a public method for getting something from the cache, we don't need to specify details in the method's name.

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -184,7 +187,7 @@ export class CacheService {
    * @param {string} [requestIdPrefix] - The optional request ID prefix.
    * @returns {Promise<any>} A Promise that resolves with the cached value or null if not found.
    */
-  public get(key: string, callingMethod: string, requestIdPrefix?: string): any {
+  private getFromInternalCache(key: string, callingMethod: string, requestIdPrefix?: string): any {
     this.cacheMethodsCounter.labels(callingMethod, CacheService.cacheTypes.LRU, CacheService.methods.GET).inc(1);
 
     return this.internalCache.get(key, callingMethod, requestIdPrefix);
```
victor-yanev, line 193: Shouldn't we also make this method `async`, since it's using the `async get()` method from `this.internalCache`?

```typescript
private async getFromInternalCache(key: string, callingMethod: string, requestIdPrefix?: string): Promise<any> {
  return this.internalCache.get(key, callingMethod, requestIdPrefix);
}
```

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -168,11 +171,11 @@ export class CacheService {
...
    * @returns {Promise<any>} A Promise that resolves with the cached value or null if not found.
    */
-  public getSharedWithFallback(key: string, callingMethod: string, requestIdPrefix?: string): any {
+  public async getSharedWithFallback(key: string, callingMethod: string, requestIdPrefix?: string): Promise<any> {
     if (this.isSharedCacheEnabled) {
-      return this.getAsync(key, callingMethod, requestIdPrefix);
+      return await this.getFromSharedCache(key, callingMethod, requestIdPrefix);
     } else {
-      return this.get(key, callingMethod, requestIdPrefix);
+      return this.getFromInternalCache(key, callingMethod, requestIdPrefix);
```
victor-yanev, line 178: And also change this to:
```
return await this.getFromInternalCache(key, callingMethod, requestIdPrefix);
```

victor-yanev: Please, also check for other such usages which haven't been updated to use `await` for the `LocalLRUCache`

victor-yanev (changes requested):
LG, I left a few suggestions

packages/relay/src/lib/services/cacheService/cacheService.ts
konstantinabl: Yes, i fixed it, thanks

packages/relay/src/lib/services/cacheService/cacheService.ts
konstantinabl: If I introduce the change like this typescript will throw an error when building, since it doesn't allow conditional initialisation in the constructor.
An option is to add non-null assertion at the declaration - like this:
`/**
   * The LRU cache used for caching items from requests.
   *
   * @private
   */
  private readonly internalCache!: ICacheClient;

  /**
   * The Redis cache used for caching items from requests.
   *
   * @private
   */
  private readonly sharedCache!: ICacheClient;`
  Are you ok with this?

konstantinabl: If I introduce the change like this typescript will throw an error when building, since it doesn't allow conditional initialisation in the constructor.
An option is to add non-null assertion at the declaration - like this:
```
/**
   * The LRU cache used for caching items from requests.
   *
   * @private
   */
  private readonly internalCache!: ICacheClient;

  /**
   * The Redis cache used for caching items from requests.
   *
   * @private
   */
  private readonly sharedCache!: ICacheClient;
```
What do you think about this?

packages/relay/src/lib/services/cacheService/cacheService.ts
konstantinabl: You are probably right that it doesn't need to specify details, so I will rename it to getAsync, since the methods is asynchronous

packages/relay/src/lib/services/cacheService/cacheService.ts
victor-yanev: I think we should use only one variable instead of two:
```
  /**
   * The client used for caching items from requests.
   *
   * @private
   */
  private readonly cacheClient: ICacheClient;

  constructor() {
    ...
      this.cacheClient = new RedisCache(logger.child({ name: 'redisCache' }), register);
    } else {
      this.cacheClient = new LocalLRUCache(logger.child({ name: 'localLRUCache' }), register);
    }
  }
```

It's unnecessary to have two variables if we are only going to use one of them.

packages/relay/src/lib/clients/cache/ICacheClient.ts
```diff
@@ -19,16 +19,21 @@
...
-  set(key: string, value: any, callingMethod: string, ttl?: number, requestIdPrefix?: string): void;
-  delete(key: string, callingMethod: string, requestIdPrefix?: string): void;
-  clear(): void;
-  multiSet(keyValuePairs: Record<string, any>, callingMethod: string, requestIdPrefix?: string): void;
+  get(key: string, callingMethod: string, requestIdPrefix?: string): Promise<any>;
+  set(
+    key: string,
+    value: any,
+    callingMethod: string,
+    ttl?: number | undefined,
```
victor-yanev, line 27:
```suggestion
    ttl?: number,
```

packages/relay/src/lib/clients/cache/IRedisCacheClient.ts
```diff
@@ -0,0 +1,5 @@
+import type { ICacheClient } from './ICacheClient';
```
victor-yanev, line 1: nit: add copyright

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -77,7 +77,7 @@ export class CacheService {
 
   private static readonly methods = {
     GET: 'get',
-    GET_ASYNC: 'getAsync',
+    GET_ASYNC: 'getFromSharedCache',
```
victor-yanev, line 80: I believe this is no longer valid since we renamed this method to `getAsync()`

victor-yanev, line 80: Shouldn't this be:
```diff
- GET_ASYNC: 'getFromSharedCache',
+ GET_ASYNC: 'getAsync',
```

packages/relay/src/lib/services/cacheService/cacheService.ts
victor-yanev: Okay, I just saw that we use the internal cache as fallback in case we get errors from redis.

Disregard the comments above. Let's revert this to as it was before - always initializing the LRU client and optionally initializing the redis client based on `this.isSharedCacheEnabled`

konstantinabl: Yeah, you are right, we are supposed to fallback to internalCache, but in some methods I think we dont actually do that
Will probably open an issue about it

packages/server/tests/acceptance/cacheService.spec.ts
```diff
@@ -22,74 +22,37 @@ import { expect } from 'chai';
...
     cacheService = new CacheService(global.logger, registry);
     await new Promise((r) => setTimeout(r, 1000));
-    const connectedClientsAfter = await getConnectedClients();
-    expect(connectedClientsAfter).to.eq(connectedClientsBefore + 1, 'successfully connected to redis server');
-
-    await clearAllTestCacheFromRedis();
-  });
-
-  afterEach(async () => {
-    await clearAllTestCacheFromRedis();
```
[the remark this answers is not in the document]
konstantinabl: @victor-yanev  Well, since I have changed the test and we're not using ioredis anymore, I don't see another way to get the redisCache instance through the CacheSevice and check if it is connected, since the redisCache property is private. If you have any suggestions I am open, maybe I can create some sort of method in the cacheService, but since it is not needed anywhere, I think it is redundant

victor-yanev (dismissed):

georgi-l95 (approved):
LG, a step in the right direction.

victor-yanev (approved):

PR #2422, opened 2024-04-26
fix: updates @open-rpc/schema-utils-js

ebadiere:
Replaces snyk PR#2358


reset.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
Nana-EC, line 1: Q: What's this file for.
No mention of it in the PR description

ebadiere: Mistake.  Closing this PR.

PR #2428, opened 2024-04-26
chore: Revert synk PR #2408.  Docker image won't build.

ebadiere:
Revert docker update to allow image to build.


AlfredoG87 (approved):
LG

Issue #2440, opened 2024-04-29
Websocket API responses contains the `method` field that makes it incompatible with the client libraries & retuns `null`


simsonraj:
### Description

Actual:
The response from Websocket connection contains the `method` field in the JSON response that makes it incompatible with blockchain client frameworks like go-ethereum
Expected:
API methods other than subscribe methods should not contain the `method` field in the response when queried via a websocket connection

### Steps to reproduce

```
package main

import (
	"context"
	"log"

	"github.com/ethereum/go-ethereum/common"
	"github.com/ethereum/go-ethereum/ethclient"
)

func main() {
	client, err := ethclient.Dial("WSS_ENDPOINT")
	if err != nil {
		log.Fatalf("Failed to connect to the Ethereum client: %v", err)
	}
	block, _ := client.BlockByNumber(context.Background(), nil)
	log.Println("Block: ", block.Number())
	txn, _, _ := client.TransactionByHash(context.Background(), common.HexToHash("0x1b8cc543ac8633a6a060629ae176e2817b4792a7841ef38d44ecfb9ec8bfcccc"))
	log.Println("Txn: ", txn.Hash())
}
```

### Hedera network

testnet

### Version

 0.46.4

### Operating system

None

PR #2444, opened 2024-04-30
fix: removed method attribute in WS response object

simsonraj:
**Description**:

The response from Websocket connection contains the method field in the JSON response that makes it incompatible with blockchain client frameworks like go-ethereum

This PR modifies JSON responses from WS for all methods except subscription to make it work with client libraries such as go-ethereum

**Related issue(s)**:
This PR fixes issue  https://github.com/hashgraph/hedera-json-rpc-relay/issues/2440
Fixes #
- it removes `method` from the response

**Notes for reviewer**:

**Checklist**

- [X] Documented (Code comments, README, etc.)

### DIFF
--- a/packages/ws-server/src/utils/utils.ts
+++ b/packages/ws-server/src/utils/utils.ts
@@ -96,7 +96,6 @@ export const sendToClient = (
   connection.send(
     JSON.stringify({
       jsonrpc: '2.0',
-      method,
       result: data,
       id: request.id,
     }),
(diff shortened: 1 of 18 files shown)


packages/ws-server/src/utils/utils.ts
```diff
@@ -96,7 +96,6 @@ export const sendToClient = (
   connection.send(
     JSON.stringify({
       jsonrpc: '2.0',
-      method,
```
quiet-node, line 99: Looks like after removing this field, 'method' param is unused. Could you please update the param object and also the doc comment?

quiet-node (approved):
LG! Thanks for the PR!

ebadiere (approved):

packages/ws-server/src/utils/utils.ts
simsonraj: Thanks for Fixing the params

PR #2467, opened 2024-05-08
fix: Populate null values in transactions.

### LINKED ISSUE #2451: Error from GoEthClient when calling getBlockByNumber on certain specific blocks
### Description

GoEthClient returns error on certain specific block for some reason. 

Here are some of the errors
```
2024/05/06 10:47:32 json: cannot unmarshal non-string into Go struct field rpcBlock.transactions of type hexutil.Uint64
exit status 1
```

```
2024/05/06 10:47:55 missing required field 'maxPriorityFeePerGas' for txdata
exit status 1
```

However, some blocks are working fine. 

### Steps to reproduce

You can try to run this repo to reproduce the problem.

https://github.com/pathornteng/hedera-ethclient-problem

### Additional context

_No response_

### Hedera network

testnet

### Version

latest

### Operating system

None

ebadiere:
Populates null values in transactions.

**Related issue(s)**:

Fixes #2451


packages/relay/src/formatters.ts
```diff
@@ -232,6 +237,36 @@ const stringToHex = (str) => {
...
+    }
+    // If it's a string of a decimal number, convert it
+    if (/^\d+$/.test(value)) {
+      return '0x' + parseInt(value, 10).toString(16);
+    }
+    throw new Error("Invalid input string format. Must be a decimal string or hex string with '0x' prefix.");
+  }
+
+  // Throw error if input type is unexpected
+  throw new Error('Invalid input type. Expected number, string, or null.');
```
quiet-node, line 267: I think this is not necessary. The `value` argument itself only accept `number | string | null`, and your `if statements` already cover all those three scenarios, so this error will never be reached.

packages/relay/src/formatters.ts
```diff
@@ -232,6 +237,36 @@ const stringToHex = (str) => {
   return hex;
 };
 
+const formatHex = (hexValue: string): string => (hexValue === '0x' ? '0x0' : hexValue);
+
+const convertToHex = (value: number | string | null): string => {
+  // Check for null value
+  if (value === null || value === undefined) {
```
quiet-node, line 244: since `value` only accepts number|string|null, value === undefined would not make sense.

packages/relay/src/lib/eth.ts
```diff
@@ -2143,6 +2146,7 @@ export class EthImpl implements Eth {
         constants.TYPE_ACCOUNT,
       ]);
       contractResult.to = await this.resolveEvmAddress(contractResult.to, requestIdPrefix);
+      contractResult.chain_id = contractResult.chain_id == null ? this.chain : contractResult.chain_id;
```
quiet-node, line 2149: Just a suggestion
```suggestion
contractResult.chain_id = contractResult.chain_id || this.chain
```

quiet-node (commented):
Nice work. Some questions and suggestions

packages/relay/src/formatters.ts
AlfredoG87: good point @quiet-node, 

anyways, the simples way to eval for both `null`, `undefined` and `empty string` would be:

```
if (!value) {...
```

packages/relay/src/formatters.ts
```diff
@@ -245,26 +280,73 @@ const isValidEthereumAddress = (address: string): boolean => {
...
+  if (typeof val === 'boolean') {
+    val = val ? 1 : 0;
+  }
+
+  if (typeof val === 'number') {
+    val = intToHex(val);
+  }
+
+  if (typeof val === 'object') {
+    val = val.toString('hex');
```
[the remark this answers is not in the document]
AlfredoG87: there should be a unit test to cover this case in any case, but agree with @quiet-node point

AlfredoG87 (commented):
LG, agreed with a few nits from @quiet-node  otherwise looking good.

packages/relay/src/formatters.ts
```diff
@@ -245,26 +280,73 @@ const isValidEthereumAddress = (address: string): boolean => {
...
+
+    if (val === '') {
+      val = '0';
+    }
+  }
+
+  return addHexPrefix(val);
+};
+
+// ganache solution
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/relay/src/formatters.ts
```diff
@@ -245,26 +280,73 @@ const isValidEthereumAddress = (address: string): boolean => {
   return new RegExp(constants.BASE_HEX_REGEX + '{40}$').test(address);
 };
 
+// ganache solution
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/relay/src/lib/eth.ts
```diff
@@ -1888,6 +1889,8 @@ export class EthImpl implements Eth {
     const fromAddress = await this.resolveEvmAddress(contractResult.from, requestIdPrefix, [constants.TYPE_ACCOUNT]);
     const toAddress = await this.resolveEvmAddress(contractResult.to, requestIdPrefix);
 
+    contractResult.chain_id = contractResult.chain_id == null ? this.chain : contractResult.chain_id;
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/relay/src/formatters.ts
ebadiere: Removed.

packages/relay/src/formatters.ts
ebadiere: Removed `value === undefined`

packages/relay/src/lib/eth.ts
ebadiere: Done.

ebadiere:
Replaced by #2495

PR #2469, opened 2024-05-10
refactor: removed specific handlers for jsonrpc methods but added a shared method

### LINKED ISSUE #2472: refactor WS server package
### Problem

At present, there are multiple redundant handlers dedicated to managing individual JSON-RPC methods. However, these handlers largely follow a consistent workflow in processing requests to the relay. Hence, replacing these individual handlers with a single dynamic shared method is a viable option.

### Solution

- remove individual handlers
- add a single dynamic shared method

### Alternatives

_No response_

quiet-node:
**Description**:
- Removed specific handlers for individual JSON-RPC methods, as they all follow the same flow. This PR adds a shared method capable of dynamically processing all supported relay methods in the WS server.
- updated corresponded acceptance tests

**Related issue(s)**:

Fixes #2472

**Notes for reviewer**:

**Checklist**


packages/ws-server/src/utils/utils.ts
```diff
@@ -68,35 +69,82 @@ export const getMultipleAddressesEnabled = (): boolean => {
...
+  response: any,
   logger: any,
   requestIdPrefix: string,
   connectionIdPrefix: string,
 ) => {
+  const TX_INFO = params[0];
+  const BLOCK_PARAM = params[1];
+  const TAG = JSON.stringify({ method, txInfo: TX_INFO, block: BLOCK_PARAM });
+
   logger.info(
```
AlfredoG87, line 93: should this log be info? or maybe trace or debug?

packages/ws-server/src/utils/utils.ts
```diff
@@ -68,35 +69,82 @@ export const getMultipleAddressesEnabled = (): boolean => {
...
+export const handleSendingRequestsToRelay = async (
+  tag: string,
+  args: any[],
+  relay: Relay,
+  logger: any,
+  rpcCallEndpoint: string,
+  requestIdPrefix: string,
+  connectionIdPrefix: string,
+): Promise<any> => {
+  logger.info(`${connectionIdPrefix} ${requestIdPrefix}: Submitting request to relay for tag=${tag}.`);
```
AlfredoG87, line 123: same as above, should be info?

packages/ws-server/src/utils/utils.ts
```diff
@@ -68,35 +69,82 @@ export const getMultipleAddressesEnabled = (): boolean => {
...
+  params: any,
+  method: any,
+  response: any,
   logger: any,
   requestIdPrefix: string,
   connectionIdPrefix: string,
 ) => {
+  const TX_INFO = params[0];
+  const BLOCK_PARAM = params[1];
+  const TAG = JSON.stringify({ method, txInfo: TX_INFO, block: BLOCK_PARAM });
```
AlfredoG87, line 91: could we add a comment as to what is the TAG and its uses?

packages/ws-server/src/utils/utils.ts
```diff
@@ -68,35 +69,82 @@ export const getMultipleAddressesEnabled = (): boolean => {
...
-      id: request.id,
-    }),
-  );
+  connection.send(JSON.stringify(response));
   connection.limiter.resetInactivityTTLTimer(connection);
 };
+
+/**
+ * Handles sending requests to the relay for processing.
+ * @param {string} tag - The tag associated with the request.
```
AlfredoG87, line 105: maybe we can place here the explanation of what is the tag and its uses.

packages/ws-server/src/utils/utils.ts
AlfredoG87, line 91: Why all capital names?

packages/ws-server/src/utils/utils.ts
```diff
@@ -68,35 +69,82 @@ export const getMultipleAddressesEnabled = (): boolean => {
...
+ * @param {any} args - The arguments to be passed to the relay.
+ * @param {Relay} relay - The relay instance used to process the request.
+ * @param {any} logger - The logger instance used for logging.
+ * @param {string} rpcCallEndpoint - The endpoint on the relay to call.
+ * @param {string} requestIdPrefix - The prefix to use for the request ID.
+ * @param {string} connectionIdPrefix - The prefix to use for the connection ID.
+ * @returns {Promise<any>} A promise that resolves to the result of the request.
+ */
+export const handleSendingRequestsToRelay = async (
+  tag: string,
```
AlfredoG87, line 115: why on this method we receive TAG already formed instead of receiving both `method` and `params` and we create the Tag here (similar to sendToClient method above), maybe we can create a method that receives the method and params and returns the TAG, that can be shared and used whenever necessary, so we can explain there the usage of TAG, and also is more descriptive to anyone reading the code for the first time.

Also if we chose to send both method and params instead of the TAG, we could get rid of `args` parameter and call the `resolvedParams` in here, further simplifying the logic a bit.

AlfredoG87 (commented):
LG, some nits and suggestions and a few questions

packages/ws-server/src/utils/utils.ts
quiet-node: Ah yeah good catch. I'm chaning them to `trace`.

packages/ws-server/src/utils/utils.ts
quiet-node: oh this is briliant why didn't I think of it! This would make the code in the server so much cleaner! Thanks so much Fredy!

AlfredoG87 (approved):
LGTM

PR #2486, opened 2024-05-13
fix:  Added the web3_clientVersion to the websocket server.

### LINKED ISSUE #2462: web3_clientVersion returns null with websocket endpoint
### Description

Expected:
web3_clientVersion to return the same value as HTTP endpoint
Current:
returns null value

### Steps to reproduce

{
	"jsonrpc":"2.0",
	"method":"web3_clientVersion",
	"params":[],
	"id":1
}

{"result":null,"jsonrpc":"2.0","id":null}

### Additional context

_No response_

### Hedera network

testnet

### Version

0.47.0-rc3

### Operating system

None

ebadiere:
Added the `web3_clientVersion` to the websocket server.

Fixes #2462

### DIFF
--- a/packages/ws-server/src/controllers/index.ts
+++ b/packages/ws-server/src/controllers/index.ts
@@ -56,11 +56,23 @@ const handleSendingRequestsToRelay = async ({
 
   try {
     const resolvedParams = resolveParams(method, params);
+    const [service, methodName] = method.split('_');
     let txRes;
-    if (method.split('_')[0] === 'eth') {
-      txRes = await relay.eth()[method.split('_')[1]](...resolvedParams, requestIdPrefix);
-    } else {
-      txRes = await relay.web3()[method.split('_')[1]](...resolvedParams, requestIdPrefix);
+    switch (service) {
+      case 'debug':
+        txRes = await relay.debug()[methodName](...resolvedParams, requestIdPrefix);
+        break;
+      case 'eth':
+        txRes = await relay.eth()[methodName](...resolvedParams, requestIdPrefix);
+        break;
+      case 'get':
+        txRes = await relay.get()[methodName](...resolvedParams, requestIdPrefix);
+        break;
+      case 'web3':
+        txRes = await relay.web3()[methodName](...resolvedParams, requestIdPrefix);
+        break;
+      default:
+        throw new Error(`Unsupported JSON RPC service type: ${service}`);
     }
     if (!txRes) {
       logger.trace(
--- a/packages/ws-server/src/controllers/index.ts
+++ b/packages/ws-server/src/controllers/index.ts
@@ -57,23 +57,9 @@ const handleSendingRequestsToRelay = async ({
   try {
     const resolvedParams = resolveParams(method, params);
     const [service, methodName] = method.split('_');
-    let txRes;
-    switch (service) {
-      case 'debug':
-        txRes = await relay.debug()[methodName](...resolvedParams, requestIdPrefix);
-        break;
-      case 'eth':
-        txRes = await relay.eth()[methodName](...resolvedParams, requestIdPrefix);
-        break;
-      case 'get':
-        txRes = await relay.get()[methodName](...resolvedParams, requestIdPrefix);
-        break;
-      case 'web3':
-        txRes = await relay.web3()[methodName](...resolvedParams, requestIdPrefix);
-        break;
-      default:
-        throw new Error(`Unsupported JSON RPC service type: ${service}`);
-    }
+
+    const txRes = await relay[service]()[methodName](...resolvedParams, requestIdPrefix);
+
     if (!txRes) {
       logger.trace(
         `${connectionIdPrefix} ${requestIdPrefix}: Fail to retrieve result for request=${JSON.stringify(
--- a/packages/ws-server/src/controllers/index.ts
+++ b/packages/ws-server/src/controllers/index.ts
@@ -56,7 +56,12 @@ const handleSendingRequestsToRelay = async ({
 
   try {
     const resolvedParams = resolveParams(method, params);
-    const txRes = await relay.eth()[method.split('_')[1]](...resolvedParams, requestIdPrefix);
+    let txRes;
+    if (method.split('_')[0] === 'eth') {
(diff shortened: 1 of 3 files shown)


quiet-node (changes requested):
It's necessary to add `WEB3_CLIENTVERSION: 'web3_clientVersion'` in WS_CONSTANTS.METHODS or else the server will return `Method web3_clientVersion not found`

AlfredoG87 (changes requested):
can we add a test for this aswell?

AlfredoG87 (changes requested):
Can we add an acceptance test aswell?

packages/ws-server/tests/acceptance/web3ClientVersion.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+describe('@web-socket-batch-2 web3_clientVersion', async function () {
+  const METHOD_NAME = 'web3_clientVersion';
+
+  let requestId: string;
+
+  before(async () => {
+    requestId = Utils.generateRequestId();
+  });
+
+  it.only(`Should execute web3_clientVersion on Standard Web Socket and handle valid requests correctly`, async () => {
```
quiet-node, line 35: remove .only

packages/ws-server/tests/acceptance/web3ClientVersion.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    const response = await WsTestHelper.sendRequestToStandardWebSocket(METHOD_NAME, []);
+    expect(response.result).to.exist;
+    expect(response.result).to.be.a('string');
+    expect(response.message).to.not.equal('Method web3_clientVersion not found');
```
quiet-node, line 39: This is probably not necessary. The `Method ... not found` is a check by the verifySupportedMethod() method, and it already has test coverage itself.

packages/ws-server/tests/acceptance/web3ClientVersion.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  let requestId: string;
+
+  before(async () => {
+    requestId = Utils.generateRequestId();
+  });
+
+  it.only(`Should execute web3_clientVersion on Standard Web Socket and handle valid requests correctly`, async () => {
+    const response = await WsTestHelper.sendRequestToStandardWebSocket(METHOD_NAME, []);
+    expect(response.result).to.exist;
+    expect(response.result).to.be.a('string');
```
quiet-node, line 38: This might not be relevant and accurate enough. I'd suggest make a call to the Relay to get the clientVersion and then compare `response` with that result.

quiet-node (changes requested):

packages/ws-server/tests/acceptance/web3ClientVersion.spec.ts
ebadiere: Done.

packages/ws-server/tests/acceptance/web3ClientVersion.spec.ts
ebadiere: Removed.

packages/ws-server/tests/acceptance/web3ClientVersion.spec.ts
ebadiere: Done.

quiet-node (approved):
LGTM

packages/ws-server/src/controllers/index.ts
```diff
@@ -56,7 +56,12 @@ const handleSendingRequestsToRelay = async ({
 
   try {
     const resolvedParams = resolveParams(method, params);
-    const txRes = await relay.eth()[method.split('_')[1]](...resolvedParams, requestIdPrefix);
+    let txRes;
+    if (method.split('_')[0] === 'eth') {
```
Nana-EC, line 60: An if isn't sufficient here and could lead to future bugs.
The relay supports `eth_`, `debug_`, `get_` and `web3_` prefixes.

You don't have to yet support all of them but we should be more specific in our execution pathway management.

Nana-EC (changes requested):
LG.
Execution pathway logic should be improved

packages/ws-server/src/controllers/index.ts
ebadiere: Updated.

packages/ws-server/src/controllers/index.ts
```diff
@@ -56,7 +56,24 @@ const handleSendingRequestsToRelay = async ({
...
+    const [service, methodName] = method.split('_');
+    let txRes;
+    switch (service) {
+      case 'debug':
+        txRes = await relay.debug()[methodName](...resolvedParams, requestIdPrefix);
+        break;
+      case 'eth':
+        txRes = await relay.eth()[methodName](...resolvedParams, requestIdPrefix);
+        break;
+      case 'get':
```
AlfredoG87, line 68: are we supporting any `get` or `debug` pathways yet?

AlfredoG87 (approved):
one question, otherwise looks good

packages/ws-server/src/controllers/index.ts
ebadiere: @Nana-EC asked that the pathways be specific, in the above comments.

quiet-node (changes requested):
a simple improvement

quiet-node (dismissed):

AlfredoG87 (dismissed):
LGTM

packages/ws-server/src/controllers/index.ts
```diff
@@ -56,7 +56,10 @@ const handleSendingRequestsToRelay = async ({
 
   try {
     const resolvedParams = resolveParams(method, params);
-    const txRes = await relay.eth()[method.split('_')[1]](...resolvedParams, requestIdPrefix);
+    const [service, methodName] = method.split('_');
+
+    const txRes = await relay[service]()[methodName](...resolvedParams, requestIdPrefix);
```
Nana-EC, line 61: nit: worth a little comment and explanation here as this relies on a given class and method structure across files

Nana-EC (approved):
LG
An extra comment would be good

Nana-EC (approved):

PR #2489, opened 2024-05-13
fix: removed chainId field for legacy EIP155 transactions (#2468)

### LINKED ISSUE #2468: Wrong formatting of pre-EIP155 transactions
### Description

When an unprotected transaction (pre-EIP155) is returned by the relay using `eth_getTransactionByHash` for example a `chain_id` property of `0x` is returned which results in incompatibility with certain tooling (for example foundry).
The proper behavior is to return a transaction object without a `chain_id` property.

### Steps to reproduce

1. Start the relay
2. Broadcast an unprotected transaction for example
```bash
cast publish --rpc-url http://localhost:7546/ 0xf8a58085174876e800830186a08080b853604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf31ba02222222222222222222222222222222222222222222222222222222222222222a02222222222222222222222222222222222222222222222222222222222222222 
```

### Additional context

_No response_

### Hedera network

other

### Version

v0.47.0

### Operating system

None

quiet-node:
**Description**:
This PR simply remove the chainId field if the transaction result is a legacy EIP155 transaction whose chainId equals to "0x".

**Related issue(s)**:

Fixes #2468

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -147,7 +147,6 @@ const formatContractResult = (cr: any) => {
   const commonFields = {
     blockHash: toHash32(cr.block_hash),
     blockNumber: nullableNumberTo0x(cr.block_number),
-    chainId: cr.chain_id === EMPTY_HEX ? undefined : cr.chain_id,
     from: cr.from.substring(0, 42),
     gas: nanOrNumberTo0x(cr.gas_used),
     gasPrice: toNullIfEmptyHex(cr.gas_price),
@@ -161,6 +160,10 @@ const formatContractResult = (cr: any) => {
     type: nullableNumberTo0x(cr.type),
     v: cr.type === null ? null : nanOrNumberTo0x(cr.v),
     value: nanOrNumberTo0x(cr.amount),
+    // for legacy EIP155 with tx.chainId=0x0, mirror-node will return a '0x' (EMPTY_HEX) value for contract result's chain_id
+    //   which is incompatibile with certain tools (i.e. foundry). By setting this field, chainId, to undefined, the end jsonrpc
+    //   object will leave out this field, which is the proper behavior for other tools to be compatible with.
+    chainId: cr.chain_id === EMPTY_HEX ? undefined : cr.chain_id,
   };
 
   switch (cr.type) {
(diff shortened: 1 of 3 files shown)


packages/relay/src/formatters.ts
```diff
@@ -147,7 +147,7 @@ const formatContractResult = (cr: any) => {
   const commonFields = {
     blockHash: toHash32(cr.block_hash),
     blockNumber: nullableNumberTo0x(cr.block_number),
-    chainId: cr.chain_id,
+    chainId: cr.chain_id === '0x' ? undefined : cr.chain_id,
```
Nana-EC, line 150: Missing UT

Nana-EC (changes requested):
Looks good.
Nice acceptance tests but there's a missing unit test.

AlfredoG87 (dismissed):
LGTM

packages/relay/src/formatters.ts
quiet-node: updated

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -1211,13 +1231,13 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
         const transaction = {
           type: 2,
           chainId: Number(CHAIN_ID),
-          nonce: await relay.getAccountNonce(accounts[2].address, requestId),
+          nonce: await relay.getAccountNonce(accounts[1].address, requestId),
```
quiet-node, line 1234: note for reviewer: since accounts[2] is the payers of most of the test, when it reaches this case, system throws Insufficient Fund error. Therefore, this change is made to update the payer to account[1] for this case.

ebadiere (dismissed):
LG.

packages/relay/src/formatters.ts
```diff
@@ -147,7 +147,7 @@ const formatContractResult = (cr: any) => {
   const commonFields = {
     blockHash: toHash32(cr.block_hash),
     blockNumber: nullableNumberTo0x(cr.block_number),
-    chainId: cr.chain_id,
+    chainId: cr.chain_id === EMPTY_HEX ? undefined : cr.chain_id,
```
Nana-EC, line 150: nit: add a comment on what this does and why so other devs understand

Nana-EC (commented):
LG. One nit on comment

Nana-EC (dismissed):

AlfredoG87 (approved):
LGTM

PR #2492, opened 2024-05-14
fix: removed `RESOURCE_NOT_FOUND` error in eth_getStorageAt when response is null

### LINKED ISSUE #2491: Inconsistent response between ethereum rpc and hedera
### Description

I've noticed that there is difference in RPC response when using method `eth_getStorageAt` between ethereum rpc node and hedera (hashio).

This case occurs only when you try to read storage at address that is "empty" (contains no data). Below I'm pasting example responses:

Executing eth_getStorageAt on **ethereum** node with address that contains **no data** returns only zeroes (which is expected behaviour):
```json
{
    "jsonrpc": "2.0",
    "result": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "id": 0
}
```

Similar request on hedera rpc node returns `error` object in the response instead of `result`. This may break some existing tools that relay on the rpc api as they might expect the field `result` to be always present.
```json
{
    "jsonrpc": "2.0",
    "id": 0,
    "error": {
        "code": -32001,
        "message": "[Request ID: d17881fd-9aa5-47d0-86ad-1949adc7c2d0] Requested resource not found. Cannot find current state for contract address 0x00099000000000000000000000000000002e7a5d at slot=0x0"
    }
}
```

Tests were performed on the `hashio.io` provider for all networks (mainnet, testnet and previewnet). Each environment behaved the same.

### Steps to reproduce

1. Execute `eth_getStorageAt` method on `https://testnet.hashio.io/api` with body
```json
{
    "jsonrpc": "2.0",
    "method": "eth_getStorageAt",
    "params": [
        "0x00099000000000000000000000000000002e7a5d",
        "0x0",
        "latest"
    ],
    "id": 0
}
```

2. Response with error object is returned witch is not expected
```json
{
    "jsonrpc": "2.0",
    "id": 0,
    "error": {
        "code": -32001,
        "message": "[Request ID: d17881fd-9aa5-47d0-86ad-1949adc7c2d0] Requested resource not found. Cannot find current state for contract address 0x00099000000000000000000000000000002e7a5d at slot=0x0"
    }
}
```

Expected result:
```json
{
    "jsonrpc": "2.0",
    "result": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "id": 0
}
```

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet

### Version

NA

### Operating system

None

quiet-node:
**Description**:
This PR refines the syntax to handle the `RESOURCE_NOT_FOUND` error in `eth_getStorageAt`. With this fix, the Relay skips processing when the response from the mirror-node is null and return a 32-byte-zero-hex string instead of throwing error. Acceptance test is added to cover the update.

**Related issue(s)**:

Fixes #2491

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -825,12 +825,7 @@ export class EthImpl implements Eth {
     await this.mirrorNodeClient
       .getContractStateByAddressAndSlot(address, slot, blockEndTimestamp, requestIdPrefix)
       .then((response) => {
-        if (response === null) {
-          throw predefined.RESOURCE_NOT_FOUND(
-            `Cannot find current state for contract address ${address} at slot=${slot}`,
-          );
-        }
-        if (response.state.length > 0) {
+        if (response !== null && response.state.length > 0) {
           result = response.state[0].value;
         }
       })
--- a/packages/relay/tests/lib/eth/eth_getStorageAt.spec.ts
+++ b/packages/relay/tests/lib/eth/eth_getStorageAt.spec.ts
@@ -20,6 +20,7 @@
 import path from 'path';
 import dotenv from 'dotenv';
 import { expect, use } from 'chai';
+import { ethers } from 'ethers';
 import sinon from 'sinon';
 import chaiAsPromised from 'chai-as-promised';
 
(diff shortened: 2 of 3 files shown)


AlfredoG87 (approved):
LGTM

Issue #2457, opened 2024-05-07
Change JSON-RPC error code response for onchain revert 


mouseless-eth:
### Problem

Some systems like ERC-4337 (Account Abstraction) use eth_call to extract onchain data by forcing a revert. ([example](https://github.com/eth-infinitism/account-abstraction/blob/2b1aac46a8b532121dab4af739bcb019f8693ae1/contracts/interfaces/IEntryPoint.sol#L204-L210)). These revert bytes are then parsed using libraries like viem. 

Currently viem can't properly parse Hedera's RPC response as the error code for an onchain revert isn't what the library expected. 

For context, here is a sample reverting eth_call response that viem can parse (it parses into the viem type `ContractFunctionRevertError`):
```
{
  "jsonrpc": "2.0",
  "id": 4337,
  "error": {
    "code": 3,
    "message": "execution reverted",
    "data": "0x6ca7b8060000000000000000000000007152e0855bab82b8e10b8692e584ad034bd22912"
  }
}
```

This is what Hedera's reverting message looks like and it is being parsed into a `CallExecutionError` viem type

```
{
  "error": {
    "code": -32008,
    "message": "[Request ID: bf1be79b-7d2f-42ee-bc82-da42c9ce5e8f] execution reverted: CONTRACT_REVERT_EXECUTED",
    "data": "0x6ca7b8060000000000000000000000007152e0855bab82b8e10b8692e584ad034bd22912"
  },
  "jsonrpc": "2.0",
  "id": 4337
}
```

[viem checks if the error code is 3](https://github.com/wevm/viem/blob/72c406dc48c8bc699a4c7567e42d14a717833a3e/src/utils/errors/getContractError.ts#L59) and if so treats it as a onchain revert and keeps the data field containing the revert bytes. Although couldn't find any eip or spec that says the error code 3 is a hard requirement.

### Solution

Change the error response code from `-32008` to `3` when a onchain revert occurs

quiet-node:
Hello @mouseless-eth, appreciate the nice catch! I did some research and indeed the value `3` is well used by other well-known services like Alchemy, Infura, etc. Here are some refferences to it:
- https://docs.alchemy.com/reference/error-reference#json-rpc-error-codes
- https://docs.infura.io/api/networks/arbitrum/json-rpc-methods/eth_call#returns

Issue #2522, opened 2024-05-21
Align error codes with standard values where neccessary


quiet-node:
### Problem

It’s necessary to align error codes with industry-standard values to ensure compatibility with various tools such as Viem, Hardhat, Foundry, and others. By doing so, we can guarantee that these tools can correctly parse and handle the error codes, improving overall interoperability and functionality.

### Solution

Research and update error codes where needed

quiet-node:
ref: https://ethereum-json-rpc.com/errors
ref: https://docs.alchemy.com/reference/error-reference#json-rpc-error-codes
Looks like the error codes are aligned with other services in the industry.

PR #2523, opened 2024-05-21
fix: updated CONTRACT_REVERT error code to the value `3`

quiet-node:
**Description**:
This PR changes the error code for CONTRACT_REVERT error to the value 3 so it can be compatible with ethereum tools like Viem, Hardhat, etc.

**Related issue(s)**: https://github.com/hashgraph/hedera-json-rpc-relay/issues/2522

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/2457

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -37,7 +37,7 @@ export class JsonRpcError {
 export const predefined = {
   CONTRACT_REVERT: (errorMessage?: string, data: string = '') =>
     new JsonRpcError({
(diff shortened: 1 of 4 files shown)


ebadiere (approved):
LG.

AlfredoG87 (approved):

AlfredoG87:
@quiet-node  we should add this on the next release `0.48` I think is a good win, and we can unblock some community members

quiet-node:
> @quiet-node we should add this on the next release `0.48` I think is a good win, and we can unblock some community members

Agree

PR #2551, opened 2024-05-30
fix: fixed requests throw error when both data and input fields are present in transaction object

### LINKED ISSUE #2547: Accept the requests with both "data" & "input" fields, ignore the extra parameter instead of throwing error
### Problem

When we try to send a request with both the Input & data fields, it errors out, we do this in order to work with some chains/client which accept only Input or only Data, but they ignore the other one, instead of throwing an error

```{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "eth_call",
    "params": [
        {
            "from": "0x51E3B84966c042C278dEb61564C106B6b82aF792",
            "input": "0x7cb9357e",
            "data": "0x7cb9357e",
            "to": "0x1f35616377069168ce06dD4b36ab8A47A36Ad92f"
        },
        "latest"
    ]
}```

### Solution

- Accept requests with both fields
- Ignore the extra param

### Alternatives

_No response_

### LINKED ISSUE #2550: eth_call mistakenly accepts tx.value as calldata when tx.data is undefined
### Description

Currently, eth_call mistakenly accepts tx.value as calldata when tx.data is undefined. `value` field is designed to hold `the integer of value sent with this transaction encoded as hexadecimal`, while calldata is `the hash of the method signature and encoded parameters`, and is represented by the `data` or `input` field.

### Steps to reproduce

[This line](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1552C4-L1552C57) set `call.value` to be the `calldata` if `call.data` is absent.

### Additional context

_No response_

### Hedera network

### Version

latest

### Operating system

None

quiet-node:
**Description**:
This PR fixed requests throw error when both data and input fields are present in transaction object. 
- transaction object has only `data` field -> no modification
- transaction object has only `input` field -> copy value of `input` to `data` & delete `input`
- transac object has both `data` and ` input` field 
    - same value -> keep `data` & delete `input`
    - different value -> copy value of `input` to `data` & delete `input`

**Related issue(s)**:

Fixes #2547 
Fixes #2550

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -561,15 +561,18 @@ export class EthImpl implements Eth {
     _blockParam: string | null,
     requestIdPrefix?: string,
   ): Promise<string | JsonRpcError> {
+    const callData = transaction.data ? transaction.data : transaction.input;
+    const callDataSize = callData ? callData.length : 0;
+
+    if (callDataSize >= constants.FUNCTION_SELECTOR_CHAR_LENGTH) {
+      this.ethExecutionsCounter
+        .labels(EthImpl.ethEstimateGas, callData.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH))
+        .inc();
+    }
+
     this.logger.trace(
       `${requestIdPrefix} estimateGas(transaction=${JSON.stringify(transaction)}, _blockParam=${_blockParam})`,
     );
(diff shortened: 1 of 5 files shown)


ebadiere (approved):
LG.

Issue #2563, opened 2024-06-04
eth_sendRawRequest fails when calldata size is in the range of [2568 bytes, 5217 bytes]


quiet-node:
### Description

Currently, eth_sendRawTransaction utilizes HFS to carry the extra calldata contents if the calldata's size is larger than a certain threshold (5kbs currently). After many experiments and for an unknown reason, if the calldata's size falls into the range of [2568 bytes, 5217 bytes], the request fails and throw `Null entity ID` error. 

### Steps to reproduce

Send eth_sendRawTransaction requests with calldata's size in the range of [2568 bytes, 5217 bytes].

### Hedera network

mainnet, testnet, previewnet, other

### Version

latest

### Operating system

None

Issue #2544, opened 2024-05-29
Proxy deployment issue with JSON RPC Relay


RaphaelMessian:
### Description

Hello, I'm currently deploying and testing Rarible's smart contracts.

I'm using this repository for deployment and testing : https://github.com/rarible/protocol-contracts/tree/master/projects/hardhat-deploy.

When deploying contracts, I can't deploy the proxies for contracts [ERC721Rarible_meta ](https://github.com/rarible/protocol-contracts/blob/04bab188c8502a2d9d1b86cc322ce6f9da4daa3d/projects/hardhat-deploy/deploy/004_deploy_erc721_Token_and_Factory.ts#L28) and [ERC1155Rarible_meta](https://github.com/rarible/protocol-contracts/blob/04bab188c8502a2d9d1b86cc322ce6f9da4daa3d/projects/hardhat-deploy/deploy/005_deploy_erc1155_Token_and_Factory.ts#L28), I manage to deploy the implementation of these 2 contracts but the proxy deployment returns nothing.

Using my own relay I see that the proxy deployment returns a transaction success but the relay tries to query a contract that is not created and the transaction displayed as success by the relay turns out to be a failure.

In the deployment scripts provided with the repository, proxy deployment for other contracts works correctly for example the [RoyaltiesRegistry contract](https://github.com/rarible/protocol-contracts/blob/04bab188c8502a2d9d1b86cc322ce6f9da4daa3d/projects/hardhat-deploy/deploy/001_deploy_RoyaltiesRegistry.ts#L13), this problem only concerns the deployment of proxies for these 2 contracts.

### Steps to reproduce

1- In the folder projects\hardhat-deploy\utils\config you need to add a file hedera.json with this config : 
`{
    "deploy_meta": true,
    "deploy_non_meta": false
  }`
2 -  You need to add the hedera network in the projects\hardhat-deploy\hardhat.config.ts, like : 
`hedera: {
      url: "https://testnet.hashio.io/api",
      timeout: 2000000000,
      accounts: [process.env.PrivateKeyECDSA]
    }`
3 - You can now run the deployment with npx hardhat deploy --network hedera.

### Additional context

[...]

PR #2566, opened 2024-06-04
fix: fixed condition to employ HFS in submitEthereumTransaction()

quiet-node:
**Description**:
Currently, when an `eth_sendRawTransaction` request contains a large `callData`, the Relay uses HFS to create a new file to handle the contents of the `callData`. The condition to use HFS is when `callData.length` exceeds the `fileChunkSize` value. However, in the original setup, the condition is evaluated by comparing the entire `ethereumTransactionData` object, not just `ethereumTransactionData.callData`. This leads to the mismatch between the `callData` size and the `fileChunkSize` value, and therefore throw `Null entity ID` error, when the `callData` size falls within the range of 2568 bytes to 5217 bytes, as described in [issue #2563](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2563).

This PR fixes the problem by using `ethereumTransactionData.callData` for the HFS condition check.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/2563
Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/2544

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -250,7 +250,8 @@ export class SDKClient {
     const interactingEntity = ethereumTransactionData.toJSON()['to'].toString();
     let fileId: FileId | null = null;
 
-    if (ethereumTransactionData.toBytes().length <= 5120) {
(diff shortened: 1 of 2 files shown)


ebadiere (approved):
LG.

PR #2586, opened 2024-06-10
feat: Golang JSON RPC methods usage example.

arianejasuwienas:
**Description**:
Golang JSON RPC methods usage example.


tools/golang-json-rpc-tests/main.go
```diff
@@ -0,0 +1,412 @@
...
+    msg := ethereum.CallMsg{
+        From: fromAddress,
+        To:   &fromAddress,
+        Data: []byte("0x"),
+    }
+    result, err := client.CallContract(context.Background(), msg, nil)
+    if err != nil {
+        log.Fatalf("Failed to call contract: %v", err)
+    }
+    fmt.Printf("eth_call result", hex.EncodeToString(result))
```
quiet-node, line 156: missing format directives

tools/golang-json-rpc-tests/main.go
```diff
@@ -0,0 +1,412 @@
...
+    if balance == nil {
+        log.Fatalf("Balance is nil")
+    }
+    fmt.Printf("Account balance: %s\n", balance.String())
+}
+
+func testEthCall(client *ethclient.Client, fromAddress common.Address) {
+    msg := ethereum.CallMsg{
+        From: fromAddress,
+        To:   &fromAddress,
```
quiet-node, line 149: Should we try a real SC's address to make sure it works in practice? Otherwise, this will return an empty array.

quiet-node: Plus I believe the program is missing a test case for `eth_sendRawTransaction`. We can use this test case to deploy an SC. This way, we can have an SC to work with and adequately cover the functionality of `eth_sendRawTransaction`.

tools/golang-json-rpc-tests/main.go
```diff
@@ -0,0 +1,412 @@
...
+
+func testCodeAt(client *ethclient.Client, address common.Address) {
+    code, err := client.CodeAt(context.Background(), address, nil)
+    if err != nil {
+        log.Fatalf("Failed to get code at address: %v", err)
+    }
+    if code == nil {
+        log.Fatalf("Code is nil")
+    }
+    fmt.Printf("Code at address %s: %s\n", address.Hex(), hex.EncodeToString(code)) // Not a SmartContract so empty string expected
```
quiet-node, line 315: If we decide to add coverage for `eth_sendRawTransaction` to deploy an SC, we can use it here so .CodeAt can return something other than empty string.

tools/golang-json-rpc-tests/main.go
```diff
@@ -0,0 +1,412 @@
...
+    logs, err := client.FilterLogs(context.Background(), query)
+    if err != nil {
+        log.Fatalf("Failed to get logs: %v", err)
+    }
+
+    fmt.Printf("Logs for address %s:\n", address.Hex())
+    if len(logs) == 0 {
+        fmt.Printf("No logs found for address %s\n", address.Hex())
+    } else {
+        for _, vLog := range logs { // Not a SmartContract so empty result expected
```
quiet-node, line 333: The SC deployed using eth_sendRawTransaction can be useful here

tools/golang-json-rpc-tests/main.go
```diff
@@ -0,0 +1,412 @@
...
+            fmt.Printf("Log Block Number: %d\n", vLog.BlockNumber)
+            fmt.Printf("Log Index: %d\n", vLog.Index)
+            fmt.Printf("Log Data: %s\n", hex.EncodeToString(vLog.Data))
+        }
+    }
+}
+
+func testStorageAt(client *ethclient.Client, address common.Address, slot string) {
+    slotHash := common.HexToHash(slot)
+    storageValue, err := client.StorageAt(context.Background(), address, slotHash, nil)
```
quiet-node, line 343: The SC deployed using eth_sendRawTransaction can be useful here

tools/golang-json-rpc-tests/main.go
```diff
@@ -0,0 +1,412 @@
...
+func testGetTransactionByBlockHashAndIndex(client *ethclient.Client, blockHash common.Hash, index uint) {
+    tx, err := client.TransactionInBlock(context.Background(), blockHash, index)
+    if err != nil {
+        log.Fatalf("Failed to get transaction by block hash and index: %v", err)
+    }
+    fmt.Printf("Transaction in block hash %s at index %d: %s\n", blockHash.Hex(), index, tx.Hash().Hex())
+}
+
+func testGetTransactionByBlockNumberAndIndex(client *ethclient.Client, blockHash common.Hash, index uint) {
+    tx, err := client.TransactionInBlock(context.Background(), blockHash, index)
```
quiet-node, line 362: Looks like it's supposed to be blockNum here, instead of blockHash?

tools/golang-json-rpc-tests/main.go
```diff
@@ -0,0 +1,412 @@
...
+    if err != nil {
+        log.Fatalf("Failed to get transaction by block number and index: %v", err)
+    }
+    fmt.Printf("Transaction in block number at index %d: %s\n", index, tx.Hash().Hex())
+}
+
+func testGetTransactionByHash(client *ethclient.Client, txHash string) {
```
quiet-node, line 369: Are this function and the `testTransactionByHash()` function serve the same?

tools/golang-json-rpc-tests/main.go
```diff
@@ -0,0 +1,412 @@
...
+    testBlockByNumber(client, blockNumber)
+    testTransactionByHash(client, signedTx.Hash().Hex())
+    testTransactionReceipt(client, signedTx.Hash().Hex())
+    testGetBalance(client, fromAddress)
+    testEthCall(client, fromAddress)
+    testEstimateGas(client, fromAddress)
+    testGetGasPrice(client)
+    testGetAccounts(client)
+    testBlockByHash(client, blockHash)
+    testFeeHistory(client, 5, newestBlock, []float64{10, 50, 90})
```
quiet-node, line 80: duplicated with line 70?

quiet-node (commented):
Great work! Left some comments and Qs.

tools/golang-json-rpc-tests/main.go
mj-blockydevs: @quiet-node 

Go SC deployment and interactions tests were already covered in this PR: 
https://github.com/hashgraph/hedera-json-rpc-relay/pull/2584 .

But thanks for your feedback. I guess I'll add sample SC to this tests as well.

AlfredoG87 (dismissed):
LGTM

tools/golang-json-rpc-tests/main.go
arianejasuwienas: @quiet-node 

Go SC deployment and interactions tests were already covered in this PR: 
https://github.com/hashgraph/hedera-json-rpc-relay/pull/2584 .

But thanks for your feedback. I guess I'll add sample SC to this tests as well.

quiet-node: Thanks it would be nice!

tools/golang-json-rpc-tests/.env.sample
```diff
@@ -0,0 +1,7 @@
+# The operator private keys start with "0x" and are in hexadecimal format
+# Your account ECDSA hex-encoded private key (Ex: 0xb46751179bc8aa9e129d34463e46cd...)
+OPERATOR_PRIVATE_KEY='OPERATOR_PRIVATE_KEY'
+
+# Sample transaction already existing in the network. It will be used to check whether some of the RPC methods inquiring
+# transactions data are working correctly.
+TRANSACTION_HASH=4b970a3c02af09d646d2bc099b94d95cd2c9b7c92d1bf42e65eee0258016a234
```
Nana-EC, line 7: Q: shouldn't this be empty?
Other than mainnet all transaction hashes are not long lived

tools/golang-json-rpc-tests/README.md
```diff
@@ -0,0 +1,103 @@
+# Hedera JSON RPC - Golang Tests Project
+
+This project offers boilerplate code for using golang client methods on Hedera JSON RPC api.
```
Nana-EC, line 3: nit
```suggestion
This project offers boilerplate code for using golang client methods on Hedera JSON RPC API.
```

tools/golang-json-rpc-tests/README.md
```diff
@@ -0,0 +1,103 @@
...
+       github.com/ethereum/go-ethereum/accounts/abi/bind \
+       github.com/ethereum/go-ethereum/crypto \
+       github.com/joho/godotenv
+```
+
+4. Get your Hedera testnet account hex encoded private key from the [Hedera Developer Portal](https://portal.hedera.com/register) and update the `.env.example` `OPERATOR_PRIVATE_KEY`
+
+5. Copy `.env.example` to `.env`
+
+6. Run the test script from the root directory of the project. The default network is set to "testnet."
```
Nana-EC, line 37:
```suggestion
6. Run the test script from the root directory of the project. The default network is set to "testnet"
```

tools/golang-json-rpc-tests/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+```shell
+go run .
+```
+# Known Issues
+
+#### Go Ethereum Client Incompatibility with Hedera JSON RPC Relay
```
Nana-EC, line 44: move this section into a ticket and reference the ticket here instead

tools/golang-json-rpc-tests/main.go
```diff
@@ -0,0 +1,412 @@
+package main
```
Nana-EC, line 1: Missing license header

Nana-EC (commented):
Some suggestions

tools/golang-json-rpc-tests/main.go
arianejasuwienas: It was missing, so I’ve added “Licensed under the Apache License,” similar to the other examples.

tools/golang-json-rpc-tests/.env.sample
arianejasuwienas: Yes, we create the transactions during tests. This environment variable is not required, so I’ve removed it.

tools/golang-json-rpc-tests/README.md
arianejasuwienas: @Nana-EC let me know if [this](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2600) issue will be sufficient, please (I've labeled it as bug).

tools/golang-json-rpc-tests/main.go
mj-blockydevs: As you proposed, I’ve added smart contract deployment to the tests. This smart contract emits an event and has a storage slot filled, so there is an actual result when collecting logs or investigating the storage of this contract.

arianejasuwienas: @quiet-node As you proposed, I’ve added smart contract deployment to the tests. This smart contract emits an event and has a storage slot filled, so there is an actual result when collecting logs or investigating the storage of this contract.

tools/golang-json-rpc-tests/main.go
mj-blockydevs: I didn't notice they were missing before. Added them now.

arianejasuwienas: I didn't notice they were missing before. Added them now.

tools/golang-json-rpc-tests/main.go
arianejasuwienas: This code is no longer empty because of: https://github.com/hashgraph/hedera-json-rpc-relay/pull/2586#discussion_r1650705127

tools/golang-json-rpc-tests/main.go
arianejasuwienas: Yes, this sample SC emits a log, so the response is no longer empty here:

https://github.com/hashgraph/hedera-json-rpc-relay/pull/2586#discussion_r1650705127

tools/golang-json-rpc-tests/main.go
arianejasuwienas: Done it: https://github.com/hashgraph/hedera-json-rpc-relay/pull/2586#discussion_r1650705127

tools/golang-json-rpc-tests/main.go
arianejasuwienas: Yes, and Go ethereum client does not seem to have this call to: eth_getTransactionByBlockNumber implemented at all. So I've just removed this test case from this PR.

tools/golang-json-rpc-tests/main.go
arianejasuwienas: Very similar, and there was no reason to keep both of them. So I've removed one.

tools/golang-json-rpc-tests/main.go
arianejasuwienas: That's right. I've removed the duplicated line.

quiet-node (approved):
Great work! LGTM!

tools/golang-json-rpc-tests/README.md
Nana-EC: It is thanks

PR #2590, opened 2024-06-11
fix: removed name field in JsonRpcError class

### LINKED ISSUE #2589: Relay `error` response is not conforming to the JSON-RPC 2.0 Specification
### Description

Right now, most of the errors are returning a error that includes a `name` field, and this field is not part of the specification, such an thing can cause unforeseeable issues with clients that expect errors to adhere to the spec strictly. as it was the case with this case found on thegraph: https://github.com/hashgraph/hedera-json-rpc-relay/issues/2330

https://www.jsonrpc.org/specification#error_object

### Steps to reproduce

```
curl --location 'https://mainnet.hashio.io/api' \
--data '{
  "id": 3635842853412497,
  "jsonrpc": "2.0",
  "method": "eth_getTransactionCount",
  "params": [
    "aaa",
    "latest"
  ]
}'
```

The response of this is just an example, but a lot of errors are returning a name:

```
{
    "error": {
        "code": -32602,
        "name": "Invalid parameter",
        "message": "[Request ID: 9cd81788-644a-4198-9098-740778d04804] Invalid parameter 0: Expected 0x prefixed string representing the address (20 bytes), value: aaa"
    },
    "jsonrpc": "2.0",
    "id": 3635842853412497
}
```

### Additional context

**Possible breaking change: If and when this is merged and released, we should place a note on the changes that it might break any client that was depending on this non-standard field for their logic.**

https://www.jsonrpc.org/specification#error_object

https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/errors/JsonRpcError.ts

### Hedera network

### Version

v0.50.0

### Operating system

None

quiet-node:
**Description**:
This PR removes the `name` field in JsonRpcError class to conform to the JSON-RPC 2.0 Specification. Many tests were updated to apply this new changes.

**Related issue(s)**:

Fixes #2589

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/charts/hedera-json-rpc-relay/postman.json
+++ b/charts/hedera-json-rpc-relay/postman.json
@@ -617,7 +617,6 @@
 							"pm.test(\"Success\", () => {",
 							"    var response = pm.response.json();",
 							"    pm.expect(response.error.code).to.equal(-32601);",
-							"    pm.expect(response.error.name).to.equal(\"Method not found\");",
 							"    pm.expect(response.error.message.endsWith(\"Unsupported JSON-RPC method\")).to.be.true;",
 							"    pm.expect(response.id).to.equal(\"test_id\");",
 							"    pm.expect(response.jsonrpc).to.equal(\"2.0\");",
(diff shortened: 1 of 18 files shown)


AlfredoG87 (commented):
PR Changes are empty 😅

quiet-node:
> PR Changes are empty 😅

😓 yes I reverted the changes to test the Tests/Tests CI and it still failed for some reasons.

AlfredoG87 (approved):
LGTM

PR #2591, opened 2024-06-11
feat: Added support for OPERATOR_KEY_FORMAT

### LINKED ISSUE #2583: Improve PrivateKey import from environment variable
### Problem

Currently to create the Hedera SDK Operator Key we are using the `PrivateKey.fromString` method, that is deprecated and only supports DER and HEX of ED25519 Key Formats, but operators would not be able to use a HEX ECDSA Private Key.

https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/services/hapiService/hapiService.ts#L211

### Solution

Include a new config param that indicates the type of Key that could be:
OPERATOR_KEY_FORMAT

and have values, `DER`, `HEX_ECDSA` or `HEX_ED25519`

by default we can set `DER` as is currently working.

and according to that value it chooses the right method to create the object `PrivateKey`

DER --> PrivateKey.fromStringDer()
HEX_ECDSA --> PrivateKey.fromStringECDSA()
HEX_ED25519 --> PrivateKey.fromStringED25519

### Alternatives

_No response_

ebadiere:
Adds support for `OPERATOR_KEY_FORMAT`

**Related issue(s)**:

Fixes #2583

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -35,11 +35,13 @@ Unless you need to set a non-default value, it is recommended to only populate o
 | `INPUT_SIZE_LIMIT`         | "1mb"            | The [koa-jsonrpc](https://github.com/Bitclimb/koa-jsonrpc) maximum size allowed for requests                                                                                                                            |
 | `MAX_BLOCK_RANGE`          | "5"              | The maximum block number greater than the mirror node's latest block to query for                                                                                                                                       |
 | `OPERATOR_ID_MAIN`         | ""               | Operator account ID used to pay for transactions. In `S.R.N` format, e.g. `0.0.1001`.                                                                                                                                   |
+| `OPERATOR_KEY_FORMAT`        | "DER"               | Operator private key format. Valid types are `DER`, `HEX_ECDSA`, or `HEX_ED25519` |
 | `OPERATOR_KEY_MAIN`        | ""               | Operator private key used to sign transactions in hex encoded DER format. This may be either an ED22519 private key or an ECDSA private key. The private key must be associated with/used to derive `OPERATOR_ID_MAIN`. |
 | `RATE_LIMIT_DISABLED`      | "false"          | Flag to disable IP based rate limiting.                                                                                                                                                                                 |
 | `REQUEST_ID_IS_OPTIONAL`   | ""               | Flag to set it the JSON RPC request id field in the body should be optional. Note, this breaks the API spec and is not advised and is provided for test purposes only where some wallets may be non compliant           |
 | `SERVER_PORT`              | "7546"           | The RPC server port number to listen for requests on. Currently a static value defaulting to 7546. See [#955](https://github.com/hashgraph/hedera-json-rpc-relay/issues/955)                                            |
 | `SERVER_REQUEST_TIMEOUT_MS`| "60000"           | The time of inactivity allowed before a timeout is triggered and the socket is closed. See [NodeJs Server Timeout](https://nodejs.org/api/http.html#serversettimeoutmsecs-callback)                                         |
+
 
 ## Relay
 
--- a/packages/relay/src/lib/services/hapiService/hapiService.ts
+++ b/packages/relay/src/lib/services/hapiService/hapiService.ts
@@ -209,7 +209,7 @@ export default class HAPIService {
    * @returns Client
    */
   private initClient(logger: Logger, hederaNetwork: string, type: string | null = null): Client {
-    let client: Client;
+    let client: Client, privateKey: PrivateKey;
     if (hederaNetwork in constants.CHAIN_IDS) {
       client = Client.forName(hederaNetwork);
     } else {
@@ -218,19 +218,15 @@ export default class HAPIService {
 
     if (type === 'eth_sendRawTransaction') {
       if (process.env.OPERATOR_ID_ETH_SENDRAWTRANSACTION && process.env.OPERATOR_KEY_ETH_SENDRAWTRANSACTION) {
-        client = client.setOperator(
-          AccountId.fromString(process.env.OPERATOR_ID_ETH_SENDRAWTRANSACTION),
-          PrivateKey.fromString(process.env.OPERATOR_KEY_ETH_SENDRAWTRANSACTION),
-        );
+        privateKey = this.initPrivateKey(process.env.OPERATOR_KEY_ETH_SENDRAWTRANSACTION);
+        client = client.setOperator(AccountId.fromString(process.env.OPERATOR_ID_ETH_SENDRAWTRANSACTION), privateKey);
       } else {
         logger.warn(`Invalid 'ETH_SENDRAWTRANSACTION' env variables provided`);
       }
     } else {
       if (process.env.OPERATOR_ID_MAIN && process.env.OPERATOR_KEY_MAIN) {
-        client = client.setOperator(
-          AccountId.fromString(process.env.OPERATOR_ID_MAIN.trim()),
-          PrivateKey.fromString(process.env.OPERATOR_KEY_MAIN),
-        );
+        privateKey = this.initPrivateKey(process.env.OPERATOR_KEY_MAIN);
+        client = client.setOperator(AccountId.fromString(process.env.OPERATOR_ID_MAIN.trim()), privateKey);
       } else {
         logger.warn(`Invalid 'OPERATOR' env variables provided`);
       }
@@ -248,6 +244,19 @@ export default class HAPIService {
     );
 
     return client;
+  }
+
+  private initPrivateKey(operatorMainKey: string): PrivateKey {
+    switch (process.env.OPERATOR_KEY_FORMAT) {
+      case 'DER':
+        return PrivateKey.fromStringDer(operatorMainKey);
+      case 'HEX_ED25519':
+        return PrivateKey.fromStringED25519(operatorMainKey);
+      case 'HEX_ECDSA':
+        return PrivateKey.fromStringECDSA(operatorMainKey);
+      default:
+        return PrivateKey.fromStringDer(operatorMainKey);
+    }
   }
 
   /**
--- a/packages/relay/src/lib/services/hapiService/hapiService.ts
+++ b/packages/relay/src/lib/services/hapiService/hapiService.ts
@@ -249,9 +249,7 @@ export default class HAPIService {
   private createPrivateKeyBasedOnFormat(operatorMainKey: string): PrivateKey {
     switch (process.env.OPERATOR_KEY_FORMAT) {
       case 'DER':
-        return PrivateKey.fromStringDer(operatorMainKey);
       case undefined:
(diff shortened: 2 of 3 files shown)


AlfredoG87 (approved):
LGTM 👍

docs/configuration.md
```diff
@@ -35,12 +35,14 @@ Unless you need to set a non-default value, it is recommended to only populate o
 | `INPUT_SIZE_LIMIT`         | "1mb"            | The [koa-jsonrpc](https://github.com/Bitclimb/koa-jsonrpc) maximum size allowed for requests                                                                                                                            |
 | `MAX_BLOCK_RANGE`          | "5"              | The maximum block number greater than the mirror node's latest block to query for                                                                                                                                       |
 | `OPERATOR_ID_MAIN`         | ""               | Operator account ID used to pay for transactions. In `S.R.N` format, e.g. `0.0.1001`.                                                                                                                                   |
+| `OPERATOR_KEY_FORMAT`        | "DER"               | Operator private key format. Valid types are `DER`, `HEX_ECDSA`, or `HEX_ED25519` |
```
quiet-node, line 38: This might not be entirely true when it says that the valid types are `DER`, `HEX_ECDSA`, or `HEX_ED22519`. Because [here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2591/files#diff-e959cfa5abcb8b82e4d8d85f4476d00f788c50a80e6e9081b390f09082f58ad2R257) is a `default` block and will read the operator key as DER type. This means `OPERATOR_KEY_FORMAT` is not necessary be `DER`, `HEX_ECDSA`, or `HEX_ED25519` because the `initPrivateKey` function will accept any arbitrary values. `OPERATOR_KEY_FORMAT` can be "HEDERA", or "HBAR", or "swirlds-labs", "bird", etc. and the `initPrivateKey` will still read the `OPERATOR_KEY_MAIN` as DER.

packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -250,6 +246,19 @@ export default class HAPIService {
...
 
+  private initPrivateKey(operatorMainKey: string): PrivateKey {
+    switch (process.env.OPERATOR_KEY_FORMAT) {
+      case 'DER':
+        return PrivateKey.fromStringDer(operatorMainKey);
+      case 'HEX_ED25519':
+        return PrivateKey.fromStringED25519(operatorMainKey);
+      case 'HEX_ECDSA':
+        return PrivateKey.fromStringECDSA(operatorMainKey);
+      default:
```
quiet-node, line 257: we should either throw an error if the `OPERATOR_KEY_FORMAT` is not `DER`, `HEX_ED25519`, or `HEX_ECDSA`, or if we want to accept arbitrary values of `OPERATOR_KEY_FORMAT`, then we can remove the case 'DER` block as it's gonna fall into the default block anyway.

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -174,4 +175,39 @@ describe('SdkClient', async function () {
...
+      const hapiService = new HAPIService(logger, registry, hbarLimiter, new CacheService(logger, registry));
+      const privateKey = (hapiService as any).initPrivateKey.call(
+        hapiService,
+        '0x08e926c84220295b5db5df25be107ce905b41e237ac748dd04d479c23dcdf2d5',
+      );
+      expect(privateKey.toString()).to.eq(
+        '3030020100300706052b8104000a0422042008e926c84220295b5db5df25be107ce905b41e237ac748dd04d479c23dcdf2d5',
+      );
+    });
+  });
```
quiet-node, line 212: I think grouping the key pairs into a variable will make the test easier and cleaner.
```suggestion
  describe('HAPIService', async () => {
    const OPERATOR_KEY_ED25519 = {
      DER: '302e020100300506032b65700422042091132178e72057a1d7528025956fe39b0b847f200ab59b2fdd367017f3087137',
      HEX_ED25519: '0x91132178e72057a1d7528025956fe39b0b847f200ab59b2fdd367017f3087137',
    };

    const OPERATOR_KEY_ECDSA = {
      DER: '3030020100300706052b8104000a0422042008e926c84220295b5db5df25be107ce905b41e237ac748dd04d479c23dcdf2d5',
      HEX_ECDSA: '0x08e926c84220295b5db5df25be107ce905b41e237ac748dd04d479c23dcdf2d5',
    };

    it('Initialize the privateKey for default which is DER', async () => {
      const hapiService = new HAPIService(logger, registry, hbarLimiter, new CacheService(logger, registry));
      const privateKey = (hapiService as any).initPrivateKey.call(hapiService, OPERATOR_KEY_ED25519.DER);
      expect(privateKey.toString()).to.eq(OPERATOR_KEY_ED25519.DER);
    });

    it('Initialize the privateKey for OPERATOR_KEY_FORMAT set to DER', async () => {
      process.env.OPERATOR_KEY_FORMAT = 'DER';
      const hapiService = new HAPIService(logger, registry, hbarLimiter, new CacheService(logger, registry));
      const privateKey = (hapiService as any).initPrivateKey.call(hapiService, OPERATOR_KEY_ECDSA.DER);
      expect(privateKey.toString()).to.eq(OPERATOR_KEY_ECDSA.DER);
    });

    it('Initialize the privateKey for OPERATOR_KEY_FORMAT set to HEX_ED25519', async () => {
      process.env.OPERATOR_KEY_FORMAT = 'HEX_ED25519';
      const hapiService = new HAPIService(logger, registry, hbarLimiter, new CacheService(logger, registry));
      const privateKey = (hapiService as any).initPrivateKey.call(hapiService, OPERATOR_KEY_ED25519.HEX_ED25519);
      expect(privateKey.toString()).to.eq(OPERATOR_KEY_ED25519.DER);
    });

    it('Initialize the privateKey for OPERATOR_KEY_FORMAT set to HEX_ECDSA', async () => {
      process.env.OPERATOR_KEY_FORMAT = 'HEX_ECDSA';
      const hapiService = new HAPIService(logger, registry, hbarLimiter, new CacheService(logger, registry));
      const privateKey = (hapiService as any).initPrivateKey.call(hapiService, OPERATOR_KEY_ECDSA.HEX_ECDSA);
      expect(privateKey.toString()).to.eq(OPERATOR_KEY_ECDSA.DER);
    });
  });
  ```

File: packages/relay/src/lib/services/hapiService/hapiService.ts [hunk 4]
```diff
@@ -250,6 +246,19 @@ export default class HAPIService {
     return client;
   }
 
+  private initPrivateKey(operatorMainKey: string): PrivateKey {
```
quiet-node, line 249: I would change the name of this method to like `resolvePrivateKey` or something like that it's just like it's not initiating a private key but instead it kinda resolves the string key to `PrivateKey` type? Hmm I'm not entirely sure about this tbh 😓

quiet-node (changes requested):
Great work! Just some suggestion

packages/relay/src/lib/services/hapiService/hapiService.ts
AlfredoG87: giving good names to variables and methods is one of the hardest things in development. I would suggest:
`resolvePrivateKeyBasedOnFormat`, maybe is long, but describes what the method is doing 😅

packages/relay/src/lib/services/hapiService/hapiService.ts
AlfredoG87: I agree that it should use the `default` when no format is explicitly specified, otherwise it would be better to throw an error. of `INVALID_PRIVATE_KEY_FORMAT` or something of sorts. that would also resolve your previous comment on documentation being wrong @quiet-node. 👍

docs/configuration.md
ebadiere: Okay, yes, will update.

packages/relay/src/lib/services/hapiService/hapiService.ts
ebadiere: Done.

docs/configuration.md
ebadiere: Handled.

packages/relay/src/lib/services/hapiService/hapiService.ts
ebadiere: Done.

AlfredoG87 (dismissed):
LGTM

packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -250,6 +246,23 @@ export default class HAPIService {
...
   }
 
+  private createPrivateKeyBasedOnFormat(operatorMainKey: string): PrivateKey {
+    switch (process.env.OPERATOR_KEY_FORMAT) {
+      case 'DER':
+        return PrivateKey.fromStringDer(operatorMainKey);
+      case undefined:
+        return PrivateKey.fromStringDer(operatorMainKey);
+      case null:
+        return PrivateKey.fromStringDer(operatorMainKey);
```
quiet-node, line 256:
```suggestion
      case 'DER':
      case undefined:
      case null:
        return PrivateKey.fromStringDer(operatorMainKey);
```

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -174,4 +176,53 @@ describe('SdkClient', async function () {
...
+      DER: '302e020100300506032b65700422042091132178e72057a1d7528025956fe39b0b847f200ab59b2fdd367017f3087137',
+      HEX_ED25519: '0x91132178e72057a1d7528025956fe39b0b847f200ab59b2fdd367017f3087137',
+    };
+
+    const OPERATOR_KEY_ECDSA = {
+      DER: '3030020100300706052b8104000a0422042008e926c84220295b5db5df25be107ce905b41e237ac748dd04d479c23dcdf2d5',
+      HEX_ECDSA: '0x08e926c84220295b5db5df25be107ce905b41e237ac748dd04d479c23dcdf2d5',
+    };
+
+    it('Initialize the privateKey for default which is DER', async () => {
```
quiet-node, line 191: Since the default is now throwing error, we should update this test case. 
I think we need:
- a test case that process.env.OPERATOR_KEY_FORMAT is not configured, i.e. undefined
- a test case that process.env.OPERATOR_KEY_FORMAT is null
- a test case that process.env.OPERATOR_KEY_FORMAT is set to arbitrary value and expect an error to be thrown

quiet-node (changes requested):

packages/relay/tests/lib/sdkClient.spec.ts
ebadiere: There is already a test to cover an arbitrary value that expects an error.  It's 'It should throw an Error when an unexpected string is set'.  The other suggestions are done.

PR #2601, opened 2024-06-17
chore: Increase batch request limit

### LINKED ISSUE #2437: Increase batch request limit
Currently by default ethers and other libraries are using batch requests by default to receive information by sending `eth_chainId`, `eth_blockNumber` and more in one batch. However we are rate limiting and counting this as a single request towards tier one (similar to `eth_sendRawTransaction`). We should split up those requests and count them to their respective tiers. Otherwise, we might degrade the performance of tools using the relay.
As we can't force users to turn off batching, we need to properly limit it.

victor-yanev:
**Description**:
In this PR we are spliting up each individual request from the batch request and counting them towards their respecitve tier. This will give us more accurate rate limiting and won't degrade the performance of the tools using it. 

Also added some technical improvements to the `KoaJsonRpc` class and added typing / fixed code smells in some tests.

**Related issue(s)**:

Fixes #2437 

**Notes for reviewer**:

**Checklist**


quiet-node (commented):
Great work! Couple of Q and request. Also, batch request is also rate limited in the WS server https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/ws-server/src/webSocketServer.ts#L149. Could you also update the WS server with this new logic?

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "root",
   "devDependencies": {
     "@types/chai-as-promised": "^7.1.5",
+    "@types/co-body": "6.1.0",
```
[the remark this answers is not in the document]
victor-yanev: The TypeScript compiler builds `.d.ts` files which serve as type definition files for JavaScript modules from all `@types/<module-name>` dependencies. They provide a way to include type information for libraries or modules that are written in plain JavaScript, enabling TypeScript to perform type-checking on such code. Therefore, this dependency only adds type information for the `co-body@6.1.0` package, which was already used.

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -196,26 +166,17 @@ export default class KoaJsonRpc {
...
-      return jsonResp(null, new IPRateLimitExceeded(BATCH_REQUEST_METHOD_NAME), undefined);
-    }
-
     const response: any[] = [];
-    ctx.state.methodName = BATCH_REQUEST_METHOD_NAME;
 
     // we do the requests in parallel to save time, but we need to keep track of the order of the responses (since the id might be optional)
-    const promises = body.map((item: any) => {
+    const promises: Promise<any>[] = body.map(async (item: any) => {
+      ctx.state.methodName = item.method;
```
Nana-EC, line 173: Q: What was the lack of this line leading to? Empty string in some logs or low metrics numbers?
Good catch.

packages/server/src/koaJsonRpc/lib/methodConfiguration.ts
```diff
@@ -24,12 +24,20 @@ dotenv.config({ path: path.resolve(__dirname, '../../../../../.env') });
 
 import CONSTANTS from '../../../../relay/dist/lib/constants';
 
-const tier1rateLimit = process.env.TIER_1_RATE_LIMIT || CONSTANTS.DEFAULT_RATE_LIMIT.TIER_1;
-const tier2rateLimit = process.env.TIER_2_RATE_LIMIT || CONSTANTS.DEFAULT_RATE_LIMIT.TIER_2;
-const tier3rateLimit = process.env.TIER_3_RATE_LIMIT || CONSTANTS.DEFAULT_RATE_LIMIT.TIER_3;
+const tier1rateLimit = parseInt(process.env.TIER_1_RATE_LIMIT ?? CONSTANTS.DEFAULT_RATE_LIMIT.TIER_1.toString());
```
Nana-EC, line 27: Q: this changes the data type from string to int right?
Was something broken here or what's the benefit of the change?
Maybe it avoids casting later?

packages/server/src/koaJsonRpc/lib/utils.ts
```diff
@@ -19,8 +19,33 @@
...
+export function hasOwnProperty(obj: any, prop: PropertyKey): boolean {
+  return Object.prototype.hasOwnProperty.call(obj, prop);
+}
 
 export function setServerTimeout(server: Server): void {
   const requestTimeoutMs = parseInt(process.env.SERVER_REQUEST_TIMEOUT_MS ?? '60000');
   server.setTimeout(requestTimeoutMs);
 }
+
+export function getBatchRequestsMaxSize(): number {
```
Nana-EC, line 33: Q: With these methods why required a cast every time it's called. These seem to be pure in nature so why not do once on class instantiation and then returned the stored const value?

packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -30,13 +30,13 @@ import { ContractId } from '@hashgraph/sdk';
 
 // Assertions and constants from local resources
 import Assertions from '../helpers/assertions';
-import RelayAssertions from '../../../../packages/relay/tests/assertions';
+import RelayAssertions from '@hashgraph/json-rpc-relay/tests/assertions';
```
Nana-EC, line 33: We keep going back and forth on this standard ever 6 months.
I believe using this identifier means you have to `npm run build` to get local changes to get picked up.
Other than avoiding the need to change the path on file moves does this have other benefits?

Nana-EC (commented):
Looking good, lots of good changes.
Left a few comments.

Also in general i'd prefer we keep PRs focused on connected changes. This should have been 2 PRs, one for the bacth stuff and one for misc improvements

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -27,104 +27,83 @@ import path from 'path';
...
-const INVALID_PARAMS_ERROR = 'INVALID PARAMS ERROR';
 const INVALID_REQUEST = 'INVALID REQUEST';
-const IP_RATE_LIMIT_EXCEEDED = 'IP RATE LIMIT EXCEEDED';
-const JSON_RPC_ERROR = 'JSON RPC ERROR';
-const CONTRACT_REVERT = 'CONTRACT REVERT';
-const METHOD_NOT_FOUND = 'METHOD NOT FOUND';
 const REQUEST_ID_HEADER_NAME = 'X-Request-Id';
 const responseSuccessStatusCode = '200';
-const BATCH_REQUEST_METHOD_NAME = 'batch_request';
+const METRIC_HISTOGRAM_NAME = 'rpc_relay_method_result';
```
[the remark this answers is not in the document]
victor-yanev: The `const BATCH_REQUEST_METHOD_NAME = 'batch_request';` variable was used only for the rate limiting logic which got removed and since it was not used anywhere else I removed the declaration.

As for the `const METRIC_HISTOGRAM_NAME = 'rpc_relay_method_result'`, it was previously a scoped variable inside the constructor but I moved it to a global variable inside the file for better overall visibility. 

Thus, no new changes have been introduced to the metric dashboards.

packages/server/src/koaJsonRpc/index.ts
victor-yanev: Previously it was set to contain `BATCH_REQUEST_METHOD_NAME` for all requests inside the batch and now I have changed it to be the corresponding method name of each request that is being executed at that moment.

packages/server/src/koaJsonRpc/lib/methodConfiguration.ts
victor-yanev: Yes, it avoids implicit casting later and potential issues which might arise from the ambiguity in the type.

packages/server/tests/acceptance/rpc_batch3.spec.ts
victor-yanev: Tbh, I don't really have a strong opinion here. 

I slightly like this approach more because we can use the declared version for the `@hashgraph/json-rpc-relay` dependency as a source of truth and we are decoupling the `server` package from the `relay` package. This allows us to completely separate the packages into their own independent modules instead of relying on files that are outside of the packages' scope.

ebadiere (commented):
Please resolve the merge conflict.

packages/server/src/koaJsonRpc/lib/utils.ts
victor-yanev: UPDATE: Adding this change breaks some of the integration and acceptance tests because they rely on modifying process env variables dynamically to verify the behaviour.

georgi-l95 (dismissed):

packages/server/src/koaJsonRpc/index.ts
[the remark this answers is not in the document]
Nana-EC: Thanks

packages/server/src/koaJsonRpc/index.ts
Nana-EC: Hmm, isn't there a single ctx for the batch method?
Will setting this as part of a map mean it's updated multiple times and the last set becomes its final value?

packages/server/src/koaJsonRpc/lib/methodConfiguration.ts
Nana-EC: Thanks

packages/server/src/koaJsonRpc/lib/utils.ts
Nana-EC: 🤦🏾‍♂️ that's unfortunate.
Can you open a ticket for us to come back to this and resolve it.
Seems very inefficient and test code shouldn't bloc this.
We might need to reassess that approach then.

packages/server/tests/acceptance/rpc_batch3.spec.ts
Nana-EC: I agree.
I like the decoupled approach better.
Can you open a ticket for this also to address later?

packages/server/src/koaJsonRpc/lib/utils.ts
victor-yanev: We only call these methods when constructing the `KoaJsonRpc` object and it is instantiated only once in the beginning of `server.ts`, so this shouldn't have any impact on performance currently. 

I agree it will be a nice improvement, though, to have these env variables captured in const values and then reusing those, I will open an issue for this.

packages/server/tests/acceptance/rpc_batch3.spec.ts
victor-yanev: Done, https://github.com/hashgraph/hedera-json-rpc-relay/issues/2685

packages/server/src/koaJsonRpc/lib/utils.ts
victor-yanev: Done, https://github.com/hashgraph/hedera-json-rpc-relay/issues/2686.

Nana-EC (approved):

PR #2611, opened 2024-06-18
fix: reworked gas comparison logic

### LINKED ISSUE #2610: baseFeePerGas is compared to a hard-coded value in batch 1 acceptance test
### Description

baseFeePerGas is compared to a hard-coded value in batch 1 acceptance test, but different environments have different gas price values. 

### Steps to reproduce

Run acceptancetest:api_batch1 gainst testnet

### Additional context

_No response_

### Hedera network

testnet

### Version

latest

### Operating system

None

quiet-node:
**Description**:
This PR reworks gas comparison logic. Originally, gas price was compared to a hard-coded value, 71 HBARs, which is only true if the environment is local-net or previewnet. The new logic retrieves the expected gas value from the Relay for the comparison so it would be correct in all environments.

**Related issue(s)**:

Fixes #2610 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance/rpc_batch1.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch1.spec.ts
@@ -56,6 +56,7 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
   let mirrorContractDetails;
   let requestId: string;
   let account2Address: string;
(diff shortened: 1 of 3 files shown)


ebadiere (approved):
LG.

PR #2616, opened 2024-06-21
fix: fixed unexpected returns for eth_getLog requests with zero address

### LINKED ISSUE #2614: Requesting logs with ZERO address returns anonymous Events
### Description

When sending a log request with ZERO address in the query, 
- it returns the anonymous logs with no topics
- It returns other logs which is not of zero address
We essentially expect no response to be received for the request pasted below
 

### Steps to reproduce

request 
``` curl --location --request POST 'https://testnet.hashio.io/api' --header 'Content-Type: application/json' --data-raw '{
    "jsonrpc": "2.0",
    "method": "eth_getLogs",
    "params": [
        {
         "fromBlock": "0x3954B2",
            "toBlock": "0x395899", "address" : "0x0000000000000000000000000000000000000000"
        }
    ],
    "id": 12
}'```

One of the blocks for this response
``` {
            "address": "0xdf903bff4404952a8d5f5585d1de2787f6a40e62",
            "blockHash": "0x9ef77e7c7838e4e0499fcb37d24494352d273c6c02765cd330525780f7a80d67",
            "blockNumber": "0x3954b2",
            "data": "0x0000000000000000000000000000000000000000000000000000000000000001",
            "logIndex": "0x0",
            "removed": false,
            "topics": [], ---> Anonymour
            "transactionHash": "0x6bc09d56768e6f9d9c999833cc454c3da0e810be04c251958125314f04514817",
            "transactionIndex": "0x1"
        },
        ```
        
And other logs which were not emitted by Zero Address

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.50.0

### Operating system

None

### PR DESCRIPTION
**Pull request by quiet-node**
**Description**:
This PR addresses the issue of unexpected records being returned from `eth_getLog` requests with zero addresses. The update ensures that if the `address` parameter equals `ZeroAddress`, the call to the mirror node is skipped, and an empty array is returned immediately.

**Related issue(s)**:

Fixes #2614

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/tests/lib/eth/eth_getLogs.spec.ts
+++ b/packages/relay/tests/lib/eth/eth_getLogs.spec.ts
@@ -56,6 +56,7 @@ import {
   DEFAULT_NULL_LOG_TOPICS,
   NOT_FOUND_RES,
 } from './eth-config';
+import { ethers } from 'ethers';
 import { generateEthTestEnv } from './eth-helpers';
 
 dotenv.config({ path: path.resolve(__dirname, '../../test.env') });
@@ -578,8 +579,16 @@ describe('@ethGetLogs using MirrorNode', async function () {
 
     const result = await ethImpl.getLogs(null, '0x5', '0x10', null, DEFAULT_LOG_TOPICS);
 
-    expect(result).to.exist;
     expectLogData1(result[0]);
     expectLogData2(result[1]);
+  });
+
+  it('Should return empty log if address = ZeroAddress', async () => {
+    restMock.onGet(BLOCKS_LIMIT_ORDER_URL).reply(200, { blocks: [latestBlock] });
+    restMock.onGet('blocks/0').reply(200, DEFAULT_BLOCK);
+    restMock.onGet('blocks/latest').reply(200, DEFAULT_BLOCK);
+    const result = await ethImpl.getLogs(null, '0x0', 'latest', ethers.ZeroAddress, DEFAULT_LOG_TOPICS);
+    expect(result.length).to.eq(0);
+    expect(result).to.deep.equal([]);
   });
 });
--- a/packages/relay/tests/lib/mirrorNodeClient.spec.ts
+++ b/packages/relay/tests/lib/mirrorNodeClient.spec.ts
@@ -31,8 +31,9 @@ import { getRequestId, mockData, random20BytesAddress } from './../helpers';
 const registry = new Registry();
 
 import pino from 'pino';
-import { SDKClientError } from '../../src/lib/errors/SDKClientError';
+import { ethers } from 'ethers';
 import { predefined } from '../../src/lib/errors/JsonRpcError';
+import { SDKClientError } from '../../src/lib/errors/SDKClientError';
 import { CacheService } from '../../src/lib/services/cacheService/cacheService';
 const logger = pino();
 const noTransactions = '?transactions=false';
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: packages/relay/tests/lib/mirrorNodeClient.spec.ts [hunk 1]
```diff
@@ -732,6 +733,11 @@ describe('MirrorNodeClient', async function () {
     expect(firstResult.contract_id).equal(log.contract_id);
     expect(firstResult.index).equal(log.index);
   });
+  it('`getContractResultsLogsByAddress` with ZeroAddress ', async () => {
+    const results = await mirrorNodeInstance.getContractResultsLogsByAddress(ethers.ZeroAddress);
+    expect(results).to.exist;
```
ebadiere, line 738: Nit.  This would be more precise and readable:
```
expect(results).to.deep.equal([]);
```

packages/relay/tests/lib/eth/eth_getLogs.spec.ts
```diff
@@ -582,4 +583,13 @@ describe('@ethGetLogs using MirrorNode', async function () {
     expectLogData1(result[0]);
     expectLogData2(result[1]);
   });
+
+  it('Should return empty log if address = ZeroAddress', async () => {
+    restMock.onGet(BLOCKS_LIMIT_ORDER_URL).reply(200, { blocks: [latestBlock] });
+    restMock.onGet('blocks/0').reply(200, DEFAULT_BLOCK);
+    restMock.onGet('blocks/latest').reply(200, DEFAULT_BLOCK);
+    const result = await ethImpl.getLogs(null, '0x0', 'latest', ethers.ZeroAddress, DEFAULT_LOG_TOPICS);
+    expect(result).to.exist;
```
ebadiere, line 592: Nit. This would be more precise and readable:

`expect(results).to.deep.equal([]);`

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -431,6 +431,41 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+            {
+              fromBlock: numberTo0x(blocksBehindLatest),
+              toBlock: 'latest',
+              address: [ethers.ZeroAddress, contractAddress2],
+            },
+          ],
+          requestId,
+        );
+        expect(logs.length).to.eq(1);
+      });
```
ebadiere, line 468: Instead of duplicating code, would it not make more sense to extract the relay call into its own function and call it with params?
```
async function getLogs(params: {
  fromBlock: string;
  toBlock: string;
  address: string[];
  requestId: string;
}): Promise<any[]> {
  return await relay.call(
    RelayCalls.ETH_ENDPOINTS.ETH_GET_LOGS,
    [
      {
        fromBlock: params.fromBlock,
        toBlock: params.toBlock,
        address: params.address,
      },
    ],
    params.requestId
  );
}
```

```
const ethLogs = await getLogs({
  fromBlock: '0x0',
  toBlock: 'latest',
  address: [ethers.ZeroAddress],
  requestId,
});

const nonZeroLogs = await getLogs({
  fromBlock: numberTo0x(blocksBehindLatest),
  toBlock: 'latest',
  address: [ethers.ZeroAddress, contractAddress2],
  requestId,
});
```

quiet-node: Great idea! But currently all the test units in eth_getLogs suite are using .call() method from the relay, and adding a shared method will result in code changes to the rest of other test units. Extracting the code to a shared component to avoid code duplication is always a good practice, but I believe it would be out of scope for this PR.

ebadiere (approved):
LG.

PR #2621, opened 2024-06-25
fix: Upgrade to node 20.

### LINKED ISSUE #2224: Upgrade to Node 20 LTS from 18
### Problem

We are currenly using Node 18 but LTS version is 20. so we need to upgrade to reap performance and memory gains, as well as better support and maintenance.

### Solution

Update to Node20 everywhere is needed.

use as a reference the previous upgrade from 16 to 18.
https://github.com/hashgraph/hedera-json-rpc-relay/pull/803/files

### Alternatives

_No response_

ebadiere:
fix: Upgrade Node version to 20

**Related issue(s)**:

Fixes #2224 

**Notes for reviewer**:


quiet-node (approved):
LGTM

Issue #2631, opened 2024-06-27
Add rateLimiter to the create file in the sdkClient


ebadiere:
### Description

A large percentage of the relay's HBar is spent on file appends.  The HBar rate limiter needs to be added to the `createFile` method in the `sdkClient`.

### Steps to reproduce

Look at the relay's dashboard, under "Usage - HBar Accounts and Burn Rates".  Notice the `fileAppendTransaction` is over 80% of the "HBar Spent by Type".

### Hedera network

mainnet

### Version

v0.50.0

### Operating system

Linux

PR #2632, opened 2024-06-27
fix: Added rateLimiter to fileCreate.

ebadiere:
Adds HBar rate limiter to the `createFile` method

**Related issue(s)**: #2631


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -257,7 +257,7 @@ export class SDKClient {
       fileId = await this.createFile(
         ethereumTransactionData.callData,
         this.clientMain,
-        requestId,
+        requestId ?? 'create-file-from-submit-ethereum-transaction',
```
Nana-EC, line 260: What's the goal of this?

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -643,60 +643,111 @@ export class SDKClient {
...
+          await this.calculateFileAppendTxTotalTinybarsCost(fileAppendTx),
+          0,
+          callerName,
+          interactingEntity,
+        );
+      }
+
+      // Ensure that the calldata file is not empty
+      if (fileId) {
+        const fileSize = await (await new FileInfoQuery().setFileId(fileId).execute(client)).size;
```
Nana-EC, line 700: Q: does this have a cost associated with it?
If so can you open a ticket so we come back and see if this is needed still and more so also capture its cost in the metric and rate limits

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -643,60 +643,111 @@ export class SDKClient {
...
+            SDKClient.transactionMode,
+            fileCreateTx.constructor.name,
+            sdkClientError.status,
+            transactionFee.toTinybars().toNumber(),
+            transactionRecord?.contractFunctionResult?.gasUsed,
+            callerName,
+            interactingEntity,
+          );
+
+          this.hbarLimiter.addExpense(transactionFee.toTinybars().toNumber(), currentDateNow);
```
Nana-EC, line 731: This only adds the file create transaction cost.
What about the fileAppend transaction(s)?

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -643,60 +643,111 @@ export class SDKClient {
...
+          Status.Success,
+          await this.calculateFileAppendTxTotalTinybarsCost(fileAppendTx),
+          0,
+          callerName,
+          interactingEntity,
+        );
+      }
+
+      // Ensure that the calldata file is not empty
+      if (fileId) {
```
Nana-EC, line 699: Q: Where's the actual logic to add the costs of the FileCreate and FileAppend to the hbar rate limiter?

Nana-EC (changes requested):
Let's confirm we're capturing all the FileCreate/FileAppend expenses

packages/relay/src/lib/clients/sdkClient.ts
ebadiere: The optional parameter on the requestId causes `Argument of type 'string | undefined' is not assignable to parameter of type 'string'.`. I can remove the optional all the way to the `sendRawTransaction`

Nana-EC: My concern is the choice of string.
Doesn't seem consistent with anything we've done.
Either the requestId is provided by a user of one is generated higher up if not present.

Why do we need this workaround only here?

packages/relay/src/lib/clients/sdkClient.ts
ebadiere: Updated.

ebadiere:
Replaced by #2634

PR #2647, opened 2024-07-01
feat: allowed eth_call to accept null/empty tx.to to simulate deploying smart contracts

### LINKED ISSUE #2334: `eth_call` errors out when `transaction.to` is set to null
### Description

Currently, `eth_call` will reject requests that have `transaction.to` set to null. However, it should accept nullable `transaction.to` in the cases of contract deployment.

### Steps to reproduce

1. Run below cURL request
```bash
curl --location 'https://testnet.hashio.io/api' \
--data '{
    "id": "test_id",
    "jsonrpc": "2.0",
    "method": "eth_call",
        "params": [{
            "from":"0x05fba803be258049a27b820088bab1cad2058871", 
            "to": null,
            "data":"0x60806040"}, "latest"]
}'
```
2. Observe the response to see the error like below
```bash
{"error":{"code":-32602,"name":"Invalid parameter","message":"[Request ID: b2a2c4cf-1ce4-4a56-bbc3-c25e51c51c53] Invalid parameter 'to' for TransactionObject: Expected 0x prefixed string representing the address (20 bytes), value: null"},"jsonrpc":"2.0","id":"test_id"}
```

### Additional context

_No response_

### Hedera network

### Version

latest

### Operating system

None

quiet-node:
**Description**:
Originally, the validation in the Relay object would stop the request if its `calldata.to` field was null or empty because the mirror-node didn't allow a nullable `to` field. However, after [this PR](https://github.com/hashgraph/hedera-mirror-node/pull/8100), the mirror-node now allows a nullable `to` field. This update changes the validation so the Relay can accept requests with a `to` field that is null or empty.

**Related issue(s)**:

Fixes #2334

**Notes for reviewer**:

**Checklist**


ebadiere (dismissed):
LG.

packages/relay/src/lib/eth.ts
```diff
@@ -1828,8 +1827,8 @@ export class EthImpl implements Eth {
    * @param requestIdPrefix
    */
   async performCallChecks(call: any): Promise<void> {
-    // The "to" address must always be 42 chars.
-    if (!call.to || call.to.length != 42) {
+    // after this mirror-node PR https://github.com/hashgraph/hedera-mirror-node/pull/8100, call.to is allowed to be empty or null
+    if (call.to !== null && !isValidEthereumAddress(call.to)) {
```
Nana-EC, line 1831: Is the first check necessary.
`isValidEthereumAddress()` should be able to handle `null` or `undefined`. Check if there are tests for this and if not we should add them to confirm.
```suggestion
    if (!isValidEthereumAddress(call.to)) {
```

Nana-EC (commented):
Suggested improvement

packages/relay/tests/lib/eth/eth_call.spec.ts
```diff
@@ -117,21 +118,6 @@ describe('@ethCall Eth Call spec', async function () {
       process.env.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE = 'true';
     });
 
-    it('eth_call with missing `to` field', async function () {
```
quiet-node, line 120: Since now `to` field is allowed to be null/undefined, this test case is no longer relavant. Instead, this [test case here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2647/files#diff-5b9587c533d686b0386eb8c09b87a02ab40ec1b9581f0a471f03049465cd7698R746) cover the new support.

packages/relay/src/lib/eth.ts
quiet-node: I understand your concern. This situation requires both checks to function correctly. Given that `call.to` can be null or undefined, the `if statement` should not throw an error if `call.to` is null or undefined. 

The function `isValidEthereumAddress(call.to)` returns `false` when `call.to` is null or undefined. Therefore, `!isValidEthereumAddress(call.to)` evaluates to `true`, making the `if statement` valid and potentially causing an error, which is incorrect.

With both checks in place:
- If `call.to` is null or undefined, the error is skipped.
- If `call.to` is not null or undefined, its validity as an Ethereum address is checked.
  - If `call.to` is a valid Ethereum address, the process continues.
  - If `call.to` is not a valid Ethereum address, an error is thrown.

Nana-EC (approved):

acuarica (approved):
lgtm

arianejasuwienas:
@acuarica 
We previously encountered issues with Smart Contract deployment, which were related to the problem described here:

https://github.com/hashgraph/hedera-json-rpc-relay/pull/2647#issue-2384754612

However, we were able to resolve it on our end, so the issue no longer exists. That's why I'm not sure if its worth mentioning (?)

PR #2649, opened 2024-07-02
fix: Improve compatibility with standard ethereum tooling

### LINKED ISSUE #2108: Research: Improve compatility with standard Ethereum tooling
### Problem

Some formatting on the relay's side can be enhanced to improve compatibility with ethereum libaries (for example go-ethereum/client).
Currently the relay returns null for certain fields if there's no value while the ethereum spec defines them as `0x0`. One such example is `maxPriorityFeePerGas` There are probably others as well.

### Solution

- Review values returned by the relay and align them with the ethereum spec

### Alternatives

_No response_

konstantinabl:
**Description**:

Currently, our validation process only checks the presence of fields returned by our API, without verifying the format of these fields. This can lead to inconsistencies and potential errors.

Changes introduced:
* Adds separate CI workflow for API conformity
* Pulls the openrpc.json from [execution-apis repo](https://github.com/ethereum/execution-apis/tree/main) in order to validate our response against it
* 

**Related issue(s)**:

Fixes #2108

### DIFF
--- a/.github/workflows/acceptance-public.yml
+++ b/.github/workflows/acceptance-public.yml
@@ -47,15 +47,6 @@ jobs:
       operator_id: ${{ inputs.operator_id }}
       operator_key: ${{ inputs.operator_key }}
 
-  rpc_api_schema_conformity:
-    name: API Conformity
-    uses: ./.github/workflows/acceptance-workflow.yml
-    with:
-      testfilter: api_conformity
-      envfile: ${{ inputs.network }}Acceptance.env
-      operator_id: ${{ inputs.operator_id }}
-      operator_key: ${{ inputs.operator_key }}
-
   erc20:
     name: ERC20
     uses: ./.github/workflows/acceptance-workflow.yml
--- a/.github/workflows/acceptance-public.yml
+++ b/.github/workflows/acceptance-public.yml
@@ -138,7 +138,6 @@ jobs:
       - api_batch_1
       - api_batch_2
       - api_batch_3
-      - rpc_api_schema_conformity
       - erc20
       - ratelimiter
       - tokencreate
--- a/.github/workflows/conformity-workflow.yml
+++ b/.github/workflows/conformity-workflow.yml
@@ -1,12 +1,11 @@
 name: Conformity Test Workflow
 
 on:
-  push:
-    branches:
-      - main
   pull_request:
-    branches:
-      - main
+    branches: [main, release/**]
+  push:
+    branches: [main, release/**]
+    tags: [v*]
 
 jobs:
   clone-and-build-execution-apis:
@@ -92,9 +91,6 @@ jobs:
       
       - name: Rename file
         run: mv openrpc.json openrpc_exec_apis.json
-
-      - name: List files in dir
-        run: ls -al
 
       - name: Run conformity tests
         run: npm run acceptancetest:rpc_api_schema_conformity
--- a/.github/workflows/conformity-workflow.yml
+++ b/.github/workflows/conformity-workflow.yml
@@ -1,0 +1,66 @@
+name: Conformity Test Workflow
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  clone-and-build-execution-apis:
+    runs-on: [self-hosted, Linux, medium, ephemeral]
+
+    steps:
+      - name: Checkout execution-apis repo
+        uses: actions/checkout@v2
+        with:
+          repository: 'ethereum/execution-apis'
+          ref: 'main'
+          path: 'execution-apis'
+
+      - name: Use Node.js TLS 18
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+        with:
+          node-version: 18
+
+      - name: Install dependencies
+        run: npm install
+        working-directory: ./execution-apis
+
+      - name: Build project
+        run: npm run build
+        working-directory: ./execution-apis
+
+      - name: Copy openrpc.json to current repo
+        run: cp ./execution-apis/openrpc.json ${{ github.workspace }}
+
+      - name: Commit openrpc.json
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add openrpc.json
+          git commit -m "Update openrpc.json" || echo "No changes to commit"
+          git push
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  build-and-test:
+    runs-on: [self-hosted, Linux, medium, ephemeral]
+    needs: clone-and-build-execution-apis
+  
+    steps:
+      - name: Checkout current repo
+        uses: actions/checkout@v2
+
+      - name: Use Node.js TLS 18
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+        with:
+          node-version: 18
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Run conformity tests
+        run: npm run acceptancetest:rpc_api_schema_conformity
--- a/.github/workflows/conformity-workflow.yml
+++ b/.github/workflows/conformity-workflow.yml
@@ -88,7 +88,9 @@ jobs:
         uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e # v4.1.7
         with:
           name: openrpc
-          path: ./openrpc_exec_apis.json
+      
+      - name: Rename file
+        run: mv openrpc.json openrpc_exec_apis.json
 
       - name: List files in dir
         run: ls -al
--- a/.github/workflows/conformity-workflow.yml
+++ b/.github/workflows/conformity-workflow.yml
@@ -14,8 +14,9 @@ jobs:
 
     steps:
       - name: Checkout execution-apis repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
         with:
+          ref: 7907424db935b93c2fe6a3c0faab943adebe8557
           repository: 'ethereum/execution-apis'
           path: 'execution-apis'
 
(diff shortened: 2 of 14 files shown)


.github/workflows/conformity-workflow.yml
```diff
@@ -0,0 +1,99 @@
...
+    branches:
+      - main
+
+jobs:
+  clone-and-build-execution-apis:
+    runs-on: [self-hosted, Linux, medium, ephemeral]
+
+    steps:
+      - name: Checkout execution-apis repo
+        uses: actions/checkout@v2
```
jeromy-cannon, line 17: pin to a commit

.github/workflows/conformity-workflow.yml
```diff
@@ -0,0 +1,99 @@
...
+          name: openrpc
+          path: ./execution-apis/openrpc.json
+
+  build-and-test:
+    runs-on: [self-hosted, Linux, medium, ephemeral]
+    needs: clone-and-build-execution-apis
+  
+    steps:
+      - name: Checkout relay repo
+        uses: actions/checkout@v2
```
jeromy-cannon, line 47: pin to a commit

georgi-l95 (commented):
LG, after this we can write some custom tests for the schema. aside from Jeromy comments, i think it looks okay. @Nana-EC has to modify the checks when this is approved in order to be merged.

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -204,6 +208,34 @@ function checkResponseFormat(fileName, actualReponse, expectedResponse) {
   return missingKeys;
 }
 
+function findSchema(file) {
+  const execApisOpenRpcData = require('../../../../openrpc_exec_apis.json');
```
victor-yanev, line 212: QQ: Why do we need to import this json file every time this function is called? Can't we just import it only once at the top and then reuse it here?

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -204,6 +208,34 @@ function checkResponseFormat(fileName, actualReponse, expectedResponse) {
...
+  expect(validate.errors).to.be.null;
+
+  return valid;
+}
+
+function validateBlockNumber() {
+  const schema = {
+    title: 'hex encoded unsigned integer',
+    type: 'string',
+    pattern: '^0x([1-9a-f]+[0-9a-f]*|0)$',
```
victor-yanev, line 231: This regex will fail for:
- hex strings with leading zeros
- the string `"0x"`

Is this the expected behaviour?

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -204,6 +208,34 @@ function checkResponseFormat(fileName, actualReponse, expectedResponse) {
...
+  };
+  const validate = ajv.compile(schema);
+  const valid = validate('d76');
+  console.log(valid);
+  console.log(validate.errors);
```
victor-yanev, line 236: Reminder to remove this

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -204,6 +208,34 @@ function checkResponseFormat(fileName, actualReponse, expectedResponse) {
...
+}
+
+function validateBlockNumber() {
+  const schema = {
+    title: 'hex encoded unsigned integer',
+    type: 'string',
+    pattern: '^0x([1-9a-f]+[0-9a-f]*|0)$',
+  };
+  const validate = ajv.compile(schema);
+  const valid = validate('d76');
```
victor-yanev, line 234: What are we validating here exactly? It seems like this is a hardcoded string.

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -288,12 +323,11 @@ describe('@api-conformity Ethereum execution apis tests', function () {
   });
   //Reading the directories within the ethereum execution api repo
   let directories = fs.readdirSync(directoryPath);
-
   const relaySupportedMethodNames = openRpcData.methods.map((method) => method.name);
 
   //Filtering in order to use only the tests for methods we support in our relay
   directories = directories.filter((directory) => relaySupportedMethodNames.includes(directory));
-
+  validateBlockNumber();
```
victor-yanev, line 330: QQ: Is this supposed to be here?

victor-yanev (changes requested):

packages/server/tests/acceptance/conformityTests.spec.ts
konstantinabl: Sorry, this is leftover from testing, I will remove it

packages/server/tests/acceptance/conformityTests.spec.ts
konstantinabl: Leftovers from testing that I will remove, thanks for pointing out

packages/server/tests/acceptance/conformityTests.spec.ts
konstantinabl: I did it in this way, because mocha loads all files and their imports even when we run the mocha command with the -g option (for a specific file). Therefore, it tries to load this file in the other workflows and the file still does not exist, it is created only in the conformity workflow. I moved it in the beginning of each test in the "it", so at least it loads for every directory and not every file. Another option is to somehow exclude this test file when other workflos are executed.

victor-yanev: Okay, I see now. Yes, for every directory would be slightly better. Let's leave it like that.

.github/workflows/conformity-workflow.yml
konstantinabl: Done

.github/workflows/acceptance-public.yml
```diff
@@ -47,15 +47,6 @@ jobs:
       operator_id: ${{ inputs.operator_id }}
       operator_key: ${{ inputs.operator_key }}
 
-  rpc_api_schema_conformity:
```
Nana-EC, line 50: Why remove vs replace?
We don't want to lose the coverage here when running this workflow.

Same note for a few other places you removed it.

.github/workflows/conformity-workflow.yml
```diff
@@ -0,0 +1,100 @@
+name: Conformity Test Workflow
+
+on:
+  push:
+    branches:
+      - main
```
Nana-EC, line 6: This should match the other workflows and should be more than just on main

.github/workflows/conformity-workflow.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Download openrpc.json artifact
+        uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e # v4.1.7
+        with:
+          name: openrpc
+      
+      - name: Rename file
+        run: mv openrpc.json openrpc_exec_apis.json
+
+      - name: List files in dir
```
Nana-EC, line 96: Q: Did you mean to remove this?

Nana-EC (commented):
nice direction. Left some comments

.github/workflows/conformity-workflow.yml
konstantinabl: Yes, thank you, removed

.github/workflows/conformity-workflow.yml
konstantinabl: Changed it so it is as the other workflows

jeromy-cannon (approved):

packages/server/package.json
```diff
@@ -38,7 +38,7 @@
     "axios-retry": "^3.5.1",
     "chai": "^4.3.6",
     "ethers": "^6.7.0",
-    "execution-apis": "git://github.com/ethereum/execution-apis.git#584905270d8ad665718058060267061ecfd79ca5",
+    "execution-apis": "git://github.com/ethereum/execution-apis.git#7907424db935b93c2fe6a3c0faab943adebe8557",
```
quiet-node, line 41: just being curious, where did you get this updated number?

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -5,6 +5,8 @@ const directoryPath = path.resolve(__dirname, '../../../../node_modules/executio
 const axios = require('axios');
```
quiet-node, line 5: This file is missing header license
```
/*-
/*-
 *
 * Hedera JSON RPC Relay
 *
 * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
 * ...
 ```

File: packages/server/tests/acceptance/conformityTests.spec.ts [hunk 13]
```diff
@@ -5,6 +5,8 @@ const directoryPath = path.resolve(__dirname, '../../../../node_modules/executio
...
 const openRpcData = require('../../../../docs/openrpc.json');
 require('dotenv').config();
+import Ajv from 'ajv';
+import addFormats from 'ajv-formats';
```
quiet-node, line 9: my editor says it cannot find this module, ajv-formats.

image

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -5,6 +5,8 @@ const directoryPath = path.resolve(__dirname, '../../../../node_modules/executio
...
 import { signTransaction } from '../../../relay/tests/helpers';
 import { expect } from 'chai';
 let currentBlockHash;
```
quiet-node, line 12: Please clean up this session. I noticed that both `import` and `require` syntaxes are being used for importing dependencies. To conform to ES6 standards, please use `import` instead of `require`. Additionally, let's improve readability by adding some empty lines between the imports section and the variable declaration section.

quiet-node (changes requested):
Great work! Just some suggestions and comments

packages/server/package.json
konstantinabl: From the execution apis repo

packages/server/tests/acceptance/conformityTests.spec.ts
konstantinabl: Honestly same, however this is installed and its in the package.json, not sure why it complains @quiet-node

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -271,15 +316,19 @@ async function processFileContent(file, content) {
...
+  const valid = needError
+    ? checkResponseFormat(response.response.data, content.response)
+    : isResponseValid(schema, response);
+  expect(valid).to.be.true;
 }
 
 describe('@api-conformity Ethereum execution apis tests', function () {
   this.timeout(240 * 1000);
   before(async () => {
+    console.log('Bytecode', bytecode);
```
Ivo-Yankov, line 331: nit: Is this intentional?

Ivo-Yankov (approved):
LGTM

packages/server/tests/acceptance/conformityTests.spec.ts
konstantinabl: no :D will remove

jeromy-cannon (approved):

quiet-node (approved):
LGTM

victor-yanev (approved):

PR #2657, opened 2024-07-02
fix: Added metrics around the HBar rate limiter to better track the

### LINKED ISSUE #2653: More visibility in the HBar rate limiter is needed.
### Problem

The relay is consuming more HBar, and it appears to be around FileAppendTransactions, but we have no idea if the HBar ratelimiter is kicking in or not.  

### Solution

Viewing the balance, the number of times it's called, and when it kicks in would allows us tro get a better idea of the relay's consumption of HBar.  Add metrics to expose this and then a panel in the dashboard to view it.

### Alternatives

None

ebadiere:
Added more metrics around the HBar rate limiter.

**Related issue(s)**:

Fixes #2653

### DIFF
--- a/packages/relay/src/lib/hbarlimiter/index.ts
+++ b/packages/relay/src/lib/hbarlimiter/index.ts
@@ -19,7 +19,7 @@
  */
 
 import { Logger } from 'pino';
-import { Registry, Counter } from 'prom-client';
+import { Registry, Counter, Gauge } from 'prom-client';
 
 export default class HbarLimit {
   private enabled: boolean = false;
@@ -29,6 +29,8 @@ export default class HbarLimit {
   private reset: number;
   private logger: Logger;
   private hbarLimitCounter: Counter;
+  private hbarLimitRemainingGauge: Gauge;
+  private hbarShouldLimitCounter: Counter;
   private readonly register: Registry;
 
   constructor(logger: Logger, currentDateNow: number, total: number, duration: number, register: Registry) {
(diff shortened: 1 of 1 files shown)


packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -52,6 +54,26 @@ export default class HbarLimit {
...
+    const rateLimiterRemainingGaugeName = 'rpc_relay_hbar_rate_remaining';
+    register.removeSingleMetric(rateLimiterRemainingGaugeName);
+    this.hbarLimitRemainingGauge = new Gauge({
+      name: rateLimiterRemainingGaugeName,
+      help: 'Relay Hbar limit total',
+      registers: [register],
+    });
+    this.hbarLimitRemainingGauge.set(this.remainingBudget);
+
+    const shouldLimitCounterName = 'rpc_relay_hbar_should_limit';
```
Nana-EC, line 68: I don't think this metric is as valuable but if you feel strongly you can split it out into a different ticket

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -52,6 +54,26 @@ export default class HbarLimit {
       registers: [register],
       labelNames: ['mode', 'methodName'],
     });
+    this.hbarLimitCounter.inc(0);
+
+    const rateLimiterRemainingGaugeName = 'rpc_relay_hbar_rate_remaining';
+    register.removeSingleMetric(rateLimiterRemainingGaugeName);
+    this.hbarLimitRemainingGauge = new Gauge({
+      name: rateLimiterRemainingGaugeName,
+      help: 'Relay Hbar limit total',
```
Nana-EC, line 63:
```suggestion
      help: 'Relay Hbar rate limit remaining budget',
```

Nana-EC (changes requested):
Nice idea.
Let's just do the remaining budget for now

packages/relay/src/lib/hbarlimiter/index.ts
ebadiere: Done.

Nana-EC (approved):

PR #2668, opened 2024-07-03
fix: Adds usage of internal cache for getFilterChanges

### LINKED ISSUE #2646: eth_newFilter / eth_getFilterChanges not working
### Description

There was a breaking change introduced in [this PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1699)

The problem is that whether shared cache is configured or not we try to write filter id to the shared cache:
https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/services/ethService/ethFilterService/index.ts#L94

But then we read the key from whatever cache is enabled:
https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/services/ethService/ethFilterService/index.ts#L212

As result if there's no shared cache set, this functionality doesn't work.

### Steps to reproduce

1. Set up Hedera JSON RPC relay so that Filters are enabled but there's no Redis cache set up
2. Create filter:
```bash
curl -H "Content-Type: application/json" -X POST -d'{"method":"eth_newFilter","params":[{"topics":["0xd78ad95fa46c994b6551d0da85fc275fe613ce37657fb8d5e3d130840159d822"],"address":["0x1e375f8947dffee57c7713f3176b01858bfba104"]}],"id":2,"jsonrpc":"2.0"}'  http://your-hedera-relay
```
3. Try to request filter changes (while specifying the filter ID obtained on the previous step):
curl -H "Content-Type: application/json" -X POST -d'{"method":"eth_getFilterChanges","params":["<your filter id>"],"id":4,"jsonrpc":"2.0"}' http://your-hedera-relay
4. Receive an error:
```
{"error":{"code":-32001,"name":"Filter not found","message":"[Request ID: 063e3a42-c3ee-4579-a9d5-cac84cd07a42] Filter not found"},"jsonrpc":"2.0","id":4}
```

### Additional context

_No response_

### Hedera network

mainnet, testnet

### Version

v0.50.1

### Operating system

Linux

konstantinabl:
**Description**:

Currently, in getFilterChanges we get the filter id from the sharedCache. However, this won't work if the sharedCache option is disabled.

The PR modifies the filterService, so we check if the sharedCache is actually enabled.

**Related issue(s)**:

Fixes #2646

### DIFF
--- a/packages/relay/src/lib/services/cacheService/cacheService.ts
+++ b/packages/relay/src/lib/services/cacheService/cacheService.ts
@@ -124,7 +124,7 @@ export class CacheService {
    * @private
    * @returns {boolean} Returns true if Redis caching is enabled, otherwise false.
    */
-  private isRedisEnabled(): boolean {
+  public isRedisEnabled(): boolean {
     const redisEnabled = process.env.REDIS_ENABLED && process.env.REDIS_ENABLED === 'true';
     const redisUrlValid = process.env.REDIS_URL && process.env.REDIS_URL !== '';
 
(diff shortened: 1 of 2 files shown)


ebadiere (approved):
LG

PR #2682, opened 2024-07-08
fix: Adds null check for param in validation

### LINKED ISSUE #2376: "Error raised during getCode for address null" is not returned in the response to the client and can cause confusion.
### Description

"Unknown error invoking RPC" is returned to the client when in the logs we see "Error raised during getCode for address null".     The "address is null" message should be returned to the client to give the user an idea of what is going on.

### Steps to reproduce

1. Run `eth_getCode` with a null address value.

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.45.2

### Operating system

Linux

konstantinabl:
**Description**:
Currently, when we pass null as a param for getCode or any other endpoint, we do not return a descriptive response to the user. Thus, this PR adds a check in the validator, that throws an error when the param passed is null.

**Files Changed** :

1. `src/validator/utils.ts` - Added a null check for the parameters
2. `rpc_batch2.spec.ts` - Fixed few failing tests

**Type of Change**:
  - [x] Bug fix
  - [ ] New feature
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Refactoring

**Related issue(s)**:
Fixes #2376

### DIFF
--- a/packages/server/src/validator/utils.ts
+++ b/packages/server/src/validator/utils.ts
@@ -14,7 +14,11 @@ export function validateParam(index: number | string, param: any, validation: an
     throw predefined.MISSING_REQUIRED_PARAMETER(index);
   }
 
-  if (param != null && Array.isArray(paramType)) {
+  if (param === null) {
+    throw predefined.INVALID_PARAMETER(index, `The value passed is not valid: ${param}.`);
+  }
+
+  if (Array.isArray(paramType)) {
     const results: any[] = [];
     for (const type of paramType) {
       const validator = Validator.TYPES[type];
@@ -27,7 +31,7 @@ export function validateParam(index: number | string, param: any, validation: an
     }
   }
 
-  if (param != null && !Array.isArray(paramType)) {
+  if (!Array.isArray(paramType)) {
     const result = isArray ? paramType.test(index, param, validation.type[1]) : paramType.test(param);
     if (result === false) {
       throw predefined.INVALID_PARAMETER(index, `${paramType.error}, value: ${param}`);
--- a/packages/server/src/validator/utils.ts
+++ b/packages/server/src/validator/utils.ts
@@ -12,13 +12,17 @@ export function validateParam(index: number | string, param: any, validation: an
 
   if (requiredIsMissing(param, validation.required)) {
     throw predefined.MISSING_REQUIRED_PARAMETER(index);
+  } else if (!validation.required && param === undefined) {
+    //if parameter is undefined and not required, no need to validate
+    //e.g estimateGas method, blockNumber is not required
+    return;
   }
 
   if (param === null) {
     throw predefined.INVALID_PARAMETER(index, `The value passed is not valid: ${param}.`);
   }
 
-  if (param !== undefined && Array.isArray(paramType)) {
+  if (Array.isArray(paramType)) {
     const results: any[] = [];
     for (const type of paramType) {
       const validator = Validator.TYPES[type];
@@ -31,7 +35,7 @@ export function validateParam(index: number | string, param: any, validation: an
     }
   }
 
-  if (param !== undefined && !Array.isArray(paramType)) {
+  if (!Array.isArray(paramType)) {
     const result = isArray ? paramType.test(index, param, validation.type[1]) : paramType.test(param);
     if (result === false) {
       throw predefined.INVALID_PARAMETER(index, `${paramType.error}, value: ${param}`);
--- a/packages/server/src/validator/utils.ts
+++ b/packages/server/src/validator/utils.ts
@@ -13,12 +13,11 @@ export function validateParam(index: number | string, param: any, validation: an
   if (requiredIsMissing(param, validation.required)) {
     throw predefined.MISSING_REQUIRED_PARAMETER(index);
   }
-
   if (param === null) {
     throw predefined.INVALID_PARAMETER(index, `The value passed is not valid: ${param}.`);
   }
 
-  if (Array.isArray(paramType)) {
+  if (param != null && Array.isArray(paramType)) {
     const results: any[] = [];
     for (const type of paramType) {
       const validator = Validator.TYPES[type];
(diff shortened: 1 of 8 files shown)


packages/server/src/validator/utils.ts
```diff
@@ -14,7 +14,11 @@ export function validateParam(index: number | string, param: any, validation: an
     throw predefined.MISSING_REQUIRED_PARAMETER(index);
   }
 
-  if (param != null && Array.isArray(paramType)) {
+  if (param === null) {
+    throw predefined.INVALID_PARAMETER(index, `The value passed is not valid: ${param}.`);
+  }
+
+  if (param !== undefined && Array.isArray(paramType)) {
```
victor-yanev, line 21: Why are we allowing `undefined` as a parameter, but not `null`?

packages/server/src/validator/utils.ts
```diff
@@ -14,7 +14,11 @@ export function validateParam(index: number | string, param: any, validation: an
     throw predefined.MISSING_REQUIRED_PARAMETER(index);
   }
 
-  if (param != null && Array.isArray(paramType)) {
+  if (param === null) {
+    throw predefined.INVALID_PARAMETER(index, `The value passed is not valid: ${param}.`);
```
georgi-l95, line 18: Hm, I think that according to the official schema, everything different than a string contating 20 byte hex should be rejected with this error. wdyt ?

victor-yanev: Do we want to throw this validation error even for parameters which are not required? It doesn't seem right.

If not, we can change `requiredIsMissing(param, validation.required)` which is already called above to also take `null` values into account:

```diff
export function requiredIsMissing(param: any, required: boolean) {
- return required && param === undefined;
+ return required && (param === undefined || param === null);
}
```

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -976,12 +976,13 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
       const signedTx = await accounts[1].wallet.signTransaction(transaction);
       const transactionHash = await relay.sendRawTransaction(signedTx, requestId);
 
-      const blockNumber = await relay.call(
+      const transactionResult = await relay.call(
```
victor-yanev, line 979: `transactionResult` -> `receipt` or `transactionReceipt`

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -1021,11 +1022,13 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
       const signedTx = await accounts[1].wallet.signTransaction(transaction);
       const transactionHash = await relay.sendRawTransaction(signedTx, requestId);
 
-      const blockHash = await relay.call(
+      const transactionResult = await relay.call(
```
victor-yanev, line 1025: nit: Same here

packages/server/src/validator/utils.ts
konstantinabl: > Hm, I think that according to the official schema, everything different than a string contating 20 byte hex should be rejected with this error. wdyt ?

@georgi-l95 do you mean in general? or for getcode in particular?

packages/server/src/validator/utils.ts
konstantinabl: Null is allowed only for to address in transaction objects and filter objects, which is cerified in the if below, so it makes sense that any other param equal to null should not be allowed. Undefined is allowed for unrequired params, so I've changed the logic a little bit to check it above in an else after the check for requiredIsMissing

packages/server/src/validator/utils.ts
konstantinabl: @georgi-l95  This file is executed for every parameter, thats why I have added param == null here

packages/server/src/validator/utils.ts
```diff
@@ -27,7 +35,7 @@ export function validateParam(index: number | string, param: any, validation: an
     }
   }
 
-  if (param != null && !Array.isArray(paramType)) {
+  if (!Array.isArray(paramType)) {
     const result = isArray ? paramType.test(index, param, validation.type[1]) : paramType.test(param);
```
victor-yanev, line 39: We already know `isArray === false` here
```suggestion
    const result = : paramType.test(param);
```

victor-yanev (approved):
LGTM, just one small nit

packages/server/src/validator/utils.ts
konstantinabl: Fixed it

packages/server/tests/integration/server.spec.ts
```diff
@@ -2144,7 +2144,7 @@ describe('RPC Server', function () {
         }
       });
 
-      it('validates address filter', async function () {
+      it.only('validates address filter', async function () {
```
victor-yanev, line 2147: remove `.only` from here

packages/server/src/validator/utils.ts
```diff
@@ -12,9 +12,17 @@ export function validateParam(index: number | string, param: any, validation: an
 
   if (requiredIsMissing(param, validation.required)) {
     throw predefined.MISSING_REQUIRED_PARAMETER(index);
+  } else if (!validation.required && param === undefined) {
+    //if parameter is undefined and not required, no need to validate
+    //e.g estimateGas method, blockNumber is not required
+    return;
   }
```
victor-yanev, line 19: nit: this will be slightly easier to read:
```suggestion
  if (param === undefined) {
    if (validation.required) {
      throw predefined.MISSING_REQUIRED_PARAMETER(index);
    } else {
      // if parameter is undefined and not required, no need to validate
      // e.g estimateGas method, blockNumber is not required
      return;
    }
  }
```

victor-yanev (approved):
LGTM, I left two small suggestions:

victor-yanev (approved):

packages/server/src/validator/utils.ts
konstantinabl: For some reason when I introduce the new change you suggested the unit tests start failing in the CI

victor-yanev (approved):

Ivo-Yankov (approved):

Issue #2671, opened 2024-07-03
Update the OPERATOR_ID_MAIN in the localAcceptance.env to be one of the local accounts created by the local node.


ebadiere:
### Problem

Testing accounts should be used when running the acceptance tests.

### Solution

Update the localAcceptance.env file.

Nana-EC:
Also ensure that the `OPERATOR_ID_MAIN` can be set as an evm address and doesn't have to be a `0.0.x` format.
We may have done this already but let's confirm

PR #2690, opened 2024-07-09
fix: Updated the account from the 0.0.2 account to an account from th…

ebadiere:
Use an account other than 0.0.2 for running tests.  Removed "Out of Memory" error check in CI.. PR #[2695](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2695) covers memory leaks. 
**Related issue(s)**: #2671


Nana-EC (dismissed):

quiet-node:
The PR title seems to got cut off. Please update it before merging.

packages/server/tests/localAcceptance.env
```diff
@@ -1,6 +1,7 @@
 HEDERA_NETWORK={"127.0.0.1:50211":"0.0.3"}
-OPERATOR_ID_MAIN=0.0.2
-OPERATOR_KEY_MAIN=302e020100300506032b65700422042091132178e72057a1d7528025956fe39b0b847f200ab59b2fdd367017f3087137
+# Account from local node. This is the account that will be used to pay for transactions.
+OPERATOR_ID_MAIN=0.0.1022
+OPERATOR_KEY_MAIN=302e020100300506032b657004220420a608e2130a0a3cb34f86e757303c862bee353d9ab77ba4387ec084f881d420d4
```
quiet-node, line 4: After local-node finishes running, I could find the account 0.0.1022 in the accounts list, and there seems to be only HEX private keys, not DER. So I'm curious about how you got this DER key?

Nana-EC: Good point. 
Also don't we now support the hex format for keys? If so let's use that instead

Ivo-Yankov (approved):
LGTM

packages/server/tests/localAcceptance.env
ebadiere: Done.  Now using ECDSA key in HEX format.

packages/server/tests/clients/servicesClient.ts
```diff
@@ -75,7 +75,7 @@ export default class ServicesClient {
     this.network = network;
 
     if (!network) network = '{}';
-    const opPrivateKey = PrivateKey.fromString(key);
+    const opPrivateKey = PrivateKey.fromStringECDSA(key);
```
quiet-node, line 78: Would this work if the key is ED2? Should we make this dynamically accept any type of key?

ebadiere: We decided to go with HEX ECDSA for testing.

ebadiere: Updated with supporting code for all formats.

packages/server/tests/clients/servicesClient.ts
```diff
@@ -84,6 +84,21 @@ export default class ServicesClient {
     this.client.setOperator(AccountId.fromString(accountId), opPrivateKey);
   }
 
+  createPrivateKeyBasedOnFormat(operatorMainKey: string): PrivateKey {
```
quiet-node, line 87: I believe this was also checked in before in hapuService, right? Could you pull it out and put in a neutral place like a util or something to share the method instead of duplicate it?

quiet-node (dismissed):
LGTM 🚀

quiet-node (approved):
LGTM

PR #2692, opened 2024-07-10
feat: add a tx gasPrice Buffer to reduce INSUFFICIENT_TX_FEE cases

### LINKED ISSUE #2674: Add a tx gasPrice Buffer to reduce INSUFFICIENT_TX_FEE cases
### Problem

Occasionally users observe an `INSUFFICIENT_TX_FEE`.
This may be for insufficient gasLimit supply but occasionally it's due to a change in exchange ate that result in the converted hbar value of the gasLimit not being sufficient.
This is hard to determine as it change infrequently but also due to caching optimizations the relay may miss this.

### Solution

Add a configureable setting `GAS_PRICE_BUFFER` that adds a percentage on top of the calculated network gasPrice.

Note: This may be a band aid option to improve user experience whiles we explore other longer term options.

### Alternatives

_No response_

natanasow:
**Description**:

Occasionally users observe an `INSUFFICIENT_TX_FEE`.
This may be for insufficient gasLimit supply but occasionally it's due to a change in the exchange rate that results in the converted hbar value of the gasLimit not being sufficient.
This is hard to determine as it changes infrequently but also due to caching optimizations the relay may miss this.

**Solution**:
Add a configureable setting `GAS_PRICE_PERCENTAGE_BUFFER` that adds a percentage on top of the calculated network gasPrice.

Note: This may be a band-aid option to improve user experience while exploring other longer-term options.

**Related issue(s)**:

Fixes #2674

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -730,6 +730,19 @@ export class EthImpl implements Eth {
...
       if (!gasPrice) {
         gasPrice = await this.getFeeWeibars(EthImpl.ethGasPrice, requestIdPrefix);
+
+        // converting to tinybar and afterward to weibar again is needed
+        // in order to handle the possibility of an invalid float number being calculated as a gas price
+        // e.g.
+        //   current gas price = 126
+        //   buffer = 10%
+        //   buffered gas price = 126 + 12.6 = 138.6 <--- invalid tinybars
+        gasPrice +=
```
Nana-EC, line 740: move to a util method outside of this class and add some tests to it

Nana-EC (commented):
Looking good.
One recommendation for coverage

packages/relay/src/lib/eth.ts
natanasow: Done.

docs/configuration.md
```diff
@@ -67,6 +67,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
 | `ETH_FEE_HISTORY_FIXED`                     | "true"                   | Flag to set if eth_feeHistory should return a fixed fee for the set of results.                                                                                                                                                                                                                        |
 | `ETH_POPULATE_SYNTHETIC_CONTRACT_RESULTS`   | "true"                   | Flag to set if the relay should populate the contract results for synthetic contracts.                                                                                                                                                                                                                 |
+| `GAS_PRICE_PERCENTAGE_BUFFER`               | "0"                      | The additional buffer that adds a percentage on top of the calculated network gasPrice. This may be an option to fix INSUFFICIENT_TX_FEE error thrown due recently changes in exchange rates.                                                                                                          |
```
quiet-node, line 70: I feel that the part "This may be an option to fix the INSUFFICIENT_TX_FEE error caused by recent changes in exchange rates." shouldn't be included because it doesn't seem relevant in a global context.

Nana-EC: I think it's a fair mention that informs why someone might use it.
Maybe it can be better worded as 
```suggestion
| `GAS_PRICE_PERCENTAGE_BUFFER`               | "0"                      | The additional buffer that adds a percentage on top of the calculated network gasPrice. This may be used by operators to reduce the chances of `INSUFFICIENT_TX_FEE` errors experienced by users caused by minor fluctuations in the exchange rate.                                                                                                          |
```

Nana-EC (approved):
Good change.
None blocking suggestions that can be done in future PRs

1. Move the `addPercentageBufferToGasPrice()` to a different util class (new one if necessary). `formatter.ts` doesn't seem like the right location
2. Adopt the README suggestion

docs/configuration.md
natanasow: Done.

ebadiere (approved):
LG.

docs/configuration.md
quiet-node: Ah with the new version, it sounds more confident!

quiet-node (approved):
LGTM

PR #2695, opened 2024-07-11
chore: Implement memory leak detection in tests

### LINKED ISSUE #2260: Implement a Memory Leak detection Test
### Problem

Sometimes changes to the codebase of the relay, as well as upgrades of libraries that are used within the relay, could provoke a memory leak issue. that is not apparent on the tests until we move the version to production environments.
That makes it harder to identify, since the release contains many changes, all of wich could be the source of the problem.

### Solution

Ideally we need a Test or set of tests that are designed to find and detect memory issues or leaks faster, that we can run on every PR, or at least on every release.

### Alternatives

We have worked on a POC that uses `node-memwatch` to log the differential of the memory heap after each request is made.
see here: https://github.com/hashgraph/hedera-json-rpc-relay/pull/2253

maybe we can extend this solution to perform many requests, simulating load, and then take the `diff` on heap. compare those results with a baseline and raise a flag in case there is something that changes too drastically.

victor-yanev:
**Description**:

This PR introduces a method `captureMemoryLeaks` to detect and report potential memory leaks in test suites, enhancing the reliability and performance of applications by ensuring early detection and resolution of memory leaks. 

It leverages the `V8` engine's garbage collection (GC) profiling capabilities to monitor memory usage before and after test execution, identifying significant increases in `totalHeapSize` that may indicate a memory leak.

1. **Initialization**: The method begins by enabling the garbage collection (GC) through the `setFlagsFromString('--expose_gc')` command. This allows for manual triggering of garbage collection within the test environment.  
2. **Test Setup**: Before each test case runs, `profiler.start()` is called to set up the GC profiler to monitor heap usage and other GC-related statistics during the test execution.  
3. **Test Teardown**: After each test case completes, several critical actions are performed:  
   - Garbage collection is manually triggered using `await gc()` to ensure that all unreferenced objects are cleared from memory. This step aims to capture the most accurate state of the heap after test execution.
   - The GC profiler is then stopped with `profiler.stop()`, and the method collects the GC statistics, including heap sizes before and after the GC.
   - The method calculates the total difference in heap size before and after the test execution. If this difference exceeds **`500KB`** (defined as `HEAP_SIZE_DIFF_MEMORY_LEAK_THRESHOLD`), it indicates a potential memory leak.
   - For tests that potentially leak memory, detailed statistics about the memory usage and the specific areas of the heap that have grown are compiled into a report. This includes information such as the type of garbage collection event, the cost (in milliseconds) of the GC, and detailed heap statistics before and after the test.
   - A comment is generated with the memory leak report and is programmatically added or updated in a pull request (PR) associated with the test suite. This comment includes a header indicating a memory leak detection, a summary of the issue, detailed statistics, and recommendations for further investigation.
   - Optionally, if enabled through an environment variable (`WRITE_SNAPSHOT_ON_MEMORY_LEAK`), a heap snapshot is taken for leaks exceeding **`1MB`** (defined as `HEAP_SIZE_DIFF_SNAPSHOT_THRESHOLD`). This snapshot can be used for a more in-depth analysis of the memory allocation and retention patterns. (this env is disabled everywhere by default and can be enabled locally by the person who is investigating the memory leak)

**Related issue(s)**:

Fixes #2260

**Notes for reviewer**:

**Checklist**


packages/relay/tsconfig.json
```diff
@@ -14,7 +14,8 @@
         "forceConsistentCasingInFileNames": true,
         "noImplicitAny": false,
         "declaration": true,
-        "strict": true
+        "strict": true,
+        "sourceMap": true
```
ebadiere, line 18: I know the project is open source, but can we confirm that it's okay to build with this as a default?  As set production builds for mainnet would include `sourceMap` information.

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -429,6 +429,8 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
           ],
           requestId,
         );
+        //resetting mirror node limit to 100
+        process.env['MIRROR_NODE_LIMIT_PARAM'] = '100';
```
ebadiere, line 433: Nice, thank you

ebadiere (commented):
Nice work!  Once this is merged it will get picked up in CI.  I'd like to discuss briefly some window of time during the first release when this is run, and you can be available if the CI breaks down.  CI updates can make releases interesting.

packages/server/tests/helpers/utils.ts
```diff
@@ -372,4 +383,229 @@ export class Utils {
...
+              ).filter((spaceStatistics) => spaceStatistics.spaceSize > 0),
+            },
+          }));
+          console.error(
+            `Total Heap Size ${Utils.formatBytes(totalDiffBytes)}: --> ` + JSON.stringify(statsDiff, null, 2),
+          );
+          // add comment on PR highlighting after which test the memory leak is happening
+          const testTitle = this.currentTest?.title ?? 'Unknown test';
+          const comment = Utils.generateMemoryLeakComment(testTitle, statsDiff);
+          await githubClient.addOrUpdateExistingCommentOnPullRequest(comment, (existing: string) =>
```
quiet-node, line 459: The addOrUpdateExistingCommentOnPullRequest() will first pull the context from GtiHub first for stuff like owner, repo, pullNumber, etc. right? Then just curious how this will play out locally when no GitHub-related-env-vars are provided?

packages/server/tests/helpers/utils.ts
```diff
@@ -372,4 +383,229 @@ export class Utils {
   static async wait(time: number): Promise<void> {
     await new Promise((r) => setTimeout(r, time));
   }
+
+  static async writeHeapSnapshotAsync(): Promise<string | undefined> {
+    return new Promise((resolve, reject) => {
+      writeSnapshot((error, fileName) => {
```
quiet-node, line 389: qq: what is fileName? Is it something predefined within writeSnapshot() method?

ebadiere (approved):

packages/server/tests/helpers/utils.ts
```diff
@@ -372,4 +383,229 @@ export class Utils {
...
+
+    return message;
+  }
+
+  /**
+   * Converts a string in camel case to title case.
+   * @param textInCamelCase The text in camel case.
+   * @return The text in title case.
+   */
+  private static camelCaseToTitleCase(textInCamelCase: string): string {
```
quiet-node, line 528: This and snakeCaseToTitleCase() can potentially be used in a global scale so maybe consider switching it to public? Or well maybe we can address it in future PRs where these are used in global scope.

packages/server/tests/integration/test.env
```diff
@@ -1,2 +1,3 @@
 E2E_SERVER_PORT=7546
 CHAIN_ID=0x12a
+TEST=true
```
quiet-node, line 3: qq: i might have missed it but what's the reason for this config again?

quiet-node (approved):
Wonderful work 🚀 🚀 !! Some qqs

packages/server/tests/integration/test.env
victor-yanev: It’s to use LRU cache instead of expecting a redis server to be up in the unit tests

packages/server/tests/helpers/utils.ts
victor-yanev: Let’s address it in future PRs where it might be used, I barely got the CI to work for this PR 😅

packages/server/tests/helpers/utils.ts
victor-yanev: Yeah, it’s an automated file name `heapdump-<timestamp>.json`

packages/server/tests/helpers/utils.ts
victor-yanev: It wouldn’t work locally, but it will only log an error, tests will still pass

packages/relay/tsconfig.json
victor-yanev: I recommend we add another tsconfig which is used for production builds:

**tsconfig.prod.json:**
```json
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "sourceMap": false
  }
}
```

victor-yanev: But let's do that in a separate issue, the CI is really unstable right now and I believe we want to merge this PR before the next release?

PR #2698, opened 2024-07-12
fix: Hbar usage tracking when transaction is larger than FILE_APPEND_CHUNK_SIZE

### LINKED ISSUE #2693: Use executeAll() instead of execute() in FileCreate/Append transactions
### Description

When creating a file with FileCreateTransaction, multiple fileAppends can be performed internally by the sdk. When using `execute()` only the first transactionId is returned, while using `executeAll()` returns a list of all possible transactionsIds, which then need to be looped over to find the sum of all fees.

Ivo-Yankov:
**Description**:
- Replaces `.execute()` with `.executeAll()` for `fileAppendTransaciton`. This way all transactions are returned in the result. The `transaction_fee` is summed and subtracted from the remaining Hbar limit in the limiter class.
- Removes some unused functions from `sdkClient.ts`.
- Adds additional acceptance tests for `HbarLimiter`. 
- Changes the relay operator when running acceptance tests to one of the local-node generated accounts.

**Related issue(s)**:

Fixes #2693 

**Notes for reviewer**:

**Checklist**


quiet-node (dismissed):
LGTM

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -687,22 +591,27 @@ export class SDKClient {
...
+        let totalTinybarsCost = 0;
+        for (let fileAppendTxResponse of fileAppendTxResponses) {
+          // get transaction fee and add expense to limiter
+          const appendFileRecord = await fileAppendTxResponse.getRecord(this.clientMain);
+          totalTinybarsCost += appendFileRecord.transactionFee.toTinybars().toNumber();
+        }
+
+        if (totalTinybarsCost > 0) {
+          this.hbarLimiter.addExpense(totalTinybarsCost, currentDateNow);
+          this.captureMetrics(
```
Nana-EC, line 605: Q: Doesn't this throw metrics of transaction count off? 
The below will capture 1 FileAppend count metric but indeed will capture the hbar expense of the collection File appends.

packages/server/tests/acceptance/rateLimiter.spec.ts
```diff
@@ -155,6 +158,46 @@ describe('@ratelimiter Rate Limiters Acceptance Tests', function () {
 
         await expect(relay.call(testConstants.ETH_ENDPOINTS.ETH_SEND_RAW_TRANSACTION, [signedTx], requestId)).to.be
           .fulfilled;
+        const remainingHbarsAfter = Number(await metrics.get(testConstants.METRICS.REMAINING_HBAR_LIMIT));
+        expect(remainingHbarsAfter).to.be.eq(remainingHbarsBefore);
```
Nana-EC, line 162: This seems off. 
I would expect the REMAINING_HBAR_LIMIT to have decreased for any call to `eth_sendRawTransaction`.
This is great that we're capturing it though

packages/server/tests/acceptance/rateLimiter.spec.ts
```diff
@@ -155,6 +158,46 @@ describe('@ratelimiter Rate Limiters Acceptance Tests', function () {
...
+
+        const largeContract = await Utils.deployContract(
+          largeContractJson.abi,
+          largeContractJson.bytecode,
+          accounts[0].wallet,
+        );
+        await largeContract.waitForDeployment();
+        const remainingHbarsAfter = Number(await metrics.get(testConstants.METRICS.REMAINING_HBAR_LIMIT));
+        expect(largeContract.target).to.not.be.null;
+        expect(remainingHbarsAfter).to.be.lt(remainingHbarsBefore);
```
Nana-EC, line 177: Over time we can improve this beyond a `lt` but for now this is great capturing the metrics usage

packages/server/tests/clients/metricsClient.ts
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Retrieves the value of a specified metric
+   * @param metric
+   * @param requestId
+   */
+  async get(metric: string, requestId?: string) {
+    const requestIdPrefix = Utils.formatRequestIdMessage(requestId);
+    this.logger.debug(`${requestIdPrefix} [GET] Read all metrics from ${this.relayUrl}/metrics`);
+    const allMetrics = (await this.client.get('')).data;
+    const allMetricsArray = allMetrics.split('\n');
```
Nana-EC, line 56: Nice idea.
Can we explain and illustrate the text layout here so it's clear what the text matching is doing?

Nana-EC (changes requested):
Great ideas.
Some questions and clarifications

packages/server/tests/acceptance/rateLimiter.spec.ts
Ivo-Yankov: This is correct. For transactions with small callData size the signer pays for the transaction. Only when for callData > 5120 do we use the `createFile + appendFile` flow. We cannot charge the signer for those transactions, because they are Hedera specific, and thus the relay operator pays for them.

packages/relay/src/lib/clients/sdkClient.ts
Ivo-Yankov: I changed it so it executes `addExpense` and `captureMetrics` for every single `fileAppendTransaction`

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -664,7 +567,8 @@ export class SDKClient {
         .setKeys(client.operatorPublicKey ? [client.operatorPublicKey] : []);
 
       const fileCreateTxResponse = await fileCreateTx.execute(client);
-      const { fileId } = await fileCreateTxResponse.getReceipt(client);
+      const fileCreateReceipt = await fileCreateTxResponse.getReceipt(client);
```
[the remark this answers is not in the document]
Ivo-Yankov: fixed

packages/server/tests/clients/metricsClient.ts
Ivo-Yankov: Added a detailed comment

packages/server/tests/acceptance/index.spec.ts
```diff
@@ -86,12 +87,22 @@ describe('RPC Server Acceptance Tests', function () {
...
   global.relay = new RelayClient(RELAY_URL, logger.child({ name: `relay-test-client` }));
   global.relayServer = relayServer;
   global.socketServer = socketServer;
   global.logger = logger;
   global.initialBalance = INITIAL_BALANCE;
 
+  global.restartLocalRelay = async function () {
+    if (global.relayIsLocal) {
+      stopRelay();
+      await new Promise((r) => setTimeout(r, 5000)); // wait for server to shutdown
```
Nana-EC, line 100: nit: seems like it could cause flakiness but we can address later in other PRs

packages/server/tests/clients/metricsClient.ts
Nana-EC: Thanks 
nit: put an example of the text in there also

packages/server/tests/localAcceptance.env
```diff
@@ -24,3 +24,7 @@ WS_NEW_HEADS_ENABLED=false
 INITIAL_BALANCE='5000000000'
 LIMIT_DURATION=90000
 SERVER_REQUEST_TIMEOUT_MS=60000
+
+# Only used in acceptance tests
```
Nana-EC, line 28: nit - note this is a static account created only in the local node tool

ebadiere (dismissed):
LG.

Nana-EC (approved):
LG, unblocking
Minor suggestions

packages/server/tests/localAcceptance.env
```diff
@@ -24,3 +24,7 @@ WS_NEW_HEADS_ENABLED=false
...
+OPERATOR_ID_MAIN_NON_GENESIS=0.0.1022
+OPERATOR_KEY_MAIN_NON_GENESIS=302e020100300506032b657004220420a608e2130a0a3cb34f86e757303c862bee353d9ab77ba4387ec084f881d420d4
```
ebadiere, line 30: I like this.  IMO it should be used in more tests.

package.json
```diff
@@ -33,7 +33,7 @@
     "acceptancetest:api_batch2": "ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@api-batch-2' --exit",
     "acceptancetest:api_batch3": "ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@api-batch-3' --exit",
     "acceptancetest:erc20": "ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@erc20' --exit",
-    "acceptancetest:ratelimiter": "ts-mocha packages/ws-server/tests/acceptance/index.spec.ts  -g '@web-socket-ratelimiter' --exit && ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@ratelimiter' --exit",
+    "acceptancetest:ratelimiter": "ts-mocha packages/ws-server/tests/acceptance/index.spec.ts -g '@web-socket-ratelimiter' --exit && HBAR_RATE_LIMIT_TINYBAR=3000000000 ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@ratelimiter' --exit",
```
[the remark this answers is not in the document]
Nana-EC: @Ivo-Yankov you removed it then re-added it

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -525,91 +524,6 @@ export class SDKClient {
     }
   };
 
-  async executeGetTransactionRecord(
```
Nana-EC, line 528: Don't remove this.
I think it's not called because of a bug and in follow up we should get logic using this method instead of repeating limiter and metrics calls everywhere.
SHould sync with @georgi-l95 PR to see incoming logic and after both we revisit again for conformity

package.json
Ivo-Yankov, line 36: This is the only way I got it to work with this env value for only this test suite. It turned out that changing `process.env` and restarting the relay instance does not actually change the initial values. If needed I can address this in a future PR

Nana-EC (dismissed):
Unblocking but see minor items before merge

packages/relay/src/lib/clients/sdkClient.ts
Ivo-Yankov: Reverted this change

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -687,22 +590,24 @@ export class SDKClient {
...
-        this.captureMetrics(
-          SDKClient.transactionMode,
-          fileAppendTx.constructor.name,
-          Status.Success,
-          await this.calculateFileAppendTxTotalTinybarsCost(fileAppendTx),
-          0,
-          callerName,
-          interactingEntity,
-        );
+        const fileAppendTxResponses = await fileAppendTx.executeAll(client);
```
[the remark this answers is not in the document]
Ivo-Yankov: I opened an issue for this

package.json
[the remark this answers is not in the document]
Ivo-Yankov: yes, I added a comment [here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2698#discussion_r1681361904)

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: So `executeGetTransactionRecord()` is the function that get transaction records using the SDK via `getRecord()` method. And like we've discussed before, `getRecord()` cost operator some fees that don't get captured in metrics and rate limitter class. That's why we wanted to replace `getRecord()` with calls to mirror-node like this [issue#2706](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2706) is trying to solve. I'm working on a solution for this ticket, and the solution makes repeated calls to mirror-node to get the correct `transactionRecord` using `transactionId`. Now this solution is expected to do the exact same job as the `executeGetTransactionRecord`, so in my opinion, this `executeGetTransactionRecord()` can be removed safely.

package.json
Nana-EC: Hmm okay, not a perfect solution so let's revisit after

Nana-EC (approved):

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC: @quiet-node agreed in the final case but let's do the SDK `getRecord()` replacement in follow up for `executeGetTransactionRecord()`.
Avoid conflict with PRs currently out also

quiet-node (approved):
LGTM

PR #2705, opened 2024-07-16
feat: track hbar expenses for submit tx

### LINKED ISSUE #2703: Expand Hbar Limiter coverage
### Description

It seems that in some cases we miss to account for the expenses that the operator is paying for.

### Steps to reproduce

1. Run relay
2. Do a transfer
3. Observe operator balance before and after (operator has to be different than 0.0.2)

### Additional context

_No response_

### Hedera network

### Version

main

### Operating system

None

georgi-l95:
**Description**:
This PR modifies submit transaction logic to account for hbar expenses. Previously we didn't track operator expenses for ethereum transactions. This PR is using executeGetTransactionRecord method, which takes the record and deduct the cost from the budget.

**Related issue(s)**:

Fixes #2703 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -468,61 +468,20 @@ export class SDKClient {
     const transactionType = transaction.constructor.name;
     const requestIdPrefix = formatRequestIdMessage(requestId);
     const currentDateNow = Date.now();
-    try {
-      const shouldLimit = this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.transactionMode, callerName);
-      if (shouldLimit) {
-        throw predefined.HBAR_RATE_LIMIT_EXCEEDED;
-      }
-
-      this.logger.info(`${requestIdPrefix} Execute ${transactionType} transaction`);
-      const resp = await transaction.execute(this.clientMain);
-
-      this.logger.info(
-        `${requestIdPrefix} ${resp.transactionId} ${callerName} ${transactionType} status: ${Status.Success} (${Status.Success._code})`,
-      );
-      return resp;
-    } catch (e: any) {
-      const sdkClientError = new SDKClientError(e, e.message);
-      let transactionFee: number | Hbar = 0;
-
-      // if valid network error utilize transaction id
-      if (sdkClientError.isValidNetworkError()) {
-        try {
-          const transactionRecord = await new TransactionRecordQuery()
-            .setTransactionId(transaction.transactionId!)
-            .setNodeAccountIds(transaction.nodeAccountIds!)
-            .setValidateReceiptStatus(false)
-            .execute(this.clientMain);
-          transactionFee = transactionRecord.transactionFee;
-
-          this.captureMetrics(
-            SDKClient.transactionMode,
-            transactionType,
-            sdkClientError.status,
-            transactionFee.toTinybars().toNumber(),
-            transactionRecord?.contractFunctionResult?.gasUsed,
-            callerName,
-            interactingEntity,
-          );
-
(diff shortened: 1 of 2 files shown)


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -468,61 +468,20 @@ export class SDKClient {
...
-      }
-
-      this.logger.info(`${requestIdPrefix} Execute ${transactionType} transaction`);
-      const resp = await transaction.execute(this.clientMain);
-
-      this.logger.info(
-        `${requestIdPrefix} ${resp.transactionId} ${callerName} ${transactionType} status: ${Status.Success} (${Status.Success._code})`,
-      );
-      return resp;
-    } catch (e: any) {
```
georgi-l95, line 484: removing try catch block, because the same operation is performed in the `executeGetTransactionRecord()` method.

ebadiere (commented):
LG, but can we add tests?

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC: Are we sure?
What happens when an executeTransaction fails for a PreCheck reason?
I believe that's what this check is for.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -541,15 +500,14 @@ export class SDKClient {
           `${requestIdPrefix} Invalid response format, expected record availability: ${JSON.stringify(resp)}`,
         );
       }
-      const shouldLimit = this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.recordMode, transactionName);
-      if (shouldLimit) {
-        throw predefined.HBAR_RATE_LIMIT_EXCEEDED;
-      }
+      const transactionRecord = await new TransactionRecordQuery()
+        .setTransactionId(resp.transactionId!)
+        .setValidateReceiptStatus(false)
```
Nana-EC, line 505: Missing the nodeIds selection.
Was this intentionally left out?
I believe the optimization allows for you to get it from the node you submitted to for shorter latency.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -541,15 +500,14 @@ export class SDKClient {
           `${requestIdPrefix} Invalid response format, expected record availability: ${JSON.stringify(resp)}`,
         );
       }
-      const shouldLimit = this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.recordMode, transactionName);
-      if (shouldLimit) {
-        throw predefined.HBAR_RATE_LIMIT_EXCEEDED;
-      }
+      const transactionRecord = await new TransactionRecordQuery()
```
Nana-EC, line 503: Use executeQuery() as it will handle metric and hbar rate limit

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC: Are we sure?
What happens when an executeTransaction fails for a PreCheck reason?
I believe that's what this check is for.

Nana-EC (commented):
Some suggestions

packages/relay/src/lib/clients/sdkClient.ts
georgi-l95: It should throw and be caught one level above in eth.ts catch handler https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1436

victor-yanev (approved):

georgi-l95:
Replaced by https://github.com/hashgraph/hedera-json-rpc-relay/pull/2714

PR #2714, opened 2024-07-18
fix: captured transactionFee in metrics and HBAR limiter class in executeTransaction and deleteFile

### LINKED ISSUE #2707: HBAR Burned Rate is not captured in executeTransaction() and deleteFile()
### Problem

Currently, the operator gets charged when an eth_sendRawTransaction is submitted. This process involves many transactions by the SDK, and each transacction would cost a different amount of HBAR. So far, there are four main transactions: FileCreate, FileAppend, ExecuteTransaction, and FileDelete. While charged HBAR is correctly added to metrics in FileCreate and FileAppend, the cost has not been captured in ExecuteTransaction and FileDelete.

### Solution

Add the HBAR to metrics and HBAR rate limitter class for ExecuteTransaction and FileDelete

### Alternatives

_No response_

quiet-node:
**Description**:
This PR is a hotfix for v0.51 that added steps to capture transactionFee in metrics and HBAR limiter class for executeTransaction and deleteFile.

To retrieve the `transactionFee` in `executeTransaction()`, the `getRecord()` method was utilized. The `getRecord()` method exposes and throws errors caused by invalid transactions on the network, which were previously ignored. These errors were typically discovered only after repeated lookups on the mirror node, with `MIRROR_NODE_GET_CONTRACT_RESULTS_RETRIES` (10) multiplied by `MIRROR_NODE_RETRY_DELAY` (2000ms), resulting in a total of 20000ms (20 seconds). This error exposure helps to reduce transaction time and ensures more accurate and secure handling of transactions.

This PR also tweaked and reused `executeGetTransactionRecord()` method to capture transactionFee of EthereumTransactions and add the fees to metrics and HBAR rate limiter class.

**Related issue(s)**:

Fixes #2707 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -469,18 +469,34 @@ export class SDKClient {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     const currentDateNow = Date.now();
     try {
-      const shouldLimit = this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.transactionMode, callerName);
-      if (shouldLimit) {
+      if (this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.transactionMode, callerName))
         throw predefined.HBAR_RATE_LIMIT_EXCEEDED;
-      }
 
       this.logger.info(`${requestIdPrefix} Execute ${transactionType} transaction`);
-      const resp = await transaction.execute(this.clientMain);
+
+      const transactionResponse = await transaction.execute(this.clientMain);
+
+      const transactionRecord = await transactionResponse.getRecord(this.clientMain);
+
+      // get transactionFee and capture it in metrics and HBAR limiter class
+      // @todo: The transactionFee includes the entire charges of the transaction, with some portions charged by tx.from, not the operator.
+      //        Determine how to separate the fee charged exclusively by the operator.
+      let transactionFee = transactionRecord.transactionFee;
+      this.hbarLimiter.addExpense(transactionFee.toTinybars().toNumber(), currentDateNow);
+      this.captureMetrics(
+        SDKClient.transactionMode,
+        transaction.constructor.name,
+        Status.Success,
+        transactionRecord.transactionFee.toTinybars().toNumber(),
+        transactionRecord?.contractFunctionResult?.gasUsed,
+        callerName,
+        interactingEntity,
+      );
 
       this.logger.info(
-        `${requestIdPrefix} ${resp.transactionId} ${callerName} ${transactionType} status: ${Status.Success} (${Status.Success._code})`,
+        `${requestIdPrefix} ${transactionResponse.transactionId} ${callerName} ${transactionType} status: ${Status.Success} (${Status.Success._code})`,
       );
-      return resp;
+      return transactionResponse;
     } catch (e: any) {
       const sdkClientError = new SDKClientError(e, e.message);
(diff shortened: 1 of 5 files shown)


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -468,19 +468,68 @@ export class SDKClient {
...
+
+        // get transactionFee and gasUsed for metrics
+        /**
+         * @todo: Determine how to separate the fee charged exclusively by the operator because
+         *        the transactionFee below includes the entire charges of the transaction,
+         *        with some portions paid by tx.from, not the operator.
+         */
+        transactionFee = transactionRecord.transactionFee.toTinybars().toNumber();
+        gasUsed = transactionRecord?.contractFunctionResult?.gasUsed.toNumber();
+      } catch (e: any) {
```
Nana-EC, line 493: Q: why the catch?
There's already a larger catch that should handle any transaction execution failures

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -468,19 +468,68 @@ export class SDKClient {
...
     try {
-      const shouldLimit = this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.transactionMode, callerName);
-      if (shouldLimit) {
+      if (this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.transactionMode, callerName))
         throw predefined.HBAR_RATE_LIMIT_EXCEEDED;
-      }
 
       this.logger.info(`${requestIdPrefix} Execute ${transactionType} transaction`);
-      const resp = await transaction.execute(this.clientMain);
+      const transactionResponse = await transaction.execute(this.clientMain);
```
Nana-EC, line 479: Q: Are you just extracting `executeGetTransactionRecord()`?
Why not just call it here as all this logic is already covered?
No longer calling `executeGetTransactionRecord()` was a regression.
```suggestion
      const transactionResponse = await transaction.execute(this.clientMain);
      executeGetTransactionRecord(transactionResponse, transactionType, callerName, interactingEntity, requestId);
```

Nana-EC (commented):
Curious why we don't just revert the regression and utilize `executeGetTransactionRecord()`

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Updated with new changes. `executeGetTransactionRecord()` is now tweaked and reused.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: updated with new changes.

georgi-l95 (approved):

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -469,18 +469,21 @@ export class SDKClient {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     const currentDateNow = Date.now();
     try {
-      const shouldLimit = this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.transactionMode, callerName);
```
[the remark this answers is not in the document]
quiet-node: oh yeah you're right that makes sense I removed this as I thought it was already added in the `executeGetTransactionRecord()` but we should definitely check it before execute the transaction. Reverted.

ebadiere (approved):

Nana-EC (approved):

PR #2720, opened 2024-07-18
feat: add real `logsBloom` where available

### LINKED ISSUE #2666: Add real logsBloom where available
### Problem

Currently when we request the block information from the mirror node, we don't return the actual logsBloom, because it was not available before

### Solution

Return the logsBloom from the mirror node response.

natanasow:
**Description**:

When we request the block information from the mirror node, we don't return the actual logsBloom, because it was not available before.

**Related issue(s)**:

Fixes #2666 

**Notes for reviewer**:

**Checklist**


ebadiere (approved):
LG.

georgi-l95 (approved):

PR #2737, opened 2024-07-22
feat: replaced getRecord network calls with calls to mirror node to avoid SDK costs 

### LINKED ISSUE #2706: Replace getRecord network calls with calls to Mirror node
### Problem

Calling the consensus node with `getRecord` charges the relay operator and not all such instances are captured by the metrics or the `hbarRateLimiter`. Find a way to call the mirror node where possible. A noteworthy place for this optimization is the `fileCreate / fileAppend` flow in the `sdkClient`

_No response_

quiet-node:
**Description**:
- Added an instance of Mirror Node client to the SDKClient class.

- Added the `retrieveAndCaptureChargedTxFee()` method to retrieve charged transaction fees by making repeated calls to the Mirror Node using the appropriate transactionId. This method is intended to replace the `getRecord()` methods called by the SDK to reduce costs.

- Refactored the createFile() and deleteFile() method to implement the new `retrieveAndCaptureChargedTxFee` logic.

**Related issue(s)**:

Fixes #2706 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -53,7 +53,7 @@ import {
 } from '@hashgraph/sdk';
 import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
 import { Logger } from 'pino';
-import { formatRequestIdMessage } from '../../formatters';
+import { formatRequestIdMessage, formatTransactionId, parseNumericEnvVar } from '../../formatters';
 import HbarLimit from '../hbarlimiter';
 import constants from './../constants';
 import { SDKClientError } from './../errors/SDKClientError';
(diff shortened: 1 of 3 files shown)


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -118,6 +133,15 @@ export class SDKClient {
...
     this.fileAppendChunkSize = Number(process.env.FILE_APPEND_CHUNK_SIZE) || 5120;
+
+    this.mirrorNodeClient = new MirrorNodeClient(
+      process.env.MIRROR_NODE_URL || '',
+      logger.child({ name: `mirror-node` }),
+      register,
+      this.cacheService,
+      undefined,
+      process.env.MIRROR_NODE_URL_WEB3 || process.env.MIRROR_NODE_URL || '',
+    );
```
victor-yanev, line 144: Please, let's inject this `MirrorNodeClient` instead of instantiating it here, I see no point in having multiple copies of this client in the heap space. Same for other places where `MirrorNodeClient` is used, we should inject the same object instance for those classes. 

In general, it's a good practice to apply dependency injection to make each class independent of its dependencies and have a loosely coupled program.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -98,9 +100,22 @@ export class SDKClient {
   private operatorAccountGauge;
   private maxChunks;
   private fileAppendChunkSize;
+  private mirrorNodeClient;
```
victor-yanev, line 103:
```suggestion
  private readonly mirrorNodeClient: MirrorNodeClient;
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -648,143 +671,91 @@ export class SDKClient {
...
-        }
+        // get all fileAppend transactions
+        fileAppendTxResponses = await fileAppendTx.executeAll(client);
       }
 
       // Ensure that the calldata file is not empty
       if (fileId) {
-        const fileSize = await (await new FileInfoQuery().setFileId(fileId).execute(client)).size;
+        // @todo: figure out the relationship between query.getCost() with the actual HBAR operator get charged as they are different => capture in metrics
+        const fileSize = (await new FileInfoQuery().setFileId(fileId).execute(client)).size;
```
victor-yanev, line 712: QQ: Why not use `await this.executeQuery` here?

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -795,47 +766,108 @@ export class SDKClient {
...
       // Create fileDeleteTx
-      const fileDeleteTx = new FileDeleteTransaction()
+      fileDeleteTx = new FileDeleteTransaction()
         .setFileId(fileId)
         .setMaxTransactionFee(new Hbar(2))
         .freezeWith(this.clientMain);
 
       // execute fileDeleteTx
-      const fileDeleteTxResponse = await fileDeleteTx.execute(this.clientMain);
+      await fileDeleteTx.execute(this.clientMain);
```
victor-yanev, line 783: Same here - don't we want to use `executeTransaction` here and also for all other transactions executed inside `SdkClient`?

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -98,9 +100,22 @@ export class SDKClient {
   private operatorAccountGauge;
   private maxChunks;
   private fileAppendChunkSize;
+  private mirrorNodeClient;
+
+  private readonly MirrorNodeRetries = parseNumericEnvVar(
+    'MIRROR_NODE_GET_CONTRACT_RESULTS_RETRIES',
+    'MIRROR_NODE_GET_CONTRACT_RESULTS_DEFAULT_RETRIES',
+  );
```
victor-yanev, line 108: This sounds like something that's a responsibility of the `MirrorNodeClient`, let's move it there.

packages/relay/src/lib/clients/sdkClient.ts
victor-yanev: Same for other lines above (98-102):
- please add types to those variables
- change them to `readonly` if we never reassign them anywhere in the class

packages/relay/src/lib/clients/sdkClient.ts
victor-yanev: We can also add a method `executeAllTransactions` and reuse it for cases where we want to call `executeAll`. 

In this way we can only define this `try`/`catch`/`finally` logic only in one place and we don't have to duplicate it everywhere, just have to make sure all transactions and queries are using `executeTransaction`/`executeAllTransactions` and `executeQuery`.

victor-yanev (changes requested):

PR #2749, opened 2024-07-24
feat: reused executetransaction for methods in sdkclient and improved logging

### LINKED ISSUE #2740: Block `gaslimit` should be `30000000` GAS
### Description

When querying block information, the block `gasLimit` is shown as 15M GAS (15000000). The correct value should be 30M GAS (30000000). This happens for all public networks (mainnet, testnet and previewnet).

### Steps to reproduce

Query the public networks for block information:

```zsh
❯ cast block latest --rpc-url https://mainnet.hashio.io/api

baseFeePerGas        1230000000000
difficulty           0
extraData            0x
gasLimit             15000000
gasUsed              0
hash                 0xbf9cd52312540ba120ab91fc59a8d00b4fe394b19e60495bec4c1d14d1c5693f
logsBloom            0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
miner                0x0000000000000000000000000000000000000000
mixHash              0x0000000000000000000000000000000000000000000000000000000000000000
nonce                0x0000000000000000
number               66858554
parentHash           0x3a4db7e2ad768dab95f2127930bc650cc8effc02b573696b0a5c8813fe8d0b66
transactionsRoot     0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421
receiptsRoot         0x0000000000000000000000000000000000000000000000000000000000000000
sha3Uncles           0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347
size                 682502
stateRoot            0x0000000000000000000000000000000000000000000000000000000000000000
timestamp            1721748748
withdrawalsRoot      0x0000000000000000000000000000000000000000000000000000000000000000
totalDifficulty      0
transactions:        []


❯ cast block latest --rpc-url https://testnet.hashio.io/api

baseFeePerGas        1230000000000
difficulty           0
extraData            0x
gasLimit             15000000
gasUsed              108617
hash                 0xbfcb7d0777f17437cd09b60361383657c9f4088fdd280cd8a81692f822e13034
logsBloom            0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
miner                0x0000000000000000000000000000000000000000
mixHash              0x0000000000000000000000000000000000000000000000000000000000000000
nonce                0x0000000000000000
number               6921714
parentHash           0x81bb2b2b3cb326a9164f879840a6056d75d491757e3d1cc6b959233c300aa9c6
transactionsRoot     0xbfcb7d0777f17437cd09b60361383657c9f4088fdd280cd8a81692f822e13034
receiptsRoot         0x0000000000000000000000000000000000000000000000000000000000000000
sha3Uncles           0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347
size                 3662
stateRoot            0x0000000000000000000000000000000000000000000000000000000000000000
timestamp            1721748756
withdrawalsRoot      0x0000000000000000000000000000000000000000000000000000000000000000
totalDifficulty      0
transactions:        [
	0x9985171cd35dd1d0d913cdf257f80195b70c0bc4f3ddc630e3cdd389b4d09374
	0x2dc53875a197b4c0ee1d9372d50c5d15f4086d3a3e9ee5dcc3c704f960715ac0
]


❯ cast block latest --rpc-url https://previewnet.hashio.io/api

baseFeePerGas        710000000000
difficulty           0
extraData            0x
gasLimit             15000000
gasUsed              0
hash                 0x38102d37d55cce332d8564d66e54abdaf12dd5fb51727dc6ddb8d768c5405600
logsBloom            0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
miner                0x0000000000000000000000000000000000000000
mixHash              0x0000000000000000000000000000000000000000000000000000000000000000
nonce                0x0000000000000000
number               477381
parentHash           0x111f0dd3651934f23dab53670bb7c709d545b77564b2c2dd5b02bc21dadfd907
transactionsRoot     0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421
receiptsRoot         0x0000000000000000000000000000000000000000000000000000000000000000
sha3Uncles           0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347
size                 1656
stateRoot            0x0000000000000000000000000000000000000000000000000000000000000000
timestamp            1721748761
withdrawalsRoot      0x0000000000000000000000000000000000000000000000000000000000000000
totalDifficulty      0
transactions:        []

```

### Additional context

_No response_

### Hedera network

mainnet, testnet, previewnet

### Version

v0.51.3

### Operating system

None

### LINKED ISSUE #2746: Reuse executeTransaction() for methods in SDKClient and improve logging
### Problem

Currently, the `fileCreate`, `fileAppend`, and `fileDelete` methods operate independently, resulting in repetitive and lengthy code. By reusing `executeTransaction()`, we can avoid code duplication, ensure these methods share the same workflow, and make logging more descriptive.

### Solution

Reuse executeTransaction in fileCreate, fileAppend, and fileDelete.

### Alternatives

_No response_

quiet-node:
**Description**:
- updated `executeTransaction()` so it can perform executeAll() for type FileAppend transactions 
- reused executeTransaction() for fileCreate, fileAppend, and fileDeelete so they can share the same workflow
- improve logging so it can be descriptive for each transactions. Add logs that contain requestId, transactionId, transactionType, transactionCost, gasUsed, etc.

**Related issue(s)**:

Fixes #2746

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -38,7 +38,10 @@ const generateRandomHex = (bytesLength = 16) => {
  * Format message prefix for logger.
  */
 const formatRequestIdMessage = (requestId?: string): string => {
-  return requestId ? `[${constants.REQUEST_ID_STRING}${requestId}]` : '';
+  if (!requestId) {
+    return '';
+  }
+  return requestId.includes(constants.REQUEST_ID_STRING) ? requestId : `[${constants.REQUEST_ID_STRING}${requestId}]`;
 };
 
 function hexToASCII(str: string): string {
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -614,6 +614,7 @@ export class SDKClient {
       }
 
       // get transactionRecord
+      this.logger.trace(`${requestId} Get transaction record: transactionId=${transactionId}`);
       const transactionRecord: TransactionRecord = await transactionResponse.getRecord(this.clientMain);
 
       // get transactionFee and gasUsed for metrics
@@ -814,7 +815,6 @@ export class SDKClient {
    */
   async deleteFile(fileId: FileId, requestId: string, callerName: string, interactingEntity: string): Promise<void> {
     // format request ID msg
-    const currentDateNow = Date.now();
     const requestIdPrefix = formatRequestIdMessage(requestId);
 
     try {
@@ -824,29 +824,12 @@ export class SDKClient {
         .setMaxTransactionFee(new Hbar(2))
         .freezeWith(this.clientMain);
 
-      // execute fileDeleteTx
-      const fileDeleteTxResponse = await fileDeleteTx.execute(this.clientMain);
-
-      // get fileDeleteTx's record
-      const deleteFileRecord = await fileDeleteTxResponse.getRecord(this.clientMain);
-
-      // capture transactionFee in metrics and HBAR limiter class
-      this.hbarLimiter.addExpense(deleteFileRecord.transactionFee.toTinybars().toNumber(), currentDateNow);
-      this.captureMetrics(
-        SDKClient.transactionMode,
-        fileDeleteTx.constructor.name,
-        Status.Success,
-        deleteFileRecord.transactionFee.toTinybars().toNumber(),
-        deleteFileRecord?.contractFunctionResult?.gasUsed,
-        callerName,
-        interactingEntity,
-      );
+      await this.executeTransaction(fileDeleteTx, callerName, interactingEntity, requestId);
 
       // ensure the file is deleted
-      const receipt = deleteFileRecord.receipt;
       const fileInfo = await new FileInfoQuery().setFileId(fileId).execute(this.clientMain);
 
-      if (receipt.status === Status.Success && fileInfo.isDeleted) {
+      if (fileInfo.isDeleted) {
         this.logger.trace(`${requestIdPrefix} Deleted file with fileId: ${fileId}`);
       } else {
         this.logger.warn(`${requestIdPrefix} Fail to delete file with fileId: ${fileId} `);
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -580,7 +580,8 @@ export class SDKClient {
       const sdkClientError = new SDKClientError(e, e.message);
 
       // get transactionId from error
-      transactionId = [e.transactionId];
+      // Note: In case of a FileAppendTransaction failure, expand the transactionId array.
+      transactionId.push(e.transactionId);
 
       // if valid network error utilize transaction id to get transactionFee and gasUsed for metrics
       if (sdkClientError.isValidNetworkError()) {
@@ -592,8 +593,9 @@ export class SDKClient {
             .execute(this.clientMain);
 
           // extract gas and txFee
-          transactionFee = transactionRecord.transactionFee.toTinybars().toNumber();
-          gasUsed = transactionRecord.contractFunctionResult
+          // Note: In case of a FileAppendTransaction failure, complement transactionFee and gasUsed.
+          transactionFee += transactionRecord.transactionFee.toTinybars().toNumber();
+          gasUsed += transactionRecord.contractFunctionResult
             ? transactionRecord.contractFunctionResult.gasUsed.toNumber()
             : 0;
         } catch (err: any) {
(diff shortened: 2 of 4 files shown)


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -459,109 +464,171 @@ export class SDKClient {
...
 
       // execute transaction
-      this.logger.info(`${requestIdPrefix} Execute ${transactionType} transaction`);
-      const transactionResponse = await transaction.execute(this.clientMain);
+      // logic: if transaction is typed FileAppendTransaction, use executeAll() to retrieve all fileAppend transaction responses
+      // logic: if transaction is any other type, use execute() to get the only transaction response
+      this.logger.info(`${requestId} Execute ${transactionType} transaction`);
+      if (transactionType === FileAppendTransaction.name) {
+        // execute transaction
+        transactionResponse = await (transaction as FileAppendTransaction).executeAll(this.clientMain);
```
Nana-EC, line 500: Q: Do we know what happens if a subset of FileAppends fail? Does executeAll throw an error?
Something to consider

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -459,109 +464,171 @@ export class SDKClient {
...
-    requestId?: string,
-  ) {
-    const requestIdPrefix = formatRequestIdMessage(requestId);
+  async executeGetTransactionRecord(transactionResponse: TransactionResponse, callerName: string, requestId: string) {
     const currentDateNow = Date.now();
     let gasUsed: any = 0;
     let transactionFee: number = 0;
     const transactionId: string = transactionResponse.transactionId.toString();
 
+    const shouldLimit = this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.recordMode, callerName);
```
Nana-EC, line 617: Interesting, now i'm wondering why we have a limit in here at all or if we're handling the logic right. Will rate limiting on getRecord cause a valid transaction flow to fail with an error?
Example you Submit an `EthereumTransaction` successfully but hit a rate limit on getRecord. What happens?

Nana-EC (commented):
Nice improvement, some additional flows to consider.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Ah yeah I have thought about this while writing this update. If executeAll() or execute() throw an error, it will fall into the catch block down below. If it fails from executtion level, then `transactionResponse` will be null and at line 582, it'll throw an internal error.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Ah yeah good point! hmm.. well to answer your question, if it hits a rate limit on `getRecord`, it will throw an `HBAR_RATE_LIMIT` error all the way to the top level and won't return any hashes, only the `jsonRpcError`. This is not ideal because the transaction was successfully submitted to the network. The `getRecord` function is like a secondary operation meant to retrieve the transaction fee for metrics purposes, and the charge is not too big.

Seems like this check is not necessary at this step.

quiet-node: removed this check in the new update.

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC: Well I mean if there are 3 File Appends and the 3rd one fails on a precheck what happens.
In that case we'd want t be able to capture the fees of the 1st and 2nd FIleAppend

quiet-node: Ah that's legit a solid point! 
Yeah that would fall into the catch block in `executeGetTransactionRecord()`, and I don't wanna calculate the transactionFee in the catch block of `executeGetTransactionRecord()` as that'd lead back to the old set up which caused capturing metrics twice. If I add the `addExpense` and `captureMetrics` inside the `for (let txResp of transactionResponse)` loop like Ivo did then it'd be a little messy. Ha interesting it's a diliema! I'll sleep on this a little more then!

quiet-node: @Nana-EC new updates are in and at [this line](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2749/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR567) and [these lines](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2749/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR580-R581), expand/add the `transactionIds` and `transactionFees` of the `FileAppendTx` even in case of an error. This way, the `finally` block will capture all the metrics and log all the `transactionIds` in the `FileAppendTransactions`.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -498,75 +503,125 @@ export class SDKClient {
...
+   * @param requestId
+   * @returns
+   */
   async executeTransaction(
     transaction: Transaction,
     callerName: string,
     interactingEntity: string,
     requestId: string,
-  ): Promise<TransactionResponse> {
+  ): Promise<TransactionResponse | TransactionResponse[]> {
```
victor-yanev, line 519: Please, define another method - `executeAllTransactions` which returns `TransactionResponse[]`. The `executeTransaction` method already has too many responsibilities and adding this change makes it even harder to maintain. 

Also, defining a separate method for executing multiple transactions can help us with type safety and avoid casting later inside `submitEthereumTransaction`:

Example usage:
```typescript
  async submitEthereumTransaction(
    transactionBuffer: Uint8Array,
    callerName: string,
    requestId: string,
  ): Promise<{ txResponses: TransactionResponse[]; fileId: FileId | null }> {
    ...
    if (transaction.constructor.name === FileAppendTransaction.name) {
      return {
        fileId,
        txResponses: await this.executeAllTransactions(
          ethereumTransaction,
          callerName,
          interactingEntity,
          requestId,
        ),
      };
    } else {
      return {
        fileId,
        txResponses: [
          await this.executeTransaction(
            ethereumTransaction,
            callerName,
            interactingEntity,
            requestId,
          ),
        ],
      };
    }
  }
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -498,75 +503,125 @@ export class SDKClient {
...
     const transactionType = transaction.constructor.name;
```
victor-yanev, line 520: Are we sure this would work? Have you tested it? 

Btw, why not just do `transaction instanceof FileAppendTransaction` instead of checking the constructor name?

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -498,75 +503,125 @@ export class SDKClient {
...
     const currentDateNow = Date.now();
     let gasUsed: number = 0;
     let transactionFee: number = 0;
-    let transactionResponse: TransactionResponse | null = null;
+    let transactionId = [] as string[];
+    let transactionResponse: TransactionResponse | TransactionResponse[] | null = null;
```
victor-yanev, line 525:
```suggestion
    let transactionId: string;
    let transactionResponse: TransactionResponse;
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -665,144 +714,45 @@ export class SDKClient {
...
-        } catch (err: any) {
-          const recordQueryError = new SDKClientError(err, err.message);
-          this.logger.error(
-            recordQueryError,
-            `${requestIdPrefix} Error raised during TransactionRecordQuery for ${fileCreateTx.transactionId}`,
-          );
-        }
-      }
+      // use executeTransaction() to execute fileAppendTx -> handle errors -> capture HBAR burned in metrics and hbar rate limit class
+      await this.executeTransaction(fileAppendTx, callerName, interactingEntity, requestId);
```
victor-yanev, line 742:
```suggestion
      await this.executeAllTransactions(fileAppendTx, callerName, interactingEntity, requestId);
```

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -2232,7 +2235,11 @@ describe('SdkClient', async function () {
         .resolves(getTransactionResponse('FileDeleteTransaction'));
 
       hbarLimitMock.expects('addExpense').withArgs(fileDeleteFee).once();
-      hbarLimitMock.expects('shouldLimit').never();
+      hbarLimitMock
+        .expects('shouldLimit')
+        .withArgs(sinon.match.any, SDKClient.transactionMode, callerName)
+        .once()
+        .returns(false);
```
victor-yanev, line 2242: We don't want to do any rate limiting on `FileDeleteTransaction`, we always want to clean up files so we don't want the rate limiter to block us in this case. 

It was discussed [here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2756#discussion_r1691468790).

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -665,144 +714,45 @@ export class SDKClient {
...
+    }
 
-      this.logger.info(`${requestIdPrefix} HBAR_RATE_LIMIT_EXCEEDED cost: ${transactionFee}`);
+    // Ensure that the calldata file is not empty
+    if (fileId) {
+      const fileSize = (await new FileInfoQuery().setFileId(fileId).execute(client)).size;
 
-      if (error instanceof JsonRpcError) {
-        throw predefined.HBAR_RATE_LIMIT_EXCEEDED;
+      if (callData.length > 0 && fileSize.isZero()) {
```
victor-yanev, line 749: QQ: Do we want to execute the `FileCreateTransaction` at all when we have `callData.length === 0`? What's the point of creating an empty file?

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Thanks for the input! Yeah I see and agree that splitting executeTransaction up and put executeAll() logic into a new method would make the code more maintainable. Pushing in updates soon!

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Yeah we've been using this `transaction.constructor.name` for quite a while. This is to get the type of the transaction for the purposes of logging and adding to metrics. 

Though I like this `transaction instanceof FileAppendTransaction` for checking if transactions are FileAppend transactions. But now we're splitting the executeTransaction so it might not bee necessary.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Yeah this check has been there as well I guess it's just another extra confirmation and its purpose is to make sure that the file created is not an empty file. 

In case of callData.length = 0 or less than 5kbs createFile() is not even involved. 

Are you suggesting we should remove this check block?

packages/relay/tests/lib/sdkClient.spec.ts
quiet-node: oh yeah nice catch! Nana and I also had a conversation about this I must've forgotten about this since the shouldLimit check is in executeTransaction() and deleteFile() now uses executeTransaction(). Will push in a solution soon!

quiet-node (commented):
Thanks for the input @victor-yanev

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -594,34 +610,126 @@ export class SDKClient {
...
+            gasUsed = transactionRecord.contractFunctionResult
+              ? transactionRecord.contractFunctionResult.gasUsed.toNumber()
+              : 0;
+          } catch (err: any) {
+            const recordQueryError = new SDKClientError(err, err.message);
+            this.logger.error(
+              recordQueryError,
+              `${formattedRequestId} Error raised during TransactionRecordQuery for ${transactionResponse.transactionId}`,
+            );
+          }
```
victor-yanev, line 681: Let's extract this into a separate method, nested try/catch blocks are really tough to read

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -665,144 +773,47 @@ export class SDKClient {
...
-            `${requestIdPrefix} ${fileCreateTx.transactionId} ${callerName} ${fileCreateTx.constructor.name} status: ${sdkClientError.status} (${sdkClientError.status._code}), cost: ${transactionFee}`,
-          );
+    // use executeTransaction() to execute fileCreateTx -> handle errors -> capture HBAR burned in metrics and hbar rate limit class
+    const fileCreateTxResponse = (await this.executeTransaction(
+      fileCreateTx,
+      callerName,
+      interactingEntity,
+      formattedRequestId,
+      true,
+    )) as TransactionResponse;
```
victor-yanev, line 791: QQ: Do we need this casting anymore?

packages/relay/src/lib/clients/sdkClient.ts
victor-yanev: Let's change it only to:
```suggestion
      if (fileSize.isZero()) {
```

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -2192,8 +2192,8 @@ describe('SdkClient', async function () {
         .resolves(Array.from({ length: fileAppendChunks }, () => getTransactionResponse('FileAppendTransaction')));
 
       hbarLimitMock.expects('addExpense').withArgs(fileCreateFee).once();
-      hbarLimitMock.expects('addExpense').withArgs(fileAppendFee).thrice();
-      hbarLimitMock.expects('shouldLimit').once().returns(false);
+      hbarLimitMock.expects('addExpense').withArgs(fileAppendFee).atLeast(fileAppendChunks);
```
victor-yanev, line 2195: QQ: Why use `atLeast` and not `exactly` here?

victor-yanev (approved):
LGTM

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Good catch! Thanks!

packages/relay/src/formatters.ts
```diff
@@ -38,7 +38,10 @@ const generateRandomHex = (bytesLength = 16) => {
  * Format message prefix for logger.
  */
 const formatRequestIdMessage = (requestId?: string): string => {
-  return requestId ? `[${constants.REQUEST_ID_STRING}${requestId}]` : '';
+  if (!requestId) {
+    return '';
+  }
+  return requestId.includes(constants.REQUEST_ID_STRING) ? requestId : `[${constants.REQUEST_ID_STRING}${requestId}]`;
```
victor-yanev, line 44: nit: We can only check the length of the string, since we know that UUID strings have a length of `36`

> A standard UUID code contains 32 hex digits along with 4 “-” symbols, which makes its length equal to 36 characters.

```suggestion
  if (requestId.length > UUID_LENGTH) {
    return requestId;
  }
  return `[${constants.REQUEST_ID_STRING}${requestId}]`;
```
```suggestion
  return requestId.length > UUID_LENGTH ? requestId : `[${constants.REQUEST_ID_STRING}${requestId}]`;
```

victor-yanev (approved):
Looks good 🎉! Just two last suggestions:

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -848,4 +812,59 @@ export class SDKClient {
...
+  }
+
+  /**
+   * Handles errors encountered during the execution of all transactions by retrieving the transaction record.
+   * @private
+   * @param {TransactionResponse} transactionResponse - The response of the transaction that encountered an error.
+   * @param {string} formattedRequestId - The formatted request ID for logging purposes.
+   * @returns {Promise<{transactionFee: number, gasUsed: number}>} A promise that resolves to an object containing the transaction fee and gas used.
+   */
+  private async handleExecuteAllError(
```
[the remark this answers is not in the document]
quiet-node: Nice I like it! This made me look back and it seems like it can be reused in executeTransaction as well. Just pushed in the update.

packages/relay/src/formatters.ts
quiet-node: Ah, I see. Thanks, that's a good point! However, I think `constants.REQUEST_ID_STRING` is already present and used. Since we know a formatted request ID will have that prefix, checking if the passed-in argument has that prefix should be sufficient enough.

victor-yanev: Checking only the length is slightly faster - takes `O(1)` time, whereas checking if the string contains the prefix is roughly `O(m)` (where `m` is the length of the prefix), but the difference is negligible cause our prefix is a very short string, so it's up to you!

victor-yanev (approved):

ebadiere (approved):

PR #2751, opened 2024-07-24
feat: added preemtiveCheck for HBAR ratelimit

quiet-node:
**Description**:
Add first round solution to preemtively limit HBAR based on callDataSize

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #2753, opened 2024-07-24
test: added E2E test for preemtive rate limit

### LINKED ISSUE #2745: Calculate the cost of an eth_sendRawTransaction based on callData byte size
### Problem

Currently, in some cases, `eth_sendRawTransaction` can spend more than the remaining rate limit balance, causing it to exceed the limit significantly after the process is completed.

### Solution

Calculate the callData size beforehand and compare it with the remaining balance to determine if it should be limited or not.

### Alternatives

_No response_

quiet-node:
**Description**:
[PR#2751](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2751) added a preemptive check for the HBAR rate limit as a hotfix but did not include E2E test to cover the solution. This PR adds the necessary E2E test to ensure comprehensive coverage.

**Related issue(s)**:

Fixes #2745

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance/hbarLimiter.spec.ts
+++ b/packages/server/tests/acceptance/hbarLimiter.spec.ts
@@ -162,6 +162,26 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
         verifyRemainingLimit(expectedCost, remainingHbarsBefore, remainingHbarsAfter);
       });
 
+      it('Should preemtively check the rate limit before submitting EthereumTransaction', async function () {
+        const remainingHbarsBefore = Number(await metrics.get(testConstants.METRICS.REMAINING_HBAR_LIMIT));
+
+        process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK = 'true';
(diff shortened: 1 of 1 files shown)


Nana-EC (dismissed):

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -162,6 +162,28 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+
+        try {
+          const largeContract = await Utils.deployContract(
+            largeContractJson.abi,
+            largeContractJson.bytecode,
+            accounts[0].wallet,
+          );
+          await largeContract.waitForDeployment();
+
+          expect(true).to.be.false;
```
acuarica, line 179: `chai` provides methods to test whether a function call throws an exception, see https://www.chaijs.com/api/bdd/#method_throw. Maybe we should use that instead of a `expect(true).to.be.false`.

ebadiere (approved):
LG.

acuarica (approved):
lg

PR #2756, opened 2024-07-25
test: expand integration tests for SDK client

### LINKED ISSUE #2747: Expand integration tests for SDK client
### Problem

Currently there is limited coverage on the SDK client in the `sdkClient.spec.ts` file.

### Solution

Add integration tests covering `createFile()` (with `callData.length` > `fileAppendChunkSize` and without), `deleteFile()`, `executeQuery`, `executeTransaction`, `executeGetTransactionRecord ` and verify that `hbarLimiter.addExpense()` has been called with the transaction cost (e.g., with a tolerance of 0.1% for delta)

### Alternatives

_No response_

victor-yanev:
**Description**:

This PR adds integration tests covering:
- `createFile()` (with `callData.length > fileAppendChunkSize` and `callData.length <= fileAppendChunkSize`)
- `deleteFile()`
- `executeQuery()`
- `executeTransaction()`
- `executeGetTransactionRecord()`.  

The tests verify that:
1. `hbarLimiter.shouldLimit()` has been called the appropriate amount of times before executing the transaction
2. `hbarLimiter.addExpense()` has been called the appropriate amount of times with the cost of each transaction

**Related issue(s)**:

Fixes #2747 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -424,13 +424,13 @@ export class SDKClient {
     );
   };
 
-  executeQuery = async (
-    query: Query<any>,
+  async executeQuery<T>(
+    query: Query<T>,
     client: Client,
     callerName: string,
     interactingEntity: string,
     requestId?: string,
-  ) => {
+  ): Promise<T> {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     const currentDateNow = Date.now();
     try {
@@ -496,14 +496,14 @@ export class SDKClient {
       }
       throw sdkClientError;
     }
-  };
+  }
 
-  executeTransaction = async (
+  async executeTransaction(
     transaction: Transaction,
     callerName: string,
     interactingEntity: string,
     requestId?: string,
-  ): Promise<TransactionResponse> => {
+  ): Promise<TransactionResponse> {
     const transactionType = transaction.constructor.name;
     const requestIdPrefix = formatRequestIdMessage(requestId);
     const currentDateNow = Date.now();
@@ -572,7 +572,7 @@ export class SDKClient {
       }
       throw sdkClientError;
     }
-  };
+  }
 
   async executeGetTransactionRecord(
     transactionResponse: TransactionResponse,
@@ -680,17 +680,17 @@ export class SDKClient {
     return balance.hbars.to(HbarUnit.Tinybar).multipliedBy(constants.TINYBAR_TO_WEIBAR_COEF);
   }
 
-  createFile = async (
+  async createFile(
     callData: Uint8Array,
     client: Client,
     requestId: string,
     callerName: string,
     interactingEntity: string,
-  ) => {
+  ): Promise<FileId | null> {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     const hexedCallData = Buffer.from(callData).toString('hex');
     const currentDateNow = Date.now();
-    let fileCreateTx, fileAppendTx;
+    let fileCreateTx: FileCreateTransaction, fileAppendTx: FileAppendTransaction;
 
     try {
       const shouldLimit = this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.transactionMode, callerName);
@@ -825,7 +825,7 @@ export class SDKClient {
       }
       throw sdkClientError;
     }
-  };
+  }
 
   /**
    * @dev Deletes `fileId` file from the Hedera Network utilizing Hashgraph SDK client
@@ -834,7 +834,7 @@ export class SDKClient {
    * @param callerName
    * @param interactingEntity
    */
-  public deleteFile = async (fileId: FileId, requestId: string, callerName: string, interactingEntity: string) => {
+  async deleteFile(fileId: FileId, requestId: string, callerName: string, interactingEntity: string): Promise<void> {
     // format request ID msg
     const currentDateNow = Date.now();
     const requestIdPrefix = formatRequestIdMessage(requestId);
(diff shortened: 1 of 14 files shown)


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -817,12 +834,18 @@ export class SDKClient {
...
    */
-  public deleteFile = async (fileId: FileId, requestId?: string, callerName?: string, interactingEntity?: string) => {
+  async deleteFile(fileId: FileId, requestId: string, callerName: string, interactingEntity: string): Promise<void> {
     // format request ID msg
     const currentDateNow = Date.now();
     const requestIdPrefix = formatRequestIdMessage(requestId);
 
     try {
+      // TODO: Limiting was missing here, not sure if it was intentional
+      const shouldLimit = this.hbarLimiter.shouldLimit(currentDateNow, SDKClient.transactionMode, callerName);
```
Nana-EC, line 844: Was intentional.
We always want to clean up fies so we don't want the rate limiter to block us in this case.

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -69,7 +74,7 @@ describe('SdkClient', async function () {
 
     client = client.setOperator(
       AccountId.fromString(process.env.OPERATOR_ID_MAIN!),
-      PrivateKey.fromString(process.env.OPERATOR_KEY_MAIN!),
+      PrivateKey.fromStringDer(process.env.OPERATOR_KEY_MAIN!),
```
Nana-EC, line 77: Q: why the change as part of this PR?
Does this force DER only?
I believe we wanted to move to a place that supported DER or hex for EVM tools.
If not related this should be reverted and tackled in another PR

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -2045,31 +2067,203 @@ describe('SdkClient', async function () {
       124, 18, 190, 114, 79, 189, 99, 27, 157, 117, 227, 107, 74, 255, 111, 74, 166, 7, 78, 25, 35, 131, 85, 47, 52,
       120, 20,
     ]);
+    const accountId = AccountId.fromString('0.0.1234');
```
Nana-EC, line 2070: Should make accountId configurable in case a test runner needs to change the account. The current value is the default

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -2045,31 +2067,203 @@ describe('SdkClient', async function () {
...
+
+      await sdkClient.executeGetTransactionRecord(
+        transactionResponse,
+        callerName,
+        interactingEntity,
+        new EthereumTransaction().constructor.name,
+      );
+
+      hbarLimitMock.verify();
+    });
```
Nana-EC, line 2267: This is really good.
Is it possible to add one test to ensure all the above checks happen in a flow for all the transactions and queries for say a transaction than needs 2 file appends?

packages/relay/tests/lib/sdkClient.spec.ts
victor-yanev: The `fromString` method was getting flagged for being deprecated by eslint, and since we use DER keys everywhere in our `.env` files I changed this to `fromStringDer`. Please let me know if I'm missing anything and I should revert this.

packages/relay/tests/lib/sdkClient.spec.ts
victor-yanev: Those are only mocked values, we don't do any real requests here

packages/relay/tests/lib/sdkClient.spec.ts
victor-yanev: Done:
- I added a new test `should rate limit before creating file and add expenses to limiter for large transaction data` which executes an `EthereumTransaction` with `callData` which needs 13 file append chunks - this is the most detailed test of all, also covers expenses for `FileContentsQuery` to retrieve the `FeeSchedules`
- I also updated the test `should execute FileCreateTransaction with callData.length > fileAppendChunkSize and add expenses to limiter` to use `callData.length := FILE_APPEND_CHUNK_SIZE * 2 + 1`

packages/relay/tests/lib/sdkClient.spec.ts
Nana-EC: Gotcha. 
It's only affecting the tests path in this case so it should be fine.
Carry on.

Nana-EC (approved):

packages/relay/tests/lib/sdkClient.spec.ts
quiet-node: Eric wrote a method to create private key based on format here https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/utils.ts#L44-L56. Can this be employed here to make it more dynamic?

victor-yanev: Replaced it with `createPrivateKeyBasedOnFormat `, thanks!

quiet-node (approved):
LGTM. 1 small QQ

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -150,24 +185,11 @@ describe('SdkClient', async function () {
...
-                    servicedata: undefined,
-                  },
-                ],
-              },
-            ],
-          },
-        };
+        return Promise.resolve(feeSchedules);
       });
+      // @ts-ignore
```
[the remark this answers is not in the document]
victor-yanev: Cause it was producing typescript warnings

ebadiere (approved):
LG.

PR #2762, opened 2024-07-29
feat: search contract logs only by transaction hash

### LINKED ISSUE #2744: Search contract logs only by transaction hash
### Problem

Currently, we are relying on requests to `eth_getBlockByNumber` to cache all synthetic transactions, which later (for a certain time, not indefinitely) are available when someone makes an `eth_getTransactionReceipt`. However, this results in a poor developer experience because if you want to search only by transaction hash, and you don’t know the block, it’s impossible.

### Solution

With the changes introduced in this [EPIC](https://github.com/hashgraph/hedera-mirror-node/issues/8758), it’s now possible to search synthetic contract logs by only knowing the transaction hash. This allows us to make the necessary changes in the relay and support this behavior.
It’s important that we add adequate tests for this as well, including transferring HTS tokens and querying the logs.

### Alternatives

_No response_

natanasow:
**Description**:

Currently, we are relying on requests to `eth_getBlockByNumber` to cache all synthetic transactions, which later (for a certain time, not indefinitely) are available when someone makes an `eth_getTransactionReceipt`. However, this results in a poor developer experience because if you want to search only by transaction hash, and you don’t know the block, it’s impossible.

**Solution**:

With the changes introduced in this [EPIC](https://github.com/hashgraph/hedera-mirror-node/issues/8758), it’s now possible to search synthetic contract logs by only knowing the transaction hash. This allows us to make the necessary changes in the relay and support this behavior.
It’s important that we add adequate tests for this as well, including transferring HTS tokens and querying the logs.

**Related issue(s)**:

Fixes #2744

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -2012,38 +2012,44 @@ export class EthImpl implements Eth {
       return cachedResponse;
     }
 
-    const cacheKeySyntheticLog = `${constants.CACHE_KEY.SYNTHETIC_LOG_TRANSACTION_HASH}${hash}`;
-    const cachedLog = await this.cacheService.getSharedWithFallback(
-      cacheKeySyntheticLog,
-      EthImpl.ethGetTransactionReceipt,
-      requestIdPrefix,
-    );
+    const receiptResponse = await this.mirrorNodeClient.getContractResultWithRetry(hash, requestIdPrefix);
+    if (receiptResponse === null || receiptResponse.hash === undefined) {
+      // handle synthetic transactions
+      const syntheticLogs = await this.common.getLogsWithParams(
+        null,
+        {
+          'transaction.hash': hash,
+        },
+        requestIdPrefix,
+      );
 
-    if (cachedLog) {
-      const gasPriceForTimestamp = await this.getCurrentGasPriceForBlock(cachedLog.blockHash);
+      // no tx found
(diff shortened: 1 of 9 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -2012,54 +2012,52 @@ export class EthImpl implements Eth {
...
-    const cacheKeySyntheticLog = `${constants.CACHE_KEY.SYNTHETIC_LOG_TRANSACTION_HASH}${hash}`;
-    const cachedLog = await this.cacheService.getSharedWithFallback(
-      cacheKeySyntheticLog,
-      EthImpl.ethGetTransactionReceipt,
-      requestIdPrefix,
-    );
+    const receiptResponse = await this.mirrorNodeClient.getContractResultWithRetry(hash, requestIdPrefix);
+    if (receiptResponse === null || receiptResponse.hash === undefined) {
+      // handle synthetic transactions
+      const syntheticLogs = await this.common.getLogsWithParams(
```
ebadiere, line 2018: So if I understand this correctly, the mirror node now supports using a transaction hash to fetch contract results in the logs, so the relay's existing calls to the mirror node didn't need to be updated.  The response just needed to be handled, correct?

ebadiere (approved):
LGTM

acuarica (approved):
lg

PR #2764, opened 2024-07-29
fix: Code coverage report is not being uploaded

### LINKED ISSUE #2699: Code coverage report is not being uploaded
### Description

Currently, the upload of the code coverage report fails with:
`Could not upload to Codecov: {'detail': ErrorDetail(string='Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 2566s.', code='throttled')}`
and we receive no code coverage reports on PRs.

### Steps to reproduce

Open a PR with some new changes in the code

**Expected:**
- Codecov report should appear in the PR highlighting the lines which were not covered and adding a comment with the summary of the report

**Actual:**
- upload fails and we have no coverage information in PRs currently

### Additional context

Ref: https://github.com/hashgraph/hedera-json-rpc-relay/actions/runs/9888198253/job/27311558978#step:13:29

### Hedera network

other

### Version

v0.51.1

### Operating system

None

victor-yanev:
**Description:**
This PR addresses the issue where the Codecov report is not being uploaded correctly. The following changes have been made:

**Changes:**
- Added `nyc` to all acceptance and integration test NPM run scripts
- Added new script for uploading coverage reports to Codecov `.github/scripts/codecov-upload.sh`
  - Download the latest Codecov uploader for Linux.
  - Check for the presence of a `coverage` directory in the root and each package directory.
  - Upload the coverage report if the `coverage` directory exists, otherwise log a message indicating no coverage report was found.
- Modified the `test.yml` and `acceptance-workflow.yml` GitHub workflows with step to run the new `codecov-upload.sh` script after the tests

**Testing:**
- Verified that the script correctly identifies and uploads coverage reports from the root and package directories.
- Ensured that appropriate messages are logged when no coverage report is found.

**Related Issues:**

Fixes #2699

**Notes:**
- The script requires `CODECOV_TOKEN` to be set in the repository secrets for successful uploads.


codecov.yml
```diff
@@ -0,0 +1,15 @@
...
+
+coverage:
+  status:
+    project:
+      default:
+        target: 80%
+        threshold: 1%
+
+ignore:
+  - "packages/*/dist"
```
AlfredoG87, line 15: blank line missing?

codecov.yml
```diff
@@ -0,0 +1,15 @@
+# Always comment on PRs
+comment:
+  layout: "diff, flags, files" # delete old and post new
+  behavior: new
```
AlfredoG87, line 4: should we set new?

I think is better to `update` the old one, so is cleaner and we don't have too many codecov comments... but that is just my opinion.

codecov.yml
```diff
@@ -0,0 +1,15 @@
...
+  require_changes: false
```
AlfredoG87, line 5: why would we want to post a coverage when there are no changes?

codecov.yml
```diff
@@ -0,0 +1,15 @@
...
+
+coverage:
+  status:
+    project:
+      default:
+        target: 80%
+        threshold: 1%
+
```
AlfredoG87, line 13: we should add:
```
patch:
      default:
        target: 80%
        threshold: 1%
```

So that every PR has at least 80% coverage. not sure if setting the needed coverage at the project level will also set it as requirement at the patch (PR) level.

AlfredoG87 (commented):
Nice Work!!

codecov.yml
victor-yanev: Thanks! Nice catch

codecov.yml
victor-yanev: What this does is delete the old comment and post a new comment. (I put the comment on the wrong line above 😆)

I like this slightly more because the new comment goes at the bottom of the PR and it can be spotted more easily, otherwise old comments get lost somewhere above in-between the commits.

codecov.yml
victor-yanev: This is the default behavior, I just thought it's better to have it explicit. 

I don't have a strong opinion here, I can change this to `true`.

codecov.yml
victor-yanev: You're right, good catch.

codecov.yml
AlfredoG87: Let's leave it if is the default behavior, I agree that is better to have the configurations explicit! 👍

codecov.yml
AlfredoG87: Yes you are right, is better to always have it at the bottom, I was not aware that it deletes the old comment, good to know!! like this approach more as well 👍 

Thank you for explaining!

.github/scripts/codecov-upload.sh
```diff
@@ -0,0 +1,12 @@
...
+curl -Os https://uploader.codecov.io/latest/linux/codecov && chmod +x codecov
+
+for dir in packages/*; do
+  if [ -d "${dir}/coverage" ]; then
+    echo "Uploading coverage report for ${dir}"
+    ./codecov --dir "${dir}" --flags "$(basename "${dir}")" --token "${CODECOV_TOKEN}" --fail-on-error
+  else
+    echo "No coverage report found for ${dir}"
+  fi
+done
```
georgi-l95, line 12: new line

georgi-l95 (approved):
LG, one nit.

.github/scripts/codecov-upload.sh
victor-yanev: thanks, fixed it!

georgi-l95 (dismissed):
LG

georgi-l95 (approved):

georgi-l95 (approved):

quiet-node (dismissed):

ebadiere (dismissed):
LG

.github/workflows/test.yml
```diff
@@ -57,8 +57,9 @@ jobs:
           if-no-files-found: ignore
 
       - name: Upload coverage report
-        if: ${{ always() && !cancelled() }}
```
rbarkerSL, line 60: Why remove this line?

If anything I would keep it but slightly modified:

```
if: ${{ !cancelled() && always() }}
```

this way it'll only upload the coverage report if the workflow isn't cancelled...

.github/workflows/acceptance-workflow.yml
```diff
@@ -109,6 +109,11 @@ jobs:
           name: Test Results
           path: test-*.xml
 
+      - name: Upload coverage report
+        run: bash .github/scripts/codecov-upload.sh
```
rbarkerSL, line 113: I would probably add in the `if: ${{ !cancelled() && always() }}` conditional here as well.

rbarkerSL (changes requested):

ebadiere (approved):

PR #2785, opened 2024-08-02
feat: calculate `logsBloom` for synthetic transactions

### LINKED ISSUE #2664: Calculate logsBloom for synthetic transactions
### Problem

Currently with our synthetic transactions, which are constructed on the fly using events from the operations, we don't return adequte and real logsBloom. This might be problem for some tools, which are requiring logsBloom for their operation.

### Solution

We can easly calculate logsBloom for each receipt, this can be done, by taking a look at how it's implemented in the javascript execution client for ethereum [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/bloom/index.ts)

natanasow:
**Description**:
Currently with our synthetic transactions, which are constructed on the fly using events from the operations, we don't return adequate and real logsBloom. This might be a problem for some tools, which requires logsBloom for their operation.

**Solution**:

We can easily calculate logsBloom for each receipt, this can be done, by taking a look at how it's implemented in the javascript execution client for Ethereum [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/vm/src/bloom/index.ts)

**Related issue(s)**:

Fixes #2664

**Notes for reviewer**:

**Checklist**


packages/relay/tests/lib/utils.spec.ts
```diff
@@ -45,4 +46,39 @@ describe('Utils', () => {
...
+    });
+
+    it('should be able to validate address and topics in generated logsBloom', () => {
+      expect(Utils.checkInLogsBloom(address, expectedLogsBloom)).to.equal(true);
+      expect(Utils.checkInLogsBloom(topics[0], expectedLogsBloom)).to.equal(true);
+      expect(Utils.checkInLogsBloom(topics[1], expectedLogsBloom)).to.equal(true);
+      expect(Utils.checkInLogsBloom(topics[2], expectedLogsBloom)).to.equal(true);
+    });
+
+    it('should be able to validate non-existing address and topic in generated logsBloom', () => {
```
konstantinabl, line 76: Do we have a test for the scenario where an emptyBloom is returned, so basically testing the first if in the _createSyntheticLogsBloom_ method?

natanasow: Added.

georgi-l95 (approved):
LGTM

packages/relay/src/utils.ts
```diff
@@ -55,4 +58,55 @@ export class Utils {
...
+   * @param topics
+   */
+  public static createSyntheticLogsBloom(address: string, topics: string[]): string {
+    if (!topics.length) {
+      return EthImpl.emptyBloom;
+    }
+
+    const items = [address, ...topics];
+    const BYTE_SIZE = 256;
+    const MASK = 0x7ff;
```
victor-yanev, line 74: nit: We can extract these in static constants inside the `Utils` class and reuse them here and inside `checkInLogsBloom`:
```
private static readonly LOGS_BLOOM_BYTE_SIZE = 256;
private static readonly LOGS_BLOOM_MASK = 0x7ff;
```

packages/relay/src/utils.ts
```diff
@@ -55,4 +58,55 @@ export class Utils {
...
+    let match = true;
+    for (let i = 0; i < 3 && match; i++) {
+      const first2bytes = new DataView(itemBuf.buffer).getUint16(i * 2);
+      const loc = MASK & first2bytes;
+      const byteLoc = loc >> 3;
+      const bitLoc = 1 << loc % 8;
+      match = (bitvectorUint8Arr[BYTE_SIZE - byteLoc - 1] & bitLoc) !== 0;
+    }
+
+    return Boolean(match);
```
victor-yanev, line 110: nit: No need to cast here, we can just do `return match;` here if we declare the variable above `let match: boolean = true` with a specified type.

packages/relay/src/utils.ts
```diff
@@ -55,4 +58,55 @@ export class Utils {
...
+
+    return prepend0x(Buffer.from(bitvector).toString('hex'));
+  }
+
+  /**
+   * Check whether an item exists in the hex encoded logs bloom bitvector
+   * @param item
+   * @param bitvector
+   */
+  public static checkInLogsBloom(item: string, bitvector: string): boolean {
```
victor-yanev, line 95: We can either add input validation here:
```typescript
  public static checkInLogsBloom(item: string, bitvector: string): boolean {
    if (!item?.length) {
      throw new Error("Item must be a non-empty string");
    }
    if (!bitvector?.length) {
      throw new Error("Bitvector must be a non-empty string");
    }
    ...
  }
```

or add a safety check and make it return `false` in those cases:
```typescript
  public static checkInLogsBloom(item: string, bitvector: string): boolean {
    if (!item?.length || !bitvector?.length) {
      return false;
    }
    ...
  }
```

packages/relay/src/utils.ts
```diff
@@ -55,4 +58,55 @@ export class Utils {
         throw new Error(`Invalid OPERATOR_KEY_FORMAT provided: ${process.env.OPERATOR_KEY_FORMAT}`);
     }
   }
+
+  /**
+   * Generate logs bloom for synthetic transaction
+   * @param address
+   * @param topics
+   */
+  public static createSyntheticLogsBloom(address: string, topics: string[]): string {
```
victor-yanev, line 67: QQ: Why not rename this to `createLogsBloom`, is there anything specific to this method that makes it work only for synthetic transactions?

victor-yanev, line 67: We can also do validation on the inputs here:
```typescript
  public static createSyntheticLogsBloom(address: string, topics: string[]): string {
    if (!address?.length) {
      throw new Error("Address must be a non-empty string");
    }
    if (!Array.isArray(topics) || topics.some(topic => typeof topic !== 'string')) {
      throw new Error("Topics must be an array of strings.");
    }
    ...
  }
```

packages/relay/tests/lib/utils.spec.ts
```diff
@@ -45,4 +47,44 @@ describe('Utils', () => {
       });
     }
   });
+
+  describe('createSyntheticLogsBloom and checkInLogsBloom', () => {
```
victor-yanev, line 51: Add tests for:
- `createSyntheticLogsBloom` with:
  - `undefined`/`null`/`empty` address string
  - invalid topics array: `[...topics, "nonHexString"]`/`[...topics, 1]`/`[...topics, {}]`
- `checkInLogsBloom` with:
  - `undefined`/`null`/`empty` item string
  - `undefined`/`null`/`empty` bit vector string

packages/relay/src/lib/eth.ts
```diff
@@ -2032,7 +2032,7 @@ export class EthImpl implements Eth {
         from: EthImpl.zeroAddressHex,
         gasUsed: EthImpl.zeroHex,
         logs: [syntheticLogs[0]],
-        logsBloom: EthImpl.emptyBloom,
+        logsBloom: Utils.createSyntheticLogsBloom(syntheticLogs[0].address, syntheticLogs[0].topics),
```
victor-yanev, line 2035: I would personally extract these methods in a separate utility class `LogsBloomUtils` for better organization and modularity. 

In this way, we create a clear separation of concerns, making the `Utils` class focused only on generic helpers, while `LogsBloomUtils` handles all specific operations related to logs bloom filters.

victor-yanev (changes requested):
LGTM, I left a few suggestions

packages/relay/src/utils.ts
victor-yanev: Or we can return an empty bloom in those cases, but I like the idea of having validations more because we can spot if there is a bug anywhere - unless it's expected that we receive invalid inputs here.

packages/relay/src/utils.ts
natanasow: The method is used only for test purposes so the better place to be is `utils.spec.ts` IMO.

packages/relay/tests/lib/utils.spec.ts
```diff
@@ -45,4 +49,70 @@ describe('Utils', () => {
...
+      const res = LogsBloomUtils.buildLogsBloom(address, topics);
+      expect(expectedLogsBloom).to.equal(res);
+    });
+
+    it('should be able to validate address and topics in generated logsBloom', () => {
+      expect(checkInLogsBloom(address, expectedLogsBloom)).to.equal(true);
+      expect(checkInLogsBloom(topics[0], expectedLogsBloom)).to.equal(true);
+      expect(checkInLogsBloom(topics[1], expectedLogsBloom)).to.equal(true);
+      expect(checkInLogsBloom(topics[2], expectedLogsBloom)).to.equal(true);
+    });
```
victor-yanev, line 108:
```suggestion
    it('should be able to validate address and topics in generated logsBloom', () => {
      const logsBloom = LogsBloomUtils.buildLogsBloom(address, topics);
      expect(checkInLogsBloom(address, logsBloom)).to.be.true;
      expect(checkInLogsBloom(topics[0], logsBloom)).to.be.true;
      expect(checkInLogsBloom(topics[1], logsBloom)).to.be.true;
      expect(checkInLogsBloom(topics[2], logsBloom)).to.be.true;
    });
```

packages/relay/tests/lib/utils.spec.ts
```diff
@@ -45,4 +49,70 @@ describe('Utils', () => {
       });
     }
   });
+
+  describe('buildLogsBloom and checkInLogsBloom', () => {
```
victor-yanev, line 53:
```suggestion
  describe('buildLogsBloom', () => {
```

packages/relay/tests/lib/utils.spec.ts
```diff
@@ -45,4 +49,70 @@ describe('Utils', () => {
...
+    });
+
+    it('should be able to validate non-existing address and topic in generated logsBloom', () => {
+      expect(checkInLogsBloom(ZeroAddress, expectedLogsBloom)).to.equal(false);
+      expect(checkInLogsBloom('0xD865b78906938EfDD065Cb443Be31440bE08a7CE', expectedLogsBloom)).to.equal(false);
+      expect(
+        checkInLogsBloom('0x0000000000000000000000C70c3C06A4db619B7879d060B9215d528F584FcC', expectedLogsBloom),
+      ).to.equal(false);
+    });
+  });
```
victor-yanev, line 117: Still missing tests for `undefined`/`null` address and topics

packages/relay/src/logsBloomUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+   * @param address
+   * @param topics
+   */
+  public static buildLogsBloom(address: string, topics: string[]): string {
+    if (!address?.length) {
+      return EthImpl.emptyBloom;
+    }
+    if (!topics.length) {
+      return EthImpl.emptyBloom;
+    }
```
victor-yanev, line 40: Let's add a log here for better traceability, in case we have a bug somewhere:
```suggestion
    if (!address?.length) {
      log.warning(`Invalid address passed: ${address}. Returning empty bloom.`);
      return EthImpl.emptyBloom;
    }
    if (!topics?.length) {
      log.warning(`Invalid topics passed: ${topics}. Returning empty bloom.`);
      return EthImpl.emptyBloom;
    }
```

victor-yanev (approved):
LGTM, few last suggestions:

packages/relay/src/logsBloomUtils.ts
natanasow: The entire project uses a Pino logger. If we want to log a warning here, we must pass the logger from a level above or we should initialize it in that utility class. Also, we can't do it in the constructor because `buildLogsBloom` is a static method. I don't think the effort is worth it but if you insist on adding these warning logs here, I'll try to do it in the most clean way.

victor-yanev (approved):
LGTM

georgi-l95 (approved):

PR #2787, opened 2024-08-05
chore: return empty trie root where applicable

### LINKED ISSUE #2665: Insert actual emptyRoot where not applicable
### Problem

Currentlty in our implementation of Block, where we don't know the actual trie root, we are ruturning some random value or 0x. This is not correct as some tools might need this information or at least require this to be in the correct format.

### Solution

We can easly implement ethereum tree library, like this https://www.npmjs.com/package/@ethereumjs/trie. Using it we can define new empty trie and either populate or return actual emtpy root.

Ivo-Yankov:
**Description**:

**Related issue(s)**:

Fixes #2665 

**Notes for reviewer**:

**Checklist**


georgi-l95 (dismissed):
LGTM

packages/relay/src/lib/eth.ts
```diff
@@ -276,6 +283,20 @@ export class EthImpl implements Eth {
     this.common = new CommonService(mirrorNodeClient, logger, cacheService);
     this.filterServiceImpl = new FilterService(mirrorNodeClient, logger, cacheService, this.common);
     this.debugServiceImpl = new DebugService(mirrorNodeClient, logger, this.common);
+
```
konstantinabl, line 286: Do we need this empty line here?

packages/relay/src/lib/eth.ts
```diff
@@ -277,6 +284,19 @@ export class EthImpl implements Eth {
     this.common = new CommonService(mirrorNodeClient, logger, cacheService);
     this.filterServiceImpl = new FilterService(mirrorNodeClient, logger, cacheService, this.common);
     this.debugServiceImpl = new DebugService(mirrorNodeClient, logger, this.common);
+    this.emptyTrieRoot = EthImpl.emptyHex;
```
victor-yanev, line 287: If we do this we can get rid of the `getEmptyTrieRoot()` function and directly use `EthImpl.emptyTrieRoot`:
```suggestion
    this.emptyTrieRoot = `0x${toHexString(new Trie({}))}`;
```

victor-yanev: I checked the source code of `ethereumjs` and it seems that the reason this `create` method is `async` is to await the connection to the DB, if any is passed through the arguments, e.g., `await Trie.create({ db: new MapDB() })`. 

But in this case, we want to use only the `EMPTY_TRIE_ROOT` constant that is inside of it, so we have no reason to call `Trie.create({})` and overcomplicate the code with this asynchronous method, instead of just initializing it through the constructor `new Trie()`.

victor-yanev: If no arguments are passed, there should be no difference between doing `await Trie.create()` and `new Trie()`

victor-yanev (dismissed):
LGTM

packages/relay/src/formatters.ts
```diff
@@ -260,6 +260,15 @@ const toNullIfEmptyHex = (value: string): string | null => {
...
 
+const toHexString = (byteArray) => {
+  if (typeof byteArray !== 'object') {
+    byteArray = Buffer.from(byteArray?.toString() ?? '', 'hex');
+  }
+
+  const encoded = Buffer.from(byteArray, 'utf8').toString('hex');
+  return encoded;
+};
+
```
quiet-node, line 271: seems like this method, toHexString, is missing UTs

packages/relay/src/formatters.ts
```diff
@@ -260,6 +260,15 @@ const toNullIfEmptyHex = (value: string): string | null => {
   return value === EMPTY_HEX ? null : value;
 };
 
+const toHexString = (byteArray) => {
```
quiet-node, line 263: could you add types for byteArray?

packages/relay/src/lib/eth.ts
```diff
@@ -251,6 +252,12 @@ export class EthImpl implements Eth {
    */
   private readonly debugServiceImpl: DebugService;
 
+  /**
+   * A 32 byte hex hash of an empty trie root
+   * @private
+   */
+  public readonly emptyTrieRoot: string;
```
quiet-node, line 259: doc says @private but declared as pubic

packages/relay/src/lib/eth.ts
```diff
@@ -56,6 +57,7 @@ import { IDebugService } from './services/debugService/IDebugService';
 import { DebugService } from './services/debugService';
 import { IFeeHistory } from './types/IFeeHistory';
 import { ITransactionReceipt } from './types/ITransactionReceipt';
+import { Trie } from '@ethereumjs/trie';
```
quiet-node, line 60: My editor says it cannot find this module

image

packages/relay/tests/lib/eth/eth-config.ts
```diff
@@ -572,6 +572,9 @@ export const DEFAULT_LOGS = {
 export const DEFAULT_ETH_GET_BLOCK_BY_LOGS = {
   logs: [DEFAULT_LOGS.logs[0], DEFAULT_LOGS.logs[1]],
 };
+
+export const DEFAULT_ROOT_HASH = '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421';
```
quiet-node, line 576: Q: Where is this root hash from? Is there a document referencing this root hash? If so, please add a comment with the link to the document above it. The hash seems a bit arbitrary and random to me.

quiet-node (changes requested):

packages/relay/src/lib/eth.ts
Ivo-Yankov: You need to install the Trie library with `npm install`

packages/relay/tests/lib/eth/eth-config.ts
Ivo-Yankov: Added a short description

quiet-node (approved):
LGTM

natanasow (approved):
LGTM

PR #2791, opened 2024-08-05
refactor: Validation logic in the `server` package

### LINKED ISSUE #2790: Refactor validation logic in the server package
### Problem

Currently the logic in the server validator is really tangled up and hard to maintain, also there are several bugs inside which need to be addressed.

### Solution

- Add typing to schema definitions
- Simplify the logic in the utils.ts file
- Give better control of the person describing how the validator should work inside the object schema

### Alternatives

_No response_

victor-yanev:
**Description**:

Currently the logic in the server validator is really tangled up and hard to maintain, also there are several bugs inside which need to be addressed.

This PR:
- Adds typing to object schema definition, method validation, and type validation
- Simplifies the logic in the `utils.ts` file
- Gives better control of the person describing how the validator should work inside the object schema

**Related issue(s)**:

Fixes #2790 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/src/validator/methods.ts
+++ b/packages/server/src/validator/methods.ts
@@ -1,4 +1,33 @@
-export const METHODS = {
+/*-
+ *
+ * Hedera JSON RPC Relay
+ *
+ * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+export type IMethodValidation = {
+  [index: number]: IMethodParamValidation;
+};
+
+export type IMethodParamValidation = {
+  type: string;
+  required?: boolean;
+};
+
+export const METHODS: { [key: string]: IMethodValidation } = {
   eth_estimateGas: {
     0: {
       type: 'transaction',
@@ -96,7 +125,7 @@ export const METHODS = {
       type: 'blockNumber',
     },
     2: {
-      type: ['array'],
+      type: 'array',
     },
   },
   eth_getBlockTransactionCountByHash: {
--- a/packages/server/src/validator/objectTypes.ts
+++ b/packages/server/src/validator/objectTypes.ts
@@ -1,109 +1,155 @@
+/*-
+ *
+ * Hedera JSON RPC Relay
+ *
+ * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
 import { Validator } from '.';
 import { predefined } from '@hashgraph/json-rpc-relay';
 
-export const OBJECTS_VALIDATIONS = {
+export type IObjectSchema = {
+  failOnEmpty?: boolean;
+  properties: {
+    [key: string]: {
+      type: string;
+      nullable: boolean;
+      required?: boolean;
+    };
+  };
+};
+
+export interface IObjectValidation {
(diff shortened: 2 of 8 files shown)


georgi-l95 (approved):
LGTM

georgi-l95 (approved):

georgi-l95 (approved):

natanasow (dismissed):
LGTM

packages/server/src/validator/methods.ts
```diff
@@ -1,4 +1,33 @@
...
+ */
+
+export type IMethodValidation = {
+  [index: number]: IMethodParamValidation;
+};
+
+export type IMethodParamValidation = {
+  type: string;
+  required?: boolean;
+};
```
quiet-node, line 28: Let's create a new `types` folder for the `server` package and then host types and interfaces there!

packages/server/src/validator/objectTypes.ts
```diff
@@ -1,109 +1,155 @@
...
+      nullable: boolean;
+      required?: boolean;
+    };
+  };
+};
+
+export interface IObjectValidation {
+  validate(): boolean;
+  name(): string;
+}
```
quiet-node, line 38: Let's create a new `types` folder for the `server` package and then host types and interfaces there!

packages/server/src/validator/types.ts
```diff
@@ -1,7 +1,32 @@
+/*-
```
quiet-node, line 1: Let's create a new `types` folder for the `server` package and then host types and interfaces there!

quiet-node (changes requested):

packages/server/src/validator/objectTypes.ts
```diff
@@ -1,109 +1,140 @@
...
 import { predefined } from '@hashgraph/json-rpc-relay';
+import { IObjectSchema, IObjectValidation } from '../types/validator/objectTypes';
 
-export const OBJECTS_VALIDATIONS = {
+export const OBJECTS_VALIDATIONS: { [key: string]: IObjectSchema } = {
   blockHashObject: {
-    blockHash: {
-      type: 'blockHash',
-      nullable: false,
+    properties: {
```
quiet-node, line 27: QQ: why do we want to wrap these in an extra nested object?

packages/server/src/validator/types.ts
```diff
@@ -1,7 +1,28 @@
...
+ * limitations under the License.
+ *
+ */
+
 import * as Constants from './constants';
 import { Validator } from '.';
+import { ITypeValidation } from '../types/validator/types';
 
-export const TYPES = {
+export const TYPES: { [key: string]: ITypeValidation } = {
```
quiet-node, line 25: Just a small nit: perhaps we could come up with a more descriptive name for this object instead of just `TYPES`, as that might be a bit too generic. Maybe something related to the Relay or the validation process—or well.. we could just keep it as `TYPES` I guess, lol

packages/server/src/types/validator/methods.ts
```diff
@@ -0,0 +1,28 @@
+/*-
```
quiet-node, line 1: Perhaps instead of having a single folder (`validator`) with multiple type files, we could consolidate them into one file (`validator.ts`) that hosts all the 'validator-related' types.

Just a thought, but as I mentioned, since these types are all related to the validator logic, it might be cleaner to keep things in one place.

packages/server/src/validator/utils.ts
```diff
@@ -1,10 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
 import { Validator } from '.';
 import { JsonRpcError, predefined } from '@hashgraph/json-rpc-relay';
+import { IObjectSchema, IObjectValidation } from '../types/validator/objectTypes';
+import { IMethodParamSchema } from '../types/validator/methods';
+import { ITypeValidation } from '../types/validator/types';
```
quiet-node, line 25: if we decide to have multiple files for the validator folder, we should include these new types into an index.ts file at the root of the `types` folder so the import statements can be a little cleaner

packages/server/src/validator/objectTypes.ts
victor-yanev: Because we want to also have the ability to configure validations for the root object (e.g., `failOnEmpty` or `failOnUnknownProperties`)

packages/server/src/validator/types.ts
victor-yanev: If you have an idea of a better name, I’m open to hearing it 😄

packages/server/src/types/validator/methods.ts
victor-yanev: We already have an `index.ts` file inside the validator folder, which servers this purpose

packages/server/src/validator/utils.ts
victor-yanev: I agree but maybe not on the root level, but rather on `types/validator` in order to have better separation in the imports

packages/server/src/validator/objectTypes.ts
quiet-node: fair thanks for confirming!

packages/server/src/validator/utils.ts
quiet-node: yeah that's fair

quiet-node (dismissed):

quiet-node (dismissed):

natanasow (approved):

quiet-node (approved):

Issue #2821, opened 2024-08-05
Transaction reverted with encoded error but string was expected


mgarbs:
#### Description 
The Axelar team is encountering an issue where they are expecting a revert message to be a string but it is coming back encoded. Below are the details of the scenario:  
- **Transaction Details**: [Transaction on Hashscan](https://hashscan.io/testnet/transaction/1722600107.393768552) 
- **Error Message**: `AssertionError: Expected transaction to be reverted with custom error 'NativeCurrencyNotAccepted', but it didn't revert` 
- **Assertion Location**: [AxelarGateway.js#L159](https://github.com/axelarnetwork/axelar-cgp-solidity/blob/hedera/test/test/AxelarGateway.js#L159) 
- **Contract Code**: The call that actually reverts is located at [AxelarGatewayProxy.sol#L54](https://github.com/axelarnetwork/axelar-cgp-solidity/blob/hedera/test/contracts/AxelarGatewayProxy.sol#L54)  
#### Steps to Reproduce 
1. Execute the transaction linked above on the Hedera testnet. 
2. Observe that the transaction reverts with the encoded version of the expected error string. 
3. Check the assertion in the `AxelarGateway.js` test file.  

#### Environment 
- **Hedera Network**: Testnet 
- **Repository**: [axelarnetwork/axelar-cgp-solidity](https://github.com/axelarnetwork/axelar-cgp-solidity)

acuarica:
We have detected the root cause of the issue. Essentially, the `eth_estimateGas` method call does not properly handle HBAR transfers when the `to` address is a contract that `revert`s execution.

### How to Reproduce

Create an empty Hardhat project, set up `hardhat.config.js` to use a Local Node network named `local`, and use the following contract and test case

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.24;

contract RevertOnReceive {
    error NativeCurrencyNotAccepted();

    receive() external payable {
        revert NativeCurrencyNotAccepted();
    }
}
```

```js
const { expect } = require("chai");
const { ethers } = require("hardhat");

describe("RevertOnReceive", function () {
  it("should revert with custom error on `receive`", async function () {
    const [signer] = await ethers.getSigners();
    const ContractFactory = await ethers.getContractFactory("RevertOnReceive");
    const contract = await ContractFactory.deploy();

    await expect(signer.sendTransaction({
      to: await contract.getAddress(),
      value: 10_000_000_001,
[...]

PR #2834, opened 2024-08-16
fix: `eth_estimateGas` opts to fallback estimation for contract revert

### LINKED ISSUE #2830: Calling eth_estimateGas opts to fallback estimation when a contract revert is executed
### Description

Many frameworks depend on `eth_call` and `eth_estimateGas` after sending a transaction which reverts to retrieve the revert reason for their sent transaction.

Currently, calling `eth_estimateGas` on a function which reverts against our relay will return the following error from the mirror-node:
```
Error raised while fetching estimateGas from mirror-node: {"detail":"","data":"0x858d70bd","statusCode":400}
```

which contains the revert reason in the `"data"` field of the thrown error.

What we currently do is treat it as a regular error from the mirror node and we opt to the fallback estimations instead of returning the revert reason to the user to give him context of what is happening (as other ETH clients would for `eth_estimateGas`)

### Steps to reproduce

1. send a transaction which calls a reverting contract function through `eth_sendRawTransaction`
2. send an `eth_estimateGas` request with the unsigned transaction body and the block number from the receipt of the executed transaction

**Expected:**
- should return the error code for contract revert and the ABI-encoded revert reason from the contract call in the `"data"` field of the error. 

Example with custom error thrown in solidity:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32000,
        "message": "Execution reverted",
        "data": "0x858d70bd"
    }
}
```

If the returned reason is encoded as a generic `Error(string)`(prefixed with `0x08c379a`) and not a custom error, we can decode it and append it to the message to be consistent with https://github.com/hyperledger/besu/pull/5706:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32000,
        "message": "Execution reverted: ERC20: transfer amount exceeds balance",
        "data": "0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000002645524332303a207472616e7366657220616d6f756e7420657863656564732062616c616e63650000000000000000000000000000000000000000000000000000"
    }
}
```

**Actual:**
- ignores the error from mirror node (which contains the revert reason) and returns a hardcoded fallback gas estimation

### Additional context

```
TRACE (relay-eth/85 on 5f4099310571): [Request ID: 9ebee5c9-3cae-4a92-b4a8-76e73666c865] estimateGas(transaction={"from":"0x17b2b8c63fa35402088640e426c6709a254c7ffb","to":"0x498825bf0af4ffaa3f2cc4ea176d6df7fed8bde2","gas":"0x61a80","gasPrice":"0x0","value":"0x0","data":"0xf8a8fd6d"}, _blockParam=undefined)
DEBUG (mirror-node/85 on 5f4099310571): [Request ID: 9ebee5c9-3cae-4a92-b4a8-76e73666c865] [POST] contracts/call Contract Revert: ( StatusCode: '400', StatusText: '', Detail: 'undefined',Data: '{"_status":{"messages":[{"message":"CONTRACT_REVERT_EXECUTED","detail":"","data":"0x858d70bd"}]}}')
ERROR (relay-eth/85 on 5f4099310571): [Request ID: 9ebee5c9-3cae-4a92-b4a8-76e73666c865] Error raised while fetching estimateGas from mirror-node: {"detail":"","data":"0x858d70bd","statusCode":400}
WARN (relay-eth/85 on 5f4099310571): [Request ID: 9ebee5c9-3cae-4a92-b4a8-76e73666c865] Returning predefined gas for contract call: 0x7a120
INFO (rpc-server/85 on 5f4099310571): [Request ID: 9ebee5c9-3cae-4a92-b4a8-76e73666c865] [POST]: eth_estimateGas 200 13 ms
```

### Hedera network

_No response_

### Version

v0.54.0

### Operating system

None

victor-yanev:
**Description**:

This PR modifies the `EthImpl` class to improve error handling for gas estimation.
* Add logic to return `predefined.CONTRACT_REVERT` JSON-RPC error with the revert reason in case of contract revert
* Extend `predefined.CONTRACT_REVERT` to cover more edge cases when decoding error messages
* Extend tests for `decodeErrorMessage` and `predefined.CONTRACT_REVERT`

**Related issue(s)**:

Fixes #2830 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -601,12 +601,17 @@ export class EthImpl implements Eth {
         this.logger.info(`${requestIdPrefix} Returning gas: ${response.result}`);
         return prepend0x(trimPrecedingZeros(response.result));
       } else {
+        this.logger.error(`${requestIdPrefix} No gas estimate returned from mirror-node: ${JSON.stringify(response)}`);
         return this.predefinedGasForTransaction(transaction, requestIdPrefix);
       }
     } catch (e: any) {
       this.logger.error(
         `${requestIdPrefix} Error raised while fetching estimateGas from mirror-node: ${JSON.stringify(e)}`,
       );
+      // in case of contract revert, we don't want to return a predefined gas but the actual error with the reason
+      if (e instanceof MirrorNodeClientError && e.isContractReverted()) {
+        return predefined.CONTRACT_REVERT(e.detail || e.message, e.data);
+      }
       return this.predefinedGasForTransaction(transaction, requestIdPrefix, e);
     }
   }
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -33,12 +33,20 @@ export class JsonRpcError {
 }
 
 export const predefined = {
-  CONTRACT_REVERT: (errorMessage?: string, data: string = '') =>
-    new JsonRpcError({
+  CONTRACT_REVERT: (errorMessage?: string, data: string = '') => {
+    let message: string;
+    if (errorMessage?.length) {
+      message = `execution reverted: ${decodeErrorMessage(errorMessage)}`;
+    } else {
+      const decodedData = decodeErrorMessage(data);
+      message = decodedData.length ? `execution reverted: ${decodedData}` : 'execution reverted';
+    }
+    return new JsonRpcError({
       code: 3,
-      message: `execution reverted: ${decodeErrorMessage(errorMessage)}`,
-      data: data,
-    }),
+      message,
+      data,
+    });
+  },
   GAS_LIMIT_TOO_HIGH: (gasLimit, maxGas) =>
     new JsonRpcError({
       code: -32005,
--- a/packages/relay/tests/lib/errors/JsonRpcError.spec.ts
+++ b/packages/relay/tests/lib/errors/JsonRpcError.spec.ts
@@ -19,7 +19,8 @@
  */
 
 import { expect } from 'chai';
-import { JsonRpcError } from '../../../src/lib/errors/JsonRpcError';
+import { JsonRpcError, predefined } from '../../../src';
+import { AbiCoder, keccak256 } from 'ethers';
 
 describe('Errors', () => {
   describe('JsonRpcError', () => {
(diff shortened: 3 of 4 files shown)


acuarica (changes requested):
Shouldn't we include some tests for this new scenario?

victor-yanev:
> Shouldn't we include some tests for this new scenario?

Good point, I will extend the tests for eth_estimateGas

packages/relay/src/lib/eth.ts
```diff
@@ -175,6 +175,10 @@ export class EthImpl implements Eth {
   private readonly estimateGasThrows = process.env.ESTIMATE_GAS_THROWS
     ? process.env.ESTIMATE_GAS_THROWS === 'true'
     : true;
+  private readonly estimateGasIgnoredErrors = [
+    MirrorNodeClientError.messages.INVALID_HEX,
+    MirrorNodeClientError.messages.INVALID_FROM,
+  ];
```
acuarica, line 181: Should we move this to the inner scope? _i.e._, inside `estimateGas`?

packages/relay/src/lib/eth.ts
```diff
@@ -601,12 +605,22 @@ export class EthImpl implements Eth {
...
     } catch (e: any) {
       this.logger.error(
         `${requestIdPrefix} Error raised while fetching estimateGas from mirror-node: ${JSON.stringify(e)}`,
       );
+      // in case of contract revert, we don't want to return a predefined gas but the actual error with the reason
+      if (
+        this.estimateGasThrows &&
+        e instanceof MirrorNodeClientError &&
+        e.isContractReverted() &&
+        !this.estimateGasIgnoredErrors.includes(e.message)
```
acuarica, line 620: If `e.isContractReverted()` is `true`, can the error message be `from field must not be empty`? Just to make sure we have right guard here.

packages/relay/src/lib/eth.ts
```diff
@@ -601,12 +605,22 @@ export class EthImpl implements Eth {
...
+        this.logger.error(`${requestIdPrefix} No gas estimate returned from mirror-node: ${JSON.stringify(response)}`);
         return this.predefinedGasForTransaction(transaction, requestIdPrefix);
       }
     } catch (e: any) {
       this.logger.error(
         `${requestIdPrefix} Error raised while fetching estimateGas from mirror-node: ${JSON.stringify(e)}`,
       );
+      // in case of contract revert, we don't want to return a predefined gas but the actual error with the reason
+      if (
+        this.estimateGasThrows &&
```
acuarica, line 617: I understand this is a global setting, but should we honor it in this case? I mean, should we throw regardless of this setting in this case?

acuarica (commented):
looking good, left some questions

acuarica:
Nit: I think we should the remove the `"Fixes #2821"` from this PR's description, and replace it by something like `"Related to #2821"`. The reason is that we wouldn't want to close #2821 when we merge this PR, but instead close it when the Axelar team can effectively confirm the issue is solved, which is when we make a new release containing this fix.

packages/relay/src/lib/eth.ts
victor-yanev: It would be weird not to honour it,`estimateGasThrows` defaults to `true` if the environment variable is `undefined`, so I don’t see why we would want to bypass this if it’s going to be true in most cases. 

If the user has set this variable explicitly to `false` in his .env file, I believe we should also honour it here, what is your opinion?

packages/relay/src/lib/eth.ts
victor-yanev: Hmmm, good point. 

It is a matter of trading heap space for execution time of the `estimateGas` method since it will have to initialise this array every time.

acuarica: In this case, I believe both heap space and time would be negligible. I was suggesting this because (at least to me) it's more clear when you have both definition and usage together. Otherwise, at first glance it may look like `estimateGasIgnoredErrors` may be used somewhere else, when it's not the case.

Nevertheless, just a suggestion.

packages/relay/src/lib/eth.ts
acuarica: This is tricky, not sure. But if the corresponding `eth_call` fails, shouldn't the `eth_estimateGas` fail as well regardless of this setting?

packages/relay/src/lib/eth.ts
acuarica: Oh, I see, thanks for the clarification.

packages/relay/src/lib/eth.ts
victor-yanev: I agree with you, readability is prioritized when the difference in complexity is negligible.

I will move the declaration inside the method.

packages/relay/src/lib/eth.ts
victor-yanev: I added a new method inside `MirrorNodeClientError`:
```typescript
  public isContractRevertOpcodeExecuted() {
    return this.message === MirrorNodeClientError.messages.CONTRACT_REVERT_EXECUTED;
  }
```

which is taken from [here](https://github.com/hashgraph/hedera-sdk-js/blob/507fdad9a4ef399b929a8dd1a167c8619109f8b5/src/Status.js#L111) and used that instead of checking the message inside the `this.estimateGasIgnoredErrors` array. 

There are too many validations done inside the mirror node and we need to add each type of validation inside this array in order to avoid bugs - I saw some dApp tests were failing because they were passing wrong transaction bodies to `eth_estimateGas` and they were receiving an error instead of getting the predefined gas.

packages/relay/src/lib/eth.ts
victor-yanev: Hmm, it's kinda tricky because we rely on the fallback estimations for some tools to work properly right now 😆

victor-yanev: See https://github.com/hashgraph/hedera-json-rpc-relay/pull/2834#discussion_r1720163125

natanasow (approved):
LGTM

konstantinabl (approved):

PR #2841, opened 2024-08-19
fix: error thrown when setting value in redis cache

### LINKED ISSUE #2838: Error thrown from Redis when setting value with TTL
### Description

This error is appearing multiple times in the logs of multiple workflows in the CI, right after doing the `getAccount` query from the mirror node.

### Steps to reproduce

1. Run the App Tests
2. Observe the logs
3. Error occurs after the cache service tries to cache the result from `eth_getTransactionCount` request

### Additional context

```bash
[2024-08-19 09:10:10.500 +0000] ERROR (cache-service/7325): [Request ID: 82112b27-7f93-4007-86cf-58b806d266be] Error occurred while setting the cache to Redis. Fallback to internal cache. Error is: Error: ERR value is not an integer or out of range
```

image

### Hedera network

_No response_

### Version

v0.54.0

### Operating system

None

victor-yanev:
**Description**:

```bash
ERROR (cache-service/7325): Error occurred while setting the cache to Redis. Fallback to internal cache. Error is: Error: ERR value is not an integer or out of range
```

This error is appearing multiple times in the logs of multiple workflows in the CI, right after doing the getAccount query from the mirror node.

After doing some investigation, it looks like when we transform the passed TTL to seconds, it was possible that we get a floating point number (`ETH_GET_TRANSACTION_COUNT_CACHE_TTL: 500`) => which resolves to `resolvedTtl = 0.5`
```typescript
const resolvedTtl = (ttl ?? this.options.ttl) / 1000; // convert to seconds
```

And the `setEx` method of Redis expects an integer, hence why this error is thrown.

This PR:
- modifies the `RedisCache#set` method to use `set` (with `PX` option which allows us to directly pass milliseconds as argument for TTL) instead of using `setEx` to avoid this issue.

**Related issue(s)**:

Fixes #2838

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/cache/redisCache.ts
+++ b/packages/relay/src/lib/clients/cache/redisCache.ts
@@ -150,9 +150,9 @@ export class RedisCache implements IRedisCacheClient {
   ): Promise<void> {
     const client = await this.getConnectedClient();
     const serializedValue = JSON.stringify(value);
-    const resolvedTtl = (ttl ?? this.options.ttl) / 1000; // convert to seconds
+    const resolvedTtl = ttl ?? this.options.ttl; // in milliseconds
 
-    await client.setEx(key, resolvedTtl, serializedValue);
+    await client.set(key, serializedValue, { PX: resolvedTtl });
     this.logger.trace(`${requestIdPrefix} caching ${key}: ${serializedValue} on ${callingMethod} for ${resolvedTtl} s`);
     // TODO: add metrics
   }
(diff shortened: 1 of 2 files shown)


isavov (dismissed):

konstantinabl (dismissed):

konstantinabl:
Nice catch with the tests! 
LGTM

konstantinabl (approved):

konstantinabl (approved):

konstantinabl (approved):

PR #2843, opened 2024-08-19
fix: Bug due to validations of `eth_call`

### LINKED ISSUE #2827: Bug due to validations of eth_call
### Description

Currently the validations we do for `eth_call` are stopping us from operating with our relay as we would with any other ethereum client through popular tools like web3 and ethers

### Steps to reproduce

1. send a transaction which calls a contract function through `eth_sendRawTransaction`
2. call `eth_getTransactionByHash` to get the transaction object for the executed call
3. pass the resulting transaction object to `eth_call`

Expected:
- we should receive a successful response from the `eth_call` request

Actual:
- the relay throws an error: `Invalid parameter 'blockHash' for TransactionObject: Unknown parameter`

### Additional context

```bash
ERROR (rpc-server/85 on 5f4099310571): [Request ID: ed92e528-f8b1-41be-a61d-29ecd76d67de] Invalid parameter 'blockHash' for TransactionObject: Unknown parameter
INFO (rpc-server/85 on 5f4099310571): [Request ID: ed92e528-f8b1-41be-a61d-29ecd76d67de] [POST]: eth_call 400 0 ms
```

### Hedera network

_No response_

### Version

v0.54.0

### Operating system

None

victor-yanev:
**Description**:

This PR modifies the object validation logic in order to support the deletion of unknown properties. 
* Adds check for `deleteUnknownProperties` in object schema
* Calls `deleteUnknownProperties` method if the schema property is set

**Related issue(s)**:

Fixes #2827

**Checklist**


ebadiere (approved):
LG.

acuarica (approved):
lg

Issue #2847, opened 2024-08-19
Get TransactionRecord via consensus node in TransactionService can be improved for better performance


quiet-node:
### Problem

Currently, the `TransactionService.getTransactionStatusAndMetrics()` method, when `GET_RECORD_DEFAULT_TO_CONSENSUS_NODE` is set to true, queries the consensus node to retrieve the operator's balance information to calculate the charged amount before and after executing `TransactionRecordQuery`. To ensure accurate account balances, a 200ms timeout is used to allow the balances to update properly. This results in a total of 400ms (before and after) for each request to calculate the difference in account balance. Depending on the size of the `callData`, each client call (e.g., `eth_sendRawTransaction`) can involve multiple requests (10+ or 20+), significantly increasing the total wait time just to calculate the balance difference.

This issue has been raised in the SDK repository under [this ticket](https://github.com/hashgraph/hedera-sdk-js/issues/2460), requesting a new API that performs the calculation natively.

### Solution

Wait for the new API to be included in the SDK and implement it in the `TransactionService` to improve performance.

PR #2848, opened 2024-08-19
fix: temporarily increased the timeout for the SdkClient test suite to 90000ms to reduce flakiness

### LINKED ISSUE #2846: Flaky HBarLimiter class due to time out issue
### Problem

The `should execute submitEthereumTransaction add expenses to limiter for large transaction data` test suite is not stable and usually throw timeout error. Many attempts that try to bump the timeout but the issue didn't seem to be resolved.

### Solution

Shorten down the size of the mocked call data

### Alternatives

_No response_

quiet-node:
**Description**:
This PR temporarily bumped timeout for the sdkClient test suite to 90000ms to reduce flakyness

Fixes #2846 

**Notes for reviewer**:
This is a temporarily solution. The issue lies in the TransactionService class that takes long time to retrieve balance information of the operator. More information is captured in this ticket https://github.com/hashgraph/hedera-json-rpc-relay/issues/2847.

**Checklist**


victor-yanev (changes requested):
Instead of extending the timeout from 45 seconds to 90 seconds to address the flakiness, I think we should focus on the fact that this transaction used to complete in less than 45 seconds, and now it’s taking twice as long.

The delay is especially evident when the relay executes multiple file appends and has to retry operator balance queries during hbar limiter metrics capture. To improve user experience and avoid unnecessary slowdowns, I suggest we handle these metrics asynchronously and address the actual problem which we have - that this transaction no longer executes in the time frame it used to.

quiet-node:
> Instead of extending the timeout from 45 seconds to 90 seconds to address the flakiness, I think we should focus on the fact that this transaction used to complete in less than 45 seconds, and now it’s taking twice as long.
> 
> The delay is especially evident when the relay executes multiple file appends and has to retry operator balance queries during hbar limiter metrics capture. To improve user experience and avoid unnecessary slowdowns, I suggest we handle these metrics asynchronously and address the actual problem which we have - that this transaction no longer executes in the time frame it used to.

Hey @victor-yanev, thanks for the comment! I also noticed that the main issue behind this timeout problem stems from the `TransactionService` class, as I mentioned in the "notes for reviewer" section. That’s why I labeled it as a "temporary" solution to help unblock the PRs for now. But yes, I definitely agree with you and appreciate the great suggestion—I’ll look into making the account balance queries asynchronous to improve the situation.

victor-yanev (dismissed):

packages/relay/src/lib/services/transactionService/transactionService.ts
```diff
@@ -113,21 +106,16 @@ export default class TransactionService {
...
-        const operatorBalanceAfter = await this.sdkClient.getBalanceInTinyBars(
-          operatorAccountId,
-          callerName,
-          requestId,
-          200,
+        // calculate transactionRecord fee
+        const hbarToTinybar = Hbar.from(1, HbarUnit.Hbar).toTinybars().toNumber();
+        const exchangeRateInCents = transactionReceipt.exchangeRate!.exchangeRateInCents;
+        txRecordChargeAmount = Math.round(
+          (Constants.TX_RECORD_QUERY_COST_IN_CENTS / exchangeRateInCents) * hbarToTinybar,
```
ebadiere, line 113: Given that it's already a method with 89 lines, would it not make sense to break out the calculation into its own function?  I could be more readable.
```
function calculateTxRecordChargeAmount(exchangeRate: ExchangeRate): number {
    const exchangeRateInCents = exchangeRate.exchangeRateInCents;
    return Math.round((TX_RECORD_QUERY_COST_IN_CENTS / exchangeRateInCents) * HBAR_TO_TINYBAR);
}
```
```
// Usage
const txRecordChargeAmount = **calculateTxRecordChargeAmount(transactionReceipt.exchangeRate!);**
```

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -2110,6 +2110,7 @@ describe('SdkClient', async function () {
     const fileCreateFee = Number(process.env.HOT_FIX_FILE_CREATE_FEE || 100000000); // 1 hbar
     const fileDeleteFee = Number(process.env.HOT_FIX_FILE_DELETE_FEE || 11000000); // 0.11 hbar
     const fileAppendFee = Number(process.env.HOT_FIX_FILE_APPEND_FEE || 120000000); // 1.2 hbar
+    const transactionRecordFee = 83333; // the precalcuated fee for transaction record queries
```
ebadiere, line 2113: Nit.  Maybe add the URL to the docs describing the fee calculation?

packages/relay/src/lib/services/transactionService/transactionService.ts
quiet-node: Great point! Updated with a new refactored function.

packages/relay/tests/lib/sdkClient.spec.ts
quiet-node: Unfortunately, there isn't a URL to a document describing the calculation. It's based on a constant, `TX_RECORD_QUERY_COST_IN_CENTS = 0.01` cents, for which I’ve included a link to the fees calculator website.

packages/relay/tests/lib/services/transactionService/transactionService.spec.ts
```diff
@@ -208,4 +219,24 @@ describe('Transaction Service', function () {
...
+      const expectedTxRecordAmount = Math.round((txQueryCostInCents / mockedExchangeRateIncents) * hbarToTinybar);
+
+      const mockedExchangeRate = {
+        hbars: 30000,
+        cents: 164330,
+        expirationTime: new Date(),
+        exchangeRateInCents: mockedExchangeRateIncents,
+      } as ExchangeRate;
+
+      const txRecordChargedAmount = transactionService.calculateTxRecordChargeAmount(mockedExchangeRate);
```
victor-yanev, line 237: Please, duplicate the calculation here instead of using the actual logic. Otherwise, introducing a bug inside the actual method would be reflected here and wont be caught by the tests.

packages/relay/tests/lib/services/transactionService/transactionService.spec.ts
```diff
@@ -208,4 +219,24 @@ describe('Transaction Service', function () {
       );
     });
   });
+
+  describe('calculateTxRecordChargeAmount', () => {
+    it('Should execute calculateTxRecordChargeAmount() to get the charge amount of transaction record', () => {
+      const mockedExchangeRateIncents = 12;
+      const txQueryCostInCents = constants.TX_RECORD_QUERY_COST_IN_CENTS;
+      const hbarToTinybar = Hbar.from(1, HbarUnit.Hbar).toTinybars().toNumber();
+      const expectedTxRecordAmount = Math.round((txQueryCostInCents / mockedExchangeRateIncents) * hbarToTinybar);
```
victor-yanev, line 228: We can use the helper method from `/relay/tests/helpers`

ebadiere (approved):

packages/relay/tests/lib/services/transactionService/transactionService.spec.ts
quiet-node: Good catch! Thanks!

ebadiere (approved):

victor-yanev (approved):

PR #2860, opened 2024-08-21
fix: The relay relies on fallback gas estimations instead of properly sending requests to mirror node

### LINKED ISSUE #2852: The relay relies on fallback gas estimations instead of properly sending requests to mirror node
### Description

Currently the JSON-RPC relay relies too much on the logic for fallback gas estimations instead of properly preparing and sending `eth_estimateGas` requests to the mirror node. At the same time, we want to rethrow errors from the mirror node in case of contract reverts to give the user feedback about failed contract calls/creations, because popular tools (e.g., hardhat) rely on the `eth_estimateGas` method to retrieve a revert reason for failed transactions. 

With the current logic which relies on the fallback estimations, this is really hard to implement, because the only way to do it is to compare the error message we receive from the mirror node with every possible error that is possible to be thrown from contract reverts. Obviously, this is not the way we want this to work, 

### Steps to reproduce

We can observe in the logs, many requests to the estimate gas endpoint in mirror node fail the validations and the relay opts to the fallback estimations (example logs are given below).

### Additional context

Example logs:
```bash
[2024-08-20 18:25:24.145 +0000] DEBUG (rpc-server/14667 on hedera-json-rpc-relay-linux-large-nvtx2-4sp7f): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] Validating method parameters for eth_estimateGas, params: [{"value":"0x18cef949392f5e000","to":"0x00000000000000000000000000000000000003ea"}]
[2024-08-20 18:25:24.145 +0000] TRACE (relay-eth/14667 on hedera-json-rpc-relay-linux-large-nvtx2-4sp7f): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] estimateGas(transaction={"value":"0x18cef949392f5e000","to":"0x00000000000000000000000000000000000003ea"}, _blockParam=undefined)
[2024-08-20 18:25:24.154 +0000] DEBUG (mirror-node/14667 on hedera-json-rpc-relay-linux-large-nvtx2-4sp7f): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] [POST] contracts/call Contract Revert: ( StatusCode: '400', StatusText: '', Detail: 'undefined',Data: '{"_status":{"messages":[{"message":"from field must not be empty","detail":"","data":""}]}}')
[2024-08-20 18:25:24.154 +0000] ERROR (relay-eth/14667 on hedera-json-rpc-relay-linux-large-nvtx2-4sp7f): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] Error raised while fetching estimateGas from mirror-node: {"detail":"","data":"","statusCode":400}
[2024-08-20 18:25:24.155 +0000] TRACE (redisCache/14667 on hedera-json-rpc-relay-linux-large-nvtx2-4sp7f): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] returning cached value account_0x00000000000000000000000000000000000003ea:"{\"account\":\"0.0.1002\",\"alias\":null,\"auto_renew_period\":7776000,\"balance\":{\"balance\":798448269546,\"timestamp\":\"1724178111.719078683\",\"tokens\":[{\"token_id\":\"0.0.1041\",\"balance\":99999999990}]},\"created_timestamp\":\"1724178041.158822890\",\"decline_reward\":false,\"deleted\":false,\"ethereum_nonce\":0,\"evm_address\":\"0x00000000000000000000000000000000000003ea\",\"expiry_timestamp\":\"1731954041.158822890\",\"key\":{\"_type\":\"ECDSA_SECP256K1\",\"key\":\"024778141fffe462dbb3817c93e4252b2d6f5f5667fa71f34b84280265c01d408f\"},\"max_automatic_token_associations\":0,\"memo\":\"\",\"pending_reward\":0,\"receiver_sig_required\":false,\"staked_account_id\":null,\"staked_node_id\":null,\"stake_period_start\":null,\"transactions\":[],\"links\":{\"next\":null}}" on eth_estimateGas call
[2024-08-20 18:25:24.155 +0000] WARN (relay-eth/14667 on hedera-json-rpc-relay-linux-large-nvtx2-4sp7f): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] Returning predefined gas for simple transfer: 0x5208

```

### Hedera network

_No response_

### Version

v0.54.2

### Operating system

None

victor-yanev:
**Description**:

Currently the relay relies too much on the logic of fallback gas estimations instead of properly preparing the `ContractCallRequest` body before sending it to the mirror node. 

Also, we want to rethrow any errors from the mirror node in case of contract reverts to give the user feedback about failed contract calls/creations, because popular tools (e.g., hardhat) rely on the `eth_estimateGas` method to retrieve a revert reason for failed transactions.

With the current logic which relies on the fallback estimations in cases of errors thrown from the mirror node, this is really hard to implement. The only way to do it currently is to compare the error message we receive from the mirror node with every possible error that is possible to be thrown from contract reverts. Obviously, this is not the way we want this to work, so this PR makes sure that requests which are sent to the mirror node cover all required validations that they do on their side.

**Additional Context**:

We can observe in the logs, many requests to estimate gas in the mirror node fail the validations and the relay opts to the fallback estimations (example logs are given below).
```bash
DEBUG (rpc-server/14667): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] Validating method parameters for eth_estimateGas, params: [{"value":"0x18cef949392f5e000","to":"0x00000000000000000000000000000000000003ea"}]

TRACE (relay-eth/14667): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] estimateGas(transaction={"value":"0x18cef949392f5e000","to":"0x00000000000000000000000000000000000003ea"}, _blockParam=undefined)

DEBUG (mirror-node/14667): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] [POST] contracts/call Contract Revert: ( StatusCode: '400', StatusText: '', Detail: 'undefined',Data: '{"_status":{"messages":[{"message":"from field must not be empty","detail":"","data":""}]}}')

ERROR (relay-eth/14667): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] Error raised while fetching estimateGas from mirror-node: {"detail":"","data":"","statusCode":400}

TRACE (redisCache/14667): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] returning cached value account_0x00000000000000000000000000000000000003ea:"{\"account\":\"0.0.1002\",\"alias\":null,\"auto_renew_period\":7776000,\"balance\":{\"balance\":798448269546,\"timestamp\":\"1724178111.719078683\",\"tokens\":[{\"token_id\":\"0.0.1041\",\"balance\":99999999990}]},\"created_timestamp\":\"1724178041.158822890\",\"decline_reward\":false,\"deleted\":false,\"ethereum_nonce\":0,\"evm_address\":\"0x00000000000000000000000000000000000003ea\",\"expiry_timestamp\":\"1731954041.158822890\",\"key\":{\"_type\":\"ECDSA_SECP256K1\",\"key\":\"024778141fffe462dbb3817c93e4252b2d6f5f5667fa71f34b84280265c01d408f\"},\"max_automatic_token_associations\":0,\"memo\":\"\",\"pending_reward\":0,\"receiver_sig_required\":false,\"staked_account_id\":null,\"staked_node_id\":null,\"stake_period_start\":null,\"transactions\":[],\"links\":{\"next\":null}}" on eth_estimateGas call

WARN (relay-eth/14667): [Request ID: 8e1d2f0d-5180-4f89-a736-e34c31f425f9] Returning predefined gas for simple transfer: 0x5208
```

**Related issue(s)**:

Fixes #2852 

**Notes for reviewer**:

**Checklist**


ebadiere (commented):
The build is broken.

Ivo-Yankov (dismissed):

packages/relay/src/lib/eth.ts
```diff
@@ -704,17 +704,29 @@ export class EthImpl implements Eth {
...
     }
     if (transaction.gasPrice) {
       transaction.gasPrice = parseInt(transaction.gasPrice.toString());
+    } else {
+      transaction.gasPrice = await this.gasPrice(requestIdPrefix).then((gasPrice) => parseInt(gasPrice));
     }
     if (transaction.gas) {
       transaction.gas = parseInt(transaction.gas.toString());
     }
+    if (!transaction.from && transaction.value && transaction.value > 0) {
```
ebadiere, line 721: Nit.  The warning "Unexpected nullable string value in conditional. Please handle the nullish/empty cases explicitly." applies here.  I know it also applies to the other conditions, but slowly cleaning up code as we work on it, is a good practice.

packages/relay/src/lib/types/IMirrorNode.ts
```diff
@@ -18,6 +18,28 @@
...
+   * Account ID string in the form `shard.realm.num`
+   */
+  account: string;
+  /**
+   * RFC4648 no-padding base32 encoded string of the account's alias.
+   */
+  alias: string;
+  balance?: IAccountBalance;
+  deleted?: boolean;
+  ethereum_nonce?: number;
```
ebadiere, line 32: Should we consider making the nonce non-optional with a default value of zero?  Could adding JSDoc comments help with documentation?

ebadiere (commented):
Looks good.  Just left a couple of questions.

packages/relay/src/lib/types/IMirrorNode.ts
victor-yanev: It’s important to note that interfaces in TypeScript are just for defining the structure of the data, like a blueprint. They don’t actually exist after the code is compiled to JavaScript, so you can't set default values in them.

If we want a default value for the nonce, like zero, we’d need to create a class that implements this interface. Then, we would be able to set the default value in the class's constructor. But I think it’s better to keep the original data that is coming from the mirror node, without modifying it unless we have a good reason to.

As for adding JSDoc comments, that’s a great idea! They’ll definitely help with documentation and make the code easier to understand.

packages/relay/src/lib/eth.ts
victor-yanev: Actually, this warning only applies if you pass the `string` directly into the conditional. Using `!` or `!!` converts the `string` to a `boolean`, which avoids the issue. We can address other conditions over time, but this one is already handled correctly.

PR #2872, opened 2024-08-22
fix: Retry query to mirror node to get correct nonce.

ebadiere:
Retry query to mirror node to get nonce.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -1002,6 +1003,18 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
       );
       expect(nonceAfter).to.be.equal('0x3');
     });
+
+    it('should deploy a contract and get the correct nonce', async function () {
```
Ivo-Yankov, line 1007: Is this test actually testing for the nonce?

Ivo-Yankov (changes requested):
I don't understand the reasoning behind the changes in `eth.ts`. Can you add some comments to describe the nonce problem and the logic.

Is it possible that querying for multiple different accounts can cause problems with `previousAccountNonce`?

Issue #2808, opened 2024-08-09
Headers Timeout Error when deploying large contracts


mgarbs:
**Problem:**
The Axelar team is encountering a consistent HeadersTimeoutError when deploying large contracts. This issue occurs across various RPC providers, including QuickNode, ValidationCloud, the public RPC and when using their own relay.

**Details:**
Despite the timeout error, the contract still deploys successfully.
Example of deployed contract: [contract deploy success](https://hashscan.io/testnet/contract/0.0.4643580)
Example bytecode: [AxelarGateawayData.txt](https://github.com/user-attachments/files/16562185/AxelarGateawayData.txt)

**Impact:**
This issue affects the deployment process by causing unnecessary errors, which could mislead developers into thinking the deployment failed when it really succeeded. Resolving this issue would also allow Axelar to not need custom configs for their rpc which would be more in parity with other evms they have integrated.

quiet-node:
I have tried reproducing this issue on the local Relay pointing to testnet, as well as on the HashIO relay, using the same bytecode from `AxelarGateWayData.txt`. However, I was unable to reproduce the timeout issue on either instance.

![Screenshot 2024-08-14 at 12:24:50 PM](https://github.com/user-attachments/assets/f3511cbb-6c37-4238-9e46-b001afcff11c)

As shown in the screenshot above, it takes around 25 seconds to complete the deployment job, and no timeout error was thrown. Here is an example of the deployed contract: [https://hashscan.io/testnet/transaction/1723656880.276860003](https://hashscan.io/testnet/transaction/1723656880.276860003).

Is it possible for you to show what the log says when a timeout error is thrown? Additionally, a list of the Relay's configuration settings would be helpful to review as well.

natanasow:
Hi @mgarbs, I might know what's happening if you're using hardhat. When we create a contract deployment, the relay under the hood executes 1x`FileCreate` + N*`FileAppend`s which takes some time. After the file (with the contract's content) has been created, the relay passes the main signed Ethereum contract deploy transaction to the services for further execution. If we're waiting for the receipt to be populated in the mirror node, the entire process might take up to 30 seconds against testnet for the bytecode you have posted above.

By default, hardhat-ethers set the `HeadersTimeoutError` to 20 seconds, so if we exceed these seconds, the error is thrown but the request has been already sent for execution and that's why the contract is deployed but the end-user receives an error for timeout.

The flow might be described as follows:
[...]

PR #2873, opened 2024-08-23
feat: added MetricService class to replace TransactionService

### LINKED ISSUE #2847: Get TransactionRecord via consensus node in TransactionService can be improved for better performance

quiet-node:
**Description**:
- Added a new `MetricService` class to replace `TransactionService`. The `MetricService` can retrieve transaction metrics for a given transaction ID by making transaction record queries to either the consensus node or mirror node. All logic related to `transactionFee` and `gasUsed` metrics has also been moved to this class.
  
- Modified `SDKClient` so that all logic for retrieving transaction metrics through transaction records, capturing those metrics in the HBAR rate limiter and metric registry, now runs asynchronously in the background. This allows `sdkClient.executeTransaction()` to return `transactionResponse` immediately after executing the transaction, without waiting for those actions. Resulting in less request time for the whole call.

- Added additional JSDoc comments to multiple methods and variables.

- Modified existing tests and added new tests to cover the new functions.

Fixes #2847 
Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/2808

**Notes for reviewer**:

**Checklist**


victor-yanev (changes requested):
Instead of passing service classes like `MetricService` through the arguments of every `SdkClient` method (which tightly couples the two classes together), I recommend using an event-driven approach that makes use of NodeJS’s EventEmitter. 

This would decouple `SdkClient` from the services it interacts with, allowing for better modularity and scalability. The `SdkClient` would emit events after transactions, and services like `MetricService` could listen for these events to perform their tasks.

This change would make it easier to add or modify services that listen to the event emitted by the `SdkClient` without altering the `SdkClient`, leading to a more maintainable codebase.

Here is a small example:
```typescript

import { EventEmitter } from 'events';

class SdkClient {
    constructor(private eventEmitter: EventEmitter) {}

    executeTransaction(…): void {
        try {
            …
        } catch {
            …
        }  finally {
            this.eventEmitter.emit(Constants.EVENTS.TRANSACTION_EXECUTED, {
                transactionId,
                callerName,
                requestId,
                txConstructorName,
                operatorAccountId,
                transactionType,
                interactingEntity,
            });
        }
    }
}

class MetricService {
    constructor(private eventEmitter: EventEmitter) {
        eventEmitter.on(Constants.EVENTS.TRANSACTION_EXECUTED , this.captureTransactionMetrics.bind(this));
    }
}
```

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,297 @@
...
+   */
+  private captureMetrics = (
+    mode: string,
+    type: string,
+    status: Status,
+    cost: number,
+    gas: any,
+    caller: string,
+    interactingEntity: string,
+  ) => {
```
konstantinabl, line 286: add a return type

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,297 @@
...
+  public addExpenseAndCaptureMetrics = (
+    executionType: string,
+    transactionId: string,
+    transactionType: string,
+    callerName: string,
+    cost: number,
+    gasUsed: number,
+    interactingEntity: string,
+    formattedRequestId: string,
+  ) => {
```
konstantinabl, line 200: add return type - void

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,297 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { Logger } from 'pino';
+import { Status } from '@hashgraph/sdk';
+import HbarLimit from '../../hbarlimiter';
```
konstantinabl, line 23: same goes for the import type here

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,297 @@
...
+    const resolvedGas = gas ? (typeof gas === 'object' ? gas.toInt() : gas) : 0;
+
+    this.consensusNodeClientHistogramCost
+      .labels(mode, type, status.toString(), caller, interactingEntity)
+      .observe(resolvedCost);
+    this.consensusNodeClientHistogramGasFee
+      .labels(mode, type, status.toString(), caller, interactingEntity)
+      .observe(resolvedGas);
+  };
+}
```
konstantinabl, line 297: mark as readonly

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,297 @@
...
+
+      if (transactionRecordMetrics) {
+        transactionFee = transactionRecordMetrics.transactionFee;
+      }
+    }
+
+    // capture metrics to HBAR rate limiter and metric registry
+    if (transactionFee !== 0) {
+      this.addExpenseAndCaptureMetrics(
+        `TransactionExecution`,
```
konstantinabl, line 155: Why are those in these ` quotes but not normal quotes -> '

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,297 @@
...
+    let transactionFee: number = 0;
+    let txRecordChargeAmount: number = 0;
+    const formattedRequestId = formatRequestIdMessage(requestId);
+
+    // check if calls are default to consensus node or not
+    const defaultToConsensusNode = process.env.GET_RECORD_DEFAULT_TO_CONSENSUS_NODE === 'true';
+
+    // retrieve metrics
+    if (defaultToConsensusNode) {
+      const transactionRecordMetrics = await this.sdkClient.getTransactionRecordMetrics(
```
konstantinabl, line 125: I think you can extract the call to this method in a separate function where you check the GET_RECORD_DEFAULT_TO_CONSENSUS_NODE. Looks more readable to me like this
E.g
```
private async getTransactionRecordMetrics(
  transactionId: string,
  callerName: string,
  requestId: string,
  txConstructorName: string,
  operatorAccountId: string,
  defaultToConsensusNode: boolean,
): Promise<{ gasUsed?: number; transactionFee?: number; txRecordChargeAmount?: number } | null> {
  if (defaultToConsensusNode) {
    return this.sdkClient.getTransactionRecordMetrics(
      transactionId,
      callerName,
      requestId,
      txConstructorName,
      operatorAccountId,
    );
  } else {
    return this.mirrorNodeClient.getTransactionRecordMetrics(
      transactionId,
      callerName,
      requestId,
      txConstructorName,
      operatorAccountId,
    );
  }
}
```
Then if the result is null you can handle it and if not
you can extract the values you want
  const { gasUsed, transactionFee, txRecordChargeAmount } = transactionRecordMetrics;

konstantinabl (changes requested):
Few suggested changes

quiet-node:
> Instead of passing service classes like `MetricService` through the arguments of every `SdkClient` method (which tightly couples the two classes together), I recommend using an event-driven approach that makes use of NodeJS’s EventEmitter.
> 
> This would decouple `SdkClient` from the services it interacts with, allowing for better modularity and scalability. The `SdkClient` would emit events after transactions, and services like `MetricService` could listen for these events to perform their tasks.
> 
> This change would make it easier to add or modify services that listen to the event emitted by the `SdkClient` without altering the `SdkClient`, leading to a more maintainable codebase.
> 
> Here is a small example:
> 
> ```ts
> import { EventEmitter } from 'events';
> 
> class SdkClient {
>     constructor(private eventEmitter: EventEmitter) {}
> 
>     executeTransaction(…): void {
>         try {
>             …
>         } catch {
>             …
>         }  finally {
>             this.eventEmitter.emit(Constants.EVENTS.TRANSACTION_EXECUTED, {
>                 transactionId,
>                 callerName,
>                 requestId,
>                 txConstructorName,
>                 operatorAccountId,
>                 transactionType,
>                 interactingEntity,
>             });
>         }
>     }
> }
> 
> class MetricService {
>     constructor(private eventEmitter: EventEmitter) {
>         eventEmitter.on(Constants.EVENTS.TRANSACTION_EXECUTED , this.captureTransactionMetrics.bind(this));
>     }
> }
> ```
@victor-yanev, awesome idea! I had been struggling to find a solution to avoid drilling the `MetricService` class down to dependent classes but couldn't come up with a viable approach. The logic of the `EventEmitter` class is not only simple and efficient but also helps avoid overlapping services. I’ve just pushed an update to implement this idea. Thanks for the great suggestion!

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -31,14 +31,21 @@ import { IOpcodesResponse } from './models/IOpcodesResponse';
...
+  ILimitOrderParams,
   IContractCallRequest,
   IContractCallResponse,
-  IContractLogsResultsParams,
   IContractResultsParams,
-  ILimitOrderParams,
+  IContractLogsResultsParams,
+  MirrorNodeTransactionRecord,
+  IMirrorNodeTransactionRecord,
 } from '../types/IMirrorNode';
```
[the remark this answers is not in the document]
quiet-node: Good point!

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1273,4 +1280,57 @@ export class MirrorNodeClient {
...
+      mirrorNodeRetries,
+      formattedRequestId,
+    );
+
+    if (!transactionRecord) {
+      this.logger.warn(
+        `${formattedRequestId} No transaction record retrieved: transactionId=${transactionId}, txConstructorName=${txConstructorName}, callerName=${callerName}`,
+      );
+    } else {
+      const transactionReceipt: IMirrorNodeTransactionRecord = transactionRecord.transactions.find(
```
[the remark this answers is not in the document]
quiet-node: Ah yeah no worries! So the `transactionRecord` is actually the response from the `this.repeatedRequest()` method, which is located above at line 1314. After executing that line, `transactionRecord` will hold an array of transactions. From there, we traverse the array and use `.find()` to locate the specific transaction we want to work with. Let me know if it makes sense!

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -176,54 +170,80 @@ export class SDKClient {
    * @param {string} callerName - The name of the caller requesting the balance.
    * @param {string} requestId - The unique request ID for tracking the request.
    * @param {number} ms - The delay in milliseconds before retrieving the balance.
+   * @param {MetricService} metricService - The metric service instance for tracking metrics.
    * @returns {Promise<number>} - A promise that resolves to the account balance in tinybars.
    */
-  async getBalanceInTinyBars(accountId: string, callerName: string, requestId: string, ms: number): Promise<number> {
+  async getBalanceInTinyBars(
```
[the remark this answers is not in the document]
quiet-node: Ah good catch! Since Victor's suggestion we don't need this anymore! Thanks for pointing it out!

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: oh it's jsut a habit they have the same function.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1265,12 +1272,67 @@ export class MirrorNodeClient {
...
+        (tx: any) => tx.transaction_id === formatTransactionId(transactionId),
+      );
+
+      const mirrorNodeTxRecord = new MirrorNodeTransactionRecord(transactionReceipt);
+
+      // get transactionFee
+      const transactionFee = getTransferAmountSumForAccount(mirrorNodeTxRecord, operatorAccountId);
+      return { transactionFee, txRecordChargeAmount: 0, gasUsed: 0 };
```
victor-yanev, line 1335: - Move the `getTransferAmountSumForAccount` method inside the `MirrorNodeTransaction` class
- Rename `txRecordChargeAmount` to `transactionRecordFee`
```typescript
      // get transactionFee
      const transactionFee = transaction.getTransferAmountSumForAccount(operatorAccountId);
      return { transactionFee, transactionRecordFee: 0, gasUsed: 0 };
```

packages/relay/src/lib/constants.ts
```diff
@@ -188,4 +188,9 @@ export default {
 
   // @source: https://docs.hedera.com/hedera/networks/mainnet/fees
   TX_RECORD_QUERY_COST_IN_CENTS: 0.01,
+
+  EVENTS: {
+    START_CAPTURING_TRANSACTION_METRICS: 'start_capturing_transaction_metrics',
+    START_ADD_EXPENSE_AND_CAPTURE_METRICS: 'start_add_expense_and_capture_metrics',
+  },
```
victor-yanev, line 195:
```suggestion
  EVENTS: {
    TRANSACTION_EXECUTED: `transaction_executed`
  },
```

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -31,14 +31,21 @@ import { IOpcodesResponse } from './models/IOpcodesResponse';
...
-import { formatRequestIdMessage, formatTransactionId } from '../../formatters';
 import Axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
 import {
+  parseNumericEnvVar,
+  formatTransactionId,
+  formatRequestIdMessage,
+  getTransferAmountSumForAccount,
+} from '../../formatters';
+import {
+  ILimitOrderParams,
```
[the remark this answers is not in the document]
konstantinabl: @quiet-node did you see this comment :? 
its a nit: so no need to fix it here, if you wont make any more changes

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,366 @@
...
+        args.transactionId,
+        args.transactionType,
+        args.callerName,
+        args.cost,
+        args.gasUsed,
+        args.interactingEntity,
+        args.formattedRequestId,
+      );
+    });
+  }
```
victor-yanev, line 142: The `SdkClient` should emit the `TRANSACTION_EXECUTED` event since it isn’t aware of `MetricService`’s existence. `MetricService` can then subscribe to this event and handle it as needed. Having separate events for each method in `MetricService` and emitting them inside `SdkClient` is counterproductive - it’s similar to making `MetricService` a dependency, which defeats the purpose of using an `EventEmitter`:
```typescript
    this.eventEmitter.on(constants.TRANSACTION_EXECUTED, (args) => {
      // capture metrics for grafana dashboard
      this.captureTransactionMetrics(
        args.transactionId,
        args.callerName,
        args.requestId,
        args.txConstructorName,
        args.operatorAccountId,
        args.transactionType,
        args.interactingEntity,
      );
      // add expense to hbar limiter spending
      this.addExpenseAndCaptureMetrics(
        args.executionType,
        args.transactionId,
        args.transactionType,
        args.callerName,
        args.cost,
        args.gasUsed,
        args.interactingEntity,
        args.formattedRequestId,
      );
    });
  }
```

packages/relay/src/lib/clients/mirrorNodeClient.ts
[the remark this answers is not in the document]
konstantinabl: Ah yes, for some reason i thought this else is when we DO NOT have a transaction record :d

packages/relay/src/lib/services/metricService/metricService.ts
victor-yanev: Also, you can change `captureTransactionMetrics` and `addExpenseAndCaptureMetrics` to accept an object argument (you can use object de-structuring in the method's arguments) and change this to:
```typescript
    this.eventEmitter.on(constants.TRANSACTION_EXECUTED, (args) => {
      // capture metrics for grafana dashboard
      this.captureTransactionMetrics(args);
      // add expense to hbar limiter spending
      this.addExpenseAndCaptureMetrics(args);
    });
  }
```

victor-yanev (changes requested):

packages/relay/src/lib/clients/mirrorNodeClient.ts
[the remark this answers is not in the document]
quiet-node: > Okay, then let's rename this interface because it's very misleading:
>
> IMirrorNodeTransactionRecord -> ITransaction
>
> and change the variable to:
>
> transactionReceipt -> transaction

@victor-yanev Thanks for the suggestion! However, I don’t see it as misleading. I specifically named it `IMirrorNodeTransactionRecord` to clearly indicate that it’s assigned to the transaction record response retrieved from querying the mirror node. This helps distinguish it from the transaction record response from the consensus node. Additionally, `ITransaction` feels a bit too general to me. That said, I do agree that `transactionReceipt` should be renamed, but I would rename it to `transactionRecord` instead of `transaction`.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1265,12 +1272,67 @@ export class MirrorNodeClient {
...
+    if (!transactionRecord) {
+      this.logger.warn(
+        `${formattedRequestId} No transaction record retrieved: transactionId=${transactionId}, txConstructorName=${txConstructorName}, callerName=${callerName}`,
+      );
+    } else {
+      const transactionReceipt: IMirrorNodeTransactionRecord = transactionRecord.transactions.find(
+        (tx: any) => tx.transaction_id === formatTransactionId(transactionId),
+      );
+
+      const mirrorNodeTxRecord = new MirrorNodeTransactionRecord(transactionReceipt);
```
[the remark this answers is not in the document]
quiet-node: https://github.com/hashgraph/hedera-json-rpc-relay/pull/2873#discussion_r1731677460

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1265,12 +1272,67 @@ export class MirrorNodeClient {
...
+    const formattedRequestId = formatRequestIdMessage(requestId);
+
+    this.logger.trace(
+      `${formattedRequestId} Get transaction record via mirror node: transactionId=${transactionId}, txConstructorName=${txConstructorName}, callerName=${callerName}`,
+    );
+
+    const mirrorNodeRetries = parseNumericEnvVar(
+      'MIRROR_NODE_GET_CONTRACT_RESULTS_RETRIES',
+      'MIRROR_NODE_GET_CONTRACT_RESULTS_DEFAULT_RETRIES',
+    );
```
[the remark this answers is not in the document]
quiet-node: yeah it makes sense! Pushed in the update!

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: @victor-yanev keep in mind that `getTransferAmountSumForAccount()` is also used in getTransactionRecordMetrics() method in sdkClient. I think keeping the getTransferAmountSumForAccount() method in a mutual place is better than include it in mirror-node-specific-class.

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: > The `SdkClient` should emit the `TRANSACTION_EXECUTED` event since it isn’t aware of `MetricService`’s existence. `MetricService` can then subscribe to this event and handle it as needed. Having separate events for each method in `MetricService` and emitting them inside `SdkClient` is counterproductive - it’s similar to making `MetricService` a dependency, which defeats the purpose of using an `EventEmitter`:
> 
> ```ts
>     this.eventEmitter.on(constants.TRANSACTION_EXECUTED, (args) => {
>       // capture metrics for grafana dashboard
>       this.captureTransactionMetrics(
>         args.transactionId,
>         args.callerName,
>         args.requestId,
>         args.txConstructorName,
>         args.operatorAccountId,
>         args.transactionType,
>         args.interactingEntity,
>       );
>       // add expense to hbar limiter spending
>       this.addExpenseAndCaptureMetrics(
>         args.executionType,
>         args.transactionId,
>         args.transactionType,
>         args.callerName,
>         args.cost,
>         args.gasUsed,
>         args.interactingEntity,
>         args.formattedRequestId,
>       );
>     });
>   }
> ```

@victor-yanev Great idea! However, `START_CAPTURING_TRANSACTION_METRICS` and `START_ADD_EXPENSE_AND_CAPTURE_METRICS` are two distinct events that require separate flows. Combining them into a single flow wouldn’t be accurate, as some functions only require one event or the other, but not both. If anything, should we consider adding an `args.type` and using switch statements to distinguish between the flows?


for example 
```typescript
    this.eventEmitter.on(constants.TRANSACTION_EXECUTED, (args) => {
     switch (args.type) {
          case "abc":
                // capture metrics for grafana dashboard
               this.captureTransactionMetrics(
                   args.transactionId,
                   args.callerName,
                   args.requestId,
                   args.txConstructorName,
                   args.operatorAccountId,
                   args.transactionType,
                   args.interactingEntity,
              );
              break;
              
          case "xyz":
               // add expense to hbar limiter spending
               this.addExpenseAndCaptureMetrics(
                   args.executionType,
                   args.transactionId,
                   args.transactionType,
                   args.callerName,
                   args.cost,
                   args.gasUsed,
                   args.interactingEntity,
                   args.formattedRequestId,
      );
    });        
     }
  }
```

victor-yanev: In which functions do we require only one? Could you clarify?

I think we should always capture any hbar spending using the `captureTransactionMetrics`, as well as add it to the hbar limiter's expenses using `addExpenseAndCaptureMetrics`

packages/relay/src/lib/clients/mirrorNodeClient.ts
victor-yanev: Yes, but right now this utility method contains two different implementations inside of it:
```typescript
const getTransferAmountSumForAccount = (
  transactionRecord: TransactionRecord | MirrorNodeTransactionRecord,
  accountId: string,
): number => {
  if (transactionRecord instanceof MirrorNodeTransactionRecord) {
    // implementation for MirrorNodeTransactionRecord
  } else {
    // implementation for SDK client's TransactionRecord
  }
};
```

We can:
- extract the relevant part for `MirrorNodeTransactionRecord` inside `MirrorNodeClient` (or directly inside `MirrorNodeTransactionRecord`)
- and extract the relevant part for the SDK `TransactionRecord` inside `SdkClient`

We won't need to do any type checking, the two methods will implement it's own logic, and only the relevant part of the logic will be in the class where it belongs.

quiet-node: Ah yeah it makes sense now! Yes I like that idea! Separating them into their own classes is a clean solution!

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1273,4 +1285,68 @@ export class MirrorNodeClient {
...
+   * @param {string} operatorAccountId - The account ID of the operator, used to calculate transaction fees.
+   * @returns {Promise<{transactionFee: number;} | undefined>} - An object containing the transaction fee if available, or `undefined` if the transaction record is not found.
+   */
+  public async getTransactionRecordMetrics(
+    transactionId: string,
+    callerName: string,
+    requestId: string,
+    txConstructorName: string,
+    operatorAccountId: string,
+  ): Promise<ITransactionRecordMetric | undefined> {
```
victor-yanev, line 1305: QQ: Why `| undefined` here?

If something goes wrong it's better to reject the `Promise` with an error instead of resolving it with `undefined`

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1273,4 +1285,68 @@ export class MirrorNodeClient {
...
+
+    this.logger.trace(
+      `${formattedRequestId} Get transaction record via mirror node: transactionId=${transactionId}, txConstructorName=${txConstructorName}, callerName=${callerName}`,
+    );
+
+    // poll mirror node to get transaction record
+    const transactionRecords = await this.repeatedRequest(
+      this.getTransactionById.name,
+      [transactionId, 0],
+      this.MIRROR_NODE_GET_CONTRACT_RETRIES,
```
victor-yanev, line 1316: QQ: Shouldn't this be `MIRROR_NODE_GET_TRANSACTION_BY_ID_RETRIES`

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1273,4 +1285,68 @@ export class MirrorNodeClient {
...
+
+    if (!transactionRecords) {
+      this.logger.warn(
+        `${formattedRequestId} No transaction record retrieved: transactionId=${transactionId}, txConstructorName=${txConstructorName}, callerName=${callerName}`,
+      );
+    } else {
+      const transactionRecord: IMirrorNodeTransactionRecord = transactionRecords.transactions.find(
+        (tx: any) => tx.transaction_id === formatTransactionId(transactionId),
+      );
+
```
victor-yanev, line 1328: What happens when `transactionRecords.transactions` doesn't contain the `transactionId` you are looking for? This should be handled:
```typescript
      const transactionRecord: IMirrorNodeTransactionRecord = transactionRecords.transactions.find(
        (tx: any) => tx.transaction_id === formatTransactionId(transactionId),
      );
      if (!transactionRecord) {
        // handle here
      }
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -650,21 +637,18 @@ export class SDKClient {
...
-       */
+      // emitting an EXECUTE_QUERY event to kick off capturing metrics process asynchronously
       if (queryCost && queryCost !== 0) {
-        this.addExpenseAndCaptureMetrics(
-          `TransactionExecution`,
-          query.paymentTransactionId!,
-          queryType,
+        this.eventEmitter.emit(constants.EVENTS.EXECUTE_QUERY, {
+          executionType: `TransactionExecution`,
+          transactionId: query.paymentTransactionId?.toString()!,
```
victor-yanev, line 644:
```suggestion
          transactionId: query.paymentTransactionId!.toString(),
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -650,21 +637,18 @@ export class SDKClient {
...
-       *       captured in the metrics and rate limiter class, even if SDK transactions fail at any point.
-       */
+      // emitting an EXECUTE_QUERY event to kick off capturing metrics process asynchronously
       if (queryCost && queryCost !== 0) {
-        this.addExpenseAndCaptureMetrics(
-          `TransactionExecution`,
-          query.paymentTransactionId!,
-          queryType,
+        this.eventEmitter.emit(constants.EVENTS.EXECUTE_QUERY, {
+          executionType: `TransactionExecution`,
```
victor-yanev, line 643: We are executing a query but here we set the execution type to `TransactionExecution`? Seems odd.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -927,6 +802,22 @@ export class SDKClient {
         `${formattedRequestId} Fail to executeAll for ${transactionType} transaction: transactionId=${transaction.transactionId}, callerName=${callerName}, transactionType=${transactionType}, status=${sdkClientError.status}(${sdkClientError.status._code})`,
       );
       throw sdkClientError;
+    } finally {
+      // Loop through transactionResponses to asynchronously capture metrics from each response
```
victor-yanev, line 806: No need to explain what is already self-explanatory from the code, please remove these redundant comments. 

Comments should explain "why" instead of "what".

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -927,6 +802,22 @@ export class SDKClient {
...
+      if (transactionResponses) {
+        for (let transactionResponse of transactionResponses) {
+          // emitting an EXECUTE_TRANSACTION event to kick off capturing metrics process asynchronously
```
victor-yanev, line 809: I think this is unnecessary, same for other such comments

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    //listen to EXECUTE_TRANSACTION event to kick off captureTransactionMetrics() process
+    this.eventEmitter.on(constants.EVENTS.EXECUTE_TRANSACTION, (args: IExecuteTransactionEventPayload) => {
+      this.captureTransactionMetrics(args);
+    });
+
+    //listen to EXECUTE_QUERY event to kick off addExpenseAndCaptureMetrics() process
+    this.eventEmitter.on(constants.EVENTS.EXECUTE_QUERY, (args: IExecuteQueryEventPayload) => {
+      this.addExpenseAndCaptureMetrics(args);
+    });
```
victor-yanev, line 125: Please remove these redundant comments

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,355 @@
...
+      txConstructorName,
+      operatorAccountId,
+    );
+
+    // capture metrics to HBAR rate limiter and metric registry
+    if (transactionRecordMetrics) {
+      const { gasUsed, transactionFee, txRecordChargeAmount } = transactionRecordMetrics;
+      if (transactionFee !== 0) {
+        this.addExpenseAndCaptureMetrics({
+          executionType: `TransactionExecution`,
```
victor-yanev, line 165: Please, define an enum with possible ExecutionTypes inside the `MetricService` and use it here:
```suggestion
          executionType: ExecutionType.Transaction,
```

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,355 @@
...
+          cost: transactionFee,
+          gasUsed,
+          interactingEntity,
+          requestId,
+        } as IExecuteQueryEventPayload);
+      }
+
+      if (txRecordChargeAmount !== 0) {
+        this.addExpenseAndCaptureMetrics({
+          executionType: `TransactionRecordQuery`,
```
victor-yanev, line 178: Same here:
```suggestion
          executionType: ExecutionType.TransactionRecordQuery,
```

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,355 @@
...
+    mode: string,
+    type: string,
+    status: Status,
+    cost: number,
+    gas: any,
+    caller: string,
+    interactingEntity: string,
+  ): void => {
+    const resolvedCost = cost ? cost : 0;
+    const resolvedGas = gas ? (typeof gas === 'object' ? gas.toInt() : gas) : 0;
```
victor-yanev, line 304: Why are we doing this `typeof gas === 'object'` check here?

This is an example of a line which should be commented with a comment explaining "why".

packages/relay/src/lib/types/IEvent.ts
```diff
@@ -0,0 +1,40 @@
+/* -
```
victor-yanev, line 1: Remove this file to `events.ts`

packages/relay/src/lib/types/IMetricService.ts
```diff
@@ -0,0 +1,25 @@
+/* -
```
victor-yanev, line 1: `IMetricService` suggests that this file contains the interface of the `MetricService` class and I don't see it here, this name is very misleading.

Rename this file to `transactionRecordMetric.ts` if it's going to contain only one interface, or `metrics.ts` to suggest that this could contain more than one metric type.

victor-yanev (changes requested):
Good job, looks much better than before, but still needs some improvements:

packages/relay/src/lib/clients/sdkClient.ts
victor-yanev: Even better, wrap this with `if (query.paymentTransactionId) { ... }` to make sure that we are not going to run into this case. And then inside of the `else` block, we can log a warning/error for situations when the `paymentTransactionId` is missing from the query

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: Right great catch! `MIRROR_NODE_GET_CONTRACT_RETRIES` has been used this whole time but you're right it is now used for other purposes rather than just get contract result. I'll come up with a more common name so it can be used for both methods.

quiet-node: updated with a more generic name, MIRROR_NODE_REQUEST_RETRY_COUNT

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: The `transactionRecords` array is a response from `MirrorNodeClient.getTransactionById()`, where the ID is the `transactionId`. Therefore, the `transactionRecords` array should contain at least one transaction with the specified `transactionId`. If `transactionRecords` does not include a transaction with the given `transactionId`, it indicates that `transactionRecords` was returned as `null` from `repeatedRequest`, which has already been handled earlier.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: The `paymentTransactionId()` is not a critical value used for looking up transactions or any other processing but rather for logging purposes in addExpenseAndCaptureMetrics(). A query may or may not have a `paymentTransactionId`, but the `queryCost` is always calculated regardless. Therefore, the metrics capturing logic should depend on `queryCost` rather than `paymentTransactionId`.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -650,21 +637,18 @@ export class SDKClient {
...
+      // emitting an EXECUTE_QUERY event to kick off capturing metrics process asynchronously
       if (queryCost && queryCost !== 0) {
-        this.addExpenseAndCaptureMetrics(
-          `TransactionExecution`,
-          query.paymentTransactionId!,
-          queryType,
+        this.eventEmitter.emit(constants.EVENTS.EXECUTE_QUERY, {
+          executionType: `TransactionExecution`,
+          transactionId: query.paymentTransactionId?.toString()!,
+          transactionType: queryType,
```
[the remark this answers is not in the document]
quiet-node: Ah, great catch! It was indeed a duplication since `transactionType` is the same as `txConstructorName`. I've updated it by removing the redundant entries.

packages/relay/src/lib/clients/sdkClient.ts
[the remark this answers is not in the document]
quiet-node: Since `IExecuteQueryEventPayload` is also reused in `captureTransactionMetrics`, which does not have `query.paymentTransactionId`, I'd keep it as `transactionId` to maintain its generic nature.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Ha good point! Probably copy and paste issue.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: yeah good points the comments were meant to be a todo list for me during programming. Removed in new updates.

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: I like this idea! Updated.

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: I like this idea! Updated.

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: Ah yeah that was part of the old implementation, but with the new update, gas is strictly typed as a number, so this is no longer needed. Great catch! I've updated it with the fix.

quiet-node: Ah yeah that was part of the old implementation, but with the new update, gas is strictly typed as a number, so this is no longer needed. Great catch! I've updated it with the fix.

packages/relay/src/lib/types/IEvent.ts
quiet-node: Updated.

packages/relay/src/lib/types/IMetricService.ts
quiet-node: Yeah good point! Renamed to `metrics.ts`

packages/relay/src/lib/clients/sdkClient.ts
victor-yanev: Okay, then let’s remove the `!` as it would throw an error if the paymentTransactionId is undefined or null:
```suggestion
          transactionId: query.paymentTransactionId?.toString(),
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -714,107 +691,51 @@ export class SDKClient {
...
-          formattedRequestId,
-        );
-      }
-
-      if (txRecordChargeAmount !== 0) {
-        this.addExpenseAndCaptureMetrics(
-          `TransactionRecordQuery`,
-          transaction.transactionId!,
-          transactionType,
+      if (transactionId !== '') {
```
victor-yanev, line 730:
```suggestion
      if (transactionId?.length) {
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -861,72 +778,34 @@ export class SDKClient {
...
 
-      // log and throw
       this.logger.warn(
-        `${formattedRequestId} Fail to executeAll for ${transactionType} transaction: transactionId=${transaction.transactionId}, callerName=${callerName}, transactionType=${transactionType}, status=${sdkClientError.status}(${sdkClientError.status._code})`,
+        `${formattedRequestId} Fail to executeAll for ${txConstructorName} transaction: transactionId=${transaction.transactionId}, callerName=${callerName}, transactionType=${txConstructorName}, status=${sdkClientError.status}(${sdkClientError.status._code})`,
       );
       throw sdkClientError;
+    } finally {
+      if (transactionResponses) {
+        for (let transactionResponse of transactionResponses) {
```
victor-yanev, line 796:
```suggestion
        for (const transactionResponse of transactionResponses) {
```

packages/relay/src/lib/relay.ts
```diff
@@ -19,47 +19,113 @@
...
+import EventEmitter from 'events';
+import constants from './constants';
+import findConfig from 'find-config';
+import HbarLimit from './hbarlimiter';
 import { Client } from '@hashgraph/sdk';
-import { Logger } from 'pino';
+import { prepend0x } from '../formatters';
 import { MirrorNodeClient } from './clients';
 import { Gauge, Registry } from 'prom-client';
+dotenv.config({ path: findConfig('.env') || '' });
```
victor-yanev, line 35: Please revert this change and move it back to the top before importing any files

packages/relay/src/lib/services/metricService/metricService.ts
victor-yanev: They are still not removed 😄

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,354 @@
...
+    const formattedRequestId = formatRequestIdMessage(requestId);
+    this.logger.trace(
+      `${formattedRequestId} Capturing HBAR charged: executionType=${executionType} transactionId=${transactionId}, txConstructorName=${txConstructorName}, callerName=${callerName}, cost=${cost} tinybars`,
+    );
+
+    this.hbarLimiter.addExpense(cost, Date.now(), requestId);
+    this.captureMetrics(
+      SDKClient.transactionMode,
+      txConstructorName,
+      Status.Success,
```
victor-yanev, line 222: Why is the status hardcoded?

packages/relay/src/lib/types/events.ts
```diff
@@ -0,0 +1,45 @@
...
+  transactionId: string;
+  txConstructorName: string;
+  callerName: string;
+  cost: number;
+  gasUsed: number;
+  interactingEntity: string;
+  requestId: string;
+}
+
+export enum ExecutionType {
```
victor-yanev, line 41: Please, change the keys and values of the string enum to be exactly the same. 

As it is right now it is possible that unexpected things happen when parsing (it has happened to me in the past), and we don’t really need them to be different, so let’s make our lives easier.

packages/relay/tests/lib/hapiService.spec.ts
```diff
@@ -18,29 +18,37 @@
...
 const registry = new Registry();
 const logger = pino();
 
 describe('HAPI Service', async function () {
   this.timeout(20000);
-  let hapiService: HAPIService | null;
   let hbarLimiter: HbarLimit;
+  let cacheService: CacheService;
+  let eventEmitter: EventEmitter;
+  let hapiService: HAPIService | null;
```
victor-yanev, line 43: I don’t see any point in setting this variable to null in the tests, please remove such declarations:
```suggestion
  let hapiService: HAPIService;
```

victor-yanev (dismissed):
LGTM, please address these last comments before merging and test these changes thoroughly, also please make sure that all common workflows are covered by acceptance tests. This PR introduces a lot of changes and it’s likely there are some small bugs somewhere, make sure that you manually test submitting different types of transactions and queries, especially make sure that contract deploys that execute multiple FileAppend transactions are working as expected.

packages/relay/src/lib/clients/mirrorNodeClient.ts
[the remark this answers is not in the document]
konstantinabl: Screenshot 2024-08-29 at 15 58 29
Importing whatever only as a type, when its used as a type only

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,354 @@
...
+        return await this.mirrorNodeClient.getTransactionRecordMetrics(
+          transactionId,
+          callerName,
+          requestId,
+          txConstructorName,
+          operatorAccountId,
+        );
+      }
+    } catch (error: any) {
+      this.logger.warn(error, `${formattedRequestId} ${error.message}`);
```
konstantinabl, line 351: Could make this more descriptive e.g Couldn't fetch transactionRecord due to error: _**the error message**_

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -645,26 +632,22 @@ export class SDKClient {
       }
 
       this.logger.debug(
-        `${requestIdPrefix} Fail to execute ${queryType} query: paymentTransactionId=${query.paymentTransactionId}, callerName=${callerName}, queryType=${queryType}, status=${sdkClientError.status}(${sdkClientError.status._code}), cost=${queryCost} tinybars`,
+        `${requestIdPrefix} Fail to execute ${queryConstructorName} query: paymentTransactionId=${query.paymentTransactionId}, callerName=${callerName}, queryConstructorName=${queryConstructorName}, status=${sdkClientError.status}(${sdkClientError.status._code}), cost=${queryCost} tinybars`,
```
victor-yanev, line 635: Same here:
```suggestion
        `${requestIdPrefix} Fail to execute ${queryConstructorName} query: paymentTransactionId=${query.paymentTransactionId}, callerName=${callerName}, status=${sdkClientError.status}(${sdkClientError.status._code}), cost=${queryCost} tinybars`,
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -714,107 +691,51 @@ export class SDKClient {
...
-        };
-
-        throw new SDKClientError(error, error.message);
-      }
+      // .getReceipt() will throw an error if, in any case, the status !== 22 (SUCCESS).
+      const transactionRceipt = await transactionResponse.getReceipt(this.clientMain);
 
       this.logger.info(
-        `${formattedRequestId} Successfully execute ${transactionType} transaction: transactionId=${transactionResponse.transactionId}, callerName=${callerName}, transactionType=${transactionType}, status=${Status.Success}(${Status.Success._code}), cost=${transactionFee} tinybars, gasUsed=${gasUsed}`,
+        `${formattedRequestId} Successfully execute ${txConstructorName} transaction: transactionId=${transactionResponse.transactionId}, callerName=${callerName}, txConstructorName=${txConstructorName}, status=${transactionRceipt.status}(${transactionRceipt.status._code})`,
```
victor-yanev, line 703: Same here and for all other occurrences in this file.

packages/relay/src/lib/clients/sdkClient.ts
victor-yanev: Is the operator account currently paying for all queries? 

If not, are we doing any checks that the `payerAccountId === operatorAccountId` before we add the query expenses to the `HbarLimiter`?

packages/relay/tests/lib/services/metricService/metricService.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+        .stub(TransactionRecordQuery.prototype, 'execute')
+        .resolves(mockedConsensusNodeTransactionRecord);
+
+      const originalBudget = hbarLimiter.getRemainingBudget();
+
+      // emitting an EXECUTE_TRANSACTION event to kick off capturing metrics process asynchronously
+      eventEmitter.emit(constants.EVENTS.EXECUTE_TRANSACTION, mockedExecuteTransactionEventPayload);
+
+      // small wait for hbar rate limiter to settle
+      await new Promise((r) => setTimeout(r, 100));
```
victor-yanev, line 246:
```suggestion
      await new Promise((r) => eventEmitter.once(constants.EVENTS.EXECUTE_TRANSACTION, r));
```

packages/relay/tests/lib/services/metricService/metricService.spec.ts
```diff
@@ -0,0 +1,358 @@
...
+    });
+
+    it('should listen to EXECUTE_QUERY event and kick off addExpenseAndCaptureMetrics()', async () => {
+      const originalBudget = hbarLimiter.getRemainingBudget();
+
+      // emitting an EXECUTE_QUERY event to kick off capturing metrics process
+      eventEmitter.emit(constants.EVENTS.EXECUTE_QUERY, mockedExecuteQueryEventPayload);
+
+      // small wait for hbar rate limiter to settle
+      await new Promise((r) => setTimeout(r, 100));
```
victor-yanev, line 330: Same here

packages/relay/tests/lib/services/metricService/metricService.spec.ts
victor-yanev: Oh nvm, we also need to wait for the addExpenseAndCaptureMetrics method to finish executing, disregard this comment

packages/relay/src/lib/clients/mirrorNodeClient.ts
victor-yanev: Make sure that this env variable is also updated in the Kubernetes secrets for previewnet/testnet/mainnet

packages/relay/src/lib/clients/mirrorNodeClient.ts
victor-yanev: Discussed offline.

packages/relay/src/lib/relay.ts
quiet-node: updated

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: Good catch! Updated with dynamic status

packages/relay/src/lib/types/events.ts
quiet-node: fair! Updated with the fix.

packages/relay/src/lib/services/metricService/metricService.ts
victor-yanev: `SDKClient.transactionMode` above also seems to be hardcoded

packages/relay/tests/lib/hapiService.spec.ts
quiet-node: True. Updated

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: I like the idea. Updated!

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -29,51 +29,60 @@ import { predefined } from '../errors/JsonRpcError';
...
+  private static readonly GET_CONTRACT_RESULT_LOGS_BY_ADDRESS_ENDPOINT = `contracts/${MirrorNodeClient.ADDRESS_PLACEHOLDER}/results/logs`;
+  private static readonly GET_CONTRACT_RESULTS_DETAILS_BY_CONTRACT_ID_ENDPOINT = `contracts/${MirrorNodeClient.CONTRACT_ID_PLACEHOLDER}/results/${MirrorNodeClient.TIMESTAMP_PLACEHOLDER}`;
+  private static readonly GET_CONTRACT_RESULTS_DETAILS_BY_ADDRESS_AND_TIMESTAMP_ENDPOINT = `contracts/${MirrorNodeClient.ADDRESS_PLACEHOLDER}/results/${MirrorNodeClient.TIMESTAMP_PLACEHOLDER}`;
+  private readonly MIRROR_NODE_RETRY_DELAY = parseNumericEnvVar(
+    'MIRROR_NODE_RETRY_DELAY',
+    'MIRROR_NODE_RETRY_DELAY_DEFAULT',
+  );
+  private readonly MIRROR_NODE_REQUEST_RETRY_COUNT = parseNumericEnvVar(
+    'MIRROR_NODE_REQUEST_RETRY_COUNT',
+    'MIRROR_NODE_REQUEST_RETRY_COUNT_DEFAULT',
```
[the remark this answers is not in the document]
quiet-node: Yes good point I have noted this myself and will let the release manager know! Thanks for the concern!

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -629,7 +616,7 @@ export class SDKClient {
         queryCost = query._queryPayment?.toTinybars().toNumber();
       }
       this.logger.info(
-        `${requestIdPrefix} Successfully execute ${queryType} query: paymentTransactionId=${query.paymentTransactionId}, callerName=${callerName}, transactionType=${queryType}, cost=${queryCost} tinybars`,
+        `${requestIdPrefix} Successfully execute ${queryConstructorName} query: paymentTransactionId=${query.paymentTransactionId}, callerName=${callerName}, queryConstructorName=${queryConstructorName}, cost=${queryCost} tinybars`,
```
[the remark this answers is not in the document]
quiet-node: Good catch! Updated!

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: updated

packages/relay/tests/lib/services/metricService/metricService.spec.ts
quiet-node: the suggestion failed with 
```bash
Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

packages/relay/src/lib/clients/mirrorNodeClient.ts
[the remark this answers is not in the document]
quiet-node: Ah, I see! Good idea! However, we can make these changes in a separate PR for the entire codebase, as this isn't the only file where we've been importing types like regular arguments.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: updated with suggestion

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: updated with suggestion

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -30,6 +30,7 @@ import {
   Transaction,
   AccountInfo,
   FeeSchedules,
+  ExchangeRate,
```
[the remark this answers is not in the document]
quiet-node: https://github.com/hashgraph/hedera-json-rpc-relay/pull/2873#discussion_r1736637552

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: https://github.com/hashgraph/hedera-json-rpc-relay/pull/2873#discussion_r1736637552

konstantinabl (dismissed):

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -627,16 +616,18 @@ export class SDKClient {
       } else {
```
victor-yanev, line 616: QQ: Shouldn't we set the status to success also in the if-block above after successfully executing the query?

```suggestion
        status = Status.Success.toString();
      } else {
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -1092,22 +965,84 @@ export class SDKClient {
...
-   * @param {string} caller - The name of the caller executing the transaction.
-   * @param {string} interactingEntity - The entity interacting with the transaction.
+   * Retrieves transaction record metrics for a given transaction ID.
+   *
+   * @param {string} transactionId - The ID of the transaction to retrieve metrics for.
+   * @param {string} callerName - The name of the caller requesting the transaction record.
+   * @param {string} requestId - The request ID for tracking the request.
+   * @param {string} txConstructorName - The name of the transaction constructor.
+   * @param {string} operatorAccountId - The account ID of the operator.
+   * @returns {Promise<{transactionFee: number; txRecordChargeAmount: number; gasUsed: number;} | undefined>} - A promise that resolves to an object containing transaction metrics.
```
victor-yanev, line 975:
```suggestion
   * @returns {Promise<ITransactionMetric>} - A promise that resolves to an object containing transaction metrics.
```

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -714,107 +696,51 @@ export class SDKClient {
...
-      if (transactionStatus === Status.WrongNonce.toString()) {
-        const error = {
-          status: Status.WrongNonce,
-          message: `receipt for transaction ${transactionResponse.transactionId} contained error status WRONG_NONCE`,
-        };
-
-        throw new SDKClientError(error, error.message);
-      }
+      // .getReceipt() will throw an error if, in any case, the status !== 22 (SUCCESS).
+      const transactionRceipt = await transactionResponse.getReceipt(this.clientMain);
```
victor-yanev, line 705: nit: small typo here - `transactionRceipt` -> `transactionReceipt`

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  /**
+   * Retrieves the gas fee metric for consensus node client operations.
+   *
+   * @returns {Histogram} - The histogram metric tracking the gas fees of consensus node client operations.
+   */
+  public getGasFeeMetric(): Histogram {
+    return this.consensusNodeClientHistogramGasFee;
+  }
```
victor-yanev, line 241: QQ: Are these getters used anywhere outside of the tests? We shouldn't expose things only because of tests. 

Let's keep those private and directly access them with `@ts-ignore` comments in the spec files.

Example usage inside `metricService.spec.ts`:
```
// @ts-ignore
const gasFeeHistogram =  metricService.consensusNodeClientHistogramGasFee;
```

packages/relay/src/lib/types/events.ts
victor-yanev: Actually, do we need these strings at all? I saw that we use them only for this log:

```
    this.logger.trace(
      `${formattedRequestId} Capturing HBAR charged: executionType=${executionType} transactionId=${transactionId}, txConstructorName=${transactionType}, callerName=${callerName}, cost=${cost} tinybars`,
    );
```

In my opinion, `txConstructorName` provides enough information about the kind of the transaction, why do we need this `executionType` string? It seems useless to me.

victor-yanev (changes requested):
LGTM! Last few comments:

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -0,0 +1,345 @@
...
+    status,
+    requestId,
+  }: IExecuteQueryEventPayload): void => {
+    const formattedRequestId = formatRequestIdMessage(requestId);
+    this.logger.trace(
+      `${formattedRequestId} Capturing HBAR charged: executionType=${executionType} transactionId=${transactionId}, txConstructorName=${txConstructorName}, callerName=${callerName}, cost=${cost} tinybars`,
+    );
+
+    this.hbarLimiter.addExpense(cost, Date.now(), requestId);
+    this.captureMetrics(executionType, txConstructorName, status, cost, gasUsed, callerName, interactingEntity);
```
victor-yanev, line 222: We shouldn't use `executionType` here imo, because this is gonna mess up the labels in the histograms. 

Old labels used to include the `transactionMode`/`queryMode`/`recordMode` strings, I think we should keep using these?
```
  static transactionMode = 'TRANSACTION';
  static queryMode = 'QUERY';
  static recordMode = 'RECORD';
```

victor-yanev (changes requested):

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Ah yeah good catch! I actually moved it outside of the if/else block as it has the same syntax.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Great catch! fixed with suggestion

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Good catch! Updated.

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: Fair. Updated.

packages/relay/src/lib/types/events.ts
quiet-node: https://github.com/hashgraph/hedera-json-rpc-relay/pull/2873#discussion_r1738907538

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Yes, by default, the operator account pays for all queries. According to [this source](https://docs.hedera.com/hedera/sdks-and-apis/sdks/client#id-2.-define-the-operator-account-id-and-private-key):

> The operator is the account that will, by default, pay the transaction fee for transactions and queries built with this client. The operator account ID is used to generate the default transaction ID for all transactions executed with this client. The operator private key is used to sign all transactions executed by this client.

victor-yanev (approved):

ebadiere (approved):
LG

Issue #2896, opened 2024-08-28
[HBAR Rate Limit Redesign] Replace usages of HbarLimit with the new HbarLimitService and delete it


victor-yanev:
Update the codebase to replace all instances and usages of the HbarLimit with the newly implemented HbarLimitService. The HbarLimitService provides improved functionality and better encapsulates the logic for handling rate limits on HBAR transactions.

Once all references to HbarLimit have been replaced, the old HbarLimit class or module should be safely deleted from the project to prevent redundancy and ensure that only the new service is utilized moving forward.

PR #2902, opened 2024-08-29
feat: Implement `HbarLimitService#addExpense`

### LINKED ISSUE #2901: [HBAR Rate Limit Redesign] Implement addExpense

victor-yanev:
#### Description

This PR implements the `addExpense` method in the `HbarLimitService` class and adds corresponding unit tests. The changes include:

- Implementing the `addExpense` method to handle expense addition for ETH addresses.
- Adding unit tests for the `addExpense` method to ensure correct functionality.
- Adding unit tests for other methods in `HbarLimitService` to improve test coverage.

#### Changes

1. **Implementation of `addExpense` Method:**
   - Added logic to create a basic spending plan if none exists for the provided ETH or IP address.
   - Updated the remaining budget and spending history after adding an expense.
   - Added error handling for cases where neither ETH nor IP address is provided.

2. **Unit Tests:**
   - Added tests for the `addExpense` method to cover various scenarios, including:
     - Adding an expense when both ETH and IP addresses are provided.
     - Adding an expense when only ETH address is provided.
     - Handling errors when adding an expense fails.
   - Added tests for other methods in `HbarLimitService` to ensure comprehensive test coverage.

#### Related Issues

Fixes #2901 

#### Notes for Reviewer

- Ensure to implement the remaining TODOs related to IP address handling in future PRs.
- Review the added unit tests to confirm they cover all edge cases and scenarios.


packages/relay/src/lib/services/hbarLimitService/IHbarLimitService.ts
```diff
@@ -19,18 +19,7 @@
...
-  /**
-   * Determines if the Hbar limit should be applied based on the provided Ethereum address
-   * and optionally an IP address.
-   *
-   * @param {string} ethAddress - The Ethereum address to check.
-   * @param {string} [ipAddress] - The optional IP address to check.
-   * @returns {Promise<boolean>} - True if the limit should be applied, false otherwise.
-   */
-  shouldLimit(ethAddress: string, ipAddress?: string): Promise<boolean>;
+  shouldLimit(mode: string, methodName: string, ethAddress: string, ipAddress?: string): Promise<boolean>;
```
ebadiere, line 23: Nit: Should we consider adding the following rule to the `eslintrc.js`? `"@typescript-eslint/method-signature-style": "off"`

This would prevent the warning below: 
`Shorthand method signature is forbidden. Use a function property instead.`

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -25,20 +25,72 @@ import { IDetailedHbarSpendingPlan } from '../../db/types/hbarLimiter/hbarSpendi
 import { SubscriptionType } from '../../db/types/hbarLimiter/subscriptionType';
 import { Logger } from 'pino';
 import { formatRequestIdMessage } from '../../../formatters';
+import { Counter, Gauge, Registry } from 'prom-client';
```
ebadiere, line 28: Nit: Warning `Import "Registry" is only used as types`. 

Perhaps add the following to `eslintrc.js`?
` "@typescript-eslint/consistent-type-imports": "off"`

ebadiere (commented):
A couple of nits, but I have a more general question.  

I assume based on the description of issue #2896 that the plan will be to simply swap out the old HBar Rate Limiter with the new HBar Rate Limiter in the `sdkClient.ts`.   If that is the case what happens if the Tier 3 general users are limited to 10K HBar per day, and a given partner is limited to 20K per day, and the general users use up all of the 10K budget before the partner has used up all of their allocated 20K.  Is the partner rate limited as well, even though the partner may have only used up 12K Hbar that day?

Should we have a test for this?

victor-yanev:
> A couple of nits, but I have a more general question.
> 
> I assume based on the description of issue #2896 that the plan will be to simply swap out the old HBar Rate Limiter with the new HBar Rate Limiter in the `sdkClient.ts`. If that is the case what happens if the Tier 3 general users are limited to 10K HBar per day, and a given partner is limited to 20K per day, and the general users use up all of the 10K budget before the partner has used up all of their allocated 20K. Is the partner rate limited as well, even though the partner may have only used up 12K Hbar that day?
> 
> Should we have a test for this?

I have also been thinking about this, but I couldn't come up with something else than eliminating the total daily limit and only keeping the tiered limits per user/partner, let's discuss it on the parking lot today.

packages/relay/src/lib/services/hbarLimitService/index.ts
victor-yanev: Done, fixed the eslint rules

packages/relay/src/lib/services/hbarLimitService/IHbarLimitService.ts
victor-yanev: Done, fixed the eslint rules

ebadiere (approved):

Ivo-Yankov (approved):
LGTM

PR #2922, opened 2024-08-30
feat: improved preemtive rate limit for file transactions

### LINKED ISSUE #2921: Improved the current preemptive rate limit logic
Currently, there is a preemptive rate limiting logic that enables the Relay to estimate the amount of HBAR a client call will cost the Relay operator in advance. This logic was introduced as a temporary solution in the hotfix for v0.52.1. It's now time to revisit this implementation to improve code readability, method functionality, and scope.

quiet-node:
**Description**:
A preemptive rate limit was introduced as a hotfix in version 0.52.1, using a logic that preemptively rate-limited file transactions by comparing a hard-coded amount of tinybars that was assumed to be the cost for `FileCreate` and `FileAppend` transactions. However, this logic was incorrect and the tinybar amount was a random figure agreed among the team for the purposes of the hotfix.

This PR improves the rate-limiting logic by leveraging the [Hedera Fee Calculator](https://hedera.com/fee) to estimate fees for `FileCreate` and `FileAppend` transactions based on a file size of `FILE_APPEND_CHUNK_SIZE` bytes. This approach ensures the estimated fees more closely align with the actual transaction fees charged for these transactions.

Fixes #2921

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -402,7 +402,7 @@ export class SDKClient {
       if (isPreemtiveCheckOn) {
         this.hbarLimiter.shouldPreemtivelyLimit(
           originalCallerAddress,
-          ethereumTransactionData.callData.length,
+          ethereumTransactionData.toString().length,
           this.fileAppendChunkSize,
           requestId,
         );
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -399,9 +399,9 @@ export class SDKClient {
     if (ethereumTransactionData.callData.length <= this.fileAppendChunkSize) {
       ethereumTransaction.setEthereumData(ethereumTransactionData.toBytes());
     } else {
-      const isPreemtiveCheckOn = process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK === 'true';
+      const isPreemptiveCheckOn = process.env.HBAR_RATE_LIMIT_PREEMPTIVE_CHECK === 'true';
 
-      if (isPreemtiveCheckOn) {
+      if (isPreemptiveCheckOn) {
         const shouldPreemptivelyLimit = this.hbarLimiter.shouldPreemptivelyLimitFileTransactions(
           originalCallerAddress,
           ethereumTransactionData.toString().length,
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -376,6 +376,7 @@ export class SDKClient {
    * @param {string} callerName - The name of the caller initiating the transaction.
    * @param {string} originalCallerAddress - The address of the original caller making the request.
    * @param {number} networkGasPriceInWeiBars - The predefined gas price of the network in weibar.
+   * @param {number} currentNetworkExchangeRateInCents - The exchange rate in cents of the current network.
    * @param {string} requestId - The unique identifier for the request.
    * @returns {Promise<{ txResponse: TransactionResponse; fileId: FileId | null }>}
    * @throws {SDKClientError} Throws an error if no file ID is created or if the preemptive fee check fails.
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -402,13 +402,17 @@ export class SDKClient {
       const isPreemtiveCheckOn = process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK === 'true';
 
       if (isPreemtiveCheckOn) {
-        this.hbarLimiter.shouldPreemtivelyLimitFileTransactions(
+        const shouldPreemptivelyLimit = this.hbarLimiter.shouldPreemptivelyLimitFileTransactions(
           originalCallerAddress,
           ethereumTransactionData.toString().length,
           this.fileAppendChunkSize,
           currentNetworkExchangeRateInCents,
           requestId,
         );
+
+        if (shouldPreemptivelyLimit) {
+          throw predefined.HBAR_RATE_LIMIT_PREEMPTIVE_EXCEEDED;
+        }
       }
 
       fileId = await this.createFile(
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -385,6 +385,7 @@ export class SDKClient {
     callerName: string,
     originalCallerAddress: string,
     networkGasPriceInWeiBars: number,
+    currentNetworkExchangeRateInCents: number,
     requestId: string,
   ): Promise<{ txResponse: TransactionResponse; fileId: FileId | null }> {
     const ethereumTransactionData: EthereumTransactionData = EthereumTransactionData.fromBytes(transactionBuffer);
@@ -400,10 +401,11 @@ export class SDKClient {
       const isPreemtiveCheckOn = process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK === 'true';
 
       if (isPreemtiveCheckOn) {
-        this.hbarLimiter.shouldPreemtivelyLimit(
+        this.hbarLimiter.shouldPreemtivelyLimitFileTransactions(
           originalCallerAddress,
           ethereumTransactionData.toString().length,
           this.fileAppendChunkSize,
+          currentNetworkExchangeRateInCents,
           requestId,
         );
       }
@@ -1030,7 +1032,7 @@ export class SDKClient {
   public calculateTxRecordChargeAmount(exchangeRate: ExchangeRate): number {
     const exchangeRateInCents = exchangeRate.exchangeRateInCents;
     const hbarToTinybar = Hbar.from(1, HbarUnit.Hbar).toTinybars().toNumber();
-    return Math.round((constants.TX_RECORD_QUERY_COST_IN_CENTS / exchangeRateInCents) * hbarToTinybar);
+    return Math.round((constants.NETWORK_FEES_IN_CENTS.TRANSACTION_GET_RECORD / exchangeRateInCents) * hbarToTinybar);
   }
 
   /**
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -402,9 +402,10 @@ export class SDKClient {
       const isPreemptiveCheckOn = process.env.HBAR_RATE_LIMIT_PREEMPTIVE_CHECK === 'true';
 
       if (isPreemptiveCheckOn) {
+        const hexCallDataLength = Buffer.from(ethereumTransactionData.callData).toString('hex').length;
         const shouldPreemptivelyLimit = this.hbarLimiter.shouldPreemptivelyLimitFileTransactions(
           originalCallerAddress,
-          ethereumTransactionData.toString().length,
+          hexCallDataLength,
           this.fileAppendChunkSize,
           currentNetworkExchangeRateInCents,
           requestId,
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -397,30 +397,15 @@ export class SDKClient {
     if (ethereumTransactionData.callData.length <= this.fileAppendChunkSize) {
       ethereumTransaction.setEthereumData(ethereumTransactionData.toBytes());
     } else {
-      // notice: this solution is temporary and subject to change.
-      const isPreemtiveCheckOn = process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK
-        ? process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK === 'true'
-        : false;
+      const isPreemtiveCheckOn = process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK === 'true';
 
       if (isPreemtiveCheckOn) {
-        const numFileCreateTxs = 1;
-        const numFileAppendTxs = Math.ceil(ethereumTransactionData.callData.length / this.fileAppendChunkSize);
-        const fileCreateFee = Number(process.env.HOT_FIX_FILE_CREATE_FEE || 100000000); // 1 hbar
-        const fileAppendFee = Number(process.env.HOT_FIX_FILE_APPEND_FEE || 120000000); // 1.2 hbar
-
-        const totalPreemtiveTransactionFee = numFileCreateTxs * fileCreateFee + numFileAppendTxs * fileAppendFee;
-
-        const shouldPreemtivelyLimit = this.hbarLimiter.shouldPreemtivelyLimit(
+        this.hbarLimiter.shouldPreemtivelyLimit(
           originalCallerAddress,
-          totalPreemtiveTransactionFee,
+          ethereumTransactionData.callData.length,
+          this.fileAppendChunkSize,
           requestId,
         );
-        if (shouldPreemtivelyLimit) {
-          this.logger.trace(
-            `${requestIdPrefix} The total preemptive transaction fee exceeds the current remaining HBAR budget due to an excessively large callData size: numFileCreateTxs=${numFileCreateTxs}, numFileAppendTxs=${numFileAppendTxs}, totalPreemtiveTransactionFee=${totalPreemtiveTransactionFee}, callDataSize=${ethereumTransactionData.callData.length}`,
-          );
-          throw predefined.HBAR_RATE_LIMIT_PREEMTIVE_EXCEEDED;
-        }
       }
 
       fileId = await this.createFile(
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -2534,7 +2534,7 @@ export class EthImpl implements Eth {
 
     if (!currentNetworkExchangeRate) {
       currentNetworkExchangeRate = (await this.mirrorNodeClient.getNetworkExchangeRate(requestId)).current_rate;
-      this.cacheService.set(cacheKey, currentNetworkExchangeRate, callingMethod, cacheTTL, requestIdPrefix);
+      await this.cacheService.set(cacheKey, currentNetworkExchangeRate, callingMethod, cacheTTL, requestIdPrefix);
     }
 
     const exchangeRateInCents = currentNetworkExchangeRate.cent_equivalent / currentNetworkExchangeRate.hbar_equivalent;
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -55,6 +55,7 @@ import {
   formatContractResult,
   weibarHexToTinyBarInt,
   isValidEthereumAddress,
+  formatRequestIdMessage,
   formatTransactionIdWithoutQueryParams,
   getFunctionSelector,
 } from '../formatters';
@@ -1509,7 +1510,7 @@ export class EthImpl implements Eth {
 
         if (!accountNonce) {
           this.logger.warn(`${requestIdPrefix} Cannot find updated account nonce.`);
-          throw predefined.INTERNAL_ERROR(`Cannot find updated account nonce for WRONT_NONCE error.`);
+          throw predefined.INTERNAL_ERROR(`Cannot find updated account nonce for WRONG_NONCE error.`);
         }
 
         if (parsedTx.nonce > accountNonce) {
@@ -1538,9 +1539,10 @@ export class EthImpl implements Eth {
    * Submits a transaction to the network for execution.
    *
    * @param transaction
-   * @param requestIdPrefix
+   * @param requestId
    */
-  async sendRawTransaction(transaction: string, requestIdPrefix: string): Promise<string | JsonRpcError> {
+  async sendRawTransaction(transaction: string, requestId: string): Promise<string | JsonRpcError> {
+    const requestIdPrefix = formatRequestIdMessage(requestId);
     if (transaction?.length >= constants.FUNCTION_SELECTOR_CHAR_LENGTH)
       this.ethExecutionsCounter
         .labels(EthImpl.ethSendRawTransaction, transaction.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH))
@@ -1563,6 +1565,7 @@ export class EthImpl implements Eth {
           EthImpl.ethSendRawTransaction,
           originalCallerAddress,
           networkGasPriceInWeiBars,
+          await this.getCurrentNetworkExchangeRateInCents(requestIdPrefix),
           requestIdPrefix,
         );
 
@@ -2513,5 +2516,17 @@ export class EthImpl implements Eth {
       .reduce((total, amount) => {
         return total + amount;
       }, 0);
+  }
+
+  /**
+   * Retrieves the current network exchange rate of HBAR to USD in cents.
+   *
+   * @param {string} requestId - The unique identifier for the request.
+   * @returns {Promise<number>} - A promise that resolves to the current exchange rate in cents.
+   */
+  private async getCurrentNetworkExchangeRateInCents(requestId: string): Promise<number> {
+    const currentNetworkExchangeRate = (await this.mirrorNodeClient.getNetworkExchangeRate(requestId)).current_rate;
+    const exchangeRateInCents = currentNetworkExchangeRate.cent_equivalent / currentNetworkExchangeRate.hbar_equivalent;
+    return exchangeRateInCents;
   }
 }
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -2525,7 +2525,18 @@ export class EthImpl implements Eth {
    * @returns {Promise<number>} - A promise that resolves to the current exchange rate in cents.
    */
   private async getCurrentNetworkExchangeRateInCents(requestId: string): Promise<number> {
-    const currentNetworkExchangeRate = (await this.mirrorNodeClient.getNetworkExchangeRate(requestId)).current_rate;
+    const requestIdPrefix = formatRequestIdMessage(requestId);
+    const cacheKey = constants.CACHE_KEY.CURRENT_NETWORK_EXCHANGE_RATE;
+    const callingMethod = this.mirrorNodeClient.getNetworkExchangeRate.name;
+    const cacheTTL = 15 * 60 * 1000; // 15 minutes
+
+    let currentNetworkExchangeRate = await this.cacheService.getAsync(cacheKey, callingMethod, requestIdPrefix);
+
+    if (!currentNetworkExchangeRate) {
+      currentNetworkExchangeRate = (await this.mirrorNodeClient.getNetworkExchangeRate(requestId)).current_rate;
+      this.cacheService.set(cacheKey, currentNetworkExchangeRate, callingMethod, cacheTTL, requestIdPrefix);
+    }
+
     const exchangeRateInCents = currentNetworkExchangeRate.cent_equivalent / currentNetworkExchangeRate.hbar_equivalent;
     return exchangeRateInCents;
   }
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -2527,7 +2527,7 @@ export class EthImpl implements Eth {
   private async getCurrentNetworkExchangeRateInCents(requestId: string): Promise<number> {
     const requestIdPrefix = formatRequestIdMessage(requestId);
     const cacheKey = constants.CACHE_KEY.CURRENT_NETWORK_EXCHANGE_RATE;
-    const callingMethod = this.mirrorNodeClient.getNetworkExchangeRate.name;
+    const callingMethod = this.getCurrentNetworkExchangeRateInCents.name;
     const cacheTTL = 15 * 60 * 1000; // 15 minutes
 
     let currentNetworkExchangeRate = await this.cacheService.getAsync(cacheKey, callingMethod, requestIdPrefix);
--- a/packages/relay/src/lib/hbarlimiter/index.ts
+++ b/packages/relay/src/lib/hbarlimiter/index.ts
@@ -132,7 +132,7 @@ export default class HbarLimit {
    * @param {number} callDataSize - The size of the call data that will be used in the transaction.
    * @param {number} fileChunkSize - The chunk size used for file append transactions.
    * @param {string} requestId - The request ID for tracing the request flow.
-   * @throws {Error} Throws an error if the total estimated transaction fee exceeds the remaining HBAR budget.
+   * @throws {JsonRpcError} Throws an error if the total estimated transaction fee exceeds the remaining HBAR budget.
    */
   shouldPreemtivelyLimitFileTransactions(
     originalCallerAddress: string,
--- a/packages/relay/src/lib/hbarlimiter/index.ts
+++ b/packages/relay/src/lib/hbarlimiter/index.ts
@@ -240,21 +240,21 @@ export default class HbarLimit {
     const fileCreateFeeInCents = constants.NETWORK_FEES_IN_CENTS.FILE_CREATE_PER_5_KB;
 
     // The first chunk goes in with FileCreateTransaciton, the rest are FileAppendTransactions
-    const fileAppend5kbTransactions = Math.floor(callDataSize / fileChunkSize) - 1;
-    const fileAppendfinalChunkSize = callDataSize % fileChunkSize;
+    const fileAppendTransactions = Math.floor(callDataSize / fileChunkSize) - 1;
+    const lastFileAppendChunkSize = callDataSize % fileChunkSize;
 
-    const fileAppendTxFeeInCents_5kb = constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_PER_5_KB;
-    const fileAppendTxFeeInCents_finalChunk =
+    const fileAppendFeeInCents = constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_PER_5_KB;
+    const lastFileAppendChunkFeeInCents =
       constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_BASE_FEE +
-      fileAppendfinalChunkSize * constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_RATE_PER_BYTE;
+      lastFileAppendChunkSize * constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_RATE_PER_BYTE;
 
-    const totalFileTransactionFeeInCents =
+    const totalTxFeeInCents =
       fileCreateTransactions * fileCreateFeeInCents +
-      fileAppendTxFeeInCents_5kb * fileAppend5kbTransactions +
-      fileAppendTxFeeInCents_finalChunk;
+      fileAppendFeeInCents * fileAppendTransactions +
+      lastFileAppendChunkFeeInCents;
 
     const estimatedTxFee = Math.round(
-      (totalFileTransactionFeeInCents / currentNetworkExchangeRateInCents) * constants.HBAR_TO_TINYBAR_COEF,
+      (totalTxFeeInCents / currentNetworkExchangeRateInCents) * constants.HBAR_TO_TINYBAR_COEF,
     );
 
     return estimatedTxFee;
--- a/packages/relay/src/lib/hbarlimiter/index.ts
+++ b/packages/relay/src/lib/hbarlimiter/index.ts
@@ -149,20 +149,20 @@ export default class HbarLimit {
       );
       return false;
     } else {
-      const { numFileCreateTxs, numFileAppendTxs, totalFeeInTinyBar } = this.estimateFileTransactionFee(
+      const { fileCreateTransactions, fileAppendTransactions, transactionFee } = this.estimateFileTransactionFee(
         callDataSize,
         fileChunkSize,
         currentNetworkExchangeRateInCents,
       );
 
-      if (this.remainingBudget - totalFeeInTinyBar < 0) {
+      if (this.remainingBudget - transactionFee < 0) {
         this.logger.warn(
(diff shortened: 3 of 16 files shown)


packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -19,7 +19,10 @@
...
 
 import { Logger } from 'pino';
-import { Counter, Gauge, Registry } from 'prom-client';
+import constants from '../constants';
+import { Hbar, HbarUnit } from '@hashgraph/sdk';
+import { predefined } from '../errors/JsonRpcError';
+import { Registry, Counter, Gauge } from 'prom-client';
 import { formatRequestIdMessage } from '../../formatters';
 
 export default class HbarLimit {
```
victor-yanev, line 28: This class is going to be deprecated and removed with the hbar limiter redesign initiative, this story is supposed to add preemptive limiting logic in the new `HbarLimitService`.

packages/relay/src/lib/eth.ts
```diff
@@ -2492,4 +2495,16 @@ export class EthImpl implements Eth {
...
+   * Retrieves the current network exchange rate of HBAR to USD in cents.
+   *
+   * @param {string} requestId - The unique identifier for the request.
+   * @returns {Promise<number>} - A promise that resolves to the current exchange rate in cents.
+   */
+  private async getCurrentNetworkExchangeRateInCents(requestId: string): Promise<number> {
+    const currentNetworkExchangeRate = (await this.mirrorNodeClient.getNetworkExchangeRate(requestId)).current_rate;
+    const exchangeRateInCents = currentNetworkExchangeRate.cent_equivalent / currentNetworkExchangeRate.hbar_equivalent;
+    return exchangeRateInCents;
+  }
```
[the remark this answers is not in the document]
quiet-node: Great idea! Pushed in the improvement.

packages/relay/src/lib/hbarlimiter/index.ts
quiet-node: This PR mainly focuses on improving current preemtive rate limit logic. This [ticket#2937](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2937) will implement the logic in the new `HbarLimitService` class.

victor-yanev: Then remove it from hbar limiter redesign epic and remove the "[HBAR Rate Limit Redesign]" prefix in the name of the issue.

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -108,44 +111,60 @@ export default class HbarLimit {
...
-   *
-   * @param {string} originalCallerAddress - The address of the original caller making the request.
-   * @param {number} transactionFee - The transaction fee in tinybars to be checked against the remaining budget.
-   * @param {string} [requestId] - An optional unique request ID for tracking the request.
-   * @returns {boolean} - Returns `true` if the rate limit should be preemptively enforced, otherwise `false`.
+   * @param {string} originalCallerAddress - The address of the caller initiating the transaction.
+   * @param {number} callDataSize - The size of the call data that will be used in the transaction.
+   * @param {number} fileChunkSize - The chunk size used for file append transactions.
+   * @param {string} requestId - The request ID for tracing the request flow.
+   * @throws {Error} Throws an error if the total estimated transaction fee exceeds the remaining HBAR budget.
```
victor-yanev, line 135:
```suggestion
   * @throws {JsonRpcError} Throws an error if the total estimated transaction fee exceeds the remaining HBAR budget.
```

packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
```diff
@@ -70,6 +70,13 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
     const value = '0x511617DE831B9E173';
     const contractResultEndpoint = `contracts/results/${transactionId}`;
```
victor-yanev, line 71:
```suggestion
    const contractResultEndpoint = `contracts/results/${transactionId}`;
    const networkExchangeRateEndpoint = 'network/exchangerate';
```

packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
```diff
@@ -105,14 +112,15 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
...
-        [txHash],
+        [txHash, getRequestId()],
       );
     });
 
     it('should return a computed hash if unable to retrieve EthereumHash from record due to contract revert', async function () {
       const signed = await signTransaction(transaction);
 
       restMock.onGet(`transactions/${transactionId}`).reply(200, null);
+      restMock.onGet(`network/exchangerate`).reply(200, mockedExchangeRate);
```
victor-yanev, line 123: Instead of adding it manually in every test, move this to a `beforeEach` block:
```typescript
beforeEach(() => {
  restMock.onGet(networkExchangeRateEndpoint).reply(200, mockedExchangeRate);
});
```

packages/relay/tests/lib/hbarLimiter.spec.ts
```diff
@@ -207,15 +211,40 @@ describe('HBAR Rate Limiter', async function () {
   });
 
   it('Should preemtively limit while expected transactionFee is greater than remaining balance', () => {
-    const validTotalTxFee = validTotal - 100;
-    const invalidTotalTxFee = validTotal + 100;
-    rateLimiter = new HbarLimit(logger, currentDateNow, validTotal, validDuration, registry);
+    rateLimiter = new HbarLimit(logger, currentDateNow, invalidTotal, validDuration, registry);
```
victor-yanev, line 214: Can we avoid instantiating this inside every test and move it to `beforeAll`? 

We can declare a separate object in the tests where we want this to be reinstantiated (e.g. after changing an env variable), but for the tests which don't do anything additional can reuse the one that was instantiated in the `beforeAll` method instead of creating a new one for every test.

packages/relay/tests/lib/hbarLimiter.spec.ts
```diff
@@ -207,15 +211,40 @@ describe('HBAR Rate Limiter', async function () {
...
+
+    try {
+      rateLimiter.shouldPreemtivelyLimitFileTransactions(
+        randomAccountAddress,
+        callDataSize,
+        fileChunkSize,
+        mockedExchangeRateInCents,
+        getRequestId(),
+      );
+      expect.fail('The rate limiter will throw HBAR_RATE_LIMIT_PREEMTIVE_EXCEEDED error');
```
victor-yanev, line 226:
```suggestion
      expect.fail('Expected an error, but no error was thrown from the hbar rate limiter');
```

packages/relay/tests/lib/hbarLimiter.spec.ts
```diff
@@ -207,15 +211,40 @@ describe('HBAR Rate Limiter', async function () {
...
+
+    try {
+      rateLimiter.shouldPreemtivelyLimitFileTransactions(
+        randomAccountAddress,
+        callDataSize,
+        fileChunkSize,
+        mockedExchangeRateInCents,
+        getRequestId(),
+      );
+      expect.fail('The rate limiter did not throw HBAR_RATE_LIMIT_PREEMTIVE_EXCEEDED error');
```
victor-yanev, line 244: Same here

packages/relay/tests/lib/hbarLimiter.spec.ts
```diff
@@ -255,19 +284,47 @@ describe('HBAR Rate Limiter', async function () {
...
+
+    try {
+      rateLimiter.shouldPreemtivelyLimitFileTransactions(
+        randomWhiteListedAccountAddress,
+        callDataSize,
+        fileChunkSize,
+        mockedExchangeRateInCents,
+        getRequestId(),
+      );
+      expect.fail('The rate limiter did not throw HBAR_RATE_LIMIT_PREEMTIVE_EXCEEDED error');
```
victor-yanev, line 298: Same here and all other occurrences.

packages/relay/tests/lib/hbarLimiter.spec.ts
```diff
@@ -255,19 +284,47 @@ describe('HBAR Rate Limiter', async function () {
...
 
-    expect(shouldNOTByPassPreemtiveRateLimit).to.be.true;
-    expect(shouldByPassPreemtiveRateLimit).to.be.false;
+  it('Should execute estimateFileTransactionFee() to estimate total fee of file transactions', async () => {
+    rateLimiter = new HbarLimit(logger, currentDateNow, validTotal, validDuration, registry);
+    const result = rateLimiter.estimateFileTransactionFee(callDataSize, fileChunkSize, mockedExchangeRateInCents);
+    const expectedResult = estimateFileTransactionsFee(callDataSize, fileChunkSize, mockedExchangeRateInCents);
+    expect(result.numFileCreateTxs).to.eq(expectedResult.numFileCreateTxs);
+    expect(result.numFileAppendTxs).to.eq(expectedResult.numFileAppendTxs);
+    expect(result.totalFeeInTinyBar).to.eq(expectedResult.totalFeeInTinyBar);
```
victor-yanev, line 328:
```suggestion
    expect(result).to.deep.eq(expectedResult);
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -288,6 +266,28 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+            }
+            expect.fail('The rate limiter will throw HBAR_RATE_LIMIT_PREEMTIVE_EXCEEDED error');
+          } catch (e) {
+            console.log(`asdaksjdhuqwiyeiqwuyhdkajshdkajhsiduuqywe`);
+            console.log(e);
+
+            expect(e.message).to.contain(predefined.HBAR_RATE_LIMIT_PREEMTIVE_EXCEEDED.message);
+          }
+
+          delete process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK;
```
victor-yanev, line 288: Please, don't do this.

First of all, we shouldn't delete this env variable, but restore it to its original state.

Also, such tests should be wrapped inside a `describe` with the pre-conditions done in `beforeAll`/`beforeEach` and then we should make sure that the original env state is always restored by placing it inside `afterAll`/`afterEach`. 

If we don't do this inside `beforeEach`/`afterEach`/`beforeAll`/`afterAll` and keep this overriding and restoring of env vars inside individual tests:
- ⚠️ any assertions that fail during this test would skip the remaining piece of code after them that restores the env variable and possibly propagate an error to all other tests which expect this env variable to be in its original state.
- ℹ️ introduce unnecessary duplication in common tests

Example of a proper test setup:
```typescript
describe('given HBAR_RATE_LIMIT_PREEMTIVE_CHECK is enabled', () => {
  let hbarRateLimitPreemptiveCheck: string | undefined;
  
  beforeAll(() => {
    hbarRateLimitPreemptiveCheck = process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK;
    process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK = 'true';
  });
  
  afterAll(() => {
    process.env.HBAR_RATE_LIMIT_PREEMTIVE_CHECK = hbarRateLimitPreemptiveCheck;
  });
  
  it('Should preemtively check the rate limit before submitting EthereumTransaction', () => {
    ...
  });

  it('<Name of some other test which needs this env variable set to true>', () => {
    ...
  });
});
```

packages/relay/src/lib/hbarlimiter/index.ts
quiet-node: Yep that's a good point! updated!

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -392,30 +393,16 @@ export class SDKClient {
...
       if (isPreemtiveCheckOn) {
-        const numFileCreateTxs = 1;
-        const numFileAppendTxs = Math.ceil(ethereumTransactionData.callData.length / this.fileAppendChunkSize);
-        const fileCreateFee = Number(process.env.HOT_FIX_FILE_CREATE_FEE || 100000000); // 1 hbar
-        const fileAppendFee = Number(process.env.HOT_FIX_FILE_APPEND_FEE || 120000000); // 1.2 hbar
-
-        const totalPreemtiveTransactionFee = numFileCreateTxs * fileCreateFee + numFileAppendTxs * fileAppendFee;
-
-        const shouldPreemtivelyLimit = this.hbarLimiter.shouldPreemtivelyLimit(
+        this.hbarLimiter.shouldPreemtivelyLimitFileTransactions(
```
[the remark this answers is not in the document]
quiet-node: Good catch! Updated.

packages/relay/src/lib/eth.ts
```diff
@@ -2492,4 +2495,27 @@ export class EthImpl implements Eth {
...
+    const requestIdPrefix = formatRequestIdMessage(requestId);
+    const cacheKey = constants.CACHE_KEY.CURRENT_NETWORK_EXCHANGE_RATE;
+    const callingMethod = this.mirrorNodeClient.getNetworkExchangeRate.name;
+    const cacheTTL = 15 * 60 * 1000; // 15 minutes
+
+    let currentNetworkExchangeRate = await this.cacheService.getAsync(cacheKey, callingMethod, requestIdPrefix);
+
+    if (!currentNetworkExchangeRate) {
+      currentNetworkExchangeRate = (await this.mirrorNodeClient.getNetworkExchangeRate(requestId)).current_rate;
+      this.cacheService.set(cacheKey, currentNetworkExchangeRate, callingMethod, cacheTTL, requestIdPrefix);
```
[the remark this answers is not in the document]
quiet-node: good catch!

packages/relay/tests/lib/hbarLimiter.spec.ts
quiet-node: I have a different perspective on this. I'm in favor of reinstantiating the `rateLimiter` object so that each unit test starts with a fresh instance, which, I guess, the OA of these tests had in mind. Well these tests are designed to modify attributes within the `rateLimiter` class like the expenses, remaining balances, etc.. Reusing the same object would carry over the updated values from one test to another, causing them to depend on each other and undermining the principle that unit tests should be independent and not influenced by changes from other tests.

Pushed in an update to clean up the test suite and add the rateLimitter instantiation to `beforeEach`, instead of `beforeAll`.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: Good point! Thanks for the suggestion!

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -108,44 +110,60 @@ export default class HbarLimit {
...
+   * @param {number} fileChunkSize - The chunk size used for file append transactions.
+   * @param {string} requestId - The request ID for tracing the request flow.
+   * @throws {JsonRpcError} Throws an error if the total estimated transaction fee exceeds the remaining HBAR budget.
    */
-  shouldPreemtivelyLimit(originalCallerAddress: string, transactionFee: number, requestId?: string): boolean {
-    if (!this.enabled) {
-      return false;
-    }
-
+  shouldPreemtivelyLimitFileTransactions(
```
victor-yanev, line 136: Please, keep the old logic which returns booleans instead of throwing errors from inside this method.

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -108,44 +110,60 @@ export default class HbarLimit {
...
     const requestIdPrefix = formatRequestIdMessage(requestId);
 
-    // check if the caller is a whitelisted caller
     if (this.isAccountWhiteListed(originalCallerAddress)) {
       this.logger.trace(
-        `${requestIdPrefix} HBAR preemtive rate limit bypassed - the caller is a whitelisted account: originalCallerAddress=${originalCallerAddress}`,
+        `${requestIdPrefix} HBAR preemptive rate limit bypassed - the caller is a whitelisted account: originalCallerAddress=${originalCallerAddress}`,
+      );
+    } else {
+      const { numFileCreateTxs, numFileAppendTxs, totalFeeInTinyBar } = this.estimateFileTransactionFee(
```
victor-yanev, line 150: numFileCreateTxs -> fileCreateTransactions
numFileAppendTxs -> fileAppendTransactions
totalFeeInTinyBar -> transactionFee

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -108,44 +110,60 @@ export default class HbarLimit {
...
+        fileChunkSize,
+        currentNetworkExchangeRateInCents,
       );
-      return false;
-    }
 
-    return this.remainingBudget - transactionFee < 0;
+      if (this.remainingBudget - totalFeeInTinyBar < 0) {
+        this.logger.warn(
+          `${requestIdPrefix} HBAR preemptive rate limit incoming call - the total preemptive transaction fee exceeds the current remaining HBAR budget due to an excessively large callData size: remainingBudget=${this.remainingBudget}, total=${this.total}, resetTimestamp=${this.reset}, callDataSize=${callDataSize}, numFileCreateTxs=${numFileCreateTxs}, numFileAppendTxs=${numFileAppendTxs}, totalFeeInTinyBar=${totalFeeInTinyBar}, exchangeRateInCents=${currentNetworkExchangeRateInCents}`,
```
victor-yanev, line 158: Please avoid nesting if-statements:
```typescript
function shouldPreemptivelyLimitFileTransactions(...): Promise<boolean> {
  if (this.isAccountWhiteListed(..) {
    this.logger.trace(...);
    return false;
  }

  if (this.remainingBudget - transactionFee < 0) {
    this.logger.warn(...);
    return true;
  }
  
  this.logger.trace(...);
  return false;
}
```

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -201,6 +219,44 @@ export default class HbarLimit {
...
+   * @param {number} currentNetworkExchangeRateInCents - The current exchange rate of HBAR to USD cents.
+   * @returns An object containing:
+   *   - `numFileCreateTxs` (number): The number of file creation transactions.
+   *   - `numFileAppendTxs` (number): The number of file append transactions.
+   *   - `totalFeeInTinyBar` (number): The estimated total transaction fee in tinybars.
+   */
+  estimateFileTransactionFee(
+    callDataSize: number,
+    fileChunkSize: number,
+    currentNetworkExchangeRateInCents: number,
```
victor-yanev, line 236: Please, can you pass `currentNetworkExchangeRate: IMirrorExchangeRate` object instead of `currentNetworkExchangeRateInCents: number` to this method and do the calculations inside?

```typescript
estimateFileTransactionFee(callDataSize: number, fileChunkSize: number, currentNetworkExchangeRate: IMirrorExchangeRate) {
  const exchangeRateInCents = currentNetworkExchangeRate.cent_equivalent / currentNetworkExchangeRate.hbar_equivalent;
  // rest of the method
}
```

victor-yanev (changes requested):

packages/server/tests/acceptance/hbarLimiter.spec.ts
victor-yanev: Still not resolved

quiet-node: Hmm strange I couldve sworn i updated this 😅 thanks for pointing it out!

quiet-node: Updated with the suggested logic. Thanks!

packages/relay/src/lib/hbarlimiter/index.ts
quiet-node: yeah good point! Pushed in the updates reverting the old logic.

packages/relay/src/lib/hbarlimiter/index.ts
quiet-node: nice naming suggestions. Updated

quiet-node: nice naming suggestions. Updated

packages/relay/tests/lib/hbarLimiter.spec.ts
```diff
@@ -31,16 +31,23 @@ describe('HBAR Rate Limiter', async function () {
   this.timeout(20000);
   let rateLimiter: HbarLimit;
   let currentDateNow: number;
-  const invalidDuration: number = 0;
+  let rateLimiterWithInvalidTotal: HbarLimit;
```
victor-yanev, line 34: `rateLimiterWithInvalidTotal` -> `rateLimiterWithEmptyBudget`

victor-yanev (changes requested):
Looks good, a couple of questions and nits

packages/relay/tests/lib/hbarLimiter.spec.ts
quiet-node: I like the name! Thanks!

packages/relay/tests/lib/openrpc.spec.ts
```diff
@@ -128,8 +129,17 @@ describe('Open RPC Specification', function () {
...
+    const eventEmitter = new EventEmitter();
 
-    clientServiceInstance = new ClientService(logger, registry, hbarLimiter, cacheService);
+    const mockedExchangeRate = {
+      current_rate: {
+        cent_equivalent: 12,
+        expiration_time: 4102444800,
+        hbar_equivalent: 1,
+      },
+    };
```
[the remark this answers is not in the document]
quiet-node: fair

packages/relay/tests/lib/openrpc.spec.ts
```diff
@@ -339,7 +351,7 @@ describe('Open RPC Specification', function () {
   });
 
   it('should execute "eth_getLogs" with no filters', async function () {
-    const response = await ethImpl.getLogs(null, null, null, null, null);
+    const response = await ethImpl.getLogs(null, 'latest', 'latest', null, null);
```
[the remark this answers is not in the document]
quiet-node: Essentially the two updated arguments do not accept nullable values. 
```typescript
async getLogs(
    blockHash: string | null,
    fromBlock: string | 'latest',
    toBlock: string | 'latest',
    address: string | string[] | null,
    topics: any[] | null,
    requestIdPrefix?: string,
  ): Promise<Log[]>
  ```
  
The previous contributor might have missed the opportunity to update the test. That's why my code editor pointed it out and I simply updated it like above.

**Review by ebadiere — APPROVED** [review]
LG.

**Review by victor-yanev — APPROVED** [review]

**Review by victor-yanev — CHANGES_REQUESTED** [review]
Btw, when I come to think about those fees we use from the hedera fee calculator - aren't those estimates for the whole fee of the transaction? What part of that is paid by the relay operator? Isn't most of that paid by the signer of the transaction? Are you sure that those estimates reflect what is actually being paid by the relay operator - please add some acceptance tests that verify the estimates are accurate by comparing the difference in the operator's balance after performing the transaction to the result from the `estimateFileTransactionFee`.

**Comment by quiet-node** [comment]
> Btw, when I come to think about those fees we use from the hedera fee calculator - aren't those estimates for the whole fee of the transaction? What part of that is paid by the relay operator? Isn't most of that paid by the signer of the transaction? Are you sure that those estimates reflect what is actually being paid by the relay operator - please add an acceptance test that verifies the estimates are accurate by comparing the difference in the operator's balance after performing the transaction to the result from the `estimateFileTransactionFee`.

@victor-yanev, very solid concern! Like I presumed in the last sync up, and after some quick experiments and checking the transaction on the explorer, relay account is indeed responsible for all the transaction fees for file transactions, FileCreate and FileAppend. `estimateFileTransactionFee()` is also only used to estimate  the fee for file transactions. Therefore, they should show the fee that are responsible only by relay account. 

However, to ensure my findings, a new acceptance test will be added soon!

**Comment by victor-yanev** [comment]
> > Btw, when I come to think about those fees we use from the hedera fee calculator - aren't those estimates for the whole fee of the transaction? What part of that is paid by the relay operator? Isn't most of that paid by the signer of the transaction? Are you sure that those estimates reflect what is actually being paid by the relay operator - please add an acceptance test that verifies the estimates are accurate by comparing the difference in the operator's balance after performing the transaction to the result from the `estimateFileTransactionFee`.
> 
> @victor-yanev, very valid concern! As I mentioned in our last sync-up and after reviewing the codebase, it is clear that the relay account is indeed responsible for all transaction fees related to file transactions, such as `FileCreate` and `FileAppend`. In the `HAPIService` class, when we call `initClient()`, the relay account is set as the operator of `clientMain`. In the sendRawTransaction flow, the Relay, on its own, initiates the file create and file append transactions to properly prepare the `callData` for the Ethereum transaction submitted to the network, making the relay account responsible for all transaction fees associated with these file transactions.
> 
> The `estimateFileTransactionFee()` function is also used solely to estimate fees for file transactions. Therefore, it should reflect only the fees that the relay account is responsible for.
> 
> > Are you sure that those estimates reflect what is actually being paid by the relay operator
> 
> There is indeed an issue with the current preemptive rate limit logic. Since the price is a hard-coded value based on the estimator for file transactions with a file size of 5120 bytes (5 KB), the estimated fee only gives a rough approximation of how much HBAR the operator will pay, rather than the exact amount. Additionally, the hard-coded value assumes a file size of 5 KB, while not all `fileAppend` transactions will carry a 5 KB chunk of content.
> 
> For example, a `callDataSize` of 17,000 bytes will result in:
> 
> * 1 `FileCreate` transaction of 5,120 bytes
> * 2 `FileAppend` transactions of 5,120 bytes each
> * 1 last `FileAppend` transaction of 1,640 bytes
> 
> However, the estimated fee treats the last `fileAppend` as a full 5 KB transaction, calculating the fee using the hard-coded 5 KB price rather than the actual 1,640-byte price. This leads to a total estimated transaction fee that is higher than what the operator actually pays.
> 
> Currently, it seems like there's no other solution to exactly estimate the accurate transaction fee to preemptively rate limit. Let me know your thoughts about this issue.

Okay, if it's overestimating the fee most of the times because we consider all chunks to be of max size (5KB), we can be even more accurate with the estimation by:
- calculating the number of full `FileAppend` transactions: 
  ```typescript
  const fullFileAppendTransactions = Math.floor(callDataSize / fileChunkSize);
  ```
- calculating the size of the remaining data for the last `FileAppend` transaction: 
  ```typescript
  const lastFileAppendSize = callDataSize % fileChunkSize;
  ```
- calculating the fee of the last `FileAppend` transaction based on the size of the remaining data:
  ```typescript
  if (lastFileAppendSize > 0) {
    const lastFileAppendFeeInCents = (lastFileAppendSize / fileChunkSize) * fileAppendFeeInCents;
    totalRequestFeeInCents += lastFileAppendFeeInCents;
  }
  ```

Full code:
```typescript
estimateFileTransactionFee(
  callDataSize: number,
  fileChunkSize: number,
  currentNetworkExchangeRateInCents: number,
) {
  const fileCreateTransactions = 1;
  const fullFileAppendTransactions = Math.floor(callDataSize / fileChunkSize);
  const lastFileAppendSize = callDataSize % fileChunkSize;
  const fileCreateFeeInCents = constants.NETWORK_FEES_IN_CENTS.FILE_CREATE_PER_5_KB;
  const fileAppendFeeInCents = constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_PER_5_KB;

  let totalRequestFeeInCents = fileCreateTransactions * fileCreateFeeInCents;
  totalRequestFeeInCents += fullFileAppendTransactions * fileAppendFeeInCents;

  let fileAppendTransactions = fullFileAppendTransactions;
  
  if (lastFileAppendSize > 0) {
    const lastFileAppendFeeInCents = (lastFileAppendSize / fileChunkSize) * fileAppendFeeInCents;
    totalRequestFeeInCents += lastFileAppendFeeInCents;
    fileAppendTransactions += 1;
  }

  const transactionFee = Math.round(
    (totalRequestFeeInCents / currentNetworkExchangeRateInCents) * constants.HBAR_TO_TINYBAR_COEF,
  );

  return {
    fileCreateTransactions,
    fileAppendTransactions,
    transactionFee,
  };
}
```

quiet-node:
> > > Btw, when I come to think about those fees we use from the hedera fee calculator - aren't those estimates for the whole fee of the transaction? What part of that is paid by the relay operator? Isn't most of that paid by the signer of the transaction? Are you sure that those estimates reflect what is actually being paid by the relay operator - please add an acceptance test that verifies the estimates are accurate by comparing the difference in the operator's balance after performing the transaction to the result from the `estimateFileTransactionFee`.
> > 
> > 
> > @victor-yanev, very valid concern! As I mentioned in our last sync-up and after reviewing the codebase, it is clear that the relay account is indeed responsible for all transaction fees related to file transactions, such as `FileCreate` and `FileAppend`. In the `HAPIService` class, when we call `initClient()`, the relay account is set as the operator of `clientMain`. In the sendRawTransaction flow, the Relay, on its own, initiates the file create and file append transactions to properly prepare the `callData` for the Ethereum transaction submitted to the network, making the relay account responsible for all transaction fees associated with these file transactions.
> > The `estimateFileTransactionFee()` function is also used solely to estimate fees for file transactions. Therefore, it should reflect only the fees that the relay account is responsible for.
> > > Are you sure that those estimates reflect what is actually being paid by the relay operator
> > 
> > 
> > There is indeed an issue with the current preemptive rate limit logic. Since the price is a hard-coded value based on the estimator for file transactions with a file size of 5120 bytes (5 KB), the estimated fee only gives a rough approximation of how much HBAR the operator will pay, rather than the exact amount. Additionally, the hard-coded value assumes a file size of 5 KB, while not all `fileAppend` transactions will carry a 5 KB chunk of content.
> > For example, a `callDataSize` of 17,000 bytes will result in:
> > 
> > * 1 `FileCreate` transaction of 5,120 bytes
> > * 2 `FileAppend` transactions of 5,120 bytes each
> > * 1 last `FileAppend` transaction of 1,640 bytes
> > 
> > However, the estimated fee treats the last `fileAppend` as a full 5 KB transaction, calculating the fee using the hard-coded 5 KB price rather than the actual 1,640-byte price. This leads to a total estimated transaction fee that is higher than what the operator actually pays.
> > Currently, it seems like there's no other solution to exactly estimate the accurate transaction fee to preemptively rate limit. Let me know your thoughts about this issue.
> 
> Okay, if it's overestimating the fee most of the times because we consider all chunks to be of max size (5KB), we can be even more accurate with the estimation by:
> 
> * calculating the number of full `FileAppend` transactions:
>   ```ts
>   const fullFileAppendTransactions = Math.floor(callDataSize / fileChunkSize);
>   ```
> * calculating the size of the remaining data for the last `FileAppend` transaction:
>   ```ts
>   const lastFileAppendSize = callDataSize % fileChunkSize;
>   ```
> * calculating the fee of the last `FileAppend` transaction based on the size of the remaining data:
>   ```ts
>   if (lastFileAppendSize > 0) {
>     const lastFileAppendFeeInCents = (lastFileAppendSize / fileChunkSize) * fileAppendFeeInCents;
>     totalRequestFeeInCents += lastFileAppendFeeInCents;
>   }
>   ```
> 
> Full code:
> 
> ```ts
> estimateFileTransactionFee(
>   callDataSize: number,
>   fileChunkSize: number,
>   currentNetworkExchangeRateInCents: number,
> ) {
>   const fileCreateTransactions = 1;
>   const fullFileAppendTransactions = Math.floor(callDataSize / fileChunkSize);
>   const lastFileAppendSize = callDataSize % fileChunkSize;
>   const fileCreateFeeInCents = constants.NETWORK_FEES_IN_CENTS.FILE_CREATE_PER_5_KB;
>   const fileAppendFeeInCents = constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_PER_5_KB;
> 
>   let totalRequestFeeInCents = fileCreateTransactions * fileCreateFeeInCents;
>   totalRequestFeeInCents += fullFileAppendTransactions * fileAppendFeeInCents;
> 
>   let fileAppendTransactions = fullFileAppendTransactions;
>   
>   if (lastFileAppendSize > 0) {
>     const lastFileAppendFeeInCents = (lastFileAppendSize / fileChunkSize) * fileAppendFeeInCents;
>     totalRequestFeeInCents += lastFileAppendFeeInCents;
>     fileAppendTransactions += 1;
>   }
> 
>   const transactionFee = Math.round(
>     (totalRequestFeeInCents / currentNetworkExchangeRateInCents) * constants.HBAR_TO_TINYBAR_COEF,
>   );
> 
>   return {
>     fileCreateTransactions,
>     fileAppendTransactions,
>     transactionFee,
>   };
> }
> ```

Nope I tried the same solution yesterday but numbers don't match. 

```typescript
...
const fileAppendTransactions = Math.ceil(callDataSize / fileChunkSize - 1);
const lastFileAppendTx = fileAppendTransactions - Math.floor(fileAppendTransactions); 
const lastFileAppendFeeInCents = (((lastFileAppendSize / fileChunkSize) * fileAppendFeeInCents) / exchangeRateInCents) * hbarToTinybar;
...
```

When running the test against the local node with `exchangeRateInCents = 12`, the `lastFileAppendFeeInCents` is approximately 26,579,589 tinybars. However, when comparing this to the last file append transaction retrieved from the mirror node, the cost is significantly higher—up to 48,438,124 tinybars. The math used to calculate the transaction fee in the services is not simply based on the size of the `callData` and exchangeRate but also depends on several other factors, such as bandwidth, storage, RAM, etc.

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -108,44 +110,63 @@ export default class HbarLimit {
...
+
+    if (this.remainingBudget - estimatedTxFeeForFileTransactions < 0) {
+      this.logger.warn(
+        `${requestIdPrefix} HBAR preemptive rate limit incoming call - the total preemptive transaction fee exceeds the current remaining HBAR budget due to an excessively large callData size: remainingBudget=${this.remainingBudget}, total=${this.total}, resetTimestamp=${this.reset}, callDataSize=${callDataSize}, estimatedTxFeeForFileTransactions=${estimatedTxFeeForFileTransactions}, exchangeRateInCents=${currentNetworkExchangeRateInCents}`,
+      );
+      return true;
+    }
+
+    this.logger.trace(
+      `${requestIdPrefix} HBAR preemptive rate limit not reached: remainingBudget=${this.remainingBudget}, total=${this.total}, resetTimestamp=${this.reset}, callDataSize=${callDataSize}, estimatedTxFeeForFileTransactions=${estimatedTxFeeForFileTransactions}, exchangeRateInCents=${currentNetworkExchangeRateInCents}`,
```
victor-yanev, line 167:
```suggestion
      `${requestIdPrefix} The remaining HBAR budget is enough to accommodate the estimated transaction fee: remainingBudget=${this.remainingBudget}, total=${this.total}, resetTimestamp=${this.reset}, callDataSize=${callDataSie}, estimatedFee=${estimatedTxFeeForFileTransactions}, exchangeRateInCents=${currentNetworkExchangeRateInCents}`,
```

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -201,6 +222,44 @@ export default class HbarLimit {
...
+  estimateFileTransactionsFee(
+    callDataSize: number,
+    fileChunkSize: number,
+    currentNetworkExchangeRateInCents: number,
+  ): number {
+    const fileCreateTransactions = 1;
+    const fileCreateFeeInCents = constants.NETWORK_FEES_IN_CENTS.FILE_CREATE_PER_5_KB;
+
+    // The first chunk goes in with FileCreateTransaciton, the rest are FileAppendTransactions
+    const fileAppend5kbTransactions = Math.floor(callDataSize / fileChunkSize) - 1;
```
victor-yanev, line 243:
```suggestion
    const fileAppendTransactions = Math.floor(callDataSize / fileChunkSize) - 1;
```

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -201,6 +222,44 @@ export default class HbarLimit {
...
+    const fileAppendfinalChunkSize = callDataSize % fileChunkSize;
```
victor-yanev, line 244:
```suggestion
    const lastFileAppendChunkSize = callDataSize % fileChunkSize;
```

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -201,6 +222,44 @@ export default class HbarLimit {
...
+
+    const fileAppendTxFeeInCents_5kb = constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_PER_5_KB;
```
victor-yanev, line 246:
```suggestion
    const fileAppendFeeInCents = constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_PER_5_KB;
```

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -201,6 +222,44 @@ export default class HbarLimit {
...
+    const fileAppendTxFeeInCents_finalChunk =
```
victor-yanev, line 247:
```suggestion
    const lastFileAppendChunkFeeInCents =
```

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -201,6 +222,44 @@ export default class HbarLimit {
...
+      constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_BASE_FEE +
+      fileAppendfinalChunkSize * constants.NETWORK_FEES_IN_CENTS.FILE_APPEND_RATE_PER_BYTE;
+
+    const totalFileTransactionFeeInCents =
```
victor-yanev, line 251:
```suggestion
    const totalFeeInCents =
```

packages/relay/src/lib/hbarlimiter/index.ts
```diff
@@ -201,6 +222,44 @@ export default class HbarLimit {
...
+      fileCreateTransactions * fileCreateFeeInCents +
+      fileAppendTxFeeInCents_5kb * fileAppend5kbTransactions +
+      fileAppendTxFeeInCents_finalChunk;
+
+    const estimatedTxFeeForFileTransactions = Math.round(
```
victor-yanev, line 256:
```suggestion
    const totalFeeInTinyBars = Math.round(
```

victor-yanev (changes requested):

victor-yanev (approved):
LGTM

victor-yanev (approved):

PR #2927, opened 2024-09-02
fix: bug where finalized and safe tags are cached

### LINKED ISSUE #2916: Discrepancy Between "Latest" and "Finalized" Block Queries in JSON RPC Relay
### Description

The Axelar team reports an issue with the JSON RPC relay on the Hedera public previewnet, where there is a discrepancy between the "latest" and "finalized" block queries. 

When querying `eth_getBlockByNumber` for "latest" and "finalized" blocks, there is a significant difference in block numbers:
- "Latest" block: `162203`
- "Finalized" block: `161682`

**Expected Behavior:**
"Latest" and "finalized" queries should return the same block number, as Hedera does not support block reversion. Both queries should point to the most recent finalized block.

### Steps to reproduce

1. Run the following command to query the latest block:
```bash
curl -s https://previewnet.hashio.io/api \
--data '{"method":"eth_getBlockByNumber","params":["latest", false],"id":1,"jsonrpc":"2.0"}' | jq -r .result.number
```
   
  2. Run the following command to query the finalized block:
  ```bash
curl -s https://previewnet.hashio.io/api \
--data '{"method":"eth_getBlockByNumber","params":["finalized", false],"id":1,"jsonrpc":"2.0"}' | jq -r .result.number
```

4. Observe the difference in the block numbers returned by the two queries.

### Additional context

**Logs and outputs:**

Polling results show the following:

- "Finalized" block query result:
  ```
  166449
  166449
  166449
  ```
  
- "Latest" block query result:
  ```
  167596
  167597
  167598
  167599
  ```

This behavior suggests an issue with the relay's handling of the API calls or possible caching problems.

```bash
2024-08-27T13:35:16.942 INFO thread-1 Finalized block: 166449
2024-08-27T13:35:38.874 INFO thread-1 Latest block: 167599
```

### Hedera network

testnet

### Version

v0.52

### Operating system

macOS

konstantinabl:
**Description**:

Currently "latest" and "pending" blocks are not cached, as it should be, since Hedera doesn't have such thing as a pending block. However, the network doesn't support finalized and safe as well, but they are cached and this leads to the issue mentioned below.
**Related issue(s)**:

Fixes #2916 

### DIFF
--- a/packages/server/tests/acceptance/rpc_batch1.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch1.spec.ts
@@ -544,6 +544,70 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
         Assertions.block(blockResult, mirrorBlock, mirrorTransactions, expectedGasPrice, false);
       });
 
+      it('should not cache "latest" block in "eth_getBlockByNumber" ', async function () {
+        const blockResult = await relay.call(
+          RelayCalls.ETH_ENDPOINTS.ETH_GET_BLOCK_BY_NUMBER,
+          ['latest', false],
+          requestId,
+        );
+        await new Promise((r) => setTimeout(r, 1000));
+
+        const blockResult2 = await relay.call(
+          RelayCalls.ETH_ENDPOINTS.ETH_GET_BLOCK_BY_NUMBER,
+          ['latest', false],
+          requestId,
+        );
+        expect(blockResult).to.not.deep.equal(blockResult2);
+      });
+
+      it('should not cache "finalized" block in "eth_getBlockByNumber" ', async function () {
+        const blockResult = await relay.call(
+          RelayCalls.ETH_ENDPOINTS.ETH_GET_BLOCK_BY_NUMBER,
+          ['finalized', false],
+          requestId,
+        );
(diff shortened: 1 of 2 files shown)


packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -544,6 +544,70 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
         Assertions.block(blockResult, mirrorBlock, mirrorTransactions, expectedGasPrice, false);
       });
 
+      it('should not cache "latest" block in "eth_getBlockByNumber" ', async function () {
+        const blockResult = await relay.call(
+          RelayCalls.ETH_ENDPOINTS.ETH_GET_BLOCK_BY_NUMBER,
+          ['latest', false],
+          requestId,
+        );
+        await new Promise((r) => setTimeout(r, 1000));
```
quiet-node, line 553: I remember there's like a .sleep() or .wait() function for these kinda waits could you please double check if not please add a util function for these waits.

quiet-node (commented):
one quick improvement.

quiet-node (approved):

ebadiere (approved):
LG.

PR #2929, opened 2024-09-02
fix: let the provider decide when to reuse an existing subscription ID

### LINKED ISSUE #2926: Websocket subscription creates the same subscription ID for new connection from the same client
### Description

when making new subscription requests from the same client, Hedera creates the same Subscription ID, whereas it should be creating a new one every time similar to other chains. this is critical because our nodes are not able to handle multiple RPC connections in a single node, causing tons of errors

### Steps to reproduce

make multiple new subscription requests to
```{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "eth_subscribe",
  "params": ["newHeads"]
}
```
when tried to connect with multiple connections for parallelisation, the existing context gets hung due to the same subscription ID
a new subscription request

### Additional context

image

### Hedera network

testnet

### Version

latest

### Operating system

None

Neurone:
**Description**:

This PR fixes #2926 (_Websocket subscription creates same subscription ID for new connection from same client_) by introducing a new environment variable that allows the JSON-RPC relay provider to choose the desired behavior.

The default value of `WS_SAME_SUB_FOR_SAME_EVENT` is set to `"true"` to match the current default behavior.

**Related issue(s)**:

Fixes #2926

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -123,6 +123,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
 | `WS_CACHE_TTL`                  | "20000"  | The time to live for cached entries.                                                                                          |
 | `WS_NEW_HEADS_ENABLED`.         | "true"   | Enables subscriptions for the latest blocks, `newHeads`.                                                                      |
 | `WS_PING_INTERVAL`              | "100000" | Interval between ping messages. Set to `0` to disable pinger.                                                                 |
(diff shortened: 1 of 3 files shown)


natanasow (approved):
LGTM

victor-yanev (approved):

PR #2931, opened 2024-09-03
feat: Add `keys()` method to `CacheService`

victor-yanev:
**Description**:

- Adds a new method to `CacheService` which retrieves all keys in the cache that match the given pattern.
- Adds the `keys()` method to the `ICacheClient` interface
- Implements the method inside `RedisClient` and `LocalLRUCache` (all implementations of the `ICacheClient`)

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/clients/cache/redisCache.ts
```diff
@@ -310,4 +310,18 @@ export class RedisCache implements IRedisCacheClient {
...
+   * Retrieves all keys matching a pattern.
+   * @param {string} pattern The pattern to match
+   * @param {string} callingMethod The name of the calling method
+   * @param {string} [requestIdPrefix] The optional request ID prefix
+   * @returns {Promise<string[]>} The list of keys matching the pattern
+   */
+  async keys(pattern: string, callingMethod: string, requestIdPrefix?: string): Promise<string[]> {
+    const client = await this.getConnectedClient();
+    const result = await client.keys(pattern);
+    this.logger.trace(`${requestIdPrefix} retrieving keys matching ${pattern} on ${callingMethod} call`);
```
quiet-node, line 324: nit: for consistency, let's tryna match up the content of the log of these keys() methods. This line and this line https://github.com/hashgraph/hedera-json-rpc-relay/pull/2931/files#diff-aa7ffa04dc980cc6bdc951ea3c201ed60d6a390851503a559371a37d73bd3014R228 can be more similar.

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -445,4 +445,27 @@ export class CacheService {
...
+   */
+  async keys(pattern: string, callingMethod: string, requestIdPrefix?: string): Promise<string[]> {
+    if (this.isSharedCacheEnabled && this.sharedCache instanceof RedisCache) {
+      try {
+        return await this.sharedCache.keys(pattern, callingMethod, requestIdPrefix);
+      } catch (error) {
+        const redisError = new RedisCacheError(error);
+        this.logger.error(
+          `${requestIdPrefix} Error occurred while getting keys from Redis. Error is: ${redisError.fullError}`,
+        );
```
quiet-node, line 464: We can add the `redisError` as the first argument of the error log so it can show full error.
```suggestion
        this.logger.error(
          redisError,
          `${requestIdPrefix} Error occurred while getting keys from Redis.`,
        );
```

quiet-node (commented):
Left some comments. Also looks like we're missing tests for these new updates.

packages/relay/src/lib/services/cacheService/cacheService.ts
victor-yanev: This wouldn't be consistent with the logging style for all other methods in this class

quiet-node: Ah, yes, I noticed that. I suggest applying the new style to the error logging in this class. Instead of embedding and converting `redisError` to a string, the new style would directly pass the error object, allowing the logger to capture the full error context. This approach is more informative, as the logger can access all properties of the error object.

victor-yanev: Okay, I will change all logs to this style then 😁

victor-yanev: Done. ✅

konstantinabl (approved):
LGTM, just make sure CI passes

natanasow (approved):
LGTM

quiet-node (approved):

PR #2939, opened 2024-09-04
feat: makes ip address of request available in application layer

### LINKED ISSUE #2867: [HBAR Rate Limit Redesign] Make the ip address of the request available in the application layer
In order to pass the ip address from the request object in Koa throughout the application, we need to change the current implementation. 
Wrap the requestId and the requestIp in a new object called requestDetails, which will be passed throughout the application

konstantinabl:
**Description**:

Makes the ip of the sender of a request, available in the application layer. Also makes requestIdPrefix required, since we need to always have it for logging purposes

**Related issue(s)**:

Fixes #2867 
Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/2940


packages/relay/src/lib/poller.ts
```diff
@@ -87,7 +88,10 @@ export class Poller {
 
           poll.lastPolled = this.latestBlock;
         } else if (event === this.NEW_HEADS_EVENT && this.newHeadsEnabled) {
-          data = await this.eth.getBlockByNumber('latest', filters?.includeTransactions ?? false);
+          data = await this.eth.getBlockByNumber('latest', filters?.includeTransactions ?? false, {
+            requestIdPrefix: '',
+            requestIp: '',
+          } as IRequestDetails);
```
victor-yanev, line 94: In those cases, `undefined` would be better instead of an object with all empty strings because it strongly suggests that this information is not available. 

```suggestion
          data = await this.eth.getBlockByNumber('latest', filters?.includeTransactions ?? false);
```

packages/relay/src/index.ts
```diff
@@ -62,31 +63,31 @@ export interface Net {
 }
 
 export interface Eth {
-  blockNumber(requestId?: string): Promise<string>;
+  blockNumber(requestIdPrefix: string): Promise<string>;
```
victor-yanev, line 66: I think it would be better to make the code base consistent and use `requestDetails` everywhere even for methods which would only use the `requestId` field of `requestDetails`.

Also, let's make it optional for methods which are used outside of the server context, e.g. in `poller.ts`

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -313,12 +313,14 @@ export class MirrorNodeClient {
...
-    requestIdPrefix?: string,
     retries?: number,
   ): Promise<T | null> {
     const start = Date.now();
     // extract request id from prefix and remove trailing ']' character
+    this.logger.debug(requestIdPrefix);
+    this.logger.debug(`${method} + ${path} + ${requestIdPrefix}`);
     const requestId =
       requestIdPrefix
         ?.split(MirrorNodeClient.REQUEST_PREFIX_SEPARATOR)[1]
```
victor-yanev, line 326: If we add a field which contains the raw `requestId` inside `requestDetails`, this logic won't be needed anymore

packages/relay/src/lib/types/IRequestDetails.ts
```diff
@@ -0,0 +1,4 @@
+export type IRequestDetails = {
+  requestIdPrefix: string;
+  requestIp: string;
+};
```
victor-yanev, line 4: Let's convert this to a class:
```typescript
export class RequestDetails {
  requestId: string;
  ipAddress: string;
  
  constructor(details?: { requestId: string, ipAddress: string }) {
    if (details) {
      this.requestId = details.requestId;
      this.ipAddress = details.ipAddress;
    }
  }
  
  get formattedRequestId(): string {
    if (!this.requestId) {
      return '';
    }
    return `[${constants.REQUEST_ID_STRING}${this.requestId}]`;
  }
}
```

Then, we can also remove the `formatRequestIdMessage` utility method and we have the flexibility of whether we want to use the raw or formatted version of the `requestId`.

victor-yanev (changes requested):

packages/relay/src/lib/poller.ts
konstantinabl: The idea was to have the request id everywhere and not have the undefined as an option, so maybe its better to generate request id

konstantinabl: I have changed it so we generate a requestId

packages/relay/src/index.ts
konstantinabl: If i make it optional for the methods in poller e.g getLogs, this will affect the whole relay, not sure if its a good idea, isnt it better to just handle the passing of requestId in the poller as I mentioned above

packages/relay/src/lib/clients/mirrorNodeClient.ts
konstantinabl: Is it worth it to add requestId and requestIdPrefix everywhere, just so we remove this. Or if you mean to pass raw requestId everywhere, then we need to format it in every method

victor-yanev: We can only add the raw `requestId` to the `RequestDetails` object and deduct the prefixed version dynamically as I suggested in https://github.com/hashgraph/hedera-json-rpc-relay/pull/2939#discussion_r1751687265.

packages/relay/src/lib/env.txt
```diff
@@ -0,0 +1,150 @@
+apiVersion: v1
```
Nana-EC, line 1: Q: please revert this configmap file addition

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -106,6 +109,7 @@ export default class KoaJsonRpc {
   rpcApp(): (ctx: Koa.Context, _next: Koa.Next) => Promise<void> {
     return async (ctx: Koa.Context, _next: Koa.Next) => {
       this.requestId = ctx.state.reqId;
+      this.requestIpAddress = ctx.request.ip;
```
Nana-EC, line 112: Q: should we only set this if the relay operator plans to rate limit by IP?

packages/server/src/server.ts
```diff
@@ -197,26 +197,29 @@ app.getKoaApp().use(async (ctx, next) => {
...
 });
 
-const logAndHandleResponse = async (methodName: any, methodParams: any, methodFunction: any) => {
-  const requestId = app.getRequestId();
-  const requestIdPrefix = requestId ? formatRequestIdMessage(requestId) : '';
+const logAndHandleResponse = async (methodName: string, methodParams: any[], methodFunction: any) => {
+  const requestDetails = app.getRequestDetails();
+  logger.debug('Method name: ' + methodName);
+  logger.debug('Request id: ' + requestDetails.requestId);
+  logger.debug('Request ip: ' + requestDetails.ipAddress);
```
Nana-EC, line 204: drop this line for now

packages/server/src/server.ts
```diff
@@ -197,26 +197,29 @@ app.getKoaApp().use(async (ctx, next) => {
   return next();
 });
 
-const logAndHandleResponse = async (methodName: any, methodParams: any, methodFunction: any) => {
-  const requestId = app.getRequestId();
-  const requestIdPrefix = requestId ? formatRequestIdMessage(requestId) : '';
+const logAndHandleResponse = async (methodName: string, methodParams: any[], methodFunction: any) => {
+  const requestDetails = app.getRequestDetails();
+  logger.debug('Method name: ' + methodName);
```
Nana-EC, line 202: Squash this into 1 line if needed.
That being said do we need it?
We should have a log that covers the method name no as well as the result

Nana-EC (commented):
Left some thoughts

packages/server/src/koaJsonRpc/index.ts
konstantinabl: @Nana-EC  the idea here is that we need this for the hbar limiter and spending plan, so we always want to have the sender's ip in order to check if its a part of the spending plan of a partner. so its not related to rate limiting but hbar limiting and we need it passed to sdkClient and the functions there

packages/server/src/server.ts
konstantinabl: yeah this was needed for debugging purposes only. will remove it

packages/server/src/koaJsonRpc/index.ts
Nana-EC: My point is if IP based hbar limiting isn't enabled then this should be skipped.

packages/relay/src/lib/types/RequestDetails.ts
```diff
@@ -0,0 +1,48 @@
+/*
+ *
+ * Hedera JSON RPC Relay
+ *
+ * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
```
Nana-EC, line 5: nit:
```suggestion
 * Copyright (C) 2024 Hedera Hashgraph, LLC
```

packages/server/src/koaJsonRpc/index.ts
victor-yanev: If we want to disable IP-based limiting we can do it inside the `HbarLimitService` for better locality of behavior.

This PR which wraps the `requestId` into a `requestDetails` object should be agnostic of the hbar limiter. This is more of an overall change for the whole project. We can still have access to the IP address, but only conditionally use it inside `HbarLimitService`.

victor-yanev (dismissed):

victor-yanev (dismissed):

natanasow (approved):
LG

Nana-EC (approved):

packages/relay/src/lib/eth.ts
```diff
@@ -1072,12 +1080,13 @@ export class EthImpl implements Eth {
 
       // save in cache the current balance for the account and blockNumberOrTag
       cachedBalance = numberTo0x(weibars);
+      this.logger.trace('Value cached balance', cachedBalance);
```
quiet-node, line 1083: formattedRequestId can be added in this log

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -113,7 +113,7 @@ export default class MetricService {
     this.consensusNodeClientHistogramGasFee = this.initGasMetric(register);
 
     this.eventEmitter.on(constants.EVENTS.EXECUTE_TRANSACTION, (args: IExecuteTransactionEventPayload) => {
-      this.captureTransactionMetrics(args);
+      this.captureTransactionMetrics(args).then();
```
quiet-node, line 116: qq: why do we need a .then() here?

packages/relay/src/lib/types/RequestDetails.ts
```diff
@@ -0,0 +1,101 @@
...
+   * The connection ID associated with the request (optional).
+   * @type {string | undefined}
+   */
+  connectionId?: string;
+}
+
+/**
+ * Represents the details of a request.
+ */
+export class RequestDetails {
```
quiet-node, line 47: Just a suggestion, thinking out loud — Personally, since RequestDetails is a full class definition, which isn't typically something I'd expect to find in a `types/` folder.

I would recommend the following changes:
1. Move the `RequestDetails` class to a more appropriate location, such as a `models` or `entities` folder (`packages/relay/src/lib/models/` or `packages/relay/src/lib/entities/`).
2. Keep the `IRequestDetails` interface in the `types/` folder and rename the file to `IRequestDetails.ts`.

However, this would involve updating all import statements, given that this file impacts the entire codebase. I'd suggest handling this update in a separate PR, as the current one is already quite large.

quiet-node (dismissed):
Excellent work, @konstantinabl! Thank you for the tremendous effort you’ve put into this PR! I really appreciate how meticulous the updates and changes are. Running the test suite and reviewing the logs was a pleasure—it's impressive to see that every single log includes the request ID now, and all IDs are formatted consistently.

I’ve left some questions and suggestions that can be addressed in follow-up PRs.

Fantastic work!

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -166,7 +164,7 @@ export class MirrorNodeClient {
     // defualt values for axios clients to mirror node
     const mirrorNodeTimeout = parseInt(process.env.MIRROR_NODE_TIMEOUT || '10000');
     const mirrorNodeMaxRedirects = parseInt(process.env.MIRROR_NODE_MAX_REDIRECTS || '5');
-    const mirrorNodeHttpKeepAlive = process.env.MIRROR_NODE_HTTP_KEEP_ALIVE === 'false' ? false : true;
+    const mirrorNodeHttpKeepAlive = process.env.MIRROR_NODE_HTTP_KEEP_ALIVE !== 'false';
```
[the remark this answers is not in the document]
konstantinabl: @quiet-node  Yes, it does imply this and what was done here is just a simplification, not a change of the logic. So it was meant to be true when undefined before, sonar suggested this simplification

packages/relay/src/lib/types/RequestDetails.ts
konstantinabl: Okay, may be addressed in a separate PR

packages/relay/src/lib/services/metricService/metricService.ts
victor-yanev: If you're calling an async function without awaiting it, Sonar suggests adding a `.then()` or `.catch()` to make it explicit that the function is intentionally not awaited. This signals to other developers (or to static analysis tools like Sonar) that you are aware of the behaviour and that this is intentional. Otherwise, they would flag a warning for “missing await on async function”.

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -358,7 +358,9 @@ export class HbarLimitService implements IHbarLimitService {
       try {
         return await this.getSpendingPlanByIPAddress(requestDetails);
       } catch (error) {
-        this.logger.warn(error, `Failed to get spending plan for IP address '${ipAddress}'`);
+        const spendingPlanId = (await this.ipAddressHbarSpendingPlanRepository.findByAddress(ipAddress, requestDetails))
```
victor-yanev, line 361: This would always be `undefined`, no point in doing this fetch.

victor-yanev (changes requested):

packages/relay/src/lib/services/hbarLimitService/index.ts
konstantinabl: ah, yes you are right. i was trying to fix this quickly and overlooked

konstantinabl: Removed

packages/relay/src/lib/clients/mirrorNodeClient.ts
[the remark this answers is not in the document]
quiet-node: Gotcha thanks for clarifying!

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: Gotcha thanks for clarifying!

quiet-node (approved):
LGTM

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -67,10 +68,10 @@ export class IPAddressHbarSpendingPlanRepository {
    * @param {IIPAddressHbarSpendingPlan} addressPlan - The plan to save.
    * @returns {Promise<void>} - A promise that resolves when the IP address is linked to the plan.
    */
-  async save(addressPlan: IIPAddressHbarSpendingPlan): Promise<void> {
+  async save(addressPlan: IIPAddressHbarSpendingPlan, requestDetails: RequestDetails): Promise<void> {
     const key = this.getKey(addressPlan.ipAddress);
-    await this.cache.set(key, addressPlan, 'save', this.threeMonthsInMillis);
-    this.logger.trace(`Saved IPAddressHbarSpendingPlan with address ${addressPlan.ipAddress}`);
+    await this.cache.set(key, addressPlan, 'save', requestDetails, this.threeMonthsInMillis);
+    this.logger.trace(`Saved IPAddressHbarSpendingPlan with ID ${addressPlan.planId}`);
```
victor-yanev, line 74:
```suggestion
    this.logger.trace(`Linked new IP address to HbarSpendingPlan with ID ${addressPlan.planId}`);
```

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -51,13 +52,13 @@ export class IPAddressHbarSpendingPlanRepository {
...
-  async findByAddress(ipAddress: string): Promise<IPAddressHbarSpendingPlan> {
+  async findByAddress(ipAddress: string, requestDetails: RequestDetails): Promise<IPAddressHbarSpendingPlan> {
     const key = this.getKey(ipAddress);
-    const addressPlan = await this.cache.getAsync<IIPAddressHbarSpendingPlan>(key, 'findByAddress');
+    const addressPlan = await this.cache.getAsync<IIPAddressHbarSpendingPlan>(key, 'findByAddress', requestDetails);
     if (!addressPlan) {
       throw new IPAddressHbarSpendingPlanNotFoundError(ipAddress);
     }
-    this.logger.trace(`Retrieved IPAddressHbarSpendingPlan with address ${ipAddress}`);
+    this.logger.trace(`Retrieved IPAddressHbarSpendingPlan with ID ${addressPlan.planId}`);
```
victor-yanev, line 61:
```suggestion
    this.logger.trace(`Retrieved link between IP address and HbarSpendingPlan with ID ${addressPlan.planId}`);
```

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -79,10 +80,14 @@ export class IPAddressHbarSpendingPlanRepository {
    * @param {string} ipAddress - The IP address to unlink the plan from.
    * @returns {Promise<void>} - A promise that resolves when the IP address is unlinked from the plan.
    */
-  async delete(ipAddress: string): Promise<void> {
+  async delete(ipAddress: string, requestDetails: RequestDetails): Promise<void> {
     const key = this.getKey(ipAddress);
-    await this.cache.delete(key, 'delete');
-    this.logger.trace(`Deleted IPAddressHbarSpendingPlan with address ${ipAddress}`);
+    const spendingPlan = await this.cache.getAsync<IIPAddressHbarSpendingPlan>(key, 'findByAddress', requestDetails);
```
victor-yanev, line 85:
```suggestion
    const ipAddressSpendingPlan = await this.cache.getAsync<IIPAddressHbarSpendingPlan>(key, 'delete', requestDetails);
```

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
konstantinabl: I would say this is kinda "overengineering" it
In terms of logs its good to have abstraction, so we basically want to "save" a plan, the method is also called save, not link

victor-yanev: Sure, but `Saved IPAddressHbarSpendingPlan with ID ${addressPlan.planId}` is conceptually wrong because this is the ID of the `HbarSpendingPlan` and not the ID of the `IPAddressHbarSpendingPlan` - it is an entity which represents the mapping between `ipAddress` and `planId`, so it is identified by a combination of (ipAddress, planId)

victor-yanev: To avoid confusion, we can change this to:
```suggestion
    this.logger.trace(`Saved IPAddressHbarSpendingPlan for HbarSpendingPlan with ID ${addressPlan.planId}`);
```

victor-yanev: But see, how my previous suggestion is much easier to follow through and understand what is happening:
`Linked new IP address to HbarSpendingPlan with ID ${addressPlan.planId}`

victor-yanev: That's why I suggested changing the style of these logs, they are not really readable unless you know what each entity represents and can be confusing to people reading them, especially the log about deleting a link between IP address and spending plan https://github.com/hashgraph/hedera-json-rpc-relay/pull/2939#discussion_r1784924436

konstantinabl: Changed it.

natanasow (approved):

victor-yanev (approved):

PR #2952, opened 2024-09-09
fix: improve `RedisCache` and simplify usage of `RedisInMemoryServer` in tests

victor-yanev:
**Description**:

This PR improves the current state of the `CacheService` and simplifies the usage of the Redis in-memory server within the project. 

The following changes have been made:
- Updated `CacheService` with error handling for the `disconnectRedisClient()` method
- Enhanced `CacheService` with new methods:
  - `getNumberOfRedisConnections()`
  - `isRedisClientConnected()`
  - `connectRedisClient()`
- Refactored `RedisInMemoryServer` Class:  
  - Added detailed logging for start and stop operations.
  - Improved error handling to log appropriate messages when the server fails to start or stop.
- Enhanced Test Helpers:  
   - Introduced `useInMemoryRedisServer` function to streamline the setup and teardown of the Redis in-memory server in tests.
   - Added `startRedisInMemoryServer` and `stopRedisInMemoryServer` functions to manage the Redis server lifecycle and environment variables.
- Test Suite Enhancements:
  - Extended tests in `redisCacche.spec.ts` to ensure proper functionality of all new methods added.
  - Extended tests in `cacheService.spec.ts` to ensure proper functionality for all new methods added.

**Notes for reviewer**:

No breaking changes introduced. All existing functionalities remain intact.

**Checklist**:

  - [x] Updated existing tests to use the new helper functions for managing the Redis in-memory server.
  - [x] Verified that the Redis in-memory server starts and stops correctly.
  - [x] Ensured that all new methods are working properly:
    - [x] Extended tests in `redisCacche.spec.ts` with scenarios for all new methods.
    - [x] Extended tests in `cacheService.spec.ts` with scenarios for for all new methods.
  - [x] Ensured that fallback to internal cache works as expected when Redis operations fail.
    - [x] Added test cases for missing scenarios in `cacheService.spec.ts` to verify Redis errors are handled correctly and all methods fallback to the internal cache.


packages/relay/src/lib/clients/cache/redisCache.ts
```diff
@@ -249,10 +249,10 @@ export class RedisCache implements IRedisCacheClient {
...
    */
   async disconnect(): Promise<void> {
-    await this.getConnectedClient().then((client) => {
-      client.disconnect();
-      client.unsubscribe();
-    });
+    if (await this.connected) {
+      const client = await this.getConnectedClient();
+      await client.disconnect();
+    }
```
ebadiere, line 255: Is there a reason we are using `client.disconnect()` instead of `client.quit()` ?  I thought `quit` was cleaner.  It waits for pending replies to be received before closing.

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -120,7 +120,12 @@ export class CacheService {
...
   async disconnectRedisClient(): Promise<void> {
     if (this.sharedCache instanceof RedisCache) {
-      await this.sharedCache.disconnect();
+      try {
+        await this.sharedCache.disconnect();
+      } catch (e) {
+        const redisError = new RedisCacheError(e);
+        this.logger.error(redisError, `Error occurred when disconnecting from Redis.`);
+      }
     }
```
ebadiere, line 129: Same question as above, quit vs disconnet?

packages/relay/tests/lib/services/cacheService/cacheService.spec.ts
```diff
@@ -244,27 +244,30 @@ describe('CacheService Test Suite', async function () {
...
-    });
+      disconnected = true;
+    };
 
     this.beforeEach(async () => {
-      await cacheService.clear();
+      multiSet = process.env.MULTI_SET;
+      process.env.MULTI_SET = 'true';
+      cacheService = new CacheService(logger.child({ name: 'cache-service' }), registry);
+      disconnected = false;
```
ebadiere, line 262: Would it be worth exploring adding some state to the cache service, such as `isConnected`, number of connections, etc, instead of tracking in the test like this?  Asking because we once ran into an issue with the relay and the number of connections to the shared cache.

packages/relay/tests/redisInMemoryServer.ts
```diff
@@ -8,67 +28,62 @@ export class RedisInMemoryServer {
...
+    this.logger.trace('Starting Redis in-memory server....');
+    const started = await this.inMemoryRedisServer.start();
+    if (started) {
+      const host = await this.getHost();
+      const port = await this.getPort();
+      this.logger.info(`Started Redis in-memory server on ${host}:${port} successfully.`);
+    } else {
+      this.logger.error("Couldn't start Redis in-memory server successfully.");
+    }
   }
```
ebadiere, line 66: Could this benefit from a health check?  It might be handy if it fails in CI.
```
private async performHealthCheck(): Promise<void> {
  try {
    // Implement a simple health check, e.g., ping the server
    const client = await this.getClient();
    await client.ping();
    this.logger.debug('Redis in-memory server health check passed');
  } catch (error) {
    this.logger.warn(`Redis in-memory server health check failed: ${error.message}`);
  }
}
```

ebadiere (commented):
Thank you for doing this.  I have a couple of questions.

packages/relay/src/lib/clients/cache/redisCache.ts
```diff
@@ -249,10 +249,10 @@ export class RedisCache implements IRedisCacheClient {
    * @returns {Promise<void>} A Promise that resolves when the client is disconnected.
    */
   async disconnect(): Promise<void> {
-    await this.getConnectedClient().then((client) => {
-      client.disconnect();
-      client.unsubscribe();
-    });
+    if (await this.connected) {
```
konstantinabl, line 252: why check this.connected when getConnectedClient actually checks it ?

packages/relay/src/lib/clients/cache/redisCache.ts
victor-yanev: Thanks for pointing this out! I changed it to use `quit()`.

packages/relay/src/lib/services/cacheService/cacheService.ts
victor-yanev: https://github.com/hashgraph/hedera-json-rpc-relay/pull/2952#discussion_r1753832210

packages/relay/tests/lib/services/cacheService/cacheService.spec.ts
victor-yanev: Done, added new methods for:
- `isConnected()`
- `getNumberOfConnections()`

packages/relay/tests/redisInMemoryServer.ts
victor-yanev: There is no `pinging` in this case, because `ping()` is called on the side of the client to check if the server is running and the `RedisInMemoryServer` class is responsible for starting the in-memory server, so it's not a client. 

But I added something like a fallback "retry" mechanism, in case the redis in-memory server couldn't be started:
```typescript
  async start(): Promise<void> {
    this.logger.trace('Starting Redis in-memory server....');
    const started = await this.inMemoryRedisServer.start();
    if (started) {
      const host = await this.getHost();
      const port = await this.getPort();
      this.logger.info(`Started Redis in-memory server on ${host}:${port} successfully.`);
    } else {
      this.logger.error("Couldn't start Redis in-memory server successfully.");
      await this.performHealthCheck();
    }
  }

  private async performHealthCheck(): Promise<void> {
    try {
      // Ensure that the instance is running -> throws error if instance cannot be started
      const instanceData = await this.inMemoryRedisServer.ensureInstance();
      this.logger.debug(`Redis in-memory server health check passed, server is running on port ${instanceData.port}.`);
    } catch (error) {
      this.logger.warn(`Redis in-memory server health check failed: ${error}`);
    }
  }
```

packages/relay/src/lib/clients/cache/redisCache.ts
victor-yanev: Good point, I removed this redundant check.

konstantinabl (approved):

ebadiere (approved):
LG

PR #2956, opened 2024-09-10
feat: capture env variables into sigleton and reusable constants

### LINKED ISSUE #2686: Capture env variables in reusable constants
### Problem

Currently, we are dynamically getting the environment variables and parsing/casting them every time. This is inefficient and also, more importantly, allows env configurations to be dynamically changed, which can modify the behaviour of the relay based on the current values of the env vars, which can be a potential exploit.

### Solution

- We should change this to only capture them once when the relay is initialised and reuse those as constants

### Alternatives

_No response_

natanasow:
**Description**:

Currently, we are dynamically getting the environment variables and parsing/casting them every time. This is inefficient and also, more importantly, allows env configurations to be dynamically changed, which can modify the behavior of the relay based on the current values of the env vars, which can be a potential exploit.

**Solution**:

We should change this to only capture them once when the relay is initialized and reuse those as constants

**Related issue(s)**:

Fixes #2686 

**Notes for reviewer**:

- created a new package that holds the `env-provider` singleton
- the `env-provider` package is added as a dependency to `relay`, `server`, and `ws-server`
- in the perfect world, the `env-provider` must be immutable and should not provide an interface for overriding or extending once-loaded environment variables but there are many tests where we dynamically override envs in order to test specific cases so we should add these interfaces as well
- the main review should be applied to the `env-provider` package
- almost all of the changes are just replaced values from `process.env.<VAR_NAME>` to `EnvProviderService.getInstance().get(<VAR_NAME>)`
- in the test cases if overrides are needed, they are done like that - `process.env.CHAIN_ID = '0x160c'` is transformed to  `EnvProviderService.getInstance().dynamicOverride('CHAIN_ID', '0x160c')`

**Checklist**


packages/env-provider/src/services/index.ts
```diff
@@ -0,0 +1,107 @@
...
+   * The singleton instance
+   * @public
+   */
+  private static instance: IEnvProviderService;
+
+  /**
+   * Copied envs from process.env
+   * @private
+   */
+  private envs: JSON;
```
victor-yanev, line 36: This property is declared as type `JSON`, which is incorrect. `JSON` is not a valid type for an object; instead, you should type it as `Record<string, string | undefined>` or `NodeJS.ProcessEnv` since `process.env` is inherently a `Record<string, string | undefined>`.

packages/env-provider/src/services/index.ts
```diff
@@ -0,0 +1,107 @@
...
+  private constructor() {
+    dotenv.config({ path: findConfig('.env') || '' });
+    this.envs = JSON.parse(JSON.stringify(process.env));
+  }
+
+  /**
+   * Get the singleton instance of the current service
+   * @public
+   */
+  public static getInstance(): IEnvProviderService {
```
victor-yanev, line 51: Let's make this method private and add a static `get` method which uses it:
```typescript
public static get(name: string): string | undefined {
  return EnvProvider.getInstance().get(name);
}
```

In this way usages of this `EnvProvider` class would be shorter and simpler to follow and we avoid the need to chain `getInstance()` every time we want to use any of the methods:

```typescript
const shouldDefaultToConsensus = EnvProvider.get('ETH_CALL_DEFAULT_TO_CONSENSUS_NODE') === 'true';
```

We can also do the same for `remove`, `dynamicOverride` and `appendEnvsFromPath`.

packages/env-provider/src/services/index.ts
```diff
@@ -0,0 +1,107 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { IEnvProviderService } from './IEnvProviderService';
+import dotenv from 'dotenv';
+import findConfig from 'find-config';
+
+export class EnvProviderService implements IEnvProviderService {
```
victor-yanev, line 25: Let's make this shorter: `EnvProviderService` -> `EnvProvider`

packages/env-provider/src/services/index.ts
```diff
@@ -0,0 +1,107 @@
...
+   */
+  private envs: JSON;
+
+  /**
+   * Fetches all envs from process.env and pushes them into the envs property
+   * @private
+   */
+  private constructor() {
+    dotenv.config({ path: findConfig('.env') || '' });
+    this.envs = JSON.parse(JSON.stringify(process.env));
```
victor-yanev, line 44: QQ: Why are we stringifying and then parsing the `process.env`? 

If the purpose is to copy the `process.env` object, you can use object de-structuring since it's only containing strings, we don't need to do a "deep" clone with `JSON.parse` and `JSON.stringify`:
```suggestion
    this.envs = { ...process.env };
```

packages/env-provider/src/services/index.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Delete an env variable, used in test cases only
+   * @param name string
+   * @returns void
+   */
+  public remove(name: string): void {
+    delete this.envs[name];
+  }
```
victor-yanev, line 92: QQ: Why would we ever want to delete an env variable, isn't `hotReload` used for reloading the original configurations in those scenarios? 

Also declaring it as a public method which makes it accessible through **_any class in the source code_** only for testing purposes is not really advisable.

packages/env-provider/src/services/index.ts
victor-yanev: Let's also make it `readonly` - we should never want to override this.
```suggestion
  private readonly envs: Record<string, string | undefined>;
```

packages/env-provider/src/services/index.ts
```diff
@@ -0,0 +1,107 @@
...
+   * @returns void
+   */
+  public appendEnvsFromPath(configPath: string): void {
+    dotenv.config({ path: configPath, override: true });
+
+    const envsToAppend = JSON.parse(JSON.stringify(process.env));
+    const merged = Object.assign(this.envs, envsToAppend);
+
+    this.envs = JSON.parse(JSON.stringify(merged));
+  }
```
victor-yanev, line 106: Again, we don't need to use `JSON.parse` and `JSON.stringify` here, and we shouldn't replace the `this.envs` object, but only modify it:
```suggestion
  public appendEnvsFromPath(configPath: string): void {
    dotenv.config({ path: configPath, override: true });
    const instance = EnvProvider.getInstance();
    Object.entries(process.env).forEach(([key, value]) => {
      instance.dynamicOverride(key, value);
    });
  }
```

packages/env-provider/src/services/index.ts
```diff
@@ -0,0 +1,107 @@
...
+    return this.envs[name];
+  }
+
+  /**
+   * Override an env variable, used in test cases only
+   * @param name string
+   * @param value string
+   * @returns void
+   */
+  public dynamicOverride(name: string, value: string | undefined): void {
```
victor-yanev, line 81: We should move this together with `hotReload`, `remove` and `appendEnvsFromPath` to a helper class in the `test` directory, these shouldn't be accessible as public methods in the source code.

victor-yanev (changes requested):
Mostly looks good, thank you for putting in the effort! Let's keep in mind that the whole purpose of this issue is to make the env variables readonly and unmodifiable, hence I've left some suggestions:

packages/env-provider/src/services/index.ts
```diff
@@ -0,0 +1,107 @@
...
+   * @private
+   */
+  private envs: JSON;
+
+  /**
+   * Fetches all envs from process.env and pushes them into the envs property
+   * @private
+   */
+  private constructor() {
+    dotenv.config({ path: findConfig('.env') || '' });
```
victor-yanev, line 43: You are using `findConfig('.env') || ''`, which means that if no `.env` file is found, an empty string is passed to `dotenv.config()`. This isn't necessarily a problem, but it might be clearer to raise an error or log a warning when `.env` is not found, as this might not be the desired behavior.

packages/env-provider/src/services/index.ts
```diff
@@ -0,0 +1,107 @@
...
+  public static hotReload(): void {
+    this.instance = new EnvProviderService();
+  }
+
+  /**
+   * Get an env var by name
+   * @param name string
+   * @returns string | undefined
+   */
+  public get(name: string): string | undefined {
```
victor-yanev, line 71: We can implement a default value or fallback mechanism to handle missing environment variables:

```typescript
public getWithDefault(name: string, defaultValue: string): string {
  const value = this.envs[name];
  if (!value) {
    console.warn(`Environment variable ${name} is not set, using default: ${defaultValue}`);
    return defaultValue;
  }
  return value;
}
```

We can also throw an error or log a warning when a required environment variable is not set:

```typescript
public getRequired(name: string): string {
  const value = this.envs[name];
  if (!value) {
    throw new Error(`Missing required environment variable: ${name}`);
  }
  return value;
}
```

packages/env-provider/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@hashgraph/env-provider",
```
Nana-EC, line 2: For consistency
```suggestion
  "name": "@hashgraph/json-rpc-env-provider",
```

packages/env-provider/src/services/index.ts
Nana-EC: Agreed.
In fact the relay shouldn't start without a valid env file so this might be a good spot to fail fast

packages/env-provider/src/services/index.ts
Nana-EC: 100% this

Nana-EC (commented):
Nice work.
I agree with most of Victors comments

ebadiere (approved):
Thank you for doing this.  I think we should also have a --help option that can further explain this environment variables and perhaps include examples.  It should probably done in another github issue.

packages/env-provider/src/services/index.ts
AlfredoG87: I know am late to the comments, Could it even be `ConfigService`? does it make more sense?

AlfredoG87 (commented):
I agree with all comments from @victor-yanev and @Nana-EC

maybe I was not part of the discussion and design however, would like to express my opinions.

1. I agree that this should be used by all other packages ( relay, server and ws-server) hence the need to keep it as a separate package itself and not only as an internal service of the `server` package. 

2. When I originally thought about the need of this, some time ago, I did so thinking it as a a Service within the `server` and calling it `ConfigService` that had the following responsabilities:
1. Read `.env` file
2. Make sure all required configurations are present.
3. Store defaults and parse override.
4. "Know" what configurations are available for what services/packages.
5. Print at startup the onto the logs the `FINAL` values for all needed configurations so is easy to debug and troubleshoot, and using the logs an operator is able to see what configurations are being used. (Secrets should be obscured before printing, so the configService provider should also know which ones are secrets)
6. Provide an interface for getting such values.

I understand that the scope of this PR might be different and as I said, am late to the discussion, however would like my comments to be taken into consideration to see what it makes sense and what not, if anything I would strongly argue against as a just passtrough service to `GET` the env values, but to make it a full fledged Configuration Service provider that is aware of the config values, its defaults and exposes an Enum or record with an existing list of configs, this way we minimize the risk of duplications, we fail fast if parsing fails, and we print the values for the envs (except secrets those should be oscured)

packages/env-provider/src/services/index.ts
natanasow: Renamed.

natanasow:
> I agree with all comments from @victor-yanev and @Nana-EC
> 
> maybe I was not part of the discussion and design however, would like to express my opinions.
> 
> I agree that this should be used by all other packages ( relay, server and ws-server) hence the need to keep it as a separate package itself and not only as an internal service of the `server` package.
> 
> When I thought about the need of this, did so thinking it as a a Service and calling it `ConfigService` that had the following responsibilities:
> 
>     1. Read `.env` file
> 
>     2. Make sure all required configurations are present.
> 
>     3. Store defaults and parse override.
> 
>     4. "Know" what configurations are available for what services/packages.
> 
>     5. Print at startup the onto the logs the `FINAL` values for all needed configurations so is easy to debug and troubleshoot, and using the logs an operator is able to see what configurations are being used. (Secrets should be obscured before printing, so the configService provider should also know which ones are secrets)
> 
>     6. Provide an interface for getting such values.
> 
> 
> I understand that the scope of this PR might be different and as I said, am late to the discussion, however would like my comments to be taken into consideration to see what it makes sense and what not, if anything I would strongly argue against as a just passtrough service to `GET` the env values, but to make it a full fledged Configuration Service provider that is aware of the config values, its defaults and exposes an Enum or record with an existing list of configs, this way we minimize the risk of duplications, we fail fast if parsing fails, and we print the values for the envs (except secrets those should be oscured)
> 
> example of usage in my mind: instead of:
> 
> ```
> const shouldDefaultToConsensus = EnvProvider.get('ETH_CALL_DEFAULT_TO_CONSENSUS_NODE') === 'true';
> ```
> 
> it would be:
> 
> ```
> const shouldDefaultToConsensus = ConfigService.get(RelayConfigs.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE);
> ```
> 
> (where RelayConfigs is a known enum) and the values are guaranteed to be present by either defaults overrides or fast failing at (ServiceConfig startup).

natanasow:
> I agree with all comments from @victor-yanev and @Nana-EC
> 
> maybe I was not part of the discussion and design however, would like to express my opinions.
> 
> I agree that this should be used by all other packages ( relay, server and ws-server) hence the need to keep it as a separate package itself and not only as an internal service of the `server` package.
> 
> When I thought about the need of this, did so thinking it as a a Service and calling it `ConfigService` that had the following responsibilities:
> 
>     1. Read `.env` file
> 
>     2. Make sure all required configurations are present.
> 
>     3. Store defaults and parse override.
> 
>     4. "Know" what configurations are available for what services/packages.
> 
>     5. Print at startup the onto the logs the `FINAL` values for all needed configurations so is easy to debug and troubleshoot, and using the logs an operator is able to see what configurations are being used. (Secrets should be obscured before printing, so the configService provider should also know which ones are secrets)
> 
>     6. Provide an interface for getting such values.
> 
> 
> I understand that the scope of this PR might be different and as I said, am late to the discussion, however would like my comments to be taken into consideration to see what it makes sense and what not, if anything I would strongly argue against as a just passtrough service to `GET` the env values, but to make it a full fledged Configuration Service provider that is aware of the config values, its defaults and exposes an Enum or record with an existing list of configs, this way we minimize the risk of duplications, we fail fast if parsing fails, and we print the values for the envs (except secrets those should be oscured)
> 
> example of usage in my mind: instead of:
> 
> ```
> const shouldDefaultToConsensus = EnvProvider.get('ETH_CALL_DEFAULT_TO_CONSENSUS_NODE') === 'true';
> ```
> 
> it would be:
> 
> ```
> const shouldDefaultToConsensus = ConfigService.get(RelayConfigs.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE);
> ```
> 
> (where RelayConfigs is a known enum) and the values are guaranteed to be present by either defaults overrides or fast failing at (ServiceConfig startup).



> I agree with all comments from @victor-yanev and @Nana-EC
> 
> maybe I was not part of the discussion and design however, would like to express my opinions.
> 
> I agree that this should be used by all other packages ( relay, server and ws-server) hence the need to keep it as a separate package itself and not only as an internal service of the `server` package.
> 
> When I thought about the need of this, did so thinking it as a a Service and calling it `ConfigService` that had the following responsibilities:
> 
>     1. Read `.env` file
> 
>     2. Make sure all required configurations are present.
> 
>     3. Store defaults and parse override.
> 
>     4. "Know" what configurations are available for what services/packages.
> 
>     5. Print at startup the onto the logs the `FINAL` values for all needed configurations so is easy to debug and troubleshoot, and using the logs an operator is able to see what configurations are being used. (Secrets should be obscured before printing, so the configService provider should also know which ones are secrets)
> 
>     6. Provide an interface for getting such values.
> 
> 
> I understand that the scope of this PR might be different and as I said, am late to the discussion, however would like my comments to be taken into consideration to see what it makes sense and what not, if anything I would strongly argue against as a just passtrough service to `GET` the env values, but to make it a full fledged Configuration Service provider that is aware of the config values, its defaults and exposes an Enum or record with an existing list of configs, this way we minimize the risk of duplications, we fail fast if parsing fails, and we print the values for the envs (except secrets those should be oscured)
> 
> example of usage in my mind: instead of:
> 
> ```
> const shouldDefaultToConsensus = EnvProvider.get('ETH_CALL_DEFAULT_TO_CONSENSUS_NODE') === 'true';
> ```
> 
> it would be:
> 
> ```
> const shouldDefaultToConsensus = ConfigService.get(RelayConfigs.ETH_CALL_DEFAULT_TO_CONSENSUS_NODE);
> ```
> 
> (where RelayConfigs is a known enum) and the values are guaranteed to be present by either defaults overrides or fast failing at (ServiceConfig startup).


Solid observations and proposals :pray:. Our next actions are defined here https://github.com/hashgraph/hedera-json-rpc-relay/issues/3023 and they cover almost all of your concerns. We'll stick to your proposals and ping you for a PR review, once we're ready. It'll be done in another PR because that one will become enormously bigger.

cc: @ebadiere

konstantinabl (changes requested):

PR #2981, opened 2024-09-13
feat: enabled retrieval of network `gasPrice` in `sendRawTransaction` using MAPI instead of HAPI.

### LINKED ISSUE #2979: Query the mirror node for gas fees
### Problem

Consensus node queries cost HBar, but mirror node queries are free.

### Solution

Query the mirror node for gas fees.

### Alternatives

_No response_

quiet-node:
**Description**:

This PR enables `sendRawTransaction` to retrieve the current `gasPrice` of the network via the mirror node, instead of querying the consensus node for the fee schedule file. By switching to MAPI, the relay operator account can avoid incurring costs associated with MAPI queries.

**Related issue(s)**:

Fixes #2979 

**Notes for reviewer**:
This PR utilizes the `gasPrice()` method in the `eth.ts` module for `sendRawTransaction` to take advantage of querying the `gasPrice` through the mirror node. If that query fails, it falls back to querying the consensus node for the fee schedule using the `getTinybarGasFee` method in the `sdkClient`.

**Checklist**

### DIFF
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -132,24 +132,23 @@ const parseNumericEnvVar = (envVarName: string, fallbackConstantKey: string): nu
 };
 
 /**
- * Parse weibar hex string to tinybar number, by applying tinybar to weibar coef.
- * Return null, if value is not a valid hex. Null is the only other valid response that mirror-node accepts.
- * @param value
- * @returns tinybarValue
+ * Converts a value in weibars (as a hex string, bigint, boolean, number, or string) to tinybars.
+ *
+ * @param {bigint | boolean | number | string} value - The value to convert, can be in various formats such as hex string, bigint, boolean, number, or string.
+ * @returns {number} - The equivalent value in tinybars, rounded to the nearest whole number.
  */
-const weibarHexToTinyBarInt = (value: bigint | boolean | number | string | null | undefined): number | null => {
-  if (value && value !== '0x') {
-    const weiBigInt = BigInt(value);
-    const coefBigInt = BigInt(constants.TINYBAR_TO_WEIBAR_COEF);
-    // Calculate the tinybar value
-    const tinybarValue = weiBigInt / coefBigInt;
-    // Check if there was a fractional part that got discarded
-    if (tinybarValue === BigInt(0) && weiBigInt > BigInt(0)) {
-      return 1; // Round up to the smallest unit of tinybar
-    }
-    return Number(tinybarValue);
+const weibarHexToTinyBarInt = (value: bigint | boolean | number | string): number => {
+  if (value === '0x') return 0;
+
+  const weiBigInt = BigInt(value);
+  const coefBigInt = BigInt(constants.TINYBAR_TO_WEIBAR_COEF);
+  // Calculate the tinybar value
+  const tinybarValue = weiBigInt / coefBigInt;
+  // Check if there was a fractional part that got discarded
+  if (tinybarValue === BigInt(0) && weiBigInt > BigInt(0)) {
+    return 1; // Round up to the smallest unit of tinybar
   }
-  return null;
+  return Number(tinybarValue);
 };
 
 const formatContractResult = (cr: any) => {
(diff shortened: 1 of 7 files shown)


ebadiere (dismissed):
LG.

packages/relay/src/lib/eth.ts
```diff
@@ -1537,7 +1544,8 @@ export class EthImpl implements Eth {
         .labels(EthImpl.ethSendRawTransaction, transaction.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH))
         .inc();
 
-    const parsedTx = await this.parseRawTxAndPrecheck(transaction, requestIdPrefix);
+    const networkGasPriceInWeiBars = await this.gasPrice(requestIdPrefix);
```
Nana-EC, line 1547: Q: Is there any dead code that needs to be removed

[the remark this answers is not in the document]
quiet-node: Yeah that's a solid concern! Thanks for pointing it out! Updated using the internal getFeeWeiBars()

quiet-node: I also looked into this and for now there doesn't appear to be any dead code. Everything seems to be in order for now. 🤞🏼

Nana-EC (approved):

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -432,9 +437,11 @@ export class SDKClient {
       ethereumTransactionData.callData = new Uint8Array();
       ethereumTransaction.setEthereumData(ethereumTransactionData.toBytes()).setCallDataFileId(fileId);
     }
+    const networkGasPriceInTinyBars = weibarHexToTinyBarInt(networkGasPriceInWeiBars)!;
```
victor-yanev, line 440: QQ: In which cases is this going to be `null` or `undefined`?

quiet-node: The `weibarHexToTinyBarInt(value)` returns `null` if value is `null` | `undefined` or the value is an empty hash, `0x`.

victor-yanev: In this case let’s handle this more explicitly by doing a check instead of using `!` here:
```typescript
    const networkGasPriceInTinyBars = weibarHexToTinyBarInt(networkGasPriceInWeiBars);
    if (!networkGasPriceInTinyBars) {
      throw predefined.SOME_RELEVANT_ERROR_HERE;
    }
```

quiet-node: Yeah that's a good point. I originally thought since the `networkGasPriceInWeiBars` was a number and a `predefine.COULD_NOT_ESTIMATE_GAS_PRICE` has already been thrown but you're right we should also handle the error here as well. Pushing in the fix soon.

ebadiere (approved):
LG

ebadiere (approved):

packages/relay/src/formatters.ts
```diff
@@ -132,24 +132,23 @@ const parseNumericEnvVar = (envVarName: string, fallbackConstantKey: string): nu
...
-    const coefBigInt = BigInt(constants.TINYBAR_TO_WEIBAR_COEF);
-    // Calculate the tinybar value
-    const tinybarValue = weiBigInt / coefBigInt;
-    // Check if there was a fractional part that got discarded
-    if (tinybarValue === BigInt(0) && weiBigInt > BigInt(0)) {
-      return 1; // Round up to the smallest unit of tinybar
-    }
-    return Number(tinybarValue);
+const weibarHexToTinyBarInt = (value: bigint | boolean | number | string): number => {
+  if (value === '0x') return 0;
```
victor-yanev, line 141: **nit:** I think we can use the `EMPTY_HEX` constant here:

https://github.com/hashgraph/hedera-json-rpc-relay/blob/f80b338bbe9158dd33183712ba074845ad312c4a/packages/relay/src/formatters.ts#L29

victor-yanev (approved):

victor-yanev (changes requested):
I'm not sure why we want to set the `maxTransactionFee` to `MAX_GAS_PER_SEC`. 🤔 

Is this a realistic scenario for the relay operator to pay that much for a transaction?

PR #2828, opened 2024-08-15
feat: Add workaround for support of isAssociated

Ivo-Yankov:
**Description**:
- Adds a temporary workaround that enables the support of direct `isAssociated` calls to the relay.
- Adds a new feature flag `ETH_CALL_FORCE_TO_CONSENSUS_BY_SELECTOR` that enables some requests to `eth_call` to go through the Consensus node instead of the Mirror node regardless of the value of `ETH_CALL_DEFAULT_TO_CONSENSUS_NODE`. When set to `true` if the function selector is found in a predetermined array then the call will go through Consensus. For now the only entry in this list is `0x4d8fdd6d`, which is the selector of `isAssociated()`

**Related issue(s)**:

Fixes #2810

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #2983, opened 2024-09-13
feat: improved eth call force to consensus by selector logic

### LINKED ISSUE #2982: Improve the eth call force to consensus by selector logic
### Problem

The force-to-consensus-by-selector-logic feature enables `eth_call` to be redirected to a consensus node rather than a mirror node if the transaction's selector is considered special. This functionality was recently implemented in pull request [#2828](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2828) as an attempt to address issues related to the special selector for `isAssociated()`, which was introduced as a workaround for pull request [#905](https://github.com/hashgraph/hedera-smart-contracts/pull/905). However, the special selector has been added as a constant and does not appear to match the correct sighash for the `isAssociated()` selector.

### Solution

- Convert the `ETH_CALL_SELECTORS_ALWAYS_TO_CONSENSUS` constant into an environment configuration variable, allowing for easier addition or removal of special selectors.

### Alternatives

_No response_

quiet-node:
**Description**:
This PR improves the `force to consensus by selector` logic in `eth_call`:
- Moved the `ETH_CALL_SELECTORS_ALWAYS_TO_CONSENSUS` constant to environment configurations.
- Added an E2E test to demonstrate the force-to-consensus-by-selector flow, utilizing the latest `isAssociated` update from HIP-719.

**Related issue(s)**: https://github.com/hashgraph/hedera-smart-contracts/pull/905 https://github.com/hashgraph/hedera-json-rpc-relay/pull/2828

Fixes #2982 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -1643,9 +1643,11 @@ export class EthImpl implements Eth {
 
     const selector = getFunctionSelector(call.data!);
 
+    // The force-to-consensus-by-selector logic is active only when process.env.ETH_CALL_FORCE_TO_CONSENSUS_BY_SELECTOR is set to 'true'.
+    // When eth_call is invoked with a selector listed in specialSelectors, it will be routed through the consensus node, regardless of ETH_CALL_DEFAULT_TO_CONSENSUS_NODE.
```
lukelee-sl, line 1647: Can we add a note to say that this is a temporary measure and will be removed once mirror node can handle eth_call for `isAssociated`?

lukelee-sl (approved):
Looks good to me.  Adding a small comment might help to remind us to remove this when it is no longer necessary

packages/server/tests/contracts/HRC719Contract.json
acuarica, line 1: Shouldn't we include the `HRC719Contract.sol` here? Or at least mention how this bytecode was generated?

I see the corresponding [contract](https://github.com/hashgraph/hedera-smart-contracts/blob/6fedf8fd5a17db475ac7d0a8630bf8adb7cb6d0d/contracts/system-contracts/hedera-token-service/examples/hrc-719/HRC719Contract.sol) on the `hedera-smart-contracts` repo does not have the `isAssociated` method, which is included here.

lukelee-sl: I think that addition may be part of this pr - https://github.com/hashgraph/hedera-smart-contracts/pull/905 but Logan would know for sure.

acuarica: Thanks for the heads up. Given that we have the `.sol` for each `.json` output already (for some contracts at least), we could include the `.sol` here as well for consistency.

packages/relay/src/lib/eth.ts
quiet-node: Ah yeah I just finished creating [a ticket](https://github.com/hashgraph/hedera-json-rpc-relay/issues/2984) just for that. It's a good idea I'll include the ticket in the note.

packages/server/tests/contracts/HRC719Contract.json
quiet-node: Ah yeah good point will include associated solidity contract.

packages/relay/src/lib/eth.ts
```diff
@@ -1643,9 +1643,13 @@ export class EthImpl implements Eth {
...
+    // note: this feature is a workaround for when a feature is supported by consensus node but not yet by mirror node.
+    // Follow this ticket https://github.com/hashgraph/hedera-json-rpc-relay/issues/2984 to revisit and remove special selectors.
+    const specialSelectors = process.env.ETH_CALL_CONSENSUS_SELECTORS || [''];
     const shouldForceToConsensus =
-      process.env.ETH_CALL_FORCE_TO_CONSENSUS_BY_SELECTOR == 'true' &&
-      constants.ETH_CALL_SELECTORS_ALWAYS_TO_CONSENSUS.indexOf(selector) !== -1;
+      process.env.ETH_CALL_FORCE_TO_CONSENSUS_BY_SELECTOR === 'true' && specialSelectors.includes(selector);
```
Ivo-Yankov, line 1652: Just a thought, but this can be simplified by removing `ETH_CALL_FORCE_TO_CONSENSUS_BY_SELECTOR` altogether. When the functionallity needs to be disabled just leave `ETH_CALL_CONSENSUS_SELECTORS` as an empty array. This way there is 1 less env variable to maintain

quiet-node: yes fair point!

quiet-node: updated with the removal of `ETH_CALL_FORCE_TO_CONSENSUS_BY_SELECTOR`

lukelee-sl (dismissed):
LGTM

packages/relay/src/lib/eth.ts
```diff
@@ -1643,9 +1643,11 @@ export class EthImpl implements Eth {
 
     const selector = getFunctionSelector(call.data!);
 
-    const shouldForceToConsensus =
-      process.env.ETH_CALL_FORCE_TO_CONSENSUS_BY_SELECTOR == 'true' &&
-      constants.ETH_CALL_SELECTORS_ALWAYS_TO_CONSENSUS.indexOf(selector) !== -1;
+    // When eth_call is invoked with a selector listed in specialSelectors, it will be routed through the consensus node, regardless of ETH_CALL_DEFAULT_TO_CONSENSUS_NODE.
+    // note: this feature is a workaround for when a feature is supported by consensus node but not yet by mirror node.
+    // Follow this ticket https://github.com/hashgraph/hedera-json-rpc-relay/issues/2984 to revisit and remove special selectors.
+    const specialSelectors = process.env.ETH_CALL_CONSENSUS_SELECTORS || [''];
```
acuarica, line 1649: This value will have different types, it can be either `string` or `string[]` (when `ETH_CALL_CONSENSUS_SELECTORS` is not defined. Is this intended?

packages/relay/src/lib/eth.ts
```diff
@@ -1643,9 +1643,11 @@ export class EthImpl implements Eth {
...
+    const shouldForceToConsensus = selector !== '' && specialSelectors.includes(selector);
```
acuarica, line 1650: I guess this makes sense when `specialSelector` is of type `string[]`, but what happens when its type is `string`?

acuarica (commented):
lg, but left a comment

ebadiere (dismissed):
LG

packages/relay/src/lib/eth.ts
quiet-node: Good point! Updated a fix for it.

acuarica (dismissed):
lg

lukelee-sl (approved):

ebadiere (approved):

PR #2990, opened 2024-09-16
fix: Skip issue matching check for thirdparty build dependency librar…

### LINKED ISSUE #2988: The PR check in CI needs to be updated to ignore the issues check on external dependencies.
### Description

Right now we have a failing dependency PR check test in CI.  It is not finding the associated issue.  The PR check is to an external library that we do not control.

### Steps to reproduce

[Here is the failing build.](https://github.com/hashgraph/hedera-json-rpc-relay/actions/runs/10856422986/job/30131037345?pr=2958)

### Additional context

_No response_

### Hedera network

other

### Version

v0.55.-

### Operating system

Linux

ebadiere:
Skip issue check for thirdparty libraries. We do not control them.

**Related issue(s)**:

Fixes #2988

### DIFF
--- a/.github/scripts/check-pr.js
+++ b/.github/scripts/check-pr.js
@@ -37,13 +37,19 @@ async function getIssueDetails(issueNumber) {
 async function run() {
     try {
         const pr = await getPRDetails();
-        const { labels: prLabels, milestone: prMilestone, body: prBody } = pr;
+        const { labels: prLabels, milestone: prMilestone, body: prBody, title: prTitle } = pr;
 
         if (prLabels.length === 0) {
             throw new Error('The PR has no labels.');
         }
         if (!prMilestone) {
             throw new Error('The PR has no milestone.');
+        }
+
+        // Thirdparty dependencies are not controlled by us so we will skip the issue matching check for them"
+        if (prTitle.includes("build(deps):")) {
+            console.log('PR is a thirdparty dependency update, skipping issue matching check.');
+            return;
         }
 
         const issueNumberMatches = prBody.match(/#(\d+)/g);


quiet-node (changes requested):
I don’t think we should allow any third-party PRs to skip the `check_pr` CI task. Instead, we should let the CI fail for these PRs to serve as a reminder to update their labels and milestones.

.github/scripts/check-pr.js
```diff
@@ -46,6 +46,12 @@ async function run() {
             throw new Error('The PR has no milestone.');
         }
 
+        // Thirdparty dependencies are not controlled by us so we will skip the issue matching check for them"
+        if (prTitle.includes("build(deps):")) {
```
quiet-node, line 50: instead of checking if prTitle includes `build(deps)s`, we might want to use a different criterion, such as checking the PR author or something like that. The build(deps) title doesn't necessarily indicate that the PR is from a third party, so relying solely on that check might not be sufficient IMO

ebadiere: Updated.

.github/scripts/check-pr.js
```diff
@@ -5,73 +5,274 @@ const { GITHUB_REPOSITORY, GITHUB_PR_NUMBER } = process.env;
...
-        console.log('PR and all associated issues have labels and milestones.');
+        console.log('All checks completed.');
     } catch (error) {
         console.error(error.message);
         process.exit(1);
     }
 }
 
-run();
+run();
```
quiet-node, line 278: nit: missing empty line at EOF

quiet-node (approved):
LGTM 1 nit can be addresed in future

.github/scripts/check-pr.js
```diff
@@ -5,73 +5,274 @@ const { GITHUB_REPOSITORY, GITHUB_PR_NUMBER } = process.env;
...
-    const url = `https://api.github.com/repos/${owner}/${repo}/pulls/${GITHUB_PR_NUMBER}`;
-    const response = await axios.get(url, {
-        headers: {
-            Authorization: `token ${githubToken}`
+async function getPRDetails(prNumber) {
+    const url = `https://api.github.com/repos/${owner}/${repo}/pulls/${prNumber}`;
+    try {
+        const response = await axios.get(url, {
+            headers: {
+                Authorization: `token ${githubToken}`,
```
nathanklick, line 13: Which workflow executes this script and what are the effective permissions of the `const githubToken = process.env.GITHUB_TOKEN;` token?

ebadiere: This is run as part of the `pr-label-milestone-check.yml` workflow.  The GITHUB_TOKEN is pulled from the secrets and assigned in that workflow:
```
     - name: Check PR labels and milestones
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          GITHUB_PR_NUMBER: ${{ github.event.number }}
        run: node .github/scripts/check-pr.js

nathanklick: @ebadiere Ok so the effective permissions are `read` for all by virtue of the `pr-label-milestone-check.yml` workflow definition.

nathanklick (approved):

PR #3014, opened 2024-09-19
refactor: configurations of `HbarLimitService`

### LINKED ISSUE #2970: [HBAR Rate Limit Redesign] Note what configurations are used
Update the design doc to note which configs are used and what new configs were added and their purpose and effect on the design.

Looking at `HbarSpendingPlan`, I'm curious about the `spentToday` field.
1. What was the logic to confine it **_Today_** and how is it used?
2. Are we still using the `HBAR_RATE_LIMIT_DURATION` config and how does that affect `spentToday` if the duration is not 1 day?
3. What are the configurations which will be used to create the spending plans for supported projects & partners

### LINKED ISSUE #3018: [HBAR Rate Limit Redesign] Make limit duration and budget configurable

victor-yanev:
**Description**:

This PR includes changes to the configurations of the `HbarLimitService` to make use of the env variables for the total budget and the limit duration that were used for the old `HbarLimit` implementation.

As a result, also had to rename `spentToday` field to `amountSpent` since we are no longer doing daily limiting and we are limiting based on a configurable time window.

It also adds a `Configurations` section in the `hbar-limiter.md doc` which describes all configurations used for the redesigned HBAR limiter, and also added the new env vars used in the `configurations.md` doc.

**Related issue(s)**:

Fixes #2970

Fixes #3018

**Notes for reviewer**:

As part of doing #2970 and #3018, I noticed a flaw in the configurations of the `HbarLimitService` - we use `number` for the type of the `totalBudget` and `remainingBudget` and if this value becomes too big, we would hit a bug of overflowing numbers.

Current limiting of the existing `HbarLimit` class is 110 ℏ per minute - if we want to keep this limit the same (but limit on a per day basis instead of per minute) that would be:

```math
totalBudget = 110 \: ℏ * 1440 \: mins \: in \: a \: day = 158400 \: ℏ = 1.58400e13 \: tℏ \: per \: day
```

These numbers get increasingly big, especially if we decide to change the limit duration to be per month instead of per day - then we would overflow the `Number.MAX_VALUE` and break the hbar limiter.

This is why in this PR, I have changed the type of `totalBudget` and `remainingBudget` from `number` to `Hbar`, which internally uses the `Long` library.

**Checklist**


docs/design/hbar-limiter.md
```diff
@@ -271,6 +275,83 @@ c. Current day's usage (increase limits if overall usage is low)
...
+HBAR_RATE_LIMIT_BASIC=1000
+HBAR_RATE_LIMIT_EXTENDED=10000
+HBAR_RATE_LIMIT_PRIVILEGED=20000
+```
+
+### Total Budget and Limit Duration
+
+The total budget and the limit duration are defined as environment variables:
+- `HBAR_RATE_LIMIT_DURATION`: The time window (in milliseconds) for which both the total bud
+- `HBAR_RATE_LIMIT_TINYBAR`: The ceiling on the total amount of HBars that can be spent in the limit duration.
```
Nana-EC, line 347: Given the context above we should be explicit and note this is the largest bucket from which others pull from.

docs/design/hbar-limiter.md
```diff
@@ -271,6 +275,83 @@ c. Current day's usage (increase limits if overall usage is low)
...
+```dotenv
+HBAR_RATE_LIMIT_BASIC=1000
+HBAR_RATE_LIMIT_EXTENDED=10000
+HBAR_RATE_LIMIT_PRIVILEGED=20000
+```
+
+### Total Budget and Limit Duration
+
+The total budget and the limit duration are defined as environment variables:
+- `HBAR_RATE_LIMIT_DURATION`: The time window (in milliseconds) for which both the total bud
```
Nana-EC, line 346: Might be worth mentioning how the duration applies. Also i think the sentence here wasn't completed

docs/design/hbar-limiter.md
```diff
@@ -271,6 +275,83 @@ c. Current day's usage (increase limits if overall usage is low)
...
+  }
+]
+```
+
+### Spending Limits of Different Tiers
+
+The spending limits for different tiers are defined as environment variables:
+- `HBAR_RATE_LIMIT_BASIC`: The spending limit (in hbars) for general users (tier 3)
+- `HBAR_RATE_LIMIT_EXTENDED`: The spending limit (in hbars) for supported projects (tier 2)
+- `HBAR_RATE_LIMIT_PRIVILEGED`: The spending limit (in hbars) for trusted partners (tier 1)
```
Nana-EC, line 334: Does this config have an unlimited mode?

docs/design/hbar-limiter.md
```diff
@@ -271,6 +275,83 @@ c. Current day's usage (increase limits if overall usage is low)
...
+  {
+    "name": "new partner name",
+    "ethAddresses": ["0x129", "0x130"],
+    "ipAddresses": ["133.0.0.1"],
+    "subscriptionTier": "ADVANCED"
+  }
+]
+```
+
+### Spending Limits of Different Tiers
```
Nana-EC, line 329: What happens when the tiered values are  larger than `HBAR_RATE_LIMIT_TINYBAR` is there an expected behaviour?

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -27,16 +27,19 @@ import { IDetailedHbarSpendingPlan } from '../../db/types/hbarLimiter/hbarSpendi
...
 
 export class HbarLimitService implements IHbarLimitService {
-  static readonly ONE_DAY_IN_MILLIS = 24 * 60 * 60 * 1000;
-  // TODO: Replace with actual values
-  static readonly DAILY_LIMITS: Record<SubscriptionType, number> = {
-    BASIC: parseInt(process.env.HBAR_DAILY_LIMIT_BASIC ?? '1000'),
-    EXTENDED: parseInt(process.env.HBAR_DAILY_LIMIT_EXTENDED ?? '10000'),
-    PRIVILEGED: parseInt(process.env.HBAR_DAILY_LIMIT_PRIVILEGED ?? '100000'),
+  // TODO: Replace with actual values - https://github.com/hashgraph/hedera-json-rpc-relay/issues/2895
+  static readonly DAILY_LIMITS: Record<SubscriptionType, Hbar> = {
```
Nana-EC, line 35: Q: are they references to daily limits accurate or are they variable based on the duration?

Nana-EC (commented):
Nice work, some suggestions and a few places to clarify to relay operators how the different configs impact logic and how they impact eachother

docs/design/hbar-limiter.md
victor-yanev: Hmm good point, ideally this scenario should not be permitted, we can add validations on the env variables as part of https://github.com/hashgraph/hedera-json-rpc-relay/issues/3023

docs/design/hbar-limiter.md
victor-yanev: Not really at the moment, but it could be easily changed if we want this behavior

packages/relay/src/lib/services/hbarLimitService/index.ts
victor-yanev: Good point - this variable name is no longer correct, I can rename it to `TIERED_LIMITS`.

docs/design/hbar-limiter.md
Nana-EC: That's fine, let's leave it for a future feature request.

docs/design/hbar-limiter.md
Nana-EC: Please add a description of the needed validation

victor-yanev: Done, added in the description of https://github.com/hashgraph/hedera-json-rpc-relay/issues/3023

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -134,21 +140,21 @@ export class HbarLimitService implements IHbarLimitService {
       {} as Record<SubscriptionType, Gauge>,
     );
 
-    // Reset the rate limiter at the start of the next day
-    this.reset = this.getResetTimestamp();
+    setInterval(() => {
```
Nana-EC, line 143: Q: what's the impact here? You're resetting metric every day regardless of other logic?

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -189,13 +195,14 @@ export class HbarLimitService implements IHbarLimitService {
       spendingPlan = await this.createBasicSpendingPlan(ethAddress, ipAddress);
     }
 
-    const dailyLimit = HbarLimitService.DAILY_LIMITS[spendingPlan.subscriptionType];
+    const dailyLimit = HbarLimitService.TIER_LIMITS[spendingPlan.subscriptionType].toTinybars();
```
Nana-EC, line 198: Still says daily.
Is it?

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -317,13 +324,21 @@ export class HbarLimitService implements IHbarLimitService {
...
+   *
    * @returns {Date} - The new reset timestamp.
    * @private
    */
   private getResetTimestamp(): Date {
-    const tomorrow = new Date(Date.now() + HbarLimitService.ONE_DAY_IN_MILLIS);
-    return new Date(tomorrow.setHours(0, 0, 0, 0));
+    const now = new Date();
+    const resetDate = new Date(now.getTime() + this.limitDuration);
+    if (this.limitDuration >= this.oneDayInMillis) {
```
Nana-EC, line 337: Q: Is this correct?
Does this mean our intention is that the hbar limit at max supports logic up to on day?
Just need to confirm and be explicit in the docs if so how one day and the other durations interact with each other

Nana-EC (commented):
Looking good, a few more specification.
Feel free to call out appropriate items that can be pulled out to separate follow up PRs so this doesn't get more complex

docs/design/hbar-limiter.md
```diff
@@ -19,6 +19,10 @@
       - [HBar Allocation Strategy](#hbar-allocation-strategy)
         - [Metrics to Track](#metrics-to-track)
         - [Allocation Algorithm](#allocation-algorithm)
+  - [Configurations](#configurations)
+    - [Pre-populating the Redis Cache with Spending Plans for Supported Projects and Partner Projects](#pre-populating-the-redis-cache-with-spending-plans-for-supported-projects-and-partner-projects)
```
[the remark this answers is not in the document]
victor-yanev: Yes, internally the repository classes use the `CacheService`, so `LRUCache` is used by default if Redis is not configured.

I will modify this to say just `Cache` instead of `Redis Cache`.

docs/design/hbar-limiter.md
```diff
@@ -271,6 +275,88 @@ c. Current day's usage (increase limits if overall usage is low)
    - Keep a small portion of the daily budget (e.g., 10%) as a reserve
    - Use this to accommodate unexpected spikes or high-priority users
 
+## Configurations
+
+### Pre-populating the Redis Cache with Spending Plans for Supported Projects and Partner Projects
+
+The following configurations will be used to automatically populate the Redis cache with `HbarSpendingPlan`, `EthAddressHbarSpendingPlan`, and `IPAddressHbarSpendingPlan` entries for the outlined supported projects and partner projects on every start-up of the relay.
+
+All other users (ETH and IP addresses which are not specified in the configuration file) will be treated as "general users" and will be assigned a basic `HbarSpendingPlan` on their first request and their ETH address and IP address will be linked to that plan for all subsequent requests.
```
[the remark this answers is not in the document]
victor-yanev: There is no order, both the IP and ETH address will be linked to the spending plan, so if one of them matches in subsequent requests we will re-use that existing plan

docs/design/hbar-limiter.md
```diff
@@ -271,6 +275,88 @@ c. Current day's usage (increase limits if overall usage is low)
...
+    "ipAddresses": ["133.0.0.1"],
+    "subscriptionTier": "ADVANCED"
+  }
+]
+```
+
+### Spending Limits of Different Tiers
+
+The spending limits for different tiers are defined as environment variables:
+- `HBAR_RATE_LIMIT_BASIC`: The spending limit (in hbars) for general users (tier 3)
```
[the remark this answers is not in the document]
victor-yanev: We can specify decimals - the `BigNumber` constructor supports decimal strings:

**constants.ts**:
```typescript
{
  ...,
  HBAR_RATE_LIMIT_BASIC: Hbar.from(BigNumber(process.env.HBAR_DAILY_LIMIT_BASIC || '1000'), HbarUnit.Hbar),
  HBAR_RATE_LIMIT_EXTENDED: Hbar.from(BigNumber(process.env.HBAR_DAILY_LIMIT_EXTENDED || '10000'), HbarUnit.Hbar),
  HBAR_RATE_LIMIT_PRIVILEGED: Hbar.from(BigNumber(process.env.HBAR_DAILY_LIMIT_PRIVILEGED || '20000'), HbarUnit.Hbar),
}
```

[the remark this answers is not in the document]
victor-yanev: We can also change this to accept tinybars, which one do you think is better? 🤔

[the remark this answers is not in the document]
victor-yanev: Or maybe add a new env variable for `HBAR_RATE_LIMIT_UNIT` which we would link to one of the following:
```typescript
declare namespace HbarUnit {
    let Tinybar: HbarUnit;
    let Microbar: HbarUnit;
    let Millibar: HbarUnit;
    let Hbar: HbarUnit;
    let Kilobar: HbarUnit;
    let Megabar: HbarUnit;
    let Gigabar: HbarUnit;
}
```

packages/relay/src/lib/services/hbarLimitService/index.ts
victor-yanev: Yeah, I thought we want daily metrics, or should we reset the metrics together with the hbar limiter based on its duration? 🤔

packages/relay/src/lib/services/hbarLimitService/index.ts
victor-yanev: Oops, good catch. I will update the name of this variable.

packages/relay/src/lib/services/hbarLimitService/index.ts
victor-yanev: It supports more than one day as well, the idea is described in the jsdocs: `If the new reset timestamp falls on a different day, the time is set to midnight.`, so what that means is:
- we start-up the relay at let's say 3:00PM
- if we set the duration to 2 days, the reset date will be after 2 days at midnight
- if we set the duration to 1 day, the reset date will be after 1 day at midnight
- if we set the duration to 4 hours, the reset date will be after 4 hours

docs/design/hbar-limiter.md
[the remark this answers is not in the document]
victor-yanev: Changed the tiered limits to be specified in tinybars

packages/relay/src/lib/services/hbarLimitService/index.ts
victor-yanev: Changed the way it is calculated, and also added docs:
```typescript
  /**
   * Calculates the next reset timestamp for the rate limiter.
   *
   * This method determines the next reset timestamp based on the current reset timestamp
   * and the limit duration. If the current reset timestamp is not defined, it initializes
   * the reset timestamp to midnight of the current day. It then iteratively adds the limit
   * duration to the reset timestamp until it is in the future.
   *
   * @returns {Date} - The next reset timestamp.
   */
  private getResetTimestamp(): Date {
    const todayAtMidnight = new Date().setHours(0, 0, 0, 0);

    let newResetDate = this.reset !== undefined ? new Date(this.reset.getTime()) : new Date(todayAtMidnight);
    while (newResetDate.getTime() < Date.now()) {
      newResetDate = new Date(newResetDate.getTime() + this.limitDuration);
    }

    return newResetDate;
  }
```

packages/relay/tests/lib/services/metricService/metricService.spec.ts
```diff
@@ -132,7 +132,7 @@ describe('Metric Service', function () {
     mock = new MockAdapter(instance);
 
     const duration = constants.HBAR_RATE_LIMIT_DURATION;
-    const total = constants.HBAR_RATE_LIMIT_TINYBAR;
+    const total = constants.HBAR_RATE_LIMIT_TOTAL.toTinybars().toNumber();
```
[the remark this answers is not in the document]
victor-yanev: Changed, thanks

packages/relay/src/lib/services/hbarLimitService/index.ts
ebadiere: Why not both?

konstantinabl (dismissed):

packages/relay/src/lib/services/hbarLimitService/index.ts
Nana-EC: I don't think we should be resetting, allow grafana to poll as needed.
We can calculated daily metrics later but resetting it doesn't seem right or we'll lose the granularity

Nana-EC (commented):
LG but the metrics reset portion is unclear

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -322,13 +328,28 @@ export class HbarLimitService implements IHbarLimitService {
...
+    let resetDate = this.reset ? new Date(this.reset.getTime()) : new Date(todayAtMidnight);
+    if (resetDate.getTime() < Date.now()) {
+      // 1. Calculate the difference between the current time and the reset time.
+      // 2. Determine how many intervals of size `limitDuration` have passed since the last reset.
+      // 3. Calculate the new reset date by adding the required intervals to the original reset date.
+      const intervalsPassed = Math.ceil((Date.now() - resetDate.getTime()) / this.limitDuration);
+      resetDate = new Date(resetDate.getTime() + intervalsPassed * this.limitDuration);
+    }
+
+    return resetDate;
```
quiet-node, line 352: Let's say on the first startup of the Relay, the `resetDate` is set to midnight of the current date. For simplicity, let's assign it a UNIX timestamp of 18000, and assume `limitDuration = 80 seconds`.

After midnight, a new request comes in at `Date.now = 18080` (80 seconds after midnight). The calculation would be:

`intervalsPassed = (18080 - 18000) / 80 = 1 interval passed`.

Then, the new `resetDate` would be:

`resetDate = 18000 + 1 * 80 = 18080` (which is the current `Date.now`). 

In this case, why don't we simply set `resetDate = Date.now` instead of calculating the number of intervals passed?

In another scenario, assuming midnight is still at 18000, if a request comes in at `18100` (100 seconds after midnight), the calculation would be:

`intervalsPassed = Math.ceil(100 / 80) = 2 intervals (or 160 seconds)`.

Thus, the new `resetDate` would be:

`resetDate = 18000 + 2 * 80 = 18160`.

This creates a gap of 60 seconds between when the request was made and the reset time. 

I guess I am confused with this logic of calculating the resetDate.

quiet-node: Aside from all that above, the `getResetTimestamp()` method is used to retrieve the `resetDate` and reset `this.reset` in the `resetLimiter()` method. This part is also somewhat confusing.

Using the same example, let's assume at startup, `getResetTimestamp()` returns a `resetDate` set to midnight (we'll use 18000 to remain consistent with the previous example). This makes sense because `this.reset = 18000`, and it’s in the future compare to startup time.

Now, let’s say a request comes in after midnight at `18080` (80 seconds after midnight). At this point, `shouldResetLimiter()` is triggered and compares `Date.now()`, which is `18080`, to `this.reset`, which is `18000`. Since `Date.now()` is greater than `this.reset`, `shouldResetLimiter()` returns `true`, which makes sense so far. 

When `shouldResetLimiter()` returns `true`, `resetLimiter()` is called, and `this.reset` is set to `this.getResetTimestamp()`. So far still makes perfect sense.

However, with the current logic in `getResetTimestamp()`, when the request arrives at `18080`, `this.reset` is still at `18000`. Therefore, it falls into the following block:
```typescript
      const intervalsPassed = Math.ceil((Date.now() - resetDate.getTime()) / this.limitDuration);
      resetDate = new Date(resetDate.getTime() + intervalsPassed * this.limitDuration);
```
This calculates the `resetDate` as `18080`, which is the same time as when the request was made. This means the `resetDate` will always be in the past from this point onward, because any future requests will come after the `18080` timestamp. As a result, `shouldResetLimiter()` will always return `true`, since `Date.now()` will always be greater than the reset timestamp, which remains at `18080`.

This doesn’t seem logical, as the `resetDate` should always be in the future and at least 24 hours (1 day) apart from the previous reset timestamp.

docs/configuration.md
```diff
@@ -68,10 +66,15 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
 | `GAS_PRICE_PERCENTAGE_BUFFER`               | "0"                      | The additional buffer that adds a percentage on top of the calculated network gasPrice. This may be used by operators to reduce the chances of `INSUFFICIENT_TX_FEE` errors experienced by users caused by minor fluctuations in the exchange rate.                                                    |
 | `GAS_PRICE_TINY_BAR_BUFFER`                 | "10000000000"            | The additional buffer range to allow during a relay precheck of gas price. This supports slight fluctuations in network gasprice calculations.                                                                                                                                                         |
+| `HBAR_RATE_LIMIT_DURATION`                  | "80000"                  | HBar budget limit duration. This creates a timestamp, which resets all limits, when it's reached. Default is to 80000 (80 seconds).                                                                                                                                                                    |
```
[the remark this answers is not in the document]
ebadiere: I recall the documentation listing 80s but we run with 60s.

packages/relay/src/lib/db/repositories/hbarLimiter/ethAddressHbarSpendingPlanRepository.ts
```diff
@@ -68,11 +68,16 @@ export class EthAddressHbarSpendingPlanRepository {
...
    * @param {IEthAddressHbarSpendingPlan} addressPlan - The plan to save.
    * @param {RequestDetails} requestDetails - The request details for logging and tracking.
+   * @param {number} [ttl] - The time-to-live for the cache entry. (default: 1 day)
    * @returns {Promise<void>} - A promise that resolves when the ETH address is linked to the plan.
    */
-  async save(addressPlan: IEthAddressHbarSpendingPlan, requestDetails: RequestDetails): Promise<void> {
+  async save(
+    addressPlan: IEthAddressHbarSpendingPlan,
+    requestDetails: RequestDetails,
+    ttl: number = this.oneDayInMillis,
```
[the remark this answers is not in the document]
victor-yanev: Hmm, might be better to set this to be `this.limitDuration` by default, because if the `limitDuration` is more than one day, we would lose the BASIC spending plans before the hbar limiter has been restarted. Good catch!

packages/relay/src/lib/services/hbarLimitService/index.ts
victor-yanev: > This calculates the `resetDate` as `18080`, which is the same time as when the request was made. This means the `resetDate` will always be in the past from this point onward, because any future requests will come after the `18080` timestamp. As a result, `shouldResetLimiter()` will always return `true`, since `Date.now()` will always be greater than the reset timestamp, which remains at `18080`.

1. this is not really possible to happen, it's an extreme edge case
2. this is not true, because what would happen is that on the next request `shouldResetLimiter()` will return true and `getResetTimestamp()` will be called again by `resetLimiter()` to calculate the next reset timestamp, which will be `18160`

victor-yanev: > Thus, the new `resetDate` would be:
> 
> `resetDate = 18000 + 2 * 80 = 18160.`
> 
> This creates a gap of 60 seconds between when the request was made and the reset time.

There is no "gap" because if `getResetTimestamp()` was called, this means that `resetLimiter()` was called and the hbar limiter was resetted.

victor-yanev: Overall think of `getResetTimestamp()` as a method which is called to give you the **_next reset timestamp_** and we call it only when we have resetted the limiter.

packages/relay/src/lib/db/repositories/hbarLimiter/ethAddressHbarSpendingPlanRepository.ts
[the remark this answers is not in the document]
victor-yanev: Changed!

packages/relay/src/lib/services/hbarLimitService/index.ts
victor-yanev: Discussed offline

docs/configuration.md
```diff
@@ -69,6 +67,11 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
+| `HBAR_RATE_LIMIT_TINYBAR`                   | "11_000_000_000"         | Total hbar budget in tinybars (110 hbars per 80 seconds).                                                                                                                                                                                                                                              |
```
Nana-EC, line 71: Nit, no need to repeat 80 seconds. If anything note `HBAR_RATE_LIMIT_DURATION` so that the text isn't prone to being outdated

Nana-EC (approved):

packages/relay/src/lib/db/repositories/hbarLimiter/ethAddressHbarSpendingPlanRepository.ts
```diff
@@ -27,7 +27,7 @@ import { RequestDetails } from '../../../types';
 
 export class EthAddressHbarSpendingPlanRepository {
   private readonly collectionKey = 'ethAddressHbarSpendingPlan';
-  private readonly threeMonthsInMillis = 90 * 24 * 60 * 60 * 1000;
+  private readonly oneDayInMillis = 24 * 60 * 60 * 1000;
```
quiet-node, line 30: Looks like this is not used anywhere anymore, right? If so it can be removed.

quiet-node (approved):
LG. 1 nit can be addressed in upcoming PR.

packages/relay/src/lib/services/hbarLimitService/index.ts
quiet-node: Was a confusion I had behind the idea of the limitDuration carrying from the old implementation. Fully resolved offline.

PR #3021, opened 2024-09-20
ci: Adding key env variables step in workflow to resolve issue with env.

### LINKED ISSUE #3020: Fix issue with acceptance-workflow because of replacement of conditional setting
There is an issue with the acceptance workflow due to [PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3012) which is breaking the workflow.
Input from: [Slack message](https://limechain.slack.com/archives/C03DN7L1T1A/p1726824240424909)

san-est:
**Description**:
The acceptance workflow is breaking because of the removal of conditional setting of non-required inputs. I've added the step in which variables are added if the CI is manual.

**Related issue(s)**:

Fixes #3020 

**Notes for reviewer**:

**Checklist**


nathanklick (changes requested):
@rbarkerSL Please review

nathanklick (approved):

nathanklick (approved):

rbarkerSL (dismissed):

victor-yanev (dismissed):
Please revert the changes to the source code, we want to be able to set an env variable and override any values from the `.env` file, not vice-versa.

Imo, since the `OPERATOR_ID` input is optional we should **optionally** set the env variable instead of modifying the source code to avoid the bug in the workflow configuration.

packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -99,12 +106,17 @@ export default class HAPIService {
     eventEmitter: EventEmitter,
   ) {
     dotenv.config({ path: findConfig('.env') || '' });
+    if (fs.existsSync(findConfig('.env') || '')) {
+      this.config = dotenv.parse(fs.readFileSync(findConfig('.env') || ''));
+    } else {
+      this.config = {};
+    }
```
victor-yanev, line 113: We don't need this additional field, please remove it.

The line above already loads those env variables into Node's `process.env`:
```typescript
dotenv.config({ path: findConfig('.env') || '' });
```

And those fallbacks that we are doing below in the added changes are essentially useless.

packages/server/tests/acceptance/index.spec.ts
```diff
@@ -50,7 +50,8 @@ import { AliasAccount } from '../types/AliasAccount';
 import { setServerTimeout } from '../../src/koaJsonRpc/lib/utils';
 
 chai.use(chaiAsPromised);
-dotenv.config({ path: path.resolve(__dirname, '../../../.env') });
+dotenv.config({ path: path.resolve(__dirname, '../../../../.env') });
+const DOT_ENV = dotenv.parse(fs.readFileSync(path.resolve(__dirname, '../../../../.env')));
```
victor-yanev, line 54: Same here, the change in the line above which modifies the path of the `.env` file is sufficient to fix the problem, fallbacks to this `DOT_ENV` variable is useless, since the same variables are already loaded into `process.env`

packages/ws-server/tests/acceptance/index.spec.ts
```diff
@@ -35,7 +35,8 @@ import { app as wsApp } from '@hashgraph/json-rpc-ws-server/dist/webSocketServer
 import ServicesClient from '@hashgraph/json-rpc-server/tests/clients/servicesClient';
 import { Utils } from '@hashgraph/json-rpc-server/tests/helpers/utils';
 import { AliasAccount } from '@hashgraph/json-rpc-server/tests/types/AliasAccount';
-dotenv.config({ path: path.resolve(__dirname, '../../../.env') });
+dotenv.config({ path: path.resolve(__dirname, '../../../../.env') });
+const DOT_ENV = dotenv.parse(fs.readFileSync(path.resolve(__dirname, '../../../../.env')));
```
victor-yanev, line 39: Same here, please remove this `DOT_ENV` variable

packages/relay/src/lib/services/hapiService/hapiService.ts
nathanklick: @victor-yanev This is a workaround which properly handles the case where the environment defines an empty variable. The behavior of the `dotenv` module is to override the `.env` file even when the environment variable is an empty value or zero length string.

This change is needed to unblock the tests in CI and will be merged.

nathanklick: @victor-yanev Additionally, the `OPERATOR_ID_MAIN` and `OPERATOR_KEY_MAIN` are truly not optional and must be supplied either via the `.env` or the actual environment variables.

The relay ceases to function if either of these variables are omitted. The current code does not properly handle or validate the absence of these required variables. This is a break fix for a problem in the code.

victor-yanev: Yes, that’s right but this `config` variable is still unneeded, because we are already loading the `.env` file into `process.env` through line 108.

ebadiere: @victor-yanev I opened an [issue to handle environment variables properly in the relay.](https://github.com/hashgraph/hedera-json-rpc-relay/issues/3023)

victor-yanev: Sorry, I meant optional inputs in the workflow file, not optional overall in the relay.

ebadiere (approved):

PR #3022, opened 2024-09-20
test: add helper method for overriding env variables

### LINKED ISSUE #3029: Add Helper Methods for Overriding Environment Variables in Tests
### Problem

In the current setup, there is no straightforward way to temporarily override environment variables for the duration of a test suite. This can lead to issues where tests inadvertently affect each other due to shared environment settings, making it difficult to isolate and reproduce test failures.

Additionally, in many tests environment variables are overridden without being reverted, propagating this behavior to other tests and making the tests dependent on their order and leading to chain-failures in cases where only one of the tests actually fails.

### Solution

Add new helper methods `overrideEnvs` and `withOverriddenEnvs` provide a mechanism to temporarily set environment variables for the duration of a test or a group of tests. These methods ensure that the original environment settings are restored after the tests have run, maintaining the integrity of the global environment.
  - `overrideEnvs`: Temporarily sets environment variables before a test and restores them afterward.
  - `withOverriddenEnvs`: A higher-level function that uses `overrideEnvs` to set environment variables for a group of tests defined within a callback function.

**Examples:**

```typescript
// Example usage of overrideEnvs
describe('test suite', () => {
  overrideEnvs({ TEST: 'true', API_URL: 'http://localhost' });
  
  it('should run with overridden env variables', () => {
  });

  it('should run with overridden env variables', () => {
  });
});

it('should run with original env variables', () => {
});
```

```typescript
// Example usage of withOverriddenEnvs
withOverriddenEnvs({ TEST: 'false', API_URL: 'http://testnet' }, () => {
  it('should run with overridden env variables', () => {
  });
  
  it('should run with overridden env variables', () => {
  });
});

it('should run with original env variables', () => {
});
```

### Alternatives

_No response_

victor-yanev:
**Description**:

In the current setup, there is no straightforward way to temporarily override environment variables for the duration of a test suite. This can lead to issues where tests inadvertently affect each other due to shared environment settings, making it difficult to isolate and reproduce test failures.

Additionally, in many tests environment variables are overridden without being reverted, propagating this behavior to other tests and making the tests dependent on their order and leading to chain-failures in cases where only one of the tests actually fails.

**Changes:**

This PR adds new helper methods `overrideEnvs` and `withOverriddenEnvs` which provide a mechanism to temporarily set environment variables for the duration of a test or a group of tests. 

ℹ️ These methods ensure that the original environment settings are restored after the tests have run, maintaining the integrity of the global environment.

- `overrideEnvs`: Temporarily sets environment variables before a test and restores them afterward.
- `withOverriddenEnvs`: A higher-level function that uses `overrideEnvs` to set environment variables for a group of tests defined within a callback function.

**Related issue(s)**:

Fixes #3029

**Checklist**


packages/relay/tests/helpers.ts
```diff
@@ -921,6 +921,52 @@ export const stopRedisInMemoryServer = async (
...
+    before(() => {
+      for (const key in envs) {
+        envsToReset[key] = process.env[key];
+        overrideEnv(process.env, key, envs[key]);
+      }
+    });
+
+    tests();
+
+    after(() => {
```
acuarica, line 962: I believe mocha allows the user to have multiple `before` and `after`. Maybe we could use a smaller wrapper that only invokes `before` and `after` to set only 1 env variable. I guess this way might be more composable. What do you think?

victor-yanev: That's true and the `tests()` method that is passed can also include `before`, `beforeEach`, `afterEach`, and `after` in addition to `it` and `describe` declarations, I don't see a problem with there being other `before` and `after` called inside.

I am not sure what you mean, could you maybe give some examples?

packages/relay/tests/helpers.ts
```diff
@@ -921,6 +921,52 @@ export const stopRedisInMemoryServer = async (
...
+      for (const key in envsToReset) {
+        overrideEnv(process.env, key, envsToReset[key]);
+      }
+    });
+  });
+};
```
acuarica, line 968: I was thinking something like this (and for each env var you want to modify you need to call this method in the test)

```suggestion
export const overriddeEnv = (key: string, value: string | undefined) => {
  let prevValue: string | undefined;

  before(() => {
    prevValue = process.env[key];
    process.env[key] = value;
  });

  after(() => {
    if (prevValue === undefined) {
      delete process.env[key];
    } else {
      process.env[key] = prevValue;
    }
  });
};
```

packages/relay/tests/helpers.ts
```diff
@@ -885,40 +885,70 @@ export const calculateTxRecordChargeAmount = (exchangeRateIncents: number) => {
...
+
+/**
+ * Overrides environment variables for the duration of the provided tests.
+ *
+ * @param {NodeJS.Dict<string>} envs - An object containing key-value pairs of environment variables to set.
+ * @param {Function} tests - A function containing the tests to run with the overridden environment variables.
+ */
+export const withOverriddenEnvs = (envs: NodeJS.Dict<string>, tests: () => void) => {
```
Nana-EC, line 942: Interesting usage.
I see lot sof changes were needed to adopt new approach.
It might be time for a `testing-guide.md` in which there's a section that tells contributors how to approach testing. In one such section you can note these 2 methods and how they should be utilized by engineers.

No need to do the whole doc but at least the how people should use these 2 methods so that contributors are well informed.

Nana-EC (commented):
Nice work, left some feedback

packages/relay/tests/helpers.ts
victor-yanev: Hmm, let's open a separate issue for that, I also have a few other ideas of what we can include in this doc

packages/relay/tests/helpers.ts
```diff
@@ -885,40 +885,70 @@ export const calculateTxRecordChargeAmount = (exchangeRateIncents: number) => {
...
-  envsToReset: { TEST?: string; REDIS_ENABLED?: string; REDIS_URL?: string },
-): Promise<void> => {
+export const stopRedisInMemoryServer = async (redisInMemoryServer: RedisInMemoryServer): Promise<void> => {
   await redisInMemoryServer.stop();
-  process.env.TEST = envsToReset.TEST;
-  process.env.REDIS_ENABLED = envsToReset.REDIS_ENABLED;
-  process.env.REDIS_URL = envsToReset.REDIS_URL;
+};
+
+export const overrideEnvs = (envs: NodeJS.Dict<string>) => {
```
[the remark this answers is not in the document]
victor-yanev: True that it's not really obvious how it works from its name, the envs would be overridden:
- for the whole test suite if it's used in the root `describe` of the spec file
- but also only for a subset of tests if it's used in a nested `describe` block

So it depends on where it's used, I'll add docs and maybe try to come up with a better name for it (if you have any ideas feel free to share!)

packages/relay/tests/lib/clients/redisCache.spec.ts
```diff
@@ -104,7 +104,7 @@ describe('RedisCache Test Suite', async function () {
     it('should be able to set cache with TTL less than 1000 milliseconds', async () => {
       const key = 'int';
       const value = 1;
-      const ttl = 500;
+      const ttl = 200;
```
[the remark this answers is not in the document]
victor-yanev: Makes the tests faster, they were unnecessarily slow. 

I am even considering to drop this even lower - to `100`.

packages/relay/tests/assertions.ts
```diff
@@ -1,11 +1,34 @@
...
+import chai, { expect } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
 import { JsonRpcError } from '../src';
 import { EthImpl } from '../src/lib/eth';
 import { Block, Transaction } from '../src/lib/model';
 import { numberTo0x } from '../src/formatters';
 import constants from '../src/lib/constants';
 import { BASE_FEE_PER_GAS_DEFAULT } from './lib/eth/eth-config';
 
+chai.use(chaiAsPromised);
```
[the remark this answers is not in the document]
victor-yanev: Yes, there are utility functions which use `expect(...).eventually` inside, but `assertions.ts` was previously relying on other spec files which import it to have called `chai.use(chaiAsPromised)` themselves, instead of doing it here. 

I think they were extracted into this helper file at some point, but the person who did it forgot to call `chai.use(chaiAsPromised)` after extracting it into a separate file.

packages/relay/tests/helpers.ts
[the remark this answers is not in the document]
victor-yanev: Done, added some comprehensive docs also including code examples

packages/relay/tests/helpers.ts
victor-yanev: https://github.com/hashgraph/hedera-json-rpc-relay/issues/3031

packages/relay/tests/helpers.ts
```diff
@@ -885,40 +885,98 @@ export const calculateTxRecordChargeAmount = (exchangeRateIncents: number) => {
...
+ *   it('should return false', () => {
+ *     expect(process.env.TEST).to.equal('false');
+ *   });
+ * });
+ *
+ * it('should return true', () => {
+ *   expect(process.env.TEST).to.equal('true');
+ * });
+ */
+export const overrideEnvs = (envs: NodeJS.Dict<string>) => {
```
Nana-EC, line 928: Q: should we make the name more descriptive?
Something like `overrideEnvsInJestDescribe` and `withOverriddenEnvsInJestTest` or better names.
If these methods were coming from the jest library or were overriding classes it'd be fine to be implicit but used outside it can be missed.

My main concern here is engineering time spent writing and debugging tests missing how the env variable in the class are affecting the tests scope.
Seeing with clarity that the use of these methods has an impact in relation to jest scoping might avoid issues.

Nana-EC (commented):
LG, final comment is around method naming to make things obvious and avoid wasted troubleshooting time form this improvement

packages/relay/tests/helpers.ts
victor-yanev: Done, renamed to `overrideEnvsInMochaDescribe` and `withOverriddenEnvsInMochaTest`.

ebadiere (commented):
LG, but now there is a merge conflict.

konstantinabl (dismissed):

natanasow (approved):

ebadiere (approved):

konstantinabl (approved):

PR #3047, opened 2024-10-01
feat: capture env variables into sigleton and reusable constants

### LINKED ISSUE #3023: All environment variables required by the relay should be initially loaded and validated.
### Problem

When updating CI we noticed with the configuration updates, that if an environment variable such as `OPERATOR_ID` was defined with not value, it would override the values in the .env file.  This caused tests to fail when the relay tried to submit a transaction.  Ideally, the relay to fail fast, during bootstrap if required environment variables are not set.

### Solution

Consolidate environment variable loading to the start of the relay.  Fail to start if any required environment variables are not set.  Better to fail fast.

All environment variables should be handled in one class that can also validate them, list their usage in a --help option, and exit appropriately if a validation fails.  

We should also replace environment variables with a yml settings file.

### Needed validations
- Make sure that `CHAIN_ID`, `HEDERA_NETWORK`, `MIRROR_NODE_URL`, `OPERATOR_ID_MAIN`, `OPERATOR_KEY_MAIN`, `SERVER_PORT` are non-empty and in the expected format
- Make sure that if `OPERATOR_KEY_FORMAT` is not specified, the provided `OPERATOR_KEY_MAIN` is in the DER format
- Make sure that `HBAR_RATE_LIMIT_TINYBAR` is more than `HBAR_RATE_LIMIT_BASIC`, `HBAR_RATE_LIMIT_EXTENDED`, `HBAR_RATE_LIMIT_PRIVILEGED`

### Implementation Steps
 - [ ] Consolidate all environment variables into one service class
 - [ ] Add validations and exit on fail of validation, to the environment service class
 - [ ] Add help/usage message and option
 - [ ] Enhance to use yml instead of dotenv.

### Alternatives

_No response_

natanasow:
**Description**:

Currently, we are dynamically getting the environment variables and parsing/casting them every time. This is inefficient and also, more importantly, allows env configurations to be dynamically changed, which can modify the behavior of the relay based on the current values of the env vars, which can be a potential exploit.

Also, when updating CI we noticed with the configuration updates, that if an environment variable such as `OPERATOR_ID` was defined with not value, it would override the values in the .env file.  This caused tests to fail when the relay tried to submit a transaction.  Ideally, the relay to fail fast, during bootstrap if required environment variables are not set.

**Solution**:

Consolidate environment variable loading to the start of the relay.  Fail to start if any required environment variables are not set.  Better to fail fast.

All environment variables should be handled in one class that can also validate them, list their usage in a --help option, and exit appropriately if a validation fails.  

We should also replace environment variables with a yml settings file.

**Notes for reviews**:
- created a new package that holds the `config-service` singleton
- the `config-service` package is added as a dependency to `relay`, `server`, and `ws-server`
- in the perfect world, the `config-service` must be immutable and should not provide an interface for overriding or extending once-loaded environment variables but there are many tests where we dynamically override envs in order to test specific cases so we should `ConfigServiceTestHelper` as well
- the main review should be applied to the `config-service` package
- almost all of the changes are just replaced values from `process.env.<VAR_NAME>` to `ConfigService.get(<VAR_NAME>)`

**Needed validations**:
- Make sure that `CHAIN_ID`, `HEDERA_NETWORK`, `MIRROR_NODE_URL`, `OPERATOR_ID_MAIN`, `OPERATOR_KEY_MAIN`, `SERVER_PORT` are non-empty and in the expected format
- Make sure that if `OPERATOR_KEY_FORMAT` is not specified, the provided `OPERATOR_KEY_MAIN` is in the DER format
- Make sure that `HBAR_RATE_LIMIT_TINYBAR` is more than `HBAR_RATE_LIMIT_BASIC`, `HBAR_RATE_LIMIT_EXTENDED`, `HBAR_RATE_LIMIT_PRIVILEGED`

### Implementation Steps
 - [x] Consolidate all environment variables into one service class
 - [x] Add validations and exit on fail of validation, to the environment service class
 - [x] Add help/usage message and option
 - [x] There are several solid proposals by @AlfredoG87 that we should take into account as well, described [here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2956#pullrequestreview-2338214042.).
 - [ ] Enhance to use yml instead of dotenv.

cc: @ebadiere  

**Related issue(s)**:

Fixes #3023 https://github.com/hashgraph/hedera-json-rpc-relay/issues/2686

**Notes for reviewer**:

**Checklist**


packages/config-service/src/services/index.ts
```diff
@@ -60,7 +62,22 @@ export class ConfigService {
...
+    ValidationService.startUp(process.env);
+
+    // transform string values to typed envs, we'll get rid off things like that:
+    // - ConfigService.get('MY_CUSTOM_VAR_1') === 'true';
+    // - Number(ConfigService.get('MY_CUSTOM_VAR_2')) == 10;
+    this.envs = ValidationService.typeCasting(process.env);
+
+    // TODO: is this the right place for printing the envs?
+    for (let i in this.envs) {
+      console.log(LoggerService.maskUpEnv(i, this.envs[i]));
```
nathanklick, line 79: Potentially prints secrets to the console.

nathanklick (changes requested):

packages/config-service/src/services/index.ts
natanasow: Yep, the `maskUpEnv` will hide the secrets. Currently, the PR is like an architecture overview, the implementations are not ready yet. Once all TODOs are resolved, I'll mark it ready for review and ping you again :pray:.

natanasow:
@AlfredoG87 @ebadiere please read the TODOs as well.

packages/config-service/src/services/relayConfigs.ts
```diff
@@ -0,0 +1,103 @@
...
+  static REDIS_ENABLED: string = 'REDIS_ENABLED';
+  static REDIS_URL: string = 'REDIS_URL';
+  static REDIS_RECONNECT_DELAY_MS: string = 'REDIS_RECONNECT_DELAY_MS';
+  static DEBUG_API_ENABLED: string = 'DEBUG_API_ENABLED';
+  static FILTER_API_ENABLED: string = 'FILTER_API_ENABLED';
+  static MULTI_SET: string = 'MULTI_SET';
+  static GAS_PRICE_PERCENTAGE_BUFFER: string = 'GAS_PRICE_PERCENTAGE_BUFFER';
+}
+
+export class RelayEnvsTypes {
```
AlfredoG87, line 52: I think we can get rid of both objects if we define a `ConfigProperty` interface like this:
```
interface ConfigProperty {
  envName: string;
  type: string;
  required: boolean;
  defaultValue: string | number | boolean | Array<number> | null; // etc
}
```

and then a static `RelayConfig` like this:

```
export class RelayConfig {
  static CONFIG: Record<string, ConfigProperty> = {
    [RelayEnvs.CHAIN_ID]: {
      envName: RelayEnvs.CHAIN_ID,
      type: 'string',
      required: true,
      defaultValue: '',
    },
[RelayEnvs.REDIS_ENABLED]: {
      envName: RelayEnvs.REDIS_ENABLED,
      type: 'boolean',
      required: false,
      defaultValue: false,
    },
    [RelayEnvs.REDIS_URL]: {
      envName: RelayEnvs.REDIS_URL,
      type: 'string',
      required: true,
      defaultValue: '',
    },
    // Continue for all other configurations...
```

maybe is possible to remove the `envName` since is also the `key` but no harm in having it both ways I guess.

AlfredoG87 (commented):
I like the improvements just a few things to consider:
1. Default values.
2. Secrets
3. Is all RelayEnvs? I see some are WS (WS_) and some are Server (_RATE_LIMIT), maybe this being a Draft is just to capture the idea and there will be more Env classes?

leaved an example of a suggestion or idea on how we can acomplish to know if something is required, its default value and if is a secret. but feel free to suggest otherwise as you see it best!

AlfredoG87 (approved):
LGTM

packages/config-service/src/services/validationService.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+import { GlobalConfig } from './globalConfig';
+
+export class ValidationService {
+  static startUp(envs: NodeJS.Dict<string>): void {
+    // make sure that mandatory fields are passed
+    Object.entries(GlobalConfig.ENTRIES).forEach(([entryName, entryInfo]) => {
+      if (entryInfo.required) {
+        if (!envs[entryName]) {
```
ebadiere, line 28: Is an existing environment variable with no value going to pass this check?  Do we need to test for something like:
```
value === undefined || value === ''
```

packages/config-service/src/services/validationService.ts
```diff
@@ -0,0 +1,95 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { GlobalConfig } from './globalConfig';
+
+export class ValidationService {
+  static startUp(envs: NodeJS.Dict<string>): void {
+    // make sure that mandatory fields are passed
```
ebadiere, line 25: Nit: // Validate mandatory fields and their types

packages/config-service/src/services/validationService.ts
```diff
@@ -0,0 +1,95 @@
...
+    Object.entries(GlobalConfig.ENTRIES).forEach(([entryName, entryInfo]) => {
+      if (entryInfo.required) {
+        if (!envs[entryName]) {
+          throw new Error(`${entryName} is a mandatory and the relay can not operate without its value.`);
```
ebadiere, line 29: `the relay can not operate `
to 
`the relay cannot operate `

packages/config-service/src/services/validationService.ts
```diff
@@ -0,0 +1,95 @@
...
+      !envs[GlobalConfig.ENTRIES.OPERATOR_KEY_FORMAT.envName] &&
+      // @ts-ignore
+      !envs[GlobalConfig.ENTRIES.OPERATOR_KEY_MAIN.envName].match(/^[0-9a-f]{96,100}$/)
+    ) {
+      throw new Error(
+        `When ${GlobalConfig.ENTRIES.OPERATOR_KEY_FORMAT.envName} is not specified, the ${GlobalConfig.ENTRIES.OPERATOR_KEY_MAIN.envName} must be in DER format.`,
+      );
+    }
+
+    // make sure that HBAR_RATE_LIMIT_TINYBAR is more than HBAR_DAILY_LIMIT_BASIC, HBAR_DAILY_LIMIT_EXTENDED, HBAR_DAILY_LIMIT_PRIVILEGED
```
ebadiere, line 50: Perhaps `// Ensure HBAR_RATE_LIMIT_TINYBAR is not less than daily limits` to be succinct?

packages/config-service/src/services/validationService.ts
```diff
@@ -0,0 +1,95 @@
...
+          throw new Error(`${entryName} is a mandatory and the relay can not operate without its value.`);
+        }
+
+        if (entryInfo.type === 'number' && isNaN(Number(envs[entryName]))) {
+          throw new Error(`${entryName} must be a valid number.`);
+        }
+      }
+    });
+
+    // make sure that if OPERATOR_KEY_FORMAT is not specified, the provided OPERATOR_KEY_MAIN is in the DER format
```
ebadiere, line 38: We now run the tests using ECDSA keys.  I don't think DER format should be a requirement.

ebadiere (commented):
Thanks for doing this.  There are a few comments to address.

packages/config-service/src/services/validationService.ts
natanasow: As far as I understand from [this](https://github.com/hashgraph/hedera-json-rpc-relay/issues/3023) ticket, we assume that if `OPERATOR_KEY_FORMAT` is not defined, the passed operator key must be in `DER` format.

There isn't a correlation between ECDSA and DER/HEX in my opinion. An ECDSA key can be encoded in both formats (DER and HEX).

You're right, we're running the tests using the ECDSA key but we've specified [here](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/server/tests/localAcceptance.env#L5) that it's HEX encoded. Another approach could be using the same key but its DER format, assigning it to `OPERATOR_KEY_MAIN` and completely removing `OPERATOR_KEY_FORMAT` from the `localAcceptance.env`.

I may be wrong, what do you think? Or are you proposing to change the default format from DER to HEX? If it is, let's open a new ticket and triage it.

packages/config-service/src/services/validationService.ts
natanasow: Fixed.

Nana-EC (commented):
Really good progress 🙌🏾 .
Some questions

packages/config-service/src/services/validationService.ts
```diff
@@ -0,0 +1,95 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { GlobalConfig } from './globalConfig';
+
+export class ValidationService {
```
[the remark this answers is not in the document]
natanasow: Here we should validate at least the mandatory fields and their formats on startup. 

What we can remove and can be done in relay/server/ws-server as validations are HBAR daily limit checks.

What do you think?

.github/workflows/image-build.yml
```diff
@@ -40,7 +40,14 @@ jobs:
 
       - name: Start server
         run: |
-          docker run -d -p 7546:7546 --name relay relay:latest
+          docker run \
```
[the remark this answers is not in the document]
natanasow: `SERVER_PORT` can be set with a default value, and we can skip passing it here.

packages/config-service/src/services/validationService.ts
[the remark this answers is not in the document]
Nana-EC: Agreed, let's remove non mandatory or format checks

packages/config-service/tests/test.env
```diff
@@ -0,0 +1 @@
+CHAIN_ID=0x160c
```
Nana-EC, line 1: Is this just a random chain id or did you mean for 0x128 for mainnet?

Nana-EC (commented):
Looking good, waiting on the changes noted in comments

packages/config-service/tests/test.env
natanasow: It's a random chain id for unit testing purposes only.

ebadiere (dismissed):
Thank you for doing this.

packages/config-service/src/services/index.ts
```diff
@@ -0,0 +1,104 @@
...
+    ValidationService.startUp(process.env);
+
+    // transform string representations of env vars into proper types
+    this.envs = ValidationService.typeCasting(process.env);
+
+    // printing current env variables, masking up sensitive information
+    for (let i in this.envs) {
+      // @ts-ignore
+      logger.info(LoggerService.maskUpEnv(i, this.envs[i]));
+    }
```
victor-yanev, line 81: - `let` -> `const`
- `i` -> `name` or `envName`

packages/config-service/src/services/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    // validate mandatory fields
+    ValidationService.startUp(process.env);
+
+    // transform string representations of env vars into proper types
+    this.envs = ValidationService.typeCasting(process.env);
+
+    // printing current env variables, masking up sensitive information
+    for (let i in this.envs) {
+      // @ts-ignore
```
victor-yanev, line 79: QQ: Why is `@ts-ignore` needed here?

packages/config-service/src/services/validationService.ts
```diff
@@ -0,0 +1,65 @@
...
+  static startUp(envs: NodeJS.Dict<string>): void {
+    // validate mandatory fields and their types
+    Object.entries(GlobalConfig.ENTRIES).forEach(([entryName, entryInfo]) => {
+      if (entryInfo.required) {
+        if (!envs.hasOwnProperty(entryName)) {
```
victor-yanev, line 28: QQ: Shouldn't we check if the entry has a default value here?

```suggestion
        if (!envs.hasOwnProperty(entryName) && !entryInfo.defaultValue) {
```

victor-yanev (dismissed):
Great work! Looks good overall, I only have a few small questions and nit suggestions:

packages/config-service/src/services/index.ts
natanasow: Fixed.

packages/config-service/src/services/index.ts
natanasow: Removed.

packages/config-service/src/services/validationService.ts
natanasow: The `if` statement above checks if the field is required so the default value check is unnecessary IMO.

ebadiere (approved):
LG.  Let's try and get this merged!

packages/relay/src/lib/clients/cache/redisCache.ts
```diff
@@ -39,7 +40,8 @@ export class RedisCache implements IRedisCacheClient {
    */
   private readonly options = {
     // Max time to live in ms, for items before they are considered stale.
-    ttl: Number.parseInt(process.env.CACHE_TTL ?? constants.CACHE_TTL.ONE_HOUR.toString()),
+    // @ts-ignore
```
quiet-node, line 43: Q: I noticed `@ts-ignore` is added wherever `new ConfigService` is used, and there are so many of them now. I think we should fix this issue so we don’t have to ignore the execution every time we get an ENV var.

packages/server/tests/localAcceptance.env
```diff
@@ -27,6 +27,6 @@ LIMIT_DURATION=90000
 SERVER_REQUEST_TIMEOUT_MS=60000
 MEMWATCH_ENABLED=true
 WRITE_SNAPSHOT_ON_MEMORY_LEAK=false
-HBAR_RATE_LIMIT_TINYBAR=5000000000# 50 HBARs
-HBAR_RATE_LIMIT_DURATION=80000# 80 seconds
-HBAR_RATE_LIMIT_BASIC=4000000000# 40 HBARs
+HBAR_RATE_LIMIT_TINYBAR=12000000000 # 100 HBARs
+HBAR_RATE_LIMIT_DURATION=80000 # 80 seconds
+HBAR_RATE_LIMIT_BASIC=1100000000 # 90 HBARs
```
quiet-node:
```suggestion
HBAR_RATE_LIMIT_TINYBAR=12000000000 # 120 HBARs
HBAR_RATE_LIMIT_DURATION=80000 # 80 seconds
HBAR_RATE_LIMIT_BASIC=1100000000 # 110 HBARs
```

packages/server/tests/localAcceptance.env
```diff
@@ -27,6 +27,6 @@ LIMIT_DURATION=90000
 SERVER_REQUEST_TIMEOUT_MS=60000
 MEMWATCH_ENABLED=true
 WRITE_SNAPSHOT_ON_MEMORY_LEAK=false
-HBAR_RATE_LIMIT_TINYBAR=5000000000# 50 HBARs
-HBAR_RATE_LIMIT_DURATION=80000# 80 seconds
-HBAR_RATE_LIMIT_BASIC=4000000000# 40 HBARs
+HBAR_RATE_LIMIT_TINYBAR=12000000000 # 100 HBARs
```
quiet-node: Why is there a need to increase these variables? This could potentially cause the test `should eventually exhaust the HBAR limit for a BASIC user after multiple deployments of large contracts` to fail. The total budget might be too high, and the duration too short, which could result in the budget resetting during the deployment loop. Additionally, with a higher `HBAR_RATE_LIMIT_TINYBAR`, the exhaustion test could take much longer than necessary.

I’d suggest reverting it back. If you believe the change is necessary, then the duration should be updated accordingly to prevent the limiter from resetting during the deployment loop. Another option is to override these variables specifically for the `acceptancetest:hbarlimiter` npm command. Something like this:

```bash
"acceptancetest:hbarlimiter": "HBAR_RATE_LIMIT_TINYBAR=5000000000 nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter' --exit",
```

quiet-node (changes requested):

packages/server/tests/localAcceptance.env
natanasow: Leftovers from testing, reverted to the values defined in the `main` branch.

packages/server/tests/localAcceptance.env
natanasow: Leftovers from testing, reverted to the values defined in the `main` branch.

packages/config-service/src/services/loggerService.ts
```diff
@@ -0,0 +1,37 @@
...
+import { GlobalConfig } from './globalConfig';
+
+export class LoggerService {
+  public static readonly SENSITIVE_FIELDS = [
+    GlobalConfig.ENTRIES.OPERATOR_KEY_MAIN.envName,
+    GlobalConfig.ENTRIES.OPERATOR_KEY_ETH_SENDRAWTRANSACTION.envName,
+    GlobalConfig.ENTRIES.GITHUB_TOKEN.envName,
+  ];
+
+  static maskUpEnv(envName: string, envValue: string | undefined): string {
```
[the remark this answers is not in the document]
natanasow: Fixed.

packages/config-service/src/services/validationService.ts
```diff
@@ -0,0 +1,65 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { GlobalConfig } from './globalConfig';
+
+export class ValidationService {
+  static startUp(envs: NodeJS.Dict<string>): void {
```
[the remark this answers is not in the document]
natanasow: Fixed.

packages/config-service/src/services/validationService.ts
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        if (entryInfo.type === 'number' && isNaN(Number(envs[entryName]))) {
+          throw new Error(`${entryName} must be a valid number.`);
+        }
+      }
+    });
+  }
+
+  static typeCasting(envs: NodeJS.Dict<string>): NodeJS.Dict<any> {
```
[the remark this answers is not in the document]
natanasow: Fixed.

packages/relay/src/lib/clients/cache/redisCache.ts
natanasow: You're right. There are unnecessary transformations like this above. The `process.env.CACHE_TTL` will be parsed to the number in the config service and here we cast it to string (via `toString()`) and then again cast it to the number. All these places must be fixed in another issue/PR, we wanted to stick to as few as possible new changes that can break something that already works.

packages/config-service/src/services/index.ts
```diff
@@ -0,0 +1,103 @@
...
+   */
+  private constructor() {
+    const configPath = findConfig(ConfigService.envFileName);
+
+    if (!configPath) {
+      logger.warn('No .env file is found. The relay cannot operate without valid .env.');
+    }
+
+    // @ts-ignore
+    dotenv.config({ path: configPath });
```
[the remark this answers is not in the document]
quiet-node: oh it's okay I believe you! Thanks much for the clarification!

packages/config-service/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@hashgraph/json-rpc-config-service",
+  "version": "0.56.0-SNAPSHOT",
```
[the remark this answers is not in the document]
natanasow: Done.

ebadiere (approved):

quiet-node (dismissed):
LGTM thanks for the great work!

quiet-node (approved):

nathanklick (approved):

PR #3057, opened 2024-10-03
fix: change unit of transaction value result

### LINKED ISSUE #3053: Change units of the value in `eth_getTransactionByHash` result
### Description

When querying a transaction by hash we send wei. It is expected that the units of the value in the result will be the same. Instead it is returned in tinybars. It would be better if it's wei in and wei out so it can be consistent.

### Steps to reproduce

run query `eth_getTransactionByHash` and check value in the result

### Example

- **request**
```
curl https://testnet.hashio.io/api \
  --header "Content-Type: application/json" \
  --data '{"method":"eth_getTransactionByHash","params":["0x2a92a7c48567e0ce4cfd1932792fd1666793152d429091e39e4456a3953f60b2"],"id":1,"jsonrpc":"2.0"}'
```

- **result**
```
{
    "result": {
        "blockHash": "0xee2fd72987d2524ee51353a28edc6564334f87448aa49217fa8396a683675dc6",
        "blockNumber": "0x8fd5b4",
        "chainId": "0x128",
        "from": "0xba76c6980428a0b10cfc5d8ccb61949677a61233",
        "gas": "0xf561",
        "gasPrice": "0x0",
        "hash": "0x2a92a7c48567e0ce4cfd1932792fd1666793152d429091e39e4456a3953f60b2",
        "input": "0xc99907ec00000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000e0000000000000000000000000000000000000000000000000000000000000014000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000e746573742d6176616c616e636865000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002a3078413146353139373263313739373562353964433438316646323838426539456145323532446343440000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000000c68656c6c6f2068656465726100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
        "nonce": "0x6e5",
        "r": "0x1588d641c8b6177a2b8a8f58f7d509cf2cf6e5a4f3fe01fe489ca66dbe49f027",
        "s": "0x78ded747213c061859ad9967b8ce4e6ed5565099da4353300bc850ffa96b4524",
        "to": "0x2202534ea92e8f026032c460bf9f79d2f81425b4",
        "transactionIndex": "0x4",
        "type": "0x2",
        "v": "0x1",
        "value": "0x5f5e100",
        "yParity": "0x1",
        "accessList": [],
        "maxPriorityFeePerGas": "0x0",
        "maxFeePerGas": "0x13e"
    },
    "jsonrpc": "2.0",
    "id": "test_id"
}
```

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.56.0

### Operating system

macOS

nadezhdapopovaa:
**Description**:

Convert value of a transaction from tinybars to weibars for consistency.

**Related issue(s)**:

Fixes #3053 

**Notes for reviewer**:

**Checklist**


quiet-node (approved):
LGTM

packages/relay/src/formatters.ts
```diff
@@ -301,6 +301,10 @@ const getFunctionSelector = (data?: string): string => {
   return data.replace(/^0x/, '').substring(0, 8);
 };
 
+const tinybarsToWeibars = (value) => {
+  return value * constants.TINYBAR_TO_WEIBAR_COEF;
```
Nana-EC, line 305: Add some checks for invalid formats.
For example null, or negative numbers, how do we handle those?

Also there's a max value for tinybar we can cap any inputs to that to avoid invalid responses

packages/relay/tests/lib/formatters.spec.ts
```diff
@@ -737,4 +738,10 @@ describe('Formatters', () => {
       expect(result).to.deep.equal({ A: '1', B: '2', C: '3' });
     });
   });
+  
+  describe('getFunctionSelector', () => {
+    it('should convert tinybars to weibars', () => {
+      expect(tinybarsToWeibars(10)).to.eql(100000000000);
```
Nana-EC, line 744: Should have more tests for the scenarios I noted

Nana-EC (commented):
Expand the corner cases handling for the conversion and the tests

packages/relay/src/formatters.ts
```diff
@@ -301,6 +301,14 @@ const getFunctionSelector = (data?: string): string => {
   return data.replace(/^0x/, '').substring(0, 8);
 };
 
+const tinybarsToWeibars = (value: number | null) => {
```
konstantinabl, line 304: I know this addresses this comment https://github.com/hashgraph/hedera-json-rpc-relay/pull/3057#discussion_r1791757104
Not sure if its relevant for a util function

packages/relay/src/formatters.ts
Nana-EC: Good point, we can mostly rely on the MN source, but doesn't hurt to guard

packages/relay/src/formatters.ts
Nana-EC: I think it's fine for now.
We can relax in a follow up PR if you think it's overkill.

Nana-EC (approved):

nadezhdapopovaa:
issue resolved in this [PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3080)

PR #3058, opened 2024-10-03
feat: implement configuration to pre-populate the cache with spending plans

### LINKED ISSUE #3025: [HBAR Rate Limit Redesign] Improive logging logic for hbarLimit related classes
### Problem

Logging in the classes within the `hbarLimiter` repositories (e.g., `EthAddressHbarSpendingPlanRepository`, `HbarSpendingPlanRepository`, `IPAddressHbarSpendingPlanRepository`) appears to be duplicated and does not include a `requestId`. This makes it difficult to track logs during debugging.

### Solution

Improve logging for HbarLimitter related classes

### Alternatives

_No response_

### LINKED ISSUE #3055: [HBAR Rate Limit Redesign] Implement configuration to pre-populate the cache with spending plans

victor-yanev:
**Description**:

Implements logic to pre-populate the cache with spending plans specified in a JSON file.

Also fixes logs in `HbarSpendingPlanRepository`, `IPAddressHbarSpendingPlanRepository`, `EthAddressHbarSpendingPlanRepository` to log the request ID in messages for better tracking.

**Related issue(s)**:

Fixes #3055

Fixes #3025

**Notes for reviewer**:

Logs without configuration file:
```bash
[2024-10-03 13:12:55.959 +0000] INFO (relay/15455 on MacBook-Pro.local): Configurations successfully loaded
[2024-10-03 13:12:55.960 +0000] TRACE (hbar-rate-limit/15455 on MacBook-Pro.local): remainingBudget=11000000000 tℏ, resetTimestamp=1727961255960
[2024-10-03 13:12:55.980 +0000] INFO (relay/15455 on MacBook-Pro.local): SDK client successfully configured to "{\"127.0.0.1:50211\":\"0.0.3\"}" for account 0.0.2 with request timeout value: 10000
[2024-10-03 13:12:55.981 +0000] INFO (mirror-node/15455 on MacBook-Pro.local): Mirror Node client successfully configured to REST url: http://localhost:5551/api/v1/ and Web3 url: http://localhost:8545/api/v1/ 
[2024-10-03 13:12:55.982 +0000] INFO (relay/15455 on MacBook-Pro.local): Relay running with chainId=0x12a
[2024-10-03 13:12:55.985 +0000] WARN (relay/15455 on MacBook-Pro.local): Failed to load pre-configured spending plans: Configuration file not found at path "spendingPlansConfig.json"
```

Logs with a configuration file:
```bash
[2024-10-04 14:52:35.405 +0000] INFO (relay/15296 on MacBook-Pro.local): Configurations successfully loaded
[2024-10-04 14:52:35.406 +0000] TRACE (hbar-rate-limit/15296 on MacBook-Pro.local): remainingBudget=11000000000 tℏ, resetTimestamp=1728053635406
[2024-10-04 14:52:35.417 +0000] INFO (relay/15296 on MacBook-Pro.local): SDK client successfully configured to "{\"127.0.0.1:50211\":\"0.0.3\"}" for account 0.0.2 with request timeout value: 10000
[2024-10-04 14:52:35.418 +0000] INFO (mirror-node/15296 on MacBook-Pro.local): Mirror Node client successfully configured to REST url: http://localhost:5551/api/v1/ and Web3 url: http://127.0.0.1:8545/api/v1/ 
[2024-10-04 14:52:35.419 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  retrieving keys matching hbarSpendingPlan:* on findAllActiveBySubscriptionType call
[2024-10-04 14:52:35.419 +0000] INFO (relay/15296 on MacBook-Pro.local): Relay running with chainId=0x12a
[2024-10-04 14:52:35.423 +0000] TRACE (hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Creating HbarSpendingPlan with ID c758c095-342c-4607-9db5-867d7e90ab9d...
[2024-10-04 14:52:35.424 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching hbarSpendingPlan:c758c095-342c-4607-9db5-867d7e90ab9d:{"id":"c758c095-342c-4607-9db5-867d7e90ab9d","subscriptionType":"PRIVILEGED","createdAt":"2024-10-04T14:52:35.423Z","active":true,"spendingHistory":[],"amountSpent":0} on create for indefinite time (cache size: 1, max: 1000)
[2024-10-04 14:52:35.424 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Created HBAR spending plan "partner name" with ID "c758c095-342c-4607-9db5-867d7e90ab9d" and subscriptionType "PRIVILEGED"
[2024-10-04 14:52:35.424 +0000] TRACE (hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Creating HbarSpendingPlan with ID a68488b0-6f7d-44a0-87c1-774ad64615f2...
[2024-10-04 14:52:35.424 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching hbarSpendingPlan:a68488b0-6f7d-44a0-87c1-774ad64615f2:{"id":"a68488b0-6f7d-44a0-87c1-774ad64615f2","subscriptionType":"PRIVILEGED","createdAt":"2024-10-04T14:52:35.424Z","active":true,"spendingHistory":[],"amountSpent":0} on create for indefinite time (cache size: 2, max: 1000)
[2024-10-04 14:52:35.424 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Created HBAR spending plan "some other partner that has given us only eth addresses" with ID "a68488b0-6f7d-44a0-87c1-774ad64615f2" and subscriptionType "PRIVILEGED"
[2024-10-04 14:52:35.424 +0000] TRACE (hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Creating HbarSpendingPlan with ID af13d6ed-d676-4d33-8b9d-cf05d1ad7134...
[2024-10-04 14:52:35.424 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching hbarSpendingPlan:af13d6ed-d676-4d33-8b9d-cf05d1ad7134:{"id":"af13d6ed-d676-4d33-8b9d-cf05d1ad7134","subscriptionType":"EXTENDED","createdAt":"2024-10-04T14:52:35.424Z","active":true,"spendingHistory":[],"amountSpent":0} on create for indefinite time (cache size: 3, max: 1000)
[2024-10-04 14:52:35.424 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Created HBAR spending plan "supported project name" with ID "af13d6ed-d676-4d33-8b9d-cf05d1ad7134" and subscriptionType "EXTENDED"
[2024-10-04 14:52:35.424 +0000] TRACE (hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Creating HbarSpendingPlan with ID 7f665aa3-6b73-41d7-bf9b-92d04cdab96b...
[2024-10-04 14:52:35.424 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching hbarSpendingPlan:7f665aa3-6b73-41d7-bf9b-92d04cdab96b:{"id":"7f665aa3-6b73-41d7-bf9b-92d04cdab96b","subscriptionType":"EXTENDED","createdAt":"2024-10-04T14:52:35.424Z","active":true,"spendingHistory":[],"amountSpent":0} on create for indefinite time (cache size: 4, max: 1000)
[2024-10-04 14:52:35.424 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Created HBAR spending plan "some other supported project that has given us only ip addresses" with ID "7f665aa3-6b73-41d7-bf9b-92d04cdab96b" and subscriptionType "EXTENDED"
[2024-10-04 14:52:35.424 +0000] TRACE (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Updating associations for HBAR spending plan 'partner name' with ID c758c095-342c-4607-9db5-867d7e90ab9d...
[2024-10-04 14:52:35.424 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  retrieving keys matching ethAddressHbarSpendingPlan:* on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.424 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ethAddressHbarSpendingPlan:0x123:{"ethAddress":"0x123","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on save for indefinite time (cache size: 5, max: 1000)
[2024-10-04 14:52:35.424 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ethAddressHbarSpendingPlan:0x124:{"ethAddress":"0x124","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on save for indefinite time (cache size: 6, max: 1000)
[2024-10-04 14:52:35.424 +0000] TRACE (eth-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked ETH address 0x123 to HbarSpendingPlan with ID c758c095-342c-4607-9db5-867d7e90ab9d
[2024-10-04 14:52:35.424 +0000] TRACE (eth-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked ETH address 0x124 to HbarSpendingPlan with ID c758c095-342c-4607-9db5-867d7e90ab9d
[2024-10-04 14:52:35.424 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'partner name' with ETH address 0x123
[2024-10-04 14:52:35.424 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'partner name' with ETH address 0x124
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  retrieving keys matching ipAddressHbarSpendingPlan:* on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on save for indefinite time (cache size: 7, max: 1000)
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on save for indefinite time (cache size: 8, max: 1000)
[2024-10-04 14:52:35.425 +0000] TRACE (ip-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked new IP address to HbarSpendingPlan with ID c758c095-342c-4607-9db5-867d7e90ab9d
[2024-10-04 14:52:35.425 +0000] TRACE (ip-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked new IP address to HbarSpendingPlan with ID c758c095-342c-4607-9db5-867d7e90ab9d
[2024-10-04 14:52:35.425 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'partner name' with IP address
[2024-10-04 14:52:35.425 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'partner name' with IP address
[2024-10-04 14:52:35.425 +0000] TRACE (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Updating associations for HBAR spending plan 'some other partner that has given us only eth addresses' with ID a68488b0-6f7d-44a0-87c1-774ad64615f2...
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  retrieving keys matching ethAddressHbarSpendingPlan:* on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x123:{"ethAddress":"0x123","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x124:{"ethAddress":"0x124","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ethAddressHbarSpendingPlan:0x125:{"ethAddress":"0x125","planId":"a68488b0-6f7d-44a0-87c1-774ad64615f2"} on save for indefinite time (cache size: 9, max: 1000)
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ethAddressHbarSpendingPlan:0x126:{"ethAddress":"0x126","planId":"a68488b0-6f7d-44a0-87c1-774ad64615f2"} on save for indefinite time (cache size: 10, max: 1000)
[2024-10-04 14:52:35.425 +0000] TRACE (eth-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked ETH address 0x125 to HbarSpendingPlan with ID a68488b0-6f7d-44a0-87c1-774ad64615f2
[2024-10-04 14:52:35.425 +0000] TRACE (eth-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked ETH address 0x126 to HbarSpendingPlan with ID a68488b0-6f7d-44a0-87c1-774ad64615f2
[2024-10-04 14:52:35.425 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'some other partner that has given us only eth addresses' with ETH address 0x125
[2024-10-04 14:52:35.425 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'some other partner that has given us only eth addresses' with ETH address 0x126
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  retrieving keys matching ipAddressHbarSpendingPlan:* on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Updating associations for HBAR spending plan 'supported project name' with ID af13d6ed-d676-4d33-8b9d-cf05d1ad7134...
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  retrieving keys matching ethAddressHbarSpendingPlan:* on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x123:{"ethAddress":"0x123","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x124:{"ethAddress":"0x124","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x125:{"ethAddress":"0x125","planId":"a68488b0-6f7d-44a0-87c1-774ad64615f2"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x126:{"ethAddress":"0x126","planId":"a68488b0-6f7d-44a0-87c1-774ad64615f2"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ethAddressHbarSpendingPlan:0x127:{"ethAddress":"0x127","planId":"af13d6ed-d676-4d33-8b9d-cf05d1ad7134"} on save for indefinite time (cache size: 11, max: 1000)
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ethAddressHbarSpendingPlan:0x128:{"ethAddress":"0x128","planId":"af13d6ed-d676-4d33-8b9d-cf05d1ad7134"} on save for indefinite time (cache size: 12, max: 1000)
[2024-10-04 14:52:35.425 +0000] TRACE (eth-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked ETH address 0x127 to HbarSpendingPlan with ID af13d6ed-d676-4d33-8b9d-cf05d1ad7134
[2024-10-04 14:52:35.425 +0000] TRACE (eth-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked ETH address 0x128 to HbarSpendingPlan with ID af13d6ed-d676-4d33-8b9d-cf05d1ad7134
[2024-10-04 14:52:35.425 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'supported project name' with ETH address 0x127
[2024-10-04 14:52:35.425 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'supported project name' with ETH address 0x128
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  retrieving keys matching ipAddressHbarSpendingPlan:* on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"af13d6ed-d676-4d33-8b9d-cf05d1ad7134"} on save for indefinite time (cache size: 13, max: 1000)
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"af13d6ed-d676-4d33-8b9d-cf05d1ad7134"} on save for indefinite time (cache size: 14, max: 1000)
[2024-10-04 14:52:35.425 +0000] TRACE (ip-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked new IP address to HbarSpendingPlan with ID af13d6ed-d676-4d33-8b9d-cf05d1ad7134
[2024-10-04 14:52:35.425 +0000] TRACE (ip-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked new IP address to HbarSpendingPlan with ID af13d6ed-d676-4d33-8b9d-cf05d1ad7134
[2024-10-04 14:52:35.425 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'supported project name' with IP address
[2024-10-04 14:52:35.425 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'supported project name' with IP address
[2024-10-04 14:52:35.425 +0000] TRACE (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Updating associations for HBAR spending plan 'some other supported project that has given us only ip addresses' with ID 7f665aa3-6b73-41d7-bf9b-92d04cdab96b...
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  retrieving keys matching ethAddressHbarSpendingPlan:* on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x123:{"ethAddress":"0x123","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x124:{"ethAddress":"0x124","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x125:{"ethAddress":"0x125","planId":"a68488b0-6f7d-44a0-87c1-774ad64615f2"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x126:{"ethAddress":"0x126","planId":"a68488b0-6f7d-44a0-87c1-774ad64615f2"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x127:{"ethAddress":"0x127","planId":"af13d6ed-d676-4d33-8b9d-cf05d1ad7134"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.425 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ethAddressHbarSpendingPlan:0x128:{"ethAddress":"0x128","planId":"af13d6ed-d676-4d33-8b9d-cf05d1ad7134"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.426 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  retrieving keys matching ipAddressHbarSpendingPlan:* on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.426 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.426 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"c758c095-342c-4607-9db5-867d7e90ab9d"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.426 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"af13d6ed-d676-4d33-8b9d-cf05d1ad7134"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.426 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Returning cached value ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"af13d6ed-d676-4d33-8b9d-cf05d1ad7134"} on populatePreconfiguredSpendingPlans call
[2024-10-04 14:52:35.426 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"7f665aa3-6b73-41d7-bf9b-92d04cdab96b"} on save for indefinite time (cache size: 15, max: 1000)
[2024-10-04 14:52:35.426 +0000] TRACE (localLRUCache/15296 on MacBook-Pro.local):  Caching ipAddressHbarSpendingPlan:<REDACTED>:{"ipAddress":"<REDACTED>","planId":"7f665aa3-6b73-41d7-bf9b-92d04cdab96b"} on save for indefinite time (cache size: 16, max: 1000)
[2024-10-04 14:52:35.426 +0000] TRACE (ip-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked new IP address to HbarSpendingPlan with ID 7f665aa3-6b73-41d7-bf9b-92d04cdab96b
[2024-10-04 14:52:35.426 +0000] TRACE (ip-address-hbar-spending-plan-repository/15296 on MacBook-Pro.local):  Linked new IP address to HbarSpendingPlan with ID 7f665aa3-6b73-41d7-bf9b-92d04cdab96b
[2024-10-04 14:52:35.426 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'some other supported project that has given us only ip addresses' with IP address
[2024-10-04 14:52:35.426 +0000] INFO (hbar-spending-plan-config-service/15296 on MacBook-Pro.local): Associated HBAR spending plan 'some other supported project that has given us only ip addresses' with IP address
[2024-10-04 14:52:35.426 +0000] INFO (relay/15296 on MacBook-Pro.local): Pre-configured spending plans populated successfully
```

**Checklist**


packages/relay/src/lib/config/preconfiguredSpendingPlans.ts
```diff
@@ -0,0 +1,109 @@
...
+ * @throws {Error} If the configuration file is not found or cannot be read.
+ */
+const loadSpendingPlansConfig = (): SpendingPlanConfig[] => {
+  const path = findConfig(SPENDING_PLANS_CONFIG_FILE);
+  if (!path || !fs.existsSync(path)) {
+    throw new Error(`Configuration file not found at path: ${path || SPENDING_PLANS_CONFIG_FILE}`);
+  }
+  const rawData = fs.readFileSync(path, 'utf-8');
+  return JSON.parse(rawData) as SpendingPlanConfig[];
+};
```
ebadiere, line 45: I think we should handle any JSON syntax parsing errors gracefully.  Perhaps:
```
  try {
    const rawData = fs.readFileSync(path, 'utf-8');
    return JSON.parse(rawData) as SpendingPlanConfig[];
  } catch (error) {
    throw new Error(`Failed to parse JSON from ${path}: ${error.message}`);
  }
  ```

File: packages/relay/src/lib/config/preconfiguredSpendingPlans.ts [hunk 2]
```diff
@@ -0,0 +1,109 @@
...
+      for (const ethAddress of ethAddresses) {
+        if (!(await ethAddressHbarSpendingPlanRepository.existsByAddress(ethAddress, requestDetails))) {
+          await ethAddressHbarSpendingPlanRepository.save({ ethAddress, planId: spendingPlan.id }, requestDetails, ttl);
+          logger.info(
+            `${requestDetails.formattedRequestId} Associated HBAR spending plan '${name}' with ETH address ${ethAddress}`,
+          );
+        } else {
+          logger.warn(
+            `${requestDetails.formattedRequestId} Skipping ETH address ${ethAddress} as it already has an HBAR spending plan associated with it`,
+          );
```
ebadiere, line 91: Nit: Is this a warning or useful "info"?

victor-yanev: It's a warning, because if we get here, then the JSON file is most likely wrongly configured.

Each unique ETH address should be linked only to one plan, we cannot have it linked to two plans because we wouldn't know which plan to use in this case.

packages/relay/src/lib/config/preconfiguredSpendingPlans.ts
victor-yanev: Good point, I will update this

packages/relay/src/lib/config/preconfiguredSpendingPlans.ts
```diff
@@ -0,0 +1,109 @@
...
+const PRE_CONFIGURED_SPENDING_PLANS_TTL = -1; // Never expire
+const SPENDING_PLANS_CONFIG_FILE = 'spendingPlansConfig.json';
+
+/**
+ * Loads the pre-configured spending plans from a JSON file.
+ * @returns {SpendingPlanConfig[]} An array of spending plan configurations.
+ * @throws {Error} If the configuration file is not found or cannot be read.
+ */
+const loadSpendingPlansConfig = (): SpendingPlanConfig[] => {
+  const path = findConfig(SPENDING_PLANS_CONFIG_FILE);
```
ebadiere, line 39: Should we allow the operator to configure where to place the spending plans file on their file system?  They may want special protections on it.  Pass in the path from an environment variable?
```
const loadSpendingPlansConfig = (configPath?: string): SpendingPlanConfig[] => {
  const path = configPath || findConfig(SPENDING_PLANS_CONFIG_FILE);
  ```

**Review by ebadiere — COMMENTED** [review]
Good work.  I have a few questions.

File: packages/relay/src/lib/config/preconfiguredSpendingPlans.ts (hunk 3 above)
Comment (victor-yanev) [reply packages/relay/src/lib/config/preconfiguredSpendingPlans.ts]: Done, made the filename configurable through `HBAR_SPENDING_PLANS_CONFIG_FILE` env variable

File: packages/relay/src/lib/types/spendingPlanConfig.ts [hunk 4]
```diff
@@ -0,0 +1,64 @@
...
+   * @type {string[]}
+   * @optional
+   */
+  ipAddresses?: string[];
+
+  /**
+   * The spending plan tier.
+   * @type {SubscriptionType}
+   */
+  subscriptionTier: SubscriptionType;
```
quiet-node, line 53: Wondering why this is called `subscriptionTier` instead of `subscriptionType` like the other references. The `HbarSpendingPlan` class refers to it as `subscriptionType`, and the enum is named `SubscriptionType`. Other spending repositories also use `SubscriptionType`. I would recommend renaming it to `subscriptionType` for better consistency across the codebase.

spendingPlansConfig.json.example
```diff
@@ -0,0 +1,24 @@
+[
```
quiet-node, line 1: qq: I understand this is an example file, but why use `.json.example` instead of `.example.json` so it would allow for syntax highlighting and better tooling support?

quiet-node (changes requested):

packages/relay/src/lib/config/preconfiguredSpendingPlans.ts
```diff
@@ -0,0 +1,127 @@
...
+      continue;
+    }
+
+    const spendingPlan = await hbarSpendingPlanRepository.create(subscriptionTier, requestDetails, ttl);
+    logger.info(`${requestDetails.formattedRequestId} Created HBAR spending plan '${name}' with ID ${spendingPlan.id}`);
+
+    if (ethAddresses) {
+      for (const ethAddress of ethAddresses) {
+        if (!(await ethAddressHbarSpendingPlanRepository.existsByAddress(ethAddress, requestDetails))) {
+          await ethAddressHbarSpendingPlanRepository.save({ ethAddress, planId: spendingPlan.id }, requestDetails, ttl);
```
quiet-node, line 102: Q: So, this means that each spending plan can be associated with multiple `ethAddresses`. For instance, a plan with `planID apple-36` is associated with the addresses `0x33`, `0x66`, and `127.0.0.1`. When these addresses make requests to the relay, all expenses from these addresses are recorded and attributed to the `apple-36` plan, correct?

Secondly, I noticed that the `spendingPlan: IDetailedHbarSpendingPlan` only tracks `amountSpent` and `spendingHistory`, but neither of these seems to reference the addresses causing the expenses. Just thinking out loud here—wouldn't it be beneficial if the `IDetailedHbarSpendingPlan` could track how much each address has spent? From the current implementation, if in any cases that the Relay wants to know how much an individual address has spent, it would have to query the cache, correct? Now including the addresses in the usage history of the plan could remove the need to query the cache and potentially offer some benefits for metrics.

packages/relay/src/lib/config/preconfiguredSpendingPlans.ts
victor-yanev: Sure, changed it to a `trace` log.

packages/relay/src/lib/config/preconfiguredSpendingPlans.ts
```diff
@@ -0,0 +1,127 @@
...
+
+const PRE_CONFIGURED_SPENDING_PLANS_TTL = -1; // Never expire
+const DEFAULT_SPENDING_PLANS_CONFIG_FILE = 'spendingPlansConfig.json';
+
+/**
+ * Loads the pre-configured spending plans from a JSON file.
+ * @returns {SpendingPlanConfig[]} An array of spending plan configurations.
+ * @throws {Error} If the configuration file is not found or cannot be read.
+ */
+const loadSpendingPlansConfig = (filename?: string): SpendingPlanConfig[] => {
```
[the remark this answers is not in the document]
victor-yanev: My bad, changed the description of `HBAR_SPENDING_PLANS_CONFIG_FILE` in the documentation, it's supposed to be a filename, not a path - `findConfig` will automatically resolve the path.

packages/relay/src/lib/config/preconfiguredSpendingPlans.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  for (const plan of spendingPlans) {
+    const { name, ethAddresses, ipAddresses, subscriptionTier } = plan;
+
+    if (!ethAddresses?.length && !ipAddresses?.length) {
+      logger.warn(
+        `${requestDetails.formattedRequestId} Skipping HBAR spending plan '${name}' as it has no ethAddresses or ipAddresses`,
+      );
+      continue;
+    }
```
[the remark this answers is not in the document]
victor-yanev: Done, I also changed the validations to check the size of the array to make sure we don't get a spending plan with two empty arrays for `ethAddresses` and `ipAddresses`

packages/relay/src/lib/config/preconfiguredSpendingPlans.ts
```diff
@@ -0,0 +1,127 @@
...
+  logger: Logger,
+  hbarSpendingPlanRepository: HbarSpendingPlanRepository,
+  ethAddressHbarSpendingPlanRepository: EthAddressHbarSpendingPlanRepository,
+  ipAddressHbarSpendingPlanRepository: IPAddressHbarSpendingPlanRepository,
+): Promise<void> => {
+  const requestDetails = new RequestDetails({ requestId: '', ipAddress: '' });
+  const ttl = PRE_CONFIGURED_SPENDING_PLANS_TTL;
+
+  const spendingPlans = loadSpendingPlansConfig(process.env.HBAR_SPENDING_PLANS_CONFIG_FILE);
+  validateSpendingPlanConfig(spendingPlans);
```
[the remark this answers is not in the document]
victor-yanev: Hmmm, I think we should block any pre-configurations in this case and let the user know that he has a wrong configuration and incentivize him to fix it

packages/relay/src/lib/types/spendingPlanConfig.ts
victor-yanev: Done, changed to `subscriptionType`

spendingPlansConfig.json.example
victor-yanev: Valid point, changed it to `spendingPlansConfig.example.json`

packages/relay/src/lib/config/preconfiguredSpendingPlans.ts
victor-yanev: Yes, this is the way it is supposed to work.

Btw, we have disabled tracking of spending history at the moment (as discussed in the daily a week ago) because we don't know yet how we want to use it, but we could include this information in the `HbarSpendingRecords` if we want to track this:
```diff
export class HbarSpendingRecord implements IHbarSpendingRecord {
  amount: number;
  timestamp: Date;
+  ethAddress?: string;
+  ipAddress?: string;

  constructor(data: IHbarSpendingRecord) {
    this.amount = data.amount;
    this.timestamp = new Date(data.timestamp);
+    this.ethAddress = data.ethAddress;
+    this.ipAddress = data.ipAddress;
  }
}
```

We could discuss it today and open a new issue for that, as it is way out of scope for the current issue.

ebadiere (commented):
LG.

packages/relay/src/lib/config/preconfiguredSpendingPlans.ts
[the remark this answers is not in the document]
quiet-node: Fair point!

spendingPlansConfig.example.json
```diff
@@ -0,0 +1,28 @@
+[
+  {
+    "id": "c758c095-342c-4607-9db5-867d7e90ab9d",
```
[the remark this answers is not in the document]
victor-yanev: CC: @Nana-EC @ebadiere 

Well, there are two approaches in which this could be handled:
1. Clear all spending plans from the cache on every start-up and re-populate them from the config file:
   - in this case we don't need to specify IDs in the config file, they can be randomly generated on each startup  
   - historical logs on mainnet and testnet will be harder to debug because we would have generated new IDs on redeploys / restarted pods and the IDs we have from the logs before this moment wouldn't be valid anymore
   - if Kubernetes pods are restarted - the cache would be re-populated with the same spending plans but with new IDs and we would lose track of the current spending as it will also be restarted
2. Only populate the "delta" from the config file (i.e., delete "obsolete" spending plans which are no longer present in the config file and create new spending plans which were not specified before):
   - in this case we need to have an ID specified in the config file, otherwise we cannot know if a given spending plan was already populated or not in the cache
   - historical logs on mainnet and testnet will be easy to debug since the same ID we are seeing in the logs, will be still valid and we can retrieve the spending plan from the cache by that ID
   - the Kubernetes pods can be freely restarted without affecting the behavior of the hbar limiter, as old spending plans wouldn't be cleared and only new ones that have appeared in the config file will be populated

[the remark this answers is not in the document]
victor-yanev: > How would someone unfamiliar with the codebase, or loading the config file for the first time, come up with such IDs

You don't need to be familiar with the codebase to come up with a random UUID: https://www.uuidgenerator.net/version4

> I was under the impression that these IDs are auto-generated by the Relay and only used internally.

They are if an `"id"` field is not specified in the config file.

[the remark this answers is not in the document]
ebadiere: If I understand this correctly, with option 2, the config file can be the single source of truth regarding the spending plan.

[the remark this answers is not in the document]
quiet-node: discussed off line and agreed on option 2 mentioned above. However, needed to come up with a plan to construct the plan ID beforehand and make it as a required field.

ebadiere (approved):
LG, but needs a conflict resolution.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -0,0 +1,315 @@
+/*
```
quiet-node, line 1: If this is also a service, should we consider moving it into the `services/` folder? Perhaps something like `services/config/hbarSpendService.ts`?

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -0,0 +1,315 @@
...
+import { EthAddressHbarSpendingPlanRepository } from '../db/repositories/hbarLimiter/ethAddressHbarSpendingPlanRepository';
+import { IPAddressHbarSpendingPlanRepository } from '../db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository';
+import { RequestDetails } from '../types';
+import { Logger } from 'pino';
+import { SubscriptionType } from '../db/types/hbarLimiter/subscriptionType';
+import { IDetailedHbarSpendingPlan } from '../db/types/hbarLimiter/hbarSpendingPlan';
+
+export class HbarSpendingPlanConfigService {
+  private readonly TTL = -1; // -1 means no TTL, i.e. the data will not expire
+  private readonly DEFAULT_SPENDING_PLANS_CONFIG_FILE = 'spendingPlansConfig.json';
```
quiet-node, line 34: Let's not foget JSDoc for these vars

quiet-node, line 34: Since `DEFAULT_SPENDING_PLANS_CONFIG_FILE` is only used once for the filename in `loadSpendingPlansConfig()`, can we simplify it to something like:

```suggestion
  private readonly TTL = -1; // -1 means no TTL, i.e. the data will not expire
  private readonly SPENDING_PLANS_CONFIG_FILE = process.env.HBAR_SPENDING_PLANS_CONFIG_FILE || 'spendingPlansConfig.json';
```
Then instead of initializing a new filename variable in `loadSpendingPlansConfig()`, we can simply reuse `this.SPENDING_PLANS_CONFIG_FILE`.

Just a suggestion I'm fine with current flow.

packages/relay/src/lib/db/repositories/hbarLimiter/ethAddressHbarSpendingPlanRepository.ts
```diff
@@ -27,7 +27,7 @@ import { RequestDetails } from '../../../types';
 
 export class EthAddressHbarSpendingPlanRepository {
   private readonly collectionKey = 'ethAddressHbarSpendingPlan';
-  private readonly threeMonthsInMillis = 90 * 24 * 60 * 60 * 1000;
+  private readonly oneDayInMillis = 24 * 60 * 60 * 1000;
```
quiet-node, line 30: remove if not used

packages/relay/src/lib/clients/cache/localLRUCache.ts
```diff
@@ -105,11 +106,10 @@ export class LocalLRUCache implements ICacheClient {
   public async get(key: string, callingMethod: string, requestDetails: RequestDetails): Promise<any> {
     const value = this.cache.get(key);
     if (value !== undefined) {
-      this.logger.trace(
-        `${requestDetails.formattedRequestId} returning cached value ${key}:${JSON.stringify(
-          value,
-        )} on ${callingMethod} call`,
-      );
+      const censoredKey = key.replace(Utils.IP_ADDRESS_REGEX, '<REDACTED>');
+      const censoredValue = JSON.stringify(value).replace(/"ipAddress":"[^"]+"/, '"ipAddress":"<REDACTED>"');
```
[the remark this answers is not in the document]
victor-yanev: I believe we haven't been saving any IP addresses in the cache before this epic?

packages/relay/src/lib/clients/cache/localLRUCache.ts
```diff
@@ -105,11 +106,10 @@ export class LocalLRUCache implements ICacheClient {
...
+      const message = `Returning cached value ${censoredKey}:${censoredValue} on ${callingMethod} call`;
```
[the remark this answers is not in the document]
victor-yanev: Nope, one is for the key which is used to store the entry in the cache, and the other for the value which is the "stringified" JSON object that is stored. See example logs in the description of the PR.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
victor-yanev: I thought about this as well, but it seems like it would be a better option to have all kinds of configurations stored in only one place - inside the `lib/config` directory. Otherwise configurations will be all scattered around in the folders and will be harder to track.

packages/relay/src/lib/clients/cache/localLRUCache.ts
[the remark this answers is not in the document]
quiet-node: That's also my understanding.

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -0,0 +1,523 @@
+/*
```
quiet-node, line 1: Great test coverage 💪🏼

quiet-node (commented):
Great work! One concern

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
victor-yanev: Thanks brother 💪🏼

packages/relay/src/index.ts
```diff
@@ -41,6 +41,8 @@ export interface Relay {
   subs(): Subs | undefined;
 
   mirrorClient(): MirrorNodeClient;
+
+  populatePreconfiguredSpendingPlans(): Promise<void>;
```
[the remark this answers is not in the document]
victor-yanev: Valid point, that’s how I did it initially but sonar flagged it as an issue that we should avoid executing asynchronous methods in the constructor, that’s why I moved it to a method which can be called on demand by the server. The method still abstracts away any logic related to the cache.

If you think it would be better to force the initialisation always by calling it asynchronously through the constructor I can move it back. Let me know your thoughts!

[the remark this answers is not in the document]
quiet-node: I understand. You're absolutely right it’s generally not a good practice to have asynchronous operations within a constructor. Ah bummer! Well I do agree with the concern, I still feel that having a specific function within the Relay object, alongside the Ethereum-specific modules, doesn't sit quite right for me. 

If we must include it, how about we abstract this logic into a `preconfiguration service module`. Urgh I know, another service!! but if we think about it, haven't we been discussing the transition of environment variables to a more structured format like YAML or some sort, where every variable is populated at boot time. Now this `preconfiguration service module` could serve as the foundation for that transition. Additionally, a dedicated service module would align better alongside the Ethereum-specific modules than just a standalone function.

I also asked my AI co-pilot this concern, and the copilot suggested using a static factory method to create and initialize the instance within the factory, rather than directly initializing it through the constructor. Here's how we can implement it:

1. Convert the constructor to private:
```typescript
private constructor(
    private readonly logger: Logger,
    register: Registry,
  ) {
    // ... existing constructor logic ...
  }
```

2. Create the static factory method:
```typescript
  public static async create(logger: Logger, register: Registry): Promise<RelayImpl> {
    const relay = new RelayImpl(logger, register);
    await relay.populatePreconfiguredSpendingPlans();
    return relay;
  }
```
This static factory instantiates the class and then calls `populatePreconfiguredSpendingPlans()` before returning the instance. It ensures that the factory still returns a `Relay` object while maintaining the logic for pre-populating the cache.

3. Simply update the Relay initialization logic in `server.ts` (HTTP server) and `webSocketServer.ts` (WS server):
```typescript
- const relay: Relay = new RelayImpl(logger.child({ name: 'relay' }), register);
+ const relay: Relay = await RelayImpl.create(logger, register);
```

So yeah, those are the two suggestions. Personally, I prefer the static factory approach. Initially, I was a bit skeptical, but it makes perfect sense and seems like a reasonable solution to address these concerns. Plus, the co-pilot mentioned this is a "common pattern." We gotta trust the AI man 😆. Jokes aside though I believe I've seen this logic used before.

packages/relay/src/lib/relay.ts
```diff
@@ -161,15 +196,45 @@ export class RelayImpl implements Relay {
...
+
+    this.ethAddressHbarSpendingPlanRepository = new EthAddressHbarSpendingPlanRepository(
+      this.cacheService,
+      logger.child({ name: 'eth-address-hbar-spending-plan-repository' }),
+    );
+
+    this.ipAddressHbarSpendingPlanRepository = new IPAddressHbarSpendingPlanRepository(
+      this.cacheService,
+      logger.child({ name: 'ip-address-hbar-spending-plan-repository' }),
+    );
```
quiet-node, line 212: Since these repository instances are only used within the constructor and not elsewhere in the code, I believe initializing them as local constants is sufficient, rather than declaring them as class properties. This reduces their scope and save some memory, as they are not needed outside the constructor.

```suggestion
    const hbarSpendingPlanRepository = new HbarSpendingPlanRepository(
      this.cacheService,
      logger.child({ name: 'hbar-spending-plan-repository' }),
    );
    const ethAddressHbarSpendingPlanRepository = new EthAddressHbarSpendingPlanRepository(
      this.cacheService,
      logger.child({ name: 'eth-address-hbar-spending-plan-repository' }),
    );
    const ipAddressHbarSpendingPlanRepository = new IPAddressHbarSpendingPlanRepository(
      this.cacheService,
      logger.child({ name: 'ip-address-hbar-spending-plan-repository' }),
    );
```

packages/relay/src/lib/types/spendingPlanConfig.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Determines if the provided object is a valid spending plan configuration.
+ * @param plan - The object to validate.
+ * @returns {boolean} - True if the object is a valid {@link SpendingPlanConfig}, false otherwise.
+ */
+export function isValidSpendingPlanConfig(plan: any): plan is SpendingPlanConfig {
+  return (
+    plan &&
+    typeof plan.id === 'string' &&
```
quiet-node, line 69: Should we validate `plan.id` against a UUID regex instead of just checking for type `string` to enhance security?

On another note, I still feel that requiring maintainers to manually construct a UUID is not the most convenient approach. For the initial loading of the configuration plan, maintainers would need to use an external tool to generate a UUID for each plan, and then continuously manage the UUID throughout the process, correct?

Is it possible to use the `name` field to look up spending plans instead? In real scenarios, each project or partner would typically have a unique name, and it's much easier to input and maintain the name of a project or organization than a randomly generated UUID. 

What are your thoughts on this?

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -0,0 +1,356 @@
...
+   * Loads the pre-configured spending plans from a JSON file.
+   *
+   * @returns {SpendingPlanConfig[]} An array of spending plan configurations.
+   * @throws {Error} If the configuration file is not found or cannot be read or parsed.
+   * @private
+   */
+  private loadSpendingPlansConfig(): SpendingPlanConfig[] {
+    const configPath = findConfig(this.SPENDING_PLANS_CONFIG_FILE);
+    if (!configPath || !fs.existsSync(configPath)) {
+      throw new Error(`Configuration file not found at path "${configPath ?? this.SPENDING_PLANS_CONFIG_FILE}"`);
```
quiet-node, line 108: If the Relay starts without any special projects or partners, meaning no `spendingPlansConfig.json` is required, this line would throw an irrelevant error. Since it’s intended to operate without special projects in such cases, we probably should simply ignore the absence of the configuration file instead of throwing an error.

packages/relay/src/index.ts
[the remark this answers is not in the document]
victor-yanev: Not a bad idea, but:
1. This will be a breaking change for people using previous versions of our repo - they will have to replace usages of the constructor with the factory method everywhere in their code since the constructor is no longer going to be public
2. Currently `await` statements cannot be used inside `server.ts` and `webSocketServer.ts`:
```
TS1378: Top-level await expressions are only allowed when the module option is set to es2022, esnext, system, node16, or nodenext, and the target option is set to es2017 or higher.
```

I think I'll just remove this new method from the relay's interface, move back this logic to the constructor and ignore the sonar warning.

packages/relay/src/lib/types/spendingPlanConfig.ts
victor-yanev: > Should we validate plan.id against a UUID regex instead of just checking for type string to enhance security?

No, as we discussed in the last meeting any unique string works and the relay operator will be responsible for what he selects for the ID, we can leave a recommendation for using a UUID string but it's up to them what they set it to.

> On another note, I still feel that requiring maintainers to manually construct a UUID is not the most convenient approach. For the initial loading of the configuration plan, maintainers would need to use an external tool to generate a UUID for each plan, and then continuously manage the UUID throughout the process, correct?

What do you mean by "continuously manage the UUID throughout the process" they only have to generate it once and enter it in the JSON file - there is no maintaining after that, it remains both in the JSON file and in the cache once it is created.

> Is it possible to use the name field to look up spending plans instead? In real scenarios, each project or partner would typically have a unique name, and it's much easier to input and maintain the name of a project or organization than a randomly generated UUID.

Using names as keys isn't ideal because they aren't guaranteed to be unique or stable—they can change or have duplicates, leading to cache issues. UUIDs are specifically designed for this purpose, ensuring reliable and consistent key generation for caching. 

Also, `"name"` is only an accessory field which is used only in the logs of HbarSpendingPlanConfigService` for better readability by people. These names are not saved in the cache or used anywhere else.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
victor-yanev: > this line would throw an irrelevant error

No, it won't throw an error because the error is caught when calling it. Please, take a look at `relay.ts` where the `populatePreconfiguredSpendingPlans()` is called:
```typescript
  public async populatePreconfiguredSpendingPlans(): Promise<void> {
    return this.hbarSpendingPlanConfigService
      .populatePreconfiguredSpendingPlans()
      .then(() => this.logger.info('Pre-configured spending plans populated successfully'))
      .catch((e) => this.logger.warn(`Failed to load pre-configured spending plans: ${e.message}`));
  }
```

Even, there are tests in `relay.spec.ts` which confirm that when no configuration file is provided the error thrown from the `HbarSpendingPlanConfigService` is only logged and not re-thrown:
```typescript
    describe('when no configuration file is provided', () => {
      beforeEach(() => {
        process.env.HBAR_SPENDING_PLANS_CONFIG_FILE = 'nonExistingFile.json';
        relay = new RelayImpl(logger, register);
      });

      it('should not throw an error', async () => {
        const populatePreconfiguredSpendingPlansSpy = sinon.spy(
          relay['hbarSpendingPlanConfigService'],
          'populatePreconfiguredSpendingPlans',
        );

        await expect(relay.populatePreconfiguredSpendingPlans()).to.not.be.rejected;

        expect(populatePreconfiguredSpendingPlansSpy.calledOnce).to.be.true;
        expect(loggerSpy.error.calledWith(`Configuration file not found at path "nonExistingFile.json"`));
      });
    });
```

packages/relay/src/index.ts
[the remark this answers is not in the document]
victor-yanev: Moved back the call to be made inside the constructor.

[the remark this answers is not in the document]
quiet-node: Fair point! Yes we can potentially discuss to improve this in future PRs.

packages/relay/src/lib/types/spendingPlanConfig.ts
quiet-node: > we can leave a recommendation for using a UUID string but it's up to them what they set it to
Ah, I see! I initially thought it had to be in UUID format, which is why I was concerned about the lack of constraints around it being a UUID.

All the other points make perfect sense, thanks for clarifying!

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
quiet-node: Oh, I apologize for not being clearer! Yes, I see that the error was handled correctly and properly! What I was suggesting is to ignore it if no config file is provided. For instance, if a Relay doesn't have any special projects, I wouldn't expect a `spendingPlansConfig.json` file to be mandatory. The file should be optional in that case. 

With this current flow, when the file isn't present, the log shows a `warn` message saying "Failed to load pre-configured...". We could make this a little more friendly by logging an `info` or `trace` message instead, saying things like "No spending plan config file found, skipping pre-populating logic." (well the wordings can be different) but more important we could choose to completely ignore it and only care/log something when the file is actually provided.

victor-yanev: Oh I get what you mean now, yeah I agree. 

Changed it so that we don’t log any warnings, but only a trace log in cases the spending plan configuration file is missing.

konstantinabl (approved):

quiet-node (approved):
Wonderful! Everything works out well! Thanks for the great work!

konstantinabl (approved):

Issue #2947, opened 2024-09-05
Create CI workflow to run a post release test suite on mainnet


ebadiere:
### Problem

When I release a new version of the JSON RPC Relay on mainnet I want to know that the most basic functions are working properly. 

### Solution

A button (workflow dispatch) that triggers a mainnet release acceptance test suite in CI.  Account credentials need to be handled appropriately

PR #3060, opened 2024-10-03
feat: Add release testing workflow for public networks

beeradb:
This PR refactors the public acceptance workflow to do the following:

- Adds support for mainnet
- Run only the "release_tests" tag against public networks.
- Use the new `TEST_INITIAL_ACCOUNT_STARTING_BALANCE` attribute to ensure tests will allow runs with only 500 hbar in the operator account.
- Provide configurations for mainnet environment.

Fixes: #2947


packages/server/tests/mainnetAcceptance.env
```diff
@@ -0,0 +1,7 @@
+HEDERA_NETWORK=mainnet
+CHAIN_ID=0x00
```
quiet-node, line 2: should we match the mainnet chain ID 0x127?

ebadiere (commented):
LG, but Logan's question about chainId should be resolved.

packages/server/tests/mainnetAcceptance.env
beeradb: good catch - fixed

ebadiere (approved):
LG.

quiet-node (approved):

PR #3084, opened 2024-10-09
feat: Added consensus node information to the logging.

### LINKED ISSUE #3083: Add consensus node to logged information in the failed transaction handling.
### Problem

It would be helpful to see what node the client was submitting a transaction to if it fails.  

### Solution

Capture the consensus node in the logging.

### Alternatives

_No response_

ebadiere:
Adds the consensus node to logging information on failed calls.
e.g.
```
{"level":50,"time":1728573006569,"pid":29184,"hostname":"Admins-Laptop-2.hsd1.co.comcast.net","code":-32603,"message":"Error invoking RPC: Invalid transactionID: ","msg":"[Request ID: testId] Failed to successfully submit sendRawTransaction for transaction 0x01f87582012a8088ad78ebc5ac6200008303d0909413212a14deaf2775a5b3becc857806d5c719d3f2890511617de831b9e17380c001a0d3ae419f17fa82fa7e89acb782d87b5df611c545a81a0250b0bc5cc9c8f9e5f8a0102ed103f88d2a6e7d8597449e3806e5ce1ef5c2ace3f8958a21838a9bdb7099 to node {\"127.0.0.1:50211\":\"0.0.3\"}
```
This PR adds the following that was not in the logged messages:
```
to node {\"127.0.0.1:50211\":\"0.0.3\"}
```

**Related issue(s)**:

Fixes #3083 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1489,11 +1489,7 @@ export class EthImpl implements Eth {
   ): Promise<string | JsonRpcError> {
     this.logger.error(
       e,
-      `${
-        requestDetails.formattedRequestId
-      } Failed to successfully submit sendRawTransaction for transaction ${transaction} to node ${JSON.stringify(
-        this.hapiService.getMainClientInstance().network,
-      )}`,
+      `${requestDetails.formattedRequestId} Failed to successfully submit sendRawTransaction for transaction ${transaction} to node ${this.hapiService.hederaNetwork}`,
     );
     if (e instanceof JsonRpcError) {
       return e;
@@ -1935,7 +1931,7 @@ export class EthImpl implements Eth {
       this.logger.error(
         e,
         `${requestIdPrefix} Failed to successfully submit contractCallQuery to consensus-node: ${JSON.stringify(
-          this.hapiService.getMainClientInstance().network,
+          this.hapiService.hederaNetwork,
         )}`,
       );
       if (e instanceof JsonRpcError) {
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1489,7 +1489,11 @@ export class EthImpl implements Eth {
   ): Promise<string | JsonRpcError> {
     this.logger.error(
       e,
-      `${requestDetails.formattedRequestId} Failed to successfully submit sendRawTransaction for transaction ${transaction}`,
+      `${
+        requestDetails.formattedRequestId
+      } Failed to successfully submit sendRawTransaction for transaction ${transaction} to node ${JSON.stringify(
+        this.hapiService.getMainClientInstance().network,
+      )}`,
     );
     if (e instanceof JsonRpcError) {
       return e;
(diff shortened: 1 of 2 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1489,7 +1489,11 @@ export class EthImpl implements Eth {
   ): Promise<string | JsonRpcError> {
     this.logger.error(
       e,
-      `${requestDetails.formattedRequestId} Failed to successfully submit sendRawTransaction for transaction ${transaction}`,
+      `${
+        requestDetails.formattedRequestId
+      } Failed to successfully submit sendRawTransaction for transaction ${transaction} to node ${JSON.stringify(
+        this.hapiService.getMainClientInstance().network,
```
Nana-EC, line 1495: I don't think this outputs what you'd expect.
Do you have an example of the output you could put in the PR details for reviewers?
Each transaction could go to any number of noes in the network depending on the SDK details.

Nana-EC (changes requested):
Let's confirm this gives us the actual node the transaction is submitted to

packages/relay/src/lib/eth.ts
ebadiere: This produces the following which may include information we do not need:
```
to node {"127.0.0.1:50211":{"shard":{"low":0,"high":0,"unsigned":false},"realm":{"low":0,"high":0,"unsigned":false},"num":{"low\":3,"high":0,"unsigned":false},"aliasKey":null,"evmAddress":null,"_checksum":null}}
```
Updated the `hapiServer.ts` to make the `hederaNetwork` public and use that in the logs which just lists:
```
to node {\"127.0.0.1:50211\":\"0.0.3\"}
```
The description has been updated

Nana-EC: That's an improvement but it's still not what you need.
network is the network the SDK is set to point to . In this case it's a local network so all transactions will go there.
if you were to have more than one node this would likely contain multiple nodes from what transaction would be sent to.

What we need is the actual node a transaction is sent to so we can say Transaction 1 was sent to Node 2 and Transaction 3 was sent to Node 4.

Try this against previewnet or testnet.

ebadiere: Okay, so yes, the node to send a transaction to is selected by the internal round-robin process.  It looks as if the `nodeId` is available in the Hedera transaction response, however the problem we're trying to solve is identifying or capturing the `nodeId` on failed transactions to identify a potentially problematic node.  The message we're seeing in the logs is:
` Transaction execution returns a null value: transactionId=...`
That message comes from a `null` or `undefined` transactionResponse, and so the `nodeId` will not be available to log.
```
      if (!transactionResponse) {
        throw predefined.INTERNAL_ERROR(
          `${requestDetails.formattedRequestId} Transaction execution returns a null value: transactionId=${transaction.transactionId}, callerName=${callerName}, txConstructorName=${txConstructorName}`,
        );
      }
```

ebadiere:
Replaced by #3106

PR #3089, opened 2024-10-11
feat: add maxAttempts parameter to the acceptance test workflow

### LINKED ISSUE #3091: Add parameter to acceptance tests to control the maximum number of retries
### Problem

Currently, acceptance tests have a hardcoded maxAttempts count of 3. It would be helpful to be able to control the number of retries.

### Solution

We should add a maxAttempts parameter to the acceptance test pipeline which governs the number of retries an acceptance test run will attempt before marking the build as a failure.

### Alternatives

_No response_

beeradb:
Fixes #3091

This PR makes the following changes:

- Adds a `maxAttempts` parameter to the acceptance test workflow, so that we can control the number of test retry attempts.
- Updates the public acceptance workflow to set the maxAtempts parameter to 1, ensuring we only do a single acceptance test run against public networks.

### DIFF
--- a/.github/workflows/acceptance-public.yml
+++ b/.github/workflows/acceptance-public.yml
@@ -28,6 +28,7 @@ jobs:
       testfilter: release_light
       envfile: ${{ inputs.network }}Acceptance.env
       operator_id: ${{ inputs.operator_id }}
+      maxAttempts: 1
     secrets:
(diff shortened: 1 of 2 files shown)


ebadiere (approved):
LG

quiet-node (approved):

tomzhenghedera (commented):
Looks good.

PR #3095, opened 2024-10-13
feat: Moving connection reset metrics from HAPIService into MetricsService.

### LINKED ISSUE #3075: Enhance the `hapiService.ts` the connection reset metrics to show meaningful metrics in grafana
### Problem

In the logs we see behavior that indicates connections being dropped or closed.    

### Solution

To better monitor the issue we need meaningful metrics in the grafana dashboard.  There is a counter but it only displays as an aggregate line in a panel at the moment.  It may need to be broken down into transaction reset counter, error counter, and duration counter.

### Alternatives

_No response_

ebadiere:
Moving the consensus client reset counter to the metric service.

**Description**:
The old client reset counter was not used and was problematic. 

- Labels Are Not Meant for Numerical Data: In Prometheus, labels are designed for categorical data used to identify or group metrics (e.g., by host, region, or service). They are not intended for storing numerical data that you want to plot or perform calculations on.

- Inability to Perform Calculations: Since label values are strings, you cannot perform arithmetic operations on them in Prometheus queries.

- Visualization Challenges: Grafana relies on numerical data points over time to generate graphs. With the current setup, it's difficult to extract and plot the numerical values from labels.

This change replaces the labels with counters.

**Related issue(s)**:

Fixes #3075 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -102,19 +138,23 @@ export default class MetricService {
    */
   constructor(
     logger: Logger,
-    sdkClient: SDKClient,
+    hapiService: HAPIService,
```
victor-yanev, line 141: QQ: Why was this dependency changed? 

It seems that we are using it only to get the `SDKClient`:

```typescript
this.sdkClient = hapiService.getSDKClient();
```

victor-yanev (commented):
Nice job, I like that we moved those to the `MetricService` to have all grafana metrics in one place.

Why did we change these to use three separate `Counters`, instead of one `clientResetCounter` which has labels for `['transactions', 'duration', 'errors']`?

packages/relay/src/lib/services/metricService/metricService.ts
ebadiere: To me it made more sense to pass in the `HAPIService` into the `MetricService` since the `HAPIService` is responsible for resetting the client.   The client functionality as a whole is better represented in the `HAPIService` IMHO.  If it is too confusing I can set it back.

I'll answer your general question here since I cannot find a way to respond to it directly.  As stated in the `Description` , "Labels Are Not Meant for Numerical Data".  That's what the original implementation was doing.  This made visualizing the metrics in grafana challenging.  There are three conditions in which the client will be reset in the service:
1. A number of transactions processed
2. Predefined error codes being returned from the consensus node
3. The duration since the last client reset expiring.

The goal is to understand how often and why the client gets reset, in order to better understand the causes of any failed transactions.

ebadiere:
Replaced by PR #3120

PR #3104, opened 2024-10-15
test: create e2e tests for basic spending plan limit

### LINKED ISSUE #3065: [HBAR Rate Limit Redesign] Create e2e tests for basic spending plan limit
```[tasklist]
### Tasks
- [x] Create a test ensuring a user not included in a plan is treated as a BASIC user
- [x] Create a test simulating transactions that accumulate to just below the limit
- [x] Create a test simulating transactions that exceed the limit
- [x] Create a test deploying X amount of contracts for a BASIC user until limited
```

konstantinabl:
**Description**:

Adds test cases described in #3065 

The PR introduces the following changes:

* Specifically sets the total limit for the hbar limiter acceptance tests in the npm command, in order for the values to be closer to real life case, as opposed to other tests, where this is not needed
* Adds new test cases for BASIC users
* Adds a deleteAll method for ip address plans, as a helper for the new tests. This is needed because the acceptance tests are ran with the same ipAddress, and new plans are not created as expected
* Adds a timeout in tests where expected cost of tx is needed, since limiter updated expenses on the background

**Related issue(s)**:

Fixes #3065


packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -102,6 +102,36 @@ export class IPAddressHbarSpendingPlanRepository {
...
 
+  /**
+   * Deletes all IP address spending plans from the cache.
+   *
+   * @param {RequestDetails} requestDetails - The request details for logging and tracking.
+   * @returns {Promise<void>} - A promise that resolves when all IP address spending plans are deleted.
+   */
+  async deleteAll(requestDetails: RequestDetails): Promise<void> {
+    const key = this.getKey('*');
+    const keys = await this.cache.keys(key, 'deleteAll', requestDetails);
```
quiet-node, line 113: Let's try to use the `name` property of this particular method instead of a string. This way if the method name is ever updated, we won't have to worry about updating the string.

```suggestion
    const keys = await this.cache.keys(key, this.deleteAll.name, requestDetails);
```

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -102,6 +102,36 @@ export class IPAddressHbarSpendingPlanRepository {
...
+    for (const key of keys) {
+      await this.cache.delete(key, 'deleteAll', requestDetails);
```
quiet-node, line 115: Let's try to use the `name` property of this particular method instead of a string. This way if the method name is ever updated, we won't have to worry about updating the string.

```suggestion
    await this.cache.delete(key, this.deleteAll.name, requestDetails);
```

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -102,6 +102,36 @@ export class IPAddressHbarSpendingPlanRepository {
     }
   }
 
+  /**
+   * Deletes all IP address spending plans from the cache.
+   *
+   * @param {RequestDetails} requestDetails - The request details for logging and tracking.
+   * @returns {Promise<void>} - A promise that resolves when all IP address spending plans are deleted.
+   */
+  async deleteAll(requestDetails: RequestDetails): Promise<void> {
```
quiet-node, line 111: To keep this class, `IPAddressHbarSpendingPlanRepository`, and the `EthAddressHbarSpendingPlanRepository` class in sync, let's also add another deleteAll() method to the `EthAddressHbarSpendingPlanRepository` class.

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -102,6 +102,36 @@ export class IPAddressHbarSpendingPlanRepository {
...
+    this.logger.trace(`${requestDetails.formattedRequestId} Deleted all IP address spending plans`);
+  }
+
+  /**
+   * Gets all IP address spending plans from the cache.
+   *
+   * @param {RequestDetails} requestDetails - The request details for logging and tracking.
+   * @returns {Promise<IPAddressHbarSpendingPlan[]>} - A promise that resolves with an array of IP address spending plans.
+   */
+  async getAllPlans(requestDetails: RequestDetails): Promise<IPAddressHbarSpendingPlan[]> {
```
quiet-node, line 126: To keep this class, `IPAddressHbarSpendingPlanRepository`, and the `EthAddressHbarSpendingPlanRepository` class in sync, let's also add another getAllPlans() method to the `EthAddressHbarSpendingPlanRepository` class.

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -102,6 +102,36 @@ export class IPAddressHbarSpendingPlanRepository {
...
+    const pattern = `${this.collectionKey}:*`;
+    const keys = await this.cache.keys(pattern, 'getAllPlans', requestDetails);
```
quiet-node, line 128: Let's try to use the `name` property of this particular method instead of a string. This way if the method name is ever updated, we won't have to worry about updating the string.

```suggestion
    const keys = await this.cache.keys(pattern, this.getAllPlans.name, requestDetails)
```

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -102,6 +102,36 @@ export class IPAddressHbarSpendingPlanRepository {
...
+    const plans = await Promise.all(
+      keys.map((key) => this.cache.getAsync<IIPAddressHbarSpendingPlan>(key, 'getAllPlans', requestDetails)),
```
quiet-node, line 130: Let's try to use the `name` property of this particular method instead of a string. This way if the method name is ever updated, we won't have to worry about updating the string.

```suggestion
    keys.map((key) => this.cache.getAsync<IIPAddressHbarSpendingPlan>(key, this.getAllPlans.name, requestDetails)),
```

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
konstantinabl: done

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
konstantinabl: done, good point

package.json
```diff
@@ -47,7 +47,7 @@
     "acceptancetest:api_batch3": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@api-batch-3' --exit",
     "acceptancetest:erc20": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@erc20' --exit",
     "acceptancetest:ratelimiter": "nyc ts-mocha packages/ws-server/tests/acceptance/index.spec.ts -g '@web-socket-ratelimiter' --exit && ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@ratelimiter' --exit",
-    "acceptancetest:hbarlimiter": "HBAR_RATE_LIMIT_TINYBAR=3000000000 nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter' --exit",
+    "acceptancetest:hbarlimiter": "HBAR_RATE_LIMIT_TINYBAR=4500000000 HBAR_RATE_LIMIT_BASIC=3000000000 nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter' --exit",
```
konstantinabl, line 50: @quiet-node honestly, my idea was to set this to 10 hbar, cause its closer to what the real limits would be. And we need this for the test that checks if a BASIC user will be limited when deploying large contracts. But lets address this in another PR. This would work for now, i would just assume that the test would be unecessarily long until it exhaust 30 hbars, also not very realistic. My idea of the acceptance tests is that they need to resemble the real life experience of a user

PR #3120, opened 2024-10-17
feat: Moved sdkClient reset metrics into the metricService

### LINKED ISSUE #3075: Enhance the `hapiService.ts` the connection reset metrics to show meaningful metrics in grafana

ebadiere:
**Description**:
Moved sdkClient reset metrics into the metricService and replaced labels with prometheus counters.

**Related issue(s)**:

Fixes #3075 

**Notes for reviewer**:
The old client reset counter was not used and was problematic.

Labels Are Not Meant for Numerical Data: In Prometheus, labels are designed for categorical data used to identify or group metrics (e.g., by host, region, or service). They are not intended for storing numerical data that you want to plot or perform calculations on.

Inability to Perform Calculations: Since label values are strings, you cannot perform arithmetic operations on them in Prometheus queries.

Visualization Challenges: Grafana relies on numerical data points over time to generate graphs. With the current setup, it's difficult to extract and plot the numerical values from labels.

This change replaces the labels with counters.

**Checklist**

### DIFF
--- a/packages/relay/src/lib/services/hapiService/hapiService.ts
+++ b/packages/relay/src/lib/services/hapiService/hapiService.ts
@@ -25,11 +25,15 @@ import findConfig from 'find-config';
 import constants from '../../constants';
 import { Utils } from './../../../utils';
 import HbarLimit from '../../hbarlimiter';
-import { Counter, Registry } from 'prom-client';
 import { SDKClient } from '../../clients/sdkClient';
 import { CacheService } from '../cacheService/cacheService';
 import { AccountId, Client, PrivateKey } from '@hashgraph/sdk';
 import fs from 'fs';
+import {
+  IExecuteConsenusClientResetDurationPayload,
+  IExecuteConsenusClientResetErrorPayload,
+  IExecuteConsenusClientResetTransactionPayload,
+} from '../../types/events';
 
 export default class HAPIService {
   private transactionCount: number;
@@ -76,11 +80,6 @@ export default class HAPIService {
    */
   private readonly eventEmitter: EventEmitter;
 
-  /**
-   * @private
-   */
-  private readonly register: Registry;
-  private clientResetCounter: Counter;
   private readonly cacheService: CacheService;
 
   /**
@@ -92,18 +91,11 @@ export default class HAPIService {
    * Constructs an instance of the class, initializes configuration settings, and sets up various services.
    *
    * @param {Logger} logger - The logger instance used for logging.
-   * @param {Registry} register - The registry instance for metrics and other services.
    * @param {HbarLimit} hbarLimiter - The Hbar rate limiter instance.
    * @param {CacheService} cacheService - The cache service instance.
    * @param {EventEmitter} eventEmitter - The event emitter instance used for emitting events.
    */
-  constructor(
-    logger: Logger,
-    register: Registry,
-    hbarLimiter: HbarLimit,
-    cacheService: CacheService,
-    eventEmitter: EventEmitter,
-  ) {
+  constructor(logger: Logger, hbarLimiter: HbarLimit, cacheService: CacheService, eventEmitter: EventEmitter) {
     dotenv.config({ path: findConfig('.env') || '' });
     if (fs.existsSync(findConfig('.env') || '')) {
       this.config = dotenv.parse(fs.readFileSync(findConfig('.env') || ''));
@@ -137,16 +129,6 @@ export default class HAPIService {
       this.isReinitEnabled = false;
     }
     this.shouldReset = false;
-
-    this.register = register;
-    const metricCounterName = 'rpc_relay_client_service';
-    this.register.removeSingleMetric(metricCounterName);
-    this.clientResetCounter = new Counter({
-      name: metricCounterName,
-      help: 'Relay Client Service',
-      registers: [register],
-      labelNames: ['transactions', 'duration', 'errors'],
-    });
   }
 
   /**
@@ -159,6 +141,9 @@ export default class HAPIService {
 
     this.transactionCount--;
     if (this.transactionCount <= 0) {
+      this.eventEmitter.emit(constants.EVENTS.RESET_CLIENT_TRANSACTIONS, {
+        transactionCount: this.transactionCount,
+      } as IExecuteConsenusClientResetTransactionPayload);
       this.shouldReset = true;
     }
   }
@@ -172,6 +157,9 @@ export default class HAPIService {
     }
 
     if (this.errorCodes.includes(statusCode)) {
+      this.eventEmitter.emit(constants.EVENTS.RESET_CLIENT_ERRORS, {
+        errorCodes: this.errorCodes,
+      } as IExecuteConsenusClientResetErrorPayload);
       this.shouldReset = true;
     }
   }
@@ -182,6 +170,9 @@ export default class HAPIService {
     }
 
     if (this.resetDuration < Date.now()) {
+      this.eventEmitter.emit(constants.EVENTS.RESET_CLIENT_DURATION, {
+        duration: this.resetDuration,
+      } as IExecuteConsenusClientResetDurationPayload);
       this.shouldReset = true;
     }
   }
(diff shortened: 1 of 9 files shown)


packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -92,18 +91,11 @@ export default class HAPIService {
...
    * @param {EventEmitter} eventEmitter - The event emitter instance used for emitting events.
    */
-  constructor(
-    logger: Logger,
-    register: Registry,
-    hbarLimiter: HbarLimit,
-    cacheService: CacheService,
-    eventEmitter: EventEmitter,
-  ) {
+  constructor(logger: Logger, hbarLimiter: HbarLimit, cacheService: CacheService, eventEmitter: EventEmitter) {
```
konstantinabl, line 98: A nit, but did you change the formatting on purpose? @ebadiere

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -48,6 +52,14 @@ export default class MetricService {
    */
   private readonly sdkClient: SDKClient;
 
+  /**
+   * HAPIService for managing the sdkClient.
+   * @type {HAPIService}
+   * @readonly
+   * @private
+   */
+  private readonly hapiService: HAPIService;
```
konstantinabl, line 61: This class property is not actually used so you can remove it

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -102,19 +138,23 @@ export default class MetricService {
...
-    sdkClient: SDKClient,
+    hapiService: HAPIService,
     mirrorNodeClient: MirrorNodeClient,
     hbarLimiter: HbarLimit,
     register: Registry,
     eventEmitter: EventEmitter,
   ) {
     this.logger = logger;
-    this.sdkClient = sdkClient;
+    this.hapiService = hapiService;
```
konstantinabl, line 148: As we remove the class property this line should be removed we only need the hapiService instance to get the sdk client in the line below

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -102,19 +138,23 @@ export default class MetricService {
...
+    this.sdkClient = hapiService.getSDKClient();
```
konstantinabl, line 149: But do we really need this, not sure we need the metric service to depend on the sdkClient. The sdkClient is used here to get details from the transaction record in the captureTransactionMetrics. However, arent they passed when emitting the event in the sdkClient?

konstantinabl (changes requested):
Few comments

packages/relay/src/lib/services/hapiService/hapiService.ts
quiet-node: It looks like he removed the `register: Registry` argument from the args list, so Prettier reformatted it into a single line. It's just a Prettier thing I've seen that a lot.

packages/relay/src/lib/services/metricService/metricService.ts
quiet-node: This is a really good point! Perhaps we can passed the `transactionRecordMetrics` in when emitting the event along with other arguments. But I guess it's gonna be out of the scope for this PR we can create a ticket for it. Great point!

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -220,6 +275,20 @@ export default class MetricService {
     this.captureMetrics(executionMode, txConstructorName, status, cost, gasUsed, callerName, interactingEntity);
   };
 
+  public incrementConsensusClientDurationResets = ({ duration }: IExecuteConsenusClientResetDurationPayload): void => {
```
quiet-node, line 278: let's try to include JSDocs for these methods

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -220,6 +275,20 @@ export default class MetricService {
...
+    this.consensusClientResetsDuration.inc(1);
+  };
+
+  public incrementConsensusClientErrorResets = ({ errorCodes }: IExecuteConsenusClientResetErrorPayload): void => {
```
quiet-node, line 282: let's try to include JSDocs for these methods

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -220,6 +275,20 @@ export default class MetricService {
...
+    this.consensusClientResetsError.inc(1);
+  };
+
+  public incrementConsensusClientTransactionResets = ({
```
quiet-node, line 286: let's try to include JSDocs for these methods

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -252,6 +321,54 @@ export default class MetricService {
     });
   }
 
+  /**
+   * Initialize consensus node reset counter for duration
+   * @param {Registry} register
+   * @returns {Counter} Consensus node client reset metric
+   */
+  private initConsensusClientResetDurationMetric(register: Registry): Counter {
```
quiet-node, line 329: The methods `initConsensusClientResetDurationMetric()`, `initConsensusClientResetErrorCodeMetric()`, and `initConsensusClientResetTransactionMetric()` are essentially the same. It might be better to group them into a more general method to avoid code repetition. We could place this general method in `./packages/relay/src/utils.ts`, making it reusable in other areas as well.

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -123,6 +163,21 @@ export default class MetricService {
     this.eventEmitter.on(constants.EVENTS.EXECUTE_QUERY, (args: IExecuteQueryEventPayload) => {
       this.addExpenseAndCaptureMetrics(args);
     });
+
+    this.eventEmitter.on(constants.EVENTS.RESET_CLIENT_DURATION, (args: IExecuteConsenusClientResetDurationPayload) => {
+      this.incrementConsensusClientDurationResets(args);
```
quiet-node, line 168: Also let's make sure to add tests to cover these execution

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -123,6 +163,21 @@ export default class MetricService {
...
+    });
+
+    this.eventEmitter.on(constants.EVENTS.RESET_CLIENT_ERRORS, (args: IExecuteConsenusClientResetErrorPayload) => {
+      this.incrementConsensusClientErrorResets(args);
```
quiet-node, line 172: Also let's make sure to add tests to cover these execution

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -123,6 +163,21 @@ export default class MetricService {
...
+    });
+
+    this.eventEmitter.on(
+      constants.EVENTS.RESET_CLIENT_TRANSACTIONS,
+      (args: IExecuteConsenusClientResetTransactionPayload) => {
+        this.incrementConsensusClientTransactionResets(args);
```
quiet-node, line 178: Also let's make sure to add tests to cover these execution

quiet-node (changes requested):

packages/relay/src/lib/services/metricService/metricService.ts
ebadiere: Done.

packages/relay/src/lib/services/metricService/metricService.ts
ebadiere: Done.

packages/relay/src/lib/services/metricService/metricService.ts
ebadiere: Done.

packages/relay/src/lib/services/metricService/metricService.ts
ebadiere: Done

packages/relay/src/lib/services/metricService/metricService.ts
ebadiere: Done

packages/relay/src/lib/services/metricService/metricService.ts
ebadiere: Good idea! Done.

PR #3140, opened 2024-10-22
fix: Protect preconfigured spending plans from being deleted automatically in `LRUCache`

### LINKED ISSUE #3096: [HBAR Rate Limit Redesign] Implement custom LRU cache with protected keys

victor-yanev:
**Description**:

This PR introduces changes to ensure that preconfigured HBAR spending plans are protected from automatic deletion by the `LRUCache` on reaching the maximum cache size. 

Includes the following updates: 
- Adds a separate `reservedCache: LRUCache<string, any>` instance inside the `LocalLRUCache` class which is used to store a cache of reserved keys without a default TTL set
- Modified `CacheService` initialization to include reserved keys for protecting preconfigured spending plans.

**Related issue(s)**:

Fixes #3096

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -72,14 +62,44 @@ export class HbarSpendingPlanConfigService {
...
+            }),
+          ];
+        }),
+      );
+    } catch (error: any) {
+      logger.error(`Failed to get pre-configured spending plan keys: ${error.message}`);
+      return new Set<string>();
+    }
+  }
+
```
ebadiere, line 94: For readability and efficiency the collection keys can be destructed to reduce redundant property access.  Then the callback within the `flatMap` can be streamlined to reduce complexity.  The use of concise arrow functions and spread operators makes the code easier to read and maintain.  I replaced the above with the snippet below and the tests passed.  Just a consideration:
```
  try {
    const { collectionKey: hbarKey } = HbarSpendingPlanRepository;
    const { collectionKey: ethKey } = EthAddressHbarSpendingPlanRepository;
    const { collectionKey: ipKey } = IPAddressHbarSpendingPlanRepository;

    const keys = this.loadSpendingPlansConfig(logger).flatMap(({ id, ethAddresses = [], ipAddresses = [] }) => [
      `${hbarKey}:${id}`,
      `${hbarKey}:${id}:amountSpent`,
      `${hbarKey}:${id}:spendingHistory`,
      ...ethAddresses.map((ethAddress) => `${ethKey}:${ethAddress.trim().toLowerCase()}`),
      ...ipAddresses.map((ipAddress) => `${ipKey}:${ipAddress}`),
    ]);

    return new Set<string>(keys);
  } catch (error: any) {
    logger.error(`Failed to get pre-configured spending plan keys: ${error.message}`);
    return new Set<string>();
  }
  ```

**Review by ebadiere — DISMISSED** [review]
Looks good.  Just one small suggestion.

File: packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts (hunk 1 above)
Comment (victor-yanev) [reply packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts:94]: Nice suggestion, applied.

**Review by natanasow — DISMISSED** [review]

File: packages/relay/tests/lib/clients/localLRUCache.spec.ts [hunk 2]
```diff
@@ -134,6 +134,37 @@ describe('LocalLRUCache Test Suite', async function () {
...
+      const key2 = await customLocalLRUCache.get('key2', callingMethod, requestDetails);
+      const key3 = await customLocalLRUCache.get('key3', callingMethod, requestDetails);
+      const key4 = await customLocalLRUCache.get('key4', callingMethod, requestDetails);
+      const key5 = await customLocalLRUCache.get('key5', callingMethod, requestDetails);
+      // expect cache to have capped at max size
+      expect(key1).to.be.equal(keyValuePairs.key1); // key1 should not have been evicted, as it is a reserved key
+      expect(key2).to.be.equal(keyValuePairs.key2); // key2 should not have been evicted, as it is a reserved key
+      expect(key3).to.be.null; // key3 should have been evicted
+      expect(key4).to.be.equal(keyValuePairs.key4);
+      expect(key5).to.be.equal(keyValuePairs.key5);
```
quiet-node, line 162: night be missing something but I thought the max reserved key size was 2, so it should only hold `key1` and `key2`, right? How did `key3` get evicted while `key4` and `key5` were set?

quiet-node (dismissed):
Nice work! A quick question above

packages/relay/tests/lib/clients/localLRUCache.spec.ts
victor-yanev: `key1` and `key2` are "reserved" and, thus, they were not evicted, `key3` is not "reserved", so it got evicted as it's the least recently used key in the cache (if we don't consider the reserved ones).

natanasow (approved):

quiet-node (approved):
LGTM

Issue #2850, opened 2024-08-19
[HBAR Rate Limit Redesign] Provide configurable rate limiting based on original caller addresses, IP addresses -- whitelist system


quiet-node:
## Problem
Currently, there's a temporary whitelist system in place that allows the Relay to recognize a list of special, whitelisted projects (EVM addresses) and bypass any HBAR rate limits. This logic was introduced in v0.54.1 as a hotfix, with the whitelist being configured via the `HBAR_RATE_LIMIT_WHITELIST` environment variable. However, this logic currently only supports EVM addresses, while the desired feature is to extend this functionality to IP addresses as well.

## Solution
Improved current logic for better performance -> extend the setup to accept IP addresses
```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/2894
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/2866
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/2869
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/2865
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/2901
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/2867
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/2868
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/2888
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/2895
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/2994
```

PR #3141, opened 2024-10-22
fix: do not link IP addresses of general users

victor-yanev:
Related issue: #2850 
**Description**:

This PR addresses the following concern:
> Yeah I believe the logic should securely handle checks on both EVM addresses and IP addresses independently, without either depending on the other.
> I thought of a scenario where a random application, like an NFT marketplace or a DEX, submits user requests from the same centralized server (i.e., the same IP address). Users connect to the platform using different wallet accounts (i.e., different EVM addresses), but all requests are submitted to the network under a single IP address. If we only validate based on the IP address, traffic becomes unfair since requests from different users would be dependent on one another. A user might reach the HBAR limit even if they have not made prior calls, simply because the spending plan associated with that IP address has exceeded its limit.

For EXTENDED and PRIVILEGED:
- this is not possible to happen due to us clearing any previous links of the ETH and IP addresses specified in the configuration file

### DIFF
--- a/packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
+++ b/packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
@@ -103,21 +103,6 @@ export class IPAddressHbarSpendingPlanRepository {
   }
 
   /**
-   * Deletes all spending plans associated with IP address from cache.
-   *
-   * @param {RequestDetails} requestDetails - The request details for logging and tracking.
-   * @returns {Promise<void>} - A promise that resolves when all IP address spending plans are deleted.
-   */
-  async deleteAll(requestDetails: RequestDetails): Promise<void> {
-    const key = this.getKey('*');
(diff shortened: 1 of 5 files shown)


quiet-node:
Might need to update the associated and related UTs as well

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -102,21 +102,6 @@ export class IPAddressHbarSpendingPlanRepository {
     }
   }
 
-  /**
```
quiet-node, line 105: Nice could you also please remove the deleteAll() in the `EthAddressHbarSpendingPlanRepository` class also? 
It was added as an attempt to synchronize the two classes.

quiet-node: resolved

quiet-node (approved):
LGTM

ebadiere (approved):
LG.

PR #3147, opened 2024-10-23
test: added e2e test for EXTENDED and PRIVILEGED tier users

### LINKED ISSUE #3066: [HBAR Rate Limit Redesign] Create e2e tests for extended spending plan limit
```[tasklist]
### Tasks
- [ ] Create a test with EXTENDED user plan with both IP and ETH addresses and accumulate transactions below the limit
- [ ] Create a test with EXTENDED user plan with both IP and ETH addresses and accumulate transactions exceeding the limit
- [ ] Create a test with EXTEDNED user plan deploying X amount of contracts
```

### LINKED ISSUE #3067: [HBAR Rate Limit Redesign] Create e2e tests for privileged spending plan limit

quiet-node:
**Description**:
This PR added e2e tests to cover usage for both EXTENDED and PRIVILEGED tier users. 

**Related issue(s)**:

Fixes #3066
Fixes #3067

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+                id: 'c758c095-342c-4607-9db5-867d7e90ab9d',
+                name: 'PRIVILEGED_ALPHA',
+                ethAddresses: ['0x7d102fe71af42790fe31b126c1f49766376ca2b5'],
+                ipAddresses: ['127.0.0.1'],
+                subscriptionTier: 'PRIVILEGED',
+              },
+              PRIVILEGED_BETA: {
+                id: 'a68488b0-6f7d-44a0-87c1-774ad64615f2',
+                name: 'PRIVILEGED_BETA',
+                ethAddresses: ['0x40183ec818c1826114767391989ff2eaebc2b91e'],
```
victor-yanev, line 535: We should have a test that verifies the same PRIVILEGED plan can be shared by multiple addresses

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+            };
+
+            const expectedNonBasicEvmAddresses = {
+              PRIVILEGED_ALPHA: ['0x7d102fe71af42790fe31b126c1f49766376ca2b5'],
+              PRIVILEGED_BETA: ['0x40183ec818c1826114767391989ff2eaebc2b91e'],
+            };
+
+            const expectedPrivilegedIpAddresses = {
+              PRIVILEGED_ALPHA: ['127.0.0.1'],
+            };
```
victor-yanev, line 547: We don't need these constants we can directly loop through the keys of `expectedNonBasicPlans` and use `expectedNonBasicPlans[key].evmAddresses` and `expectedNonBasicPlans[key].ipAddresses`

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+              { ethAddress: aliasAccount.address, planId: hbarSpendingPlan.id },
+              requestDetails,
+              mockTTL,
+            );
+
+            return { aliasAccount, hbarSpendingPlan };
+          };
+
+          describe('Pre-configured spending plans', async () => {
+            const expectedNonBasicPlans = {
```
victor-yanev, line 524: Please, instead of duplicating the content of `testSpendingPlansConfig.json` here, load the file and parse it into `expectedNonBasicPlans` dynamically. 

In this way we can only modify the JSON file to extend the tests with more plans and we don't need to modify the code of the tests.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
             }
           });
         });
+
+        describe('NON-BASIC Tiers', () => {
```
victor-yanev, line 503:
```suggestion
        describe('Preconfigured Tiers', () => {
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+            it('Should successfully populate all pre-configured spending plans', async () => {
+              Object.values(expectedNonBasicPlans).forEach(async (plan) => {
+                const hbarSpendingPlan = await hbarSpendingPlanRepository.findByIdWithDetails(plan.id, requestDetails);
+                expect(hbarSpendingPlan.id).to.eq(plan.id);
+                expect(hbarSpendingPlan.active).to.be.true;
+                expect(hbarSpendingPlan.subscriptionTier).to.eq(plan.subscriptionTier);
+              });
+            });
+
+            it('Should successfully link all pre-configured spending plans to correct NON-BASIC EVM addresses', async () => {
```
victor-yanev, line 558:
```suggestion
            it('Should successfully link all EVM addresses in the pre-configured spending plans', async () => {
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+            await ethAddressSpendingPlanRepository.save(
+              { ethAddress: aliasAccount.address, planId: hbarSpendingPlan.id },
+              requestDetails,
+              mockTTL,
+            );
+
+            return { aliasAccount, hbarSpendingPlan };
+          };
+
+          describe('Pre-configured spending plans', async () => {
```
victor-yanev, line 523:
```suggestion
          describe('given a valid JSON file with pre-configured spending plans', async () => {
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+              );
+
+              expect(plan.ethAddress).to.eq(priviledgedEvmAccount.address);
+              expect(plan.planId).to.eq(privilegedHbarSpendingPlan.id);
+
+              const spendingPlan = await hbarSpendingPlanRepository.findByIdWithDetails(plan.planId, requestDetails);
+              expect(spendingPlan.active).to.be.true;
+              expect(spendingPlan.amountSpent).to.eq(0);
+              expect(spendingPlan.subscriptionTier).to.eq(SubscriptionTier.PRIVILEGED);
+            });
```
victor-yanev, line 613: QQ: What's the point of this test? We are not doing anything here other than verifying the test preparations.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+
+            it('Should increase the amount spent by the privileged spending plan', async () => {
```
victor-yanev, line 615:
```suggestion
            it('Should increase the amount spent of the spending plan by the transaction cost', async () => {
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+
+              const spendingPlan = await hbarSpendingPlanRepository.findByIdWithDetails(
+                privilegedHbarSpendingPlan.id,
+                requestDetails,
+              );
+
+              const expectedCost = await getExpectedCostOfLastLargeTx(contract.deploymentTransaction()!.data);
+              const amountSpent = spendingPlan.amountSpent;
+
+              expect(amountSpent).to.be.approximately(expectedCost, 0.009 * expectedCost);
```
victor-yanev, line 629:
```suggestion
              const tolerance = 0.009;
              expect(amountSpent).to.be.approximately(expectedCost, tolerance * expectedCost);
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+            });
+
+            it('Should eventually exhaust the hbar limit for PRIVILEDGED user', async () => {
+              const fileChunkSize = Number(ConfigService.get('FILE_APPEND_CHUNK_SIZE')) || 5120;
```
victor-yanev, line 633: The default value (`5120`) should be specified inside `globalConfig.ts` so that it's automatically picked up by the `ConfigService`, we should not specify defaults here or anywhere else.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+                exchangeRateInCents,
+              );
+
+              try {
+                for (let i = 0; i < 50; i++) {
+                  await deployContract(largeContractJson, priviledgedEvmAccount.wallet);
+                }
+                expect.fail(`Expected an error but nothing was thrown`);
+              } catch (e: any) {
+                expect(e.message).to.contain(predefined.HBAR_RATE_LIMIT_EXCEEDED.message);
```
victor-yanev, line 656: we should verify here that `i` is greater than the `<expectedAmountOfLargeContractsWhichCanBeSubmittedByPrivilegedTier>`

packages/server/tests/previewnetAcceptance.env
```diff
@@ -20,6 +20,7 @@ SERVER_REQUEST_TIMEOUT_MS=60000
 MEMWATCH_ENABLED=false
 WRITE_SNAPSHOT_ON_MEMORY_LEAK=false
 TEST_INITIAL_ACCOUNT_STARTING_BALANCE=500
-HBAR_RATE_LIMIT_TINYBAR=5000000000# 50 HBARs
-HBAR_RATE_LIMIT_DURATION=80000# 80 seconds
+HBAR_RATE_LIMIT_TINYBAR=9000000000# 90 HBARs
+HBAR_RATE_LIMIT_DURATION=180000# 180 seconds
```
victor-yanev, line 24: Wouldn't the 180 seconds slow down the tests a lot? 🤔

victor-yanev (changes requested):

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: This is a very good point! Updated!

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: Good point! Removed and reused contents from the JSON file.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: Good point! Removed and reused contents from the JSON file.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: hmm good point added more addresses to the config file

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: removed and included this test in `Should successfully populate all pre-configured spending plans`

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +499,186 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+                    evmAddress,
+                    requestDetails,
+                  );
+                  expect(associatedPlanByEVMAddress.planId).to.eq(expectedNonBasicPlans[key].id);
+                  expect(associatedPlanByEVMAddress.ethAddress).to.eq(evmAddress);
+                });
+              });
+            });
+
+            it('Should successfully link all pre-configured spending plans to correct NON-BASIC IP addresses', async () => {
```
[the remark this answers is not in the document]
quiet-node: removed and included this test in `Should successfully populate all pre-configured spending plans`

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: This test ensures that newly created accounts are successfully loaded into the cache. However, this logic can be moved into `createAliasForNonBasicPlans` to ensure the plans are properly populated. Updated.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: The default value of this var in globalConfig.ts was set to `null`, actually. 
```typescript
FILE_APPEND_CHUNK_SIZE: {
      envName: 'FILE_APPEND_CHUNK_SIZE',
      type: 'number',
      required: false,
      defaultValue: null,
    },
```

> we should not specify defaults here or anywhere else
I agree but gotta follow what it was implemented for now.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: Are you suggesting we add the `expectedAmountOfLargeContractsWhichCanBeSubmittedByPrivilegedTier` constant?

One concern is that the number of large contract deployments depends on the `HBAR_RATE_LIMIT_PRIVILEGED`, which is currently set to a lower value (50 HBAR) for testing purposes. Therefore, the value of `expectedAmountOfLargeContractsWhichCanBeSubmittedByPrivilegedTier` might not be accurate in this context.

packages/server/tests/previewnetAcceptance.env
quiet-node: I don't believe so. This variable only affects the reset timestamp, and we’re not actually waiting on it during the tests. It shouldn’t impact the tests duration.

package.json
```diff
@@ -47,7 +47,7 @@
     "acceptancetest:api_batch3": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@api-batch-3' --exit",
     "acceptancetest:erc20": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@erc20' --exit",
     "acceptancetest:ratelimiter": "nyc ts-mocha packages/ws-server/tests/acceptance/index.spec.ts -g '@web-socket-ratelimiter' --exit && ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@ratelimiter' --exit",
-    "acceptancetest:hbarlimiter": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter' --exit",
+    "acceptancetest:hbarlimiter": "HBAR_RATE_LIMIT_TINYBAR=5000000000 HBAR_RATE_LIMIT_BASIC=1000000000 HBAR_RATE_LIMIT_EXTENDED=1500000000 HBAR_RATE_LIMIT_PRIVILEGED=2000000000 nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter' --exit",
```
quiet-node, line 50: Note: Override spending limit environment variables in the `acceptancetest:hbarlimiter` script to set lower HBAR limits at boot time. The HbarLimiter test suite includes several tests that exhaust the total HBAR limit for multiple plans, and using high spending limits can generate unnecessary traffic on the local node and extend the test duration. Overriding with lower limit values improves performance and reduces test execution time.

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -209,7 +209,11 @@ export class HbarLimitService implements IHbarLimitService {
         exceedsLimit ? 'has' : 'has NOT'
       } exceeded HBAR rate limit threshold: user=${user}, amountSpent=${
         spendingPlan.amountSpent
-      }, estimatedTxFee=${estimatedTxFee}, spendingLimit=${spendingLimit}, txConstructorName=${txConstructorName}, mode=${mode}, methodName=${methodName}`,
+      }, estimatedTxFee=${estimatedTxFee}, spendingLimit=${spendingLimit}, spandingPlanId=${
```
quiet-node, line 212: A log improvement for easier debugging

packages/server/tests/acceptance/hbarLimiter.spec.ts
victor-yanev: Most likely this was missed in the env refactor PR which adds the ConfigService, 

Let's open a new issue to fix this, cause it was most likely missed not only for this one but for some other variables as well, and `globalConfig.ts` should be the single source of truth for default env values.

packages/server/tests/acceptance/hbarLimiter.spec.ts
victor-yanev: No, more like a dynamically calculated value based on the estimated tx fee and `HBAR_RATE_LIMIT_PRIVILEGED`:

```typescript
const largeContractDeployCost = estimateFileTransactionsFee(
  contract.deploymentTransaction()!.data.length,
  fileChunkSize,
  exchangeRateInCents
);
const expectedAmountOfDeploys = Math.floor(HBAR_RATE_LIMIT_PRIVILEGED / largeContractDeployCost);

let i = 0;
try {
  for (i = 0; i < expectedAmountOfDeploys; i++) {
    ...
  }
} catch (e: any) {
  expect(e.message).to.contain(predefined.HBAR_RATE_LIMIT_EXCEEDED.message);
  expect(i).to.be.eq(expectedAmountOfDeploys);
  ...
}
```

victor-yanev: It would be nice to have this for other tests as well, because we want to confirm not only that at some point they reach the limit but also after how many deploys they have reached it (this will verify that the tier limit we have set is being met)

packages/server/tests/localAcceptance.env
```diff
@@ -30,3 +30,4 @@ WRITE_SNAPSHOT_ON_MEMORY_LEAK=false
 HBAR_RATE_LIMIT_TINYBAR=5000000000# 50 HBARs
 HBAR_RATE_LIMIT_DURATION=80000# 80 seconds
 HBAR_RATE_LIMIT_BASIC=4000000000# 40 HBARs
+HBAR_SPENDING_PLANS_CONFIG_FILE=./packages/server/tests/testSpendingPlansConfig.json
```
victor-yanev, line 33: **nit:** `findConfig` will do it's job and we don't need to specify the full path here:
```suggestion
HBAR_SPENDING_PLANS_CONFIG_FILE=testSpendingPlansConfig.json
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -503,6 +501,160 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+          });
+
+          describe('EXTENDED Tier', () => {
+            reusableTestsForNonBasicTiers(SubscriptionTier.EXTENDED, maxExtendedSpendingLimit);
+          });
+
+          describe('PRIVILEGED Tier', () => {
+            reusableTestsForNonBasicTiers(SubscriptionTier.PRIVILEGED, maxPrivilegedSpendingLimit);
+          });
+        });
```
victor-yanev, line 657: Great job, I really like how you reused the tests for both tiers.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -194,10 +198,6 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
       };
 
       before(async function () {
-        // Restart the relay to reset the limits
```
konstantinabl, line 197: why is this deleted?

ebadiere (commented):
These are all good comments, but I think new issues should be opened to address them to move forward.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: This is a very good point thanks for the suggestion! Updated the tests.

packages/server/tests/localAcceptance.env
quiet-node: synced offline and findout findConfig returns undefined with fileName and works fine with filePath.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: There’s no longer a need to restart the Local Relay, as it’s now a standalone CI. When the CI starts, the Local Relay is already a fresh instance.

Also, `await cacheService.clear(requestDetails);` was removed because the original cache is needed for the preconfigured plans test. Additionally, IP addresses are no longer automatically linked to BASIC plans, so clearing the cache is no longer necessary.

victor-yanev (approved):

packages/server/tests/acceptance/hbarLimiter.spec.ts
konstantinabl: perfect :)

ebadiere (approved):
LG

PR #3150, opened 2024-10-23
chore: add active users counter per rate limiter duration

### LINKED ISSUE #3109: [HBAR Rate Limit Redesign] Add a counter to expose DAU metrics.
### Problem

Currently, there is no metrics to track how many DAUs interact with the Relay on daily basis. 

### Solution

- Add a counter (Prometheus counter) to track unique ETH addresses, and decide how often to reset it (possibly based on `HBAR_RATE_LIMIT_DURATION`).
- Once the counter is exposed, add a visual representation of it in the dashboard.
- Two possible metrics:
    - DAUs (Daily Active Users) across all APIs.
    - DAUs for `eth_sendRawTransaction()`. This could potentially be added alongside the HbarLimiter service when a new HbarSpendingPlan is created.

### Alternatives

_No response_
```[tasklist]
### Tasks
- [ ] DAUs (Daily Active Users) across all APIs.
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/3151
```

### LINKED ISSUE #3151: Implement active users counter for eth_sendRawTransaction().

nadezhdapopovaa:
**Description**:
Add counter for active users based on the rate limiter duration period. 

**Related issue(s)**:

Fixes #3151 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -227,7 +227,7 @@ export class EthImpl implements Eth {
   private readonly chain: string;
 
   /**
-   * The ethExecutionsCounter used to track the number of active contract execution requests.
+   * The ethExecutionsCounter used to track the number of daily active users and active contract execution requests.
    * @private
    */
   private readonly ethExecutionsCounter: Counter;
@@ -272,7 +272,11 @@ export class EthImpl implements Eth {
     this.cacheService = cacheService;
     this.mirrorNodeClient = mirrorNodeClient;
     this.precheck = new Precheck(mirrorNodeClient, logger, chain);
-    this.ethExecutionsCounter = this.initEthExecutionCounter(registry);
+    this.ethExecutionsCounter = this.initCounter(
+      'rpc_relay_eth_executions',
+      ['method', 'function', 'from', 'to'],
+      registry,
+    );
     this.common = new CommonService(mirrorNodeClient, logger, cacheService);
     this.debugServiceImpl = new DebugService(mirrorNodeClient, logger, this.common);
     this.filterServiceImpl = new FilterService(mirrorNodeClient, logger, cacheService, this.common);
@@ -283,13 +287,12 @@ export class EthImpl implements Eth {
     return !CommonService.blockTagIsLatestOrPendingStrict(tag) && !CommonService.isDevMode;
   }
 
-  private initEthExecutionCounter(register: Registry): Counter {
-    const metricCounterName = 'rpc_relay_eth_executions';
+  private initCounter(metricCounterName: string, labelNames: string[], register: Registry): Counter {
     register.removeSingleMetric(metricCounterName);
     return new Counter({
       name: metricCounterName,
       help: `Relay ${metricCounterName} function`,
-      labelNames: ['method', 'function'],
+      labelNames: labelNames,
       registers: [register],
     });
   }
@@ -579,7 +582,12 @@ export class EthImpl implements Eth {
 
     if (callDataSize >= constants.FUNCTION_SELECTOR_CHAR_LENGTH) {
       this.ethExecutionsCounter
-        .labels(EthImpl.ethEstimateGas, callData!.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH))
+        .labels(
+          EthImpl.ethEstimateGas,
+          callData!.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH),
+          transaction.from || '',
+          transaction.to || '',
+        )
         .inc();
     }
 
@@ -1643,17 +1651,22 @@ export class EthImpl implements Eth {
    */
   async sendRawTransaction(transaction: string, requestDetails: RequestDetails): Promise<string | JsonRpcError> {
     const requestIdPrefix = requestDetails.formattedRequestId;
-    if (transaction?.length >= constants.FUNCTION_SELECTOR_CHAR_LENGTH)
-      this.ethExecutionsCounter
-        .labels(EthImpl.ethSendRawTransaction, transaction.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH))
-        .inc();
-
     const networkGasPriceInWeiBars = Utils.addPercentageBufferToGasPrice(
       await this.getFeeWeibars(EthImpl.ethGasPrice, requestDetails),
     );
-
     const parsedTx = await this.parseRawTxAndPrecheck(transaction, requestDetails, networkGasPriceInWeiBars);
     const originalCallerAddress = parsedTx.from?.toString() || '';
+    const toAddress = parsedTx.to?.toString() || '';
+
+    this.ethExecutionsCounter
+      .labels(
+        EthImpl.ethSendRawTransaction,
+        parsedTx.data.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH) || '',
+        originalCallerAddress,
+        toAddress,
+      )
+      .inc();
+
     const transactionBuffer = Buffer.from(EthImpl.prune0x(transaction), 'hex');
 
     let fileId: FileId | null = null;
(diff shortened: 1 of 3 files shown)


natanasow (dismissed):
LGTM

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -140,6 +146,16 @@ export class HbarLimitService implements IHbarLimitService {
     logger.info(
       `HBAR Limiter successfully configured: totalBudget=${totalBudget}, maxLimitForBasicTier=${HbarLimitService.TIER_LIMITS.BASIC}, maxLimitForExtendedTier=${HbarLimitService.TIER_LIMITS.EXTENDED}, maxLimitForprivilegedTier=${HbarLimitService.TIER_LIMITS.PRIVILEGED}, limitDuration=${limitDuration}, resetTimeStamp=${this.reset}.`,
     );
+
+    const activeUsersPerLimitPeriodCounterName = 'rpc_relay_hbar_rate_limiter_active_users';
+    this.register.removeSingleMetric(activeUsersPerLimitPeriodCounterName);
+    this.activeUsersPerLimitPeriodCounter = new Counter({
+      name: activeUsersPerLimitPeriodCounterName,
+      help: 'Relay Hbar rate limiter active users counter',
```
Nana-EC, line 154: note this isn't specific to the spending plan.
This should be in general, it just happens that eth_sendRawTransaction and others might spend hBAR.
Does have it in the spending plan mean we'd miss calls that don't spend HBAR.
Maybe this should be placed higher up in the server?

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -359,6 +378,14 @@ export class HbarLimitService implements IHbarLimitService {
     Object.values(SubscriptionTier).forEach((tier) => this.uniqueSpendingPlansCounter[tier].reset());
   }
 
+  /**
+   * Resets the number of active users per rate limiter duration.
+   * @private
+   */
+  private resetActiveUsers(): void {
+    this.activeUsersPerLimitPeriodCounter.reset();
```
Nana-EC, line 386: I don't think we should be resetting in code. 
Shouldn't we be able to group this within the metric collection tool e.g. Grafana
For instance if the duration is not 24hrs the intent breaks

Nana-EC (changes requested):
Good change but not sure it's at the right level and might have side effects

ebadiere (commented):
Is there a way to have this managed by the metrics service?

packages/relay/src/lib/eth.ts
```diff
@@ -274,7 +280,8 @@ export class EthImpl implements Eth {
     this.cacheService = cacheService;
     this.mirrorNodeClient = mirrorNodeClient;
     this.precheck = new Precheck(mirrorNodeClient, logger, chain);
-    this.ethExecutionsCounter = this.initEthExecutionCounter(registry);
+    this.ethExecutionsCounter = this.initCounter('rpc_relay_eth_executions', ['method', 'function'], registry);
```
Nana-EC, line 283: Scratch `rpc_relay_daily_active_users` and let's reuse `rpc_relay_eth_executions` in the following form
```suggestion
    this.ethExecutionsCounter = this.initCounter('rpc_relay_eth_executions', ['method', 'function', 'from', 'to'], registry);
```

The rest of the changes seem good expect I think we should be able to add a test case to confirm we're capturing it

Nana-EC (commented):
I think we've got to an agreement on the metric choice now

Nana-EC (dismissed):
LG

packages/relay/src/lib/eth.ts
```diff
@@ -285,13 +289,12 @@ export class EthImpl implements Eth {
     return !CommonService.blockTagIsLatestOrPendingStrict(tag) && !CommonService.isDevMode;
   }
 
-  private initEthExecutionCounter(register: Registry): Counter {
-    const metricCounterName = 'rpc_relay_eth_executions';
+  private initCounter(metricCounterName: string, labelNames: string[], register: Registry): Counter {
```
quiet-node, line 292: qq: why did we want to rename the method?

packages/relay/src/lib/eth.ts
```diff
@@ -1553,17 +1561,22 @@ export class EthImpl implements Eth {
    */
   async sendRawTransaction(transaction: string, requestDetails: RequestDetails): Promise<string | JsonRpcError> {
     const requestIdPrefix = requestDetails.formattedRequestId;
-    if (transaction?.length >= constants.FUNCTION_SELECTOR_CHAR_LENGTH)
```
quiet-node, line 1556: I see this increment of the log has been moved to line 1569 but didn't understand why we removed this check

packages/relay/src/lib/eth.ts
```diff
@@ -1655,12 +1668,15 @@ export class EthImpl implements Eth {
...
-    if (callDataSize >= constants.FUNCTION_SELECTOR_CHAR_LENGTH) {
-      this.ethExecutionsCounter
-        .labels(EthImpl.ethCall, callData!.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH))
-        .inc();
-    }
+
+    this.ethExecutionsCounter
+      .labels(
+        EthImpl.ethCall,
+        callData ? callData.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH) : '',
```
quiet-node, line 1675:
```suggestion
        callData?.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH) ?? '',
```

packages/relay/src/lib/eth.ts
```diff
@@ -1655,12 +1668,15 @@ export class EthImpl implements Eth {
     // log call data size
     const callDataSize = callData ? callData.length : 0;
     this.logger.trace(`${requestIdPrefix} call data size: ${callDataSize}`);
-    // metrics for selector
-    if (callDataSize >= constants.FUNCTION_SELECTOR_CHAR_LENGTH) {
```
quiet-node, line 1659: Same here why is this check removed again?

quiet-node (commented):
LGTM some question

packages/relay/src/lib/eth.ts
nadezhdapopovaa: firstly, I implemented a second counter and reworked this method so I can reuse it and not have repetitive code. Afterwords we decided to use one counter with different labels. I left this change, so we can use it if we decide to refactor the metrics service

packages/relay/src/lib/eth.ts
nadezhdapopovaa: removing the checks, because we are adding more labels to the existing counter. the checks were for contract call and creation. now we want the counter to be incremented on a method call and depending on the labels we'll filter it in grafana.

packages/relay/src/lib/eth.ts
quiet-node: Nice! sounds good to me! Thanks for the explanation!

packages/relay/src/lib/eth.ts
quiet-node: Ah so earlier we want to only capture this metric for contract call which has a calldata > than constants.FUNCTION_SELECTOR_CHAR_LENGTH (basically the 4 bytes selectors). But now we can reuse the execution counter to capture metric in different methods. Okay good point! Thanks for confirming!

packages/relay/src/lib/eth.ts
```diff
@@ -1553,17 +1561,22 @@ export class EthImpl implements Eth {
...
     );
-
     const parsedTx = await this.parseRawTxAndPrecheck(transaction, requestDetails, networkGasPriceInWeiBars);
     const originalCallerAddress = parsedTx.from?.toString() || '';
+    const toAddress = parsedTx.to?.toString() || '';
+
+    this.ethExecutionsCounter
+      .labels(
+        EthImpl.ethSendRawTransaction,
+        transaction.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH) || '',
```
quiet-node, line 1574: Hmm, then this may not apply directly to `sendRawTransaction`. The `transaction` parameter in `sendRawTransaction` isn’t encoded the same way as in `eth_call`. `transaction` for sendRawTransaction would not have the initial 4-byte hash segment indicating the function selector. Therefore in this case, the `transaction.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH) || '',` would not yield any useful information as the result would not be a 4 byte transaction selector. 

However, this is not in the scope of this PR so please create a ticket to capture this and it can be address in the next PR.

quiet-node (dismissed):
LGTM! One thing can be addressed in follow PR.

Great work! 🙌🏼

victor-yanev (approved):

quiet-node (approved):
LGTM

konstantinabl (approved):

quiet-node (dismissed):
LGTM 1 question

quiet-node (approved):

konstantinabl (approved):

PR #3153, opened 2024-10-23
feat: Added ability to load spending plans from environment variable.

### LINKED ISSUE #3152: Add the ability for spending plans to be defined using environment variables.
### Problem

We don't have access to create new config maps in Hashio, but we can add environment variables. 

### Solution

 For testing spending plans we should be able to read in the plans as environment variables.

### Alternatives

_No response_

ebadiere:
This enhancement adds the ability to load HBar Rate Limiter spending plans from an environment variable. 
**Related issue(s)**:

Fixes #3152 

**Checklist**


docs/configuration.md
```diff
@@ -75,6 +75,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
 | `HBAR_RATE_LIMIT_EXTENDED`                  | "3200000000"               | Individual limit (in tinybars) for spending plans with a EXTENDED tier. Defaults to 32 HBARs.                                                                                                                                                                                                          |
 | `HBAR_RATE_LIMIT_PRIVILEGED`                | "8000000000"               | Individual limit (in tinybars) for spending plans with a PRIVILEGED tier. Defaults to 80 HBARs.                                                                                                                                                                                                        |
+| `HBAR_SPENDING_PLANS_CONFIG_JSON`           | "" | Spending plans can be loaded from environment varaibles in JSON format, based on the spendingPlansConfig.example.json file.                                    |
```
Nana-EC, line 78: Is it possible to reuse `SPENDING_PLANS_CONFIG_FILE` and rename it to something more general?
In that way if the logic realizes it's a json file it'll read it, if not it'll parse JSON text.
Could this work? Are there tradeoffs?
I'm mindful of the number of configs and the result of both cases is a valid spending plan

Nana-EC (commented):
Config suggestion, other than that looks good.

docs/configuration.md
victor-yanev, line 78: **nit:** Format the column widths of this row to be consistent with other rows in the table.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -58,6 +58,14 @@ export class HbarSpendingPlanConfigService {
   private readonly SPENDING_PLANS_CONFIG_FILE: string =
     ConfigService.get('HBAR_SPENDING_PLANS_CONFIG_FILE') || 'spendingPlansConfig.json';
 
+  /**
+   * The environment variable that contains the spending plans configuration JSON.
+   *
+   * @type {string}
+   * @private
+   */
+  private readonly SPENDING_PLANS_CONFIG_JSON: string | undefined = process.env.HBAR_SPENDING_PLANS_CONFIG_JSON;
```
victor-yanev, line 67: Please, use `ConfigService` instead of `process.env` here, this should be the standard way of accessing env variables now:
```suggestion
  private readonly SPENDING_PLANS_CONFIG_JSON: string | undefined = ConfigService.get('HBAR_SPENDING_PLANS_CONFIG_JSON');
```

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -502,6 +520,107 @@ describe('HbarSpendingPlanConfigService', function () {
         });
       });
     });
+
+    describe('Environment variable defined spending plans', function () {
+      const spendingPlansConfigExample = [
```
victor-yanev, line 525: For better type checking:
```suggestion
      const spendingPlansConfigExample: SpendingPlanConfig[] = [
```

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -502,6 +520,107 @@ describe('HbarSpendingPlanConfigService', function () {
...
+
+      describe('when HBAR_SPENDING_PLANS_CONFIG_JSON is invalid JSON', function () {
+        before(function () {
+          delete process.env.HBAR_SPENDING_PLANS_CONFIG_JSON;
+          process.env.HBAR_SPENDING_PLANS_CONFIG_JSON = 'invalid JSON';
+        });
+
+        after(function () {
+          delete process.env.HBAR_SPENDING_PLANS_CONFIG_JSON;
+        });
```
victor-yanev, line 573:
```suggestion
        overrideEnvsInMochaDescribe({
          HBAR_SPENDING_PLANS_CONFIG_JSON: 'invalid JSON',
        });
```

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -502,6 +520,107 @@ describe('HbarSpendingPlanConfigService', function () {
...
+          subscriptionTier: 'EXTENDED',
+        },
+      ];
+
+      describe('when HBAR_SPENDING_PLANS_CONFIG_JSON is not set', function () {
+        overrideEnvsInMochaDescribe({
+          HBAR_SPENDING_PLANS_CONFIG_JSON: undefined,
+        });
+
+        it('should not throw an error if the environment variable is not set', async function () {
```
victor-yanev, line 559: **nit:** It works better, if we don't repeat the condition in the title of the test case, since the wrapper describe already specifies that `HBAR_SPENDING_PLANS_CONFIG_JSON is not set`
```suggestion
        it('should not throw an error', async function () {
```

When executing the tests it would look like:
- when HBAR_SPENDING_PLANS_CONFIG_JSON is not set
  - should not throw an error

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -502,6 +520,107 @@ describe('HbarSpendingPlanConfigService', function () {
...
+          expect(process.env.HBAR_SPENDING_PLANS_CONFIG_JSON).to.be.undefined;
```
victor-yanev, line 560:
```suggestion
          expect(ConfigService.get('HBAR_SPENDING_PLANS_CONFIG_JSON')).to.be.undefined;
```

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -502,6 +520,107 @@ describe('HbarSpendingPlanConfigService', function () {
...
+          delete process.env.HBAR_SPENDING_PLANS_CONFIG_JSON;
+          process.env.HBAR_SPENDING_PLANS_CONFIG_JSON = 'invalid JSON';
+        });
+
+        after(function () {
+          delete process.env.HBAR_SPENDING_PLANS_CONFIG_JSON;
+        });
+
+        it('should throw an error if environment variable contains invalid JSON', async function () {
+          hbarSpendingPlanConfigService = await reloadHBarSPendingPlanConfigService(
```
victor-yanev, line 576: It's better not to override the outer instance here:
```suggestion
          const hbarSpendingPlanConfigService = await reloadHBarSPendingPlanConfigService(
```

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -502,6 +520,107 @@ describe('HbarSpendingPlanConfigService', function () {
...
+      });
+
+      describe('when HBAR_SPENDING_PLANS_CONFIG_JSON is valid JSON', function () {
+        before(function () {
+          process.env.HBAR_SPENDING_PLANS_CONFIG_JSON = JSON.stringify(spendingPlansConfigExample);
+        });
+
+        after(function () {
+          delete process.env.HBAR_SPENDING_PLANS_CONFIG_JSON;
+        });
```
victor-yanev, line 596: Same here, use `overrideEnvsInMochaDescribe`

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -502,6 +520,107 @@ describe('HbarSpendingPlanConfigService', function () {
...
+
+        it('should populate the database with pre-configured spending plans from the HBAR_SPENDING_PLANS_CONFIG_JSON environment variable', async function () {
+          hbarSpendingPlanConfigService = await reloadHBarSPendingPlanConfigService(
```
victor-yanev, line 599: Same here - it's better not to override the outer instance
```suggestion
          const hbarSpendingPlanConfigService = await reloadHBarSPendingPlanConfigService(
```

victor-yanev (changes requested):
Looks good, just a few suggestions:

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
victor-yanev: Please also apply this for other occurrences below 🙏

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
victor-yanev: Actually, we already have this available in the `spendingPlansConfig` variable above:

```typescript
  const spendingPlansConfigFile = 'spendingPlansConfig.example.json';
  const path = findConfig(spendingPlansConfigFile);
  const spendingPlansConfig = JSON.parse(fs.readFileSync(path!, 'utf-8')) as SpendingPlanConfig[];
```

docs/configuration.md
ebadiere: It's renamed to simply `SPENDING_PLANS_CONFIG`.  Yes it can work, and yes there are trade offs.

**Pros**
- simplicity
- flexibility
- maintainability

**Cons**
- ambiguity:  An input string could validly be both a file path and a JSON string.  Tests cover both scenarios and JSON content is validated when it's parsed.
- Performance Overhead: Checking if a file exists (fs.existsSync) introduces additional I/O operations, however this happens only once at start up.

ebadiere: Fixed.  Removed line when renamed to `HBAR_SPENDING_PLANS_CONFIG`

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
ebadiere: Done

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -108,10 +109,16 @@ export class HbarSpendingPlanConfigService {
    * @private
    */
   private loadSpendingPlansConfig(): SpendingPlanConfig[] {
-    const configPath = findConfig(this.SPENDING_PLANS_CONFIG_FILE);
+    const configPath = findConfig(this.SPENDING_PLANS_CONFIG);
```
Nana-EC, line 112: Can clean this up. 
If the config is undefined then we throw undefined error.
Beyond that it's either a file or valid json.
If the file path exist try and parse.
If not then try for a valid JSON parse.
If both cases fail then the config value is malformed.
Right now the validation paths are muddled up

Nana-EC (dismissed):
Looking good, just a non-blocking clean up suggestion on the logic.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
ebadiere: Done.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -108,16 +109,19 @@ export class HbarSpendingPlanConfigService {
    * @private
    */
   private loadSpendingPlansConfig(): SpendingPlanConfig[] {
-    const configPath = findConfig(this.SPENDING_PLANS_CONFIG_FILE);
-    if (!configPath || !fs.existsSync(configPath)) {
-      this.logger.trace(`Configuration file not found at path "${configPath ?? this.SPENDING_PLANS_CONFIG_FILE}"`);
-      return [];
+    const configPath = findConfig(this.SPENDING_PLANS_CONFIG);
+    if (!this.SPENDING_PLANS_CONFIG) {
```
Nana-EC, line 113: Q: shouldn't this be the returned value since the value is extracted for you already 
```suggestion
    if (!this.spendingPlanConfig) {
```

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -108,16 +109,19 @@ export class HbarSpendingPlanConfigService {
    * @private
    */
   private loadSpendingPlansConfig(): SpendingPlanConfig[] {
-    const configPath = findConfig(this.SPENDING_PLANS_CONFIG_FILE);
-    if (!configPath || !fs.existsSync(configPath)) {
-      this.logger.trace(`Configuration file not found at path "${configPath ?? this.SPENDING_PLANS_CONFIG_FILE}"`);
-      return [];
+    const configPath = findConfig(this.SPENDING_PLANS_CONFIG);
```
Nana-EC, line 112: nit: to make the variable name more accurate
```suggestion
    const spendingPlanConfig = findConfig(this.SPENDING_PLANS_CONFIG);
```

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -108,16 +109,19 @@ export class HbarSpendingPlanConfigService {
...
+    if (!this.SPENDING_PLANS_CONFIG) {
+      throw new Error('SPENDING_PLANS_CONFIG is undefined');
     }
     try {
-      const rawData = fs.readFileSync(configPath, 'utf-8');
-      return JSON.parse(rawData) as SpendingPlanConfig[];
+      // Try to parse the value as a file path
+      if (configPath && fs.existsSync(configPath)) {
```
Nana-EC, line 118: If the above check is correct you shouldn't need to recheck the config value
```suggestion
      if (fs.existsSync(configPath)) {
```

packages/config-service/src/services/globalConfig.ts
```diff
@@ -321,8 +321,8 @@ export class GlobalConfig {
       required: true,
       defaultValue: null,
     },
-    HBAR_SPENDING_PLANS_CONFIG_FILE: {
-      envName: 'HBAR_SPENDING_PLANS_CONFIG_FILE',
+    HBAR_SPENDING_PLANS_CONFIG: {
+      envName: 'HBAR_SPENDING_PLANS_CONFIG',
       type: 'string',
       required: false,
       defaultValue: null,
```
victor-yanev, line 328:
```suggestion
      defaultValue: 'spendingPlansConfig.json',
```

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -49,15 +49,16 @@ export class HbarSpendingPlanConfigService {
...
    * @private
    */
-  // @ts-ignore
-  private readonly SPENDING_PLANS_CONFIG_FILE: string =
-    ConfigService.get('HBAR_SPENDING_PLANS_CONFIG_FILE') || 'spendingPlansConfig.json';
-
+  private readonly SPENDING_PLANS_CONFIG: string | undefined =
+    typeof ConfigService.get('HBAR_SPENDING_PLANS_CONFIG') === 'string'
+      ? (ConfigService.get('HBAR_SPENDING_PLANS_CONFIG') as string)
+      : undefined;
```
victor-yanev, line 61: We should have `spendingPlansConfig.json` as a default value, and this should never be `undefined`.

- If the env variable is set:
  - It will always be parsed as a string: https://github.com/hashgraph/hedera-json-rpc-relay/blob/6442c98d95a9b8d6fcb12bb94c77e92fd67edfef/packages/config-service/src/services/globalConfig.ts#L326
- If the env variable is not set 
  - it will automatically resolve to the default value in `ValidationService#typeCasting` if we apply the suggestion from my previous comment: https://github.com/hashgraph/hedera-json-rpc-relay/blob/6442c98d95a9b8d6fcb12bb94c77e92fd67edfef/packages/config-service/src/services/validationService.ts#L53

Thus, we can replace this with:
```suggestion
  private readonly SPENDING_PLANS_CONFIG: string = ConfigService.get('HBAR_SPENDING_PLANS_CONFIG') as string;
```

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
victor-yanev: +1, we should check the result from `findConfig` here, not the constant that holds the env variable.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -108,16 +109,19 @@ export class HbarSpendingPlanConfigService {
...
     }
     try {
-      const rawData = fs.readFileSync(configPath, 'utf-8');
-      return JSON.parse(rawData) as SpendingPlanConfig[];
+      // Try to parse the value as a file path
+      if (configPath && fs.existsSync(configPath)) {
+        return JSON.parse(fs.readFileSync(configPath, 'utf-8')) as SpendingPlanConfig[];
+      }
+      // If it's not a valid file path, try to parse it directly as JSON
+      return JSON.parse(this.SPENDING_PLANS_CONFIG) as SpendingPlanConfig[];
```
victor-yanev, line 122: Missing test coverage

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
victor-yanev: You resolved this comment, but you still haven't removed this constant and reused `spendingPlansConfig`, could you explain?

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
victor-yanev: Still not resolved

victor-yanev (changes requested):

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
ebadiere: Removed and reused existing tests as suggested.

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
ebadiere: No longer relevant.  Reused existing tests as suggested.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
ebadiere: Both the `spendingPlanConfig` and `findConfig` we also used here.

packages/config-service/src/services/globalConfig.ts
ebadiere: Done.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
ebadiere: Now covered.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
ebadiere: Replaced by simply loading it in the `loadSpendingPlansConfig()` and added the default value `spendingPlansConfig.json` to the `GlobalConfig`

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -504,11 +514,19 @@ describe('HbarSpendingPlanConfigService', function () {
...
+
+  describe('with shared cache enabled and spending plan config variable', function () {
+    tests(true, false);
   });
 
-  describe('with shared cache disabled', function () {
-    tests(false);
+  describe('with shared cache disabled and spending plan config file', function () {
+    tests(false, true);
   });
```
victor-yanev, line 531: Imo, it would look better if these were grouped by the cache type:
```typescript
  describe('using Redis cache', function () {
    describe('and with a spending plan config file', function () {
      tests(true, true);
    });

    describe('and with a spending plan config variable', function () {
      tests(true, false);
    });
  });

  describe('using LRU cache', function () {
    describe('and with a spending plan config file', function () {
      tests(false, true);
    });
    
    describe('and with a spending plan config variable', function () {
      tests(false, false);
    });
  });
```

packages/relay/tests/lib/relay.spec.ts
```diff
@@ -133,9 +120,11 @@ describe('RelayImpl', () => {
 
         expect(populatePreconfiguredSpendingPlansSpy.calledOnce).to.be.true;
         await expect(populatePreconfiguredSpendingPlansSpy.returnValues[0]).not.to.be.rejected;
+        const failedPreConfiguredSpendingPlansMessage = `Failed to load pre-configured spending plans: `;
+        const failedToLoadEnvVarSpendingPlan = `Failed to load HBAR_SPENDING_PLAN. JSON parse error: Unexpected token 's', "spendingPl"... is not valid JSON; `;
```
victor-yanev, line 124: Why have we changed the expected message here? 

We shouldn't be getting this error, instead we should expect this message:
```typescript
const cause = `Failed to parse JSON from ${path}: Unexpected token 'i', "invalid JSON" is not valid JSON`;
```

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -36,7 +36,7 @@ import { ConfigService } from '@hashgraph/json-rpc-config-service/dist/services'
  * It reads the pre-configured spending plans from a JSON file and populates the cache with them.
  *
  * @see SpendingPlanConfig
- * @see SPENDING_PLANS_CONFIG_FILE
+ * @see SPENDING_PLANS_CONFIG
```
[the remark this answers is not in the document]
ebadiere: Removed

packages/config-service/tests/src/services/loggerService.spec.ts
```diff
@@ -37,7 +37,7 @@ describe('LoggerService tests', async function () {
   });
 
   it('should be able to mask every value if it starts with known secret prefix', async () => {
-    const { envName } = GlobalConfig.ENTRIES.HBAR_SPENDING_PLANS_CONFIG_FILE;
+    const { envName } = GlobalConfig.ENTRIES.HBAR_SPENDING_PLANS_CONFIG;
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -93,18 +98,23 @@ describe('HbarSpendingPlanConfigService', function () {
...
+        if (isSpendingPlansConfigFile) {
+          overrideEnvsInMochaDescribe({ HBAR_SPENDING_PLANS_CONFIG: 'non-existent-file.json' });
+        }
+        // Only relevant when the configuration is a file
+        if (isSpendingPlansConfigFile) {
+          it('should throw an error if the configuration file is not found', async function () {
+            await expect(hbarSpendingPlanConfigService.populatePreconfiguredSpendingPlans()).to.be.rejectedWith(
+              /Failed to load HBAR_SPENDING_PLAN. JSON parse error: /,
+            );
+          });
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/relay/tests/lib/relay.spec.ts
ebadiere: Done.

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
ebadiere: Done.

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -108,16 +98,30 @@ export class HbarSpendingPlanConfigService {
...
+        if (configFilePath && fs.existsSync(configFilePath)) {
+          const fileContent = fs.readFileSync(configFilePath, 'utf-8');
+          return JSON.parse(fileContent) as SpendingPlanConfig[];
+        } else {
+          throw new Error(`Configuration file not found at path "${configFilePath ?? spendingPlanConfig}"`);
+        }
+      } catch (fileError: any) {
+        throw new Error(
+          `Failed to load HBAR_SPENDING_PLAN. JSON parse error: ${jsonParseError.message}; File error: ${fileError.message}`,
+        );
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -93,18 +98,23 @@ describe('HbarSpendingPlanConfigService', function () {
...
 
+          it('should throw an error if configuration file is not a parsable JSON', async function () {
+            await expect(hbarSpendingPlanConfigService.populatePreconfiguredSpendingPlans()).to.be.rejectedWith(
+              /Failed to load HBAR_SPENDING_PLAN. JSON parse error: /,
+            );
+          });
+        }
```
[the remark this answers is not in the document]
ebadiere: Done.

victor-yanev (changes requested):

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -125,17 +146,19 @@ describe('HbarSpendingPlanConfigService', function () {
 
     describe('populatePreconfiguredSpendingPlans', function () {
       describe('negative scenarios', function () {
-        it('should not throw an error if the configuration file is not found', async function () {
-          sinon.stub(fs, 'existsSync').returns(false);
-          await expect(hbarSpendingPlanConfigService.populatePreconfiguredSpendingPlans()).not.to.be.rejected;
-        });
```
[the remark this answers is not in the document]
ebadiere: Previously reviewed code had missing configuration files throwing errors.  It can be updated to log messages instead:
```
      try {
        const configFilePath = findConfig(spendingPlanConfig);
        if (configFilePath && fs.existsSync(configFilePath)) {
          const fileContent = fs.readFileSync(configFilePath, 'utf-8');
          return JSON.parse(fileContent) as SpendingPlanConfig[];
        } else {
          throw new Error(`Configuration file not found at path "${configFilePath ?? spendingPlanConfig}"`);
        }
      } catch (fileError: any) {
        throw new Error(`File error: ${fileError.message}`);
      }
```

packages/relay/tests/lib/relay.spec.ts
```diff
@@ -105,23 +105,10 @@ describe('RelayImpl', () => {
...
-      overrideEnvsInMochaDescribe({ HBAR_SPENDING_PLANS_CONFIG_FILE: nonExistingFile });
-
-      it('should not throw an error', async () => {
-        expect((relay = new RelayImpl(logger, register))).to.not.throw;
-
-        expect(populatePreconfiguredSpendingPlansSpy.calledOnce).to.be.true;
-        await expect(populatePreconfiguredSpendingPlansSpy.returnValues[0]).to.not.be.rejected;
-        expect(loggerSpy.warn.notCalled).to.be.true;
-      });
-    });
```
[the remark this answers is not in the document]
ebadiere: Same as above.  I can restore these if we agree that missing files should be logged.

packages/config-service/tests/src/services/loggerService.spec.ts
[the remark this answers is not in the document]
ebadiere: The author of this particular test just grabbed the first env var he found to use in this test, and it happened to be `HBAR_SPENDING_PLANS_CONFIG_FILE`. I had to update it to the new value in order for the tests to run, but I can change it to something more relevant to masks like `OPERATOR_KEY`

packages/relay/tests/lib/relay.spec.ts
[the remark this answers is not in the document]
ebadiere: Added back/

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -125,17 +146,19 @@ describe('HbarSpendingPlanConfigService', function () {
...
-        it('should throw an error if configuration file is not a parsable JSON', async function () {
-          sinon.stub(fs, 'readFileSync').returns('invalid JSON');
-          await expect(hbarSpendingPlanConfigService.populatePreconfiguredSpendingPlans()).to.be.rejectedWith(
-            `Failed to parse JSON from ${path}: Unexpected token 'i', "invalid JSON" is not valid JSON`,
-          );
-        });
+        withOverriddenEnvsInMochaTest(
+          {
+            HBAR_SPENDING_PLANS_CONFIG: 'non-existent-file.json',
+          },
```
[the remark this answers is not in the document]
ebadiere: Good catch.  Updated to use the existing file declared in `spendingPlansConfigFile` and updated error message.

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -356,6 +384,11 @@ describe('HbarSpendingPlanConfigService', function () {
         };
 
         it('should populate the database with pre-configured spending plans', async function () {
+          // Clear existing spending plans.  Needed in CI environment where the plans are already populated
+          for (const plan of spendingPlansConfig) {
+            await hbarSpendingPlanRepository.delete(plan.id, emptyRequestDetails);
+          }
```
victor-yanev, line 390: Not really sure why this is needed? 

1. We are initializing a completely new redis server for these tests, so there shouldn't be anything populated in the cache in the beginning of this suite
2. We should be clearing the state using `await cacheService.clear(requestDetails)` after every test, in order not to have interdependency between tests

We are either reusing the same redis server, or we are not properly clearing the cache after every test. Please remove this from here and fix the root problem. 🙏

victor-yanev (changes requested):
LGTM, one last thing

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -30,13 +30,6 @@ import { SubscriptionTier } from '../db/types/hbarLimiter/subscriptionTier';
 import { IDetailedHbarSpendingPlan } from '../db/types/hbarLimiter/hbarSpendingPlan';
 import { ConfigService } from '@hashgraph/json-rpc-config-service/dist/services';
 
-/**
```
Nana-EC, line 33: Q: why the removal of the class comments?

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
```diff
@@ -127,17 +120,31 @@ export class HbarSpendingPlanConfigService {
...
+      logger.trace(
+        `Failed to parse HBAR_SPENDING_PLAN as JSON: ${jsonParseError.message}, now treating it as a file path...`,
+      );
+      try {
+        const configFilePath = findConfig(spendingPlanConfig);
+        if (configFilePath && fs.existsSync(configFilePath)) {
+          const fileContent = fs.readFileSync(configFilePath, 'utf-8');
+          return JSON.parse(fileContent) as SpendingPlanConfig[];
+        } else {
+          throw new Error(`Configuration file not found at path "${configFilePath ?? spendingPlanConfig}"`);
```
Nana-EC, line 143: nit:
```suggestion
          throw new Error(`HBAR Spending Configuration file not found at path "${configFilePath ?? spendingPlanConfig}"`);
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -232,12 +232,12 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
       });
 
-      describe('Total HBAR Limit', () => {
+      describe('@hbarlimiter-batch1 Total HBAR Limit', () => {
         it('should execute "eth_sendRawTransaction" without triggering HBAR rate limit exceeded', async function () {
           const parentContract = await deployContract(parentContractJson, accounts[0].wallet);
 
           // awaiting for HBAR limiter to finish updating expenses in the background
-          await Utils.wait(6000);
+          await Utils.wait(10000);
```
Nana-EC, line 240: Maybe not an issue to address in this PR but waits are a code smell in tests and needing to increase them confirms there's a coordination logic not being properly incorporated into the tests.
Can we open a ticket to address this in the future. What are we actually waiting on and can that status be polled for instead of a generic wait?

packages/relay/src/lib/config/hbarSpendingPlanConfigService.ts
ebadiere: Accidentally removed when the `HBAR_SPENDING_CONFIG` property initial reading was moved to the `loadSpendingPlansConfig` method.  Restored.

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
ebadiere: Removed.

tomzhenghedera (approved):

victor-yanev (approved):

quiet-node (approved):

PR #3158, opened 2024-10-25
fix: Inconsistent handling of failed EVM transactions

### LINKED ISSUE #3115: Inconsistent Handling of Failed EVM Transactions in JSON-RPC Relay
Title: 

## Description:

I'm encountering an issue where the same EVM transaction appears in consecutive blocks on the Hedera testnet, and the JSON-RPC relay handles these duplicates inconsistently. Specifically, a transaction that fails due to WRONG_NONCE is included in the next block but is misrepresented as successful in both the JSON-RPC responses and the block explorer.

## Steps to Reproduce:

Observe the Duplicate Transactions:

Visit the following block in HashScan: [Block 9337619](https://hashscan.io/testnet/block/9337619)
Notice that the EVM transaction appears twice:
[Transaction Success](https://hashscan.io/testnet/transaction/1726642023.105113000)
[Transaction Failure](https://hashscan.io/testnet/transaction/1726642023.211917000)

Query the Transaction Receipt by Hash:
```bash
curl -s https://testnet.hashio.io/api \
  --header "Content-Type: application/json" \
  --data '{"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":["0xb3390118380397cb4e483d01aa0f2f95fef2d3b2ce681fac5c60b5a7ee34492c"],"id":1}' | jq
```

The receipt indicates success, even for the failed transaction.
Retrieve EVM Transactions in Block 9337619:
```bash
curl -s https://testnet.hashio.io/api \
  --data '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x8e7b13", true]}' | jq
```

Only one instance of the transaction is returned, despite there being two occurrences.
Retrieve EVM Transactions in Block 9337620 (Next Block):

```bash
curl -s https://testnet.hashio.io/api \
  --data '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x8e7b14", true]}' | jq
```

The failed duplicate transaction is included in this block as well.

## Expected Behavior:
Failed EVM transactions due to WRONG_NONCE should not be included in subsequent blocks or should be accurately represented as failed in the JSON-RPC responses.
The JSON-RPC relay should not misrepresent the status of failed transactions.
Fees should not be charged for transactions that fail due to WRONG_NONCE.

## Actual Behavior:
The same EVM transaction appears in two consecutive blocks.
The failed transaction is included and misrepresented as successful in the JSON-RPC responses and on the block explorer.
Events emitted during the failed transaction are displayed, causing confusion.
It's unclear whether fees are being charged for the failed transaction.

## Impact:
This issue affects transaction receipt and event validation, which are crucial for cross-chain message validation.
Applications relying on accurate transaction data may experience inconsistencies or erroneous behavior.

## Additional Information:
Transaction Hash: 0xb3390118380397cb4e483d01aa0f2f95fef2d3b2ce681fac5c60b5a7ee34492c
Blocks Affected: 9337619 and 9337620 (Testnet)
RPC used: https://testnet.hashio.io/api

## Relevant Links:
[HashScan Block 9337619](https://hashscan.io/testnet/block/9337619)
[Successful Transaction Details](https://hashscan.io/testnet/transaction/1726642023.105113000)
[Failed Transaction Details](https://hashscan.io/testnet/transaction/1726642023.211917000)

## Questions:
Can failed EVM transactions be excluded from JSON-RPC responses to prevent confusion?:

Please investigate this issue to ensure that:

Failed EVM transactions are accurately represented in JSON-RPC responses.
Duplicate transactions are handled consistently across blocks.

victor-yanev:
**Description**:

This PR fixes the issue discussed in the comments of #3115 

**Related issue(s)**:

Fixes #3115

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -48,6 +48,7 @@ import {
   formatContractResult,
   formatTransactionIdWithoutQueryParams,
   getFunctionSelector,
+  hexToASCII,
   isHex,
   isValidEthereumAddress,
   nanOrNumberTo0x,
@@ -55,6 +56,7 @@ import {
   numberTo0x,
   parseNumericEnvVar,
   prepend0x,
+  strip0x,
   toHash32,
   trimPrecedingZeros,
   weibarHexToTinyBarInt,
@@ -326,10 +328,9 @@ export class EthImpl implements Eth {
     requestDetails: RequestDetails,
   ): Promise<IFeeHistory | JsonRpcError> {
     const requestIdPrefix = requestDetails.formattedRequestId;
-    const maxResults =
-      ConfigService.get('TEST')
-        ? constants.DEFAULT_FEE_HISTORY_MAX_RESULTS
-        : Number(ConfigService.get('FEE_HISTORY_MAX_RESULTS'));
+    const maxResults = ConfigService.get('TEST')
+      ? constants.DEFAULT_FEE_HISTORY_MAX_RESULTS
+      : Number(ConfigService.get('FEE_HISTORY_MAX_RESULTS'));
 
     this.logger.trace(
       `${requestIdPrefix} feeHistory(blockCount=${blockCount}, newestBlock=${newestBlock}, rewardPercentiles=${rewardPercentiles})`,
@@ -2297,9 +2298,20 @@ export class EthImpl implements Eth {
       throw predefined.MAX_BLOCK_SIZE(blockResponse.count);
     }
 
+    const isWrongNonce = (contractResult: { result: string; error_message: any }) => {
+      return (
+        contractResult.result === constants.TRANSACTION_RESULT_STATUS.WRONG_NONCE ||
+        hexToASCII(strip0x(contractResult.error_message ?? '')) === constants.TRANSACTION_RESULT_STATUS.WRONG_NONCE
+      );
+    };
+
     // prepare transactionArray
     let transactionArray: any[] = [];
     for (const contractResult of contractResults) {
+      if (isWrongNonce(contractResult)) {
+        // skip wrong nonce transactions as they are not valid transactions which should be included in the block
+        continue;
+      }
       contractResult.from = await this.resolveEvmAddress(contractResult.from, requestDetails, [constants.TYPE_ACCOUNT]);
       contractResult.to = await this.resolveEvmAddress(contractResult.to, requestDetails);
       contractResult.chain_id = contractResult.chain_id || this.chain;
@@ -2308,7 +2320,7 @@ export class EthImpl implements Eth {
     }
 
     transactionArray = this.populateSyntheticTransactions(showDetails, logs, transactionArray, requestDetails);
-    transactionArray = showDetails ? _.uniqBy(transactionArray, 'hash') : _.uniq(transactionArray);
+    transactionArray = showDetails ? transactionArray : _.uniq(transactionArray);
 
     const formattedReceipts: IReceiptRootHash[] = ReceiptsRootUtils.buildReceiptRootHashes(
       transactionArray.map((tx) => (showDetails ? tx.hash : tx)),
--- a/packages/relay/tests/lib/eth/eth_getBlockByNumber.spec.ts
+++ b/packages/relay/tests/lib/eth/eth_getBlockByNumber.spec.ts
@@ -567,65 +567,35 @@ describe('@ethGetBlockByNumber using MirrorNode', async function () {
     );
   });
 
-  it('eth_getBlockByNumber should skip transactions with wrong nonces when showDetails = false', async function () {
-    // mirror node request mocks
-    restMock.onGet(`blocks/${BLOCK_NUMBER}`).reply(200, DEFAULT_BLOCK);
-    restMock.onGet(BLOCKS_LIMIT_ORDER_URL).reply(200, MOST_RECENT_BLOCK);
-    restMock.onGet(CONTRACT_RESULTS_WITH_FILTER_URL).reply(200, {
-      results: [
-        ...defaultContractResults.results,
-        { ...defaultContractResults.results[0], result: 'WRONG_NONCE' },
-        { ...defaultContractResults.results[0], error_message: prepend0x(ASCIIToHex('WRONG_NONCE')) },
-      ],
+  [false, true].forEach((showDetails) => {
+    it(`eth_getBlockByNumber should skip wrong nonce transactions when showDetails = ${showDetails}`, async () => {
+      // mirror node request mocks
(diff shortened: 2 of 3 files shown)


ebadiere (dismissed):
LG.

natanasow (dismissed):
LGTM, but we should consider adding an exception for this check as well https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/TransactionProcessor.java#L233 in a follow-up PR.

packages/relay/tests/lib/eth/eth_getBlockByNumber.spec.ts
```diff
@@ -566,4 +566,66 @@ describe('@ethGetBlockByNumber using MirrorNode', async function () {
       args,
     );
   });
+
+  it('eth_getBlockByNumber should skip transactions with wrong nonces when showDetails = false', async function () {
```
acuarica, line 570: Aren't these two test cases basically the same module `showDetails=true|false`? If that's the case, maybe it's worth to merge them into a single test case, something like

```
[false, true].forEach(showDetails => {
  it(`eth_getBlockByNumber should skip transactions with wrong nonces when showDetails = ${showDetails}`, async function () {
  });
});
```

packages/relay/src/lib/eth.ts
```diff
@@ -2308,7 +2320,7 @@ export class EthImpl implements Eth {
     }
 
     transactionArray = this.populateSyntheticTransactions(showDetails, logs, transactionArray, requestDetails);
-    transactionArray = showDetails ? _.uniqBy(transactionArray, 'hash') : _.uniq(transactionArray);
+    transactionArray = showDetails ? transactionArray : _.uniq(transactionArray);
```
acuarica, line 2323: I'm not sure `showDetails` should change the length of the array, it concerns only with returning either transaction hashes or the transaction objects.

victor-yanev:
> LGTM, but we should consider adding an exception for this check as well https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/TransactionProcessor.java#L233 in a follow-up PR.

@natanasow Could you open a new issue for that, we could prioritize it in the next sprint. 😄

natanasow:
> > LGTM, but we should consider adding an exception for this check as well https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/TransactionProcessor.java#L233 in a follow-up PR.
> 
> @natanasow Could you open a new issue for that, we could prioritize it in the next sprint. 😄

Okay.

packages/relay/tests/lib/eth/eth_getBlockByNumber.spec.ts
victor-yanev: Fair point, that would simplify things a bit

packages/relay/src/lib/eth.ts
```diff
@@ -326,10 +328,9 @@ export class EthImpl implements Eth {
     requestDetails: RequestDetails,
   ): Promise<IFeeHistory | JsonRpcError> {
     const requestIdPrefix = requestDetails.formattedRequestId;
-    const maxResults =
-      ConfigService.get('TEST')
-        ? constants.DEFAULT_FEE_HISTORY_MAX_RESULTS
-        : Number(ConfigService.get('FEE_HISTORY_MAX_RESULTS'));
+    const maxResults = ConfigService.get('TEST')
```
Nana-EC, line 331: nit: for a different PR this feels like something the ConfigService should handle internally.
Is that possible?

packages/relay/src/lib/eth.ts
Nana-EC: Hmm, I agree with Luis here. The size should be the same regardless of `showDetails`.
I'm not sure why this logic exists, shouldn't they always be unique transactions based on hash?
Maybe something to follow up on, why should we ever have duplicates when `showDetails` is true?

victor-yanev: Btw, something that could be informative is that in ethereum `WRONG_NONCE` transactions will be discarded and won't make it out of the mempool and thus it's not possible to have such transactions inside a block.

But for `INVALID_ACCOUNT` transactions, it's kind of different. I'm not sure if we want to discard them from the response of `getBlockByHash`/`getBlockByNumber`, even though they are creating duplicated transaction hashes. What is your opinion?

packages/relay/src/lib/eth.ts
victor-yanev: Yes, we could add a `defaultValue` for `FEE_HISTORY_MAX_RESULTS` in the **globalConfig.ts** file

packages/relay/src/lib/eth.ts
quiet-node: >Btw, something that could be informative is that in ethereum WRONG_NONCE transactions will be discarded and won't make it out of the mempool and thus it's not possible to have such transactions inside a block.

Yes, this is expected. In our case, the Relay also quickly fails `WRONG_NONCE` transactions, thanks to `precheck.nonce()`. Transactions with a `WRONG_NONCE` code are stopped at the precheck stage and don’t reach the consensus node, meaning they’re not included in a block. I assume this may have happened in some blocks because we previously had a nonce buffer in the precheck, which allowed transactions with `nonce + 1` to pass the precheck but then ultimately get flagged as `WRONG_NONCE` by consensus nodes and included in a block. However, the nonce buffer was removed in a recent PR, so this shouldn’t occur again.

quiet-node (approved):

packages/relay/src/lib/eth.ts
victor-yanev: Opened an issue for that https://github.com/hashgraph/hedera-json-rpc-relay/issues/3176

packages/relay/src/lib/eth.ts
victor-yanev: To sum up, it's safe to use the lodash `uniq` method when the `transactionArray` contains only strings of EVM hashes, but when `showDetails` is `true` and the `transactionArray` contains actual objects, we want to show everything that has not been explicitly filtered by the for-loop above. 

If there are still some transaction objects with duplicated hashes, that means we have missed to filter something and we have a bug in the code, otherwise these bugs will be "hidden" if we keep using the `_.uniqBy(transactionArray, 'hash')` method.

natanasow (approved):

Issue #3118, opened 2024-10-17
Getting timeout exceeded in the sdk client even when transaction is submitted


konstantinabl:
### Description

It has been noticed that an 'SDK timeout exceeded' error appears sporadically in our CI, but also on mainnet. The transaction is actually passed to the consensus node, but we get a null value from the SDK thus it is assumed the transaction is not submitted.

### Steps to reproduce

Currently, not easily reproducable.
CI logs - https://github.com/hashgraph/hedera-json-rpc-relay/actions/runs/11142264400/job/30964850632
Mainnet logs - https://production.grafana.hedera-ops.com/goto/5H7WN0kNR?orgId=1

### Additional context

SDK Issue - https://github.com/hashgraph/hedera-sdk-js/issues/2614

_No response_

### Hedera network

mainnet

### Version

0.57

### Operating system

None

Nana-EC:
@konstantinabl can we connect the sdk bug here?

quiet-node:
@konstantinabl Was this got resolved?

Nana-EC:
@konstantinabl let's get a few logs for instances of transactions and pass it on to services team for investigations.
With the workaround this is less of a bug but we still want to close this off

konstantinabl:
As far as I understand services team says this is expected behaviour, so Im closing the issue
@Nana-EC @Ferparishuertas

PR #3161, opened 2024-10-28
fix: Adds retry to send raw transaction

konstantinabl:
**Description**:

Currently, we experience the SDk timing out in the CI and on mainnet, eventhough the transactions that fail do not take more than 10 seconds. More info in the issue below
**Related issue(s)**:

Workaround until  #3118 is solved

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -501,7 +501,7 @@ export class SDKClient {
         retryCount,
         requestDetails,
       );
-      const isFailed = transaction !== null ? true : false;
+      const isFailed = transaction !== null;
       return isFailed;
     } catch (e: any) {
       this.logger.error(
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -473,9 +473,10 @@ export class SDKClient {
       if (e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded())) {
         if (!(await this.isFailedTransaction(requestDetails, e.transactionId))) {
           txResponse = { transactionId: e.transactionId };
+        } else {
+          throw e;
         }
       }
-      throw e;
     }
 
     return {
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -724,7 +724,7 @@ export class SDKClient {
         throw e;
       }
 
-      const sdkClientError = new SDKClientError(e, e.message);
+      const sdkClientError = new SDKClientError(e, e.message, transaction.transactionId?.toString());
 
       // Throw WRONG_NONCE error as more error handling logic for WRONG_NONCE is awaited in eth.sendRawTransactionErrorHandler().
       if (sdkClientError.status && sdkClientError.status === Status.WrongNonce) {
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -795,9 +795,13 @@ export class SDKClient {
       );
 
       if (!transactionResponse) {
-        throw predefined.INTERNAL_ERROR(
-          `${requestDetails.formattedRequestId} Transaction execution returns a null value: transactionId=${transaction.transactionId}, callerName=${callerName}, txConstructorName=${txConstructorName}`,
-        );
+        if (sdkClientError.isConnectionDropped() || sdkClientError.isTimeoutExceeded()) {
+          throw sdkClientError;
+        } else {
+          throw predefined.INTERNAL_ERROR(
+            `${requestDetails.formattedRequestId} Transaction execution returns a null value: transactionId=${transaction.transactionId}, callerName=${callerName}, txConstructorName=${txConstructorName}`,
+          );
+        }
       }
       return transactionResponse;
     } finally {
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -69,6 +69,8 @@ import {
   ITransactionRecordMetric,
   RequestDetails,
 } from '../types';
+import { MirrorNodeClient } from './mirrorNodeClient';
+import { Registry } from 'prom-client';
 
 const _ = require('lodash');
 
@@ -123,6 +125,14 @@ export class SDKClient {
   private readonly hbarLimitService: HbarLimitService;
 
   /**
+   * An instance of the MirrorNodeClient
+   * @private
+   * @readonly
+   * @type {MirrorNodeClient}
+   */
+  private readonly mirrorNodeClient: MirrorNodeClient;
+
+  /**
    * Constructs an instance of the SDKClient and initializes various services and settings.
    *
    * @param {Client} clientMain - The primary Hedera client instance used for executing transactions and queries.
@@ -136,6 +146,7 @@ export class SDKClient {
     cacheService: CacheService,
     eventEmitter: EventEmitter,
     hbarLimitService: HbarLimitService,
+    register: Registry,
   ) {
     this.clientMain = clientMain;
 
@@ -151,6 +162,13 @@ export class SDKClient {
     this.hbarLimitService = hbarLimitService;
     this.maxChunks = Number(ConfigService.get('FILE_APPEND_MAX_CHUNKS')) || 20;
     this.fileAppendChunkSize = Number(ConfigService.get('FILE_APPEND_CHUNK_SIZE')) || 5120;
+    this.mirrorNodeClient = new MirrorNodeClient(
+      // @ts-ignore
+      ConfigService.get('MIRROR_NODE_URL') || '',
+      logger,
+      register,
+      cacheService,
+    );
   }
 
   /**
@@ -441,17 +459,56 @@ export class SDKClient {
       Hbar.fromTinybars(Math.floor(networkGasPriceInTinyBars * constants.MAX_GAS_PER_SEC)),
     );
 
-    return {
-      fileId,
-      txResponse: await this.executeTransaction(
+    let txResponse;
+    try {
+      txResponse = await this.executeTransaction(
         ethereumTransaction,
         callerName,
         interactingEntity,
         requestDetails,
         true,
         originalCallerAddress,
-      ),
+      );
+    } catch (e: any) {
+      if (e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded())) {
+        if (!(await this.isFailedTransaction(requestDetails, e.transactionId))) {
+          txResponse = { transactionId: e.transactionId };
+        }
+      }
+      throw e;
+    }
+
+    return {
+      fileId,
+      txResponse,
     };
+  }
+
+  /**
+   * Checks if a transaction has failed by querying the mirror node.
+   *
+   * @param {RequestDetails} requestDetails - The details of the request.
+   * @param {string} [transactionId] - The ID of the transaction to check.
+   * @returns {Promise<boolean>} - A promise that resolves to `true` if the transaction has failed, `false` otherwise.
+   */
+  async isFailedTransaction(requestDetails: RequestDetails, transactionId?: string): Promise<boolean> {
+    const retryCount = 5;
+    try {
+      const transaction = await this.mirrorNodeClient.repeatedRequest(
+        this.mirrorNodeClient.getTransactionById.name,
+        [transactionId, requestDetails],
+        retryCount,
+        requestDetails,
+      );
+      const isFailed = transaction !== null ? true : false;
+      return isFailed;
+    } catch (e: any) {
+      this.logger.error(
+        e,
+        `${requestDetails.formattedRequestId} Failed to check if transaction ${transactionId} is failed`,
+      );
+      return true;
+    }
   }
 
   /**
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -471,11 +471,13 @@ export class SDKClient {
       );
     } catch (e: any) {
       if (e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded())) {
-        if (!(await this.isFailedTransaction(requestDetails, e.transactionId))) {
-          txResponse = { transactionId: e.transactionId };
-        } else {
+        const isFailed = await this.isFailedTransaction(requestDetails, e.transactionId);
+        if (isFailed) {
           throw e;
         }
+        txResponse = { transactionId: e.transactionId };
+      } else {
+        throw e;
       }
     }
 
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1689,12 +1689,12 @@ export class EthImpl implements Eth {
     {
       maxAttempts = 2,
       backOff = 500,
-      canRetry,
+      canRetry = (error: unknown) => Promise.resolve(true),
       onError = (e: SDKClientError) => {},
     }: {
       maxAttempts?: number;
       backOff?: number;
-      canRetry: (error: unknown) => Promise<boolean>;
+      canRetry?: (error: unknown) => Promise<boolean>;
       onError?: (error: SDKClientError) => void;
     },
   ): Promise<T | undefined> {
@@ -1703,7 +1703,8 @@ export class EthImpl implements Eth {
       try {
         return await sendRawTransaction();
       } catch (e: unknown) {
-        if (!canRetry(e) || count === maxAttempts - 1) {
+        const shouldRetry = await canRetry(e);
+        if (!shouldRetry || count === maxAttempts - 1) {
           throw e;
         }
         onError(e as SDKClientError);
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1568,16 +1568,30 @@ export class EthImpl implements Eth {
     let fileId: FileId | null = null;
     let txSubmitted = false;
     try {
-      const sendRawTransactionResult = await this.hapiService
-        .getSDKClient()
-        .submitEthereumTransaction(
-          transactionBuffer,
-          EthImpl.ethSendRawTransaction,
-          requestDetails,
-          originalCallerAddress,
-          networkGasPriceInWeiBars,
-          await this.getCurrentNetworkExchangeRateInCents(requestDetails),
-        );
+      const sendRawTransactionResult = await this.sendRawTransactionWithRetry(
+        async () =>
+          await this.hapiService
+            .getSDKClient()
+            .submitEthereumTransaction(
+              transactionBuffer,
+              EthImpl.ethSendRawTransaction,
+              requestDetails,
+              originalCallerAddress,
+              networkGasPriceInWeiBars,
+              await this.getCurrentNetworkExchangeRateInCents(requestDetails),
+            ),
+        {
+          canRetry: (e: unknown) => {
+            return e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded());
+          },
+          onError: (e: SDKClientError) => {
+            this.logger.warn(
+              `${requestIdPrefix} SDK Client has probably timed out, trying again with a new instance...`,
+            );
+            this.hapiService.decrementErrorCounter(e.statusCode);
+          },
+        },
+      );
 
       txSubmitted = true;
       fileId = sendRawTransactionResult!.fileId;
@@ -1630,6 +1644,47 @@ export class EthImpl implements Eth {
           .getSDKClient()
           .deleteFile(fileId, requestDetails, EthImpl.ethSendRawTransaction, fileId.toString(), originalCallerAddress)
           .then();
+      }
+    }
+  }
+
+  /**
+   * Sends a raw transaction with retry logic.
+   *
+   * @template T The type of the result returned by the sendRawTransaction function.
+   * @param {Function} sendRawTransaction The function responsible for sending the raw transaction.
+   * @param {object} options The options for retrying the transaction.
+   * @param {number} [options.maxAttempts=2] The maximum number of attempts to send the transaction.
+   * @param {number} [options.backOff=500] The backoff period in milliseconds between retry attempts.
+   * @param {(error: unknown) => boolean} [options.canRetry=(error) => true] A function that determines whether a retry attempt can be made based on the error received.
+   * @param {(error: SDKClientError) => void} [options.onError=(error) => {}] A function to handle errors that occur during retry attempts.
+   * @returns {Promise<T>} A promise resolving to the result of the transaction.
+   */
+  private async sendRawTransactionWithRetry<T>(
+    sendRawTransaction: () => T,
+    {
+      maxAttempts = 2,
+      backOff = 500,
+      canRetry = (e: unknown) => true,
+      onError = (e: SDKClientError) => {},
+    }: {
+      maxAttempts?: number;
+      backOff?: number;
+      canRetry?: (error: unknown) => boolean;
+      onError?: (error: SDKClientError) => void;
+    },
+  ): Promise<T | undefined> {
+    const delay = (backOff) => new Promise((resolve) => setTimeout(resolve, backOff));
+    for (let count = 0; count < maxAttempts; count++) {
+      try {
+        return await sendRawTransaction();
+      } catch (e: unknown) {
+        if (!canRetry(e) || count === maxAttempts - 1) {
+          throw e;
+        }
+        onError(e as SDKClientError);
+        // eslint-disable-next-line no-await-in-loop
+        await delay(backOff);
       }
     }
   }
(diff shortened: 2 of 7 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1567,16 +1566,30 @@ export class EthImpl implements Eth {
...
+              EthImpl.ethSendRawTransaction,
+              requestDetails,
+              originalCallerAddress,
+              networkGasPriceInWeiBars,
+              await this.getCurrentNetworkExchangeRateInCents(requestDetails),
+            ),
+        {
+          canRetry: (e: unknown) => {
+            return e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded());
+          },
```
victor-yanev, line 1584: To make sure we don't duplicate transactions when a timeout occurs, we can get the `transactionId` from the `SDKClientError`: https://github.com/hashgraph/hedera-sdk-js/blob/15cb9c91209dd68667e138f191b606c04d73a8d6/src/transaction/TransactionResponse.js#L87 and poll the mirror node to confirm the transaction has actually failed before retrying:

**MirrorNodeClient.ts:**
```typescript
public async isFailedTransaction(transactionId: string, requestDetails: RequestDetails): Promise<boolean> {
  const transactionDetails = await this.repeatedRequest(
    this.getTransactionById.name,
    [transactionId, requestDetails],
    this.MIRROR_NODE_REQUEST_RETRY_COUNT,
    requestDetails,
  );
  return transactionDetails[0]?.result !== 'SUCCESS';
}
```
**eth.ts**:
```suggestion
          canRetry: async (e: unknown) => {
            return e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded()) && await this.mirrorNodeClient.isFailedTransaction(e.transactionId);
          },
```

victor-yanev (commented):
LGTM, the following suggestion will make sure we don't do unnecessary repeats and avoids submitting duplicated transactions.

Also, please add unit tests which simulate a `'timeout exceeded'`/`'Connection dropped'` error from the SDK client and confirm that:
- the transaction is not retried when the mirror node returns the transaction with a `{ result: 'SUCCESS' }`
- the transaction is retried when the transaction is not found in the mirror node or a result different from `'SUCCESS'` is returned

packages/relay/src/lib/eth.ts
```diff
@@ -1633,6 +1646,47 @@ export class EthImpl implements Eth {
...
+      backOff = 500,
+      canRetry = (e: unknown) => true,
+      onError = (e: SDKClientError) => {},
+    }: {
+      maxAttempts?: number;
+      backOff?: number;
+      canRetry?: (error: unknown) => boolean;
+      onError?: (error: SDKClientError) => void;
+    },
+  ): Promise<T | undefined> {
```
ebadiere, line 1674: Given the function returns either a transaction or throws an error, is the `undefined` generic necessary?

packages/relay/src/lib/eth.ts
```diff
@@ -1633,6 +1646,47 @@ export class EthImpl implements Eth {
     }
   }
 
+  /**
```
ebadiere, line 1649: Nit. Do we want to document why we are doing this?  I am asking because the `maxAttempts` and `backOff` values are hard coded.  I believe this is a workaround until we get the requested change from the SDK team.

ebadiere (commented):
Good work.  A couple of questions.

packages/relay/src/lib/eth.ts
konstantinabl: However, what you have linked is the ReceiptStatusError and in a timeout GrpcServiceError is thrown https://github.com/hashgraph/hedera-sdk-js/blob/main/src/Executable.js#L662
But I will find a way to get the transacion id anyways

packages/relay/src/lib/eth.ts
konstantinabl: Well the reason is that the return statement is in the try, so if this fails we throw an error, but do not return anything and in this case typescript is warning us that it is possible for the return type to be undefined, thus we add this. Just trying to make TS happy :)

packages/relay/src/lib/eth.ts
konstantinabl: What do you mean by document them @ebadiere

ebadiere: Is this not a workaround until we get requested changes from the SDK team?  Perhaps a comment stating that might be appropriate, because the `backOff`, `maxAttempts` and `retry`  values are hard coded.

packages/relay/src/lib/eth.ts
```diff
@@ -1567,16 +1566,34 @@ export class EthImpl implements Eth {
...
-        .getSDKClient()
-        .submitEthereumTransaction(
-          transactionBuffer,
-          EthImpl.ethSendRawTransaction,
-          requestDetails,
-          originalCallerAddress,
-          networkGasPriceInWeiBars,
-          await this.getCurrentNetworkExchangeRateInCents(requestDetails),
-        );
+      const sendRawTransactionResult = await this.sendRawTransactionWithRetry(
```
Nana-EC, line 1569: We should not be retrying `eth_sendRawTransaction`.
We should leave users and clients to do that based on the initial attempt response.
Retrying can add unintended latency and cost implications.
The transaction either fails gracefully or succeeds.
Let's discuss.

Nana-EC (changes requested):
Nice change although I don't think we should be doing a retry.
Let's discuss

packages/relay/src/lib/eth.ts
konstantinabl: @Nana-EC we only retry if we get a timeout exceeded and the transaction is not found in the mirror node
On [this](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3161/files#diff-dd8c7990f7764960d05dcb4c1b5ae7eee48b425b1b83d6dcc7e7ccc442330ef0R1582) line you can check on what occasions we retry. Does this make sense?

Nana-EC: It does however that exact scenario highlights the scenarios I'm referring to.
For instance if the try logic now makes the E2E time over the timeout set by a cloud provider or the users client the experience wouldn't be great and it would spend unnecessary hbars and lead to the current confusion of a successful transaction on network but with a failure response.

packages/relay/src/lib/eth.ts
konstantinabl: Added a comment :)

packages/relay/src/lib/eth.ts
Nana-EC: I think you should pull the retry logic out.
Check for a timeout occurrence and check if the MN has the transactions within a reasonable period.
If the transaction is found the transaction was successful and we shouldn't fail, if it's not found then throw the original timeout error back.

packages/relay/src/lib/eth.ts
ebadiere: Okay.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -441,19 +459,59 @@ export class SDKClient {
...
+  async isFailedTransaction(requestDetails: RequestDetails, transactionId?: string): Promise<boolean> {
+    const retryCount = 5;
+    try {
+      const transaction = await this.mirrorNodeClient.repeatedRequest(
+        this.mirrorNodeClient.getTransactionById.name,
+        [transactionId, requestDetails],
+        retryCount,
+        requestDetails,
+      );
+      const isFailed = transaction !== null ? true : false;
```
ebadiere, line 504: Nit:  I think this can be simply:
` const isFailed = transaction !== null;`

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -151,6 +162,13 @@ export class SDKClient {
     this.hbarLimitService = hbarLimitService;
     this.maxChunks = Number(ConfigService.get('FILE_APPEND_MAX_CHUNKS')) || 20;
     this.fileAppendChunkSize = Number(ConfigService.get('FILE_APPEND_CHUNK_SIZE')) || 5120;
+    this.mirrorNodeClient = new MirrorNodeClient(
```
quiet-node, line 165: Ahh the classic dilemma of overlapping the mirror node and SDK client! This does go against our design of keeping these two separate. I’m sorry but I'd have to kindly object this implementation. Let's discuss tho! 

It looks like the mirror node was added here to poll for transactions if they time out. Since `sdkClient.submitEthereumTransaction()` already throws errors if there’s an issue with the transaction, and with the previous implementation (without this [try-catch](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3161/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR462-R484)), errors would eventually be captured in `EthImpl.sendRawTransaction()` at this [catch block here](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1614-L1623). Since the error would remain consistent across all levels, how about handling the `isFailedTransaction` logic in that block instead?

Handling this in the `EthImpl` class would be beneficial, as there’s already a `mirrorNodeClient` instance available, which helps us avoid the dilemma of overlapping the two clients.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -441,19 +459,59 @@ export class SDKClient {
...
-      ),
+      );
+    } catch (e: any) {
+      if (e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded())) {
+        if (!(await this.isFailedTransaction(requestDetails, e.transactionId))) {
+          txResponse = { transactionId: e.transactionId };
+        } else {
+          throw e;
+        }
+      }
```
quiet-node, line 479: This block doesn't sseem sufficient enough for handling errors. This [block here](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L723-L725) also throw a JsonRpcError, and this block doesn't catch that. Maybe the `else` on line 476 need to be 1 level out.

quiet-node (changes requested):
Nice changes but sorry gotta block. Please fine more details below.

quiet-node:
@konstantinabl Now that I’ve had a chance to delve into this “`Transaction execution returns a null value...`” error and to examine more closely how `sdkClient.executeTransaction()` handles errors, [this block](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L729-L732) caught my attention. It seems we only throw an error if it has the `WRONG_NONCE` error code. As I mentioned above, there are several Ethereum-related errors a transaction can encounter while still reaching the consensus node and being retrievable by the mirror node. These error codes may include [WRONG_NONCE](https://hashscan.io/testnet/transaction/1730409597.694986000), [INSUFFICIENT_TX_FEE](https://hashscan.io/testnet/transaction/1730409339.234582176), `CONTRACT_REVERT_EXECUTED`, and others. These errors fail the transaction but still reach the consensus node. Thus, the line [transactionResponse = await transaction.execute(this.clientMain);](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L711) still returns a non-null value assigned to `transactionResponse`.

In the catch block, [this part](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L730-L732) only throws an error if it has the `WRONG_NONCE` code. I initially designed this logic under the assumption that errors like `WRONG_NONCE` or `CONTRACT_REVERT_EXECUTED` were the main issues causing transactions to fail while still making their way to the consensus node, so `transactionResponse` wouldn’t be null and wouldn’t trigger [this block](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L739-L743).

However, this may be a regression and not entirely accurate, as it appears there are errors that cause transactions to fail and not reach the consensus node, i.e., the `transactionResponse` is null. I think it’s worth investigating which other errors can hit this catch block without making it to the consensus node. Once we better understand these errors, we can design a more robust error-handling solution instead of throwing an `INTERNAL_ERROR` as we currently do in [this line](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L739-L743).

I hope this clarifies things and stays within the scope of the `Transaction execution returns a null value...` error.

packages/relay/src/lib/clients/sdkClient.ts
konstantinabl: Two things on this concern:
1. This is supposed to be a workaround, so its a temporary fix. I was also reluctant to adding mirrorNodeClient instance here, however it is inevitable because of #2 below
2. In the submitEthereumTransaction we have the following
`return {
      fileId,
      txResponse: await this.executeTransaction(
        ethereumTransaction,
        callerName,
        interactingEntity,
        requestDetails,
        true,
        originalCallerAddress,
      ),
    };`
and as you can see if we fail on this.executeTransaction(...) as it happens now, fileId wont'be returned and in this edge case we actually need it. If the transaction is found in the MN we will get the transactionId, but we need the fileId in order to proceed

packages/relay/src/lib/clients/sdkClient.ts
konstantinabl: The ideas is to not catch all errors and scenarios, here we only want to check the MN for the transaction if we get a timeout exceeded, becuase that when a transaction may have been submitted but the SDK dropped the connection and we didnt get a response. Thats the issue descirbed by Nana on mainnet, we have a timeout exceeded but the trasnaction can be found on hashscan. The other types of errors are handled here https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1482

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -724,7 +782,7 @@ export class SDKClient {
         throw e;
       }
 
-      const sdkClientError = new SDKClientError(e, e.message);
+      const sdkClientError = new SDKClientError(e, e.message, transaction.transactionId?.toString());
```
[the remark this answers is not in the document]
konstantinabl: It does here -> https://github.com/hashgraph/hedera-json-rpc-relay/pull/3161/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR474

packages/relay/src/lib/clients/sdkClient.ts
konstantinabl: @ebadiere You are right, i just like to be explicit when doing conditionals. But will change it, maybe its better

konstantinabl:
> @konstantinabl Now that I’ve had a chance to delve into this “`Transaction execution returns a null value...`” error and to examine more closely how `sdkClient.executeTransaction()` handles errors, [this block](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L729-L732) caught my attention. It seems we only throw an error if it has the `WRONG_NONCE` error code. As I mentioned above, there are several Ethereum-related errors a transaction can encounter while still reaching the consensus node and being retrievable by the mirror node. These error codes may include [WRONG_NONCE](https://hashscan.io/testnet/transaction/1730409597.694986000), [INSUFFICIENT_TX_FEE](https://hashscan.io/testnet/transaction/1730409339.234582176), `CONTRACT_REVERT_EXECUTED`, and others. These errors fail the transaction but still reach the consensus node. Thus, the line [transactionResponse = await transaction.execute(this.clientMain);](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L711) still returns a non-null value assigned to `transactionResponse`.
> 
> In the catch block, [this part](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L730-L732) only throws an error if it has the `WRONG_NONCE` code. I initially designed this logic under the assumption that errors like `WRONG_NONCE` or `CONTRACT_REVERT_EXECUTED` were the main issues causing transactions to fail while still making their way to the consensus node, so `transactionResponse` wouldn’t be null and wouldn’t trigger [this block](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L739-L743).
> 
> However, this may be a regression and not entirely accurate, as it appears there are errors that cause transactions to fail and not reach the consensus node, i.e., the `transactionResponse` is null. I think it’s worth investigating which other errors can hit this catch block without making it to the consensus node. Once we better understand these errors, we can design a more robust error-handling solution instead of throwing an `INTERNAL_ERROR` as we currently do in [this line](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L739-L743).
> 
> I hope this clarifies things and stays within the scope of the `Transaction execution returns a null value...` error.

@quiet-node Not quite sure what the relation to this PR is, but I believe we also handle these error here -> https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1482

packages/relay/src/lib/eth.ts
```diff
@@ -34,7 +34,7 @@ import { Transaction as EthersTransaction } from 'ethers';
 import HAPIService from './services/hapiService/hapiService';
 import { JsonRpcError, predefined } from './errors/JsonRpcError';
 import { Block, Log, Transaction, Transaction1559 } from './model';
-import { FileId, Hbar, PrecheckStatusError } from '@hashgraph/sdk';
+import { FileId, Hbar, PrecheckStatusError, TransactionId } from '@hashgraph/sdk';
```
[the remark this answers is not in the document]
konstantinabl: Correct. Removed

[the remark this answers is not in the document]
konstantinabl: Correct. Removed

packages/relay/src/lib/clients/sdkClient.ts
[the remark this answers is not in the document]
quiet-node: > It does here -> https://github.com/hashgraph/hedera-json-rpc-relay/pull/3161/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR474

The `e` in that line does not appear to be the same instance as the `sdkClientError` here. The `e` was caught when `executeTransaction()` throws an error, but in this case, the error is actually thrown [here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3161/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR781-R783) and [here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3161/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR788-R790) (WRONG_NONCE case). Therefore, the `e` in the line you mentioned would not have a reference to the `transactionId` as the `sdkClientError` instance does here.

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: This still doesn’t seem to handle `JsonRpcError` correctly. If the error thrown back is an instance of `SDKClientError`, it will remain in this catch block and won’t be thrown again.

packages/relay/src/lib/clients/sdkClient.ts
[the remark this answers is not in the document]
konstantinabl: You are correct. Fixed

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: **Note:** Synced offline and agreed this is a temporary workaround. I still wanted to document my suggestion here for reference.

Still thinking up a solution that makes sense and avoids using the `mirror node client` within the `sdkClient`. It seems that when `executeTransaction()` throws an error, we're primarily concerned with the condition `e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded())`. Given this, [this block](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3161/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR473-R482) could be modified so the `transactionId` could be directly returned to `eth.sendRawTransaction()` without involving mirror node polling.

```diff
      if (e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded())) {
-        const isFailed = await this.isFailedTransaction(requestDetails, e.transactionId);
-        if (isFailed) {
-          throw e;
-        }
+        // Commenting out the polling block
+        // const isFailed = await this.isFailedTransaction(requestDetails, e.transactionId);
+        // if (isFailed) {
+        //  throw e;
+        // }
+      // Focusing only on transactionId if the error is due to timeout or connection drop
        txResponse = { transactionId: e.transactionId };
      } else {
        throw e;
      }
    }
```

The `transactionId` would then be relayed to `EthImpl.sendRawTransaction()` and used at [this line](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1586). Finally, the polling logic can be applied in [this block here](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1594-L1599). This approach would allow us to handle timeout errors while reusing the existing polling logic in `sendRawTransaction`.

Note that this is a top-of-mind suggestion, and I haven't tested it. If you'd like to try it, please test and adjust the code as needed. 

Also, this is purely for reference. I understand the urgency and there’s no hard pressure to implement this immediately. However, if this approach might help streamline the work, please consider it.

quiet-node (approved):
Approve to unblock.

ebadiere (approved):

packages/relay/src/lib/clients/sdkClient.ts
konstantinabl: Will followup in another comment

Nana-EC: @quiet-node is right and I prefer his suggestion.
Let's follow up in a separate PR in this sprint to address that if possible.
If it requires a significant refactor we can explore in a different sprint.
@konstantinabl please note the ticket to address this here, thanks.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -441,19 +459,61 @@ export class SDKClient {
...
         callerName,
         interactingEntity,
         requestDetails,
         true,
         originalCallerAddress,
-      ),
+      );
+    } catch (e: any) {
+      if (e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded())) {
+        const isFailed = await this.isFailedTransaction(requestDetails, e.transactionId);
```
Nana-EC, line 474: An alternative approach is to simply call the `await this.mirrorNodeClient.repeatedRequest(...)` here instead. 
It makes the logic a positive approach and might avoid the need for an extra method in here.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -737,9 +797,13 @@ export class SDKClient {
       );
 
       if (!transactionResponse) {
-        throw predefined.INTERNAL_ERROR(
-          `${requestDetails.formattedRequestId} Transaction execution returns a null value: transactionId=${transaction.transactionId}, callerName=${callerName}, txConstructorName=${txConstructorName}`,
-        );
+        if (sdkClientError.isConnectionDropped() || sdkClientError.isTimeoutExceeded()) {
```
Nana-EC, line 800: Q: do we need to check on the type of `sdkClientError`?
Is the point to just return an `sdkClientError` if applicable and a general error otherwise.
If so you only need to check for a valid `sdkClientError` object

Nana-EC (approved):
Nice improvements. Left some comments.
I'm unblocking as the transaction is no longer being retried

packages/relay/src/lib/clients/sdkClient.ts
konstantinabl: @Nana-EC @quiet-node here is the issue https://github.com/hashgraph/hedera-json-rpc-relay/issues/3207

packages/relay/src/lib/clients/sdkClient.ts
konstantinabl: @Nana-EC Extracting the logic in a separate method and naming it properly helps the code to be more readable in my opinion. Now if someone checks this code they know why we call the mirror node, namely to check if this transaction has failed. Thats my motivataion for this, tell me if you don't agree.

packages/relay/src/lib/clients/sdkClient.ts
konstantinabl: My idea here was to check the message of the error via these methods. So we want to throw an SDKClientError and catch this particular error in the upper level method, and we want to do this only for a timeout exceeded or connection dropped, thats when we observed a trnasaciton being successful but the SDK client returns an error

packages/relay/src/lib/clients/sdkClient.ts
konstantinabl: @Nana-EC  Extracting the logic in a separate method and naming it properly helps the code to be more readable in my opinion. Now if someone checks this code they know why we call the mirror node, namely to check if this transaction has failed. Thats my motivataion for this, tell me if you don't agree.

Nana-EC: Sounds good

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC: Okay

Nana-EC:
@konstantinabl we should fix the title of the PR as no retry logic was added

PR #3173, opened 2024-10-29
chore: Adds eslint rules in eslintrc

### LINKED ISSUE #3064: Add eslint to be run pre-commit
### Problem

Currently, we have eslint but it is not actually used when committing and thus not enforcing any useful rules like for example sort-imports, which leads to different PRs rearranging imports and flooding PRs with unneeded changes. On top of this, eslint offers many other benefits that ensure code quality and consistency.

### Solution

Add eslint to the husky pre-commit script

### Alternatives

_No response_

konstantinabl:
**Description**:

This PR adds relevant rules to the .eslintrc.js file, adds lint script to package.json in different packages and adds eslint to husky hook
This is much needed to ensure good practices are followed, as well as styling and formatting of the code

**Related issue(s)**:

Fixes #3064


.eslintrc.js
```diff
@@ -49,9 +50,20 @@ module.exports = {
         "@typescript-eslint/ban-ts-comment": "off",
         "@typescript-eslint/no-empty-interface": "off",
         "@typescript-eslint/no-explicit-any": "off",
-        "@typescript-eslint/no-var-requires": "off",
         "@typescript-eslint/semi": "off",
+        "@typescript-eslint/no-var-requires": "warn",
```
Nana-EC, line 54: Q: is there a standard we're following?
Styling is always relative just curious what are the baselines and preferences

Nana-EC (approved):
Nice

quiet-node (approved):
Wonderful! Thanks for doing this! 

Q: So now will it automatically lint the whole codebase or do I still have to manually run the `lint` command?

.eslintrc.js
konstantinabl: I have left what was added before in this file and just added a few more rules, like no-trailing-spaces and prefer-const which are generally recommended. Above in the extended property you can see we support what is recommended by defaul and we turn off some of those rules for our projects, like semicolons, i am open to discussion on what should be on and off

konstantinabl:
> Wonderful! Thanks for doing this!
> 
> Q: So now will it automatically lint the whole codebase or do I still have to manually run the `lint` command?

@quiet-node yes, currently its not automatic yet, i will do the eslint --fix in a separate pr together with the hook that will fix linting issues before each commit. Its better to do it this way in order to not flood this PR with changes

Issue #3182, opened 2024-10-30
ci (bug): issue with pushing docker image


andrewb1269hg:
There is a potentially wrong token name in at least this file, possibly other files as well:
`.github/workflows/release-production.yml`

PR #3124, opened 2024-10-18
ci: Updaiting workflow permissions and adding hardend runner per step-security.

san-est:
**Description**:
Updating several workflows and their permissions per step-security requirements and adding hardend runner to one of the workflows.

**Related issue(s)**:

Fixes #3121 

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #3183, opened 2024-10-30
ci(bug): Update permissions in JSON rpc relay for GITHUB_TOKEN

rbarkerSL:
**Description**:
Added `packages: write` permission per [the github documentation](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic) May need to modify other files modified by #3124

**Related Issue(s)**:
Relates to #3124
Relates to #3182

### DIFF
--- a/.github/workflows/release-production.yml
+++ b/.github/workflows/release-production.yml
@@ -11,6 +11,7 @@ env:
 
 permissions:
   contents: read
+  packages: write
 
 jobs:
   docker-image-publish:


ebadiere (approved):
LGTM

quiet-node (approved):

.github/workflows/release-production.yml
```diff
@@ -11,6 +11,7 @@ env:
 
 permissions:
   contents: read
+  packages: write
```
Nana-EC, line 14: Q: Shouldn't this be the only change?
Not sure why there are version changes of dependencies included in the PR also.
Can those be removed?

Nana-EC (commented):
Change seems right but additional dependency changes seem out of place.

quiet-node (approved):

Nana-EC (approved):
LG

PR #3184, opened 2024-10-30
fix: 3152 spendingplan as env var acceptance test fix

### LINKED ISSUE #3152: Add the ability for spending plans to be defined using environment variables.

ebadiere:
This PR is a branch of PR#3153 with the HBar Rate Limiter fix in the acceptance tests.  Using a new branch because there is odd behavior in github ATM.

**Related issue(s)**:

Fixes #3152


ebadiere:
Covered by #3201

PR #3189, opened 2024-10-31
fix: stabilized HBAR Rate Limit acceptance tests

### LINKED ISSUE #3188: HbarLimiter tests are very flaky
### Problem

With the new implementation of `HbarLimitService` enabled, the HBAR Limiter test has become highly flaky. Since adding expenses and updating spending plans are executed asynchronously, and the suite includes exhaustion tests with numerous consecutive calls to the network, the test client sometimes fails to retrieve the correct amounts. The tests are flaky to the extent that they either fail or only pass after multiple retries.

### Alternatives

_No response_

quiet-node:
**Description**:
With the new implementation of `HbarLimitService` enabled, the HBAR Limiter test has become highly flaky. Since adding expenses and updating spending plans are executed asynchronously, and the suite includes exhaustion tests with numerous consecutive calls to the network, the test client sometimes fails to retrieve the correct amounts. 

The tests are flaky to the extent that they either fail or only pass after multiple retries. To address this, this PR includes measures to stabilize the tests by:
    - Adding polling methods to verify the remaining HBARs and amounts spent by `hbarSpendingPlans`.
    - Increasing tolerance in `verifyRemainingLimit()`. 
    - Adding a 1500ms wait time after each test case to allow asynchronous expense updates to complete in the background.
    - Introducing a `TEST_TRANSACTION_RECORD_COST_TOLERANCE` configuration variable for tolerance, allowing for reuse throughout the test and facilitating easy adjustments without requiring code changes.

**Related issue(s)**:

Fixes #3188

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -220,36 +220,49 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
-      beforeEach(async function () {
-        const basicPlans = await hbarSpendingPlanRepository.findAllActiveBySubscriptionTier(
-          [SubscriptionTier.BASIC],
-          requestDetails,
-        );
-        for (const plan of basicPlans) {
-          await hbarSpendingPlanRepository.delete(plan.id, requestDetails);
-          await ethAddressSpendingPlanRepository.deleteAllByPlanId(plan.id, 'before', requestDetails);
-          await ipSpendingPlanRepository.deleteAllByPlanId(plan.id, 'before', requestDetails);
-        }
```
quiet-node, line 232: **Note for reviewer:** 
Moved this logc to `BASIC Tier` so it can only be executed within the `BASIC Tier` scope. This change was made as it impacts the tests for `multiple tiered users`.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -103,7 +103,7 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
     };
 
     const verifyRemainingLimit = (expectedCost: number, remainingHbarsBefore: number, remainingHbarsAfter: number) => {
-      const delta = 0.02 * expectedCost;
+      const delta = 0.05 * expectedCost;
```
quiet-node, line 106: **Note for Reviewers:** 
Bumped tolerance to fix flaky test

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -220,36 +220,49 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
-        it('should execute "eth_sendRawTransaction" without triggering HBAR rate limit exceeded', async function () {
-          const parentContract = await deployContract(parentContractJson, accounts[0].wallet);
+        const pollForProperRemainingHbar = async (initialRemainingHbars: number, expectedTxCost: number) => {
+          let updatedRemainingHbars = Number(await metrics.get(testConstants.METRICS.REMAINING_HBAR_LIMIT));
+
+          // Note: expectedTxCost may be retrieved from mirror node which doesn't include the getRecord transaction fee.
+          //       calculating delta = expectedTxCost * tolerance to account for this difference in transaction costs.
+          const delta = expectedTxCost * 0.002;
+
+          while (initialRemainingHbars - updatedRemainingHbars > expectedTxCost + delta) {
```
quiet-node, line 240: **Note for reviewer:** 
- `initialRemainingHbars`: the remaining HBARs retrieved at the begining of the test
- `updatedRemainingHbarsAfter`: the remaining HBARs retrieved AFTER making calls to the network
- `expectedTxCost`: expected total transacction cost retrieved from mirror node
- `delta`: `expectedTxCost` may be retrieved from mirror node which doesn't include the transaction fee from getRecord queries. Delta is calculated to account for this difference in transaction costs.

Ideally, after a call to the network, `initialRemainingHbars - updatedRemainingHbarsAfter` should be greater than `expectedTxCost + delta`. However, the test client sometimes fails to retrieve the correct amounts, meaning `updatedRemainingHbarsAfter` could be fetched slightly too early, before the total burnt HBAR amount has fully updated in the metrics. In such cases, continue polling until the ideal amount is retrieved.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -366,7 +371,74 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+          expect(plan.planId).to.eq(hbarSpendingPlan.id);
+          const spendingPlan = await hbarSpendingPlanRepository.findByIdWithDetails(plan.planId, requestDetails);
+          expect(spendingPlan.active).to.be.true;
+          expect(spendingPlan.amountSpent).to.eq(0);
+          expect(spendingPlan.subscriptionTier).to.eq(subscriptionTier);
+
+          return { aliasAccount, hbarSpendingPlan };
+        };
+
+        const pollForProperAmountSpent = async (
```
quiet-node, line 402: **Note for reviewers:**
Same idea with pollForProperRemainingHbar() but for hbarSpendingPlans

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -449,210 +528,170 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
-            const plan = await ethAddressSpendingPlanRepository.findByAddress(aliasAccount.address, requestDetails);
-            expect(plan.ethAddress).to.eq(aliasAccount.address);
-            expect(plan.planId).to.eq(hbarSpendingPlan.id);
-            const spendingPlan = await hbarSpendingPlanRepository.findByIdWithDetails(plan.planId, requestDetails);
-            expect(spendingPlan.active).to.be.true;
-            expect(spendingPlan.amountSpent).to.eq(0);
-            expect(spendingPlan.subscriptionTier).to.eq(subscriptionTier);
-
-            return { aliasAccount, hbarSpendingPlan };
-          };
```
quiet-node, line 535: **Note for reviewer:** 
This function was moved to a more general location and renamed `createAliasAndAssociateSpendingPlan()` to enable reuse in other tests.

packages/server/tests/localAcceptance.env
```diff
@@ -27,7 +27,7 @@ LIMIT_DURATION=90000
 SERVER_REQUEST_TIMEOUT_MS=60000
 MEMWATCH_ENABLED=true
 WRITE_SNAPSHOT_ON_MEMORY_LEAK=false
-HBAR_RATE_LIMIT_TINYBAR=5000000000# 50 HBARs
-HBAR_RATE_LIMIT_DURATION=80000# 80 seconds
-HBAR_RATE_LIMIT_BASIC=4000000000# 40 HBARs
+HBAR_RATE_LIMIT_TINYBAR=9000000000# 90 HBARs
+HBAR_RATE_LIMIT_DURATION=86400000# 24 hours
+HBAR_RATE_LIMIT_BASIC=6000000000# 60 HBARs
```
quiet-node, line 32: **Note for Reviewers:** 
- The values of HBAR_RATE_LIMIT_TINYBAR has been increased due to the addition of more exhaustion tests in the hbar limiter. 

- Increasing HBAR_RATE_LIMIT_DURATION prevents the limiter from resetting during tests.

- Increasing HBAR_RATE_LIMIT_BASIC prevents api_batch1 from becoming flaky and encountering HBAR rate limit errors.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -449,210 +528,170 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
-                  expectedPlan.ipAddresses.forEach(async (ipAddress) => {
-                    const associatedPlanByIpAddress = await ipSpendingPlanRepository.findByAddress(
-                      ipAddress,
-                      requestDetails,
-                    );
-                    expect(associatedPlanByIpAddress.planId).to.eq(expectedPlan.id);
-                    expect(associatedPlanByIpAddress.ipAddress).to.eq(ipAddress);
-                  });
-                }
+            if (configPath) {
```
quiet-node, line 657: **Note for Reviewers:** 
Just simply add a check to run this test suite only if configPath is not null

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -449,210 +528,170 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
-                //    a. if remainingHbarsBefore > maxSpendingLimit ===> (totalHbarSpentByPrivilegedPlan + estimatedTxFee) > maxSpendingLimit
-                //    b. if remainingHbarsBefore <= maxSpendingLimit ===> remainingBudget < estimatedTxFee
+                //    a. if remainingHbarsBefore > maxSpendingLimit ===> (totalHbarSpentByPrivilegedPlan + expectedTxCost) > maxSpendingLimit
+                //    b. if remainingHbarsBefore <= maxSpendingLimit ===> remainingBudget < expectedTxCost
                 if (remainingHbarsBefore > maxSpendingLimit) {
-                  expect(amountSpent + estimatedTxFee).to.be.gt(maxSpendingLimit);
+                  expect(amountSpent + expectedTxCost).to.be.gt(maxSpendingLimit);
                 } else {
-                  expect(remainingHbarsAfter).to.be.lt(estimatedTxFee);
+                  expect(remainingHbarsAfter).to.be.lt(expectedTxCost);
```
quiet-node, line 643: **Note for reviewers:**
The variable expectedTxCost has been renamed around just simply for the sake of consistency.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -449,210 +528,170 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
-                  await deployContract(largeContractJson, aliasAccount.wallet);
+                  const tx = await deployContract(largeContractJson, aliasAccount.wallet);
+                  await tx.waitForDeployment();
+
+                  expectedTxCost ||= await getExpectedCostOfLastLargeTx(largeContractJson.bytecode);
+                  await pollForProperAmountSpent(hbarSpendingPlan, deploymentCounts + 1, expectedTxCost);
                 }
                 expect.fail(`Expected an error but nothing was thrown`);
               } catch (e: any) {
+                logger.error(e.message);
```
quiet-node, line 628: **Note for reviewers:**
Enhanced error logging for improved debugging.

packages/server/tests/acceptance/hbarLimiter.spec.ts
Nana-EC: Something like this could be a TEST_Config to allow tweaking without code changes

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -220,36 +220,49 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
-        );
-        for (const plan of basicPlans) {
-          await hbarSpendingPlanRepository.delete(plan.id, requestDetails);
-          await ethAddressSpendingPlanRepository.deleteAllByPlanId(plan.id, 'before', requestDetails);
-          await ipSpendingPlanRepository.deleteAllByPlanId(plan.id, 'before', requestDetails);
-        }
+      afterEach(async () => {
+        hbarSpendingPlanRepository.resetAmountSpentOfAllPlans(requestDetails);
+
+        // Note: Since the total HBAR budget is shared across the entire Relay instance by multiple test cases,
```
Nana-EC, line 226: Q: Are we running multiple instances in acceptance tests or are you referring to parallel runs?
If parallel aren't those instances separate from each other?

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -220,36 +220,49 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+        //       and expense updates occur asynchronously, the wait below ensures that the HBAR amount has sufficient time
+        //       to update properly after each test.
+        await Utils.wait(1500);
```
Nana-EC, line 229: nit: as usual I'm not a fan of waits.
Hmm, this means that we're going to add 1.5 * number of tests to this execution run time right?
I guess if it's only for this class it's acceptable right now to unblock but we should come back at some point to make this more stable without a wait

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -220,36 +220,49 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
 
       describe('@hbarlimiter-batch1 Total HBAR Limit', () => {
-        it('should execute "eth_sendRawTransaction" without triggering HBAR rate limit exceeded', async function () {
-          const parentContract = await deployContract(parentContractJson, accounts[0].wallet);
+        const pollForProperRemainingHbar = async (initialRemainingHbars: number, expectedTxCost: number) => {
+          let updatedRemainingHbars = Number(await metrics.get(testConstants.METRICS.REMAINING_HBAR_LIMIT));
+
+          // Note: expectedTxCost may be retrieved from mirror node which doesn't include the getRecord transaction fee.
+          //       calculating delta = expectedTxCost * tolerance to account for this difference in transaction costs.
+          const delta = expectedTxCost * 0.002;
```
Nana-EC, line 238: Q: why 0.002 and not some other number. It's unclear also could the delta from before be reused?
If not, same thing a targeted config might allow for easy troubleshooting and testing to find the right value without code changes

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -220,36 +220,49 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+
+          while (initialRemainingHbars - updatedRemainingHbars > expectedTxCost + delta) {
+            logger.warn(
+              `Fail to retrieve proper updated remaining HBARs. Polling for the proper updated remaining HBARs: expectedTxCost=${expectedTxCost}, delta=${delta}, initialRemainingHbars=${initialRemainingHbars}, currentUpdatedRemainingHbars=${updatedRemainingHbars}, properUpdatedRemainingHbar=${
```
Nana-EC, line 242:
```suggestion
              `Failed to retrieve proper updated remaining HBARs. Polling for the proper updated remaining HBARs: expectedTxCost=${expectedTxCost}, delta=${delta}, initialRemainingHbars=${initialRemainingHbars}, currentUpdatedRemainingHbars=${updatedRemainingHbars}, properUpdatedRemainingHbar=${
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -220,36 +220,49 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+                initialRemainingHbars - expectedTxCost - delta
+              }`,
+            );
+            await Utils.wait(1000);
+            updatedRemainingHbars = Number(await metrics.get(testConstants.METRICS.REMAINING_HBAR_LIMIT));
+          }
+
+          logger.info(
+            `Successfully retrieve proper updated remaining HBARs: expectedTxCost=${expectedTxCost}, delta=${delta}, initialRemainingHbars=${initialRemainingHbars}, currentUpdatedRemainingHbars=${updatedRemainingHbars}, properUpdatedRemainingHbar=${
```
Nana-EC, line 251: nit
```suggestion
            `Successfully retrieved proper updated remaining HBARs: expectedTxCost=${expectedTxCost}, delta=${delta}, initialRemainingHbars=${initialRemainingHbars}, currentUpdatedRemainingHbars=${updatedRemainingHbars}, properUpdatedRemainingHbar=${
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
Nana-EC, line 628: Q: why `console` vs `logger`?

Nana-EC (commented):
Some comments.
I understand the unblocking driver but let's capture all these needed waits and come back in the future to remove them as this likely just pushes the issue down the road

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: Ah we only run one instance of the Relay in the suite. The issue is that the tests in the suite share and depend on the total HBAR budget controlled by the `HbarLimitService` class. All expense additions occur asynchronously by emitting an event [like this](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L747-L755) and catching the event in the `MetricService` class [like this](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/services/metricService/metricService.ts), which happens asynchronously in the background to avoid unnecessary wait time.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: Understood! I agree we should comeback to make it stable without waits. Will create a ticket for it.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: You're right this is indeed unclear! After reviewing logs, I observed that the fee for getting a record is either 83,333 tinybars or 9 tinybars. I then calculated by dividing 83,333 by the `expectedTxCost`—which can vary over time—and arrived at approximately 0.0015, which I rounded up to 0.002 for better flexibility. That’s how 0.002 was determined.

Will update it to use a targeted configuration for improved debugging.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: updated

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: updated.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: Ah yeah I used `logger` first for consistency as the whole suite uses it. The second and main reason is to maintain asynchronous order. With `console`, logs appear immediately upon execution, but with `logger`, logs wait for other `logger`-initiated entries. This creates a more organized order, making it easier to navigate through logs during debugging.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: Yeah I agree. Pushed in an update which adds TEST_TRANSACTION_RECORD_COST_TOLERANCE config variable which can be reused through out the test.

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: Created a ticket for it https://github.com/hashgraph/hedera-json-rpc-relay/issues/3199

packages/server/tests/acceptance/hbarLimiter.spec.ts
quiet-node: Updated. Added TEST_TRANSACTION_RECORD_COST_TOLERANCE config variable which can be reused through out the test.

Nana-EC (approved):

docs/configuration.md
```diff
@@ -182,13 +182,14 @@ See [`.env.previewnet.sample`](./examples/.env.previewnet.sample).
...
+| Name                                     | Default | Description                                                                                        |
+| ---------------------------------------- | ------- | -------------------------------------------------------------------------------------------------- |
+| `LOCAL_NODE`                             | ""      | Flag if relay is hosted in the Hedera local node setup.                                            |
+| `E2E_RELAY_HOST`                         | ""      | Remote relay url to point to.                                                                      |
+| `DEV_MODE`                               | "false" | Flag if relay should operate in developer optimization mode.                                       |
+| `TEST_WS_SERVER`                         | "false" | Flag config for enable or disable the WS server tests.                                             |
+| `TEST_GAS_PRICE_DEVIATION`               | 0.2     | Value to use as deviation when comparing gas prices in the rpc-batch1.spec.ts                      |
```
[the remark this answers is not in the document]
quiet-node: Are you suggesting some thing like GlobalConfig.TEST.config-name, instead of GlobalConfig.ENTRIES.config-name?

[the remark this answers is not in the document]
quiet-node: If so, I think it'd be better addressing that in the next PR by pulling all test-related configurations into that section and adding appropriate tests as well. Please let me know if that's the case you're talking about.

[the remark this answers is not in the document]
Nana-EC: NO i meant the table.
Just so as a user I don't read config and worry about test configs when i'm only concerned with running the product

packages/server/tests/acceptance/hbarLimiter.spec.ts
Nana-EC: Hmm, okay.
This is the impact of the async emission of events I guess

packages/server/tests/acceptance/hbarLimiter.spec.ts
Nana-EC: Yeah I've observed this also.
I wonder if this means other logs should be converted for easier troubleshooting

docs/configuration.md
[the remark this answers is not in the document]
quiet-node: Ah I see your point well however this table is actually located in the `#Testing` section and contains only test-related configurations, separate from the table containing the main configurations.

packages/config-service/src/services/globalConfig.ts
```diff
@@ -604,6 +604,12 @@ export class GlobalConfig {
       required: false,
       defaultValue: null,
     },
+    TEST_TRANSACTION_RECORD_COST_TOLERANCE: {
+      envName: 'TEST_TRANSACTION_RECORD_COST_TOLERANCE',
+      type: 'boolean',
```
ebadiere, line 609: Is this really a boolean?

ebadiere (commented):
One question.

packages/config-service/src/services/globalConfig.ts
quiet-node: Ah good catch! Updated.

Nana-EC (approved):

packages/server/tests/localAcceptance.env
konstantinabl: @quiet-node  I thought we can add specific values for the hbar limiter acceptance tests directly in the command which runs this suite. This way we will mimic the actual behaviour of the limiter, which is also the goal of acceptance tests. So instead of just bumping the values all the time, we can just restart the limiter maybe just set a higher duration but not such high values. wdyt?

quiet-node: @konstantinabl Thank you for your comment! I understand your concern. Currently, the duration is set to 24 hours by default, and the total limit is 8,000 HBARs. I hope this adjustment will be the final time we modify these variables. However, if you prefer, I can certainly revert it back to the command. Please let me know!

konstantinabl (approved):

PR #3198, opened 2024-10-31
fix: Capture the npm_package_version

### LINKED ISSUE #3197: The version from the npm package is not getting read by the config service.
### Description

Because of this the method `web3_clientVersion` is not returning the relay version.

### Steps to reproduce

Run the `web3_clientVersion` on release  `0.50.0-rc2`.

```
curl --location 'http://localhost:7546' \
--header 'Request-Id: testRequest' \
--data '{
    "id": "test_id",
    "jsonrpc": "2.0",
    "method": "web3_clientVersion",
    "params": []
}'
```

### Additional context

_No response_

### Hedera network

previewnet

### Version

v0.59.0-rc2

### Operating system

macOS

ebadiere:
**Description**:
The ConfigService was using an all caps version of the `npm_package_version` of the env var definition but in a node context the env var is lowercase: `npm_package_version`.  Nothing was being returned from the `web3_clientVersion` method.

**Related issue(s)**:

Fixes #3197

**Notes for reviewer**:
Simply changing the case of the env var definition in the `GlobalConfig.ts` class prevents updating conditional logic that already works.  Adding the version number to the root package.json file allows it to be loaded by the `ConfigService.ts` which loads before the acceptance tests framework.

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "root",
+  "version": "0.60.0-SNAPSHOT",
   "devDependencies": {
     "@hashgraph/hedera-local": "^2.31.0",
     "@open-rpc/schema-utils-js": "^1.16.1",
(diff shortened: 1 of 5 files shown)


quiet-node (approved):

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "root",
+  "version": "0.60.0-SNAPSHOT",
```
quiet-node, line 3: qq: Did it require this addition?

quiet-node (approved):

package.json
ebadiere: Yes, I updated the description, but in short, the `ConfigService` is loaded before the acceptance tests framework, and without a version number the acceptance tests won't run.  

I really think the application version number needs to be required in order to allow the `web_clientVersion` method to work.  Making it required means it needs to be available in the `ConfigService` before the acceptance test framework is loaded.

natanasow (commented):
LGTM

natanasow (dismissed):

natanasow:
@ebadiere could run `npm install` locally to fix the `package-lock.json` as well.

natanasow (approved):

quiet-node (approved):

PR #3200, opened 2024-10-31
fix: 3198 cherry pick npm package version

ebadiere:
**Description**:
Cherry-picked fix for the `web3_clientVersion` method.

The ConfigService was using an all caps version of the npm_package_version of the env var definition but in a node context the env var is lowercase: npm_package_version. Nothing was being returned from the web3_clientVersion method.

**Related issue(s)**:

Fixes #3198 

**Notes for reviewer**:
Simply changing the case of the env var definition in the GlobalConfig.ts class prevents updating conditional logic that already works. Adding the version number to the root package.json file allows it to be loaded by the ConfigService.ts which loads before the acceptance tests framework.

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "root",
+  "version": "0.60.0-SNAPSHOT",
   "devDependencies": {
     "@hashgraph/hedera-local": "^2.31.0",
     "@open-rpc/schema-utils-js": "^1.16.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "root",
+  "version": "0.59.0-rc2",
```
quiet-node, line 3: Forgot to ask in the earlier PR, did you make sure the .bump-version script will bump this version here?

quiet-node (approved):

package.json
ebadiere: This should get updated with the next RC build.  It's needed here to run the acceptance tests.

packages/config-service/src/services/globalConfig.ts
```diff
@@ -495,10 +495,11 @@ export class GlobalConfig {
       required: false,
       defaultValue: null,
     },
-    NPM_PACKAGE_VERSION: {
+    // the actual env var in the node process is npm_package_version
+    npm_package_version: {
```
Nana-EC, line 499: Q: is this highlighting a bug in the Config Service?
Or is this by design and an oversight in the initial implementation?

packages/config-service/src/services/globalConfig.ts
```diff
@@ -495,10 +495,11 @@ export class GlobalConfig {
...
       envName: 'npm_package_version',
       type: 'string',
-      required: false,
+      required: true,
```
Nana-EC, line 502: Q: does this affect the product logic or is this only required on the test side?
I see test changes for validation but it's not clear what bug was occurring and how we are confirming it won't happen again in the product side

ebadiere: The bug that was occurring was the `web3_clientVersion` method was returning nothing, so I would say it affects the product.  Checking and confirming the correct product version is deployed, is one of the first steps in the release process.    The fix simply uses the new `ConfigService` and forces the version to be defined in order to start the application.  The idea here is to not have this problem again, especially when cutting a release.  There should always be a version defined in the package.json files.   A test was added to ensure a product version can be read, so that should prevent this issue from ever happening again.

packages/config-service/src/services/globalConfig.ts
ebadiere: I would say it was a simple oversight in the initial implementation.  In a node context the environment variable is called `npm_package_version`. All lowercase.  By making this simple update no conditional logic, that was already tested and working, needed to change.

packages/config-service/src/services/globalConfig.ts
Nana-EC: Okay. I had expect to see the place where the error was happening (web3_clientVersion in this case) have a test but I guess the claim is with coverage in the `validationService.spec.ts` there's no need anywhere else since the ConfigService would be used in that case.

Nana-EC (approved):

PR #3210, opened 2024-11-05
fix: added null check for nullable values in formatContractResult()

### LINKED ISSUE #3209: formatContractResult() throws `Cannot read properties of null (reading 'substring')`
### Problem

In some cases, contract results passed to the `formatContractResult()` method contain null fields (most of the time it's the `from` field). The `formatContractResult()` method then calls functions on these null fields, which cause `Cannot read properties of null (reading 'substring')` errors

### Solution

Add a null check before calling functions on potential null values.

### Alternatives

_No response_

quiet-node:
**Description**:
This PR adds null checks for nullable values in `formatContractResult()` to ensure that, if null values are passed in, they are handled correctly rather than causing a `Cannot read properties of null (reading...)` error.

**Related issue(s)**:

Fixes #3209

**Notes for reviewer**:

**Checklist**


ebadiere (dismissed):
LG.

ebadiere (approved):

Nana-EC (approved):

Issue #3219, opened 2024-11-05
File created to hold large calldata for sendRawTransaction won't get deleted if an error is thrown


quiet-node:
### Problem

For `eth_sendRawTransaction`, if the calldata exceeds 5120 bytes, HFS is used to temporarily store the bytecode on the network. The intention is to delete the file created to hold the calldata regardless of whether the transaction succeeds or fails. However, the current implementation only deletes the file if no error is thrown; if an error occurs, the file deletion is skipped.

### Solution

Here are a couple of potential solutions:

- Attach the `fileId` to the error being thrown, allowing access to it directly from the error object.
- Use a cache to temporarily store the `fileId`, allowing it to be re-accessed as needed, with the record removed from the cache after deletion.

Ferparishuertas:
As soon as JumbTx appear this will be fully refacorized

Ferparishuertas:
[#3761](https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3761)

Issue #3202, opened 2024-11-01
[EPIC] Fast return on eth_sendRawTransaction


Nana-EC:
### Problem

Currently the eth_sendRawTransaction endpoint is plagued with timeouts from many different sources due to server, netowkr and client timeouts settings.
In the case of a large callData call eth_sendRawTransaction logic employs HFS to submit create and append file transctions to hold the callData temporarily.
This result in increased latency and a subpar experience for some users and operators.

### Solution

The team discussed the option of returning early on the transaction instead of waiting for the relay to sequentially handle all the orchestration.
If the transactions passes the needed relay precheck validations could it return a response to the user but in the background continue to submit the `EthereumTransaction` and or HFS transactions to the network?

Most EVM tools will utilize `eth_getTransactionRecipt` waiting to find the status of the transaction.
Could this be utilized to reduce latency and improve the DevEx and UX?

```[tasklist]
### Tasks
- [x] Research current approach by other products
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/3281
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/3270
```

quiet-node:
After conducting research and gathering insights from the team, I've compiled an overview of the typical processing flow for `eth_sendRawTransaction`. Below is an outline of how Ethereum generally handles this process.

### Standard Processing in Ethereum

When a client submits an `eth_sendRawTransaction` request, the JSON-RPC server immediately returns the Keccak-256 hash of the signed raw transaction. This response is sent without waiting for the transaction to be mined or fully processed, highlighting that all subsequent transaction handling steps occur asynchronously.

#### 1. Immediate Hash Return
The JSON-RPC server verifies the transaction’s signature, computes the transaction hash using Keccak-256, and promptly returns this hash to the client as a unique identifier. At this point, the server has not yet broadcasted or validated the transaction for mining.

#### 2. Asynchronous Processing
   - **Transaction Validation**: The transaction undergoes verification checks, including nonce validation, gas limit checks, and balance checks to ensure sufficient funds.
   - **Broadcasting**: Once validated, the transaction is broadcast across the Ethereum network.
   - **Mining**: Miners pick up the transaction from the pool, attempt to include it in a block, and engage in solving the proof-of-work problem. If mined successfully, the transaction is included in a block and confirmed by the network.

#### 3. Client Responsibilities
[...]

PR #3229, opened 2024-11-06
fix: modified sendRawTransaction to also poll MN for transaction's validity for Timeout and ConnectionDropped error

### LINKED ISSUE #3207: fix: remove usage of MN client in SDK client
Mirror node client was added to the SDK client for a temporary fix regarding timeout exceeded issue -> https://github.com/hashgraph/hedera-json-rpc-relay/issues/3118

Since it is not a good practice to introduce the MN client in the SDk Client, a more elegant solution has to be found

quiet-node:
**Description**:
- Modifies EthImpl.sendRawTransaction() to poll the MN for the transaction's validity not only for transactions that were successfully submitted to the network but also in cases of Timeout and ConnectionDropped errors.
- Removes the MN instance in SDKClient.
- Fixes unit tests in eth_sendRawTransaction.spec.ts to align with the new update.

**Related issue(s)**:

Fixes #3207

**Notes for reviewer**:
- The primary focus is on the `EthImpl.sendRawTransaction`. Please refer to the code comments for a clearer understanding of the rationale behind the changes.
- The sdkClient.ts was only modified to remove the MN instance and update a couple of log message and code comments

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -69,8 +69,6 @@ import {
   ITransactionRecordMetric,
   RequestDetails,
 } from '../types';
-import { MirrorNodeClient } from './mirrorNodeClient';
-import { Registry } from 'prom-client';
 
 const _ = require('lodash');
 
@@ -125,14 +123,6 @@ export class SDKClient {
   private readonly hbarLimitService: HbarLimitService;
 
   /**
-   * An instance of the MirrorNodeClient
-   * @private
-   * @readonly
-   * @type {MirrorNodeClient}
-   */
-  private readonly mirrorNodeClient: MirrorNodeClient;
-
-  /**
    * Constructs an instance of the SDKClient and initializes various services and settings.
    *
    * @param {Client} clientMain - The primary Hedera client instance used for executing transactions and queries.
@@ -146,7 +136,6 @@ export class SDKClient {
     cacheService: CacheService,
     eventEmitter: EventEmitter,
     hbarLimitService: HbarLimitService,
-    register: Registry,
   ) {
     this.clientMain = clientMain;
 
@@ -162,13 +151,6 @@ export class SDKClient {
     this.hbarLimitService = hbarLimitService;
     this.maxChunks = Number(ConfigService.get('FILE_APPEND_MAX_CHUNKS')) || 20;
     this.fileAppendChunkSize = Number(ConfigService.get('FILE_APPEND_CHUNK_SIZE')) || 5120;
-    this.mirrorNodeClient = new MirrorNodeClient(
-      // @ts-ignore
-      ConfigService.get('MIRROR_NODE_URL') || '',
-      logger,
-      register,
-      cacheService,
-    );
   }
 
   /**
@@ -459,59 +441,17 @@ export class SDKClient {
       Hbar.fromTinybars(Math.floor(networkGasPriceInTinyBars * constants.MAX_GAS_PER_SEC)),
     );
 
-    let txResponse;
-    try {
-      txResponse = await this.executeTransaction(
+    return {
+      fileId,
+      txResponse: await this.executeTransaction(
         ethereumTransaction,
         callerName,
         interactingEntity,
         requestDetails,
         true,
         originalCallerAddress,
-      );
-    } catch (e: any) {
-      if (e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded())) {
-        const isFailed = await this.isFailedTransaction(requestDetails, e.transactionId);
-        if (isFailed) {
-          throw e;
-        }
-        txResponse = { transactionId: e.transactionId };
-      } else {
-        throw e;
-      }
-    }
-
-    return {
-      fileId,
-      txResponse,
+      ),
     };
-  }
-
-  /**
-   * Checks if a transaction has failed by querying the mirror node.
-   *
-   * @param {RequestDetails} requestDetails - The details of the request.
-   * @param {string} [transactionId] - The ID of the transaction to check.
-   * @returns {Promise<boolean>} - A promise that resolves to `true` if the transaction has failed, `false` otherwise.
-   */
-  async isFailedTransaction(requestDetails: RequestDetails, transactionId?: string): Promise<boolean> {
-    const retryCount = 5;
-    try {
-      const transaction = await this.mirrorNodeClient.repeatedRequest(
-        this.mirrorNodeClient.getTransactionById.name,
-        [transactionId, requestDetails],
-        retryCount,
-        requestDetails,
-      );
-      const isFailed = transaction !== null;
-      return isFailed;
-    } catch (e: any) {
-      this.logger.error(
-        e,
-        `${requestDetails.formattedRequestId} Failed to check if transaction ${transactionId} is failed`,
-      );
-      return true;
-    }
   }
 
   /**
@@ -781,22 +721,28 @@ export class SDKClient {
       return transactionResponse;
     } catch (e: any) {
       if (e instanceof JsonRpcError) {
+        this.logger.warn(
+          e,
+          `${requestDetails.formattedRequestId} Failed to execute ${txConstructorName} transaction due to a JsonRpcError: transactionId=${transaction.transactionId}, callerName=${callerName}, status=${e.code} message=${e.message}`,
+        );
         throw e;
       }
 
       const sdkClientError = new SDKClientError(e, e.message, transaction.transactionId?.toString());
+      this.logger.warn(
+        sdkClientError,
+        `${requestDetails.formattedRequestId} Failed to execute ${txConstructorName} transaction due to an SDKClientError: transactionId=${transaction.transactionId}, callerName=${callerName}, status=${sdkClientError.status}(${sdkClientError.status._code}) message=${sdkClientError.message}`,
+      );
 
-      // Throw WRONG_NONCE error as more error handling logic for WRONG_NONCE is awaited in eth.sendRawTransactionErrorHandler().
+      // WRONG_NONCE is one of the special errors where the SDK still returns a valid transactionResponse.
+      // Throw the WRONG_NONCE error, as additional handling logic is expected in a higher layer.
       if (sdkClientError.status && sdkClientError.status === Status.WrongNonce) {
         throw sdkClientError;
       }
 
-      this.logger.warn(
-        sdkClientError,
-        `${requestDetails.formattedRequestId} Fail to execute ${txConstructorName} transaction: transactionId=${transaction.transactionId}, callerName=${callerName}, status=${sdkClientError.status}(${sdkClientError.status._code}) message=${sdkClientError.message}`,
-      );
-
       if (!transactionResponse) {
+        // Transactions may experience "SDK timeout exceeded" or "Connection Dropped" errors from the SDK, yet they may still be able to reach the consensus layer.
+        // Throw Connection Drop and Timeout errors as additional handling logic is expected in a higher layer.
         if (sdkClientError.isConnectionDropped() || sdkClientError.isTimeoutExceeded()) {
           throw sdkClientError;
         } else {
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -721,18 +721,10 @@ export class SDKClient {
       return transactionResponse;
     } catch (e: any) {
       if (e instanceof JsonRpcError) {
-        this.logger.warn(
-          e,
-          `${requestDetails.formattedRequestId} Failed to execute ${txConstructorName} transaction due to a JsonRpcError: transactionId=${transaction.transactionId}, callerName=${callerName}, status=${e.code} message=${e.message}`,
-        );
         throw e;
       }
 
       const sdkClientError = new SDKClientError(e, e.message, transaction.transactionId?.toString());
-      this.logger.warn(
-        sdkClientError,
-        `${requestDetails.formattedRequestId} Failed to execute ${txConstructorName} transaction due to an SDKClientError: transactionId=${transaction.transactionId}, callerName=${callerName}, status=${sdkClientError.status}(${sdkClientError.status._code}) message=${sdkClientError.message}`,
-      );
 
       // WRONG_NONCE is one of the special errors where the SDK still returns a valid transactionResponse.
       // Throw the WRONG_NONCE error, as additional handling logic is expected in a higher layer.
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1585,9 +1585,6 @@ export class EthImpl implements Eth {
       fileId = sendRawTransactionResult.fileId;
       submittedTransactionId = sendRawTransactionResult.txResponse.transactionId?.toString();
       if (!constants.TRANSACTION_ID_REGEX.test(submittedTransactionId)) {
-        this.logger.warn(
-          `${requestIdPrefix} Transaction successfully submitted but returned invalid transactionID: transactionId==${submittedTransactionId}`,
-        );
         throw predefined.INTERNAL_ERROR(
           `Transaction successfully submitted but returned invalid transactionID: transactionId==${submittedTransactionId}`,
         );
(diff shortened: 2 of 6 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1630,6 +1613,61 @@ export class EthImpl implements Eth {
...
+          );
+        }
+
+        if (contractResult.hash == null) {
+          this.logger.error(
+            `${requestIdPrefix} Transaction returned a null transaction hash: transactionId=${submittedTransactionId}`,
+          );
+          throw predefined.INTERNAL_ERROR(
+            `Transaction returned a null transaction hash: transactionId=${submittedTransactionId}`,
+          );
```
victor-yanev, line 1653: missing test case

packages/relay/src/lib/eth.ts
```diff
@@ -1630,6 +1613,61 @@ export class EthImpl implements Eth {
...
+          }
+
+          this.logger.warn(
+            `${requestIdPrefix} No matching transaction record retrieved: transactionId=${submittedTransactionId}`,
+          );
+
+          throw predefined.INTERNAL_ERROR(
+            `No matching transaction record retrieved: transactionId=${submittedTransactionId}`,
+          );
+        }
```
victor-yanev, line 1645: Imo, this logic should be inside the if-statement in the `catch` block on line 1597. We don't want to be slowing down all `eth_sendRawTransaction` requests with this polling, only when we get the "connection dropped" and "timeout exceeded" errors:

```typescript
} catch (e: any) {
  // new logic should be only here
  if (e instanceof SDKClientError && (e.isConnectionDropped() || e.isTimeoutExceeded())) {
    const formattedTransactionId = formatTransactionId(e.transactionId);
    const contractResult = await this.mirrorNodeClient.repeatedRequest(
       this.mirrorNodeClient.getContractResult.name,
       [formattedTransactionId, requestDetails],
       this.mirrorNodeClient.getMirrorNodeRequestRetryCount(),
       requestDetails,
     );
     if (!contractResult?.hash) {
       throw e;
     }
     return contractResult.hash;
  }
  // rest of the `catch` logic remains the same as in the `main` branch
}
// rest of the sendRawTransaction logic remains the same as in the `main` branch
```

victor-yanev (changes requested):
Good job, I like where this PR is headed but I have some suggestion on how we can improve it further

packages/relay/src/lib/eth.ts
konstantinabl: I agree with Victor and also currently the method is even more confusing to me. Maybe as stated above if we move the new logic to the catch it will be slightly better to read

packages/relay/src/lib/eth.ts
```diff
@@ -1578,46 +1583,24 @@ export class EthImpl implements Eth {
...
-
-      if (!contractResult) {
-        this.logger.warn(`${requestIdPrefix} No record retrieved`);
-        throw predefined.INTERNAL_ERROR(`No matching record found for transaction id ${txId}`);
-      }
-
-      if (contractResult.hash == null) {
-        this.logger.error(
-          `${requestIdPrefix} The ethereumHash can never be null for an ethereum transaction, and yet it was!!`,
+      fileId = sendRawTransactionResult.fileId;
```
konstantinabl, line 1586: @quiet-node So with this change won't the fileId be null when the _**executeTransaction**_ in the **_submitTransaction_** throws null? And this will result in the file not being deleted?

konstantinabl:
I like the idea to remove the MN client from the sdkClient, however I think we should only change whats in the catch block of the sendRawTransaction. Current changes are confusing and worsening the readability at least from my POV. We can maybe discuss and brainstorm a better way, where we only change the catch block, since we only want new logic when error is thrown.

packages/relay/src/lib/eth.ts
quiet-node: Yes, you're correct! However, the original implementation has a flaw: if an error is thrown, `fileId` does not get updated and thus eventually doesn't get deleted. The previous implementation didn't fully address this issue; it allowed Connection Dropped and Timeout errors to carry the created `fileId` but didn't account for other errors. I've noted this and captured it in this ticket https://github.com/hashgraph/hedera-json-rpc-relay/issues/3219, which we can triage and address in the next sprint.

quiet-node:
> @quiet-node I like the idea to remove the MN client from the sdkClient, however I think we should only change whats in the catch block of the sendRawTransaction. Current changes are confusing and worsening the readability at least from my POV. We can maybe discuss and brainstorm a better way, where we only change the catch block, since we only want new logic when error is thrown.

@konstantinabl 
Thanks for the concern yeah it’s a valid one! 
However, I believe modifying the current implementation of `sendRawTransaction` in this way is essential for integrating the new logic, while still supporting the previous functionality. Sticking to the old implementation just because it's less confusing could limit our ability to capture and handle errors robustly. This adjustment enables us to securely capture the correct errors, handle them effectively, and reuse the mirror node polling logic in a more logical and consistent way. While it may initially seem unfamiliar, I believe this modification actually streamlines the process rather than adding complexity, allowing the codebase to evolve and adapt to new requirements.

packages/relay/src/lib/eth.ts
quiet-node: Ah good point! Well it was from the previous implementation so I thought it was covered! Will add tests where appropriate!

packages/relay/src/lib/eth.ts
quiet-node: Yeah great point! NGL but that was actually my initial thought as well to only include it here, and as you mentioned, leaving it here in the catch block would only allow for mirror node polling when we encounter "connection dropped" or "timeout exceeded" errors. However, the previous implementation also polled the mirror node for transaction validity in successful cases (see [this block](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/eth.ts#L1592-L1609)) to make sure the transaction is fully processed by CN and MN before the transaction hash is returned to the client. 

This PR expands the polling logic to support both the old and new logic without losing any functionality. That’s why I’ve moved it outside of the try-catch-finally block to enable the polling logic to be reused for both successful cases and Timeout or Connection Dropped scenarios.

victor-yanev: We can query the transaction receipt directly from the consensus node free of cost (and way faster), I see no reason to poll MN in positive scenarios 🤷‍♂️

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -781,22 +721,28 @@ export class SDKClient {
       return transactionResponse;
     } catch (e: any) {
       if (e instanceof JsonRpcError) {
+        this.logger.warn(
+          e,
+          `${requestDetails.formattedRequestId} Failed to execute ${txConstructorName} transaction due to a JsonRpcError: transactionId=${transaction.transactionId}, callerName=${callerName}, status=${e.code} message=${e.message}`,
+        );
         throw e;
       }
```
victor-yanev, line 729: QQ: What's the point of logging a warning here and throwing the error?

packages/relay/src/lib/eth.ts
```diff
@@ -1578,46 +1583,24 @@ export class EthImpl implements Eth {
...
+      submittedTransactionId = sendRawTransactionResult.txResponse.transactionId?.toString();
+      formattedTransactionId = formatTransactionIdWithoutQueryParams(submittedTransactionId);
+      if (!formattedTransactionId) {
+        this.logger.warn(
+          `${requestIdPrefix} Transaction successfully submitted but returned invalid transactionID: transactionId==${submittedTransactionId}`,
+        );
+        throw predefined.INTERNAL_ERROR(
+          `Transaction successfully submitted but returned invalid transactionID: transactionId==${submittedTransactionId}`,
```
victor-yanev, line 1594: Also here, why are we logging a warning here if we are immediately throwing an error with exactly the same message?

packages/relay/src/lib/eth.ts
```diff
@@ -1630,6 +1613,61 @@ export class EthImpl implements Eth {
...
+    //   - The transaction encountered "SDK timeout exceeded" or "Connection Dropped" errors from the SDK but still potentially reached the consensus level.
+    // In both scenarios, polling the MN is required to verify the transaction's validity before return the transaction hash to clients.
+    if (formattedTransactionId) {
+      try {
+        const contractResult = await this.mirrorNodeClient.repeatedRequest(
+          this.mirrorNodeClient.getContractResult.name,
+          [formattedTransactionId, requestDetails],
+          this.mirrorNodeClient.getMirrorNodeRequestRetryCount(),
+          requestDetails,
+        );
```
victor-yanev, line 1628:
```suggestion
    if (transactionId) {
      try {
        const formattedId = formatTransactionIdWithoutQueryParams(transactionId);
        const contractResult = await this.mirrorNodeClient.repeatedRequest(
          this.mirrorNodeClient.getContractResult.name,
          [formattedId, requestDetails],
          this.mirrorNodeClient.getMirrorNodeRequestRetryCount(),
          requestDetails,
        );
```

packages/relay/src/lib/eth.ts
quiet-node: > We can query the transaction receipt directly from the consensus node free of cost (and way faster), I see no reason to poll MN in positive scenarios 🤷‍♂️

Great idea indeed! This is actually a valuable addition to #3202 to enhance `eth_sendRawTransaction`. However, for the scope of this PR, I aim to maintain the original design while expanding it to accommodate the new behavior.

packages/relay/src/lib/eth.ts
quiet-node: https://github.com/hashgraph/hedera-json-rpc-relay/pull/3229#discussion_r1834622780

packages/relay/src/lib/eth.ts
```diff
@@ -1578,46 +1583,24 @@ export class EthImpl implements Eth {
...
-        this.logger.warn(`${requestIdPrefix} No record retrieved`);
-        throw predefined.INTERNAL_ERROR(`No matching record found for transaction id ${txId}`);
-      }
-
-      if (contractResult.hash == null) {
-        this.logger.error(
-          `${requestIdPrefix} The ethereumHash can never be null for an ethereum transaction, and yet it was!!`,
+      fileId = sendRawTransactionResult.fileId;
+      submittedTransactionId = sendRawTransactionResult.txResponse.transactionId?.toString();
+      formattedTransactionId = formatTransactionIdWithoutQueryParams(submittedTransactionId);
```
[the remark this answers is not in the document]
quiet-node: I also considered using the same submitted transactionID and reusing it outside the try-catch block as suggested in [your comment down here](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3229#discussion_r1833946368). However, there's a check for `Transaction successfully submitted but returned invalid transactionID` that is only relevant within the scope of this try block. 

Technically, I don’t believe it's possible for a successfully submitted transaction to return an invalid transactionID, but that doesn’t mean we shouldn’t account for potential edge cases. Additionally, this check is part of the original implementation, and I’d prefer not to remove it within the scope of this PR.

packages/relay/src/lib/eth.ts
quiet-node: https://github.com/hashgraph/hedera-json-rpc-relay/pull/3229#discussion_r1834650414

packages/relay/src/lib/clients/sdkClient.ts
victor-yanev: I think we should have a stack trace inside of the error that is being thrown and this warning log would only clutter the logs with duplicated errors, correct me if I'm wrong

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
victor-yanev: What I meant is that we can define only one field - `transactionId` and use only that field throughout the method. Since `formattedTransactionId` is used only for the sake of calling the mirror node, I don't see a reason we would define it as a global variable in the method. 

Instead, what I suggested is to format it right before we send the request to the mirror node as shown in https://github.com/hashgraph/hedera-json-rpc-relay/pull/3229#discussion_r1833946368 and have it as a local variable in that part of the logic, instead of maintaining it globally on the method level (I hope that makes sense to you)

packages/relay/src/lib/eth.ts
quiet-node: @victor-yanev noted this idea down in [this comment](https://github.com/hashgraph/hedera-json-rpc-relay/issues/3202#issuecomment-2465234136) in the #3202 ongoing discussion. If you have any other idea on how to improve e2e latency of the sendRawTransaction, please few free to edit and add more to [the comment](https://github.com/hashgraph/hedera-json-rpc-relay/issues/3202#issuecomment-2465234136)

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
quiet-node: Yeah that makes total sense! I initially kept `formattedTransactionId` because I was relying on `formatTransactionIdWithoutQueryParams()` to verify the transaction’s validity by checking `transactionID` against `constants.TRANSACTION_ID_REGEX`. That’s why I reused the formatted ID globally. However, after looking into `formatTransactionIdWithoutQueryParams()`, it seems I can simply check `transactionID` directly within the try block without needing to format it.

Thanks for making me to reconsider this you’re absolutely right! There’s no need to maintain `formattedTransactionId` globally.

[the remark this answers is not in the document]
quiet-node: Pushed in an update in this commit https://github.com/hashgraph/hedera-json-rpc-relay/pull/3229/commits/e56d4a5e184b28021f9c0e43b0bd69b9504d7476. 

Nice job pointing it out 🙌🏼

packages/relay/src/lib/eth.ts
quiet-node: updated in this commit https://github.com/hashgraph/hedera-json-rpc-relay/pull/3229/commits/e56d4a5e184b28021f9c0e43b0bd69b9504d7476

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Yes, you're correct that the error includes a stack trace. However, that trace is only visible if explicitly logged. If a higher-level handler catches the error and performs logic without logging it, no log would be generated, leading to a loss of context and detail. This lack of logging can make debugging more challenging, as the original context of the error isn't captured. Logging a warning here ensures the relevant details are preserved, even if the error handling at higher levels omits logging.

victor-yanev: > If a higher-level handler catches the error and performs logic without logging it, no log would be generated, leading to a loss of context and detail.

Usually that's a responsibility of the piece of code which handles the error - they can choose to either re-throw the error or log a warning and ignore it or do something else. Logging a warning before every error, can easily lead to many duplicated logs and make it harder to trace the application flow.

If you insist on keeping this, let's change it from a `warn` to a `trace` log, so that these "safety/traceability logs" are shown only on `trace` level.

packages/relay/src/lib/eth.ts
victor-yanev: https://github.com/hashgraph/hedera-json-rpc-relay/pull/3229#discussion_r1836126524

packages/relay/src/lib/clients/sdkClient.ts
konstantinabl: @victor-yanev I would agree with Victor. It is responsibility of the handler to decide what to do. I think we shouldn't clutter the code with logs, since it can affect memory and performance and lead to confusion when investigating an error through logs

quiet-node: Updated in https://github.com/hashgraph/hedera-json-rpc-relay/pull/3229/commits/9f9f1622d8b4e3c692abe19ae86e5ca80399e08e. Thanks for the suggestion!

packages/relay/src/lib/eth.ts
quiet-node: Updated in https://github.com/hashgraph/hedera-json-rpc-relay/pull/3229/commits/9f9f1622d8b4e3c692abe19ae86e5ca80399e08e. Thanks for the suggestion!

konstantinabl (approved):

victor-yanev (approved):

PR #3240, opened 2024-11-08
fix: 3239 tests need to strip leading zeros

### LINKED ISSUE #3239: Some tests do not strip leading zeros in signature s and v values leading to failures that appear to be flaky tests.
### Description

Leading zeros on signature r and s values are stripped in the `formatters.formatContractResult` in the relay before the results are returned.  On the mirror node they are stored with the leading zero.  e.g.
The transaction hash `0xc5a3df675c99f634d9d81a8092fc65db54d7c28dc21126439329e92598976d2d` on testnet returns a signature s value of:
`0x091027721aff181da255f7bda0cf5b834a0d5099b46dc636b1dbd50a30e33383`  from the [mirror](https://testnet.mirrornode.hedera.com/api/v1/contracts/results/0xc5a3df675c99f634d9d81a8092fc65db54d7c28dc21126439329e92598976d2d) node.
The `eth_getTransactionReceipt` response from the relay returns a signature s value of `0x91027721aff181da255f7bda0cf5b834a0d5099b46dc636b1dbd50a30e33383`
```
curl --location 'https://testnet.hashio.io/api' \
--header 'Request-Id: testRequest' \
--data '{
  "id": 1,
  "jsonrpc": "2.0",
  "params": [
    "0xc5a3df675c99f634d9d81a8092fc65db54d7c28dc21126439329e92598976d2d"
  ],
  "method": "eth_getTransactionByHash"
}'
```
The leading zeros in the v and s signature values will cause `json: cannot unmarshal hex number with leading zero digits into Go struct field txJSON.s of type *hexutil.Big` in the go-ethereum client.  

Some tests verify the signature v and s values using the response from the mirror node, but these values need to strip the leading zeros before being evaluated.

### Steps to reproduce

1. Run the `acceptancetest:release_ws` tests.
2. These failures require signature v and s values with leading zeros so you may need to run these tests several times.  This gives the impression of "flaky tests".
3. Look for test failures comparing values and failing because one value has a leading zero while the other does not.

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.59.4

### Operating system

macOS

ebadiere:
**Description**:
Test fix.  Ethereum transaction signatures v and s values need to have leading zeros stripped before being evaluated.
Detailed description can be found in the issue.

**Related issue(s)**:

Fixes #3239


ebadiere:
Replaced by #[3421](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3241)

PR #3252, opened 2024-11-12
fix: Allow the HBar Rate Limiter to be disabled.

### LINKED ISSUE #3251: HBar Rate Limiter should have the ability to be disabled.
### Description

As a relay operator I should have the ability to disable all HBar Rate Limiting.

### Steps to reproduce

Currently not available, but it was.  As a workaround the spending tiers can have high values set.

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.59.0

### Operating system

Linux

ebadiere:
This fix allows the HBar Rate Limiter to be disabled by setting the `HBAR_RATE_LIMIT_TINYBAR` to `0`

**Related issue(s)**:

Fixes #3251 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "acceptancetest:ratelimiter": "nyc ts-mocha packages/ws-server/tests/acceptance/index.spec.ts -g '@web-socket-ratelimiter' --exit && ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@ratelimiter' --exit",
     "acceptancetest:hbarlimiter_batch1": "HBAR_RATE_LIMIT_BASIC=1000000000 HBAR_RATE_LIMIT_EXTENDED=1500000000 HBAR_RATE_LIMIT_PRIVILEGED=2000000000 nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter-batch1' --exit",
     "acceptancetest:hbarlimiter_batch2": "HBAR_RATE_LIMIT_BASIC=1000000000 HBAR_RATE_LIMIT_EXTENDED=1500000000 HBAR_RATE_LIMIT_PRIVILEGED=2000000000 nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter-batch2' --exit",
+    "acceptancetest:hbarlimiter_batch3": "HBAR_RATE_LIMIT_TINYBAR=0 HBAR_RATE_LIMIT_BASIC=1000000000 HBAR_RATE_LIMIT_EXTENDED=1500000000 HBAR_RATE_LIMIT_PRIVILEGED=2000000000 nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter-batch3' --exit",
     "acceptancetest:tokencreate": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@tokencreate' --exit",
     "acceptancetest:tokenmanagement": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@tokenmanagement' --exit",
     "acceptancetest:htsprecompilev1": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@htsprecompilev1' --exit",
--- a/packages/relay/src/lib/constants.ts
+++ b/packages/relay/src/lib/constants.ts
@@ -145,7 +145,7 @@ export default {
   // @ts-ignore
   HBAR_RATE_LIMIT_DURATION: parseInt(ConfigService.get('HBAR_RATE_LIMIT_DURATION') || '86400000'), // 1 day
   // @ts-ignore
-  HBAR_RATE_LIMIT_TOTAL: BigNumber(ConfigService.get('HBAR_RATE_LIMIT_TINYBAR') || '800000000000'), // 8000 HBARs
+  HBAR_RATE_LIMIT_TOTAL: BigNumber(ConfigService.get('HBAR_RATE_LIMIT_TINYBAR') ?? '800000000000'), // 8000 HBARs
   // @ts-ignore
   HBAR_RATE_LIMIT_BASIC: BigNumber(ConfigService.get('HBAR_RATE_LIMIT_BASIC') || '1120000000'), // 11.2 HBARs
   // @ts-ignore
--- a/packages/relay/src/lib/constants.ts
+++ b/packages/relay/src/lib/constants.ts
@@ -145,6 +145,8 @@ export default {
   // @ts-ignore
   HBAR_RATE_LIMIT_DURATION: parseInt(ConfigService.get('HBAR_RATE_LIMIT_DURATION') || '86400000'), // 1 day
   // @ts-ignore
+  // The logical OR operator || returns the first truthy value and 0 is falsy.
+  // The nullish coalescing operator ?? falls back to the default value when the left-hand operand is null or undefined, not when it's 0 or any other falsy value.
   HBAR_RATE_LIMIT_TOTAL: BigNumber(ConfigService.get('HBAR_RATE_LIMIT_TINYBAR') ?? '800000000000'), // 8000 HBARs
   // @ts-ignore
   HBAR_RATE_LIMIT_BASIC: BigNumber(ConfigService.get('HBAR_RATE_LIMIT_BASIC') || '1120000000'), // 11.2 HBARs
--- a/packages/relay/src/lib/services/hbarLimitService/IHbarLimitService.ts
+++ b/packages/relay/src/lib/services/hbarLimitService/IHbarLimitService.ts
@@ -31,4 +31,5 @@ export interface IHbarLimitService {
     estimatedTxFee?: number,
   ): Promise<boolean>;
   addExpense(cost: number, ethAddress: string, requestDetails: RequestDetails, ipAddress?: string): Promise<void>;
+  isEnabled(): boolean;
 }
--- a/packages/relay/src/lib/services/hbarLimitService/index.ts
+++ b/packages/relay/src/lib/services/hbarLimitService/index.ts
@@ -96,7 +96,7 @@ export class HbarLimitService implements IHbarLimitService {
     this.reset = this.getResetTimestamp();
     this.remainingBudget = this.totalBudget;
 
-    if (this.remainingBudget.toTinybars().isZero()) {
+    if (this.remainingBudget.toTinybars().lte(0)) {
       this.disableRateLimiter = true;
     }
 
@@ -153,6 +153,13 @@ export class HbarLimitService implements IHbarLimitService {
   }
 
   /**
+   * Checks if the rate limiter is enabled.
+   */
+  isEnabled(): boolean {
+    return !this.disableRateLimiter;
+  }
+
+  /**
    * Resets the {@link HbarSpendingPlan#amountSpent} field for all existing plans.
    * @param {RequestDetails} requestDetails - The request details used for logging and tracking.
    * @returns {Promise<void>} - A promise that resolves when the operation is complete.
@@ -191,7 +198,7 @@ export class HbarLimitService implements IHbarLimitService {
     requestDetails: RequestDetails,
     estimatedTxFee: number = 0,
   ): Promise<boolean> {
-    if (this.disableRateLimiter) {
+    if (!this.isEnabled()) {
       return false;
     }
 
--- a/packages/relay/src/lib/services/hbarLimitService/index.ts
+++ b/packages/relay/src/lib/services/hbarLimitService/index.ts
@@ -257,6 +257,10 @@ export class HbarLimitService implements IHbarLimitService {
    * @returns {Promise<void>} - A promise that resolves when the expense has been added.
    */
   async addExpense(cost: number, ethAddress: string, requestDetails: RequestDetails): Promise<void> {
+    if (!this.isEnabled()) {
+      return;
+    }
+
     const newRemainingBudget = this.remainingBudget.toTinybars().sub(cost);
     this.remainingBudget = Hbar.fromTinybars(newRemainingBudget);
     this.hbarLimitRemainingGauge.set(newRemainingBudget.toNumber());
--- a/packages/relay/src/lib/services/hbarLimitService/index.ts
+++ b/packages/relay/src/lib/services/hbarLimitService/index.ts
@@ -38,6 +38,12 @@ export class HbarLimitService implements IHbarLimitService {
   };
 
   /**
+   * Disables the rate limiter.
+   * @private
+   */
+  private readonly disableRateLimiter: boolean = false;
+
+  /**
    * Counts the number of times the rate limit has been reached.
    * @private
    */
@@ -89,6 +95,10 @@ export class HbarLimitService implements IHbarLimitService {
   ) {
     this.reset = this.getResetTimestamp();
     this.remainingBudget = this.totalBudget;
+
+    if (this.remainingBudget.toTinybars().isZero()) {
+      this.disableRateLimiter = true;
+    }
 
     const metricCounterName = 'rpc_relay_hbar_rate_limit';
     this.register.removeSingleMetric(metricCounterName);
@@ -181,6 +191,10 @@ export class HbarLimitService implements IHbarLimitService {
     requestDetails: RequestDetails,
     estimatedTxFee: number = 0,
   ): Promise<boolean> {
+    if (this.disableRateLimiter) {
+      return false;
+    }
+
     const ipAddress = requestDetails.ipAddress;
     if (await this.isTotalBudgetExceeded(mode, methodName, txConstructorName, estimatedTxFee, requestDetails)) {
       return true;
--- a/packages/relay/src/lib/services/hbarLimitService/index.ts
+++ b/packages/relay/src/lib/services/hbarLimitService/index.ts
@@ -96,7 +96,7 @@ export class HbarLimitService implements IHbarLimitService {
     this.reset = this.getResetTimestamp();
     this.remainingBudget = this.totalBudget;
 
-    if (this.remainingBudget.toTinybars().lte(0)) {
+    if (this.totalBudget.toTinybars().lte(0)) {
       this.disableRateLimiter = true;
     }
 
--- a/packages/relay/src/lib/services/hbarLimitService/index.ts
+++ b/packages/relay/src/lib/services/hbarLimitService/index.ts
@@ -38,7 +38,7 @@ export class HbarLimitService implements IHbarLimitService {
   };
 
   /**
-   * Disables the rate limiter.
+   * Flag to turn off the HBarRateLimitService.
    * @private
    */
   private readonly disableRateLimiter: boolean = false;
--- a/packages/relay/src/lib/services/hbarLimitService/index.ts
+++ b/packages/relay/src/lib/services/hbarLimitService/index.ts
@@ -41,7 +41,7 @@ export class HbarLimitService implements IHbarLimitService {
    * Flag to turn off the HBarRateLimitService.
    * @private
    */
-  private readonly disableRateLimiter: boolean = false;
+  private readonly isHBarRateLimiterEnabled: boolean = true;
 
   /**
    * Counts the number of times the rate limit has been reached.
@@ -97,7 +97,7 @@ export class HbarLimitService implements IHbarLimitService {
     this.remainingBudget = this.totalBudget;
 
     if (this.totalBudget.toTinybars().lte(0)) {
-      this.disableRateLimiter = true;
+      this.isHBarRateLimiterEnabled = false;
     }
 
     const metricCounterName = 'rpc_relay_hbar_rate_limit';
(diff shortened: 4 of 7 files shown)


packages/relay/src/lib/constants.ts
```diff
@@ -145,7 +145,7 @@ export default {
   // @ts-ignore
   HBAR_RATE_LIMIT_DURATION: parseInt(ConfigService.get('HBAR_RATE_LIMIT_DURATION') || '86400000'), // 1 day
   // @ts-ignore
-  HBAR_RATE_LIMIT_TOTAL: BigNumber(ConfigService.get('HBAR_RATE_LIMIT_TINYBAR') || '800000000000'), // 8000 HBARs
+  HBAR_RATE_LIMIT_TOTAL: BigNumber(ConfigService.get('HBAR_RATE_LIMIT_TINYBAR') ?? '800000000000'), // 8000 HBARs
```
Nana-EC, line 148: Q: why this change and does it affect all the other configs?

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -90,6 +96,10 @@ export class HbarLimitService implements IHbarLimitService {
     this.reset = this.getResetTimestamp();
     this.remainingBudget = this.totalBudget;
 
+    if (this.remainingBudget.toTinybars().isZero()) {
```
Nana-EC, line 99: nit: as a guard I'd say zero or less

packages/relay/tests/lib/services/hbarLimitService/hbarLimitService.spec.ts
```diff
@@ -494,6 +494,29 @@ describe('HBAR Rate Limit Service', function () {
...
+        ethAddressHbarSpendingPlanRepositoryStub,
+        ipAddressHbarSpendingPlanRepositoryStub,
+        logger,
+        register,
+        Hbar.fromTinybars(0),
+        limitDuration,
+      );
+      it('should return false if the rate limiter is disabled by setting HBAR_RATE_LIMIT_TINYBAR to zero', async function () {
+        // hbarLimitServiceDisabled.disableRateLimiter();
+        const result = await hbarLimitServiceDisabled.shouldLimit(
```
Nana-EC, line 510: This actually doesn't really test that the rate limiter is working.
For instance if a future code change was made and the return was false for some other reason that a 0 value this logic would not catch it.
I think we should reconsider the layout on the product side.

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -181,6 +191,10 @@ export class HbarLimitService implements IHbarLimitService {
     requestDetails: RequestDetails,
     estimatedTxFee: number = 0,
   ): Promise<boolean> {
+    if (this.disableRateLimiter) {
```
Nana-EC, line 194: I would suggest adding an `isEnabled()` function to `IHbarLimitService`.
Then this method can call it and it can be tested.

Nana-EC (commented):
Good approach. 
Question on the config changes.
Also I provided a suggestion to be a bit more strict

packages/relay/src/lib/constants.ts
ebadiere: While working on the acceptance test I noticed that the zero value for the `HBAR_RATE_LIMIT_TINYBAR` was not getting picked up.  Most likely because the logical OR operator `||` returns the first truthy value and 0 is falsy.  The nullish coalescing operator `??` falls back to the default value when the left-hand operand is `null` or `undefined`, not when it's `0` or any other falsy value.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -887,5 +914,77 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
         });
       });
     });
+
+    describe('@hbarlimiter-batch3 BASIC Tier', () => {
```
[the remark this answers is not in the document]
ebadiere: I had to add it to another branch because of the order of loading environment variables in the acceptance tests.  The `relay.ts` reads and sets the `HBAR_RATE_LIMIT_TINYBAR` at boot time.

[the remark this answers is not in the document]
quiet-node: I believe he is addressing the scenario in which the limiter is turned off, allowing an account to make calls without being HBAR rate limited.

@ebadiere, if that's true and the suite is needed, I suggest renaming the test suite from "BASIC Tier" to something more descriptive, such as "Unlimited Limit" or "Limiter Off" or something like that. The term "BASIC Tier" is confusing, as we have another suite with the same name.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -887,5 +914,77 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+            hbarSpendingPlan = await hbarSpendingPlanRepository.findByIdWithDetails(
+              ethSpendingPlan.planId,
+              requestDetails,
+            );
+          }
+
+          await pollForProperAmountSpent(hbarSpendingPlan, deploymentCounts + 1, expectedTxCost);
+        }
+        // Verify that the amount spent exceeds the HBAR limit
+        const amountSpent = await pollForProperAmountSpent(hbarSpendingPlan, deploymentCounts, expectedTxCost);
```
quiet-node, line 978: after the loop hbarSpendingPlan can still be null in cases that no loop has got through. Please add this block above this line so the null error can be discarded
```typescript
      if (!hbarSpendingPlan) {
             const ethSpendingPlan = await ethAddressSpendingPlanRepository.findByAddress(
             accounts[2].wallet.address,
             requestDetails,
             );
             hbarSpendingPlan = await hbarSpendingPlanRepository.findByIdWithDetails(
                 ethSpendingPlan.planId,
                 requestDetails,
        );
     }
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -887,5 +914,77 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+        expect(amountSpent).to.be.gte(maxBasicSpendingLimit);
+
+        // Verify that remaining HBAR limit is zero or negative
+        const remainingHbarsAfter = Number(await metrics.get(testConstants.METRICS.REMAINING_HBAR_LIMIT));
+        expect(remainingHbarsAfter).to.be.lte(0);
+
+        // Confirm that deployments continued even after the limit was exhausted
+        expect(deploymentCounts).to.be.gte(1);
```
quiet-node, line 986: Just an extra proof saying that the account can still make calls
```typescript
await expect(deployContract(largeContractJson, same_account.wallet)).to.be.fulfilled;
```

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -887,5 +914,77 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+        logger.info(`${requestDetails.formattedRequestId} Creating accounts`);
+        logger.info(
+          `${requestDetails.formattedRequestId} HBAR_RATE_LIMIT_TINYBAR: ${ConfigService.get(
+            'HBAR_RATE_LIMIT_TINYBAR',
+          )}`,
+        );
+
+        const initialAccount: AliasAccount = global.accounts[0];
+
+        const neededAccounts: number = 3;
```
quiet-node, line 931: might have missed it but it doesn't seem to need that many alias accounts. Maybe just one account is good enough.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -887,5 +914,77 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+        const basicPlans = await hbarSpendingPlanRepository.findAllActiveBySubscriptionTier(
+          [SubscriptionTier.BASIC],
+          requestDetails,
+        );
+        for (const plan of basicPlans) {
+          await hbarSpendingPlanRepository.delete(plan.id, requestDetails);
+          await ethAddressSpendingPlanRepository.deleteAllByPlanId(plan.id, 'before', requestDetails);
+          await ipSpendingPlanRepository.deleteAllByPlanId(plan.id, 'before', requestDetails);
+        }
+      });
```
quiet-node, line 951: Correct me if I'm wrong but the plans might not need to be reset. It was for something special tested in batch1 BASIC tier.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -887,5 +914,77 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+    describe('@hbarlimiter-batch3 BASIC Tier', () => {
+      const accounts: AliasAccount[] = [];
+
+      beforeEach(async function () {
+        logger.info(`${requestDetails.formattedRequestId} Creating accounts`);
+        logger.info(
+          `${requestDetails.formattedRequestId} HBAR_RATE_LIMIT_TINYBAR: ${ConfigService.get(
+            'HBAR_RATE_LIMIT_TINYBAR',
+          )}`,
+        );
```
quiet-node, line 927: Since this suite provides coverage when the limiter is off, there should likely be assertions confirming that `HBAR_RATE_LIMIT_TINYBAR` is set to 0 and that `isDisable` is set to true. For example, we could include an assertion like `expect(HBAR_RATE_LIMIT_TINYBAR).to.eq(0)`. If it's feasible to access `isDisable`, we could also add an assertion for that as well (not sure if the isDisable is reachable tho)

package.json
```diff
@@ -50,6 +50,7 @@
     "acceptancetest:ratelimiter": "nyc ts-mocha packages/ws-server/tests/acceptance/index.spec.ts -g '@web-socket-ratelimiter' --exit && ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@ratelimiter' --exit",
     "acceptancetest:hbarlimiter_batch1": "HBAR_RATE_LIMIT_BASIC=1000000000 HBAR_RATE_LIMIT_EXTENDED=1500000000 HBAR_RATE_LIMIT_PRIVILEGED=2000000000 nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter-batch1' --exit",
     "acceptancetest:hbarlimiter_batch2": "HBAR_RATE_LIMIT_BASIC=1000000000 HBAR_RATE_LIMIT_EXTENDED=1500000000 HBAR_RATE_LIMIT_PRIVILEGED=2000000000 nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter-batch2' --exit",
+    "acceptancetest:hbarlimiter_batch3": "HBAR_RATE_LIMIT_TINYBAR=0 HBAR_RATE_LIMIT_BASIC=1000000000 HBAR_RATE_LIMIT_EXTENDED=1500000000 HBAR_RATE_LIMIT_PRIVILEGED=2000000000 nyc ts-mocha packages/server/tests/acceptance/index.spec.ts -g '@hbarlimiter-batch3' --exit",
```
quiet-node, line 53: This will need to be added in the workflow as well so it can be shown in the CI board

packages/relay/tests/lib/services/hbarLimitService/hbarLimitService.spec.ts
ebadiere: Updated test to include a condition that would normally trigger rate limiting, but verifies that `shouldLimit` returns false since rate limiting is not enabled.

packages/relay/src/lib/services/hbarLimitService/index.ts
ebadiere: Done.

packages/relay/src/lib/services/hbarLimitService/index.ts
ebadiere: Done.

packages/server/tests/acceptance/hbarLimiter.spec.ts
ebadiere: Yes, one is good enough.

packages/server/tests/acceptance/hbarLimiter.spec.ts
ebadiere: You are correct. Initially, I observed some instability when running the tests. To address this, I thought that clearing out the repositories—so that the spending plans would be recreated during the contract deployments—would provide a deterministic solution. However, after removing the code that cleared the repositories and running the tests multiple times, I achieved consistent results, indicating that the instability was unrelated to the repository state.

packages/server/tests/acceptance/hbarLimiter.spec.ts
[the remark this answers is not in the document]
ebadiere: @quiet-node Done.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -57,6 +57,33 @@ import mediumSizeContract from '../contracts/hbarLimiterContracts/mediumSizeCont
 config({ path: resolve(__dirname, '../localAcceptance.env') });
 const DOT_ENV = dotenv.parse(fs.readFileSync(resolve(__dirname, '../localAcceptance.env')));
 
+const pollForProperAmountSpent = async (
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/server/tests/acceptance/hbarLimiter.spec.ts
ebadiere: Done.

packages/server/tests/acceptance/hbarLimiter.spec.ts
ebadiere: Done.

packages/server/tests/acceptance/hbarLimiter.spec.ts
ebadiere: Done.

package.json
ebadiere: Done.

packages/server/tests/acceptance/hbarLimiter.spec.ts
ebadiere: I added ` expect(ConfigService.get('HBAR_RATE_LIMIT_TINYBAR')).to.eq(0);` to the test.  I don't think it's easily feasible to check `isEnabled()` from an acceptance test.  The running instance of the `HBarRateLimiter` would have to be captured in the server package This is covered in the unit test in the relay.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -887,5 +887,85 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+            neededAccounts,
+            initialBalance,
+            requestDetails,
+          )),
+        );
+        global.accounts.push(...accounts);
+        const basicPlans = await hbarSpendingPlanRepository.findAllActiveBySubscriptionTier(
+          [SubscriptionTier.BASIC],
+          requestDetails,
+        );
```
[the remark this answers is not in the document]
ebadiere: Correct, thanks.  Removed.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -887,5 +887,85 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+        const neededAccounts: number = 1;
+        accounts.push(
+          ...(await Utils.createMultipleAliasAccounts(
+            mirrorNode,
+            initialAccount,
+            neededAccounts,
+            initialBalance,
+            requestDetails,
+          )),
+        );
```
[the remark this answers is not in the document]
ebadiere: Yes.  Removed and just using `global.accounts[0]`.  Thanks.

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -142,6 +152,14 @@ export class HbarLimitService implements IHbarLimitService {
...
   }
 
+  /**
+   * Checks if the rate limiter is enabled.
+   * returns {boolean} - `true` if the rate limiter is enabled, otherwise `false`.
+   */
+  isEnabled(): boolean {
+    return !this.disableRateLimiter;
+  }
+
```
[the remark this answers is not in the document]
ebadiere: This was recommended by @Nana-EC in one of the earlier comments.  It makes sense to me.

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -37,6 +37,12 @@ export class HbarLimitService implements IHbarLimitService {
     PRIVILEGED: Hbar.fromTinybars(constants.HBAR_RATE_LIMIT_PRIVILEGED),
   };
 
+  /**
+   * Disables the rate limiter.
```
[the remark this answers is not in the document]
ebadiere: Updated comments to be more descriptive.

packages/relay/src/lib/services/hbarLimitService/index.ts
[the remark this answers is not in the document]
Nana-EC: Yeah, I think it's cleaner interface wise also since the service isn't required to be on. I'm open to debate

packages/relay/src/lib/services/hbarLimitService/IHbarLimitService.ts
```diff
@@ -31,4 +31,5 @@ export interface IHbarLimitService {
     estimatedTxFee?: number,
   ): Promise<boolean>;
   addExpense(cost: number, ethAddress: string, requestDetails: RequestDetails, ipAddress?: string): Promise<void>;
+  isEnabled(): boolean;
```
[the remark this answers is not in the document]
Nana-EC: Yeah see https://github.com/hashgraph/hedera-json-rpc-relay/pull/3252#discussion_r1841253642

packages/server/tests/acceptance/hbarLimiter.spec.ts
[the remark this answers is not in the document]
ebadiere: Removed.

mishomihov00 (approved):
Reviewed and approved only this file:
`.github/workflows/acceptance.yml`

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -90,6 +96,10 @@ export class HbarLimitService implements IHbarLimitService {
     this.reset = this.getResetTimestamp();
     this.remainingBudget = this.totalBudget;
 
+    if (this.remainingBudget.toTinybars().lte(0)) {
+      this.disableRateLimiter = true;
+    }
```
victor-yanev, line 101:
```suggestion
    if (this.totalBudget.toTinybars().lte(0)) {
      this.disableRateLimiter = true;
    }
```

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -181,6 +199,10 @@ export class HbarLimitService implements IHbarLimitService {
     requestDetails: RequestDetails,
     estimatedTxFee: number = 0,
   ): Promise<boolean> {
+    if (!this.isEnabled()) {
+      return false;
+    }
```
victor-yanev, line 204: Please, add this check to `addExpense` as well:
```typescript
async addExpense(cost: number, ethAddress: string, requestDetails: RequestDetails): Promise<void> {
  if (!this.isEnabled()) {
    return;
  }
  ...
}
```

ebadiere: Done.

packages/relay/src/lib/services/hbarLimitService/index.ts
ebadiere: Done

packages/relay/src/lib/services/hbarLimitService/index.ts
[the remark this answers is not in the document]
quiet-node: Ah I see yeah a getter is good in terms of encapsulation and readability. Thanks for the explanation!

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -887,5 +887,67 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
         });
       });
     });
+
+    describe('@hbarlimiter-batch3 Unlimited', () => {
+      const accounts: AliasAccount[] = [];
```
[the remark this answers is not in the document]
ebadiere: Done.

packages/server/tests/acceptance/hbarLimiter.spec.ts
```diff
@@ -887,5 +887,67 @@ describe('@hbarlimiter HBAR Limiter Acceptance Tests', function () {
...
+
+      beforeEach(async function () {
```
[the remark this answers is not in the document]
ebadiere: Done.

quiet-node (approved):

packages/relay/src/lib/constants.ts
Nana-EC: Nice catch.
Can you make a note here on that. Also i think you should open a quick issue to note that all integer numbers will hit this issue so we should do a quick check and see if there are other items that need to be fixed in a separate PR

Nana-EC (approved):
LG, see note on config issue to open

quiet-node (approved):
LGTM! Great work!

Nana-EC (approved):

victor-yanev (approved):

PR #3256, opened 2024-11-13
chore: Reduce initial account funding

beeradb:
This PR reduces the initial account funding for public acceptance tests from 500 to 250hbar.

#3256


Nana-EC (approved):

ebadiere (approved):
LG.

PR #3258, opened 2024-11-13
feat: Creates design doc for enabling access to newly created hts token add…

### LINKED ISSUE #3257: Create design doc enabling returning of HTS token address in receipt
Currently when createNonFungibleToken is called directly in the 0x167 contract there is no way for users to get the contract address of the newly created token via ethers. Needs to design a solution

konstantinabl:
…ress

**Description**:

Creates a design doc for new change to trasnaction receipt format, related to hts token address
Reference: https://github.com/hashgraph/hedera-smart-contracts/issues/901

**Related issue(s)**:

Fixes #3257

### DIFF
--- a/docs/design/hts_address_tx_receipt.md
+++ b/docs/design/hts_address_tx_receipt.md
@@ -1,0 +1,36 @@
+# HTS Token Creation via SystemContract Receipt Enhancement
+
+## Purpose
+The purpose of this enhancement is to improve the interoperability between Hedera Token Service (HTS) and Ethereum-compatible tools, like ethers, by ensuring access to the necessary information about newly created tokens, in this case token address. This change specifically addresses the gap in Web3 tooling support when creating tokens through HTS precompile contracts, making user experience more smooth.
+
+
+## Problem Statement
+When users create tokens (fungible and non-fungible) by directly calling the 0x167 contract there is no way for to get the contract address of the newly created token via ethers. This is because the transaction receipt's contractAddress field is not being populated correctly for HTS token creation operations, since essentially they are just a contract call, but not a contract create (deploy).
+
+## Current Behaviour
+
+Currently, when creating a token via HTS the address of the system contract ic returned as a contractAddress in the transaction receipt.
+
+## Proposed Solution
+
+### Technical Details
+
+1. Function Signature Detection
+
+Detect in the transactionResponse from the mirror node, if the transaction was calling any of the HTS method e.g createFungibleToken/createNonFungibleToken etc.
+
+N.B Currently, HTS supports both v1 and v2 security model function selectors
+
+1. Extract the token address from the call_result field in the transaction response.
+
+2. Add the token address to the receipt
+
+
+## Testing Requirements
+
+1. Unit tests
+   1. Test token creation function signature detection
+
+2. Acceptance tests
+   1. Test fungible token creation flow via ethers and correct receipt response
+   2. Test non fungible token creation flow via ethers and correct receipt response


docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,36 @@
+# HTS Token Creation via SystemContract Receipt Enhancement
+
+## Purpose
+The purpose of this enhancement is to improve the interoperability between Hedera Token Service (HTS) and Ethereum-compatible tools, like ethers, by ensuring access to the necessary information about newly created tokens, in this case token address. This change specifically addresses the gap in Web3 tooling support when creating tokens through HTS precompile contracts, making user experience smoother.
+
+
+## Problem Statement
+When users create tokens (fungible and non-fungible) by directly calling the 0x167 contract there is no way for to get the contract address of the newly created token via ethers. This is because the transaction receipt's contractAddress field is not being populated with the HTS token address, since essentially the operations is just a contract call, but not a contract create (deploy).
```
Nana-EC, line 8:
```suggestion
When users create tokens (fungible and non-fungible) by directly calling the 0x167 contract there is no way for them to get the evm address of the newly created token via standard EVM tooling e.g. ethers. This is because the transaction receipt's contractAddress field is not being populated with the HTS token address, since essentially the operations is just a contract call, but not a contract create (deploy).
```

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,36 @@
...
+
+Currently, when creating a token via HTS the address of the system contract is returned as a contractAddress in the transaction receipt.
+
+## Proposed Solution
+
+### Technical Details
+
+1. Function Signature Detection
+
+Detect in the transactionResponse from the mirror node, if the transaction was calling any of the HTS method e.g createFungibleToken/createNonFungibleToken etc.
```
Nana-EC, line 20: Please list all current function that this intends to support

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,36 @@
...
+
+N.B Currently, HTS supports both v1 and v2 security model function selectors
+
+1. Extract the token address from the call_result field in the transaction response.
```
Nana-EC, line 24: Please lease an example of a call_result and what is looks like before and after extraction

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,36 @@
...
+
+2. Add the token address to the receipt
+
+
```
Nana-EC, line 28: Are there existing interfaces or higher level classes of not that should be modified to support this.
Essentially you should be telling us what high level changes to the code base you are suggesting

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,36 @@
...
+## Testing Requirements
+
+1. Unit tests
```
Nana-EC, line 31: UT's aren't necessary in the design doc.
Acceptance tests highlighting the difference scenarios is what should be highlighted.

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,36 @@
...
+   1. Test token creation function signature detection
+
+2. Acceptance tests
+   1. Test fungible token creation flow via ethers and correct receipt response
```
Nana-EC, line 35: There should be a test case for every create method being supported.

docs/design/hts_address_tx_receipt.md
Nana-EC, line 24: What approach are you intending to utilize to extract?
Is it just brute force string manipulation or are you using additional logic to parse the bytes into an expected format and pull the address from that?

Nana-EC (changes requested):
Nice start,
More details need to be highlighted in the design doc to give a clear idea of approach

docs/design/hts_address_tx_receipt.md
konstantinabl: Done.

docs/design/hts_address_tx_receipt.md
konstantinabl: Yes, of course
I just didnt list them in detail, thought its not necessary, but will do it now :) its good to have a detailed test plan

docs/design/hts_address_tx_receipt.md
konstantinabl: Brute force manipulation I would say, Improved the description

docs/design/hts_address_tx_receipt.md
konstantinabl: I dont't think we need to change any higher level classes. Only the getTransactionReceipt method in eth.ts file
So not sure what else should I add in terms of this note?

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Proposed Solution
+
+### Technical Details
+
+### 1. Function Signature Detection
+
+Detect in the transactionResponse from the mirror node, if the transaction was calling any of the HTS methods - createFungibleToken/createNonFungibleToken/createFungibleTokenWithCustomFees/createNonFungibleTokenWithCustomFees
+
+N.B Currently, HTS supports both v1 and v2 security model function selectors
```
Nana-EC, line 22: Suggestion to be more specific. Also it's not security model related
```suggestion
N.B The following forms of the [HTS system contract](https://github.com/hashgraph/hedera-smart-contracts/tree/v0.10.1/contracts/system-contracts/hedera-token-service) create token function selectors will be supported. This is as of the 0.56 version of the services.
```

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,61 @@
...
+  "createFungibleTokenWithCustomFees(...)",
+  "createNonFungibleTokenWithCustomFees(...)"
+];
+const functionSelector = receiptResponse.function_parameters.substring(0, FUNCTION_SELECTOR_CHAR_LENGTH);
+const isTokenCreation = HTS_CREATE_FUNCTIONS_SIGNATURE.some(signature => 
+  Utils.calculateFunctionSelector(signature) === functionSelector
+);
+```
+
+### 2. Extract the token address from the call_result field in the transaction response. The extractions can happen with brute force since we know what the function response is beforehand and we know the last 40 characters of the response are the token address.
```
Nana-EC, line 37:
```suggestion
### 2. Extract the token address from the `call_result` field in the transaction response. 
The extractions can performed with string matching logic since we know what the function response is beforehand and we know the last 40 characters of the response are the token address.
```

docs/design/hts_address_tx_receipt.md
Nana-EC: Sounds good

Nana-EC (commented):
Nice improvements.
Signing off so as to not block but please adopt some form of my suggestions

Nana-EC (approved):

Nana-EC:
> This design handles the direct precompile call flow well.
> 
> There might be a case where we create an HTS token within the contract's constructor. What will we return in that case? Is the newly created contract's address or the HTS token address?
> 
> In terms of the Ethereum equivalence, we always must return the contract's address. Maybe we should add a section in the readme with something like "In case of HTS token creation within the constructor, the user should manually emit an event and handle the token address exposure on the contract's side"?
> 
> @Nana-EC actually, are we thinking of handling an HTS token creation within a contract and populating the `contractAddress` field with the newly created token address on the top-level transaction's receipt?

Hmmm, I don't think we need to handle in contract calls as there it is the responsibility of the developer to handle the functions response. And yeah if they want they can emit an event to externalize the token address.

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,62 @@
...
+
+```javascript
+const tokenAddress = receiptResponse.call_result.substring(receiptResponse.call_result.length - 40);
+```
+After extraction, the expected output is:
+```
+"0x000000000000000000000000000000000000000000000000000000000000040a"
+```
+
+### 3. Add the token address to the receipt
```
Nana-EC, line 53: nit: nothing in this section.
Did you mean to highlight that you are replacing the default `0x167` value with this?

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Current Behaviour
+
+Currently, when creating a token via HTS the address of the system contract is returned as a contractAddress in the transaction receipt.
+
+## Proposed Solution
+
+### Technical Details
+
+### 1. Function Signature Detection
```
Nana-EC, line 18: Notably, this should be done in a way that can be generalized to other future system contracts e.g. createSchedule or createTopic

Nana-EC (approved):
Approach generally looks good to me.
Unblocking and looking to at least 2 other team members from @hashgraph/hedera-smart-contracts-product group to sign off here to represent consensus on approach

natanasow (dismissed):
LGTM

docs/design/hts_address_tx_receipt.md
konstantinabl: Well, i have shown in the code block below that my idea is to have an array with the function signatures and with the help of it we get the function selector. So in the future any new system contract methods can be easily added. Is this what you mean?

natanasow (dismissed):

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,62 @@
...
+## Problem Statement
+When users create tokens (fungible and non-fungible) by directly calling the 0x167 contract there is no way for them to get the evm address of the newly created token via standard EVM tooling e.g ethers. This is because the transaction receipt's contractAddress field is not being populated with the HTS token address, since essentially the operations is just a contract call, but not a contract create (deploy).
+
+## Current Behaviour
+
+Currently, when creating a token via HTS the address of the system contract is returned as a contractAddress in the transaction receipt.
+
+## Proposed Solution
+
+### Technical Details
```
quiet-node, line 16: Is it possible to mention somewhere the endpoints are being modified for this logic?

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,62 @@
...
+
+### 1. Function Signature Detection
+
+Detect in the transactionResponse from the mirror node, if the transaction was calling any of the HTS methods - createFungibleToken/createNonFungibleToken/createFungibleTokenWithCustomFees/createNonFungibleTokenWithCustomFees
```
quiet-node, line 20: `if the transaction was calling any of the HTS methods` this could be a little misleading as the listed methods are only about token creation. 
```suggestion
Detect in the transactionResponse from the mirror node, if the transaction was calling any of the token creation HTS methods - createFungibleToken/createNonFungibleToken/createFungibleTokenWithCustomFees/createNonFungibleTokenWithCustomFees
```

docs/design/hts_address_tx_receipt.md
quiet-node: qq: I see Nana's suggestion above was marked as resolved but didn't see the changes nor updates. What's the approach here?

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,62 @@
...
+
+N.B Currently, HTS supports both v1 and v2 security model function selectors
+
+```javascript
+const HTS_CREATE_FUNCTIONS_SIGNATURE = [
+  "createFungibleToken(...)",
+  "createNonFungibleToken(...)",
+  "createFungibleTokenWithCustomFees(...)",
+  "createNonFungibleTokenWithCustomFees(...)"
+];
```
quiet-node, line 30: If possible, let's include the correct params for these methods instead of `(...)`

docs/design/hts_address_tx_receipt.md
quiet-node: same here suggestion was marked resolve but no updates nor comments

quiet-node (commented):
Some questions and suggestions

docs/design/hts_address_tx_receipt.md
konstantinabl: Well, this affects the getTransactionReceipt endpoint only
As it is mentioned in a few places that we want to return it there?
Could you suggest a specific place?

docs/design/hts_address_tx_receipt.md
konstantinabl: I think the purpose of this sentence is to name the specific methods in the end.
If you really want it to be clear i can change it to
-
```
Detect in the transactionResponse from the mirror node, if the transaction was calling any of the FOLLOWING HTS methods - createFungibleToken/createNonFungibleToken/createFungibleTokenWithCustomFees/createNonFungibleTokenWithCustomFees
```
As Nana mentioned in another comment this may change to include other methods as well, so i am reluctant to specifying if they are token creation or anything else. What do you think

docs/design/hts_address_tx_receipt.md
konstantinabl: yes, the commit for some reason got lost, because i didnt pull it locally and then forced push

docs/design/hts_address_tx_receipt.md
quiet-node: Ahh okay I see. I would suggest creating a dedicated section to clarify that the feature focuses on specific endpoint. As when I read through the doc, I was wondering what endpoints are being focused. So perhaps we could add a section and title it "Affected Endpoints," "Impacted Endpoints," "Targeted Endpoints," or something like that.

docs/design/hts_address_tx_receipt.md
quiet-node: Nice! I like it! Thanks for the update!

docs/design/hts_address_tx_receipt.md
quiet-node: Okay thanks for confirming!

quiet-node (approved):
LGTM just [one suggestion](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3258#discussion_r1848601401).

docs/design/hts_address_tx_receipt.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Current Behaviour
+
+Currently, when creating a token via HTS the address of the system contract is returned as a contractAddress in the transaction receipt.
+
+## Proposed Solution
+
+### Targeted endpoints
+
+[getTransactionReceipt]
```
quiet-node, line 18: Perhaps below.
```suggestion
### Targeted endpoints

- eth_getTransactionReceipt
```

quiet-node (approved):
LGTM! Thanks much for the good work!

natanasow (approved):

nadezhdapopovaa (approved):

PR #3261, opened 2024-11-14
feat: add eslint to be run pre commit

### LINKED ISSUE #3064: Add eslint to be run pre-commit

konstantinabl:
**Description**:

Added the `eslint --fix` command to lint-staged and husky. This means that whenever a staged file is committed it will be checked by eslint for any broken rules and they will be fixed, if possible. (some rules are not fixable, like unused-vars for example).

**Related issue(s)**:

Fixes #3064 

**Notes for reviewer**:

Note the following things:
1. eslint.yml file was deleted, since we shouldn't have two eslint config files
2. There are changes in sdkClient.spec.ts, since eslint --fix was tested on it


Nana-EC (dismissed):
Nice

quiet-node (approved):
LGTM thanks for the good work

Nana-EC (approved):

victor-yanev (approved):

Issue #3266, opened 2024-11-15
Relay is logging IP addresses


quiet-node:
### Problem

Currently, ipaddress are encapsulated in the RequestDetail class and being logged out here and there in the Relay.

### Solution

Redacted all the ipaddresses being logged out

PR #3268, opened 2024-11-15
fix: masks ip address in request details passed as argument to MN client

konstantinabl:
**Description:**

The issue arises from the fact requestDetails are passed as an argument to the repeatedRequest method in MN client, because they are needed for the next method call

However, further down in the method the args are logged.

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-json-rpc-relay/issues/3266

### DIFF
--- a/packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
+++ b/packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
@@ -39,7 +39,7 @@ import sinon from 'sinon';
 
 import { Eth, JsonRpcError, predefined } from '../../../src';
 import { formatTransactionIdWithoutQueryParams } from '../../../src/formatters';
-import { MirrorNodeClient, SDKClient } from '../../../src/lib/clients';
+import { SDKClient } from '../../../src/lib/clients';
 import constants from '../../../src/lib/constants';
 import { SDKClientError } from '../../../src/lib/errors/SDKClientError';
 import { CacheService } from '../../../src/lib/services/cacheService/cacheService';
@@ -63,13 +63,11 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
     hapiServiceInstance,
     ethImpl,
     cacheService,
-    mirrorNodeInstance,
   }: {
     restMock: MockAdapter;
     hapiServiceInstance: HAPIService;
     ethImpl: Eth;
     cacheService: CacheService;
-    mirrorNodeInstance: MirrorNodeClient;
   } = generateEthTestEnv();
 
   const requestDetails = new RequestDetails({ requestId: 'eth_sendRawTransactionTest', ipAddress: '0.0.0.0' });


packages/relay/src/lib/eth.ts
```diff
@@ -1722,7 +1722,7 @@ export class EthImpl implements Eth {
         const formattedTransactionId = formatTransactionIdWithoutQueryParams(submittedTransactionId);
         const contractResult = await this.mirrorNodeClient.repeatedRequest(
           this.mirrorNodeClient.getContractResult.name,
-          [formattedTransactionId, requestDetails],
+          [formattedTransactionId],
```
Nana-EC, line 1725: Just checking is this the only place to remove it?
Other approach is to set IP undefined after it's utilized in the rate limiter if possible.

konstantinabl: No, its not the only place, i think there is one more, working on it
I will check your suggestion as well

quiet-node (dismissed):
LGTM! Thanks for the quick great work!

packages/relay/src/lib/eth.ts
```diff
@@ -1720,9 +1721,16 @@ export class EthImpl implements Eth {
     if (submittedTransactionId) {
       try {
         const formattedTransactionId = formatTransactionIdWithoutQueryParams(submittedTransactionId);
+
+        // Create a modified copy of requestDetails
+        const modifiedRequestDetails = {
```
Nana-EC, line 1726: This works but is there a place where we can do this once before it's passed into the eth.ts but after it's utilized locally?

Nana-EC (commented):
Looking good but I won't if there is a single location we can set this oce

packages/relay/src/lib/eth.ts
konstantinabl: Well, I haven't come up with anything, cause actually we use the ip address in the shouldLimit method, the flow is the following
sendRawTransaction -> submitEthereumTransaction -> executeTransaction -> shouldLimit
So if i change it before sendRawTransaction a.k.a in eth.ts it will be changed for shouldLimit as well
Also I did it in the fastest way possible without affecting a lot of tests, because i thought it should go to mainnet asap

packages/relay/src/lib/eth.ts
```diff
@@ -1720,9 +1721,16 @@ export class EthImpl implements Eth {
...
+          ...requestDetails,
+          ipAddress: 'xxx.xxx.xxx.xxx',
```
natanasow, line 1728: The format can be defined in a constant file

packages/relay/src/lib/eth.ts
natanasow, line 1726: The variable is used only once so we can skip its declaration and pass the object straight to the `repreatRequest` method.

packages/relay/src/lib/eth.ts
```diff
@@ -1720,9 +1721,16 @@ export class EthImpl implements Eth {
...
+        };
+
         const contractResult = await this.mirrorNodeClient.repeatedRequest(
```
natanasow, line 1731: Can we add a unit test?

natanasow (dismissed):
Approved with some nits which might be fixed in a follow-up PR due to the urgency of this one.

victor-yanev (approved):

packages/relay/src/lib/eth.ts
konstantinabl: Done.

natanasow (approved):

nadezhdapopovaa (approved):

quiet-node (approved):

PR #3272, opened 2024-11-17
fix: updated precheck.nonce() to handle higher nonce cases

### LINKED ISSUE #3271: `precheck.nonce()` needs to also handle higher nonce cases
### Problem

Currently, `precheck.nonce()` only rejects requests with a lower nonce. Requests with a higher nonce should also be rejected with the value `NONCE_TOO_HIGH`.

### Solution

Update prechec.nonce() to handle higher nonce cases

### Alternatives

_No response_

quiet-node:
**Description**:
This PR updates precheck.nonce() to reject requests that have nonces higher than current nonce of the signer

**Related issue(s)**:

Fixes #3271 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/precheck.ts
+++ b/packages/relay/src/lib/precheck.ts
@@ -18,13 +18,14 @@
  *
  */
 
-import { JsonRpcError, predefined } from './errors/JsonRpcError';
-import { MirrorNodeClient } from './clients';
-import { EthImpl } from './eth';
-import { Logger } from 'pino';
-import constants from './constants';
 import { ethers, Transaction } from 'ethers';
+import { Logger } from 'pino';
+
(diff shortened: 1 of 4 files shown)


Nana-EC (changes requested):
We shouldn’t reject too high because we don’t know how transactions are being asynchronously submitted by users and therefore a transaction with a high nonce at a particular relays precheck could still be correct at the time of network processing.

quiet-node:
> We shouldn’t reject too high because we don’t know how transactions are being asynchronously submitted by users and therefore a transaction with a high nonce at a particular relays precheck could still be correct at the time of network processing.

Yes copy that! Thanks for confirming! Closing this PR as misunderstanding the design.

PR #3273, opened 2024-11-18
feat: enabled fast return on eth_sendRawTransaction

### LINKED ISSUE #3270: Add feature flag and move all transaction processing logic in background if the feature flag is enabled.

quiet-node:
### **Description**:  
The `eth_sendRawTransaction` endpoint currently faces timeouts from various sources, including server, network, and client settings. For large callData, it uses Hedera File Service (HFS) to create and append file transactions, increasing latency and degrading user experience.  

This PR introduces a new concepture feature design for `eth_sendRawTransaction`, enabling the API to return the transaction hash immediately upon passing prechecks. All subsequent processing (e.g., broadcasting to the network, handling file transactions, polling the mirror node, tracking HBAR consumption, etc.) is handled asynchronously in the background. 

The primary impact of this feature is a significant reduction in end-to-end client wait times while ensuring backend integrity remains intact.

### **Important Note for Clients**:  
With the transaction hash returned instantly, clients must monitor the transaction status using this hash (e.g., via `eth_getTransactionReceipt` or monitoring confirmations).  

If a request encounters internal errors (e.g., exceeding `maxChunkSize`, HBAR limits, SDK errors), these will no longer appear in the response. Instead, a transaction hash is returned, and `eth_getTransactionReceipt` may return null indefinitely. That said, it remains the responsibility of the users to handle such cases appropriately.

Fixes #3270 

### **Related issue(s)**: 
Additional aspects and considerations of the feature are highlighted in the ongoing discussion in #3202. Please feel free to share your insights in the thread if you have any.

### **Notes for reviewer**:
TL;DR; The primary change in this PR is a refactor, mainly within `eth.ts`. The remaining work involves updating tests to align with the new immediate return logic. Most efforts are directed toward adding polling logic to verify the validity of transactions before proceeding to the next instruction.

- The feature is designed for backward compatibility, allowing it to be toggled on or off while ensuring the end-to-end flow remains consistent.  
- A new global configuration, `USE_ASYNC_TX_PROCESSING`, acts as the feature flag for this functionality.  
- The `EthImpl.sendRawTransaction` method has been refactored to extract transaction processing logic into a new function, `sendRawTransactionTxProcessor()`. When `USE_ASYNC_TX_PROCESSING` is enabled, `sendRawTransactionTxProcessor()` is called without `await`, allowing transaction processing to run asynchronously in the background. If the flag is disabled, the function is awaited, maintaining the original synchronous flow.  
- The remainder of the work focuses on updating tests to accommodate the new feature.  

### **Checklist**


packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -435,30 +436,32 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+                fromBlock: numberTo0x(blocksBehindLatest),
+                toBlock: 'latest',
+                address: [contractAddress, contractAddress2],
+              },
+            ],
+            requestIdPrefix,
+          );
+          expect(logs.length).to.be.greaterThan(2);
+        });
+      }
```
quiet-node, line 464: Note for reviewers: currently disable this test of fast return is turned on.

Open to discussion: I have been testing this particular test multiple times and discovered that, for some reason, when the fast return logic is enabled, more blocks are generated, which increases the total range of blocks being processed. Consequently, the original tests that access the last 10 blocks do not provide enough logs (more than 2, as required by the test).

This issue could be easily be resolved by increasing the total log access from 10 to 15 or 20. 

However, it is more important to address the underlying logic of how faster returns generate more blocks.

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -1398,22 +1396,27 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+          const error = predefined.INTERNAL_ERROR();
 
-        await Assertions.assertPredefinedRpcError(error, sendRawTransaction, false, relay, [signedTx, requestDetails]);
-      });
+          await Assertions.assertPredefinedRpcError(error, sendRawTransaction, false, relay, [
+            signedTx,
+            requestDetails,
+          ]);
+        });
+      }
```
quiet-node, line 1419: Note for reviewers: currently disable this test of fast return is turned on.

Open for Discussion: This test submits an `eth_sendRawTransaction` request with a calldata size of 60,000 bytes. Originally, this test would throw an SDKError indicating that the maximum file chunk size exceeds the limit of 20. However, with fast return enabled, no errors are thrown, only precheck errors. Currently, prechecks do not account for this.

The question is: what is a suitable solution to update this test? Should we revise the logic to state, "It should still return a transaction hash if the calldata exceeds the maximum file chunk size," since errors are no longer thrown when fast return is enabled? Or should we retain the existing logic and assert that the transaction receipt retrieved using the returned transaction hash will be null?

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -1544,21 +1547,26 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+
+            const signedTx = await accounts[2].wallet.signTransaction(transaction);
+            const error = predefined.NONCE_TOO_HIGH(nonce + 100, nonce);
+
+            await Assertions.assertPredefinedRpcError(error, sendRawTransaction, true, relay, [
+              signedTx,
+              requestDetails,
+            ]);
+          });
+        }
```
quiet-node, line 1569: Note for reviewers: currently disable this test of fast return is turned on.

Open for discussion: With fast return enabled, no errors are thrown—only precheck errors. Currently, prechecks do not account for higher nonce transactions.

Should we remove the test case for higher nonce transactions, or should we retain the existing logic and assert that the transaction receipt retrieved using the returned transaction hash will be null?

quiet-node (commented):
Left some notes and questions for reviewers.

victor-yanev (approved):

packages/server/tests/acceptance/rpc_batch1.spec.ts
natanasow: Just random thoughts.

In the Ethereum ecosystem, there is no such thing as nonce too high. We can submit a transaction with whatever nonce above the current one we want, and the transaction will reach the mempool. Once it's in the mempool, the validator will pick the transaction with a nonce of current + 1, so if there are enough transactions filling the gap between, the transaction will be executed in the near future. They always will be executed ordered by nonce.

On another hand, in the Hedera ecosystem, there is no such thing as mempool. That means if we send a transaction, it's being sent straight to the node for its execution. It turned out that if we asynchronously send N transactions without timeout between them, there is a high chance of failures due to missing "native" nonce ordering like in Ethereum. If we off-chain "ordered" them with some timeout between them (e.g. 50ms), they might succeed if there is no unexpected race condition.

Also, we should think about submitting multiple off-chain ordered transactions within a single block. In that case, if there is a nonce too high check precheck, we will limit the user to submit only one transaction per block because "the expected" nonce is fetched from the mirror node where it's up-to-date with the latest block, not the current one.

IMO nonce too high precheck must not be added, and throwing nonce too high as an error might be misleading for the end-user when multiple transactions are submitted and can be removed (or changed to WRONG_NONCE which is the actual error from the services). Also, I'm not sure the `NONCE_TOO_HIGH` error even exists in the Ethereum JSON-RPC context.

@Nana-EC @quiet-node What do you think? Does it make sense to you to rename or remove the current nonce too high error?

quiet-node: Hey @natanasow, thanks for the comment! That makes complete sense. In fact, on the Relay, we agreed not to reject requests with higher nonces anymore and removed the `NONCE_TOO_HIGH` precheck. This test is believed to be a leftover from that work. 

Currently, the test still passes and throws a `NONCE_TOO_HIGH` error, but that happens because the transaction is rejected at the Services level, not on the Relay. Since there’s no longer a high nonce precheck on the Relay, the transaction gets forwarded to Services, where it is ultimately rejected with a `WRONG_NONCE` status code. The Relay then interprets this error, compares the current account nonce, and throws a `NONCE_TOO_HIGH` error when the transaction nonce is higher than the account nonce. As a result, the test passes with the expected error but with an incorrect underlying intention.

Now the point you've raised is indeed a solid and exellent point showing a gap in our EVM equivalence goals. However, since the Relay no longer enforces the `NONCE_TOO_HIGH` check and the rejection occurs at the Services level, it might be more appropriate to move this conversation to Services to bring broader attention to this gap.

natanasow:
LG, several tests are failing due to an `undefined variable`. Let's fix them before the final approvals.

packages/server/tests/acceptance/rpc_batch1.spec.ts
konstantinabl: Maybe we should check that the transaction receipt shows that the transaction was not successful due to chunk size
@quiet-node

packages/relay/src/lib/eth.ts
```diff
@@ -1638,24 +1636,39 @@ export class EthImpl implements Eth {
...
+   * @async
+   * @param {Buffer} transactionBuffer - The raw transaction data as a buffer.
+   * @param {EthersTransaction} parsedTx - The parsed Ethereum transaction object.
+   * @param {number} networkGasPriceInWeiBars - The current network gas price in wei bars.
+   * @param {RequestDetails} requestDetails - Details of the request for logging and tracking purposes.
+   * @returns {Promise<string | JsonRpcError>} A promise that resolves to the transaction hash if successful, or a JsonRpcError if an error occurs.
+   * @throws {JsonRpcError} If there's an error during transaction processing.
    */
-  async sendRawTransaction(transaction: string, requestDetails: RequestDetails): Promise<string | JsonRpcError> {
+  async sendRawTransactionTxProcessor(
```
konstantinabl, line 1660: nit: may you fix this

quiet-node:
> LG, several tests are failing due to an `undefined variable`. Let's fix them before the final approvals.

@natanasow Thanks for the good catch! Missed that during rebasing! Updated!

konstantinabl (approved):

packages/server/tests/acceptance/rpc_batch1.spec.ts
quiet-node: @Nana-EC Should I keep this test (which is only enabled when fast-return is off) for additional coverage, or should I remove it to clarify that the Relay no longer fails at higher nonce values?

packages/relay/src/lib/eth.ts
quiet-node: Updated! Thanks for the catch!

natanasow (approved):
LGTM. Thanks for the great work!

PR #3285, opened 2024-11-21
chore: extend the format checking via openrpcjson

### LINKED ISSUE #2728: Extend the format checking via openrpc.json
Currently, in this [PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/2649) we use the openprc.json from this [repo](https://github.com/ethereum/execution-apis/tree/main/tests) in order to check the format of our responses and if they are compliant with ethereum. We should create an openprc.json with the formats as shown in other relays we conform to (Infura, Alchemy) and compare our response formats

nadezhdapopovaa:
**Description**:
- updates conformity test to use relay's openrpc file for methods, which are not tested via ethereum/execution-api repo. We check if result is matching the pattern from our openrpc file;
- there were few methods which were duplicated and update of few patterns.

**Related issue(s)**:

Fixes #2728

**Checklist**

### DIFF
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -944,7 +944,7 @@
         "name": "The current client version.",
         "schema": {
           "type": "string",
-          "pattern": "relay/[0-9]\\.[0-9]\\.[0-9]"
+          "pattern": "relay/[0-9]+\\.[0-9]+\\.[0-9]+(-[a-zA-Z0-9-]+)?$"
         }
       }
     },
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -943,16 +943,8 @@
       "result": {
         "name": "The current client version.",
         "schema": {
-          "anyOf": [
-            {
-              "type": "string",
-              "pattern": "relay\\/[0-9]+\\.[0-9]+\\.[0-9]+-SNAPSHOT"
-            },
-            {
-              "type": "string",
-              "pattern": "relay/[0-9]\\.[0-9]\\.[0-9]"
-            }
-          ]
+          "type": "string",
+          "pattern": "relay/[0-9]\\.[0-9]\\.[0-9]"
         }
       }
     },
@@ -1886,7 +1878,7 @@
         "oneOf": [
           {
             "title": "Any Topic Match",
-            "type": "null" 
+            "type": "null"
           },
           {
             "title": "Single Topic Match",
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -57,7 +57,9 @@
         "name": "Accounts",
         "description": "Always returns an empty array",
         "schema": {
-          "$ref": "#/components/schemas/addresses"
+          "title": "empty array",
+          "type": "array",
+          "pattern": "^\\[\\s*\\]$"
         }
       }
     },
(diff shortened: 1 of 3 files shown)


docs/openrpc.json
```diff
@@ -651,8 +657,9 @@
       "description": "![](https://raw.githubusercontent.com/hashgraph/hedera-json-rpc-relay/main/docs/images/http_label.png) ![](https://raw.githubusercontent.com/hashgraph/hedera-json-rpc-relay/main/docs/images/ws_label.png)",
       "params": [],
       "result": {
-        "name": "Always returns '0x0'. Hedera doesn't have a concept of tipping nodes to promote any behavior",
+        "name": "eth_maxPriorityFeePerGas result",
```
quiet-node, line 660: QQ: I see the name has been changed around. Is this because we want the name to be explicit or because we have to align with the specification?

docs/openrpc.json
```diff
@@ -1015,120 +1031,6 @@
...
-        "name": "logs",
-        "schema": {
-          "type": "array",
-          "items": {
-            "$ref": "#/components/schemas/Log"
-          }
-        },
-        "description": "Array of log objects that match the filter."
-      }
-    },
```
quiet-node, line 1131: qq why are we removing these?

quiet-node (approved):
LGTM some QQs

docs/openrpc.json
nadezhdapopovaa: - decided to change it to be consistent with the ethereum one and ours more or less;
- visually it's too long;
- in playground it makes more sense when you open the result to have additional info about it together with the type and pattern;
- and in my opinion this sentence is a some kind of a description or additional info about the result, not the title.

docs/openrpc.json
nadezhdapopovaa: the schema for those methods was written twice; here we're removing one of them --> compared both to ethereum specification and left the one consistent with it

docs/openrpc.json
quiet-node: nice thanks for confirming!

docs/openrpc.json
quiet-node: nice thanks for explaning!

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -242,11 +244,11 @@ function checkResponseFormat(actualReponse, expectedResponse) {
   }
 }
 
-function findSchema(file) {
+const findSchema = function (file) {
```
konstantinabl, line 247: is there a specific reason we changed thi?

docs/openrpc.json
```diff
@@ -936,8 +943,16 @@
       "result": {
         "name": "The current client version.",
         "schema": {
-          "type": "string",
-          "pattern": "relay/[0-9]\\.[0-9]\\.[0-9]"
+          "anyOf": [
+            {
+              "type": "string",
+              "pattern": "relay\\/[0-9]+\\.[0-9]+\\.[0-9]+-SNAPSHOT"
```
Nana-EC, line 949: Q: if you added one for snapshot do we need one for rc versions?

docs/openrpc.json
```diff
@@ -1015,120 +1031,6 @@
         "description": "True if the subscription was successfully cancelled, otherwise false."
       }
     },
-    {
-      "name": "eth_newFilter",
```
Nana-EC, line 1019: Q: why are we removing these filter endpoints?

Nana-EC (commented):
Looking good, some questions

packages/server/tests/acceptance/conformityTests.spec.ts
nadezhdapopovaa: I was testing different approaches. with the one uploaded we don't need it, so I'll revert the change. thanks!

docs/openrpc.json
nadezhdapopovaa: decided to write a check for the beginning of the string and that way we'll have one pattern for all

konstantinabl (approved):

victor-yanev (approved):

PR #3292, opened 2024-11-22
feat: expose node account id in relay logging of failed transactions

### LINKED ISSUE #3278: Expose Node Account ID in Relay Logging for Failed Transactions
### Problem

The Relay server currently lacks visibility into which node account ID a transaction was submitted to when it fails due to an SDK error. With the [recent update](https://github.com/hashgraph/hedera-sdk-js/pull/2631) to the `hedera-sdk-js`, the error object now includes the node account ID for failed transactions. 

This new capability in the Relay's logging system is expected to improve troubleshooting and provide better insights into node-specific transaction failures.

### Solution

Update the Relay server's logging system to extract and log the node account ID from the error object provided by the SDK

### Alternatives

_No response_

natanasow:
**Description**:

The Relay server currently lacks visibility into which node account ID a transaction was submitted to when it fails due to an SDK error. With the [recent update](https://github.com/hashgraph/hedera-sdk-js/pull/2631) to the `hedera-sdk-js`, the error object now includes the node account ID for failed transactions. 

This new capability in the Relay's logging system is expected to improve troubleshooting and provide better insights into node-specific transaction failures.

**Solution**:

Update the Relay server's logging system to extract and log the node account ID from the error object provided by the SDK

**Related issue(s)**:

Fixes #3278 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/package.json
+++ b/packages/relay/package.json
@@ -50,7 +50,7 @@
   "dependencies": {
     "@hashgraph/json-rpc-config-service": "file:../config-service",
     "@ethersproject/asm": "^5.7.0",
-    "@hashgraph/sdk": "^2.53.0",
+    "@hashgraph/sdk": "^2.54.0-beta.1",
     "@keyvhq/core": "^1.6.9",
     "axios": "^1.4.0",
     "axios-retry": "^3.5.1",
--- a/packages/relay/package.json
+++ b/packages/relay/package.json
@@ -50,7 +50,7 @@
   "dependencies": {
     "@hashgraph/json-rpc-config-service": "file:../config-service",
     "@ethersproject/asm": "^5.7.0",
-    "@hashgraph/sdk": "^2.50.0-beta.3",
+    "@hashgraph/sdk": "^2.53.0",
     "@keyvhq/core": "^1.6.9",
     "axios": "^1.4.0",
     "axios-retry": "^3.5.1",
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -728,6 +728,13 @@ export class SDKClient {
         throw e;
       }
 
+      // Log the target node account ID, right now, it's populated only for MaxAttemptsOrTimeout error
+      if (e?.nodeAccountId) {
+        this.logger.info(
+          `${requestDetails.formattedRequestId} Transaction failed to execute against node with id: ${e.nodeAccountId}`
+        );
+      }
+
       const sdkClientError = new SDKClientError(e, e.message, transaction.transactionId?.toString());
 
       // WRONG_NONCE is one of the special errors where the SDK still returns a valid transactionResponse.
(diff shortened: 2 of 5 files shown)


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -728,6 +728,13 @@ export class SDKClient {
         throw e;
       }
 
+      // Log the target node account ID, right now, it's populated only for MaxAttemptsOrTimeout error
+      if (e?.nodeAccountId) {
+        this.logger.info(
+          `${requestDetails.formattedRequestId} Transaction failed to execute against node with id: ${e.nodeAccountId}`
+        );
+      }
+
```
quiet-node, line 737: This is in executeTransaction, should we do it for executeAllTransaction as well? 

Or actually maybe in a more general place like in the EthImpl.sendRawTransactionErrorHandler()?

packages/relay/package.json
```diff
@@ -50,7 +50,7 @@
   "dependencies": {
     "@hashgraph/json-rpc-config-service": "file:../config-service",
     "@ethersproject/asm": "^5.7.0",
-    "@hashgraph/sdk": "^2.50.0-beta.3",
+    "@hashgraph/sdk": "^2.53.0",
```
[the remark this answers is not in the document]
natanasow: Bumped, thank you!

packages/relay/src/lib/clients/sdkClient.ts
natanasow: Added in the `executeAllTransaction`, good catch!

`sendRawTransactionErrorHandler` uses the `SDKClientError` wrapper which doesn't have the plain SDK error object. I'm thinking about adding it but I decided it's not worth it adding more complexity for just one log.

quiet-node: Ah okay that makes sense! Okay executeTransaction and executeAllTransaction should be enough then as most of the methods in SDKClient class would eventually reroute to those two execution methods! Thanks for the explanation!

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -728,6 +728,13 @@ export class SDKClient {
         throw e;
       }
 
+      // Log the target node account ID, right now, it's populated only for MaxAttemptsOrTimeout error
+      if (e?.nodeAccountId) {
```
Nana-EC, line 732: This is okay given the beta version and end of sprint but I'd prefer we move this logic into `SDKError` and add the node to the existing error logs

Nana-EC (approved):
LG.
Please open a ticket to move logicinto SDKError & unpdate existing error logs later

packages/relay/src/lib/clients/sdkClient.ts
Nana-EC: @nickeynikolovv FYI

PR #3304, opened 2024-11-28
fix: let 0 value transactions pass

### LINKED ISSUE #3105: Let 0-value Ethereum transactions pass
### Problem

Currently, the JSON-RPC relay forces the Ethereum transaction value to be greater than 0 in order to be accepted. 

This isn't a real network constraint, both Hedera and Ethereum accept 0-value transactions, but it is enforced in two parts of the code (during the [validate API request](https://github.com/hashgraph/hedera-json-rpc-relay/blob/7796aca3d8028e9732c58fb6855bef4520c4b3cc/packages/relay/src/lib/eth.ts#L643) and during the [precheck transaction](https://github.com/hashgraph/hedera-json-rpc-relay/blob/7796aca3d8028e9732c58fb6855bef4520c4b3cc/packages/relay/src/lib/precheck.ts#L64)).

I removed these checks and was able to send a [correct 0-value transaction](https://hashscan.io/testnet/tx/0x29d9264df7244583e69bdb41822887df2511b6ce8b10463280bc8260abb3c778) to the network.

### Solution

- Let the JSON-RPC relay operator decide whether to accept 0-value transactions.
- Set the default for this parameter to `true`: 0-value transactions are accepted.

### Alternatives

_No response_

simzzz:
**Description**:
This PR fixes the case where transactions with a value of 0 would not pass. This is needed because they would pass as valid transactions on the Ethereum network as 0 is a valid value there.

**Related issue(s)**:

Fixes #3105 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -162,7 +162,7 @@ export const predefined = {
   }),
   VALUE_TOO_LOW: new JsonRpcError({
     code: -32602,
-    message: "Value can't be negative or between 1 wei and 10_000_000_000 wei which is 1 tinybar",
+    message: "Value can't be non-zero and less than 10_000_000_000 wei which is 1 tinybar",
   }),
   INVALID_CONTRACT_ADDRESS: (address) => {
     let message = `Invalid Contract Address: ${address}.`;
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -162,7 +162,7 @@ export const predefined = {
   }),
   VALUE_TOO_LOW: new JsonRpcError({
     code: -32602,
-    message: 'Value below 10_000_000_000 wei which is 1 tinybar',
+    message: "Value can't be negative or between 1 wei and 10_000_000_000 wei which is 1 tinybar",
   }),
   INVALID_CONTRACT_ADDRESS: (address) => {
     let message = `Invalid Contract Address: ${address}.`;
--- a/packages/relay/src/lib/precheck.ts
+++ b/packages/relay/src/lib/precheck.ts
@@ -18,13 +18,14 @@
  *
  */
 
-import { JsonRpcError, predefined } from './errors/JsonRpcError';
-import { MirrorNodeClient } from './clients';
-import { EthImpl } from './eth';
-import { Logger } from 'pino';
-import constants from './constants';
 import { ethers, Transaction } from 'ethers';
+import { Logger } from 'pino';
+
 import { prepend0x } from '../formatters';
+import { MirrorNodeClient } from './clients';
+import constants from './constants';
+import { JsonRpcError, predefined } from './errors/JsonRpcError';
+import { EthImpl } from './eth';
 import { RequestDetails } from './types';
 
 /**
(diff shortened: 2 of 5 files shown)


packages/relay/src/lib/precheck.ts
```diff
@@ -56,16 +57,6 @@ export class Precheck {
...
 
-  /**
-   * Checks if the value of the transaction is valid.
-   * @param {Transaction} tx - The transaction.
-   */
-  value(tx: Transaction): void {
-    if (tx.data === EthImpl.emptyHex && tx.value < constants.TINYBAR_TO_WEIBAR_COEF) {
-      throw predefined.VALUE_TOO_LOW;
-    }
-  }
```
vyanev1, line 67: I think we still want to check that the `tx.value` is not negative, instead of completely removing the pre-check

vyanev1 (changes requested):
LG, one concern:

packages/relay/src/lib/precheck.ts
victor-yanev, line 67: I think we still want to confirm that the value is not negative instead of completely removing the pre-check

victor-yanev (changes requested):
LG, just once concern:

packages/relay/src/lib/precheck.ts
simzzz: good point, the value shouldn't be below under 1 tinybar but 0 should be a valid value. I will push a new commit along with new tests for this

nadezhdapopovaa (dismissed):

packages/relay/src/lib/precheck.ts
```diff
@@ -61,7 +62,10 @@ export class Precheck {
    * @param {Transaction} tx - The transaction.
    */
   value(tx: Transaction): void {
-    if (tx.data === EthImpl.emptyHex && tx.value < constants.TINYBAR_TO_WEIBAR_COEF) {
+    if (
+      tx.data === EthImpl.emptyHex &&
+      ((tx.value > 0 && tx.value < constants.TINYBAR_TO_WEIBAR_COEF) || tx.value < 0)
+    ) {
```
victor-yanev, line 68: Don't we also want to perform this pre-check on non-zero data transaction? 

What if the solidity method is marked as payable and we pass a negative value or a value between 0 and `TINYBAR_TO_WEIBAR_COEF`? Don't we want to confirm that the passed value is correct also in this case?

victor-yanev (changes requested):

victor-yanev (approved):

nadezhdapopovaa (dismissed):

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -162,7 +162,7 @@ export const predefined = {
   }),
   VALUE_TOO_LOW: new JsonRpcError({
     code: -32602,
-    message: 'Value below 10_000_000_000 wei which is 1 tinybar',
+    message: "Value can't be negative or between 1 wei and 10_000_000_000 wei which is 1 tinybar",
```
Nana-EC, line 165: nit: Maybe this is more clear
```suggestion
    message: "Value can't be non-zero and less than 10_000_000_000 wei which is 1 tinybar",
```

Nana-EC (dismissed):
LG, one nit

quiet-node (approved):
Great work! LGTM

nadezhdapopovaa (approved):

Issue #3030, opened 2024-09-26
Ethers v6 Filter API not working with 0.55.0


tknoch-symphony:
### Description

While running our instance of the latest version of Hedera JSON-RPC node (0.55.0), filter api does not work, even when filter api flag is turned on. The client calls are made by ethers v6.

```
@TODO Error: could not coalesce error (error={ "code": -32001, "message": "[Request ID: 038fa6e7-6804-47e9-8c36-a2a43f4fc4a0] Filter not found", "name": "Filter not found" }, payload={ "id": 74408, "jsonrpc": "2.0", "method": "eth_getFilterChanges", "params": [ "0x0a7c11483e1472a3cf6836580d1c0e50" ] }, code=UNKNOWN_ERROR, version=6.13.2)
    at makeError (/node_modules/ethers/lib.commonjs/utils/errors.js:129:21)
    at JsonRpcProvider.getRpcError (/node_modules/ethers/lib.commonjs/providers/provider-jsonrpc.js:730:41)
    at /node_modules/ethers/lib.commonjs/providers/provider-jsonrpc.js:302:45
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'UNKNOWN_ERROR',
  error: {
    code: -32001,
    name: 'Filter not found',
    message: '[Request ID: 038fa6e7-6804-47e9-8c36-a2a43f4fc4a0] Filter not found'
  },
  payload: {
    method: 'eth_getFilterChanges',
    params: [ '0x0a7c11483e1472a3cf6836580d1c0e50' ],
    id: 74408,
    jsonrpc: '2.0'
  },
  shortMessage: 'could not coalesce error'
}
```

### Steps to reproduce

1. install Hedera JSON-RPC node (0.55.0). Turn on filter api
2. Use ethers v6 to subscribe to a smart contract event filter
3. See errors in Hedera JSON-RPC node logs mentioning filter api is not implemented

### Hedera network

testnet

### Version

0.55.0

### Operating system

Linux

Nana-EC:
Thanks @tknoch-symphony 
We'll take a look into this, let us know if you're still experiencing this or have additional and recent data.

simzzz:
you can assign me

simzzz:
Hello, @tknoch-symphony I have investigated your issue using the versions provided and I had no problems accessing the Filter API, however I could reproduce your exact error in the following scenarios:

[...]

PR #3308, opened 2024-12-05
fix: added cache refreshing after subsequent getFilterLogs calls for the same filter

### LINKED ISSUE #3311: Cache doesn't get refreshed after subsequent calls to getFilterLogs
### Problem

When a call to `getFilterChanges` is made, we refresh the Filter's TTL in the cache which is expected, however we don't for `getFilterLogs `. I think we should have consistent behavior across such methods.

### Solution

- Refresh the cache after calling `getFilterLogs` as well

### Alternatives

_No response_

simzzz:
**Description**:

During debugging of #3030 an inconsistency was found where we automatically refresh the TTL for our filter cache entries for `getFilterChanges` but not for `getFilterLogs`. This PR solves this inconsistency.

**Related issue(s)**:
 #3030
Fixes #3311 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/services/ethService/ethFilterService/index.ts
+++ b/packages/relay/src/lib/services/ethService/ethFilterService/index.ts
@@ -18,17 +18,18 @@
  *
  */
 
+import { ConfigService } from '@hashgraph/json-rpc-config-service/dist/services';
 import { Logger } from 'pino';
+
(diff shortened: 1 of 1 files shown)


nadezhdapopovaa (approved):

quiet-node (approved):
LGTM

PR #3310, opened 2024-12-05
chore: add precheck for receivers account in sendRawTransactionCheck

### LINKED ISSUE #3286: Add prechecks to the `sendRawTransaction` method to verify the validity of the receiver's account address.
### Problem

For crypto transfer transactions submitted through eth_sendRawTransaction endpoint, the Relay lacks prechecks to verify the validity of the receiver account. 

### Solution

Add precheck methods to verify the validity of the receiver account

### Alternatives

_No response_

nadezhdapopovaa:
**Description**:
Added precheck for receiver's account in sendRawTransactionCheck

Fixes #3286

### DIFF
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -238,6 +238,10 @@ export const predefined = {
     code: -39013,
     message: 'Invalid block range',
   }),
+  RECEIVER_SIGNATURE_REQUIRED: new JsonRpcError({
+    code: -32000,
+    message: "Receiver's signature is required.",
+  }),
   FILTER_NOT_FOUND: new JsonRpcError({
     code: -32001,
     message: 'Filter not found',
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -240,7 +240,7 @@ export const predefined = {
   }),
   RECEIVER_SIGNATURE_ENABLED: new JsonRpcError({
     code: -32000,
-    message: "Receiver's signature is enabled.",
+    message: "Operation is not supported when receiver's signature is enabled.",
   }),
   FILTER_NOT_FOUND: new JsonRpcError({
     code: -32001,
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -238,9 +238,9 @@ export const predefined = {
     code: -39013,
     message: 'Invalid block range',
   }),
-  RECEIVER_SIGNATURE_REQUIRED: new JsonRpcError({
+  RECEIVER_SIGNATURE_ENABLED: new JsonRpcError({
     code: -32000,
-    message: "Receiver's signature is required.",
+    message: "Receiver's signature is enabled.",
   }),
   FILTER_NOT_FOUND: new JsonRpcError({
     code: -32001,
--- a/packages/relay/src/lib/precheck.ts
+++ b/packages/relay/src/lib/precheck.ts
@@ -388,7 +388,7 @@ export class Precheck {
       const verifyAccount = await this.mirrorNodeClient.getAccount(tx.to, requestDetails);
 
       // When `receiver_sig_required` is set to true, the receiver's account must sign all incoming transactions.
-      if (verifyAccount !== null && verifyAccount.receiver_sig_required === true) {
+      if (verifyAccount && verifyAccount.receiver_sig_required) {
         throw predefined.RECEIVER_SIGNATURE_ENABLED;
       }
     }
--- a/packages/relay/src/lib/precheck.ts
+++ b/packages/relay/src/lib/precheck.ts
@@ -389,7 +389,7 @@ export class Precheck {
 
       // When `receiver_sig_required` is set to true, the receiver's account must sign all incoming transactions.
       if (verifyAccount !== null && verifyAccount.receiver_sig_required === true) {
-        throw predefined.RECEIVER_SIGNATURE_REQUIRED;
+        throw predefined.RECEIVER_SIGNATURE_ENABLED;
       }
     }
   }
--- a/packages/relay/src/lib/precheck.ts
+++ b/packages/relay/src/lib/precheck.ts
@@ -20,7 +20,6 @@
 
 import { ethers, Transaction } from 'ethers';
 import { Logger } from 'pino';
-
 import { prepend0x } from '../formatters';
 import { MirrorNodeClient } from './clients';
 import constants from './constants';
@@ -85,6 +84,9 @@ export class Precheck {
     this.value(parsedTx);
     this.gasPrice(parsedTx, networkGasPriceInWeiBars, requestDetails);
     this.balance(parsedTx, mirrorAccountInfo, requestDetails);
+    if (parsedTx.to) {
+      await this.receiverAccount(parsedTx, requestDetails);
+    }
   }
 
   /**
@@ -374,6 +376,19 @@ export class Precheck {
         );
       }
       throw predefined.UNSUPPORTED_TRANSACTION_TYPE;
+    }
+  }
+
+  /**
+   * Checks if the receiver account exists.
+   * @param {Transaction} tx - The transaction.
+   * @param {RequestDetails} requestDetails - The request details for logging and tracking.
+   */
+  async receiverAccount(tx: Transaction, requestDetails: RequestDetails) {
+    const verifyAccount = await this.mirrorNodeClient.getAccount(tx.to!, requestDetails);
+
+    if (verifyAccount !== null && verifyAccount.receiver_sig_required === true) {
+      throw predefined.INTERNAL_ERROR("Receiver's signature is required.");
     }
   }
 }
--- a/packages/relay/src/lib/precheck.ts
+++ b/packages/relay/src/lib/precheck.ts
@@ -84,9 +84,7 @@ export class Precheck {
     this.value(parsedTx);
     this.gasPrice(parsedTx, networkGasPriceInWeiBars, requestDetails);
     this.balance(parsedTx, mirrorAccountInfo, requestDetails);
-    if (parsedTx.to) {
-      await this.receiverAccount(parsedTx, requestDetails);
-    }
+    await this.receiverAccount(parsedTx, requestDetails);
   }
 
   /**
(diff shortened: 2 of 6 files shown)


packages/relay/src/lib/precheck.ts
```diff
@@ -85,6 +86,9 @@ export class Precheck {
     this.value(parsedTx);
     this.gasPrice(parsedTx, networkGasPriceInWeiBars, requestDetails);
     this.balance(parsedTx, mirrorAccountInfo, requestDetails);
+    if (parsedTx.to) {
+      await this.receiverAccount(parsedTx, requestDetails);
+    }
```
quiet-node, line 91: Let's move the parsedTx.to check inside the receiverAccount() precheck instead.

packages/relay/src/lib/precheck.ts
```diff
@@ -376,4 +380,17 @@ export class Precheck {
...
     }
   }
+
+  /**
+   * Checks if the receiver account exists.
+   * @param {Transaction} tx - The transaction.
+   * @param {RequestDetails} requestDetails - The request details for logging and tracking.
+   */
+  async receiverAccount(tx: Transaction, requestDetails: RequestDetails) {
+    const verifyAccount = await this.mirrorNodeClient.getAccount(tx.to!, requestDetails);
```
quiet-node, line 390: Let's add the parsedTx.to check here. We can make it only run this precheck of parsedTx.to is found so in line 390 we don't have to use the `!` operator.

packages/relay/src/lib/precheck.ts
```diff
@@ -376,4 +380,17 @@ export class Precheck {
...
+
+    if (verifyAccount !== null && verifyAccount.receiver_sig_required === true) {
+      throw predefined.INTERNAL_ERROR("Receiver's signature is required.");
```
quiet-node, line 393: Let's throw a `RESOURCE_NOT_FOUND` predefined error instead of `INTERNAL_ERROR`. Also let's include the value of tx.to.

```suggestion
      throw predefined.RESOURCE_NOT_FOUND(`Receiver's signature is required: tx.to: ${tx.to}`);
```

packages/relay/src/lib/precheck.ts
```diff
@@ -376,4 +380,17 @@ export class Precheck {
...
+
+  /**
+   * Checks if the receiver account exists.
+   * @param {Transaction} tx - The transaction.
+   * @param {RequestDetails} requestDetails - The request details for logging and tracking.
+   */
+  async receiverAccount(tx: Transaction, requestDetails: RequestDetails) {
+    const verifyAccount = await this.mirrorNodeClient.getAccount(tx.to!, requestDetails);
+
+    if (verifyAccount !== null && verifyAccount.receiver_sig_required === true) {
```
quiet-node, line 392: Could you kindly did some extra research and add a comment here stating why we want to make sure `receiver_sig_required` attribute is turned off?

packages/relay/src/lib/precheck.ts
```diff
@@ -376,4 +380,17 @@ export class Precheck {
       throw predefined.UNSUPPORTED_TRANSACTION_TYPE;
     }
   }
+
+  /**
+   * Checks if the receiver account exists.
+   * @param {Transaction} tx - The transaction.
+   * @param {RequestDetails} requestDetails - The request details for logging and tracking.
+   */
+  async receiverAccount(tx: Transaction, requestDetails: RequestDetails) {
```
quiet-node, line 389: Please add test coverage for this method in the precheck.spec.ts

quiet-node (changes requested):
Great work! Some blocking requests

packages/relay/src/lib/precheck.ts
```diff
@@ -376,4 +380,17 @@ export class Precheck {
...
+    const verifyAccount = await this.mirrorNodeClient.getAccount(tx.to!, requestDetails);
+
+    if (verifyAccount !== null && verifyAccount.receiver_sig_required === true) {
+      throw predefined.INTERNAL_ERROR("Receiver's signature is required.");
+    }
```
quiet-node, line 394: Let's add another check to see if verifyAccount is null then throw a `RESOURCE_NOT_FOUND` and include the value of tx.to.

```
      throw predefined.RESOURCE_NOT_FOUND(`Cannot find receiver account: tx.to=${tx.to}`);
```

packages/relay/src/lib/precheck.ts
Nana-EC: yeah, there may be an existing error that could be used but if not add a new one

packages/relay/src/lib/precheck.ts
Nana-EC: Careful on this one since there are valid cases for `tx.to` not being set e.g. contract create.

quiet-node: Ahh yeah you're right tx.to is nullable if the transaction is a contract create transaction. Then maybe in the test make the tx.to a non-null address but mock not found. @nadezhdapopovaa please let me know if this is clear enough

quiet-node: Oh wait this block only executes when `tx.to` is not null; therefore, it is still possible for the mirror node to return a null result if `tx.to` is not found on the network. I'd suggest instead of checking `if (verifyAccount !== null && verifyAccount.receiver_sig_required === true)` for the sig required, we can pull the verifyAccount !== null out and add a new block to check the validity of the verifyAccount. If it's null then throw RESOURCE_NOT_FOUND like how we did with the tx.from.

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -1586,6 +1593,46 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+          const signedTx = await accounts[0].wallet.signTransaction(tx);
+          await new Promise((r) => setTimeout(r, 3000));
+
+          const error = predefined.RECEIVER_SIGNATURE_REQUIRED;
+
+          await Assertions.assertPredefinedRpcError(error, sendRawTransaction, false, relay, [
+            signedTx,
+            requestDetails,
+          ]);
+        });
```
quiet-node, line 1635: Great test! Let's add another test to ensure it works fine when `receiver.to` has `receiver_sig_required` disabled.

Additionally, if we decide to throw a `RESOURCE_NOT_FOUND` error for `tx.to`, we should include another test that uses an address not present in the system to verify that it blocks the request.

quiet-node (changes requested):
Great work! However, additional tests are still needed. Also a batch of unit tests is required in the `precheck.spec.ts` file to functionally test the method.

packages/relay/src/lib/precheck.ts
quiet-node: Please disregard the above! I completely overlooked the fact that `tx.to` can be a hollow address (with no associated account). We can send HBAR to that hollow account, and the system can lazily create the account as needed. Please proceed with your original flow. I apologize for any inconvenience this may have caused.

packages/server/tests/acceptance/rpc_batch1.spec.ts
nadezhdapopovaa: talked to Logan and we decided that `RESOURCE_NOT_FOUND` change won't be necessary, which means the test itself won't be necessary

packages/relay/src/lib/precheck.ts
```diff
@@ -376,4 +377,20 @@ export class Precheck {
...
+   * Checks if the receiver account exists and has receiver_sig_required set to true.
+   * @param {Transaction} tx - The transaction.
+   * @param {RequestDetails} requestDetails - The request details for logging and tracking.
+   */
+  async receiverAccount(tx: Transaction, requestDetails: RequestDetails) {
+    if (tx.to) {
+      const verifyAccount = await this.mirrorNodeClient.getAccount(tx.to, requestDetails);
+
+      // When `receiver_sig_required` is set to true, the receiver's account must sign all incoming transactions.
+      if (verifyAccount !== null && verifyAccount.receiver_sig_required === true) {
```
natanasow, line 391: Can be simplified to :point_down:  IMO.
```
if (verifyAccount && verifyAccount.receiver_sig_required)
```

packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
```diff
@@ -127,6 +128,13 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
         balance: Hbar.from(100_000_000_000, HbarUnit.Hbar).to(HbarUnit.Tinybar),
       },
     };
+    const RECEIVER_ACCOUNT_RES = {
+      account: ACCOUNT_ADDRESS_1,
+      balance: {
+        balance: Hbar.from(100_000_000_000, HbarUnit.Hbar).to(HbarUnit.Tinybar),
```
natanasow, line 134: Just to be on the same page. The mirror node response for balance is denominated in tinybars (not weibars). 

Here we mock `100_000_000_000` hbars which is `100_000_000_000 * 100_000_000 * 10_000_000_000` weibars in the ethers world.

E.g. for that test I would mock 1 hbar
```typescript
balance: Hbar.fromTinybars(100_000_000).to(HbarUnit.Tinybar)
// or
balance: Hbar.from(1, HbarUnit.Hbar).to(HbarUnit.Tinybar)
```

natanasow (dismissed):

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -238,6 +238,10 @@ export const predefined = {
     code: -39013,
     message: 'Invalid block range',
   }),
+  RECEIVER_SIGNATURE_ENABLED: new JsonRpcError({
+    code: -32000,
+    message: "Receiver's signature is enabled.",
+  }),
```
victor-yanev, line 244:
```suggestion
  RECEIVER_SIGNATURE_REQUIRED: new JsonRpcError({
    code: -32000,
    message: "Receiver's signature is required.",
  }),
```

victor-yanev (approved):
LGTM, one small nit

packages/relay/src/lib/errors/JsonRpcError.ts
acuarica: Suggestion from the team `Operation not supported when receiver's signature is enabled.`

konstantinabl (approved):

quiet-node (approved):

konstantinabl (approved):

quiet-node (approved):
LGTM! Thanks for the great work 🚀 💪🏼 🙌🏼

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -1586,6 +1593,90 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+          if (!receipt.accountId) {
+            throw new Error('Failed to create new account - accountId is null');
+          }
+
+          const toAddress = Utils.idToEvmAddress(receipt.accountId.toString());
+          const verifyAccount = await mirrorNode.get(`/accounts/${toAddress}`, requestId);
+
+          if (verifyAccount && !verifyAccount.account) {
+            verifyAccount == (await mirrorNode.get(`/accounts/${toAddress}`, requestId));
+          }
```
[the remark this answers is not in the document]
nadezhdapopovaa: In some cases, we attempt to fetch the data before it is saved to the mirror node. The refetch ensures that there’s enough time for the data to be saved before proceeding with the test. The if statement serves as a safeguard to confirm that an account has been created—there shouldn’t be an account without an ID.

PR #3331, opened 2024-12-12
refactor: clean up configurations

### LINKED ISSUE #3099: [HBAR Rate Limit Redesign] Add OPERATOR tier
Instead of tracking the remaining total budget in-memory inside `HbarLimitService` (we would have to divide it by the number of Kubernetes instances, etc - it is creating unnecessary complexity in the configurations), we could create a new subscription tier (`SubscriptionTier.OPERATOR`) and add a pre-configured spending plan in the JSON configuration file which links the ETH address(es) of the relay operator account(s) to it.

In this way we can only set the total limit for the `OPERATOR` tier to `HBAR_RATE_LIMIT_TINYBAR` and it will be shared by all instances of the relay.

victor-yanev:
**Description**:

This PR cleans up the messed up configurations that we have for the operator account, and fixes the underlying root cause of this spaghetti code coming from the GitHub workflow files.

It also stabilizes some flaky unit tests + sets the Pino logger instances to `silent` mode in the unit tests to avoid cluttering the logs in the CI.

**Related issue(s)**:

Fixes #3099 

**Notes for reviewer**:

**Checklist**


.github/workflows/acceptance-workflow.yml
```diff
@@ -31,27 +31,23 @@ on:
...
   acceptance-workflow:
     runs-on: smart-contracts-linux-large
     timeout-minutes: 50
     permissions:
       contents: read
       checks: write
-    env:
-      OPERATOR_ID_MAIN: ${{ inputs.operator_id }}
     steps:
+      - name: Set env variables
```
mishomihov00, line 46: Is there a reason to remove the `env` statement and use this `sh` snippet?

victor-yanev: Yes, it's an optional input variable which is wrongly set always even when its not provided, causing further difficulties in the configurations, as seen in the deleted code that was trying to "workaround" this issue from the workflow

victor-yanev: Whereas, doing:
```
    env:
       OPERATOR_ID_MAIN: ${{ inputs.operator_id }}
```

would override the one from the `.env` file even if the input is empty, which is causing the difficulties in the code we are trying to fix.

packages/relay/src/lib/services/hbarLimitService/index.ts
```diff
@@ -99,12 +99,11 @@ export class HbarLimitService implements IHbarLimitService {
...
-    const operatorKey = ConfigService.get('OPERATOR_KEY_MAIN');
-    if (operatorId) {
-      this.setOperatorAddress(AccountId.fromString(operatorId as string).toSolidityAddress());
-    } else if (operatorKey) {
-      this.setOperatorAddress(Utils.createPrivateKeyBasedOnFormat(operatorKey as string).publicKey.toEvmAddress());
+    const operator = Utils.getOperator(logger);
+    if (operator) {
+      this.operatorAddress = prepend0x(AccountId.fromString(operator.accountId.toString()).toSolidityAddress());
+    } else {
+      this.operatorAddress = zeroAddress();
```
quiet-node, line 106: QQ: Why do we want to set it to zero here instead of a more critical option? I understand this is a sub-flow which shouldn't be stopping the requests but could this be misleading if we aim to track the operator's balance through the cache or Prometheus metrics?

quiet-node (approved):
Definitely feels and looks better with the new changes! Great work!

mishomihov00 (approved):
Review applies to:
`.github/workflows/acceptance-workflow.yml`
`charts/hedera-json-rpc-relay/environments/minikube.yaml`
`charts/hedera-json-rpc-relay/templates/secret.yaml`

quiet-node (dismissed):
LGMT! Thanks for the great work!

quiet-node:
@victor-yanev Hey Victor, hope you’re doing well! Could you please rebase this branch and resolve the unsigned commits? It would be greatly appreciated if you could!

rbarkerSL (dismissed):
Approval applies to `.github/workflows/acceptance-workflow.yml`

quiet-node (approved):

.github/workflows/acceptance-workflow.yml
```diff
@@ -38,20 +38,22 @@ on:
...
+        run: |
+          if [ -n "${{ inputs.operator_id }}" ]; then
+            echo "OPERATOR_ID_MAIN=${{ inputs.operator_id }}" >> $GITHUB_ENV
```
andrewb1269hg, line 52: Can we switch this away from using `$GITHUB_ENV` and replace it with `$GITHUB_OUTPUT` instead?

To use it in a later step, you'll need to add `id:` field to this step, then access the output like so:
`${{ steps.<id name>.outputs.OUTPUT_ID_MAIN }}`

andrewb1269hg (changes requested):

andrewb1269hg (approved):
Review and approve `.github/workflows/acceptance-workflow.yml`

quiet-node:
> Review and approve `.github/workflows/acceptance-workflow.yml`

Thanks @andrewb1269hg. Just wanted to make sure, this line https://github.com/hashgraph/hedera-json-rpc-relay/pull/3331/files#diff-892cb201169a51e1b74870905dd4aba3e187217c08f44cea9bfdf568bafa2847R55 wouldn't be a security risk, right?

andrewb1269hg (changes requested):
`.github/workflows/acceptance-workflow.yml` line 55:
`echo "OPERATOR_KEY_MAIN=${{ secrets.operator_key }}" >> $GITHUB_ENV`

I *think* this will echo out the key and may be a security risk. Blocking this merge so we can get someone to look at this.

andrewb1269hg:
> > Review and approve `.github/workflows/acceptance-workflow.yml`
> 
> Thanks @andrewb1269hg. Just wanted to make sure, this line https://github.com/hashgraph/hedera-json-rpc-relay/pull/3331/files#diff-892cb201169a51e1b74870905dd4aba3e187217c08f44cea9bfdf568bafa2847R55 wouldn't be a security risk, right?

Thanks @quiet-node . I'm going to get someone else to confirm this isn't an issue before I unblock this PR.

.github/workflows/acceptance-workflow.yml
```diff
@@ -38,20 +38,22 @@ on:
...
+          fi
+          if [ -n "${{ secrets.operator_key }}" ]; then
+            echo "OPERATOR_KEY_MAIN=${{ secrets.operator_key }}" >> $GITHUB_ENV
```
andrewb1269hg, line 55: This needs to be changed.

Use the effective per step `env:` blocks as needed. The above exposes the key to all steps and all third-party steps

andrewb1269hg (changes requested):
There is an issue with the secrets being passed to `$GITHUB_ENV` that needs to be resolved.

.github/workflows/acceptance-workflow.yml
```diff
@@ -99,7 +96,12 @@ jobs:
         with:
           max_attempts: ${{ inputs.maxAttempts }}
           timeout_minutes: 30
-          command: npm run acceptancetest:${{ inputs.testfilter }}
+          command: |
+            if [ -n "${{ secrets.operator_key }}" ] && [ "${{ secrets.operator_key }}" != "" ]; then
+              echo "Overriding OPERATOR_KEY_MAIN with the secret value"
+              export OPERATOR_KEY_MAIN="${{ secrets.operator_key }}"
```
rbarkerSL, line 102: This environment variable should be set in the `env` section on the action. `${{ secrets.* }}` values should not be used directly like this in a bash command.

rbarkerSL (changes requested):

.github/workflows/acceptance-workflow.yml
quiet-node: Ah, I see. But this brings us back to the dilemma discussed here: https://github.com/hashgraph/hedera-json-rpc-relay/pull/3331/files#r1890465396. Setting it as a per-step environment variable would override the value from the .env file that’s used by Relay.

I tried something like this:
```
env:
  OPERATOR_KEY_MAIN: ${{ secrets.operator_key && secrets.operator_key != '' && secrets.operator_key || env.OPERATOR_KEY_MAIN }}
```
The OPERATOR_KEY_MAIN still gets overridden even when secrets.operator_key is empty. Do you have any suggestions on how to handle this?

andrewb1269hg, line 102: I don't think this resolves the problem. Please use the `env:` stanza to fix instead.

andrewb1269hg (changes requested):

.github/workflows/acceptance-workflow.yml
quiet-node: Thanks @andrewb1269hg . I stated the dilemma here https://github.com/hashgraph/hedera-json-rpc-relay/pull/3331#discussion_r1925907565

rbarkerSL (approved):

quiet-node (approved):

andrewb1269hg (approved):

PR #3349, opened 2024-12-20
fix: enhanced reliability of eth RPC methods with null checks and retry mechanisms

### LINKED ISSUE #3345: [Failed-Request-Alert-Tuning] `eth_getBlockByHash` && `eth_getBlockByNumber` method returns too many 500 errors
### Problem  
- Approximately 60,000 log entries on `mainnet` display the error: `Received an invalid integer type: null`, without providing additional diagnostic details. This issue is a major contributor to 500 errors for both `eth_getBlockByHash` and `eth_getBlockByNumber`. The root cause stems from the shared `EthImpl.getBlock()` method used by both endpoints. The closest `trace` log indicates, `Synthetic transaction hashes will be populated in the block response`, suggesting that the issue arises after the `populateSyntheticTransactions()` process. It is likely rooted in the `ReceiptsRootUtils.buildReceiptRootHashes()` method or during the instantiation of a new `Block()`.

- Another issue identified in the `mainnet` logs is the error: `Cannot read properties of null (reading 'toString')`, which occurs on `eth_getBlock` endpoints. This error is logged immediately after the Mirror Node (MN) call is submitted and is likely originating within the `EthImpl.getBlock()` method. Example Request ID: `2504dbdb-e3a4-4d6c-8f2f-af8ba8c7d05b`.

### Solution  
Investigate and fix the underlying issue in the `EthImpl.getBlock()` function that causes the Internal Error, ensuring proper handling and accurate processing of block data.

### LINKED ISSUE #3351: [Failed-Request-Alert-Tuning] eth_getTransactionReceipt method returns too many 500 errors
### Problem  
On the mainnet, an excessive number of log entries indicate 500 errors for the `eth_getTransactionReceipt` endpoint. One notable error message is:  
- `Error invoking RPC: Invalid parameter: hashOrNumber` – This appears to stem from an empty `blockHash` in the synthetics log. 
- `Cannot read properties of null (reading 'toString')` - This could potentially fixed by https://github.com/hashgraph/hedera-json-rpc-relay/pull/3349

**Updated**: The `Error invoking RPC: Invalid parameter: hashOrNumber` error was also triggered in scenarios where `block_hash` was set as an empty hex string ("0x"), as detailed in https://github.com/hashgraph/hedera-json-rpc-relay/pull/3349. Consequently, this issue is expected to be resolved by the changes implemented in https://github.com/hashgraph/hedera-json-rpc-relay/pull/3349.

### Solution  
Investigate and resolve the root cause of the 500 errors in the `eth_getTransactionReceipt` endpoint. Ensure robust error handling and accurate processing of block data to prevent invalid parameters and enhance system reliability.

quiet-node:
**Description**:
Numerous requests for `eth_getBlockByNumber` and `eth_getBlockByHash` have been reported as failing with errors such as `Cannot read properties of null` and `Received an invalid integer type: null`. 

Additionally, `eth_getTransactionReceipt` is also reported to be failing with `Error invoking RPC: Invalid parameter: hashOrNumber`.

These issues stem from regressions in the `getBlock()` method, which lacks adequate null checks in certain areas. This pull request addresses the problem by implementing proper null checks and enhancing fallback mechanisms and error handling.

Partially, the results of contracts and log responses from the Mirror Node occasionally contain undefined fields, such as transaction_index, block_hash, block_number, and log_index. This issue may arise because the Mirror Node database does not have sufficient time to save all the information before it is fetched. This PR adds a retry mechanism to the fetching methods that checks for these cases, waits for a specified duration, and then retries the fetch operation.

**Related issue(s)**:

Fixes #3345
Fixes #3351

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -2,7 +2,7 @@
  *
  * Hedera JSON RPC Relay
  *
- * Copyright (C) 2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -757,21 +757,40 @@ export class MirrorNodeClient {
    * In some very rare cases the /contracts/results api is called before all the data is saved in
    * the mirror node DB and `transaction_index` or `block_number` is returned as `undefined` or `block_hash` as `0x`.
    * A single re-fetch is sufficient to resolve this problem.
-   * @param {string} transactionIdOrHash - The transaction ID or hash
-   * @param {RequestDetails} requestDetails - The request details for logging and tracking.
+   *
+   * @param {string} methodName - The name of the method used to fetch contract results.
+   * @param {any[]} args - The arguments to be passed to the specified method for fetching contract results.
+   * @param {RequestDetails} requestDetails - Details used for logging and tracking the request.
+   * @returns {Promise<any>} - A promise resolving to the fetched contract result, either on the first attempt or after a retry.
    */
-  public async getContractResultWithRetry(transactionIdOrHash: string, requestDetails: RequestDetails) {
-    const contractResult = await this.getContractResult(transactionIdOrHash, requestDetails);
-    if (
-      contractResult &&
-      !(
-        contractResult.transaction_index &&
-        contractResult.block_number &&
-        contractResult.block_hash != EthImpl.emptyHex
-      )
-    ) {
-      return this.getContractResult(transactionIdOrHash, requestDetails);
+  public async getContractResultWithRetry(
+    methodName: string,
+    args: any[],
+    requestDetails: RequestDetails,
+  ): Promise<any> {
+    const shortDelay = 500;
+    const contractResult = await this[methodName](...args);
+    const contractObjects = Array.isArray(contractResult) ? contractResult : [contractResult];
+
+    for (const contractObject of contractObjects) {
+      if (
+        contractObject &&
+        (contractObject.transaction_index == null ||
+          contractObject.block_number == null ||
+          contractObject.block_hash == EthImpl.emptyHex)
+      ) {
+        if (this.logger.isLevelEnabled('debug')) {
+          this.logger.debug(
+            `${requestDetails.formattedRequestId} Contract result contains undefined transaction_index, block_number, or block_hash set to 0x: transaction_hash:${contractObject.hash}, transaction_index:${contractObject.transaction_index}, block_number=${contractObject.block_number}, block_hash=${contractObject.block_hash}. Retrying after a delay of ${shortDelay} ms `,
+          );
+        }
+
+        // Backoff before repeating request
+        await new Promise((r) => setTimeout(r, shortDelay));
+        return await this[methodName](...args);
+      }
     }
+
     return contractResult;
   }
 
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -770,24 +770,26 @@ export class MirrorNodeClient {
   ): Promise<any> {
     const shortDelay = 500;
     const contractResult = await this[methodName](...args);
-    const contractObjects = Array.isArray(contractResult) ? contractResult : [contractResult];
 
-    for (const contractObject of contractObjects) {
-      if (
-        contractObject &&
-        (contractObject.transaction_index == null ||
-          contractObject.block_number == null ||
-          contractObject.block_hash == EthImpl.emptyHex)
-      ) {
-        if (this.logger.isLevelEnabled('debug')) {
-          this.logger.debug(
-            `${requestDetails.formattedRequestId} Contract result contains undefined transaction_index, block_number, or block_hash set to 0x: transaction_hash:${contractObject.hash}, transaction_index:${contractObject.transaction_index}, block_number=${contractObject.block_number}, block_hash=${contractObject.block_hash}. Retrying after a delay of ${shortDelay} ms `,
-          );
+    if (contractResult) {
+      const contractObjects = Array.isArray(contractResult) ? contractResult : [contractResult];
+      for (const contractObject of contractObjects) {
+        if (
+          contractObject &&
+          (contractObject.transaction_index == null ||
+            contractObject.block_number == null ||
+            contractObject.block_hash == EthImpl.emptyHex)
+        ) {
+          if (this.logger.isLevelEnabled('debug')) {
+            this.logger.debug(
+              `${requestDetails.formattedRequestId} Contract result contains undefined transaction_index, block_number, or block_hash is an empty hex (0x): transaction_hash:${contractObject.hash}, transaction_index:${contractObject.transaction_index}, block_number=${contractObject.block_number}, block_hash=${contractObject.block_hash}. Retrying after a delay of ${shortDelay} ms `,
+            );
+          }
+
+          // Backoff before repeating request
+          await new Promise((r) => setTimeout(r, shortDelay));
+          return await this[methodName](...args);
         }
-
-        // Backoff before repeating request
-        await new Promise((r) => setTimeout(r, shortDelay));
-        return await this[methodName](...args);
       }
     }
 
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -1,8 +1,8 @@
-/* -
+/*-
  *
  * Hedera JSON RPC Relay
  *
- * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -620,7 +620,10 @@ export class MirrorNodeClient {
       requestDetails,
     );
 
-    await this.cacheService.set(cachedLabel, block, MirrorNodeClient.GET_BLOCK_ENDPOINT, requestDetails);
+    if (block) {
+      await this.cacheService.set(cachedLabel, block, MirrorNodeClient.GET_BLOCK_ENDPOINT, requestDetails);
+    }
+
     return block;
   }
 
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -894,14 +894,25 @@ export class MirrorNodeClient {
     return this.getQueryParams(queryParamObject);
   }
 
-  public async getContractResultsLogs(
+  /**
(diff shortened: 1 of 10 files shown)


Nana-EC (commented):
Good catch.
Some considerations

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -620,7 +620,10 @@ export class MirrorNodeClient {
       requestDetails,
     );
 
-    await this.cacheService.set(cachedLabel, block, MirrorNodeClient.GET_BLOCK_ENDPOINT, requestDetails);
+    if (block) {
```
acuarica, line 623: not sure if I understand correctly, but the errors described in the issue were because of caching null blocks?

packages/relay/src/receiptsRootUtils.ts
```diff
@@ -93,16 +94,25 @@ export class ReceiptsRootUtils {
   public static buildReceiptRootHashes(txHashes: string[], contractResults: any[], logs: Log[]): IReceiptRootHash[] {
     const receipts: IReceiptRootHash[] = [];
 
-    for (let i in txHashes) {
+    for (const i in txHashes) {
       const txHash: string = txHashes[i];
       const logsPerTx: Log[] = logs.filter((log) => log.transactionHash == txHash);
       const crPerTx: any[] = contractResults.filter((cr) => cr.hash == txHash);
+
+      let transactionIndex: any = null;
```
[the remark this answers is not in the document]
acuarica: +1, maybe we could describe why the fix solves the issue.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -754,20 +757,39 @@ export class MirrorNodeClient {
...
+    for (const contractObject of contractObjects) {
+      if (
+        contractObject &&
+        (contractObject.transaction_index == null ||
+          contractObject.block_number == null ||
+          contractObject.block_hash == EthImpl.emptyHex)
+      ) {
+        if (this.logger.isLevelEnabled('debug')) {
+          this.logger.debug(
+            `${requestDetails.formattedRequestId} Contract result contains undefined transaction_index, block_number, or block_hash set to 0x. Retrying after a delay of ${shortDelay} ms.`,
```
Nana-EC, line 785: Might as well log the 3 fields as we won't know which field is null in this case

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -886,6 +919,31 @@ export class MirrorNodeClient {
       1,
       MirrorNodeClient.mirrorNodeContractResultsLogsPageMax,
     );
+
+    for (const log of logResults) {
+      if (log && (log.transaction_index == null || log.block_number == null || log.index == null)) {
+        console.log(logResults);
+        if (this.logger.isLevelEnabled('debug')) {
+          this.logger.debug(
+            `${requestDetails.formattedRequestId} Contract result log contains undefined transaction_index, block_number, or index. Retrying after a delay of ${shortDelay} ms.`,
```
Nana-EC, line 928: From this log we won't know which field is undefined

packages/relay/src/lib/eth.ts
```diff
@@ -1,8 +1,8 @@
-/* -
+/*-
  *
  * Hedera JSON RPC Relay
  *
- * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
```
Nana-EC, line 5: nit: why change. Should revert the license changes

packages/relay/src/lib/eth.ts
```diff
@@ -2531,11 +2544,12 @@ export class EthImpl implements Eth {
     if (blockResponse == null) return null;
     const timestampRange = blockResponse.timestamp;
     const timestampRangeParams = [`gte:${timestampRange.from}`, `lte:${timestampRange.to}`];
-    const contractResults = await this.mirrorNodeClient.getContractResults(
+    const contractResults = await this.mirrorNodeClient.getContractResultWithRetry(
```
Nana-EC, line 2547: Are we just being cautious or does this case actually need to have retry support?

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1,8 +1,8 @@
-/* -
+/*-
  *
  * Hedera JSON RPC Relay
  *
- * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
```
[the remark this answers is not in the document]
quiet-node: Ah I believe it's an auto change by the eslint udpate got pushed in recently. Will revert and create a ticket to report this.

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
```diff
@@ -334,7 +335,7 @@ export class CommonService implements ICommonService {
         new Log({
           address: log.address,
           blockHash: toHash32(log.block_hash),
-          blockNumber: numberTo0x(log.block_number),
+          blockNumber: nullableNumberTo0x(log.block_number),
```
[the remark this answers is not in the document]
quiet-node: Yes, this is one of the strange cases where `block_number`, `transaction_index`, or `blockHash` are falsy, leading to 500 errors. The retry mechanism should address the issue. Should I revert this change and throw a more informative error message even if those fields remain falsy after the retry? Hmm maybe that would be better so the clients can resubmit the requests themselves other than having null in the response. Wdyt?

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: This is an issue I noticed and an attempt securing the code. While it may or may not be directly related to the problem, we should ensure the block is not null before storing it in the cache. Otherwise, we risk repeatedly retrieving null from the cache, which likely has a TTL.

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: You're right that would give more info. Added those fields and an extra field for transaction hash for easier debug.

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: You're right that would give more info. Added those fields and an extra field for transaction hash for easier debug.

packages/relay/src/lib/eth.ts
quiet-node: Gotcha reverted it!

packages/relay/src/lib/eth.ts
quiet-node: The `contractResults` here was a contributing factor to the 500 errors. In the rare cases mentioned, `contractResults` contains null fields. When passed to `buildReceiptRootHashes`, the function does not handle these null values correctly, resulting in 500 errors. Therefore, this scenario requires a retry if such a rare case occurs.

quiet-node: Since these cases occur only on the mainnet and at the time of the transaction, I plan to release these updates that include additional debugging information for improved analysis. If the issues persist, it will be easier to pinpoint the root causes and escalate them to the mainnet team more effectively.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -724,6 +724,11 @@ export class SDKClient {
       );
       return transactionResponse;
     } catch (e: any) {
+      this.logger.warn(
+        e,
+        `${requestDetails.formattedRequestId} Transaction failed while executing transaction via the SDK: transactionId=${transaction.transactionId}, callerName=${callerName}, txConstructorName=${txConstructorName}`,
+      );
+
```
quiet-node, line 731: **Note for Rereviewers**: 
This is an effort to enhance debugging.

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -1672,6 +1672,8 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
 
           const signedTx = await accounts[0].wallet.signTransaction(tx);
           const transactionHash = await relay.sendRawTransaction(signedTx, requestId);
+          await relay.pollForValidTransactionReceipt(transactionHash);
+
```
quiet-node, line 1676: **Note for Rereviewers**: 
This is an effort to stablize the flaky test.

packages/relay/src/receiptsRootUtils.ts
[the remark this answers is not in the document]
quiet-node: pushed in a comment to explain the logic. Thanks for pointing it out!

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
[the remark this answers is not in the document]
Nana-EC: Hmmm, if a retry is ensured to resolve the issue then maybe we can manage one iteration in the short term.
We should open up an issue in the mirror node repo to investigate this issue so this can be resolved.

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
```diff
@@ -175,6 +177,20 @@ export class CommonService implements ICommonService {
     returnLatest?: boolean,
   ): Promise<any> {
     if (!returnLatest && this.blockTagIsLatestOrPending(blockNumberOrTagOrHash)) {
+      if (this.logger.isLevelEnabled('debug')) {
```
acuarica, line 180: nit: why is it needed to query for the log level? can we just interpolate the log message? https://github.com/pinojs/pino/blob/main/docs/api.md#interpolationvalues-any

This would be just to simplify the log logic.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -754,21 +757,42 @@ export class MirrorNodeClient {
...
-      )
-    ) {
-      return this.getContractResult(transactionIdOrHash, requestDetails);
+  public async getContractResultWithRetry(
+    methodName: string,
+    args: any[],
+    requestDetails: RequestDetails,
+  ): Promise<any> {
+    const shortDelay = 500;
+    const contractResult = await this[methodName](...args);
```
acuarica, line 772: What was this issue with calling `getContractResult` directly? I see all calls to `getContractResultWithRetry` use the same `methodName=this.mirrorNodeClient.getContractResults.name`, is there any other benefit of calling it indirectly?

quiet-node: Ah, yes, there’s a rare case where the contract result contains null fields, as mentioned here: https://github.com/hashgraph/hedera-json-rpc-relay/pull/3349#discussion_r1894698244. Calling getContractResultWithRetry() will invoke getContractResult, and if this rare case occurs, it will attempt the call again. The plan is to release this change first to gather more information about the getContractResult issue and eventually open a ticket on the MN side for further investigation.

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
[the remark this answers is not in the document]
natanasow: We should not return a nullable (or casted to 0x) block number IMO. This might seriously mess up third-party tools like the graph.

natanasow (commented):
LG, one question/thought that might be resolved before merging.

natanasow (approved):

Nana-EC (approved):

PR #3350, opened 2024-12-21
Fix: Resolve Type Warnings for ConfigService.get()

### LINKED ISSUE #3142: Fix all typescript warnings and errors generated by type mismatching on the `ConfigService.get(<envName>)` method in test specs
##  Description
There are several places in current tests where expected values are in different types than those defined in the `ConfigService.get(<envName>)` method.

### Solution:
Identify all places where there are warnings or errors due to type mismatches. Remove any unnecessary casting, or add casting if required, to align with the types defined in `ConfigService.get(<envName>)`.

### Example Implementation
Refactor test cases to ensure type consistency with `ConfigService.get(<envName>)`, addressing any type warnings or errors.

belloibrahv:
**Description**:
TL;DR; Enhancing Type Inference, Config Consistency, and Establishing a Single Source of Truth in ConfigService  

This PR significantly improves type inference in `ConfigService.get()`, ensuring more reliable and explicit type resolution across configurations.  

It also updates the entire codebase to adhere to a single source of truth for configuration management, enhancing consistency, reliability, and maintainability. 

Key improvements include:  
- Strengthened type inference in `ConfigService.get()`, introducing explicit TypeScript types and interfaces that precisely map config keys to their expected types, reducing manual casting and type mismatches.  
- Unified configuration values across the codebase, ensuring all components rely on a single, authoritative source of truth.  
- Updated default values for all global configs to align with `configurations.md`, enhancing correctness and consistency.  
- Replaced hardcoded constants with corresponding config values, reducing redundancy and improving maintainability.  
- Alphabetized configuration tables in `configurations.md`, ensuring proper documentation and organization.  

These enhancements improve type safety, reliability, and maintainability, making the configuration system more robust and easier to manage across the entire codebase.

---

**Related issue(s)**:  
Fixes #3142

---

**Checklist**:  
- [x] **Documented**: Added/updated relevant code comments and documentation.  
- [x] **Tested**: Verified changes with updated and new unit tests to ensure correctness.


belloibrahv:
@Nana-EC @ebadiere @quiet-node I've submitted a PR addressing issue #3142 to resolve type warnings for ConfigService.get(). It ensures type consistency across test cases by refining type usage and adding/removing the necessary casting. Let me know if further changes are needed. Thank you for reviewing!

packages/relay/tests/lib/config/hbarSpendingPlanConfigService.spec.ts
```diff
@@ -1,8 +1,8 @@
-/*
+/*-
  *
  * Hedera JSON RPC Relay
  *
- * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
```
Nana-EC, line 5: nit: revert

packages/server/tests/acceptance/conformityTests.spec.ts
```diff
@@ -1,8 +1,8 @@
-/*
+/*-
  *
  * Hedera JSON RPC Relay
  *
- * Copyright (C) 2022-2024 Hedera Hashgraph, LLC
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
```
Nana-EC, line 5: revert this and any other license header changes where you're changing an old header date into this year.
Might be a bug in an automated script

packages/server/tests/acceptance/serverConfig.spec.ts
```diff
@@ -17,16 +17,18 @@
...
+
+import { ConfigName } from '../../../config-service/src/services/configName';
 import { Utils } from '../helpers/utils';
-import { ConfigService } from '@hashgraph/json-rpc-config-service/dist/services';
 
 describe('@server-config Server Configuration Options Coverage', function () {
   describe('Koa Server Timeout', () => {
     it('should timeout a request after the specified time', async () => {
-      const requestTimeoutMs: number = parseInt(ConfigService.get('SERVER_REQUEST_TIMEOUT_MS') || '3000');
+      const requestTimeoutMs: number = parseInt(ConfigService.get(ConfigName.SERVER_REQUEST_TIMEOUT_MS) as string || '3000');
```
Nana-EC, line 29: I wonder if we should add a getString and do the casting in the ConfigService to make this easier on devs.
Thinking out loud and getting thoughts.
No need to change in this PR

packages/config-service/src/services/globalConfig.ts
```diff
@@ -26,7 +28,7 @@ export interface ConfigProperty {
 }
 
 export class GlobalConfig {
-  public static readonly ENTRIES: Record<string, ConfigProperty> = {
+  public static readonly ENTRIES: Record<ConfigName, ConfigProperty> = {
```
acuarica, line 31: Instead of enumerating the keys manually, can we do something like
```typescript

const _CONFIG = {
  BATCH_REQUESTS_ENABLED: {
    envName: 'BATCH_REQUESTS_ENABLED',
    type: 'boolean',
    required: false,
    defaultValue: null,
  },
  //...
  WS_SUBSCRIPTION_LIMIT: {
    envName: 'WS_SUBSCRIPTION_LIMIT',
    type: 'number',
    required: false,
    defaultValue: null,
  },
} satisfies { [key: string]: ConfigProperty };

export type ConfigKey = keyof typeof _CONFIG;

export class GlobalConfig {
  public static readonly ENTRIES: Record<ConfigKey, ConfigProperty> = _CONFIG;
}

```
?

So we retain type safety for the keys (even if they're strings).

Nana-EC (commented):
Thanks for the contribution.

packages/server/tests/acceptance/serverConfig.spec.ts
belloibrahv: Thank you for the suggestion about `ConfigService.getString`. I agree that it could streamline type handling. I suggest we open a separate issue to implement and test this change comprehensively.

belloibrahv:
@Nana-EC @acuarica @quiet-node, Could you please review my updates and let me know if further adjustments are needed?

acuarica (changes requested):
Hi @belloibrahv, thanks for taking the time to send this, it's looking good. Left a couple of suggestions we can continue iterating on to improve the config.

acuarica:
Also @belloibrahv please take a look at https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350#discussion_r1894545838.

quiet-node:
@belloibrahv DCO is failing https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/checks?check_run_id=34857409075

belloibrahv:
@quiet-node @acuarica @Nana-EC @shemnon , thank you for the suggestion. I've implemented the following changes based on your feedback:

1. Updated `ConfigService.get` to accept `name: ConfigKey`, removing the need for `as ConfigKey` assertions.
2. Added a missing configuration entry for `SERVER_HOST` in `_CONFIG`.
3. Adjusted test cases to align with the updated method definition.

Let me know if there's anything else you'd like me to address!

belloibrahv:
> Also @belloibrahv please take a look at [#3350 (comment)](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350#discussion_r1894545838).

@acuarica, Thank you for the suggestion! I think adding a `getString` (or similar methods like `getNumber`, `getBoolean`) in `ConfigService` could significantly simplify casting and improve developer experience. It could also centralize type handling and reduce repetitive casting logic.  

Would you suggest we implement these methods for this task or address it as a follow-up? Also, should these methods throw errors for mismatched types, or return defaults like `null`? Happy to explore this further and incorporate it if needed!

Nana-EC:
> > Also @belloibrahv please take a look at [#3350 (comment)](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350#discussion_r1894545838).
> 
> 
> 
> @acuarica, Thank you for the suggestion! I think adding a `getString` (or similar methods like `getNumber`, `getBoolean`) in `ConfigService` could significantly simplify casting and improve developer experience. It could also centralize type handling and reduce repetitive casting logic.  
> 
> 
> 
> Would you suggest we implement these methods for this task or address it as a follow-up? Also, should these methods throw errors for mismatched types, or return defaults like `null`? Happy to explore this further and incorporate it if needed!  

Let's do it in a separate PR after this.

Also appreciate the patience.
The team is on holiday for the Christmas season so things are a bit slow.

belloibrahv:
Hi @quiet-node, @acuarica, @Nana-EC, @shemnon, Hope you all had a wonderful holiday! When you have a moment, please review and provide feedback on the requested changes to the PR. Thank you for your understanding!

packages/ws-server/tests/acceptance/subscribeNewHeads.spec.ts
```diff
@@ -32,6 +32,7 @@ import { ethers } from 'ethers';
 import WebSocket from 'ws';
 
 import { WsTestHelper } from '../helper';
+import { ConfigKey } from '../../../config-service/src/services/globalConfig';
```
acuarica, line 35: I see many changes due to only `import`s, I guess this is a leftover from a previous commit. If there is not actual change, we should revert these changes for easier review.

tools/brownie-example/LICENSE
```diff
@@ -1,6 +1,6 @@
 Hedera Brownie Example
 
-Copyright (C) 2024 Hedera Hashgraph, LLC
+Copyright (C) 2022-2024 Hedera Hashgraph, LLC
```
acuarica, line 3: Also please revert these license header changes. See Nana's comment for reference https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350#discussion_r1894545838.

acuarica (commented):
Hi @belloibrahv, hope you had a wonderful holiday too! It's looking good, left some comments.

We appreciate your patience on this. Next week the full team will be back from holidays, so you'll get the appropriate reviews.

acuarica:
Another change we could implement to improve type safety, could be to avoid casting the return type of `ConfigService.get`.

Probably better off in another PR, but I wanted to leave the general idea here for discussion.

We need some type definitions

```typescript
type TypeStrToType<Tstr extends string> = Tstr extends 'string'
  ? string
  : Tstr extends 'boolean'
    ? boolean
    : Tstr extends 'number'
      ? number
      : Tstr extends 'array'
        ? unknown[] // could be also `any[]`
        : never;

type GetTypeStrOfKey<K extends string> = K extends keyof typeof _CONFIG ? typeof _CONFIG[K]['type'] : never;
export type TypeOfKey<K extends string> = TypeStrToType<GetTypeStrOfKey<K>>;
```

These type definitions would allows to get right type based on the config key. For example

image


Then, we can update our `ConfigService.get` definition to something like

```typescript
  public static get<K extends ConfigKey>(name: K): TypeOfKey<K> | undefined
```

So now the correct return type is inferred based on the config key

image

Nana-EC:
> Hi @quiet-node, @acuarica, @Nana-EC, @shemnon, Hope you all had a wonderful holiday! When you have a moment, please review and provide feedback on the requested changes to the PR. Thank you for your understanding!

Thanks @belloibrahv , hope you had a good one too.
Changes look good and it seems we all agree there's some follow ups post this PR.
@acuarica and @quiet-node will be reviewing this to help you get this through.
Thanks again for the contributions

packages/config-service/src/services/index.ts
```diff
@@ -97,7 +98,7 @@ export class ConfigService {
    * @param name string
    * @returns string | undefined
    */
-  public static get(name: string): string | number | boolean | null | undefined {
+  public static get(name: ConfigKey): string | number | boolean | null | undefined {
     return this.getInstance().envs[name];
   }
```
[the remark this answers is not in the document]
quiet-node: Two Options:
1.	Add Specialized Getter Methods customized getter methods for specific types:
	•	ConfigService.getString(name: ConfigKey): string
	•	ConfigService.getNumber(name: ConfigKey): number
	•	ConfigService.getBoolean(name: ConfigKey): boolean
This approach improves type inference clarity, ensuring type-specific methods are used. However, it may add complexity, as developers would need to determine which method to call based on the expected return type.

2.	Typecasting Across the Codebase
Identify all instances of ConfigService.get() throughout the codebase and explicitly typecast the return values as needed. For example, consider the case in the provided screenshot:

```diff
const mirrorNodeInstance = new MirrorNodeClient(
-  ConfigService.get('MIRROR_NODE_URL') || '',
+  String(ConfigService.get('MIRROR_NODE_URL')) || '',
    logger.child({ name: `mirror-node` }),
    registry,
    cacheService,
);
```
This option is straightforward but requires meticulous updates across the codebase and may introduce some verbosity. However, I would recommend this approach as it adds minimal overhead.

quiet-node (changes requested):
Great work! Please revert the unnecessary changes and additions. I’ve added a suggestion on the ConfigService.get() method.

packages/config-service/src/services/index.ts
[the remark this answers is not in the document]
quiet-node: Luis adds a very good solution which can fix the issue, please refer to his solution here https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350#issuecomment-2569976576

packages/config-service/tests/src/services/configService.spec.ts
```diff
@@ -55,7 +56,7 @@ describe('ConfigService tests', async function () {
   });
 
   it('should return undefined for non-existing variable', async () => {
-    const res = ConfigService.get('NON_EXISTING_VAR');
+    const res = ConfigService.get('NON_EXISTING_VAR' as ConfigKey);
```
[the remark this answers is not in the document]
acuarica: Totally, I missed that, thanks Logan.

belloibrahv:
Hello @acuarica and @quiet-node, I’ve addressed the recent feedback, incorporating the suggested changes to enhance type safety and eliminate manual casting in `ConfigService.get()`. Please review the updates at your earliest convenience and let me know if there’s anything else to adjust. Thank you

quiet-node:
> Hello @acuarica and @quiet-node, I’ve addressed the recent feedback, incorporating the suggested changes to enhance type safety and eliminate manual casting in `ConfigService.get()`. Please review the updates at your earliest convenience and let me know if there’s anything else to adjust. Thank you

🚀 🚀 
Great work @belloibrahv! Thanks a lot for the effort! Just did a quick overview and seems like these are marked as resolved but the code didn't get reverted to original 
- https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/files/c7e6a08ad1cf2488806636155d7e01f6149b77f4#r1905864286. 
- https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/files/c7e6a08ad1cf2488806636155d7e01f6149b77f4#r1905844973
- https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/files/c7e6a08ad1cf2488806636155d7e01f6149b77f4#r1905854941
- and so on

belloibrahv:
> > Hello @acuarica and @quiet-node, I’ve addressed the recent feedback, incorporating the suggested changes to enhance type safety and eliminate manual casting in `ConfigService.get()`. Please review the updates at your earliest convenience and let me know if there’s anything else to adjust. Thank you
> 
> 🚀 🚀 Great work @belloibrahv! Thanks a lot for the effort! Just did a quick overview and seems like these are still not reverted as original
> 
>     * https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/files#r1905856733
> 
>     * https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/files#diff-aab9893229af0cbfd7691831e7a10de9ce96c3ee56cadd4aee34c880cd5c0535R5
> 
>     * https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/files#diff-a381493ecdd7f78305e86f7108029d14c33c6a67c503946507017df414dafa41R5
> 
> 
> And many more. Let’s try reverting them back to their original state! If you encounter any difficulties, let me know. I understand the auto-linting might have caused these changes, not you, so if you need assistance with reverting them, I’m definitely here to help.

@quiet-node Thank you for the feedback! I suspect the auto-linting might have caused these changes. Could you please guide me on the best way to revert the code to its original state without it being reverted again by the auto-linting? Any specific steps or tools to handle this would be greatly appreciated.

Thanks in advance for your help!

quiet-node:
> > > Hello @acuarica and @quiet-node, I’ve addressed the recent feedback, incorporating the suggested changes to enhance type safety and eliminate manual casting in `ConfigService.get()`. Please review the updates at your earliest convenience and let me know if there’s anything else to adjust. Thank you
> > 
> > 
> > 🚀 🚀 Great work @belloibrahv! Thanks a lot for the effort! Just did a quick overview and seems like these are still not reverted as original
> > ```
> > * https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/files#r1905856733
> > 
> > * https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/files#diff-aab9893229af0cbfd7691831e7a10de9ce96c3ee56cadd4aee34c880cd5c0535R5
> > 
> > * https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/files#diff-a381493ecdd7f78305e86f7108029d14c33c6a67c503946507017df414dafa41R5
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And many more. Let’s try reverting them back to their original state! If you encounter any difficulties, let me know. I understand the auto-linting might have caused these changes, not you, so if you need assistance with reverting them, I’m definitely here to help.
> 
> @quiet-node Thank you for the feedback! I suspect the auto-linting might have caused these changes. Could you please guide me on the best way to revert the code to its original state without it being reverted again by the auto-linting? Any specific steps or tools to handle this would be greatly appreciated.
> 
> Thanks in advance for your help!

Hi @belloibrahv,

I’ve reported this issue to the team, and we will try to address the auto-linting problem soon. For now, as a temporary workaround, I would have to have you follow these steps:
- 	Review all unnecessarily modified files in the PR and revert them to their original state (you can simply copy and paste the old values back).
- 	Before committing, navigate to the ./husky/pre-commit file and comment out the following line:
```
npx lint-staged --verbose
```
This will disable auto-linting during the commit process.

- 	When committing, ensure that the commented-out change to the ./husky/pre-commit file is not included in the staged files.
- 	Once your commit is complete, restore the ./husky/pre-commit file to its original state by uncommenting the npx lint-staged --verbose line.
- 	Avoid opening or modifying unrelated files during the commit process, as the auto-linting tool may still detect and alter them unnecessarily.

Please note that this is a workaround and should only be applied to files with undesired modifications in this PR. The team will implement a proper fix to eliminate the need for such workarounds. Let me know if you encounter any issues, and I’ll be happy to assist.

packages/config-service/src/services/globalConfig.ts
```diff
@@ -742,5 +763,44 @@ export class GlobalConfig {
       required: false,
       defaultValue: null,
     },
-  };
+} satisfies { [key: string]: ConfigProperty };
+
+export type ConfigKey = keyof typeof _CONFIG;
+
+export class ConfigService {
```
acuarica, line 770: No need for this extra class. The runtime conversion is already done by `ValidationService.typeCasting`.

packages/config-service/src/services/globalConfig.ts
```diff
@@ -742,5 +763,44 @@ export class GlobalConfig {
...
+  private static config: typeof _CONFIG = _CONFIG;
+
+  public static get<K extends ConfigKey>(name: K): TypeOfKey<K> | undefined {
```
acuarica, line 773: This signature should be placed in the other `ConfigService`.

quiet-node:
Hey @belloibrahv just reacing out to see how things are going. Let us know if you face any challenges or blockers! The team is here to help.

belloibrahv:
@acuarica, @quiet-node I have refactored the code and reverted the requested changes. Kindly review the updates at your convenience. Thank you!

quiet-node:
@belloibrahv nice thanks for the effort! However, all the CIs are broken now. It seems like the build of the Relay is failing. Didn't have a chance to go through your new updates yet but seems like they cause the build failure. Please take a look.

belloibrahv:
[quiet-node](https://github.com/quiet-node) Resolved CI issues. Kindly help review the changes. Thank you

belloibrahv:
@acuarica @quiet-node @Nana-EC I've addressed the failing CI checks in this PR. Please take another look at the updated code. I'm eager to get feedback and address any remaining concerns. Thank You.

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -102,16 +102,16 @@ describe('SdkClient', async function () {
   overrideEnvsInMochaDescribe({ GET_RECORD_DEFAULT_TO_CONSENSUS_NODE: true });
 
   before(() => {
-    const hederaNetwork = ConfigService.get('HEDERA_NETWORK')!;
+    const hederaNetwork = ConfigService.get('HEDERA_NETWORK')! as string;
```
acuarica, line 105: is this type assertion needed? if not, we should remove it.

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -102,16 +102,16 @@ describe('SdkClient', async function () {
...
     if (hederaNetwork in constants.CHAIN_IDS) {
       client = Client.forName(hederaNetwork);
     } else {
       client = Client.forNetwork(JSON.parse(hederaNetwork));
     }
 
     client = client.setOperator(
-      AccountId.fromString(ConfigService.get('OPERATOR_ID_MAIN')!),
-      Utils.createPrivateKeyBasedOnFormat(ConfigService.get('OPERATOR_KEY_MAIN')!),
+      AccountId.fromString(ConfigService.get('OPERATOR_ID_MAIN')! as string),
```
acuarica, line 113: is this type assertion needed? if not, we should remove it.

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -102,16 +102,16 @@ describe('SdkClient', async function () {
...
+      Utils.createPrivateKeyBasedOnFormat(ConfigService.get('OPERATOR_KEY_MAIN')! as string),
```
acuarica, line 114: is this type assertion needed? if not, we should remove it.

acuarica (commented):
It's looking good, left a couple of questions. We are getting close!

belloibrahv:
@acuarica I have added the required change, Kindly check when you free.

packages/config-service/tests/src/services/loggerService.spec.ts
```diff
@@ -47,8 +47,8 @@ describe('LoggerService tests', async function () {
   });
 
   it('should be able to return plain information', async () => {
-    const envName = GlobalConfig.ENTRIES.CHAIN_ID.envName;
-    const res = ConfigService.get(envName);
+    const envName = 'CHAIN_ID';
+    const res = ConfigService.get(envName) as string | undefined;
```
acuarica, line 51: is this one needed?

acuarica (dismissed):

acuarica (dismissed):
Looks good! Thanks for your contribution.

quiet-node:
@belloibrahv The build process is failing: https://github.com/hashgraph/hedera-json-rpc-relay/actions/runs/12903875068/job/36018850517?pr=3350#step:12:49. 

It looks like the createPrivateKeyBasedOnFormat() function is failing because it couldn’t parse the correct type from ConfigService.get('OPERATOR_KEY_FORMAT'). It seems that ConfigService.get('OPERATOR_KEY_FORMAT') is returning undefined, as shown in the screenshot below: image

This is resulting in the following errors:
image


@acuarica Could there be anything missing from the approach you proposed? Just want to make sure we’re covering all the bases!

packages/config-service/src/services/globalConfig.ts
```diff
@@ -742,5 +763,10 @@ export class GlobalConfig {
       required: false,
       defaultValue: null,
     },
-  };
+} satisfies { [key: string]: ConfigProperty };
```
acuarica, line 766: for this approach to work, `_CONFIG` must be declared `as const`

```suggestion
} as const satisfies { [key: string]: ConfigProperty };
```

acuarica:
@quiet-node see https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350/files#r1925959856.

packages/config-service/src/services/globalConfig.ts
belloibrahv: @acuarica Should I go ahead to add the suggested changes as specified? and is there things else that need to be added.

acuarica: make sure it compiles correctly locally, see failing job here https://github.com/hashgraph/hedera-json-rpc-relay/pull/3350#issuecomment-2608208535 to find out the command to run.

acuarica: > Should I go ahead to add the suggested changes as specified?

Yes, that would be great, thanks.

belloibrahv:
@acuarica @quiet-node I have added the required changes as suggested, kindly help check it out.

belloibrahv:
@acuarica @quiet-node The CI build is currently failing. Could you please take a look and see if you can spot anything I might have missed? I'm concerned that I may have introduced a change that's causing the failures.

quiet-node:
@belloibrahv so to I was able to resolve all the bugs that were causing the build process to fail and updated all global configs to have their proper default values. However, the tests are still failing and no worries I'm aware of that. I’ll be addressing them soon so let's bear with me.

quiet-node:
@belloibrahv okay the tests are all resolved and looking quite bright now. Let's wait for the team to do review on this next week!

packages/config-service/src/services/globalConfig.ts
```diff
@@ -18,729 +18,804 @@
...
+  : StringTypeToActualType<ExtractTypeStringFromKey<K>>;
+
+/**
+ * Interface defining the structure of a configuration property.
+ * Each property includes the environment variable name, its type,
+ * whether it is required, and its default value.
+ *
+ * @interface ConfigProperty
+ * @property {string} envName - The name of the environment variable.
+ * @property {'string' | 'number' | 'boolean'} type - The data type of the configuration property.
```
quiet-node, line 78: Note for reviewers: Removed array type, as array config values are essentially stringified versions that will ultimately be cast to the string type.

packages/config-service/src/services/validationService.ts
```diff
@@ -30,26 +30,33 @@ export class ValidationService {
...
+   * @param envs - Dictionary of environment variables and their string values
+   * @returns Dictionary with environment variables cast to their proper types
    */
   static typeCasting(envs: NodeJS.Dict<string>): NodeJS.Dict<any> {
     const typeCastedEnvs: NodeJS.Dict<any> = {};
 
     Object.entries(GlobalConfig.ENTRIES).forEach(([entryName, entryInfo]) => {
       if (!envs.hasOwnProperty(entryName)) {
-        if (entryInfo.defaultValue) {
+        if (entryInfo.defaultValue != null) {
```
acuarica, line 59: nit: any specific reason to use `!=` instead of `!==`?

packages/config-service/src/services/index.ts
```diff
@@ -93,11 +95,35 @@ export class ConfigService {
...
+    const configEntry = GlobalConfig.ENTRIES[name];
+    let value = this.getInstance().envs[name] == undefined ? configEntry?.defaultValue : this.getInstance().envs[name];
+
+    if (value == undefined && configEntry?.required) {
+      throw new Error(`Configuration error: ${name} is a mandatory configuration for relay operation.`);
+    }
+
+    if (name === 'CHAIN_ID' && value !== undefined) {
+      value = `0x${Number(value).toString(16)}`;
+    }
```
acuarica, line 125: it looks like it's doing too much. Not sure we should change the values provided by the user. If this it's actually needed, maybe we can do it in the `typeCasting` method.

packages/config-service/src/services/globalConfig.ts
```diff
@@ -18,729 +18,804 @@
...
-  type: string;
+  type: 'string' | 'number' | 'boolean';
   required: boolean;
   defaultValue: string | number | boolean | null;
 }
 
+// Configuration object defining various properties and their metadata.
+// Each property is an object that contains information about the environment variable,
+// its type, whether it is required, and its default value.
+// Example of a configuration entry:
```
acuarica, line 92: nit: I'd use TS docs if possible

packages/config-service/src/services/globalConfig.ts
```diff
@@ -18,729 +18,804 @@
...
+/**
+ * Interface defining the structure of a configuration property.
+ * Each property includes the environment variable name, its type,
+ * whether it is required, and its default value.
+ *
+ * @interface ConfigProperty
+ * @property {string} envName - The name of the environment variable.
+ * @property {'string' | 'number' | 'boolean'} type - The data type of the configuration property.
+ * @property {boolean} required - Indicates if the configuration property is required.
+ * @property {string | number | boolean | null} defaultValue - The default value for the configuration property, which can be a string, number, boolean, or null.
```
acuarica, line 80: nit: we can use TS doc (inline the comment right in the field to avoid repeating the type definition.

packages/config-service/src/services/globalConfig.ts
```diff
@@ -18,729 +18,804 @@
  *
  */
 
+// This type extracts the type string associated with a specific key
+// in the _CONFIG object. It checks if the provided key K is a valid
+// key in _CONFIG. If it is, it retrieves the 'type' property of the
+// corresponding configuration object; otherwise, it resolves to 'never'.
+// Example:
+// - For key '‘OPERATOR_ID_MAIN’', it returns 'string' if defined in _CONFIG.
+// - For an invalid key 'INVALID_KEY', it returns never.
```
acuarica, line 27: nit: ts doc format?

acuarica (commented):
Thanks Logan for taking care of this.

belloibrahv:
Great job @quiet-node 🚀

packages/config-service/src/services/validationService.ts
quiet-node: Ah yeah, != ensures that entryInfo.defaultValue is neither null nor undefined, making the condition more inclusive. On the other hand, !== null would only check for null, allowing the condition to pass if defaultValue is undefined. This makes sure in the future if any new config comes in with `undefined` defaultValue, this would work as expected.

acuarica: Mmm, then I'd leave it as before `if (entryInfo.defaultValue)` that's already inclusive. Although if we are checking for `null` then we should be strict using the `!==` operator.

quiet-node: Ah yeah, I meant to mention the issue with if (entryInfo.defaultValue) in my last comment but forgot. Some defaultValue values, such as `false` or `0`, are falsy, which can cause the condition to be bypassed and override the type-casted value.

packages/config-service/src/services/index.ts
quiet-node: Yeah, I initially placed it in the typeCasting method, but in many test scenarios, ConfigService.get() is called directly right after an environment override, like in this case:

withOverriddenEnvsInMochaTest({ CHAIN_ID: '123' }, () => {
    it('should set chainId from CHAIN_ID environment variable', () => {
      Relay = new RelayImpl(logger, new Registry());
      const actualNetVersion = Relay.net().version();
      expect(actualNetVersion).to.equal('123');
    });
  });
Here, the call to `Relay.net().version()` essentially triggers `ConfigService.get("CHAIN_ID")`. Since the `ConfigService` class is initialized at startup along with the `typeCasting()` method, it initially retrieves the correct CHAIN_ID from the .env file. However, when CHAIN_ID is overridden with the value '123', the change doesn’t take effect, and the conversion to hexadecimal format is skipped, resulting in a mismatched value.

That’s why I moved this logic to the get() method. You’re right—it may seem somewhat arbitrary to place it there, but since many scenarios call ConfigService.get() with overridden values, typeCasting wouldn’t always be involved.

quiet-node: > Not sure we should change the values provided by the user
This got me thinking, though. The purpose of this logic is to streamline the use of ConfigService.get("CHAIN_ID") throughout the codebase. These changes help clean up the codebase, as shown below.

```diff
- const hederaNetwork: string = (ConfigService.get('HEDERA_NETWORK') || '{}').toLowerCase();
- this.chainId = ConfigService.get('CHAIN_ID') || constants.CHAIN_IDS[hederaNetwork] || '298';
- if (this.chainId.startsWith('0x')) this.chainId = parseInt(this.chainId, 16).toString();
+ this.chainId = parseInt(ConfigService.get('CHAIN_ID'), 16).toString();
```


or 
```diff
-     // @ts-ignore
-    const hederaNetwork: string = (ConfigService.get('HEDERA_NETWORK') || '{}').toLowerCase();
-    const configuredChainId = ConfigService.get('CHAIN_ID') || constants.CHAIN_IDS[hederaNetwork] || '298';
-    const chainId = prepend0x(Number(configuredChainId).toString(16));
+   const chainId = ConfigService.get('CHAIN_ID');
```

quiet-node: It's obviously a valid concern if we should modify the value users provide but it would help clean up the code and the value can be consistent. What do you think @acuarica?

packages/config-service/src/services/globalConfig.ts
quiet-node: Gotcha good point! Updated!

packages/config-service/src/services/globalConfig.ts
quiet-node: Solid! Thanks for the suggestions! Updated!

packages/relay/src/lib/clients/cache/localLRUCache.ts
```diff
@@ -40,11 +39,9 @@ export class LocalLRUCache implements ICacheClient {
    */
   private readonly options: LimitedByCount & LimitedByTTL = {
     // The maximum number (or size) of items that remain in the cache (assuming no TTL pruning or explicit deletions).
-    // @ts-ignore
-    max: Number.parseInt(ConfigService.get('CACHE_MAX') ?? constants.CACHE_MAX.toString()),
+    max: ConfigService.get('CACHE_MAX'),
```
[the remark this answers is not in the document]
quiet-node: Good question! So the idea is to use the actual configuration value instead of a constant. This ensures that if the CACHE_MAX config is set to a different value in the .env file, the application picks up that value dynamically. Using a constant here wouldn’t allow for flexibility, as it would always remain the same regardless of changes in the environment configuration.

packages/relay/src/lib/net.ts
```diff
@@ -30,10 +29,7 @@ export class NetImpl implements Net {
   constructor(client: Client) {
     this.client = client;
 
-    // @ts-ignore
-    const hederaNetwork: string = (ConfigService.get('HEDERA_NETWORK') || '{}').toLowerCase();
-    this.chainId = ConfigService.get('CHAIN_ID') || constants.CHAIN_IDS[hederaNetwork] || '298';
-    if (this.chainId.startsWith('0x')) this.chainId = parseInt(this.chainId, 16).toString();
+    this.chainId = parseInt(ConfigService.get('CHAIN_ID'), 16).toString();
```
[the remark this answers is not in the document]
quiet-node: Ah yeah great question! So this here would actually do the opposite of what we’re doing in ConfigService. The formatting logic in ConfigService ensures that the ChainID, whether in decimal or hexadecimal, is always converted to a hexadecimal string with a 0x prefix. However, this line—`parseInt(ConfigService.get('CHAIN_ID'), 16).toString()`—converts a hexadecimal value into a decimal format, as net_version returns a decimal value, not hexadecimal.

packages/relay/src/lib/clients/cache/localLRUCache.ts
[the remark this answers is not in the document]
acuarica: I understand that. I was referring to how and where to set the default value. Maybe we could delete `constants.CACHE_MAX` altogether if we use `ConfigService.get('CACHE_MAX') everywhere.

packages/config-service/src/services/index.ts
quiet-node: Yeah, it had me scratching my head since it doesn’t happen in production. But given the benefits of more structured code and consistent values, I decided to go with it.

packages/relay/src/lib/clients/cache/localLRUCache.ts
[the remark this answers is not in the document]
quiet-node: Yeah, I thought of that too—since multiple variables now use defaultValue, the constants are no longer in use. However, I didn’t remove them initially, as I thought they might still be useful for testing or other purposes.

Now that you’ve brought this up, I agree—we should remove any unused constants. Great point indeed!

acuarica (dismissed):
Thanks @belloibrahv and @quiet-node for putting the effort to get this done. Great stuff!

natanasow (approved):
Fantastic work!

belloibrahv:
@quiet-node @acuarica @natanasow @Nana-EC 🚀 🚀🚀 Thank You all

acuarica (dismissed):

acuarica (approved):

natanasow (approved):

Ferparishuertas (approved):

quiet-node:
Thanks, team, for the great responses and feedback! 

Congrats on the merge, @belloibrahv! Appreciate your effort and patience—looking forward to more of your contributions in the future!

belloibrahv:
Thank You all for helping out, I look forward to working and collaborating on more work.

PR #3387, opened 2025-01-14
feat: transaction with `CONTRACT_NEGATIVE_VALUE` breaks some routes

### LINKED ISSUE #3367: Transaction with CONTRACT_NEGATIVE_VALUE prevents entire block from being retrieved
### Description

It seems like `eth_getBlockByNumber` fails because one of the transactions in the block was a CONTRACT_NEGATIVE_VALUE failure. The expectation is for the block to be successfully returned.

### Steps to reproduce

1. call `eth_getBlockByNumber` on block [73298898](https://hashscan.io/mainnet/block/73298898)
2. see "Error invoking RPC: Invalid value - cannot pass negative number" response

The problematic transaction is [here](https://hashscan.io/mainnet/transaction/1735241436.856862230)

### Additional context

_No response_

### Hedera network

mainnet

### Version

v0.61.0

### Operating system

Linux

natanasow:
**Description**:

It seems like `eth_getBlockByNumber` fails because one of the transactions in the block was a CONTRACT_NEGATIVE_VALUE failure. The expectation is for the block to be successfully returned.

**Steps to reproduce**:

1. call `eth_getBlockByNumber` on block [73298898](https://hashscan.io/mainnet/block/73298898)
2. see "Error invoking RPC: Invalid value - cannot pass negative number" response

The problematic transaction is [here](https://hashscan.io/mainnet/transaction/1735241436.856862230)

**Related issue(s)**:

Fixes #3367 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -268,7 +268,7 @@ const nanOrNumberTo0x = (input: number | BigNumber | bigint | null): string => {
 
 const nanOrNumberInt64To0x = (input: number | BigNumber | bigint | null): string => {
   // converting to string and then back to int is fixing a typescript warning
-  if (input && parseInt(input.toString()) < 0) {
+  if (input && Number(input) < 0) {
     // the hex of a negative number can be obtained from the binary value of that number positive value
     // the binary value needs to be negated and then to be incremented by 1
 
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -268,11 +268,11 @@ const nanOrNumberTo0x = (input: number | BigNumber | bigint | null): string => {
 
 const nanOrNumberInt64To0x = (input: number | BigNumber | bigint | null): string => {
   if (input && input < 0) {
-    // the hex of a negative number can be obtained from the binary value of that number positive value,
-    // the binary value needs to be negated and then, to add 1,
+    // the hex of a negative number can be obtained from the binary value of that number positive value
+    // the binary value needs to be negated and then to be incremented by 1
 
     // how the transformation works (using 16 bits)
-    // a 16 bits integer variables has values from -32768 to +32767, so:
+    // a 16 bits integer variables have values from -32768 to +32767, so:
     // 0      - 0x0000 - 0000 0000 0000 0000
     // 32767  - 0x7fff - 0111 1111 1111 1111
     // -32768 - 0x8000 - 1000 0000 0000 0000
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -267,7 +267,8 @@ const nanOrNumberTo0x = (input: number | BigNumber | bigint | null): string => {
 };
 
 const nanOrNumberInt64To0x = (input: number | BigNumber | bigint | null): string => {
-  if (input && input < 0) {
+  // converting to string and then back to int is fixing a typescript warning
+  if (input && parseInt(input.toString()) < 0) {
     // the hex of a negative number can be obtained from the binary value of that number positive value
     // the binary value needs to be negated and then to be incremented by 1
 
@@ -281,11 +282,11 @@ const nanOrNumberInt64To0x = (input: number | BigNumber | bigint | null): string
     // converting int16 -10 will be done as following:
     // - make it positive = 10
     // - 16 bits binary value of 10 = 0000 0000 0000 1010
-    // - inverse the bytes = 1111 1111 1111 0101
+    // - inverse the bits = 1111 1111 1111 0101
     // - adding +1 = 1111 1111 1111 0110
     // - 1111 1111 1111 0110 bits = 0xfff6
 
-    // we're using 64 bits integer because that the type returned by the mirror node - int64
+    // we're using 64 bits integer because that's the type returned by the mirror node - int64
     const bits = 64;
     return numberTo0x((BigInt(input.toString()) + (BigInt(1) << BigInt(bits))) % (BigInt(1) << BigInt(bits)));
   }
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -288,6 +288,12 @@ const nanOrNumberInt64To0x = (input: number | BigNumber | bigint | null): string
 
     // we're using 64 bits integer because that's the type returned by the mirror node - int64
     const bits = 64;
+    // this mathematical expression serves as a shortcut for performing the two’s complement conversion
+    // e.g. input = -10
+    // we have: (BigInt(1) << BigInt(bits)) = 1 << 64 = 2^64 = 18446744073709551616
+    // then: (BigInt(input.toString()) + (BigInt(1) << BigInt(bits))) = -10 + 2^64 = 18446744073709551606
+    // this effectively represents -10 in an unsigned 64-bit representation:18446744073709551606 = 0xFFFFFFFFFFFFFFF6
+    // finally, the modulo operation: % (1 << 64)
     return numberTo0x((BigInt(input.toString()) + (BigInt(1) << BigInt(bits))) % (BigInt(1) << BigInt(bits)));
   }
 
--- a/packages/relay/src/formatters.ts
+++ b/packages/relay/src/formatters.ts
@@ -18,11 +18,12 @@
  *
  */
 
-import crypto from 'crypto';
-import constants from './lib/constants';
-import { BigNumber as BN } from 'bignumber.js';
-import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
 import { ConfigService } from '@hashgraph/json-rpc-config-service/dist/services';
+import { BigNumber } from '@hashgraph/sdk/lib/Transfer';
+import { BigNumber as BN } from 'bignumber.js';
+import crypto from 'crypto';
+
+import constants from './lib/constants';
 import { Transaction, Transaction1559, Transaction2930 } from './lib/model';
 
 const EMPTY_HEX = '0x';
@@ -178,7 +179,7 @@ const formatContractResult = (cr: any) => {
     transactionIndex: nullableNumberTo0x(cr.transaction_index),
     type: cr.type === null ? '0x0' : nanOrNumberTo0x(cr.type),
     v: cr.v === null ? '0x0' : nanOrNumberTo0x(cr.v),
-    value: nanOrNumberTo0x(tinybarsToWeibars(cr.amount)),
+    value: nanOrNumberInt64To0x(tinybarsToWeibars(cr.amount, true)),
     // for legacy EIP155 with tx.chainId=0x0, mirror-node will return a '0x' (EMPTY_HEX) value for contract result's chain_id
     //   which is incompatibile with certain tools (i.e. foundry). By setting this field, chainId, to undefined, the end jsonrpc
     //   object will leave out this field, which is the proper behavior for other tools to be compatible with.
@@ -265,6 +266,33 @@ const nanOrNumberTo0x = (input: number | BigNumber | bigint | null): string => {
   return input == null || Number.isNaN(input) ? numberTo0x(0) : numberTo0x(input);
 };
 
+const nanOrNumberInt64To0x = (input: number | BigNumber | bigint | null): string => {
+  if (input && input < 0) {
+    // the hex of a negative number can be obtained from the binary value of that number positive value,
+    // the binary value needs to be negated and then, to add 1,
+
+    // how the transformation works (using 16 bits)
+    // a 16 bits integer variables has values from -32768 to +32767, so:
+    // 0      - 0x0000 - 0000 0000 0000 0000
+    // 32767  - 0x7fff - 0111 1111 1111 1111
+    // -32768 - 0x8000 - 1000 0000 0000 0000
+    // -1     - 0xffff - 1111 1111 1111 1111
+
+    // converting int16 -10 will be done as following:
+    // - make it positive = 10
+    // - 16 bits binary value of 10 = 0000 0000 0000 1010
+    // - inverse the bytes = 1111 1111 1111 0101
+    // - adding +1 = 1111 1111 1111 0110
+    // - 1111 1111 1111 0110 bits = 0xfff6
+
+    // we're using 64 bits integer because that the type returned by the mirror node - int64
+    const bits = 64;
+    return numberTo0x((BigInt(input.toString()) + (BigInt(1) << BigInt(bits))) % (BigInt(1) << BigInt(bits)));
+  }
+
+  return nanOrNumberTo0x(input);
+};
+
 const toHash32 = (value: string): string => {
   return value.substring(0, 66);
 };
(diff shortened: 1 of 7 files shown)


packages/relay/src/formatters.ts
```diff
@@ -265,6 +266,33 @@ const nanOrNumberTo0x = (input: number | BigNumber | bigint | null): string => {
   return input == null || Number.isNaN(input) ? numberTo0x(0) : numberTo0x(input);
 };
 
+const nanOrNumberInt64To0x = (input: number | BigNumber | bigint | null): string => {
+  if (input && input < 0) {
```
quiet-node, line 270: This got a warning 
```
Operator '<' cannot be applied to types 'number | bigint | BigNumber' and 'number'.ts(2365)
```

packages/relay/src/formatters.ts
```diff
@@ -265,6 +266,33 @@ const nanOrNumberTo0x = (input: number | BigNumber | bigint | null): string => {
...
+    // -1     - 0xffff - 1111 1111 1111 1111
+
+    // converting int16 -10 will be done as following:
+    // - make it positive = 10
+    // - 16 bits binary value of 10 = 0000 0000 0000 1010
+    // - inverse the bytes = 1111 1111 1111 0101
+    // - adding +1 = 1111 1111 1111 0110
+    // - 1111 1111 1111 0110 bits = 0xfff6
+
+    // we're using 64 bits integer because that the type returned by the mirror node - int64
```
quiet-node, line 288: great explanation 🙌

quiet-node (changes requested):
Great work overall. Blocking as one warning needed to be addressed. 

Also, is it feasible to add an e2e test to avoid regression?

packages/relay/src/formatters.ts
```diff
@@ -265,6 +266,33 @@ const nanOrNumberTo0x = (input: number | BigNumber | bigint | null): string => {
...
+    const bits = 64;
+    return numberTo0x((BigInt(input.toString()) + (BigInt(1) << BigInt(bits))) % (BigInt(1) << BigInt(bits)));
```
acuarica, line 290: question: do we expect `number` bigger than `53` bits? so it's not automatically coerced to double https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

natanasow: Yep, I don't know how often it will happen but we definitely must handle values bigger than `Number.MAX_SAFE_INTEGER`. 

The max Int64 positive value is `0x7fffffffffffffff` which has 63 "usable" bits. 
```
0111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111
```

That's why we always convert the incoming value to BigInt here :point_down: 
```typescript
BigInt(input.toString())
```

Hope that makes sense or am I wrong?

packages/relay/src/formatters.ts
natanasow: Fixed.

natanasow:
> Great work overall. Blocking as one warning needed to be addressed.
> 
> Also, is it feasible to add an e2e test to avoid regression?

Adding an e2e test is not a straightforward process because that kind of revert can't be reproduced with an `EthereumTransaction.` In order to re-create it we have to execute `ContractCall` SDK call which is out of the relay scope IMO.

I have no problem adding that test if you insist :rocket: :pray:.

packages/relay/src/formatters.ts
acuarica: Mmm, I'm not sure that `BigInt(input.toString())` works as expected when `input` being a `number` uses more than `53` bits. Say

```js
> const x = 0x7fffffffffffffff
> typeof(x)
'number'
> BigInt(x.toString()) - 0x7fffffffffffffffn
193n
```

but the difference should be zero, right?

natanasow: You're right, the difference should be zero. Nice catch :clap: and thank you. That's something I must add more unit tests tomorrow.

Here are some more quick tests, it seems that `.toString()` messed things up, also using a shadow conversion like postfixing `n` doesn't seem to work as expected.
```typescript
> const x = 0x7fffffffffffffff
undefined
> typeof(x)
'number'
> BigInt(x.toString()) - 0x7fffffffffffffffn
193n
> BigInt(x) - 0x7fffffffffffffffn
1n
> BigInt(x) - BigInt(0x7fffffffffffffff)
0n
> BigInt(x) - BigInt(9223372036854776000)
0n
```

Do you have an idea how this can be completely fixed? I think removing `.toString()` is enough but that will be confirmed when I push the unit tests covering wide-bit range numbers.

natanasow: Makes sense. Will try to refactor it in an elegant way and once I'm done will ping you again for review.

quiet-node:
> > Great work overall. Blocking as one warning needed to be addressed.
> > Also, is it feasible to add an e2e test to avoid regression?
> 
> Adding an e2e test is not a straightforward process because that kind of revert can't be reproduced with an `EthereumTransaction.` In order to re-create it we have to execute `ContractCall` SDK call which is out of the relay scope IMO.
> 
> I have no problem adding that test if you insist 🚀 🙏.

Ah I see. Just curious how would you reproduice the problem?

packages/relay/src/formatters.ts
```diff
@@ -265,6 +266,34 @@ const nanOrNumberTo0x = (input: number | BigNumber | bigint | null): string => {
   return input == null || Number.isNaN(input) ? numberTo0x(0) : numberTo0x(input);
 };
 
+const nanOrNumberInt64To0x = (input: number | BigNumber | bigint | null): string => {
+  // converting to string and then back to int is fixing a typescript warning
+  if (input && parseInt(input.toString()) < 0) {
```
quiet-node, line 271: Would below work? 
```suggestion
  if (input && Number(input) < 0) {
```

natanasow: Fixed.

natanasow:
> > > Great work overall. Blocking as one warning needed to be addressed.
> > > Also, is it feasible to add an e2e test to avoid regression?
> > 
> > 
> > Adding an e2e test is not a straightforward process because that kind of revert can't be reproduced with an `EthereumTransaction.` In order to re-create it we have to execute `ContractCall` SDK call which is out of the relay scope IMO.
> > I have no problem adding that test if you insist 🚀 🙏.
> 
> Ah I see. Just curious how would you reproduice the problem?

Sending a negative amount of ContractCallTransaction will be enough while it accepts int64 (not uint64).

Here is the link to the doc https://docs.hedera.com/hedera/sdks-and-apis/hedera-api/readme-1-1/contractcall#contractcalltransactionbody.

packages/relay/src/formatters.ts
acuarica: What would be the advantage of using `json-bigint` over `bigint`? Only parsing? doesn't `bigint` already comes with parsing capabilities?

natanasow: `json-bigint` is doing `JSONBigInt.parse(data)` but under the hood, it transforms numeric values to `bigint` instead of JS Number.

If we're doing `JSON.parse()` and then manually transforming numeric values to Number, `JSON.parse()` already rounded the number and we lost it.

`json-bigint` doesn't replace `bigint` and they are not correlated. `json-big-int` is used instead of `JSON.parse()` due to its capability of converting string json input directly to `bigint` instead of JS Number.

acuarica: Thanks for the clarification.

quiet-node:
> > > > Great work overall. Blocking as one warning needed to be addressed.
> > > > Also, is it feasible to add an e2e test to avoid regression?
> > > 
> > > 
> > > Adding an e2e test is not a straightforward process because that kind of revert can't be reproduced with an `EthereumTransaction.` In order to re-create it we have to execute `ContractCall` SDK call which is out of the relay scope IMO.
> > > I have no problem adding that test if you insist 🚀 🙏.
> > 
> > 
> > Ah I see. Just curious how would you reproduice the problem?
> 
> Sending a negative amount via ContractCallTransaction will be enough while it accepts int64 (not uint64).
> 
> Here is the link to the doc https://docs.hedera.com/hedera/sdks-and-apis/hedera-api/readme-1-1/contractcall#contractcalltransactionbody.

Ah I see. So if it's feasible and do not take crazy amount of time, I would encourage we have an e2e just to make sure no regression later. Thanks much Nikki!

Nana-EC:
> > > > > Great work overall. Blocking as one warning needed to be addressed.
> 
> > > > > Also, is it feasible to add an e2e test to avoid regression?
> 
> > > > 
> 
> > > > 
> 
> > > > Adding an e2e test is not a straightforward process because that kind of revert can't be reproduced with an `EthereumTransaction.` In order to re-create it we have to execute `ContractCall` SDK call which is out of the relay scope IMO.
> 
> > > > I have no problem adding that test if you insist 🚀 🙏.
> 
> > > 
> 
> > > 
> 
> > > Ah I see. Just curious how would you reproduice the problem?
> 
> > 
> 
> > Sending a negative amount via ContractCallTransaction will be enough while it accepts int64 (not uint64).
> 
> > 
> 
> > Here is the link to the doc https://docs.hedera.com/hedera/sdks-and-apis/hedera-api/readme-1-1/contractcall#contractcalltransactionbody.
> 
> 
> 
> Ah I see. So if it's feasible and do not take crazy amount of time, I would encourage we have an e2e just to make sure no regression later. Thanks much Nikki!

At the least we could mock the CN response that would cause this and have an integration test that ensures the relay responds as expected.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -361,6 +362,28 @@ export class MirrorNodeClient {
...
+              if (data) {
+                try {
+                  // try to parse it, if the json is valid, numbers within it will be converted
+                  // this case will happen on almost every GET mirror node call
+                  return JSONBigInt.parse(data);
+                } catch (e) {
+                  // in some unit tests, the mocked returned json is not property formatted
+                  // so we have to preprocess it here with JSON.stringify()
+                  return JSONBigInt.parse(JSON.stringify(data));
+                }
```
quiet-node, line 381: Hmmm then should we parse the stringified data at the begining?

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -361,6 +362,28 @@ export class MirrorNodeClient {
         if (pathLabel == MirrorNodeClient.GET_CONTRACTS_RESULTS_OPCODES) {
           response = await this.web3Client.get<T>(path, axiosRequestConfig);
         } else {
+          // JS supports up to 53 bits integers, once a number with more bits is converted to a js Number type, the initial value is already lost
+          // in order to fix that, we have to use a custom JSON parser that hooks up before
+          // the default axios JSON.parse conversion where the value will be rounded and lost
+          // JSONBigInt reads the string representation of received JSON and formats each number to BigNumber
```
quiet-node, line 368: I personally find this block is a tiny bit confusing, would you consider update to something like bellow
```suggestion
          // JavaScript supports integers only up to 53 bits. When a number exceeding this limit is converted to a JS Number type, precision is lost due to rounding.
          // To prevent this, `transformResponse` is used to intercept and process the response before Axios’s default JSON.parse conversion.
          // JSONBigInt reads the string representation from the received JSON and converts large numbers into BigNumber objects to maintain accuracy.```

File: packages/server/tests/acceptance/rpc_batch1.spec.ts [hunk 7]
```diff
@@ -726,6 +727,43 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+          requestId,
+        );
+        const txHash = mirrorResult.hash;
+        const blockResult = await relay.call(
+          RelayCalls.ETH_ENDPOINTS.ETH_GET_BLOCK_BY_NUMBER,
+          [numberTo0x(mirrorResult.block_number), true],
+          requestIdPrefix,
+        );
+        expect(blockResult.transactions).to.not.be.empty;
+        expect(blockResult.transactions.map((tx) => tx.hash)).to.contain(txHash);
```
quiet-node, line 765: Could it possible to see the negative value (or its hex representation) from here?

quiet-node (commented):
Great work overall! Left some comments and suggestions

packages/relay/src/formatters.ts
```diff
@@ -265,6 +266,34 @@ const nanOrNumberTo0x = (input: number | BigNumber | bigint | null): string => {
...
+    // converting int16 -10 will be done as following:
+    // - make it positive = 10
+    // - 16 bits binary value of 10 = 0000 0000 0000 1010
+    // - inverse the bits = 1111 1111 1111 0101
+    // - adding +1 = 1111 1111 1111 0110
+    // - 1111 1111 1111 0110 bits = 0xfff6
+
+    // we're using 64 bits integer because that's the type returned by the mirror node - int64
+    const bits = 64;
+    return numberTo0x((BigInt(input.toString()) + (BigInt(1) << BigInt(bits))) % (BigInt(1) << BigInt(bits)));
```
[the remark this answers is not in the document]
natanasow: It's all correct. Added the example.

packages/relay/src/lib/clients/mirrorNodeClient.ts
natanasow: Fixed.

packages/server/tests/acceptance/rpc_batch1.spec.ts
natanasow: Yep, and I added an expectation.

packages/relay/src/lib/clients/mirrorNodeClient.ts
natanasow: I don't think so, if the passed data can be parsed with no exception that means it's in the right format. We need to stringify it only if the returned data is not already stringified e.g. here https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/tests/lib/mirrorNodeClient.spec.ts#L277 and that happens only in tests (mirror node HTTP requests always return the data stringified).

Another approach is to refactor all tests where non-stringified mocked data is returned. E.g. here https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/tests/lib/mirrorNodeClient.spec.ts#L275

E.g. this:
```typescript
    mock.onGet(`accounts/${alias}${noTransactions}`).reply(200, {
      transactions: [
        {
          nonce: 3,
        },
      ],
      links: {
        next: null,
      },
    });
```
should be refactored to:
```typescript
    mock.onGet(`accounts/${alias}${noTransactions}`).reply(200, JSON.stringify({
      transactions: [
        {
          nonce: 3,
        },
      ],
      links: {
        next: null,
      },
    }));
```
then we can completely get rid of:
```typescript
              // if the data is not valid, just return it to stick to the current behaviour
              if (data) {
                try {
                  // try to parse it, if the json is valid, numbers within it will be converted
                  // this case will happen on almost every GET mirror node call
                  return JSONBigInt.parse(data);
                } catch (e) {
                  // in some unit tests, the mocked returned json is not property formatted
                  // so we have to preprocess it here with JSON.stringify()
                  return JSONBigInt.parse(JSON.stringify(data));
                }
              }

              return data;
```
and will become just to:
```typescript
              // if the data is not valid, just return it to stick to the current behaviour
              if (data) {
                return JSONBigInt.parse(data);
              }

              return data;
```

Does it make sense now? I can create a follow-up issue/PR removing the try/catch and fixing all tests or I can do it in this one but it will become pretty big. 401 tests that need to be fixed if we get rid of try/catch.
```
  822 passing (21s)
  401 failing
```

natanasow: @quiet-node just created the follow-up PR https://github.com/hashgraph/hedera-json-rpc-relay/pull/3435.

quiet-node: Great point! Thanks for the effort!

packages/relay/src/formatters.ts
```diff
@@ -265,6 +266,40 @@ const nanOrNumberTo0x = (input: number | BigNumber | bigint | null): string => {
...
+    // - 1111 1111 1111 0110 bits = 0xfff6
+
+    // we're using 64 bits integer because that's the type returned by the mirror node - int64
+    const bits = 64;
+    // this mathematical expression serves as a shortcut for performing the two’s complement conversion
+    // e.g. input = -10
+    // we have: (BigInt(1) << BigInt(bits)) = 1 << 64 = 2^64 = 18446744073709551616
+    // then: (BigInt(input.toString()) + (BigInt(1) << BigInt(bits))) = -10 + 2^64 = 18446744073709551606
+    // this effectively represents -10 in an unsigned 64-bit representation:18446744073709551606 = 0xFFFFFFFFFFFFFFF6
+    // finally, the modulo operation: % (1 << 64)
```
quiet-node, line 296: This could explain a little more what it does.

quiet-node (approved):
LGTM 1 nit

konstantinabl (approved):

PR #3394, opened 2025-01-15
feat: added DEPENDENT_SERVICE_IMMATURE_RECORDS predefined error for immature records scenarios

### LINKED ISSUE #3392: [Failed-Request-Alert-Tuning] Assign Appropriate HTTP Status Code for Immature Records error
Description

The error caused by immature records in the Mirror Node (MN) is currently returning an HTTP 500 status code. This triggers the failed-request-too-high alerts unnecessarily, making it difficult to identify and prioritize critical issues. Since this scenario is expected behavior and not indicative of a server error, it should use a more appropriate HTTP status code.

This change will ensure that alerts are only triggered for critical issues while maintaining visibility of immature record occurrences through metrics for observability.

Solution
	•	Define a dedicated error type specifically for scenarios involving immature records.
	•	Update the HTTP status code for the “immature records found” error to 502 (Bad Gateway) instead of 500.

quiet-node:
**Description**:
This PR introduces a new predefined error, DEPENDENT_SERVICE_IMMATURE_RECORDS, with the JSON-RPC code `-32015`, mapped to the HTTP status code `503`. This improves error handling for cases where the dependent Mirror Node service returns immature records.

- `-32015`: Chosen as it fits within the JSON-RPC application-defined server error range (-32000 to -32099) and avoids conflicts with existing error codes.
- `HTTP 503`: Represents a temporary unavailability of a dependent service, signaling clients to retry.

**Related issue(s)**:

Fixes #3392 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/tests/lib/mirrorNodeClient.spec.ts
+++ b/packages/relay/tests/lib/mirrorNodeClient.spec.ts
@@ -686,8 +686,8 @@ describe('MirrorNodeClient', async function () {
 
   it('`getContractResultsWithRetry` should return immature records after exhausting maximum retry attempts', async () => {
     const hash = '0x2a563af33c4871b51a8b108aa2fe1dd5280a30dfb7236170ae5e5e7957eb6393';
-    // Mock 11 sequential calls that return immature records - as default polling counts (10)
-    [...Array(11)].reduce((mockChain) => {
+    // Mock 10 sequential calls that return immature records - equals to the default polling counts (10) - should throw an error at the last polling attempt
+    [...Array(10)].reduce((mockChain) => {
       return mockChain.onGet(`contracts/results/${hash}`).replyOnce(200, {
         ...detailedContractResult,
         transaction_index: null,
@@ -696,17 +696,19 @@ describe('MirrorNodeClient', async function () {
       });
     }, mock);
 
-    const result = await mirrorNodeInstance.getContractResultWithRetry(
-      mirrorNodeInstance.getContractResult.name,
-      [hash, requestDetails],
-      requestDetails,
-    );
+    try {
+      await mirrorNodeInstance.getContractResultWithRetry(
+        mirrorNodeInstance.getContractResult.name,
+        [hash, requestDetails],
+        requestDetails,
+      );
+      expect.fail('should have thrown an error');
+    } catch (error) {
+      expect(error).to.exist;
+      expect(error).to.eq(predefined.DEPENDENT_SERVICE_IMMATURE_RECORDS);
+    }
 
-    expect(result).to.exist;
-    expect(result.transaction_index).equal(null);
-    expect(result.block_number).equal(null);
-    expect(result.block_hash).equal('0x');
-    expect(mock.history.get.length).to.eq(11);
+    expect(mock.history.get.length).to.eq(10);
   });
 
   it('`getContractResults` detailed', async () => {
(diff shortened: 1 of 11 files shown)


Ferparishuertas:
503 is Service Unavaialable. How will we distinguis the normal 503 and the inmature logs?
Can mirror node distinguish the error and send differente codes?
on the other hand yesterday we mentiones 502s, on mirror node. I suggest to add an error code too

Related to this, whats the current alert logic? What will happen with 502s for example?

quiet-node:
> 503 is Service Unavaialable. How will we distinguish the normal 503 and the inmature logs? Can mirror node distinguish the error and send differente codes? on the other hand yesterday we mentiones 502s, on mirror node. I suggest to add an error code too
> 
> Related to this, whats the current alert logic? What will happen with 502s for example?

Thanks @Ferparishuertas for the comment.

> **503 is Service Unavailable. How will we distinguish the normal 503 and the immature logs?**  

The intent here is to distinguish errors arising from immature records from other internal errors within the Relay module. Up until this PR, any internal error caused by the Relay module has been consistently mapped to a 500 Internal Server Error. This change introduces a 503 status code specifically for immature records scenarios, making it the first such instance where a dependent service issue is explicitly identified with a 503.  

> **Can the Mirror Node distinguish the error and send different codes?**  

While the Mirror Node likely has the capability to distinguish errors and send different codes, it is important to note that they are actively addressing the root cause of the replica synchronization issue in their database. Their long-term solution aims to prevent such scenarios altogether. However, as a downstream service, the Relay module is responsible for standardizing error handling, and this PR assigns the 503 code to these cases in alignment with HTTP semantics.  

> **On the other hand, yesterday we mentioned 502s on the Mirror Node.**  

After some exploration and research, I find that the 503 status code is quite more appropriate for these scenarios. While a 502 Bad Gateway indicates that the server received an invalidd response from an upstream server, typically refers to scenarios where the response is fundamentally invalid or cannot be processed—such as malformed headers, incomplete payloads, or protocol-level errors. In contrast, immature records, while incomplete, are still HTTP-valid responses that conform to protocol standards but lack the expected data due to transient synchronization delays. As a result, a 503 Service Unavailable status code is believed to be better aligns with this condition, signaling a temporary issue with the dependent Mirror Node service. Notably, when the request is retried after some time (off the flow), the records are fully complete, with no nullable or empty fields, further reinforcing the appropriateness of the 503 designation for these cases. Let me know your thoughts; I’m open to discussing this further.

> **I suggest adding an error code too.**  

Could you clarify if you are suggesting introducing a new code for 502 scenarios or proposing something else?  

> **Related to this, what’s the current alert logic?**  

The current alerting mechanism calculates the proportion of non-acceptable (TBD on this name) HTTP response codes (status codes not equal to 200, 400, or 501) for specific RPC relay methods relative to the total number of response codes for those methods and namespaces within a specified time range. If the failed requests exceed 5% of total requests within this range, an alert is triggered and sent to the alerting channel.  

> **What will happen with 502s, for example?**  

Since 502 is not currently included in the specific list of acceptable codes (200, 400, or 501), it would be categorized as a failed request. If the cumulative percentage of all failed requests, including 502s, crosses the 5% threshold, an alert would be issued accordingly.

Ferparishuertas:
Thanks for all the clarifications. That said, I still have the same
question. Let me formalize it in a more precise manner

- Are we going to distinguish 503 errors at mirror node level at metrics
level from immature logs where we map it to 503?
- If mirror node gives a 502, which error code are we showing

My overall question is how to distinguish our errors, vs dependant service
errors

quiet-node:
Hey @Ferparishuertas thanks for the quick response.

> Are we going to distinguish 503 errors at mirror node level at metrics level from immature logs where we map it to 503?

Ahhh so currently, any failure from upstream servers (e.g. MN or CN) is mapped to a 500 Internal Server Error by the Relay module. It’s important to note that immature records are not technically errors; they represent valid responses returned by the Mirror Node, albeit incomplete due to synchronization delays.  

In the Relay module, immature records are explicitly treated as "errors" for the purpose of standardization and are mapped to a 503 Service Unavailable status code. Thus:  
- A 503 error directly from the Mirror Node is mapped to a 500 error by the Relay.  
- Immature records from the Mirror Node are mapped to a 503 error by the Relay.  

> If mirror node gives a 502, which error code are we showing
As mentioned above, this will be 500.

Ferparishuertas (approved):
LGTM

packages/relay/tests/lib/mirrorNodeClient.spec.ts
```diff
@@ -696,17 +696,19 @@ describe('MirrorNodeClient', async function () {
       });
     }, mock);
 
-    const result = await mirrorNodeInstance.getContractResultWithRetry(
-      mirrorNodeInstance.getContractResult.name,
-      [hash, requestDetails],
-      requestDetails,
-    );
+    try {
```
acuarica, line 699: nit: you may want to use https://www.chaijs.com/api/bdd/#method_throw to be more explicit about the intent.

acuarica (approved):
left one nit

packages/relay/tests/lib/mirrorNodeClient.spec.ts
quiet-node: Great point! Will address in future PR!

PR #3431, opened 2025-01-28
feat: enhanced eth_getLogs with timestamp range validation and new error handling

### LINKED ISSUE #3428: [eth_getLogs Failure] High Failure Rate with eth_getLogs Requests
### Problem

The Relay is experiencing a high failure rate (currently 17%) with eth_getLogs requests. Logs show approximately 200K entries. 

So far there are two case scenarios:
- Case 1: When fromBlock refers to an excessively old block with a block.timestamp beyond the 7-day range, eth_getLogs fails with a 500 error because the Mirror Node responds with a 400 error. The outdated range triggers the error: Cannot search topics without a valid timestamp range: Timestamp range by the lower and upper bounds must be positive and within 7 days.

- Case 2: If a fromBlock is not found within the getHistoricalBlock() method, the Relay is designed to detect these failures, classify them as accepted errors, and immediately return a nullable value. However, the logs indicate that even after encountering the accepted error, another request was sent to the Mirror Node, which responded with a 400 error. This subsequently caused the end-to-end request to fail with a 500 status. Further investigation is needed. 

`Update for case 2 `: The accepted error originated from the getHistoricalBlock method, but it pertains to toBlock, not fromBlock. While a nullable value is returned immediately if fromBlock is not found, the behavior for toBlock is different—it simply does nothing if null was returned back. Since toBlock is critical for determining the gte field in the Mirror Node request parameter timestamp, the absence of toBlock causes the gte parameter to be omitted. This leads to the issue: Cannot search topics without a valid timestamp range: Timestamp range must have gt (or gte) and lt (or lte), or eq operator.

### Solution

 Investigate the Relay Logic and improve the eth_getLogs method

```[tasklist]
### Tasks
- [x] Investigate all scenarios where eth_getLogs requests fail.
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/3429
- [ ] https://github.com/hashgraph/hedera-json-rpc-relay/issues/3430
```

### LINKED ISSUE #3430: [eth_getLogs Failure] Handle non-existent toBlock in validateBlockRangeAndAddTimestampToParams method
### Desctiption

The eth_getLogs functionality fails when the toBlock specified in the validateBlockRangeAndAddTimestampToParams method does not exist in the network. Currently, the method returns null for a non-existent toBlock, which causes the gte parameter to be omitted in the Mirror Node request. This results in the following error:
“Cannot search topics without a valid timestamp range: Timestamp range must have gt (or gte) and lt (or lte), or eq operator.”

This issue also leads to a 400 response from the Mirror Node, which cascades into a 500 status error for the end-to-end request.

### Solution
 Ensuring eth_getLogs gracefully handles cases where toBlock is not provided. Without toBlock, the `params.lte` field is omitted, leading to request rejected by MN. To address this, the function now returns an empty response to the client.

quiet-node:
**Description**:
This pull request improves the eth_getLogs functionality by:
- Introducing a new error: TIMESTAMP_RANGE_TOO_LARGE for handling large timestamp ranges.
- Adding validation to ensure the timestamp range between fromBlock and toBlock is within acceptable limits, 7 days (604800 seconds). (Fixes #3428)
- Ensuring eth_getLogs gracefully handles cases where toBlock is not provided. Without toBlock, the `params.lte` field is omitted, leading to request rejected by MN. To address this, the function now returns an empty response to the client. (Fixes #3430)

Tests and comments have been added to support and clarify these changes.

**Related issue(s)**:

Fixes #3428

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -135,7 +135,7 @@ export const predefined = {
     }),
   TIMESTAMP_RANGE_TOO_LARGE: (fromBlock: string, fromTimestamp: number, toBlock: string, toTimestamp: number) =>
     new JsonRpcError({
-      code: -32000,
+      code: -32004,
       message: `The provided fromBlock and toBlock contain timestamps that exceed the maximum allowed duration of 7 days (604800 seconds): fromBlock: ${fromBlock} (${fromTimestamp}), toBlock: ${toBlock} (${toTimestamp})`,
     }),
   REQUEST_BEYOND_HEAD_BLOCK: (requested: number, latest: number) =>
--- a/packages/relay/src/lib/errors/JsonRpcError.ts
+++ b/packages/relay/src/lib/errors/JsonRpcError.ts
@@ -133,6 +133,11 @@ export const predefined = {
       code: -32000,
       message: `Exceeded maximum block range: ${blockRange}`,
     }),
+  TIMESTAMP_RANGE_TOO_LARGE: (fromBlock: string, fromTimestamp: number, toBlock: string, toTimestamp: number) =>
+    new JsonRpcError({
+      code: -32000,
+      message: `The provided fromBlock and toBlock contain timestamps that exceed the maximum allowed duration of 7 days (604800 seconds): fromBlock: ${fromBlock} (${fromTimestamp}), toBlock: ${toBlock} (${toTimestamp})`,
+    }),
   REQUEST_BEYOND_HEAD_BLOCK: (requested: number, latest: number) =>
     new JsonRpcError({
       code: -32000,
--- a/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
+++ b/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
@@ -140,9 +140,13 @@ export class CommonService implements ICommonService {
     } else {
       fromBlockNum = parseInt(fromBlockResponse.number);
       const toBlockResponse = await this.getHistoricalBlockResponse(requestDetails, toBlock, true);
-      if (toBlockResponse != null) {
-        params.timestamp.push(`lte:${toBlockResponse.timestamp.to}`);
-        toBlockNum = parseInt(toBlockResponse.number);
+      if (!toBlockResponse) {
+        return false;
+      }
+
+      params.timestamp.push(`lte:${toBlockResponse.timestamp.to}`);
+      toBlockNum = parseInt(toBlockResponse.number);
+
       }
 
       if (fromBlockNum > toBlockNum) {
--- a/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
+++ b/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
@@ -173,7 +173,7 @@ export class CommonService implements ICommonService {
       }
 
       if (fromBlockNum > toBlockNum) {
-        return false;
+        throw predefined.INVALID_BLOCK_RANGE;
       }
 
       const blockRangeLimit = this.getLogsBlockRangeLimit();
--- a/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
+++ b/packages/relay/src/lib/services/ethService/ethCommonService/index.ts
@@ -190,6 +190,53 @@ export class CommonService implements ICommonService {
     return true;
   }
 
+  public async validateBlockRange(fromBlock: string, toBlock: string, requestDetails: RequestDetails) {
+    let fromBlockNumber: any = null;
+    let toBlockNumber: any = null;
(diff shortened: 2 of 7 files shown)


packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -133,6 +133,11 @@ export const predefined = {
       code: -32000,
       message: `Exceeded maximum block range: ${blockRange}`,
     }),
+  TIMESTAMP_RANGE_TOO_LARGE: (fromBlock: string, fromTimestamp: number, toBlock: string, toTimestamp: number) =>
+    new JsonRpcError({
+      code: -32000,
```
Nana-EC, line 138: Q: shouldn't code be unique?
I realize we made this mistake with `RANGE_TOO_LARGE` also as it's the same as `NO_MINING_WORK`

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -133,6 +133,11 @@ export const predefined = {
       code: -32000,
       message: `Exceeded maximum block range: ${blockRange}`,
     }),
+  TIMESTAMP_RANGE_TOO_LARGE: (fromBlock: string, fromTimestamp: number, toBlock: string, toTimestamp: number) =>
```
Nana-EC, line 136: Would `RANGE_TOO_LARGE` not suffice for the situation given the API call utilizes locks not timestamp?

Nana-EC (commented):
Looking good some questions

packages/relay/src/lib/errors/JsonRpcError.ts
quiet-node: Ah yeah good point! Let me tryna find a unique code for it within the appropriate range

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
```diff
@@ -140,9 +140,29 @@ export class CommonService implements ICommonService {
...
+      if (!toBlockResponse) {
+        return false;
+      }
+
+      params.timestamp.push(`lte:${toBlockResponse.timestamp.to}`);
+      toBlockNum = parseInt(toBlockResponse.number);
+
+      // Add timestamp range validation (7 days = 604800 seconds)
+      const timestampDiff = toBlockResponse.timestamp.to - fromBlockResponse.timestamp.from;
+      if (timestampDiff > 604800) {
```
natanasow, line 159: nit: 7 days as seconds may be defined as constant

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
```diff
@@ -140,9 +140,29 @@ export class CommonService implements ICommonService {
...
+        throw predefined.TIMESTAMP_RANGE_TOO_LARGE(
+          `0x${fromBlockNum.toString(16)}`,
```
natanasow, line 161: nit: we can use `prepend0x` here

natanasow (dismissed):
LG, some nits that can be addressed in a follow-up PR due to the urgency of this one.

packages/relay/src/lib/errors/JsonRpcError.ts
quiet-node: Yeah I at first thought of using it to but then realized RANGE_TOO_LARGE does have a generic name, but its message specifically addresses the block range exceeding the limit. In this case, the issue stemmed from the timestamps within the blocks (fromBlock and toBlock). So I thought a different error would be more appropriate. I aimed for a more precise and specific message to ensure clarity for end clients and eliminate any need for guessing

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
quiet-node: Great point! Updated with a constant

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
quiet-node: Ahhh nice catch yikes I knew it didn't look quite right loool thanks for pointing it out! Updated with prepend0x

nadezhdapopovaa (dismissed):

natanasow (dismissed):

konstantinabl (dismissed):

natanasow (approved):

konstantinabl (approved):
LGTM, we just have to discuss if we will limit users to 7 days or we somehow handle this limiti from the MN side

packages/relay/src/lib/errors/JsonRpcError.ts
quiet-node: Updated with -32004. Thanks for pointing it out!

PR #3433, opened 2025-01-28
fix: fixed eth_getCode to now use blockNumber when address is an HTS token

### LINKED ISSUE #3155: `eth_getCode` does not use `blockNumber` when `address` corresponds to an HTS token
### Description

The JSON-RPC method `eth_getCode` does not take into account the `blockNumber` when the `address` corresponds to an HTS token. It always returns the redirect bytecode, specified by [HIP-719](https://hips.hedera.com/hip/hip-719), regardless of the provided `blockNumber`.

This means that is the `blockNumber` provided is **less** than the block number where the HTS token was created, the returned bytecode is incorrect.

### Steps to reproduce

We use [USDC](https://hashscan.io/mainnet/token/0.0.456858) on `mainnet` as an example. It was created at block [`22193323`](https://hashscan.io/mainnet/block/22193323). Running the following command we get the right bytecode

```console
curl https://mainnet.hashio.io/api -X POST --data '{
    "method": "eth_getCode", "id": "1", "jsonrpc": "2.0",
    "params": ["0x000000000000000000000000000000000006f89a", "latest"]
}'
{"result":"6080604052348015600f57600080fd5b506000610167905077618dc65e000000000000000000000000000000000006f89a600052366000602037600080366018016008845af43d806000803e8160008114605857816000f35b816000fdfea2646970667358221220d8378feed472ba49a0005514ef7087017f707b45fb9bf56bb81bb93ff19a238b64736f6c634300080b0033","jsonrpc":"2.0","id":"1"}
```

However, when we specify `22193322 (0x152a4aa)` as the `blockNumber` (the previous block where USDC was created), we still get the same bytecode as before

```console
curl https://mainnet.hashio.io/api -X POST --data '{
    "method": "eth_getCode", "id": "1", "jsonrpc": "2.0",
    "params": ["0x000000000000000000000000000000000006f89a", "0x152a4aa"]
}'
```

However, the response in this case should have been `0x`, empty bytecode.

### Additional context

When querying the Mirror Node, it correctly uses the `timestamp` (determined by the `blockNumber`).

**With block `22193323` (block where USDC Token was created).**

We get the `timestamp` for block `22193323`

```console
$ curl https://mainnet.mirrornode.hedera.com/api/v1/blocks/22193323 | jq .timestamp.to
"1632175205.855270000"
```

and use it to query for the USDC Token at that timestamp

```console
$ curl 'https://mainnet.mirrornode.hedera.com/api/v1/tokens/0.0.456858?timestamp=1632175205.855270000' | jq '.token_id,.name,.symbol'
"0.0.456858"
"USD Coin"
"USDC"
```

**With block `22193322` (just before USDC Token was created).**

```console
$ curl https://mainnet.mirrornode.hedera.com/api/v1/blocks/22193322 | jq .timestamp.to
"1632175203.847228000"
```

```console
$ curl 'https://mainnet.mirrornode.hedera.com/api/v1/tokens/0.0.456858?timestamp=1632175203.847228000'
{"_status":{"messages":[{"message":"Not found"}]}}%
```

### Hedera network

_any_

### Version

`v0.58.0`

### Operating system

macOS

simzzz:
**Description**:
If we pass an HTS token address to getCode, we do not take into account the block number, resulting in a bug where it would return the smart contract code even if the block number provided is earlier than the one where the token was created.

**Related issue(s)**:

Fixes #3155 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/services/ethService/ethCommonService/index.ts
```diff
@@ -389,4 +389,9 @@ export class CommonService implements ICommonService {
 
     return this.getLogsWithParams(address, params, requestDetails);
   }
+
+  public addressToTokenId(address: string): string {
+    const tokenNum = parseInt(address.slice(-8), 16);
+    return `0.0.${tokenNum}`;
+  }
```
quiet-node, line 396: Feels like this should be in formatter or utils instead of commonService. Also we need unit tests for this method

packages/relay/src/lib/services/ethService/ethCommonService/index.ts
```diff
@@ -389,4 +389,9 @@ export class CommonService implements ICommonService {
 
     return this.getLogsWithParams(address, params, requestDetails);
   }
+
+  public addressToTokenId(address: string): string {
+    const tokenNum = parseInt(address.slice(-8), 16);
```
quiet-node, line 394: What if this address is not long zero encoded alias? .slice(-8) doesn't seem to scale. Maybe look for an approach with the hashgraph/sdk package.

packages/relay/tests/lib/eth/eth_getCode.spec.ts
```diff
@@ -155,5 +155,69 @@ describe('@ethGetCode using MirrorNode', async function () {
...
+        await ethImpl.getCode(HTS_TOKEN_ADDRESS, futureBlockNumber, requestDetails);
+        expect(true).to.eq(false);
+      } catch (error: any) {
+        const expectedError = predefined.UNKNOWN_BLOCK(`Block number ${futureBlockNumber} does not exist`);
+        expect(error).to.exist;
+        expect(error instanceof JsonRpcError);
+        expect(error.code).to.eq(expectedError.code);
+        expect(error.message).to.eq(expectedError.message);
+      }
+    });
```
quiet-node, line 221: Let's try await expect().to.be.rejectedWith() instead of regular trycatch

quiet-node, line 221: Great test but let's also add acceptance test for beetter e2e coverage.

quiet-node (changes requested):

packages/server/tests/acceptance/eth_getCode.spec.ts
```diff
@@ -0,0 +1,71 @@
+/*-
```
quiet-node, line 1: Ah acceptance test for server package should be in the rpc_batch1/2/3.ts. `eth_getCode` should be in batch 2. Let's move the test cases to batch 2 instead of a new file

packages/relay/src/lib/eth.ts
```diff
@@ -1211,6 +1211,25 @@ export class EthImpl implements Eth {
       ]);
       if (result) {
         if (result?.type === constants.TYPE_TOKEN) {
+          if (blockNumber && !this.common.blockTagIsLatestOrPending(blockNumber)) {
```
acuarica, line 1214: what happens if `blockNumber` is `earliest`? https://ethereum.org/en/developers/docs/apis/json-rpc/#default-block

packages/relay/src/lib/eth.ts
```diff
@@ -1211,6 +1211,25 @@ export class EthImpl implements Eth {
...
+            const blockNumberInt = parseInt(blockNumber, 16);
+            const blockInfo = await this.mirrorNodeClient.getBlock(blockNumberInt, requestDetails);
+
+            if (!blockInfo) {
+              throw predefined.UNKNOWN_BLOCK(`Block number ${blockNumber} does not exist`);
+            }
+
+            const tokenId = Utils.addressToTokenId(address);
```
acuarica, line 1222: This assumes `address` is a long zero address, right?

packages/relay/src/lib/eth.ts
```diff
@@ -1211,6 +1211,25 @@ export class EthImpl implements Eth {
...
+            const tokenInfo = await this.mirrorNodeClient.getTokenById(tokenId, requestDetails);
```
acuarica, line 1223: can we use the `timestamp` argument in the `tokens/{tokenId}` endpoint? this way we avoid having the logic below to compare timestamps.

packages/relay/src/lib/eth.ts
simzzz: Correct

packages/relay/src/lib/eth.ts
simzzz: It was passing through the if, but the parseInt was returning "234" which is why it slipped through my testing. Good catch!

packages/relay/src/lib/eth.ts
```diff
@@ -1209,8 +1209,33 @@ export class EthImpl implements Eth {
         constants.TYPE_CONTRACT,
```
acuarica, line 1209: This query

```ts
      const result = await this.mirrorNodeClient.resolveEntityType(address, EthImpl.ethGetCode, requestDetails, [
        constants.TYPE_CONTRACT,
        constants.TYPE_TOKEN,
      ]);
```

does not depend on the `blockNumber`. Does that mean that this issue happen with contracts as well? If that's the case, we should fixed it here. And the logic you included to fetch block+timestamp can be reused for both cases, contract and tokens.

packages/relay/src/lib/eth.ts
```diff
@@ -1211,11 +1211,23 @@ export class EthImpl implements Eth {
       ]);
       if (result) {
         if (result?.type === constants.TYPE_TOKEN) {
+          const blockInfo = await this.getBlockInfo(blockNumber, requestDetails);
+          if (blockInfo) {
+            if (parseFloat(result.entity?.created_timestamp) > parseFloat(blockInfo.timestamp.to)) {
```
acuarica, line 1216: what would happen if `blockNumber` is the block number where the token was created? shouldn't we use `timestamp.from` instead?

packages/relay/src/lib/eth.ts
```diff
@@ -1211,11 +1211,23 @@ export class EthImpl implements Eth {
...
+              return EthImpl.emptyHex;
+            }
+          }
           if (this.logger.isLevelEnabled('trace')) {
             this.logger.trace(`${requestIdPrefix} Token redirect case, return redirectBytecode`);
           }
           return EthImpl.redirectBytecodeAddressReplace(address);
         } else if (result?.type === constants.TYPE_CONTRACT) {
+          const blockInfo = await this.getBlockInfo(blockNumber, requestDetails);
```
acuarica, line 1225: we could move this bit right after the `if (result) {` given it's the same for both cases.

packages/relay/src/lib/eth.ts
acuarica, line 1216: Mmm, I'm not convinced about the timestamp comparison. I guess the issue is with `resolveEntityType`, which should get the `blockNumber`/`timestamp`. I'm not sure if by doing the comparison here we are missing any edge case. If it gets too complicated, then we can leave like this and think of a refactor in the future. WDYT?

simzzz: In that case it would return the bytecode

This condition returns `emptyHex` if we pass a blockNumber when the token/contract wasn't created, so if the creation timestamp is after `timestamp.to` then the block will be earlier. If we use `timestamp.from` it would return the empty bytecode instead

simzzz: Generally I agree and it also makes the method more confusing, however I did try just passing the timestamp to `resolveEntityType`, but at the end of the try block there is this fallback logic which gets the contract bytecode from the SDK as fallback in case the mirror node returns empty bytecode or 404. 

```
      const bytecode = await this.hapiService
        .getSDKClient()
        .getContractByteCode(0, 0, address, EthImpl.ethGetCode, requestDetails);
      return prepend0x(Buffer.from(bytecode).toString('hex'));
```

However if I pass the timestamp there might be a case where the fallback logic gets triggered for an earlier block and this would be harder to check.  I would suggest to leave it like this and possibly refactor/simplify the whole method

acuarica: Saw it backwards, thanks for the clarification!

acuarica: Sounds good to me.

acuarica (dismissed):

acuarica (dismissed):

packages/relay/src/lib/eth.ts
```diff
@@ -2864,4 +2868,16 @@ export class EthImpl implements Eth {
...
+    if (blockNumber && !this.common.blockTagIsLatestOrPending(blockNumber)) {
+      const blockNumberInt = blockNumber === EthImpl.blockEarliest ? 0 : parseInt(blockNumber, 16);
+      const blockInfo = await this.mirrorNodeClient.getBlock(blockNumberInt, requestDetails);
+      if (!blockInfo) {
+        throw predefined.UNKNOWN_BLOCK(`Block number ${blockNumber} does not exist`);
+      }
+      return blockInfo;
+    }
+    return null;
+  }
```
quiet-node, line 2882: I believe CommonService.getHistoricalBlockResponse() can be used here for the same result but with broader precheck coverage. Let’s ensure CommonService.getHistoricalBlockResponse() can do the job, or can be modified to do the job, to avoid introducing additional methods.

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -790,6 +790,24 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
...
+
+    it('should return empty bytecode for contract when a block earlier than the contract creation is passed', async function () {
+      const res = await relay.call(
+        RelayCalls.ETH_ENDPOINTS.ETH_GET_CODE,
+        [mainContractAddress, '0x123'], // a very early block number
+        requestId,
+      );
+      expect(res).to.equal(EthImpl.emptyHex);
+    });
+
```
quiet-node, line 810: Nice test. However, `0x123` is quite arbitrary and might potentially be flaky. Let's instead get the block number at birth for these token and contract, and then use that to get an earlier block number which can be used as the param here.

quiet-node: Secondly, these tests are great candidates for the release test suite. Please kindly add the `@release` tag to the test so they can be picked up during release test.

quiet-node (changes requested):
Great work as always! Just some questions and suggestions.

packages/relay/src/lib/eth.ts
```diff
@@ -1210,6 +1210,10 @@ export class EthImpl implements Eth {
         constants.TYPE_TOKEN,
       ]);
       if (result) {
+        const blockInfo = await this.getBlockInfo(blockNumber, requestDetails);
```
acuarica, line 1213: another thing came to mind, what if the token was/is deleted, but in the `blockNumber` requested the token still exists? In this case, the result of `resolveEntityType` would be falsy, so there would be no block comparison.

If this is an issue, I don't mind tackling this in another PR.

simzzz: Hmm, I just tested with some tokens that were later deleted that I found on testnet (for example - https://hashscan.io/testnet/token/0.0.4933576) and I passed the creation block and it returned the bytecode. The result was not falsy,  it just had `"deleted": true,`

acuarica: ok, so that means if the now the `blockNumber` is after the token was deleted, we still be getting the bytecode when we shoudn't, right?

simzzz: Yes, that is correct.

packages/relay/src/lib/eth.ts
```diff
@@ -2864,4 +2868,15 @@ export class EthImpl implements Eth {
...
+  private async getBlockInfo(blockNumber: string | null, requestDetails: RequestDetails): Promise<any> {
+    if (blockNumber && !this.common.blockTagIsLatestOrPending(blockNumber)) {
+      const blockInfo = await this.common.getHistoricalBlockResponse(requestDetails, blockNumber, true);
+      if (!blockInfo) {
+        throw predefined.UNKNOWN_BLOCK(`Block number ${blockNumber} does not exist`);
+      }
+      return blockInfo;
+    }
+    return null;
+  }
```
quiet-node, line 2881: Looks like I didn’t get an answer for this [comment](https://github.com/hashgraph/hedera-json-rpc-relay/pull/3433#discussion_r1946776059), but it got resolved. However, getBlockInfo is still present.

simzzz: I misunderstood your comment, it should be resolved now, mb

quiet-node (approved):
LGTM! Nice work!

acuarica (approved):

packages/relay/src/lib/eth.ts
simzzz: Leaving this unsolved for now since getting token by ID would always return `deleted: true` even for an older date

PR #3435, opened 2025-01-30
feat: refactor mirrorNodeClient axios `transformResponse` and fix all unit tests

### LINKED ISSUE #3367: Transaction with CONTRACT_NEGATIVE_VALUE prevents entire block from being retrieved

natanasow:
**Description**:

Follow up on https://github.com/hashgraph/hedera-json-rpc-relay/pull/3387 regarding  this discussion https://github.com/hashgraph/hedera-json-rpc-relay/pull/3387#discussion_r1934738767.

**Related issue(s)**:

Fixes #3367

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -25,16 +25,16 @@
 # NOTE: Must be placed last to ensure enforcement over all other rules
 
 # Protection Rules for Github Configuration Files and Actions Workflows
-/.github/                                       @hashgraph/devops-ci @hashgraph/devops-ci-committers @hashgraph/release-engineering @hashgraph/release-engineering-managers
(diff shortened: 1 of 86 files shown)


quiet-node (changes requested):
Let's wait for the other PR to get merged first so we can cut a hotfix then we can target this against main after that as this has lower priority.

quiet-node (approved):
Thanks for the great work!

konstantinabl (approved):

PR #3469, opened 2025-02-11
feat: adds proper structure to automatic bumping PRs

### LINKED ISSUE #3319: [SMALL]Automatic bumping PRs are not properly structured
### Problem

The `hedera-json-rpc-relay` utilizes automatic Snyk PRs to update dependencies alongside Dependabot. However, Snyk PRs are not well-structured, requiring significant manual effort to prepare them for review.

- Snyk PRs lack proper conventional type prefixes in their titles (e.g., `build:` or `build(dep):`).  
- Snyk PRs are improperly signed, causing the DCO job to fail consistently.
- Snyk PRs don't have approprate labels (e.g., dependencies)

### Solution

Approve Snyk PRs to resolve the issues

### Alternatives

_No response_

simzzz:
**Description**:
Snyk PR titles now automatically add proper conventional type prefixes and appropriate labels. 

**Related issue(s)**:

Fixes #3319 

**Notes for reviewer**:
Judging from the closed PR's by Snyk, I have made the assumption that they would always start with [Snyk] in their title and that they would all be a version upgrade or a security upgrade. 

**Checklist**

### DIFF
--- a/.github/scripts/check-pr.js
+++ b/.github/scripts/check-pr.js
@@ -6,274 +6,304 @@ const { GITHUB_REPOSITORY, GITHUB_PR_NUMBER } = process.env;
 const [owner, repo] = GITHUB_REPOSITORY.split('/');
 
 async function getPRDetails(prNumber) {
-    const url = `https://api.github.com/repos/${owner}/${repo}/pulls/${prNumber}`;
-    try {
-        const response = await axios.get(url, {
-            headers: {
-                Authorization: `token ${githubToken}`,
-            },
-        });
-        return response.data;
-    } catch (error) {
-        if (error.response && error.response.status === 404) {
-            console.log(`PR #${prNumber} not found in repository ${owner}/${repo}, skipping...`);
-            return null;
-        } else {
-            throw error;
-        }
+  const url = `https://api.github.com/repos/${owner}/${repo}/pulls/${prNumber}`;
+  try {
+    const response = await axios.get(url, {
+      headers: {
+        Authorization: `token ${githubToken}`,
+      },
+    });
+    return response.data;
+  } catch (error) {
+    if (error.response && error.response.status === 404) {
+      console.log(`PR #${prNumber} not found in repository ${owner}/${repo}, skipping...`);
+      return null;
+    } else {
+      throw error;
     }
+  }
 }
 
 async function getIssueDetails(issueOwner, issueRepo, issueNumber) {
-    try {
-        const url = `https://api.github.com/repos/${issueOwner}/${issueRepo}/issues/${issueNumber}`;
-        const response = await axios.get(url, {
-            headers: {
-                Authorization: `token ${githubToken}`,
-            },
-        });
-        return response.data;
-    } catch (error) {
-        if (error.response && error.response.status === 404) {
-            console.log(`Issue #${issueNumber} not found in repository ${issueOwner}/${issueRepo}, skipping...`);
-            return null;
-        } else {
-            throw error;
-        }
+  try {
+    const url = `https://api.github.com/repos/${issueOwner}/${issueRepo}/issues/${issueNumber}`;
+    const response = await axios.get(url, {
+      headers: {
+        Authorization: `token ${githubToken}`,
+      },
+    });
+    return response.data;
+  } catch (error) {
+    if (error.response && error.response.status === 404) {
+      console.log(`Issue #${issueNumber} not found in repository ${issueOwner}/${issueRepo}, skipping...`);
+      return null;
+    } else {
+      throw error;
     }
+  }
 }
 
 async function getContributors() {
-    const url = `https://api.github.com/repos/${owner}/${repo}/contributors`;
-    const response = await axios.get(url, {
-        headers: {
-            Authorization: `token ${githubToken}`,
-        },
-    });
-    return response.data;
+  const url = `https://api.github.com/repos/${owner}/${repo}/contributors`;
+  const response = await axios.get(url, {
+    headers: {
+      Authorization: `token ${githubToken}`,
+    },
+  });
+  return response.data;
 }
 
 function stripHTMLTags(text) {
-    return text.replace(/<\/?[^>]+(>|$)/g, '');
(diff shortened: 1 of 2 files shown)


.github/scripts/check-pr.js
```diff
@@ -6,274 +6,304 @@ const { GITHUB_REPOSITORY, GITHUB_PR_NUMBER } = process.env;
...
+    await axios.post(
+      url,
+      { labels: ['dependencies'] },
+      {
+        headers: {
+          Authorization: `token ${githubToken}`,
+        },
+      },
+    );
+  }
```
quiet-node, line 276: I know the file needs to get reformated but it made it very hard to detect the changes. Could you please confirm that, except for the linting, this is the only block got added, and of course the `await fixSnykPR(pr);` in run(), everything else stays the same?

.github/scripts/check-pr.js
```diff
@@ -6,274 +6,304 @@ const { GITHUB_REPOSITORY, GITHUB_PR_NUMBER } = process.env;
...
+    await axios.patch(
+      url,
+      { title },
+      {
+        headers: {
+          Authorization: `token ${githubToken}`,
+        },
+      },
+    );
+  }
```
quiet-node, line 260: Ha this's a clever way to update the PRs.

I might be mistaken, but this isn’t quite what I initially envisioned. I thought Snyk is like a tool integrated into the repo, allowing us to reconfigure the Snyk PRs through their API rather than having to make patch requests to GitHub like this.

Have you confirmed that we can’t reconfigure Snyk through its API? Or maybe we should check with the CI team to ensure this is the only option? 

Sorry, this approach feels a bit like a hack to me and I’d expect there to be a more streamlined way to configure the Snyk tool.

quiet-node, line 260: Ha this's a clever way to update the PRs.

I might be mistaken, but this isn’t quite what I initially envisioned. I thought Snyk is like a tool integrated into the repo, allowing us to reconfigure the Snyk PRs through their API rather than having to make patch requests to GitHub like this.

Have you confirmed that we can’t reconfigure Snyk through its API? Or maybe we should check with the CI team to ensure this is the only option? 

Sorry, this approach feels a bit like a hack to me and I’d expect there to be a more streamlined way to configure the Snyk tool.

quiet-node (commented):
Thanks for getting to this! This definitely helps save time by reducing the need to rebase the PR just to sign off and rename them.

got have a few questions tho

.github/scripts/check-pr.js
simzzz: Yes, it got formatted automatically after comitting, but these are the only changes

.github/scripts/check-pr.js
simzzz: I took a look initially but it seems very limited. It doesn't accept labels and we can pass only static strings with limited variables.

It can work for the title but we won't be able to detect if it's an Upgrade or a Security upgrade which Snyk currently does.
For the labels - we would still need to keep the hack or do them manually. 

https://docs.snyk.io/scan-with-snyk/pull-requests/snyk-pull-or-merge-requests/customize-pr-templates/apply-a-custom-pr-template 

https://docs.snyk.io/scan-with-snyk/pull-requests/snyk-pull-or-merge-requests/customize-pr-templates/variables-list-and-description

.github/scripts/check-pr.js
quiet-node: Nice thanks for confirming!

.github/scripts/check-pr.js
quiet-node: Ah, I see.

On a different note, how about we remove Snyk entirely and rely only on dependabot? I haven’t seen Snyk PRs in Mirror Node or Mirror Node Explorer anymore but only by dependa bots. 

Would it be worth removing it altogether?
@beeradb @andrewb1269hg @rbarker-dev, what do you think?

quiet-node (approved):
LGTM great work

.github/scripts/check-pr.js
quiet-node: @simzzz Andrew from the Platform CI team asked if you have tested this to confirm that the token is neither echoed nor visible in the query.

```typescript
const response = await axios.get(url, {
  headers: {
    Authorization: `token ${githubToken}`,
  },
});
```

simzzz: It doesn't, the same call already existed for some of the functions.

.github/snyk_pull_request_template.yaml
```diff
@@ -0,0 +1,4 @@
+commitMessage: |
```
andrewb1269hg, line 1: Where is this `commitMessage` template being called from the scripts? I don't see it anywhere.

simzzz: According to the documentation, Snyk will automatically use this template for all future pull requests and override the default commit message with the one from the template

https://docs.snyk.io/scan-with-snyk/pull-requests/snyk-pull-or-merge-requests/customize-pr-templates/apply-a-custom-pr-template#customize-using-a-yaml-pr-template-file

.github/snyk_pull_request_template.yaml
```diff
@@ -0,0 +1,4 @@
...
+  Update dependencies for {{ snyk_project_name }}
+
+  Signed-off-by: Logan Nguyen <logan.nguyen@swirldslabs.com>
```
quiet-node, line 4: Ha, I missed this! I believe we can sign with Snyk’s signature instead of mine for better clarity. I usually sign because Snyk PRs don’t get signed.
```suggestion
  Signed-off-by: Swirlds Automation <swirlds-eng-automation@swirlds.com>
```
Check this commit for the signature example: https://github.com/hashgraph/hedera-json-rpc-relay/pull/3473/commits/f6d08111169be67b95c10c2514c83bf0358ee82a

quiet-node (changes requested):

konstantinabl (approved):

quiet-node (approved):

andrewb1269hg (approved):

Issue #3283, opened 2024-11-21
Transactions are getting failed with INSUFFICIENT_TX_FEE


anshuman-supraoracles:
### Description

We are feeding pair details into a smart contract on hedera. All the txns we are pushing is having the calldata size <5kb. Most of the times the txns are getting executed perfectly but only certain times its getting failed due to **INSUFFICIENT_TX_FEE** error. Although both the accounts(relayer and txn initiator) are having sufficient balances.

We are using the dedicated relayer from **Validation Cloud** 

These are the txns that got faied

https://hashscan.io/mainnet/transaction/1732151941.129055000
https://hashscan.io/mainnet/transaction/1732152145.106706521
https://hashscan.io/mainnet/transaction/1732152160.921207000
https://hashscan.io/mainnet/transaction/1732152177.426895000

You can check other successful txns in this contract

https://hashscan.io/mainnet/contract/0.0.4323024?pf=1&p=1&k=1732163002.082358000

### Steps to reproduce

NA

### Hedera network

mainnet

### Version

Validation Cloud

### Operating system

Linux

simzzz:
I'm picking this up. You can assign me

simzzz:
Hello, @anshuman-supraoracles. Would it be possible to provide the address of the logic contract that you passed in your contract's constructor so I can conduct further debugging? Thank you!

anshuman-supraoracles:
@simzzz Logic contract 0xdee317b28dcce292859bd82f517de2037cc46b98
We are using the relayer from Validation cloud

simzzz:
It seems the issue might be related to a configuration variable called `GAS_PRICE_PERCENTAGE_BUFFER`. This variable adds a buffer to the estimated gas price to ensure transactions have sufficient fees to execute successfully.

Since you are using the Validation Cloud, it’s possible they might have an issue and need to increase the buffer, which could be causing the `INSUFFICIENT_TX_FEE` error. In our own implementation, this variable is set to 0, and we have not been able to reproduce the issue on our end. We will coordinate with the Validation Cloud team to address this configuration issue and work toward a resolution and keep you updated.

Nana-EC:
Closing based on above progress in terms of observation and support via the config value.
Please fee free to reopen or mention @Ferparishuertas or @simzzz if this comes up again

anshuman-supraoracles:
[...]

PR #3482, opened 2025-02-17
fix: reduced TTL for feeHistory, getTinyBarGasFee, and getPrice

simzzz:
**Description**:
This PR reduces the TTL for caching fee history, gas fee, and gas price in an attempt to reduce the likelihood of INSUFFICIENT_TX_FEE errors.

We suspect that a reason that some transactions fail is because there is a fluctuation on the network gas price, while we return the cached value resulting in a too low gas estimation. By reducing the TTL in a few places where the gas fee is used it would reduce the likelihood of such problem.

**Related issue(s)**:
#3283 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -356,13 +357,19 @@ export class SDKClient {
...
         const exchangeRates = await this.getExchangeRate(callerName, requestDetails);
         const tinyBars = this.convertGasPriceToTinyBars(schedule.fees[0].servicedata, exchangeRates);
 
-        await this.cacheService.set(constants.CACHE_KEY.GET_TINYBAR_GAS_FEE, tinyBars, callerName, requestDetails);
+        await this.cacheService.set(
+          constants.CACHE_KEY.GET_TINYBAR_GAS_FEE,
+          tinyBars,
+          callerName,
+          requestDetails,
+          parseInt(constants.ETH_GAS_FEE_TTL),
```
quiet-node, line 371: Why don't we use CACHE_TTL.HALF_HOUR constant here but introduce new constants?

packages/relay/src/lib/eth.ts
```diff
@@ -379,7 +379,13 @@ export class EthImpl implements Eth {
           );
         }
         if (newestBlock != EthImpl.blockLatest && newestBlock != EthImpl.blockPending) {
-          await this.cacheService.set(cacheKey, feeHistory, EthImpl.ethFeeHistory, requestDetails);
+          await this.cacheService.set(
+            cacheKey,
+            feeHistory,
+            EthImpl.ethFeeHistory,
+            requestDetails,
+            parseInt(constants.ETH_FEE_HISTORY_TTL),
```
quiet-node, line 387: Looks like constants.ETH_FEE_HISTORY_TTL is set to CACHE_TTL.ONE_HOUR, but the default ttl value is already one hour, isn't it? Why do we want to add the same value here?

quiet-node:
@simzzz, the PR description doesn’t explain how lowering the TTL would help resolve the issue. Additionally, there are no end-to-end tests to verify the fix. Could you update the description with more details on how this change addresses the problem? If possible, adding e2e tests would be beneficial to prevent regressions. Thanks!

packages/relay/src/lib/eth.ts
simzzz: We don't, I must have set the wrong enum value in the constants file. Good catch!

quiet-node (approved):

natanasow (approved):

PR #3495, opened 2025-02-18
fix: fixes flaky CI tests in RPC Batch 2

### LINKED ISSUE #3353: Intermittent Test Failures in Acceptance Suite Batches
### Problem

Two tests in the acceptance suite are exhibiting significant flakiness:  

- **Test:** `'should execute "eth_sendRawTransaction" if receiver's account has receiver_sig_required disabled'`  
- **Issue:** The `beforeAll()` block in the `eth_getCode` suite is timing out.  

### Solution

- Investigate and resolve the issue

### Alternatives

_No response_

simzzz:
**Description**:
The beforeAll() block in the eth_getCode suite is timing out.

By optimizing the hook in the RPC API batch 2 and removing unnecessary and expensive calls, and then rerunning the job multiple times - now it fails much less often.

**Related issue(s)**:

Fixes partially #3353 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/server/tests/acceptance/rpc_batch2.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch2.spec.ts
@@ -57,7 +57,6 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
     servicesNode,
     mirrorNode,
     relay,
-    logger,
     initialBalance,
   }: {
     servicesNode: ServicesClient;
@@ -72,10 +71,7 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
   let requestId;
   let htsAddress;
   let expectedGasPrice: string;
-  let basicContract: ethers.Contract;
-  let basicContractAddress: string;
   let parentContractAddress: string;
-  let parentContractLongZeroAddress: string;
   let createChildTx: ethers.ContractTransactionResponse;
   let accounts0StartBalance: bigint;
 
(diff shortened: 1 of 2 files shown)


quiet-node (dismissed):
Nice attempt! Would it be possible to also take this opportunity to fix the `should execute "eth_sendRawTransaction" if the receiver's account has receiver_sig_required disabled'` test? I’ve noticed it fails quite often. If you think it’s better addressed in a separate PR, that works too.

simzzz:
@quiet-node Yes, I did debug that one as well but I still haven't determined an exact cause. I will get back to it in a separate PR

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -110,73 +106,79 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
...
+      accounts[0].wallet.provider?.getBlockNumber(),
+      accounts[0].wallet.provider?.getBalance(accounts[0].address),
+    ]);
 
-    // @ts-ignore
-    createChildTx = await parentContract.createChild(1);
+    createChildTx = childTx;
     await relay.pollForValidTransactionReceipt(createChildTx.hash);
 
-    if (global.logger.isLevelEnabled('trace')) {
```
konstantinabl, line 153: Why delete this?

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -110,73 +106,79 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
...
+    accounts0StartBalance = balance as bigint;
 
-    tokenId = await servicesNode.createToken(1000, requestId);
     htsAddress = Utils.idToEvmAddress(tokenId.toString());
+    await Promise.all([
+      accounts[0].client.associateToken(tokenId, requestId),
+      accounts[1].client.associateToken(tokenId, requestId),
+    ]);
 
-    logger.info('Associate and transfer tokens');
```
konstantinabl, line 162: Also this - Do you think its unecessary?

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -110,73 +106,79 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
...
-    const neededAccounts: number = 4;
-    accounts.push(
-      ...(await Utils.createMultipleAliasAccounts(
-        mirrorNode,
-        initialAccount,
-        neededAccounts,
-        initialBalance,
-        requestId,
-      )),
+    const [gasPrice, initialAccount] = await Promise.all([
```
konstantinabl, line 110: Initial account is not needed here imo

konstantinabl (changes requested):
a few comments

packages/server/tests/acceptance/rpc_batch2.spec.ts
simzzz: I felt like they were unnecessary - if something fails we usually look at the error. This log in particular adds no value. The ones that log the hashes/addresses still aren't very helpful as it is tricky to access the local node instance that the CI is using. Also the log files for this batch is already long and detailed. If you disagree I don't mind bringing them back.

konstantinabl: Okay, seems fair : )

konstantinabl (approved):

quiet-node (approved):
LGTM thanks for looking into this.

quiet-node:
@simzzz woops Batch2 still fails prett consistently

PR #3497, opened 2025-02-19
feat: filter out some methods from batch requests

### LINKED ISSUE #1885: Batch Requests - Method not allowed for batch request
### Problem

Some methods should not be allowed to be requested as part of a batch.

Debug and Filter APIs to start with, but we might want to leave them in a configurable property that we can extend if needed.

### Solution

- New error  
  code: -32007 
  message: Method `<method>` is not permitted as part of batch requests

- New Configurable property with the list of the methods that are not allowed for batch.

- Add a validation on each request, right before calling `getRequestResult`
Current:
```
const promises = body.map((item: any) => this.getRequestResult(item, ctx.ip));
```

Proposed:
```
    const promises = body.map((item: any) => {
      //validate method is allowed for batch requests
      if(this.methodIsAllowed(item.method)) {
        this.getRequestResult(item, ctx.ip)
      } else {
        return jsonResp(item.id, new MethodNotAllowedForBatch(item.method), undefined);
      }
    });
```

### Alternatives

_No response_

natanasow:
**Description**:

Some methods should not be allowed to be requested as part of a batch.

Debug and Filter APIs to start with, but we might want to leave them in a configurable property that we can extend if needed.

**Solution**:

- New error  
  code: -32007 
  message: Method `<method>` is not permitted as part of batch requests

- New Configurable property with the list of the methods that are not allowed for batch.

- Add a validation on each request, right before calling `getRequestResult`

**Related issue(s)**:

Fixes #1885 

**Notes for reviewer**:

**Checklist**


packages/server/tests/acceptance/rpc_batch3.spec.ts
```diff
@@ -2036,6 +2036,29 @@ describe('@api-batch-3 RPC Server Acceptance Tests', function () {
...
+      for (let index = 0; index < disallowedMethods.length; index++) {
+        expect(res[index]).to.haveOwnProperty('error');
+        expect(res[index].id).to.equal(index);
+        expect(res[index].error.code).to.equal(-32007);
+        expect(res[index].error.message).to.equal(
+          `Method ${disallowedMethods[index]} is not permitted as part of batch requests`,
+        );
+      }
+    });
+
```
quiet-node, line 2061: Nice test could you please also add a test to the WS package and also add these into the release test by adding a @release tag to the tests?

quiet-node (changes requested):
Nice work!

packages/server/tests/acceptance/rpc_batch3.spec.ts
natanasow: Added both test and `@release` tag.

Also, I had to add an additional check to the websocket server itself because I had missed that.

quiet-node (approved):
Perfect! LGTM! Thanks for the great work!

konstantinabl (approved):
LGTM

Ferparishuertas (approved):
LGTM

PR #3510, opened 2025-02-25
feat: adds logic for `engine_*` rpc methods which indicates they're not supported 

### LINKED ISSUE #3492: Implement standard response for all engine_* endpoints

simzzz:
This PR adds handling for the JSON RPC engine_ methods by returning UNSUPPORTED_METHOD using regex pattern matching

**Related issue(s)**:

Closes #3492 

**Notes for reviewer**:

**Checklist**


packages/server/tests/integration/server.spec.ts
```diff
@@ -485,6 +485,81 @@ describe('RPC Server', function () {
     expect(res.data.result).to.be.equal('0x0');
   });
 
+  it('should execute "engine_getPayloadV1" and return UNSUPPORTED_METHOD', async function () {
```
natanasow, line 488: We can remove duplications by creating a list of methods, iterating them, and creating an `it` statement for each, similar to [here](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/tests/lib/mirrorNodeClient.spec.ts#L60).

natanasow (changes requested):

packages/server/src/server.ts
```diff
@@ -793,6 +793,10 @@ app.useRpc('eth_coinbase', async () => {
   return logAndHandleResponse('eth_coinbase', [], (requestDetails) => relay.eth().coinbase(requestDetails));
 });
 
+app.useRpcRegex(/^engine_.*$/, async () => {
```
acuarica, line 796: If the method name starts with `engine_` but is not part of the Ethereum JSON-RPC spec https://ethereum.github.io/execution-apis/api-documentation/, for example, `engine_foo`, do we want to return method not found or unsupported method?

packages/relay/src/index.ts
```diff
@@ -45,6 +47,11 @@ export interface Net {
   version(): string;
 }
 
+export interface Engine {
+  // This interface is intentionally empty as all engine_* methods are unsupported
+  // and will return UNSUPPORTED_METHOD error
+}
```
acuarica, line 53: IMO we should remove this (and all its usages) given this is not even called.

acuarica (changes requested):
Thanks for sending this, couple of comments. Also, should we rename the PR? because the current name actually suggests that we are going to implement those endpoints.

packages/server/src/server.ts
simzzz: This was discussed and the reason for this was to avoid having not found if a future engine_ method is added. But if you believe this is incorrect behavior I can change it.

acuarica: Sounds about right, thanks for the clarification.

natanasow (approved):

packages/server/src/koaJsonRpc/lib/methodConfiguration.ts
```diff
@@ -166,4 +166,16 @@ export const methodConfiguration: IMethodRateLimitConfiguration = {
...
+  },
+  engine_newPayloadV1: {
+    total: tier2rateLimit,
+  },
+  engine_forkchoiceUpdatedV1: {
+    total: tier2rateLimit,
+  },
+  engine_exchangeCapabilities: {
+    total: tier2rateLimit,
+  },
```
quiet-node, line 180: QQ: Why do we explicitly specify these methods? Isn't this here https://github.com/hashgraph/hedera-json-rpc-relay/pull/3510/files#diff-b610a40338d7e27cd291a9b987285e717764ac37bbb9d8d476618d5284eaf987R98 always gives engine_* a total of defaultRateLimit?

packages/server/tests/integration/server.spec.ts
```diff
@@ -485,6 +485,28 @@ describe('RPC Server', function () {
...
+          params: [null],
+        });
+
+        Assertions.expectedError();
+      } catch (error: any) {
+        BaseTest.unsupportedJsonRpcMethodChecks(error.response);
+      }
+    });
+  });
+
```
quiet-node, line 509: Nice tests! Should we group this up in a dedicated describe like describe("Unsupported engine methods") or something like that?

quiet-node: Also is it possible to add some tests in the acceptance tests for more e2e coverage?

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -87,6 +87,17 @@ export default class KoaJsonRpc {
...
+   * Register a regex pattern for RPC method names
+   * @param pattern - Regex pattern to match method names
+   * @param func - Function to handle the request
+   */
+  useRpcRegex(pattern: RegExp, func: (params?: any) => Promise<any>): void {
+    this.registry[pattern.toString()] = func;
+    // Use default rate limit for regex patterns
+    this.registryTotal[pattern.toString()] = this.defaultRateLimit;
+  }
+
```
quiet-node, line 100: Nice move! Can we also allow this for the WS server as well?

quiet-node (changes requested):
Nice work left some Qs

packages/server/src/koaJsonRpc/index.ts
quiet-node: Or maybe not—WS is more request-based, so if there’s demand for it, we can implement it. Let's see what the team say.

packages/server/src/koaJsonRpc/lib/methodConfiguration.ts
simzzz: I noticed we add tier 2 rate limit for most unsupported methods here. I'm not sure why though, we can remove this and change the value in `useRpcRegex` if needed.

packages/server/tests/integration/server.spec.ts
simzzz: Yeah, I'm on it

packages/server/src/koaJsonRpc/lib/methodConfiguration.ts
quiet-node: Ah, I see—that’s strange! I believe other unsupported requests will still be reached, but engine_* calls will already be blocked by useRpcRegex through the /engine_*/ key, right? That’s why I think these explicit values aren’t necessary, but correct me if I’m wrong.

acuarica (approved):

packages/server/src/koaJsonRpc/lib/methodConfiguration.ts
simzzz: Yeah, that's correct. They have been removed

quiet-node (approved):
LGTM! Great work!

Issue #3002, opened 2024-09-17
Set the `maxTransactionFee` to a configurable value instead of `constants.MAX_GAS_PER_SEC`


victor-yanev:
### Problem:

Currently we are setting the `maxTransactionFee` to the hbar equivalent of 15 million gas, which is the maximum gas per second that the network can handle, and this is the HAPI fee which is paid by the relay operator account, but we don't want to freely let the operator spend this huge amount, we should change this value to something more sensible and also make it configurable.

### Solution:

- Add a new environment variable `MAX_TRANSACTION_FEE_THRESHOLD` and use it instead of `constants.MAX_GAS_PER_SEC` in `sendEthereumTransaction` in the [following place](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/src/lib/clients/sdkClient.ts#L437)

### Version

v0.55.0

PR #3528, opened 2025-03-06
feat: adds MAX_TRANSACTION_FEE_THRESHOLD env variable

simzzz:
**Description**:
This PR adds a MAX_TRANSACTION_FEE_THRESHOLD environment variable which replaces `constants.MAX_GAS_PER_SEC` in `submitEthereumTransaction(...)`

**Related issue(s)**:

Solves #3002

**Notes for reviewer**:

**Checklist**


charts/hedera-json-rpc-relay/value-test.yaml
```diff
@@ -137,6 +137,7 @@ config:
   FILTER_API_ENABLED: true
   MULTI_SET: true
   GAS_PRICE_PERCENTAGE_BUFFER: 0
+  MAX_TRANSACTION_FEE_THRESHOLD: 15000000
```
konstantinabl, line 140: In the original issue https://github.com/hashgraph/hedera-json-rpc-relay/issues/3002
Its mentioned that this value may be too high to be the default, has it been discussed as part of the issue, if we should change it? If not, maybe it makes sense to

konstantinabl (changes requested):
A question before merging

konstantinabl (dismissed):

charts/hedera-json-rpc-relay/value-test.yaml
quiet-node, line 140: Let's remove this from the charts as the config values are going to be overriden in the ConfigService anyway. We specify these configs in the charts if we want something different compared to the default.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -426,7 +426,9 @@ export class SDKClient {
     const networkGasPriceInTinyBars = weibarHexToTinyBarInt(networkGasPriceInWeiBars);
 
     ethereumTransaction.setMaxTransactionFee(
-      Hbar.fromTinybars(Math.floor(networkGasPriceInTinyBars * constants.MAX_GAS_PER_SEC)),
```
quiet-node, line 429: If MAX_GAS_PER_SEC is supposed to be replaced with the new config, should we remove MAX_GAS_PER_SEC from the constant object?

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -426,7 +426,9 @@ export class SDKClient {
...
+      Hbar.fromTinybars(
+        Math.floor(networkGasPriceInTinyBars * Number(ConfigService.get('MAX_TRANSACTION_FEE_THRESHOLD'))),
```
quiet-node, line 430: Hmmm, ConfigService.get() should already auto-assign the type, right? Why do we need to wrap it with Number()?

packages/relay/src/lib/constants.ts
```diff
@@ -230,4 +230,6 @@ export default {
     RECORD: `RECORD`,
     TRANSACTION: `TRANSACTION`,
   },
+
+  MAX_TRANSACTION_FEE_THRESHOLD: ConfigService.get('MAX_TRANSACTION_FEE_THRESHOLD'),
```
quiet-node, line 234: I don’t see this new constant being used anywhere—why initialize it without using it? Also, since we’re already defining this constant here, can we use it here https://github.com/hashgraph/hedera-json-rpc-relay/pull/3528/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR430?

quiet-node (changes requested):

packages/relay/src/lib/clients/sdkClient.ts
simzzz: TypeScript didn't recognize it as Number but it should work when I take it from constants

simzzz: Hmm yeah, we don't

packages/relay/src/lib/clients/sdkClient.ts
simzzz: MAX_GAS_PER_SEC still represents the technical network limit and it is used in `getCappedBlockGasLimit` while MAX_TRANSACTION_FEE_THRESHOLD is meant to limit the spending exposure of the operator so I think it should stay.

konstantinabl (approved):

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Nice one yeah probably better to keep a constant for that rather than a config thanks for the clarification.

quiet-node (approved):
Nice one! LGTM

PR #3531, opened 2025-03-07
feat: adds support for arrays in ConfigService

### LINKED ISSUE #3178: Handle and parse env variables of type "array" in the config service
### Problem

Currently, there is no mechanism to parse and use environment variables from the type "array." They are used as plain strings, and they are explicitly parsed on each call/usage.

### Solution

- add a new case for type "array" here https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/config-service/src/services/validationService.ts#L59
- fix all places where `JSON.parse` is being called on a plain string but the env variable is defined as an "array" in the globalConfig.ts

### Alternatives

_No response_

simzzz:
**Description**:
This PR adds support for array type environment variables in the Hedera JSON RPC Relay configuration system. Previously, array-type environment variables were stored as strings and had to be explicitly parsed with JSON.parse() on each usage. This implementation provides a more type-safe approach .

**Related issue(s)**:

Fixes #3178 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/config-service/src/services/globalConfig.ts
+++ b/packages/config-service/src/services/globalConfig.ts
@@ -11,12 +11,13 @@
 type ExtractTypeStringFromKey<K extends string> = K extends keyof typeof _CONFIG ? (typeof _CONFIG)[K]['type'] : never;
 
 /**
- * Maps string representations of types (`'string'`, `'boolean'`, `'number'`) to their actual TypeScript types.
+ * Maps string representations of types (`'string'`, `'boolean'`, `'number'`, `'array'`) to their actual TypeScript types.
  *
  * Example:
  * - `'string'` → string
  * - `'boolean'` → boolean
  * - `'number'` → number
+ * - `'array'` → any[]
  */
 type StringTypeToActualType<Tstr extends string> = Tstr extends 'string'
   ? string
@@ -24,6 +25,8 @@ type StringTypeToActualType<Tstr extends string> = Tstr extends 'string'
   ? boolean
   : Tstr extends 'number'
   ? number
+  : Tstr extends 'array'
+  ? any[]
   : never;
 
 /**
@@ -62,9 +65,9 @@ export type GetTypeOfConfigKey<K extends string> = CanBeUndefined<K> extends tru
  */
 export interface ConfigProperty {
   envName: string; // Environment variable name
-  type: 'string' | 'number' | 'boolean'; // Data type of the configuration property
+  type: 'string' | 'number' | 'boolean' | 'array'; // Data type of the configuration property
   required: boolean; // Whether the property is required
-  defaultValue: string | number | boolean | null; // Default value (if any)
+  defaultValue: string | number | boolean | readonly any[] | null; // Default value (if any)
 }
 
 /**
@@ -75,17 +78,17 @@ export interface ConfigProperty {
 const _CONFIG = {
   BATCH_REQUESTS_DISALLOWED_METHODS: {
     envName: 'BATCH_REQUESTS_DISALLOWED_METHODS',
-    type: 'string',
+    type: 'array',
     required: false,
-    defaultValue: `[
-      "debug_traceTransaction",
-      "eth_newFilter",
-      "eth_uninstallFilter",
-      "eth_getFilterChanges",
-      "eth_getFilterLogs",
-      "eth_newBlockFilter",
-      "eth_newPendingTransactionFilter"
-    ]`,
+    defaultValue: [
+      'debug_traceTransaction',
+      'eth_newFilter',
+      'eth_uninstallFilter',
+      'eth_getFilterChanges',
+      'eth_getFilterLogs',
+      'eth_newBlockFilter',
+      'eth_newPendingTransactionFilter',
+    ],
   },
   BATCH_REQUESTS_ENABLED: {
     envName: 'BATCH_REQUESTS_ENABLED',
@@ -185,9 +188,9 @@ const _CONFIG = {
   },
   ETH_CALL_ACCEPTED_ERRORS: {
     envName: 'ETH_CALL_ACCEPTED_ERRORS',
-    type: 'string',
+    type: 'array',
     required: false,
-    defaultValue: '[]',
+    defaultValue: [],
   },
   ETH_CALL_CACHE_TTL: {
     envName: 'ETH_CALL_CACHE_TTL',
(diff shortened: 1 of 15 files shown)


packages/config-service/src/services/globalConfig.ts
```diff
@@ -11,19 +11,22 @@
...
+ * - `'array'` → any[]
  */
 type StringTypeToActualType<Tstr extends string> = Tstr extends 'string'
   ? string
   : Tstr extends 'boolean'
   ? boolean
   : Tstr extends 'number'
   ? number
+  : Tstr extends 'array'
+  ? any[]
```
acuarica, line 29: should we refine this to `string[]|number[]`? this is to avoid `any` and to be more declarative.

Or should we even have two different array types, `'num-array'` and `'str-array'` to distinguish between them.

packages/config-service/src/services/validationService.ts
```diff
@@ -11,13 +11,21 @@ export class ValidationService {
...
           throw new Error(`Configuration error: ${entryName} is a mandatory configuration for relay operation.`);
         }
 
         if (entryInfo.type === 'number' && isNaN(Number(envs[entryName]))) {
           throw new Error(`Configuration error: ${entryName} must be a valid number.`);
         }
+
+        if (entryInfo.type === 'array' && envs[entryName]) {
+          try {
+            JSON.parse(envs[entryName] as string);
```
acuarica, line 24: we are doing this `JSON.parse` twice, right?

packages/config-service/src/services/validationService.ts
```diff
@@ -51,8 +60,17 @@ export class ValidationService {
         case 'boolean':
           typeCastedEnvs[entryName] = envs[entryName] === 'true';
           break;
+        case 'array':
+          try {
+            const value = envs[entryName];
+            typeCastedEnvs[entryName] = value ? JSON.parse(value) : [];
+          } catch (e) {
+            // If parsing fails, fall back to the string value
+            typeCastedEnvs[entryName] = envs[entryName];
```
acuarica, line 69: this would defeat the purpose right? because we might have `string`s as well

packages/config-service/src/services/globalConfig.ts
simzzz: Yes, I would go with two different types as it's more precise

packages/config-service/src/services/validationService.ts
simzzz: Yes, once in startUp for validation and once in typeCasting for the actual conversion

packages/config-service/src/services/validationService.ts
simzzz: yes, it's fixed in my next commit

packages/relay/tests/lib/hapiService.spec.ts
```diff
@@ -103,9 +103,10 @@ describe('HAPI Service', async function () {
     });
   });
 
-  withOverriddenEnvsInMochaTest({ HAPI_CLIENT_ERROR_RESET: '[50]' }, () => {
+  withOverriddenEnvsInMochaTest({ HAPI_CLIENT_ERROR_RESET: [50] }, () => {
     it('should be able to reinitialise SDK instance upon error status code encounter', async function () {
-      const hapiClientErrorReset: Array<number> = JSON.parse(ConfigService.get('HAPI_CLIENT_ERROR_RESET'));
+      // Use type assertion for backward compatibility until all code is updated
```
acuarica, line 108: q: where is the type assertion?

packages/config-service/src/services/validationService.ts
```diff
@@ -28,7 +46,8 @@ export class ValidationService {
    * - If the env var is missing but has a default value, use the default
    * - For 'number' type, converts to Number
    * - For 'boolean' type, converts 'true' string to true boolean
-   * - For 'string' and 'array' types, keeps as string
+   * - For 'array' type, parses JSON string to array
```
acuarica, line 49: nit

```suggestion
   * - For 'numArray' or 'strArray' types, parses JSON string to array
```

packages/config-service/src/services/validationService.ts
```diff
@@ -11,13 +11,31 @@ export class ValidationService {
...
+
+        if ((entryInfo.type === 'strArray' || entryInfo.type === 'numArray') && envs[entryName]) {
+          let parsed;
+          try {
+            parsed = JSON.parse(envs[entryName] as string);
+          } catch (e) {
+            throw new Error(`Configuration error: ${entryName} must be a valid JSON array.`);
+          }
+          if (!Array.isArray(parsed)) {
+            throw new Error(`Configuration error: ${entryName} must be a valid JSON array.`);
```
acuarica, line 30: I'd use different error messages to distinguish where it failed.

acuarica (approved):
Nice addition!

Left some comments we may want to address.

packages/relay/tests/lib/hapiService.spec.ts
simzzz: I planned to separate the new types and replacing them in the code in separate PR's initially so this is a leftover comment. I will remove it.

acuarica (dismissed):

packages/config-service/src/services/validationService.ts
```diff
@@ -11,13 +11,31 @@ export class ValidationService {
...
+          }
+          if (entryInfo.type === 'numArray' && !parsed.every((item) => typeof item === 'number')) {
+            throw new Error(`Configuration error: ${entryName} must contain only numbers.`);
+          }
+          if (entryInfo.type === 'strArray' && !parsed.every((item) => typeof item === 'string')) {
+            throw new Error(`Configuration error: ${entryName} must contain only strings.`);
+          }
+        }
```
quiet-node, line 38: This looks a bit cluttered. Clade suggested the following which I think it's pretty clean. wdyt?

```typescript
if ((entryInfo.type === 'strArray' || entryInfo.type === 'numArray') && envs[entryName]) {
          try {
            const parsed = JSON.parse(envs[entryName] as string);
            
            if (!Array.isArray(parsed)) {
              throw new Error(`Configuration error: ${entryName} must be a valid JSON array.`);
            }
            
            const isCorrectType = entryInfo.type === 'numArray' 
              ? parsed.every((item) => typeof item === 'number')
              : parsed.every((item) => typeof item === 'string');
              
            if (!isCorrectType) {
              const expectedType = entryInfo.type === 'numArray' ? 'numbers' : 'strings';
              throw new Error(`Configuration error: ${entryName} must contain only ${expectedType}.`);
            }
          } catch (e) {
            if (e instanceof SyntaxError) {
              throw new Error(`Configuration error: ${entryName} must be a valid JSON string.`);
            }
            throw e;
          }
        }
```

quiet-node (commented):
Nice one! Definitely a nice touch for the ConfigService! Left a suggestion though!

packages/config-service/src/services/validationService.ts
simzzz: Cool, I like it.

quiet-node (approved):
Perfecto! LGTM

konstantinabl (approved):

Issue #3516, opened 2025-02-27
Design a deterministic WHBAR deployment script or flow


natanasow:
### Problem

Currently, the WHBAR address depends on the signer's nonce which on cross-chain deployment will resolve in different WHBAR addresses.

### Solution

Create an action plan and flow on how to deploy WHBAR on multiple chains with the same address.

Due to the urgency of that task, we decided to stick with the clean account and nonce 0 for cross-chain deployment. For now, we will not use `create2` factory.

PR #3546, opened 2025-03-13
feat: add `WHBAR` contract deployment example under tools folder

### LINKED ISSUE #3517: Execute `WHBAR` tests against already existing contract
### Problem

Currently, there is no way to execute the `WHBAR.js` against an already existing contract.

We need to implement a custom test script to test against an already deployed contract, Take into consideration that  we will need to store somehow the final addresses in all the envs and networks of the whbar

### Solution

Tasks:
- [x] - Research how this can be achieved.
- [x] - Propose a solution e.g. scripts under /scripts or GitHub Action or something else.
- [x] - Define needed .env or action arguments like the signer's private key, receiver's address, target network, and contract's address.
- [x] - Run some tests against already existing `WHBAR` contract on testnet

### Alternatives

_No response_

natanasow:
**Description**:

Add `WHBAR` deployment flow.

**Related issue(s)**:

Fixes #3517 and #3516

**Notes for reviewer**:

**Checklist**


andrewb1269hg (approved):
Review and approve `tools/whbar-hardhat-example/.gitignore`

konstantinabl (approved):

PR #3547, opened 2025-03-13
feat: change acceptance.yml to use matrix and tweak configurations for a more optimal CI flow

### LINKED ISSUE #3353: Intermittent Test Failures in Acceptance Suite Batches

### LINKED ISSUE #3536: Set max parallel jobs on Github to a lower value
### Problem

While working on https://github.com/hashgraph/hedera-json-rpc-relay/issues/3353 I noticed tests fail more often when all of them are ran in parallel but when you rerun them manually one by one they almost always pass. It might seem that they are competing for the limited resources that the runner has

The team from solo have a limit for this reason - https://github.com/hashgraph/solo/blob/main/.github/workflows/flow-pull-request-checks.yaml#L95 and implementing a similar limit could improve the flakiness that we are experiencing.

### Solution

- Set max-parallel value to 5 or another value

### Alternatives

_No response_

simzzz:
**Description**:
This PR changes the structure of the acceptance.yml file to use a matrix, which makes the configuration look more clean and readable, and also adds a limit to the max amount of jobs that can run in parallel so they don't compete for resources.
The current limit with great results I'm using is 6. This doesn't slow down the full flow by much since if a job completes a new one is picked up right away, it doesn't wait for the first 6 to finish before picking up the next 6 etc. 

Other tweaks that showed great results:
- Set `USE_ASYNC_TX_PROCESSING` to false  for the job. For some reason the runner doesn't like this value when we set it to true, causing timeouts and errors when communicating with the mirror node sometimes. This doesn't happen locally and the feature has test coverage and otherwise works as expected.
- Increase `HBAR_RATE_LIMIT_TINYBAR`, for some longer and more expensive jobs such as API Batch 1. In case they get retried we often get `-32606 HBAR_RATE_LIMIT_EXCEEDED` error on the retries. We can bump it up a bit just to be safe.

**Related issue(s)**:

Fixes #3536, fixes #3353

**Notes for reviewer**:

**Checklist**


quiet-node:
> Set USE_ASYNC_TX_PROCESSING to false for the job. For some reason the runner doesn't like this value when we set it to true, causing timeouts and errors when communicating with the mirror node sometimes. This doesn't happen locally and the feature has test coverage and otherwise works as expected.

Nice catch! Yes, USE_ASYNC_TX_PROCESSING=true might cause a timeout. If an internal exception is thrown during the sendRawTransaction request flow, a valid transaction hash is still returned, but the transaction never reaches consensus, leaving no record on MN. As a result, the poller continuously checks for the transaction record in an endless loop since the polling result will always be null. More info about this can be found in [#3318](https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3318).

.github/workflows/acceptance-workflow.yml
```diff
@@ -51,6 +51,8 @@ jobs:
           if [ -n "${{ inputs.operator_id }}" ]; then
             echo "OPERATOR_ID_MAIN=${{ inputs.operator_id }}" >> $GITHUB_ENV
           fi
+          echo "USE_ASYNC_TX_PROCESSING=false" >> $GITHUB_ENV
+          echo "HBAR_RATE_LIMIT_TINYBAR=5000000000000" >> $GITHUB_ENV
```
quiet-node, line 55: Hm I don't know if setting config here would scale or not? Did you try updating it in the localAcceptance.env instead?

.github/workflows/acceptance.yml
```diff
@@ -12,168 +12,41 @@ concurrency:
...
+          - { name: 'Precompile Calls', testfilter: 'precompile-calls' }
+          - { name: 'Websocket Batch 1', testfilter: 'ws_batch1', test_ws_server: true }
+          - { name: 'Websocket Batch 2', testfilter: 'ws_batch2', test_ws_server: true }
+          - { name: 'Websocket Batch 3', testfilter: 'ws_batch3', test_ws_server: true }
+          - { name: 'Cache Service', testfilter: 'cache-service' }
+          - { name: 'Server Config', testfilter: 'serverconfig' }
+    uses: ./.github/workflows/acceptance-workflow.yml
+    with:
+      testfilter: ${{ matrix.test.testfilter }}
+      test_ws_server: ${{ matrix.test.test_ws_server || false }}
```
quiet-node, line 42: nice one definitely looks cleaner!

quiet-node (changes requested):
LGTM but one question

quiet-node (approved):
Nice work! LGTM!

quiet-node:
Hmmm still see it takes up to 6 attempts for the tests. Does the mattrix actually help? Also should we consider breaking up the tests in smaller batch like maybe adding 1 more batch like API Batch 4 or something?

PavelSBorisov (approved):

Issue #3551, opened 2025-03-17
Epic: Implement Dispatcher Pattern for Relay Package and Improve Error Handling and Mapping


quiet-node:
### Description:
The current Relay package uses a monolithic request-handling approach, which limits scalability and flexibility. This proposal introduces a Dispatcher Pattern to centralize request routing and improve modularity. The Dispatcher will decouple request processing from core logic, making it easier to extend, test, and maintain. The goal is to enhance scalability, flexibility, and overall code maintainability by introducing a centralized routing mechanism and refactoring existing handlers to work with it.

### Solution:

Create a Dispatcher class to handle dynamic request routing and error handling.
Refactor existing handlers to work with the Dispatcher.
Improve logging, error handling, and test coverage.

PR #3555, opened 2025-03-18
refactor: normalize routes and reorganize service implementations

### LINKED ISSUE #3552: Code Refactoring & Route Normalization
### Description:
The first stage of the dispatcher implementation involves refactoring existing code and normalizing route definitions to prepare for the dispatcher. This includes reorganizing namespaces, simplifying parameter handling, and establishing a consistent method invocation pattern across all routes. These changes will make the codebase cleaner and easier to integrate with the new dispatcher.

### Solution:

- Reorganize debugServiceImpl under relay.ts instead of eth.ts.
- Normalize debug_traceTransaction parameter decoding and simplify debugRoutes.ts.
- Introduce a facade method for eth_newFilter and standardize filtering-related calls.
- Ensure requestDetails is always the last parameter in method calls.
- Refactor eth_estimateGas logic into eth().estimateGas().
- Standardize route definitions to directly pass parameters to method implementations.

quiet-node:
### **Description**:
This PR implements the first stage of the dispatcher pattern by refactoring existing code and normalizing route handling across the codebase. The changes focus on separating concerns, improving type safety, and establishing consistent parameter patterns to prepare for future dispatcher integration.

### **Changes**:
Service Organization:
 - Decoupled DebugImpl service from EthImpl, making it a standalone service in the relay
 - Added proper facade methods in EthImpl for FilterService operations

Parameter Standardization:
 - Consolidated getLogs parameters into a typed IGetLogsParams object
 - Consolidated newFilter parameters into a typed INewFilterParams object
 - Moved Metamask empty data hotfix from route handler to EthImpl implementation
 - Ensured requestDetails is consistently the last parameter in method calls

Route Simplification:
 - Normalized debug_traceTransaction parameter handling
 - Standardized route definitions to directly pass parameters to method implementations
 - Simplified request parameter validation

**Related issue(s)**: EPIC #3551

Fixes #3552

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/services/ethService/ethFilterService/index.ts
```diff
@@ -94,28 +95,16 @@ export class FilterService implements IFilterService {
...
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(
-        `${requestIdPrefix} newFilter(fromBlock=${fromBlock}, toBlock=${toBlock}, address=${address}, topics=${topics})`,
-      );
-    }
+  async newFilter(params: INewFilterParams, requestDetails: RequestDetails): Promise<string> {
     try {
       FilterService.requireFiltersEnabled();
 
+      const fromBlock = params?.fromBlock === undefined ? 'latest' : params?.fromBlock;
```
natanasow, line 105: nit: We can use `CommonService.blockLatest` for 'latest', I know it's not changed in this PR and it's not in this PR scope but it might be worth fixing it with minimum effort.

natanasow (approved):
LG, great work as always!

packages/relay/src/lib/services/ethService/ethFilterService/index.ts
quiet-node: hey great catch! I definitely overlooked this! Thanks for pointing it out!

quiet-node: Updated with the changes! Thanks 🫡!

natanasow (dismissed):

packages/relay/src/lib/debug.ts
```diff
@@ -40,25 +42,23 @@ export class DebugService implements IDebugService {
   private readonly common: CommonService;
 
   /**
-   * Creates an instance of DebugService.
+   * Creates an instance of DebugImpl.
```
konstantinabl, line 45: I dont quite agree with completely removing the DebugService. I thought the idea is to introduce DebugImpl as an abstraction over the service (kinda like a controller, as we plan for EthImpl) which routes requests to the service. In this way we stick to the general approach that we want, having services. It doesn't make sense to me to have EthImpl which will call different services, but then DebugImpl directly executes business logic
Wdyt @quiet-node and @acuarica

quiet-node: Yeah I see your point.

However, aren't the services involved in EthImpl, like common, filter, cache, and HAPI services, more like utility services rather than a dedicated EthService? DebugService was, as we believed, originally introduced as a utility service for EthImpl because until now debug_traceTransaction still calls Relay.eth().debugService().traceTransaction(). As we’ve discussed, that wasn’t the right architecture—debug_traceTransaction should be calling Relay.debug().traceTransaction() instead.

Regarding the solution that keeps DebugService and uses DebugImpl as an abstraction layer, I don’t think it aligns with our current setup and may not be necessary. Our architecture doesn’t have a dedicated service for each implementation; instead, it relies on multiple utility services.

konstantinabl: We will have EthImpl containing services calling the method that a user needs as per the design doc we discussed for eth.ts refactor
If someone calls eth_getBlockByNumber the flow will be server.ts -> relay.ts -> eth.ts -> blockService.ts
Why should we have a different approach for debug then?
server.ts -> relay.ts -> debug.ts
In my proposal for eth.ts refactor we separate the business logic in services, so I think we should do this for debug as well, and maybe even for the rest of the implementations. But since we havent discussed or planned, maybe it wont happen now, anyways we already have the debugService

packages/relay/src/lib/relay.ts
```diff
@@ -25,7 +26,7 @@ import { SubscriptionController } from './subscriptionController';
 import { RequestDetails } from './types';
 import { Web3Impl } from './web3';
 
-export class RelayImpl implements Relay {
+export class RelayImpl {
```
acuarica, line 29: Given we don't have an interface here, we can rename the class to just `Relay`.

packages/relay/src/lib/eth.ts
```diff
@@ -2774,24 +2846,25 @@ export class EthImpl implements Eth {
...
    *     - If the timestamp range between `fromBlock` and `toBlock` exceeds 7 days, a predefined error `TIMESTAMP_RANGE_TOO_LARGE` is thrown.
    *
-   * @param {string | null} blockHash - The block hash to prioritize log retrieval.
-   * @param {string | 'latest'} fromBlock - The starting block for log retrieval.
-   * @param {string | 'latest'} toBlock - The ending block for log retrieval.
-   * @param {string | string[] | null} address - The address(es) to filter logs by.
-   * @param {any[] | null} topics - The topics to filter logs by.
-   * @param {RequestDetails} requestDetails - The details of the request.
-   * @returns {Promise<Log[]>} - A promise that resolves to an array of logs or an empty array if no logs are found.
+   * @param {IGetLogsParams} params - The parameters for the getLogs method.
```
acuarica, line 2849: I'd put the JSdocs in the type itself rather than here

packages/ws-server/src/webSocketServer.ts
```diff
@@ -32,7 +32,7 @@ const mainLogger = pino({
 });
 const register = new Registry();
 const logger = mainLogger.child({ name: 'rpc-ws-server' });
-const relay: Relay = new RelayImpl(logger, register);
+const relay: RelayImpl = new RelayImpl(logger, register);
```
acuarica, line 35: nit

```suggestion
const relay = new RelayImpl(logger, register);
```

packages/server/src/routes/debugRoutes.ts
```diff
@@ -1,14 +1,16 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { Relay } from '@hashgraph/json-rpc-relay';
-import { ITracerConfig, RequestDetails } from '@hashgraph/json-rpc-relay/dist/lib/types';
+import { RelayImpl } from '@hashgraph/json-rpc-relay';
+import { TracerType } from '@hashgraph/json-rpc-relay/dist/lib/constants';
```
acuarica, line 4: is `TracerType` duplicated in `../validator`? If yes, we should keep only one.

packages/relay/src/lib/eth.ts
```diff
@@ -782,6 +770,90 @@ export class EthImpl implements Eth {
...
+   * @param {INewFilterParams} params - The parameters for the new filter
+   * @param {RequestDetails} requestDetails - Details about the request for logging and tracking
+   * @returns {Promise<string>} A filter ID that can be used to query for changes
+   */
+  async newFilter(params: INewFilterParams, requestDetails: RequestDetails): Promise<string> {
+    const requestIdPrefix = requestDetails.formattedRequestId;
+    if (this.logger.isLevelEnabled('trace')) {
+      this.logger.trace(`${requestIdPrefix} newFilter(params=${JSON.stringify(params)})`);
+    }
+    return this.filterService.newFilter(params, requestDetails);
```
acuarica, line 785: We should destructure `params` here, to avoid just passing through to `filterService.newFilter`.

packages/relay/src/lib/types/requestParams.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Interface representing parameters for the getLogs method.
+ * Used to filter and retrieve logs from the blockchain.
+ *
+ * @param blockHash - Hash of the block to get logs from. If null, logs are not filtered by block hash.
+ * @param fromBlock - The block number or 'latest' to start fetching logs from.
+ * @param toBlock - The block number or 'latest' to stop fetching logs at.
+ * @param address - Contract address or list of addresses to filter logs by. If null, logs are not filtered by address.
+ * @param topics - Array of topics to filter logs by. If null, logs are not filtered by topics.
+ */
+export interface IGetLogsParams {
```
acuarica, line 13: I'd put these where the `interface Eth` lives, this would be to avoid indirection and just importing/export types.

acuarica (commented):
Looking good, left some comments

packages/relay/src/lib/debug.ts
acuarica: > We will have EthImpl containing services calling the method that a user needs as per the design doc we discussed for eth.ts refactor

That's true, but only applies to `EthImp`. That's because its size and complexity that we decided to split it into smaller more manageable services. But it doesn't mean that all namespaces will have the same structure. For example both `net` and `web3` are also implemented "inplace" because there is no need for any services there.

> If someone calls eth_getBlockByNumber the flow will be Relay.eth().blockService().getBlockByNumber()

True, but that dispatch lives inside `eth.ts`. It's `EthImpl` who decides that `eth_getBlockByNumber` should be resolved by the `blockService`. `debug` is another top-level namespace, and it can decide how it dispatches its calls independently of `EthImpl`.

konstantinabl: I dont agree, but whatever floats your boat

quiet-node: > If someone calls eth_getBlockByNumber the flow will be Relay.eth().blockService().getBlockByNumber()

Again, EthImpl is a much larger and more complex implementation class. Also services like blockService(), transactionService(), filterService(), etc. work more as utility services. As Luis mentioned, we chose to break them down due to EthImpl’s complexity, as it interacts with multiple services and aspects, such as block and transaction-related operations. So having utility "services" classes would be clean and efficient.

However, having utility services doesn’t mean every namespace should have them, even when a namespace primarily handles a single type of operation or are simply small enough. In cases like `net`, `web3`, and `debug`, I believe it would be cleaner for these smaller namespaces to have direct access to their methods rather than introducing service classes, which could be unnecessary overhead.

> Why should we have a different approach for debug then?
> Relay.debug().traceTransaction()

Because debug_traceTransaction now belongs to the `debug` namespace rather than `eth`, and the approach differs from how it was previously handled.

> In my proposal for eth.ts refactor we separate the business logic in services, so I think we should do this for debug as well, and maybe even for the rest of the implementations.

Maintaining a consistent layout would require introducing dedicated services for `net` and `web3` as well. However, given the scope and size of these namespaces compared to `eth`, I think it would be overkilled to have dedicated services for them.

> But since we havent discussed or planned, maybe it wont happen now, anyways we already have the debugService.

`debugService` was originally introduced because `traceTransaction()` was mistakenly placed under the `eth` namespace. However, it is meant to exist at the same level as `EthImpl`, `Web3Impl`, and `NetImpl` , which is why we now have `DebugImpl`. The term `service` in `debugService` was intended to indicate a utility for `EthImpl`, but it shouldn’t have been named with `Service` at the beginning.

packages/relay/src/lib/eth.ts
quiet-node: Yeah I already have JSDocs for the type itself. Are you suggesting the below?

Instead of having all,
```
   ...
   * @param {IGetLogsParams} params - The parameters for the getLogs method.
   * @param {string | null} params.blockHash - Hash of the block to get logs from. If null, logs are not filtered by block hash.
   * @param {string | 'latest'} params.fromBlock - The block number or 'latest' to start fetching logs from.
   * @param {string | 'latest'} params.toBlock - The block number or 'latest' to stop fetching logs at.
   * @param {string | string[] | null} params.address - Contract address or list of addresses to filter logs by. If null, logs are not filtered by address.
```

Just only
```
   ...
   * @param {IGetLogsParams} params - The parameters for the getLogs method.
```

packages/ws-server/src/webSocketServer.ts
quiet-node: NIce catch!

packages/relay/src/lib/eth.ts
acuarica: Yes, this avoids duplicating both the param type and its docs.

packages/server/src/routes/debugRoutes.ts
quiet-node: Indeed! Removed one

packages/ws-server/src/webSocketServer.ts
quiet-node: updated.

packages/relay/src/lib/relay.ts
quiet-node: updated.

packages/relay/src/lib/types/requestParams.ts
quiet-node: Hmmm I think index.ts is already quite large with just interface definitions for the APIs, adding detailed parameter interfaces would make it quite clutter. 

I feel that keeping index.ts limited to only the main interfaces would help maintain clarity, especially since we already have several interfaces in the centralized types/ folder.

wdyt?

acuarica: So it depends on usage. `IGetLogsParams` is "part" of the `Eth` interface (in the sense that every time you import `IGetLogsParams` you also import or define `Eth`). So I believe it's more clear if we put the public interface of `Eth` in one place, otherwise it's gets tricker to follow the dependencies.

In any case, we can change this later if we feel like. Going to mark this as resolved.

packages/relay/src/lib/eth.ts
acuarica: It was two fold, first avoiding make more modifications on an already big PR. Second because of usage. Internally we use `filterService.newFilter` and there is no need to create an object just to satisfy `filterService.newFilter` param type.

However, seeing the tests it might actually make more sense to have it as an object (it's clear that way to see the meaning of each argument).

packages/relay/src/lib/eth.ts
quiet-node: yeah bet! Updated with the removal of unnecessary docs. Thanks for pointing it out

packages/relay/src/lib/eth.ts
quiet-node: Ah I see yeah solid points will keep that in mind!

packages/relay/src/lib/types/requestParams.ts
quiet-node: Yeah okay I see but yeah let's change it later as again I just don't wanna clutter index.ts lol. Thanks for bringing this up tho

konstantinabl (approved):

acuarica (approved):

PR #3556, opened 2025-03-18
chore: removes redundant env config values from helm charts

### LINKED ISSUE #2476: Remove redundant env config values from helm charts
### Problem

The relay has the recommended and default values hardcoded within the app. however we are overriding these with the one with the chart values, what is forcing devs to update the defaults in both places when there is a change or optimization, and in some cases there is an inconsistency between them due to devs forget about the charts.

### Solution

review and remove all duplicate and default values from the chart values files.

review files:

- https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/charts/hedera-json-rpc-relay/values.yaml
- https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/charts/hedera-json-rpc-relay/value-test.yaml
- https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/charts/hedera-json-rpc-relay-websocket/values.yaml

leave as few configs on the values as possible only when there is a real need to have it different than the defaults on app.

### Alternatives

_No response_

simzzz:
**Description**:
This PR removes values that are duplicated in the helm charts config and the default config

**Related issue(s)**:

Fixes #2476 

**Notes for reviewer**:

**Checklist**


charts/hedera-json-rpc-relay-websocket/values.yaml
```diff
@@ -8,32 +8,70 @@ autoscaling:
...
-  WS_CONNECTION_LIMIT: 100
-  WS_MAX_INACTIVITY_TTL: 300000
-  WS_MULTIPLE_ADDRESSES_ENABLED: false
-  WS_SUBSCRIPTION_LIMIT: 100
-  WS_PING_INTERVAL: 1000
+
+  # Operator Account (Required for transaction execution)
+  # OPERATOR_ID_MAIN:
+  # OPERATOR_KEY_MAIN:
+  # OPERATOR_KEY_FORMAT:
```
quiet-node, line 22: I think we should uncomment this just to indicate they should be filled out too along with other requried configs otherwise ConfigService will not be happy

charts/hedera-json-rpc-relay-websocket/values.yaml
```diff
@@ -8,32 +8,70 @@ autoscaling:
...
+
+  # WebSocket Configuration
+  SUBSCRIPTIONS_ENABLED: true
```
quiet-node, line 25: Should we comment this out along with other not required ones?

charts/hedera-json-rpc-relay-websocket/values.yaml
```diff
@@ -8,32 +8,70 @@ autoscaling:
...
+  # MIRROR_NODE_AGENT_CACHEABLE_DNS:
+
+  # ========== ETH CALL CONFIGURATION ==========
+  # ETH_CALL_DEFAULT_TO_CONSENSUS_NODE:
+  # CONSENSUS_MAX_EXECUTION_TIME:
+  # SDK_REQUEST_TIMEOUT:
+  # CONTRACT_QUERY_TIMEOUT_RETRIES:
+
+  # ========== REDIS CONFIGURATION ==========
   REDIS_ENABLED: false
```
quiet-node, line 64: Should we comment this out along with other not required ones?

charts/hedera-json-rpc-relay-websocket/values.yaml
```diff
@@ -8,32 +8,70 @@ autoscaling:
...
-  USE_ASYNC_TX_PROCESSING: true
+  # REDIS_URL:
+  # REDIS_RECONNECT_DELAY_MS:
+  # MULTI_SET:
+
+  # ========== OTHER SETTINGS ==========
+  # CLIENT_TRANSPORT_SECURITY:
+  # USE_ASYNC_TX_PROCESSING:
+  # DEV_MODE:
+  LOG_LEVEL: ''
```
quiet-node, line 73: Should we comment this out along with other not required ones?

charts/hedera-json-rpc-relay-websocket/values.yaml
Ferparishuertas: @quiet-node it is WS, so it should be true.

charts/hedera-json-rpc-relay-websocket/values.yaml
Ferparishuertas: if you uncomment but not set will fail. Is you want only a rpc-relay for query, this should work,
Comment is explanatory enoygh. I dont see the need to keep them uncommented

charts/hedera-json-rpc-relay-websocket/values.yaml
Ferparishuertas: Here I agreed to comment this out with the default value

Ferparishuertas (dismissed):
LGTM
Only 1 comment out comment

charts/hedera-json-rpc-relay-websocket/values.yaml
simzzz: This is true by default but should be false for now since redis hasn't been added as a dependency yet so I think it makes sense to override it.

quiet-node: Ah I see! Okay that makes sense!

quiet-node (approved):
Perfecto LGTM!

natanasow (approved):

konstantinabl (approved):

PR #3581, opened 2025-03-22
feat: implemented decorator-based RPC method dispatcher and parameter handling system

### LINKED ISSUE #3553: Implement Decorator-Based Dispatcher
### Description:
The second stage involves implementing a dynamic dispatcher using a decorator-based approach to centralize request handling. This will allow the dispatcher to dynamically collect and route RPC methods, simplifying routing logic and ensuring consistent parameter validation and logging across all methods.

### Solution:
Create a Dispatcher class using decorators to register and route RPC methods.
Centralize method definitions within each namespace.
Ensure consistent parameter validation, caching, and error handling at the dispatcher level.

### LINKED ISSUE #3575: Ensure RequestID is generated and part of the Response
### Problem

When I experience any error as Json RPC relay user i want to be able to share my errors with a linked id, enabling the service provider to easily trace my request and give me support.

### Solution

Add the requestId to any failing request, as a standard method of traceability

### Alternatives

TBD

quiet-node:
## Overview
This PR introduces a fully decorator-based API architecture for JSON-RPC method handling. It provides a centralized solution that ensures a clean, type-safe approach to defining, exposing, validating, and dispatching RPC methods across the Relay package.

###  New Single & Centralized API Entry Point
- `RpcMethodDispatcher` is a centralized and robust class that serves as the sole public API entry point for server packages and all consumers. It provides a structured solution for executing RPC methods by autonomously validating requests and routing them to the appropriate operation handler.
- Streamlined request handling flow, incorporating pre-execution method validation, execution, and error-handling phases.
- Standardized response format, ensuring that methods either return a valid result or throw a JsonRpcError back to consumers.

```typescript
// RpcMethodDispatcher used in Relay to create the signle API Entry Point
// relay.ts
public async executeRpcMethod(
    rpcMethodName: string,
    rpcMethodParams: any,
): Promise<any> {
    return this.rpcMethodDispatcher.dispatch(rpcMethodName, rpcMethodParams, requestDetails);
}


// Usage from server or other consumer
// koaJsonRpc.ts in Server package
const result = await relay.executeRpcMethod('eth_getBalance', ['0x123...', 'latest'], requestId);
```

### Decorator-Based Components
- `@rpcMethod` annotates methods for RPC exposure and automatic registration.
- `@rpcParamValidationRules` defines validation rules for method parameters, ensuring data integrity.
- `@rpcParamLayoutConfig ` manages parameter arrangement and transformation for correct parameters layout.
- All decorators function cohesively to establish a declarative, self-documenting API.

```typescript
// Class implementation with decorators
class EthImpl {
    // decorators usage
    @rpcMethod
    chainId(requestDetails: RequestDetails): string {
        // Implementation
    }

    @rpcMethod
    @rpcParamValidationRules({
      0: { type: 'blockHash', required: true },
      1: { type: 'boolean', required: false }
    })
    @rpcParamLayoutConfig(RPC_LAYOUT.custom((params) => [params[0], params[1]]))
    async estimateGas(
        transaction: IContractCallRequest,
        _blockParam: string | null,
        // Implementation
    }
}
```

### Centralized Method Registry
- `rpcMethodRegistry` serves as the single source of truth for all RPC method implementations
- The method registerRpcMethod():
   - Scans implementation classes for methods decorated with `@rpcMethod`
   - Preserves validation rules from `@rpcParamValidationRules ` for runtime parameter validation
   - Retains parameter layout configurations from `@rpcParamLayoutConfig` for correct invocation
   - Maps RPC method names to their implementations in a centralized registry

### Parameter Handling
- Clean separation between validation and parameter structuring
- Autonomous parameter transformation based on method signature
- Support for special parameter layouts (request-details-only, custom params layout, etc.)
- Advanced parameter parsing for complex methods like `traceTransaction`

### Server Package Simplification
- **Eliminated Manual Method Registration:**
   - Removed the `logAndHandleResponse()` and `useRpc()` pattern previously used to register RPC methods
   - Deleted route files (ethRoutes.ts, debugRoutes.ts, netRoutes.ts, etc.) as methods are now auto-registered in the Relay pacakge
   - Server no longer needs to maintain method mappings or handle routing logic
   
- **Validator Migration:**
   - Migrated the entire Validator service from Server package to Relay
   - Validator now works directly with `@rpcParamValidationRules` decorators
   - Validation happens automatically during the dispatch process
   - Simplified error handling with consistent response formats

### Additional Context
**Related issue(s)**: EPIC #3551

Fixes #3553
Fixes #3575

**Notes for reviewer**:
- The Validator service was migrated from the Server package to the Relay component by component, ensuring the preservation of validation logic.
- MirrorNodeClientError and SdkClientError are currently handled at the dispatcher only for timeout errors. A more comprehensive error mapping will be introduced in a follow-up PR for Stage 3.

## Acknowledgments

Special kudos to @acuarica for initially proposing this approach and for the excellent work on the POC PR (#3548). His contributions laid the foundation for this implementation and significantly influenced the design of this PR.


packages/relay/src/lib/net.ts
```diff
@@ -3,32 +3,60 @@
...
 
   constructor() {
     this.chainId = parseInt(ConfigService.get('CHAIN_ID'), 16).toString();
   }
 
   /**
+   * Indicates whether the client is actively listening for network connections.
    * We always return true for this.
+   *
+   * @rpcMethod Exposed as net_listening RPC endpoint
```
acuarica, line 21: nit: do we want to duplicate this doc comment everywhere?

packages/relay/src/lib/net.ts
```diff
@@ -3,32 +3,60 @@
...
   peerCount(): JsonRpcError {
     return predefined.UNSUPPORTED_METHOD;
   }
+
+  /**
+   * Returns the namespace identifier for this implementation.
+   *
+   * @returns {string} The namespace string for the Net implementation.
+   */
+  getNamespace(): string {
```
acuarica, line 59: Not sure if this bit should be here. A direct user of `NetImpl` shoudln't care about its namespace. I think we should do namespace mgmt in the Relay class.

packages/relay/src/lib/debug.ts
```diff
@@ -79,6 +85,12 @@ export class DebugImpl implements Debug {
    * @example
    * const result = await traceTransaction('0x123abc', TracerType.CallTracer, {"tracerConfig": {"onlyTopCall": false}}, some request id);
    */
+  @rpcMethod
+  @rpcParamValidationRules({
+    0: { type: ParamType.TRANSACTION_HASH_OR_ID, required: true },
```
acuarica, line 90: wouldn't it be better if this is an array? as opposed of an object where keys are indexes.

packages/relay/src/lib/web3.ts
```diff
@@ -4,15 +4,49 @@ import { keccak256 } from '@ethersproject/keccak256';
...
+   *
+   * @rpcMethod Exposed as web3_sha3 RPC endpoint
+   * @rpcParamValidationRules Applies JSON-RPC parameter validation according to the API specification
+   *
+   * @param {string} input - The input string to hash.
+   * @returns {string} The SHA3 hash of the input.
+   */
+  @rpcMethod
+  @rpcParamValidationRules({
+    0: { type: ParamType.HEX, required: true },
```
acuarica, line 38: Given that most params are `required: true`, we can accept two variants. See https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3548/files#diff-68236793abaab30e17a8f16cc850d0f238f1615d36c6d4376ea44abb7220099eR17 for details. This would simplify current code, and also it will simplify future RPC method additions.

packages/relay/src/lib/services/registryService/rpcMethodRegistryService.ts
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
```
acuarica, line 1: I'd move this to its parent folder. No need to have a `index.ts` just to export one file/class.

packages/relay/src/lib/services/registryService/rpcMethodRegistryService.ts
```diff
@@ -0,0 +1,72 @@
...
+   * Creates and registers RPC methods from the provided service implementations
+   *
+   * This static method scans each implementation instance for methods decorated with
+   * @rpcMethod and registers them in a map using the convention namespace_operationName
+   * (e.g., eth_blockNumber). The namespace is derived from the implementation class name.
+   *
+   * @param {RpcImplementation[]} implementations - Array of implementation instances containing RPC methods to register
+   * @returns {RpcMethodRegistry} A map of RPC method names to their bound function implementations
+   */
+  public static register(implementations: RpcImplementation[]): RpcMethodRegistry {
```
acuarica, line 21: I'd convert this to a regular function, no need to have a class with no state.

packages/server/tests/postman.json
```diff
@@ -862,7 +862,7 @@
               "pm.test(\"Success\", () => {",
               "    var response = pm.response.json();",
               "    pm.expect(JSON.stringify(response.error)).to.equal(undefined);",
-              "    pm.expect(response.result).to.equal(\"false\");",
+              "    pm.expect(response.result).to.equal(false);",
```
acuarica, line 865: do we want to include this type change in this PR? So, here we are changing the actual `result` of `net_listening`. I understand that this is not heavily used. However, this change should be independent of the refactor, just to keep better keep track of functional changes. WDYT?

packages/server/tests/postman.json
```diff
@@ -628,7 +628,7 @@
               "pm.test(\"Success\", () => {",
               "    var response = pm.response.json();",
               "    pm.expect(response.error.code).to.equal(-32601);",
-              "    pm.expect(response.error.message).to.equal(\"Method unsupported_function not found\");",
+              "    pm.expect(response.error.message).to.contain(\"Method unsupported_function not found\");",
```
acuarica, line 631: q: what's the change here that we needed to relax the condition?

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -188,37 +165,19 @@ export default class KoaJsonRpc {
 
   async getRequestResult(request: IJsonRpcRequest, ip: string): Promise<IJsonRpcResponse> {
```
acuarica, line 166: I think we should `unknown` here given `request` has not been validate it yet

```suggestion
  async getRequestResult(request: unknown, ip: string): Promise<IJsonRpcResponse> {
```

We should change its type after `validateJsonRpcRequest`.

packages/relay/tsconfig.json
```diff
@@ -15,7 +15,8 @@
         "noImplicitAny": false,
         "declaration": true,
         "strict": true,
-        "sourceMap": true
+        "sourceMap": true,
+        "experimentalDecorators": true
```
acuarica, line 19: we can leave this for now, but we should update to TypeScript 5 where decorators behave a bit different https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#differences-with-experimental-legacy-decorators

packages/relay/src/lib/dispatcher/index.ts
```diff
@@ -0,0 +1,3 @@
+// SPDX-License-Identifier: Apache-2.0
+
+export * from './rpcMethodDispatcher';
```
acuarica, line 3: would move one level up and remove this `index.ts`

acuarica (commented):
Initial round of reviews, thanks for the effort! we are on good track.

packages/relay/src/lib/net.ts
quiet-node: Yeah, it may be repetitive, but I believe it clarifies the intent of the annotation and helps prevent confusion

packages/relay/src/lib/net.ts
quiet-node: Good point! I’ve updated it to pass the namespaces as an argument to the registry method, so this has been removed.

packages/relay/src/lib/debug.ts
quiet-node: As discussed I believe we should keep this as-is for now to avoid introducing new logic on top of the PR which is already quite too big. Created this ticket here to keep track https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3592

packages/relay/src/lib/web3.ts
quiet-node: As discussed offline, I believe keeping the `required` field improves clarity within the method’s context. Additionally, I want to avoid introducing new logic in this large PR and instead preserve the existing functionality also discussed here https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3581#discussion_r2013085555

packages/relay/src/lib/services/registryService/rpcMethodRegistryService.ts
quiet-node: I think we should enforce an architecture where each folder includes an index.ts file that exports all related components within that folder. This approach ensures a unified and clear import structure across the codebase.

Even if the folder currently contains only one file, adding an index.ts future-proofs the structure. As more components are added, they can be seamlessly exported through index.ts, preventing the need to update imports across multiple files. 

I think it's a pretty clean structure.

packages/relay/src/lib/services/registryService/rpcMethodRegistryService.ts
quiet-node: Good point! Refactored and turned this into a stand-alone function.

packages/server/tests/postman.json
quiet-node: Yep definitely a good point! Will be updated upon merging of https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3591

packages/server/tests/postman.json
quiet-node: Discussed offline, but for context:

Now that the JSON Error Object from the Relay package always includes a request ID for better traceability and debugging, this has been updated to use .contain() instead of .equal().

However, a better approach suggested during the discussion is to use regex to ensure the request ID is properly attached. Great suggestion!

packages/server/src/koaJsonRpc/index.ts
quiet-node: This was already defined as IJsonRpcRequest from the top in the rpcApp() method when parsed from the request context object. Why would we wanna change it to unknown? If it's the case, the entire pipeline would need to be updated from the top, which is beyond the scope of this PR.

packages/relay/tsconfig.json
quiet-node: Nice one! Thanks for pointing it out! Created this ticket to keep track https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3593

packages/relay/src/lib/dispatcher/index.ts
quiet-node: I did agree to this during our offline discussion but now looking back at it I think having a dedicated folder for the dispatcher would improve clarity and reduce clutter in the root directory, which already contains all the service implementation classes. 

We can discussed index.ts here https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3581#discussion_r2013097003

natanasow (commented):
At first sight, it looks good. Thanks for the crazy amount of work you have already done :rocket:.

packages/server/tests/postman.json
quiet-node: Again thanks for bringing this up! Since PR https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3591 got merged, this has been resolved

packages/relay/src/lib/eth.ts
```diff
@@ -1048,10 +1276,19 @@ export class EthImpl implements Eth {
...
+   *
    * @param {string} account The account to get the balance from
    * @param {string | null} blockNumberOrTagOrHash The block number or tag or hash to get the balance from
    * @param {RequestDetails} requestDetails The request details for logging and tracking
+   * @returns {Promise<string>} A promise that resolves to the balance of the account in hexadecimal format.
    */
+  @rpcMethod
+  @rpcParamValidationRules({
+    0: { type: ParamType.ADDRESS, required: true },
+    1: { type: ParamType.BLOCK_NUMBER_OR_HASH, required: true },
```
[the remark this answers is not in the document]
quiet-node: Ah great catch! Hmmm I have migrated these over from the Server in the previous implementation! If it was true from the previous implementation, I might have to keep it as-is to preserve the logic and avoid adding new behavior. Anywho thanks for pointing it out will definitely double check!

packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
```diff
@@ -0,0 +1,219 @@
...
+   */
+  public async dispatch(
+    rpcMethodName: string,
+    rpcMethodParams: any[] = [],
+    requestDetails: RequestDetails,
+  ): Promise<any | JsonRpcError> {
+    try {
+      /////////////////////////////// Pre-execution Phase ///////////////////////////////
+      const operationHandler = this.validateRpcMethod(rpcMethodName, rpcMethodParams, requestDetails);
+
```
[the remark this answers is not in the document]
quiet-node: Ahhh yeah good point hmmm I hadn’t thought about it that way, but it definitely makes sense and yeah it could be for a different PR. 

Since this has already a big one, for now I try to preserve the existing logic while integrating the dispatcher’s pre-processing for RPC method param validation and post-processing for error handling. Plus, response modification and debugging logs are already handled within each operation handler, which, I believe, keeps things cleaner.

A centralized response handler sounds appealing, but it would require numerous conditionals to ensure proper formatting per operation handler. Keeping it within each handler feels more intuitive.

Currently, the post-processor primarily handles errors and maps them to the appropriate JsonRpcError.

packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
```diff
@@ -0,0 +1,219 @@
...
+   * @returns Promise that resolves to the method execution result or a JsonRpcError instance
+   */
+  public async dispatch(
+    rpcMethodName: string,
+    rpcMethodParams: any[] = [],
+    requestDetails: RequestDetails,
+  ): Promise<any | JsonRpcError> {
+    try {
+      /////////////////////////////// Pre-execution Phase ///////////////////////////////
+      const operationHandler = this.validateRpcMethod(rpcMethodName, rpcMethodParams, requestDetails);
```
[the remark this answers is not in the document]
quiet-node: Ohhh this is golden! You’re right the name can be misleading since additional prechecks happen within each operation handler. Great point!

I like your naming ideas and also want something both generic and descriptive since it also checks for method existence, not just parameter validation. What do you think about `precheckRpcMethod`, or maybe `preVerifyRpcMethod` or `verifyRpcMethod` as they sort of verifying the RPC method for their validity and existence. Naming is hard 😅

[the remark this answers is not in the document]
natanasow: Same, naming is kinda the hardest part 😅. 

`precheckRpcMethod` sounds good to me, and I would go with it.

packages/relay/src/lib/web3.ts
acuarica: Sounds good to me.

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
acuarica: Hi @natanasow, why are you saying that it is required? Looking at the Ethereum JSON-RPC spec https://ethereum.github.io/execution-apis/api-documentation/, both `eth_getBalance` (and `eth_getCode` as mentioned below) have the block number or hash parameter as required.

packages/relay/src/lib/dispatcher/index.ts
acuarica: Will reply in the other discussion and resolve this.

packages/relay/tsconfig.json
acuarica: Awesome, thanks!

packages/relay/src/lib/services/registryService/rpcMethodRegistryService.ts
acuarica: IMO it just adds more overhead, by having a folder, `index.ts` and the actual code as opposed to the actual code.

> As more components are added, they can be seamlessly exported through index.ts, preventing the need to update imports across multiple files.

You can convert a single file module into a folder module with an `index.ts` without changing imports not requires. This is how Node resolution works, so I don't think it's a problem, that we need to future-proof the module structure.

packages/relay/src/lib/debug.ts
acuarica: Fair enough.

packages/relay/src/lib/net.ts
acuarica: Great stuff! thanks for the hard work!

packages/relay/src/lib/services/registryService/rpcMethodRegistryService.ts
quiet-node: Oh, I see! You’re suggesting moving the index.ts under the services/ folder not just within this folder I think I have read the original comment to quickly and initially had a different impression. 

That makes sense—we can place index.ts in services/ and export everything from there. Plus, no need to update other references since it will work as expected.

quiet-node: Removed index.ts for now since more services exist in the folder.

For the dispatcher, I think it makes sense to keep index.ts as it’s the root folder for it. Also, as mentioned, having a dedicated folder for the dispatcher keeps the root lib folder cleaner and less cluttered.

Let me know wyt

packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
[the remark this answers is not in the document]
quiet-node: Nice updated to precheckRpcMethod! Thanks for pointing it out!

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
quiet-node: As discussed with the team, we agreed to follow the Ethereum JSON-RPC spec, which requires block params for these endpoints. Therefore, no changes will be made.

packages/relay/src/lib/eth.ts
```diff
@@ -1227,13 +1464,23 @@ export class EthImpl implements Eth {
...
+   * @param {string} address - The Ethereum address of the contract.
+   * @param {string | null} blockNumber - The block number from which to retrieve the contract code.
+   * @param {RequestDetails} requestDetails - The details of the request for logging and tracking.
+   * @returns {Promise<string>} A promise that resolves to the contract code in hexadecimal format, or an empty hex string if not found.
+   * @throws {Error} Throws an error if the block number is invalid or if there is an issue retrieving the contract code.
+   */
+  @rpcMethod
+  @rpcParamValidationRules({
+    0: { type: ParamType.ADDRESS, required: true },
+    1: { type: ParamType.BLOCK_NUMBER_OR_HASH, required: true },
```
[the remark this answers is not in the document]
quiet-node: As discussed with the team, we agreed to follow the Ethereum JSON-RPC spec, which requires block params for these endpoints. Therefore, no changes will be made.

natanasow (dismissed):

packages/relay/src/utils.ts
```diff
@@ -160,4 +163,53 @@ export class Utils {
...
+   *
+   * @param method - The RPC method handler
+   * @param rpcParams - Parameters from the RPC request
+   * @param requestDetails - Request context information
+   * @returns Array of parameters arranged for the method
+   */
+  public static arrangeRpcParams(method: Function, rpcParams: any[] = [], requestDetails: RequestDetails): any[] {
+    const layout = method[RPC_PARAM_LAYOUT_KEY];
+
+    if (method.name === 'traceTransaction') {
```
simzzz, line 177: Is this the only special case? If not it might be a good idea to extract special case handlers to a map to avoid it getting too complex in the future.

packages/relay/src/lib/services/registryService/rpcMethodRegistryService.ts
acuarica: > For the dispatcher, I think it makes sense to keep index.ts as it’s the root folder for it.

To me this clutters things even more, having a folder and two files (where `index.ts` doesn't add value)  as opposed to a single file. In any case, let's move forward, resolving this.

packages/relay/src/lib/types/validation.ts
```diff
@@ -0,0 +1,114 @@
...
+   * Optional custom error message
+   */
+  errorMessage?: string;
+}
+
+/**
+ * Standard parameter types supported by the validator
+ * These are the basic types used in parameter validation schemas
+ */
+export enum ParamType {
```
acuarica, line 28: Not sure if we want to have a list of some/all used combinations. Maybe better to extract the types from `TYPES` directly

```
export type ParamType = (keyof typeof TYPES)[];
```

By doing this, we should also `|` to arrays instead.

packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
```diff
@@ -0,0 +1,219 @@
...
+   * @param requestDetails - Additional context about the request
+   * @returns Promise resolving to the result of the operation handler
+   */
+  private async processRpcMethod(
+    operationHandler: OperationHandler,
+    rpcMethodParams: any[],
+    requestDetails: RequestDetails,
+  ): Promise<any> {
+    // Rearrange the parameters as needed for the specific operation handler
+    const rearramgedRpcParams = Utils.arrangeRpcParams(operationHandler, rpcMethodParams, requestDetails);
```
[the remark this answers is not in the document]
quiet-node: Nice good catch! Updated!

packages/relay/src/utils.ts
quiet-node: Yeah, good point and yes, this is the only case that requires special rearrangement.

packages/relay/src/lib/types/validation.ts
quiet-node: Ah I see yeah it would make it cleaner and less repetitive.

However, I’d save this for a task in https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3592 which refactors the validator service holistically. 

For this PR, I want to preserve the logic and migrate the Validator component by component and avoid introducing new logic to the service itself..

acuarica: Awesome!

acuarica (approved):

natanasow (approved):

packages/relay/tests/lib/utils.spec.ts
```diff
@@ -190,17 +192,57 @@ describe('Utils', () => {
...
             const networkName = Utils.getNetworkNameByChainId();
             expect(networkName).to.equal(networkName);
           });
-        }
+        },
       );
     }
   });
+
+  describe('arrangeRpcParams', () => {
```
simzzz, line 211: Should we also add tests for `traceTransaction` here since it's the most complex case?

PR #3591, opened 2025-03-25
fix: fixed net_listening to return boolean `true`

### LINKED ISSUE #3590: net_listening returns ‘false’ as a string instead of a boolean
### Description:

net_listening is incorrectly returning a string instead of a boolean, with the value `'false'` instead of `true`. It should always return true when the client is actively listening for network connections, which is the expected behavior for Hedera.

### Solution:
Updated net_listening to return a boolean and set it to true.

quiet-node:
**Description**:
This PR updates `net_listening` to return a boolean `true` instead of the previous string `'false'`. It also updates all test coverage to align with the new logic.

**Related issue(s)**:

Fixes #3590

**Notes for reviewer**:

**Checklist**


acuarica (approved):
Nice stuff! I like that we return `true` now.

simzzz (approved):

PR #3594, opened 2025-03-26
feat: Improved MirrorNodeClient Error Handling and HTTP Error Mapping

### LINKED ISSUE #3441: [Relay White Noise Error] Enhance JSON RPC Relay Stability and Error Resilience [NEEDS MONITORING AND GLOBAL MAPPING ReView]
#### **Overview**  
The JSON RPC Relay service experiences intermittent errors due to issues from remote endpoints, primarily the Mirror Node. These "white noise errors" result in various failure responses, leading to degraded reliability. The goal of this epic is to analyze these errors, improve handling mechanisms, and enhance the overall stability of the Relay service.  

#### **Key Findings**  

a. `eth_call`  
- **Frequent eth_call Failures Due to Mirror Node 5xx Errors**  
  Many `eth_call` requests are rejected with a **500 Relay error** caused by **Mirror Node HTTP 5xx series errors**. The most common error codes are **500, 502, 503, and 504**, with **502 accounting for almost 99%** of all 5xx responses.  

- **Invalid Data Causing 501 Errors**  
  Some `eth_call` requests return **501 Not Implemented** from the Mirror Node due to invalid hex string parameters (e.g., `0x01ffc9a7d9b67a26...`). The Relay then retries with the Consensus Node, which also rejects the request, ultimately leading to a **500 error in the Relay**. Example `txRequestId`: `443ca4b5-5a19-459e-aa4e-59ee4420a49f`.  

- **Incorrect Mapping of 429 Errors**  
  `eth_call` requests receiving **429 Too Many Requests** from the Mirror Node are currently mapped to **409 Conflict** in the Relay. We need to evaluate whether the Relay should **return 429 instead of 409** for better alignment with rate-limiting semantics.  

b. `eth_getBalance`  
- **Frequent eth_getBalance Failures Due to Mirror Node 504 Errors**  
  Many `eth_getBalance` requests fail with a **504 Gateway Timeout** from the Mirror Node, indicating that the request is not completing within the expected timeframe. This suggests potential performance issues or bottlenecks in the Mirror Node's handling of balance queries. Further investigation is needed to determine if these failures are transient or indicative of a deeper infrastructure limitation.  

c. `eth_getBlockByNumber`  
- **Frequent eth_getBlockByNumber Failures Due to Mirror Node 504 Errors**  
  Similar to `eth_getBalance`, many `eth_getBlockByNumber` requests are failing with a **504 Gateway Timeout** error from the Mirror Node. This points to possible delays or timeouts in processing block data at the Mirror Node level. Investigating the root cause of these timeouts is crucial to understand if the issue is related to high load or other performance constraints within the node infrastructure.  

d. `eth_getLogs`  
- **Frequent eth_getLogs Failures Due to Mirror Node 502 Errors**  
  Many `eth_getLogs` requests are failing with a **502 Bad Gateway** error from the Mirror Node. This suggests an issue with the Mirror Node's ability to process and return log data reliably. The high occurrence of 502 errors indicates that the node may be experiencing upstream failures or instability when handling log queries. Further investigation is required to determine the root cause and potential mitigations.  

e. **Unexpected HTTP 567 Errors**  
- **Intermittent 567 Errors in the Relay**  
  Occasionally, the Relay logs show **HTTP 567 errors**, but these are not consistently reproducible. The cause of this status code is unclear, as it is not a standard HTTP error. It may indicate an upstream service issue, misconfigured proxy behavior, or an internal Relay anomaly. Further logging and investigation are needed to determine when and why these errors occur.  

#### **Core Issues**  
The key findings highlight four main issues affecting the Relay’s error handling and stability:  

1. **Inconsistent Error Code Mapping for 5xx Failures**  
   - Mirror Node errors (500, 502, 503, 504) are currently mapped to **500** in the Relay, losing specific failure context. A better mapping mechanism is needed to provide more precise error feedback to end clients.  

2. **Invalid Data Causing 501 Errors**  
   - Some `eth_call` requests return **501 Not Implemented** from the Mirror Node due to invalid hex string parameters. The Relay retries with the Consensus Node, which also fails, ultimately leading to a **500 error**. Improved validation is needed to detect invalid requests earlier.  

3. **Incorrect Mapping of 429 Errors**  
   - `eth_call` requests receiving **429 Too Many Requests** from the Mirror Node are incorrectly mapped to **409 Conflict** in the Relay. This needs correction to ensure proper rate-limiting responses.  

4. **Intermittent HTTP 567 Errors**  
   - The Relay occasionally logs **HTTP 567 errors**, but their cause is unclear. Further investigation is required to determine their origin and potential mitigation.  

#### **Objectives**  
This epic aims to enhance the Relay’s stability and resilience by addressing the core issues identified above. The main objectives include:  

- **Implementing better error code mapping** to ensure that Mirror Node failures provide accurate failure reasons to end clients.  
- **Improving request validation** to prevent invalid `eth_call` parameters from propagating and causing unnecessary retries.  
- **Correcting the mapping of rate-limiting errors** to align with HTTP semantics and improve client handling of 429 responses.  
- **Investigating and mitigating the cause of HTTP 567 errors** to eliminate unexplained anomalies in the Relay.  

### Alternatives

_No response_

### LINKED ISSUE #3442: [Relay White Noise Error] Improve Error Code Mapping for Mirror Node Failures
### Problem

The JSON RPC Relay currently maps all Mirror Node 5xx errors (500, 502, 503, 504) to a generic **500 Internal Server Error** when responding to clients. This lack of granular error mapping makes it difficult for clients to differentiate between transient failures, gateway issues, and actual server-side problems.  

From our investigation, the majority of failures stem from:  
- **502 Bad Gateway (99% of 5xx errors)** – Indicates an upstream Mirror Node failure.  
- **503 Service Unavailable** – Suggests that the Mirror Node is temporarily overloaded or undergoing maintenance.  
- **504 Gateway Timeout** – Indicates that the Mirror Node request took too long to process.  

By improving how these errors are mapped, the Relay can provide more informative responses, allowing clients to handle failures more effectively and reducing unnecessary retries.  

### Solution

Enhance the Relay's error handling by implementing a more precise mapping mechanism - return **502, 503, and 504** to clients instead of a generic **500**, preserving the original failure reason.  

### Alternatives

_No response_

### LINKED ISSUE #3444: [Relay White Noise Error] Correct Mapping of Rate-Limiting Errors (429 Handling)
### Problem

Currently, the Relay incorrectly maps **429 Too Many Requests** responses from the Mirror Node to **409 Conflict** before returning them to clients. This misalignment causes:  
- Clients being unaware of rate limits, potentially leading to excessive retries.  
- A lack of clarity regarding whether failures are due to conflicts or rate limits.  
- Potential inconsistencies in error handling across different RPC endpoints. 

### Solution

- Ensuring that **429 Too Many Requests** from the Mirror Node is correctly **returned as 429** to clients.  
- Define an internal code, and text that differentiated MN, RN throttle, and add it to the response
- Maintaining consistency in how rate-limiting responses are handled across different endpoints.  
- Improving client awareness of rate limits, reducing unnecessary load on the Relay and Mirror Node.  

### Alternatives

_No response_

### LINKED ISSUE #3445: Investigate and Address Intermittent HTTP 567 Errors
### Problem

The Relay intermittently logs **HTTP 567 errors**, but these occurrences are inconsistent and not tied to any specific request pattern. Since **567 is not a standard HTTP status code**, its presence suggests:  
- A misconfigured proxy or upstream service.  
- An internal anomaly in how the Relay handles certain responses.  
- Potential transient failures that are not properly logged or diagnosed.  

### Solution

Identify the root cause of the **HTTP 567** errors and determine the appropriate mitigation by:  
- **Analyzing logs** to detect patterns and common factors associated with these occurrences.  
- **Reviewing upstream dependencies** to determine whether an external service is generating the error.  
- **Ensuring proper handling or suppression** if 567 is an unintended response from internal mechanisms.  

### Alternatives

_No response_

### LINKED ISSUE #3470: [Relay White Noise Error] eth_getCode shows some 500 errors with timeout exceeded
### Problem

Occasionally, more than 5% of eth_getCode requests fail due to timeout exceeded.

### Solution

Investigate and resolve the underlying issue. This may also involve improving the mapping of internal errors to client errors.

### LINKED ISSUE #3519: [Feature] NEW HTTP Error mapping logic
Based on the study we elaborated, defining the error codes that we raise or receive, we define on fthe ollowing lines  a new http mapping proposal:

**Mapping**

MN

- Map 429 errors from MN to 429 errors on RPC relay, including a new 300XX code explanation, saying “MN throttled the request”
- Map 567 errors trom MN to 500 errors on RPC relay, including a new 300XX code explanation, saying “MN raised a 567  timeout”
- Map 502 errors trom MN to 500 errors on RPC relay, including a new 300XX code explanation, saying “MN raised a 502 error” ( 32602?)
- All unsupported methods should return 405
- All not yet implemented methods should retun 501 explicitly
- Not found methods 404
- 503 on MN, should raise, 500 on our side internal error, specifing 503 form MN
- All other MN error codes, should map to the same HTTP code as received, with internañ code + “MN error with HTTP code xxx”

Extras
- We should never raise a 503/502 due this is reserved to load balancer
- When rate limit error is generated from the relay rate limiting, response shoul 429, but with other message

### **Metrics**

Add to the metrics

- Counter for all internal rpc codes
    - To have a global overview
- Counter for all internal rpc codes + message
    - To have detailed view
        - number.message with_ instead spaces
- Counter For all the MN HTTP error codes
    - to have all MN HTTP codes and be able to distinguished errors, and alert on production support
- Counter per unsupported method
    - To have the visibility of called methods

Grafana

- Add internal rpc codes graph
- Add internal rpc codes.message graph
- Add MN error code graph
- Add graph for unsupported methods

Alerts

- Generate a new alert when 429 errors from MN are raised for more than 1% in the last 15 mins
- Generate a new alert when 429 are raised from Relay only ( not MN based)  for more than 15% in the last 1h

**Technical specs**

- Design a central point of error on mirronode client and json rpc, to detect and map errors
- Add all error codes, an descriptions to our repo documentation enabling users to have error descritpion
-

### LINKED ISSUE #3551: Epic: Implement Dispatcher Pattern for Relay Package and Improve Error Handling and Mapping

quiet-node:
### **Description**:
This PR improves error handling throughout the system, with a focus on Mirror Node client errors and their proper mapping to HTTP status codes. It introduces a standardized approach to error conversion and ensures consistent error responses across the API.

### Key Changes

- **Enhanced HTTP Error Mapping**
  - Introduced new logic to properly map JSON-RPC error codes to HTTP status codes
  - Implemented clear separation between internal error codes and external responses

- **Improved Mirror Node Error Handling**
  - Added `convertMirrorNodeError()` method to dispatcher for centralized error processing
  - Reworked `MirrorNodeClientError` class for better error classification and information
  - Preserved complete error context when propagating Mirror Node errors through RPC methods
  - Removed invalid status code 567 that was causing confusion in error handling

- **New Error Types**
  - Added `MIRROR_NODE_UPSTREAM_FAIL` JsonRpcError type to specifically handle Mirror Node failures
  - Created appropriate error mapping for various Mirror Node error conditions (rate limits, server errors, etc.)

Fixes #3519
Fixes #3470
Fixes #3444
Fixes #3442
Fixes #3445 
Fixes #3441 (EPIC)
Fixes #3551 (EPIC)

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
```diff
@@ -132,7 +132,16 @@ export class RpcMethodDispatcher {
     const rearrangedParams = Utils.arrangeRpcParams(operationHandler, rpcMethodParams, requestDetails);
 
     // Execute the operation handler with the rearranged parameters
-    return await operationHandler(...rearrangedParams);
+    const result = await operationHandler(...rearrangedParams);
+
+    // *Note: In some cases, the operation handler may return an exception instead of throwing.
+    // To ensure proper and centralized error handling in the dispatcher, preserve and rethrow the error,
+    // regardless of whether the operation handler returns or throws it.
+    if (result instanceof JsonRpcError || result instanceof SDKClientError || result instanceof MirrorNodeClientError) {
```
acuarica, line 140: If we make `JsonRpcError extends Error` then this can be just `result instanceof Error`. In any case, I guess `JsonRpcError` should extend Error anyway.

quiet-node: Ah, good point—it would be cleaner. However, I attempted to quickly extend JsonRpcError from the Error object, but ran into some caveats and pitfalls. The findings was noted down in this ticket https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3617.

Given the issues and the possibility of more unexpected behaviors—I’d prefer to address this in a separate PR for a more comprehensive solution.

packages/relay/src/lib/errors/MirrorNodeClientError.ts
```diff
@@ -41,43 +68,55 @@ export class MirrorNodeClientError extends Error {
...
+  public isFailInvalid() {
     return this.message === 'FAIL_INVALID';
   }
 
-  isInvalidTransaction() {
+  public isInvalidTransaction() {
     return this.message === 'INVALID_TRANSACTION';
   }
+
+  public isInternalServerError() {
```
natanasow, line 103: nit: missing return type

packages/server/src/koaJsonRpc/lib/httpErrorMapper.ts
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: Apache-2.0
+
+// Define constants for frequently used values
```
natanasow, line 3: q: we could wrap the entire file in a class, making it more OOP, right?

packages/relay/src/lib/errors/MirrorNodeClientError.ts
quiet-node: Oh, indeed! Just updated MirrorNodeClientError with TSDoc and return types throughout. Thanks much for pointing it out!

packages/server/src/koaJsonRpc/lib/httpErrorMapper.ts
quiet-node: Ah, good point. However, I don’t think it’s necessary in this case. The `translateRpcErrorToHttpStatus()` function does one thing well without relying on module states. It’s a pure function that takes inputs and returns outputs. It also doesn’t require any special initialization—no setup or teardown is needed for a constructor. Additionally, no state needs to be maintained between calls. Moreover, constants like `HTTP_STATUS`, `ERROR_CODE_MAP`, and `MIRROR_NODE_ERROR_MAP` remain unchanged and don’t really need encapsulation.

That said, I think `httpErrorMapper` would be better suited to a JavaScript functional approach.

natanasow (approved):

packages/server/src/koaJsonRpc/lib/httpErrorMapper.ts
```diff
@@ -0,0 +1,63 @@
...
+  [-32601]: HTTP_STATUS.BAD_REQUEST, // Method not found
+  [-32605]: HTTP_STATUS.TOO_MANY_REQUESTS, // Rate limit exceeded
+};
+
+// Map Mirror Node error data to HTTP status codes
+// - 429 from the Mirror Node maps to HTTP 429
+// - 501 from the Mirror Node maps to HTTP 501
+// - Any other error data from the Mirror Node will be mapped to HTTP 500 by default
+const MIRROR_NODE_ERROR_MAP: Record<string, number> = {
+  '429': HTTP_STATUS.TOO_MANY_REQUESTS,
```
acuarica, line 27: I don't see the point of having a key that maps to that same value.

packages/server/src/koaJsonRpc/index.ts
```diff
@@ -104,17 +104,21 @@ export default class KoaJsonRpc {
...
+      const httpStatusCodeAndMessage = translateRpcErrorToHttpStatus(
+        errorOrResult.code,
+        errorOrResult.message,
+        this.getRequestDetails().formattedRequestId,
+        errorOrResult.data,
+      );
+
       ctx.status = httpStatusCodeAndMessage.statusCode;
-      ctx.state.status = `${ctx.status} (${httpStatusCodeAndMessage.StatusName})`;
+      ctx.state.status = `${ctx.status} (${httpStatusCodeAndMessage.statusErrorMessage})`;
```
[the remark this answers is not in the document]
quiet-node: > what is ctx.state.status used for?

`ctx.state.status` is used for debugging purposes, as seen here https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3594/files#diff-d6749baa2bf74a3f75f55f27e1f3c89512ad671f6dbec8db7e42ca6fa3377c38R66-R68. It was previously misused as ctx.status, resulting in logs missing descriptive messages.

Before the fix:
```
[Request ID: d2569344-9e98-4e3d-858e-f680bea694e4] [POST]: eth_blockNumber 501 18 ms
```

After the fix:
```
[Request ID: d2569344-9e98-4e3d-858e-f680bea694e4] [POST]: eth_blockNumber 501 (Mirror node upstream failure: statusCode=501, message=Not Supported) 18 ms
```
This makes it easier to diagnose issues from the logs.


> Essentially I was expecting translateRpcErrorToHttpStatus to just receive errorOrResult.code, not the whole error object
Right your expectation makes sense, but during development I found room for improving clarity by using the full error object, including `errorMessage` and `errorData`. The previous error message setup was done manually and I found it wasn’t scalable, so using `errorMessage` directly provides better context and more autonomous. `errorData` is optional and mainly used for MIRROR_NODE_UPSTREAM_FAILURE. 

Well now that you mention it, the parameter can just be the error object itself—I’ll update it.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -357,19 +370,21 @@ export class MirrorNodeClient {
       return response.data;
     } catch (error: any) {
       const ms = Date.now() - start;
+
+      // Calculate effective status code
       const effectiveStatusCode =
-        error.response?.status ||
-        MirrorNodeClientError.ErrorCodes[error.code] ||
-        MirrorNodeClient.unknownServerErrorHttpStatusCode;
+        error.response?.status || MirrorNodeClientError.HttpStatusResponses[error.code]?.statusCode || 500; // Use standard 500 as fallback
```
Ferparishuertas, line 376: How will we distinguish a real 500 from the fallback?

packages/server/src/koaJsonRpc/lib/httpErrorMapper.ts
quiet-node: Yeah good observation! It may seem redundant, but using a map makes the code cleaner. Some MN status codes don’t have direct mappings, so based on user stories, we map:
- MN 429 → HTTP 429
- MN 501 → HTTP 501
- MN 404 → HTTP 400
- All others from MN (e.g. 502, 503, 504, etc.) default to HTTP 500

Without a map, we’d need explicit status checks like something like below:

``` typescript
if (jsonRpcError.code === -32020 && jsonRpcError.data) {
    let statusCode = HTTP_STATUS.INTERNAL_SERVER_ERROR;
    switch (jsonRpcError.data) {
      case '404':
        statusCode = HTTP_STATUS.BAD_REQUEST;
        break;
      case '429':
        statusCode = HTTP_STATUS.TOO_MANY_REQUESTS;
        break;
      case '501':
        statusCode = HTTP_STATUS.NOT_IMPLEMENTED;
        break;
    }
    return { statusCode, statusErrorMessage };
}
```

While it functions fine, it’s a bit cluttered compared to the map solution:

```typescript
if (jsonRpcError.code === -32020 && jsonRpcError.data) {
    return {
      statusCode: MIRROR_NODE_ERROR_MAP[jsonRpcError.data] || HTTP_STATUS.INTERNAL_SERVER_ERROR,
      statusErrorMessage,
    };
}
```

Though the map solution can be cleaner but your point is valid. So let me know if you have a suggestion on this.

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: Okay restored the custom MirrorNodeClient.unknownServerErrorHttpStatusCode. Thanks for pointing it out.

packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
```diff
@@ -216,4 +226,35 @@ export class RpcMethodDispatcher {
...
+
+  /**
+   * Maps a MirrorNodeClientError to a standardized JsonRpcError.
+   *
+   *
+   * @param error - The MirrorNodeClientError instance to map to a JsonRpcError.
+   * @returns A JsonRpcError representing the mapped Mirror Node error, including
+   *          the appropriate error code and message.
+   */
+  private mapMirrorNodeError(error: MirrorNodeClientError): JsonRpcError {
```
simzzz, line 238: Wouldn't it be better for the `MirrorNodeClientError` class to be responsible for its own transformation and have this method there?

packages/relay/src/lib/eth.ts
```diff
@@ -2422,10 +2428,6 @@ export class EthImpl implements Eth {
           return EthImpl.emptyHex;
         }
 
-        if (e.isRateLimit()) {
```
konstantinabl, line 2425: @quiet-node so why are we deleting this?

packages/relay/src/lib/eth.ts
```diff
@@ -2449,9 +2451,10 @@ export class EthImpl implements Eth {
           }
           return await this.callConsensusNode(call, gas, requestDetails);
         }
-      }
 
-      this.logger.error(e, `${requestIdPrefix} Failed to successfully submit eth_call`);
+        // for any other Mirror Node upstream server errors (429, 500, 502, 503, 504, etc.), preserve the original error and re-throw to the upper layer
+        throw e;
```
konstantinabl, line 2456: Probably Im missing something but the comment suggest here we throw any other MN upstream server error, arent we covering all of them in the MirrorNodeClientError class?

packages/relay/tests/lib/eth/eth_getBlockByHash.spec.ts
```diff
@@ -288,11 +294,15 @@ describe('@ethGetBlockByHash using MirrorNode', async function () {
         `contracts/results?timestamp=gte:${randomBlock.timestamp.from}&timestamp=lte:${randomBlock.timestamp.to}&limit=100&order=asc`,
       )
       .abortRequest();
-    await RelayAssertions.assertRejection(predefined.INTERNAL_ERROR(), ethImpl.getBlockByHash, false, ethImpl, [
-      BLOCK_HASH,
-      false,
-      requestDetails,
-    ]);
+
+    await expect(ethImpl.getBlockByHash(BLOCK_HASH, true, requestDetails))
```
konstantinabl, line 298: So why not use the assertRejection?

konstantinabl (commented):
Just a few questions

packages/server/src/koaJsonRpc/lib/httpErrorMapper.ts
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { JsonRpcError } from '@hashgraph/json-rpc-relay/dist';
+
+import { JsonRpcError as JsonRpcErrorServer } from './RpcError';
+
+// Define constants for frequently used values
+const HTTP_STATUS = {
```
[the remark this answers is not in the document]
quiet-node: Yeah, enums are great but not sure if switching to one would offer any significant practical benefits here. Object literals are also widely used and more idiomatic in JavaScript. Plus, the code already follows this pattern, and consistency is valuable as well. Let me know if you see significant benefit. Open to update.

packages/relay/src/lib/eth.ts
quiet-node: Based on the user story, we want to map MN 429 to HTTP 429 with a more descriptive message. We can see that the removed component was technically mapping it to `predefined.IP_RATE_LIMIT_EXCEEDED`, which could misrepresent the original domain (i.e. Mirror Node or Relay). That’s why I removed it and threw the error instead, allowing it to be caught and properly mapped to a corresponding JsonRpcError in the dispatcher’s error handling.

packages/relay/src/lib/eth.ts
quiet-node: Not sure how the question relates to this code but yes all cases are already covered in the MirrorNodeClientError class, but this code simply preserves and re-throws the error so it can be caught and translated by the dispatcher. 

Let me know if this answers your question

natanasow (dismissed):

konstantinabl (approved):

packages/server/src/koaJsonRpc/lib/httpErrorMapper.ts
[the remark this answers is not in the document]
simzzz: Just a preference, we can keep it as an object. I'm resolving

simzzz (dismissed):

packages/server/src/koaJsonRpc/index.ts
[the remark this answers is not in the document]
quiet-node: This has been updated to refactor the request ID stripping into a separate component. Resolving.

simzzz (approved):

packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
```diff
@@ -163,14 +172,17 @@ export class RpcMethodDispatcher {
...
+    if (error instanceof SDKClientError && error.isGrpcTimeout()) {
       return this.createJsonRpcError(predefined.REQUEST_TIMEOUT, requestDetails.requestId);
     }
 
-    // @TODO: handle SDKClientError by mapping to the correct JsonRpcError
-    if (error instanceof SDKClientError && error.isGrpcTimeout()) {
-      return this.createJsonRpcError(predefined.REQUEST_TIMEOUT, requestDetails.requestId);
+    // Handle MirrorNodeClientError by mapping to the correct JsonRpcError
+    if (error instanceof MirrorNodeClientError) {
+      return this.createJsonRpcError(
```
acuarica, line 182: nit: should we also make `createJsonRpcError` static?

packages/relay/tests/lib/dispatcher/rpcMethodDispatcher.spec.ts
```diff
@@ -185,12 +185,13 @@ describe('RpcMethodDispatcher', () => {
...
 
-    it('should directly return JsonRpcError if the handler returns a JsonRpcError', async () => {
+    it('should preserve and rethrow exception even when the handler returns an exception', async () => {
       const jsonRpcError = new JsonRpcError({ code: -32000, message: 'Handler error' });
       operationHandler.returns(jsonRpcError);
 
-      const result = await (dispatcher as any).processRpcMethod(operationHandler, TEST_PARAMS, TEST_REQUEST_DETAILS);
-      expect(result).to.equal(jsonRpcError);
+      await expect(
+        (dispatcher as any).processRpcMethod(operationHandler, TEST_PARAMS, TEST_REQUEST_DETAILS),
```
acuarica, line 193: nit: why do we need the `as any`? can we use its type definition here?

acuarica (approved):
lgtm, left some comments/questions in case we want to address them.

packages/relay/tests/lib/dispatcher/rpcMethodDispatcher.spec.ts
quiet-node: Yes this is for the purposes of accessing private methods for testing purposes which is the same as one of the discussions we have earlier in the subscription and poller service PR. Anywho created a ticket to address this separately https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3659

packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
quiet-node: Technical debt ticket is created here https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3658

PR #3595, opened 2025-03-26
feat: create `FeeService` class

### LINKED ISSUE #3587: Refactor: Create FeeService class
### Problem

Currently, all of the fee-related methods logic is inside eth.ts.

### Solution

Create a FeeService class.

### Alternatives

_No response_

natanasow:
**Description**:

Currently, all of the fee-related methods logic is inside eth.ts.

**Solution**:

Create a FeeService class.

**Related issue(s)**:

Fixes #3587 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -241,7 +232,7 @@ export class EthImpl implements Eth {
     logger: Logger,
     chain: string,
     registry: Registry,
-    cacheService: CacheService,
+    cacheService: CacheService
```
konstantinabl, line 235: How did this get removed, is it es-lint, I think its set to not delete trailing commas?

packages/relay/src/lib/eth.ts
```diff
@@ -282,6 +278,10 @@ export class EthImpl implements Eth {
     return this.debugServiceImpl;
   }
 
+  feeService(): IFeeService {
+    return this.feeServiceImpl;
```
konstantinabl, line 282: I think the name should be just feeService, we don't have anything else than an implementation of the service, so why add Impl in the end?

packages/relay/src/lib/services/factories/transactionFactory.ts
```diff
@@ -27,8 +27,10 @@ export class TransactionFactory {
               ? '0x0'
               : prepend0x(trimPrecedingZeros(fields.maxFeePerGas)),
         }); // eip 1559 fields
-      default:
+      case null:
```
konstantinabl, line 30: In what case would this be null? I am not sure it makes sense to pass null as a type?

packages/relay/src/lib/services/factories/transactionFactory.ts
```diff
@@ -27,8 +27,10 @@ export class TransactionFactory {
...
         return new Transaction(fields); //hapi
     }
+
+    return null;
```
konstantinabl, line 34: And here maybe im missing it, in what cases would we want to return null?

konstantinabl (changes requested):
Generally, looks good, left some comments

packages/relay/src/lib/eth.ts
natanasow: Fixed.

packages/relay/src/lib/eth.ts
natanasow: Just sticking to the current implementations convention https://github.com/hiero-ledger/hiero-json-rpc-relay/blame/poc-eth-service-apporach/packages/relay/src/lib/eth.ts#L281.

packages/relay/src/lib/services/factories/transactionFactory.ts
natanasow: With `null` we handle HAPI transactions.

packages/relay/src/lib/services/factories/transactionFactory.ts
natanasow: If a new type of transaction is introduced and we have not yet handled it in the relay, we should return null (e.g., blob or eip 7702 tx).

konstantinabl (approved):

PR #3605, opened 2025-03-27
fix: added safe guard for rewardPercentiles size limit

### LINKED ISSUE #3527: Optimize memory usage for feeHistory reward field
### Description
In the current implementation of the feeHistory method in our JSON-RPC Relay, the reward field is populated with a nested array determined by the length of the rewardPercentiles parameter. Each inner array is filled with a fixed value (0x0), which avoids the need for additional per-item analysis. However, when the rewardPercentiles array is large, the resulting reward field can grow significantly in size, leading to increased memory usage. To enhance memory efficiency and improve overall performance, we can introduce a reasonable limit on the size of the rewardPercentiles array.

### Solution
Implement a reasonable cap on the rewardPercentiles array length to optimize memory usage and improve response times. This adjustment aligns with typical use patterns and enhances overall system efficiency.

quiet-node:
**Description**:
This PR adds a guard to the feeHistory method to limit the size of the rewardPercentiles array. The maximum allowed size is set to 100 in a new constant, FEE_HISTORY_REWARD_PERCENTILES_MAX_SIZE, preventing excessive memory usage when the rewardPercentiles parameter contains a large number of entries.

**Related issue(s)**:

Fixes #3527 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/eth.ts
```diff
@@ -296,13 +293,21 @@ export class EthImpl implements Eth {
...
+    const maxRewardPercentilesSize = constants.FEE_HISTORY_REWARD_PERCENTILES_MAX_SIZE;
 
     if (this.logger.isLevelEnabled('trace')) {
       this.logger.trace(
         `${requestIdPrefix} feeHistory(blockCount=${blockCount}, newestBlock=${newestBlock}, rewardPercentiles=${rewardPercentiles})`,
       );
     }
 
+    if (rewardPercentiles && rewardPercentiles.length > maxRewardPercentilesSize) {
+      return predefined.INVALID_PARAMETER(
```
Ferparishuertas, line 305: When merges with requestid, I think we need to add the requestid

Ferparishuertas (approved):
LGTM
Please review the requestId comment

packages/relay/src/lib/eth.ts
quiet-node: solid concern but yeah the dispatcher's centralized error handling should take care of this task and automatically attaches requestID to any JsonRpcError https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/main/packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts#L206-L218.

natanasow (approved):

acuarica (approved):

PR #3626, opened 2025-04-04
feat: added IS_MODULARIZED axios header for MirrorNodeClient

### LINKED ISSUE #3623: Enforce Non Modularization FLOW on MN
Mirrror node has enabled modularized flow 10% in previewnet. Mirror node was lacking of all the modularization done at consensus node and was missing some gas estimations calcs, generating some gas estimate issues.
Perhaps gas estimation is not fully the same as on consensus node, it is 95% equal now.

As we added an extra 10% of margin on estimation, to avoid any possible issue, if we send traffic to any new modulira MNzation node (10% of node but will be progressively going to 100%), we will overcharge the user.

The idea is to set the following httoo header Is-Modularized: false when invoking /api/v1/contracts/call to enforce to old flow until mirror node has tested fully the changes.

Action
Change the header on a new version and deploy it to all environments.
Once MN has tested correctly on all envs, and set to 100% we will set it to true

quiet-node:
### Description

This PR adds support for the `Is-Modularized` header in Mirror Node requests through a new configuration flag, `USE_MIRROR_NODE_MODULARIZED_SERVICES`, allowing controlled routing of traffic to modularized services within the Mirror Node.

### Context

Currently, only a small percentage of Mirror Node traffic on Previewnet is routed to modularized services, with a gradual increase planned until full adoption across all networks.

For now, the Relay explicitly sets this to `false` to ensure traffic follows the non-modularized flow. Once the Mirror Node fully transitions, this will be updated to `true` as a default value.

### Changes
- Added `USE_MIRROR_NODE_MODULARIZED_SERVICES` configuration flag (default: false)
- Modified request headers to include `Is-Modularized` with appropriate string conversion
- Updated documentation to reflect the new configuration option

**Related issue(s)**:

Fixes #3623 

**Checklist**


Ferparishuertas (approved):
LGTM

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -307,6 +314,7 @@ export class MirrorNodeClient {
       const axiosRequestConfig: AxiosRequestConfig = {
         headers: {
           [MirrorNodeClient.REQUESTID_LABEL]: requestDetails.requestId,
+          [MirrorNodeClient.IS_MODULARIZED]: ConfigService.get('USE_MIRROR_NODE_MODULARIZED_SERVICES').toString(),
```
natanasow, line 317: q: do we need that `.toString()`? Is it type hinted in the `AxiosRequestConfig` type?

natanasow (approved):
LGTM!

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: Yeah, I did some research and found that technically it accepts any value—Axios will internally convert it to a string, which aligns with the HTTP protocol for headers. However, I thought it’s a good practice not to rely on implicit conversion, so I chose to convert it to a string explicitly to ensure clear alignment with the protocol. Good Qs tho.

natanasow: Got it, thank you for the explanation!

Issue #3622, opened 2025-04-03
Refactor: Create TransactionReceipt Factory


konstantinabl:
Currently, we have methods, where we create Receipts in the method itself manually. Its best to have a factory that creates the needed objects

Issue #3570, opened 2025-03-20
Refactor: Create TransactionService class


konstantinabl:
<img width="645" alt="Image" src="https://github.com/user-attachments/assets/78f2d7dd-adae-45e2-a8c9-bdd1449dcd35" />

Implement the class as shown in the diagram above

PR #3646, opened 2025-04-07
refactor: create TransactionService class

konstantinabl:
**Description**:

This PR refactors the JSON-RPC relay by extracting transaction-related functionality from the eth.ts file into a dedicated TransactionService class and moving shared code into a CommonService. This improves code organization, testability, and maintainability.

**Changes**:

- Created new TransactionService to handle all transaction-related RPC methods
- Moved methods used only from the transaction related methods to the TransactionService - sendRawTransactionProcessor, prune0x, parseRawTxAndPrecheck, sendRawTransactionErrorHandler
- Moved common static variables into a CommonService class
- Changed the way we use ethExecutionsCounter - moved it to the metric service as an event listener, similar to other metrics we have
-  Created TransactionReceiptFactory to standardize creation of transaction receipts
- Added JSDoc where missing

**Moved methods to TransactionService**

- getTransactionByBlockHashAndIndex
- getTransactionByBlockNumberAndIndex
- getTransactionByHash
- getTransactionReceipt
- sendRawTransaction
- sendTransaction
- signTransaction
- sign

**Fixes**:

#3570
#3622


packages/relay/src/lib/services/factories/ITransactionReceiptFactoryParams.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Parameters specific to creating a regular transaction receipt from mirror node data
+ */
+interface IRegularTransactionReceiptParams {
+  effectiveGas: string;
+  from: string;
+  logs: Log[];
+  receiptResponse: any;
+  to: string;
+}
+
```
quiet-node, line 22: Regarding interface naming, I understand we currently use the `I` prefix convention, but I've been reading around and believe it’s largely considered outdated in the modern TypeScript community. Some suggest avoiding the `I` prefix unless the interface is meant to be implemented by a class. Some suggest to completely avoid it. Even official docs from  `typescriptlang.org` don’t follow this convention anymore. Same goes for file naming.

Just a suggestion—but let’s avoid introducing more of this pattern. We could follow the blockFactory approach where BlockFactoryParams is declared in the same file and without `I` prefix. If we prefer keeping interfaces separate for clarity, then let’s move them (including blockFactory’s) to a shared types/ folder for consistency. We shouldn’t have blockFactory structured one way and transactionReceiptFactory another but let's follow a consistent approach.

Let me know what you think. Open to discuss.

Some references:
- https://www.typescriptlang.org/static/TypeScript%20Interfaces-34f1ad12132fb463bd1dfe5b85c5b2e6.png
- https://www.typescriptlang.org/docs/handbook/2/objects.html
- https://makecode.com/extensions/naming-conventions
- https://dev.to/mscamargo/why-you-should-avoid-using-the-i-prefix-for-interfaces-in-typescript-43gd
- https://www.shaneyu.com/blog/post/modern-typescript-naming/

packages/relay/src/lib/services/transactionService/TransactionService.ts
```diff
@@ -0,0 +1,859 @@
...
+    }
+    return predefined.UNSUPPORTED_METHOD;
+  }
+
+  /**
+   * Creates a transaction object from a log entry
+   * @param log The log entry containing transaction data
+   * @returns {Transaction1559 | null} A Transaction1559 object or null if creation fails
+   */
+  private createTransactionFromLog(log: Log): Transaction1559 | null {
```
quiet-node, line 436: Would it make sense to move such function to the factory?

packages/relay/src/lib/services/transactionService/TransactionService.ts
```diff
@@ -0,0 +1,859 @@
...
+    });
+  }
+
+  /**
+   * Gets the current gas price for a block
+   * @param blockHash The block hash
+   * @param requestDetails The request details for logging and tracking
+   * @returns {Promise<string>} A promise that resolves to the gas price as a hex string
+   */
+  private async getCurrentGasPriceForBlock(blockHash: string, requestDetails: RequestDetails): Promise<string> {
```
quiet-node, line 489: Hmmm I think we should think about migrating these private methods, e.g. getCurrentGasPriceForBlock(), getCurrentNetworkExchangeRateInCents(), etc., to a different module as to me these are not necessarily related to `transaction` per se but more of helper or utility function. wdyt?

packages/relay/src/lib/services/index.ts
```diff
@@ -2,9 +2,23 @@
...
+  IBlockService,
+  ICommonService,
+  CommonService,
+  IFilterService,
+  FilterService,
+  IContractService,
+  ContractService,
+  TransactionService,
+  MetricService,
+};
```
quiet-node, line 24: Can we do something like below for cleaner code?

```suggestion
export * from './blockService/BlockService'
export * from './blockService/IBlockService'
export * from './contractService/ContractService'
export * from './contractService/IContractService'
export * from './ethService/ethCommonService/CommonService'
export * from './ethService/ethCommonService/ICommonService'
export * from './ethService/ethFilterService'
export * from './ethService/ethFilterService/IFilterService'
```

packages/relay/src/lib/eth.ts
```diff
@@ -181,6 +161,10 @@ export class EthImpl implements Eth {
    */
   private readonly feeService: FeeService;
 
+  private readonly transactionService: TransactionService;
+
+  private readonly eventEmitter: EventEmitter;
```
quiet-node, line 166: Can't mark the lines but I can see private properties like precheck, hapiService, cacheService, etc. are not being used anymore. Should consider removing them

packages/relay/src/lib/eth.ts
```diff
@@ -209,34 +193,30 @@ export class EthImpl implements Eth {
     chain: string,
     registry: Registry,
```
quiet-node, line 194: unused var warning

packages/relay/src/lib/eth.ts
```diff
@@ -1,39 +1,32 @@
 // SPDX-License-Identifier: Apache-2.0
```
quiet-node, line 1: Nice one! This file looks much cleaner and more organized now. However, there are still several unused variables and methods remaining from functions that were migrated elsewhere.

Also, I think the eth.ts file is supposed to contain only functions annotated with @rpcMethod, i.e., actual RPC methods. Private or helper functions should be moved to appropriate modules like helpers or utils.

packages/relay/src/lib/eth.ts
```diff
@@ -354,14 +334,13 @@ export class EthImpl implements Eth {
...
-        )
-        .inc();
+      this.eventEmitter.emit(constants.EVENTS.ETH_EXECUTION, {
+        method: EthImpl.ethEstimateGas,
+        functionSelector: callData!.substring(0, constants.FUNCTION_SELECTOR_CHAR_LENGTH),
+        from: transaction.from || '',
+        to: transaction.to || '',
+        requestDetails: requestDetails,
+      });
     }
```
quiet-node, line 344: Nice, I like this centralized solution, but it does make me a bit angsty. Let’s ensure the metrics work as expected and use the correct keyword for the metric names. Otherwise, Grafana might fail to collect them.

packages/relay/src/logsBloomUtils.ts
```diff
@@ -1,24 +1,25 @@
...
 import { keccak256 } from 'ethers';
+
 import { prepend0x, strip0x } from './formatters';
-import { EthImpl } from './lib/eth';
 
 export class LogsBloomUtils {
   public static readonly BYTE_SIZE = 256;
   public static readonly MASK = 0x7ff;
 
+  private static readonly EMPTY_BLOOM = '0x' + '0'.repeat(512);
```
quiet-node, line 10: I noticed EMPTY_BLOOM is used across multiple modules like BlockService, BlockFactory, TransactionService, etc.. How about we refactor it into the centralized constants.ts file and reuse it throughout the codebase instead of redeclaring it in multiple places.

quiet-node (commented):
Great work! Much cleaner and more organized now. Left some Qs and suggestions around.

packages/relay/src/lib/services/transactionService/TransactionService.ts
```diff
@@ -0,0 +1,859 @@
...
+  private readonly mirrorNodeClient: MirrorNodeClient;
+
+  /**
+   * An instance of EventEmitter used for emitting and handling events within the class.
+   *
+   * @private
+   * @readonly
+   * @type {EventEmitter}
+   */
+  private readonly eventEmitter: EventEmitter;
```
simzzz, line 74: nit: Properties are not sorted alphabetically

packages/relay/src/lib/services/transactionService/TransactionService.ts
```diff
@@ -0,0 +1,859 @@
...
+  }
+
+  /**
+   * Parses a raw transaction and performs prechecks
+   * @param transaction The raw transaction string
+   * @param networkGasPriceInWeiBars The current network gas price in wei bars
+   * @param requestDetails The request details for logging and tracking
+   * @returns {Promise<EthersTransaction>} A promise that resolves to the parsed Ethereum transaction
+   */
+  async parseRawTxAndPrecheck(
```
simzzz, line 569: Missing `private`

packages/relay/src/lib/eth.ts
```diff
@@ -1797,30 +1201,6 @@ export class EthImpl implements Eth {
     return input.startsWith(CommonService.emptyHex) ? input.substring(2) : input;
```
simzzz, line 1201: `prune0x` and `getTransactionCountFromBlockResponse` are no longer used here and can be removed

packages/relay/src/lib/services/transactionService/TransactionService.ts
```diff
@@ -0,0 +1,859 @@
...
+        requestDetails,
+      );
+
+      // no tx found
+      if (!syntheticLogs.length) {
+        if (this.logger.isLevelEnabled('trace')) {
+          this.logger.trace(`${requestIdPrefix} no tx for ${hash}`);
+        }
+        return null;
+      }
```
simzzz, line 215: This logic is repeated here and in `getTransactionReceipt`, perhaps it could be abstracted?

packages/relay/src/lib/services/transactionService/TransactionService.ts
```diff
@@ -0,0 +1,859 @@
...
+    });
+  }
+
+  /**
+   * Gets a transaction receipt by hash
+   * @param hash The transaction hash
+   * @param requestDetails The request details for logging and tracking
+   * @returns {Promise<ITransactionReceipt | null>} A promise that resolves to a transaction receipt or null if not found
+   */
+  async getTransactionReceipt(hash: string, requestDetails: RequestDetails): Promise<ITransactionReceipt | null> {
```
simzzz, line 239: Perhaps we can split up this method in some way, since it's quite complex. I'm thinking this could be a good start?

```typescript
//...
  let receipt: ITransactionReceipt | null;

  if (receiptResponse === null || receiptResponse.hash === undefined) {
    receipt = await this.handleSyntheticTransactionReceipt(hash, requestDetails);
  } else {
    receipt = await this.handleRegularTransactionReceipt(receiptResponse, requestDetails);
  }
// ...rest of logic
```

packages/relay/src/lib/services/index.ts
konstantinabl: yeah, will change it

packages/relay/src/lib/eth.ts
konstantinabl: fixed

packages/relay/src/lib/eth.ts
konstantinabl: fixed

packages/relay/src/lib/services/transactionService/TransactionService.ts
konstantinabl: fixed

packages/relay/src/lib/services/transactionService/TransactionService.ts
```diff
@@ -0,0 +1,859 @@
...
+      throw this.common.genericErrorHandler(
+        error,
+        `${requestIdPrefix} Failed to retrieve contract result for blockHash ${blockHash} and index=${transactionIndex}`,
+      );
+    }
+  }
+
+  /**
+   * Gets a transaction by block number and transaction index
+   * @param blockNumber The block number
```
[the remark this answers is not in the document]
konstantinabl: fixed

packages/relay/src/lib/services/transactionService/TransactionService.ts
konstantinabl: fixed

packages/relay/src/lib/services/transactionService/TransactionService.ts
konstantinabl: Moved it : )

packages/relay/src/lib/eth.ts
konstantinabl: Not sure what exactly do you think should be changed?

packages/relay/src/lib/services/transactionService/TransactionService.ts
konstantinabl: thanks, great suggestion! fixed

packages/relay/src/lib/services/factories/ITransactionReceiptFactoryParams.ts
konstantinabl: Yeah, you are right. I moved the interfaces to the transactionReceiptFactory file. Will create an issue to change all interface names

packages/relay/src/lib/services/factories/ITransactionReceiptFactoryParams.ts
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
konstantinabl: Fixed - used 1st approach

packages/relay/src/lib/services/transactionService/TransactionService.ts
konstantinabl: Honestly, i think its a very small scope and the strings are slightly different, so creating a function to asbtract it will complicate the code in my opinion.

simzzz (approved):

packages/relay/src/lib/eth.ts
quiet-node: Oh it's not about changing things but just make sure that we don't accidentally rename the metrics name or labels or arguments. Make sure we preserve them so no issue on Grafana.

packages/relay/src/logsBloomUtils.ts
acuarica: Agree with Kosi, we should move away from having all constants in a dedicated file. Over time it's just a bag of things difficult to follow. If it's used across multiple places, it could be just `public` in this class.

packages/relay/src/lib/services/transactionService/TransactionService.ts
```diff
@@ -0,0 +1,860 @@
...
+  private readonly common: ICommonService;
+
+  /**
+   * An instance of EventEmitter used for emitting and handling events within the class.
+   *
+   * @private
+   * @readonly
+   * @type {EventEmitter}
+   */
+  private readonly eventEmitter: EventEmitter;
```
acuarica, line 53: I guess we could start typing the events, so at its definition is clear what is being emitted.

packages/relay/src/lib/services/transactionService/TransactionService.ts
```diff
@@ -0,0 +1,860 @@
...
+  private static ethGasPrice: string;
+
+  /**
+   * Constructor for the TransactionService class.
+   */
+  constructor(
+    cacheService: CacheService,
+    chain: string,
+    common: ICommonService,
+    eventEmitter: EventEmitter,
```
acuarica, line 99: why is this a parameter? I mean, is the emitter instance shared with some other class? Otherwise, this class should create it.

quiet-node (approved):
Great work! Just left some notes and suggestions if you agree let's create tickets to keep track! Otherwise all good on my end

packages/relay/src/lib/services/transactionService/TransactionService.ts
```diff
@@ -0,0 +1,860 @@
...
+
+  /**
+   * Removes the '0x' prefix from a string if present
+   * @param input The input string
+   * @returns {string} The input string without the '0x' prefix
+   */
+  private prune0x(input: string): string {
+    return input.startsWith(CommonService.EMPTY_HEX) ? input.substring(2) : input;
+  }
+
```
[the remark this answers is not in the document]
konstantinabl: This would require using common service in sdk client and do more changes, can we address in another PR?

packages/relay/src/lib/services/transactionService/TransactionService.ts
konstantinabl: I think we should open an issue and do this in a following PR, so we type events everywhere

packages/relay/src/lib/services/transactionService/TransactionService.ts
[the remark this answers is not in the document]
quiet-node: Gotcha — yeah, it was just a quick observation. We should do it in another PR if it requires many changes. Also I see several topics that need to be addressed in different PRs particularly for this topic. Let’s create a centralized technical debt issue for the ETH refactor and house all those topics there. We can create the actual tickets when we actually work on them so we have clear priorities and don’t clutter the ticket board too much.

packages/relay/src/lib/services/factories/ITransactionReceiptFactoryParams.ts
konstantinabl: https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3705 Issue opened

packages/relay/src/lib/services/transactionService/TransactionService.ts
konstantinabl: Issue opened https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3706

packages/relay/src/lib/services/transactionService/TransactionService.ts
konstantinabl: I think both passing on eventEmitter instance and creating a new one for each class are valid cases and we need to discuss this on a global leveл. Thus I opened an issue - > https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3707

PR #3660, opened 2025-04-08
fix: enhanced JsonRpcError handling and message formatting

### LINKED ISSUE #3658: Refactor createJsonRpcError to be static and move to utils module
## Description

The `createJsonRpcError` method in the `RpcMethodDispatcher` class (packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts) is currently an instance method but doesn't use any class state or properties.

This method should be either:

- Made static since it's not related to the dispatcher's state
- Or potentially moved to the utils module so it can be reused by other components

## Proposed Changes

- Refactor `createJsonRpcError` to be a static method
- Consider migrating it to the utils module for better code reusability
- Update any references to this method throughout the codebase

quiet-node:
### Description:
This PR improves JsonRpcError handling by standardizing message formatting to prevent request ID duplication and turning `createJsonRpcError()` to `JsonRpcError.newWithRequestId` static method for clarity. 

### Changes
- Prevent duplicate requestId prefixes in error messages by adding validation in the JsonRpcError constructor
- Rename `createJsonRpcError` to `newJsonRpcErrorWithRequestId` to better convey its purpose
- Move error creation method to the Utils class for better organization
- Add comprehensive test coverage for both changes

Fixes #3658

### DIFF
--- a/packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
+++ b/packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
@@ -169,24 +169,24 @@ export class RpcMethodDispatcher {
 
     // If error is already a JsonRpcError, use it directly
     if (error instanceof JsonRpcError) {
-      return this.createJsonRpcError(error, requestDetails.requestId);
+      return Utils.newJsonRpcErrorWithRequestId(error, requestDetails.requestId);
     }
 
     // Handle GRPC timeout errors
     if (error instanceof SDKClientError && error.isGrpcTimeout()) {
-      return this.createJsonRpcError(predefined.REQUEST_TIMEOUT, requestDetails.requestId);
+      return Utils.newJsonRpcErrorWithRequestId(predefined.REQUEST_TIMEOUT, requestDetails.requestId);
     }
 
     // Handle MirrorNodeClientError by mapping to the correct JsonRpcError
     if (error instanceof MirrorNodeClientError) {
-      return this.createJsonRpcError(
+      return Utils.newJsonRpcErrorWithRequestId(
         predefined.MIRROR_NODE_UPSTREAM_FAIL(error.statusCode, error.message || 'Mirror node upstream failure'),
         requestDetails.requestId,
       );
     }
 
     // Default to internal error for all other error types
-    return this.createJsonRpcError(predefined.INTERNAL_ERROR(errorMessage), requestDetails.requestId);
+    return Utils.newJsonRpcErrorWithRequestId(predefined.INTERNAL_ERROR(errorMessage), requestDetails.requestId);
   }
 
   /**
@@ -213,19 +213,5 @@ export class RpcMethodDispatcher {
 
     // Default response for truly unknown methods
     throw predefined.METHOD_NOT_FOUND(methodName);
-  }
-
-  /**
-   * Creates a new JsonRpcError with the request ID attached to assist with tracing and debugging
-   */
-  private createJsonRpcError(error: JsonRpcError, requestId: string): JsonRpcError {
-    return new JsonRpcError(
-      {
-        code: error.code,
-        message: error.message,
-        data: error.data,
-      },
-      requestId,
-    );
   }
 }
--- a/packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
+++ b/packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
@@ -169,24 +169,24 @@ export class RpcMethodDispatcher {
 
     // If error is already a JsonRpcError, use it directly
     if (error instanceof JsonRpcError) {
-      return Utils.newJsonRpcErrorWithRequestId(error, requestDetails.requestId);
+      return JsonRpcError.newWithRequestId(error, requestDetails.requestId);
     }
 
     // Handle GRPC timeout errors
     if (error instanceof SDKClientError && error.isGrpcTimeout()) {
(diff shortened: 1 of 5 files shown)


packages/relay/src/utils.ts
```diff
@@ -212,4 +213,23 @@ export class Utils {
     // No configuration, use default behavior
     return [...rpcParams, requestDetails];
   }
+
+  /**
+   * Creates a new JsonRpcError with the request ID attached to assist with tracing and debugging.
+   *
+   * @param error - The original JsonRpcError instance to bind the request ID to
```
konstantinabl, line 220: Should we add the param types here in the docs? In other places we have done so and I usually add them, maybe we can discuss but its best to stick to a unified approach?

acuarica: > Should we add the param types here in the docs?

I guess it's duplicating the param definitions already provided in the function/method. If that changes, then we need to change it also here.

> maybe we can discuss but its best to stick to a unified approach?

Definitely.

quiet-node: Gotcha okay good catch let's stick to the current approach for now. Updated.

konstantinabl (dismissed):

packages/relay/src/utils.ts
```diff
@@ -212,4 +213,23 @@ export class Utils {
...
+
+  /**
+   * Creates a new JsonRpcError with the request ID attached to assist with tracing and debugging.
+   *
+   * @param {JsonRpcError} error - The original JsonRpcError instance to bind the request ID to
+   * @param {string} requestId - The unique identifier of the request that generated this error
+   * @returns A new JsonRpcError instance with identical properties but with the request ID
+   *          prefixed to the error message
+   */
+  public static newJsonRpcErrorWithRequestId(error: JsonRpcError, requestId: string): JsonRpcError {
```
acuarica, line 225: I think this would be better placed in `JsonRpcError` class itself. It's easier to understand and more cohesive (no unrelated stuff in the `Utils` class). And given it would be part of `JsonRpcError` class, then it could be renamed to `newWithRequestId` or `withRequestId`.

packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
```diff
@@ -169,24 +169,24 @@ export class RpcMethodDispatcher {
 
     // If error is already a JsonRpcError, use it directly
     if (error instanceof JsonRpcError) {
-      return this.createJsonRpcError(error, requestDetails.requestId);
+      return Utils.newJsonRpcErrorWithRequestId(error, requestDetails.requestId);
```
acuarica, line 172: no more `this` yay :)

acuarica (commented):
nice stuff! left a suggestion

acuarica (dismissed):
Actually, I'll approve it, feel free to implement the suggestion or discard it.

packages/relay/src/utils.ts
```diff
@@ -10,6 +10,7 @@ import { Logger } from 'pino';
 import { hexToASCII, prepend0x, strip0x } from './formatters';
 import constants, { TracerType } from './lib/constants';
 import { RPC_LAYOUT, RPC_PARAM_LAYOUT_KEY } from './lib/decorators';
+import { JsonRpcError } from './lib/errors/JsonRpcError';
```
acuarica, line 13: nit: leftover, remove

wondering why doesn't the linter catch this kind thing.

packages/relay/tests/lib/utils.spec.ts
```diff
@@ -10,6 +10,7 @@ import { ASCIIToHex, prepend0x } from '../../src/formatters';
 import constants, { TracerType } from '../../src/lib/constants';
 import { RPC_PARAM_LAYOUT_KEY } from '../../src/lib/decorators/rpcParamLayoutConfig.decorator';
 import { RPC_LAYOUT } from '../../src/lib/decorators/rpcParamLayoutConfig.decorator';
+import { JsonRpcError } from '../../src/lib/errors/JsonRpcError';
```
acuarica, line 13: nit: leftover, remove

packages/relay/src/lib/errors/JsonRpcError.ts
```diff
@@ -10,8 +10,33 @@ export class JsonRpcError {
 
   constructor(args: { code: number; message: string; data?: string }, requestId?: string) {
     this.code = args.code;
-    this.message = requestId ? `[${constants.REQUEST_ID_STRING}${requestId}] ` + args.message : args.message;
     this.data = args.data;
+
+    const hasRequestId = requestId && !args.message.includes(constants.REQUEST_ID_STRING);
+    this.message = hasRequestId ? `[${constants.REQUEST_ID_STRING}${requestId}] ${args.message}` : args.message;
```
acuarica, line 16: Sweet, I definitely like it better this way.

I see that almost all usages of this constructor don't pass the `requestId` argument. There is one in `validators.ts` and the other in `webSocketServer.ts` (this one actually passes `undefined`). So if we want to go one step further, I'd move this logic into `newWithRequestId` and update the call sites.

packages/relay/tests/lib/utils.spec.ts
quiet-node: Good catch! Updated!

packages/relay/src/utils.ts
quiet-node: Good catch! Updated

simzzz (approved):
LGTM

PR #3697, opened 2025-04-16
feat: change eth_maxPriorityFeePerGas to return UNSUPPORTED_METHOD

### LINKED ISSUE #3487: Investigate implementation for eth_maxPriorityFeePerGas
Create endpoint eth_maxPriorityFeePerGas and implement response UNSUPPORTED_METHOD

simzzz:
**Description**:
This PR changes eth_maxPriorityFeePerGas to return UNSUPPORTED_METHOD

**Related issue(s)**:

Fixes #3487 

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -3240,7 +3240,7 @@ export class EthImpl implements Eth {
    */
   @rpcMethod
   @rpcParamLayoutConfig(RPC_LAYOUT.REQUEST_DETAILS_ONLY)
-  maxPriorityFeePerGas(requestDetails: RequestDetails): JsonRpcError {
+  public maxPriorityFeePerGas(requestDetails: RequestDetails): JsonRpcError {
     if (this.logger.isLevelEnabled('trace')) {
       this.logger.trace(`${requestDetails.formattedRequestId} maxPriorityFeePerGas()`);
     }


packages/relay/src/lib/eth.ts
```diff
@@ -3240,11 +3240,11 @@ export class EthImpl implements Eth {
    */
   @rpcMethod
   @rpcParamLayoutConfig(RPC_LAYOUT.REQUEST_DETAILS_ONLY)
-  async maxPriorityFeePerGas(requestDetails: RequestDetails): Promise<string> {
+  maxPriorityFeePerGas(requestDetails: RequestDetails): JsonRpcError {
```
konstantinabl, line 3243: Should we add a visibility keyword here?

konstantinabl (changes requested):

packages/relay/src/lib/eth.ts
simzzz: yes, it's a good practice

packages/relay/tests/lib/eth/eth_common.spec.ts
```diff
@@ -84,7 +84,10 @@ describe('@ethCommon', async function () {
 
     it('should execute "eth_maxPriorityFeePerGas"', async function () {
       const result = await relay.eth().maxPriorityFeePerGas(requestDetails);
```
quiet-node, line 86: seems like the await can be dropped

quiet-node, line 86: seems like the await can be dropped

quiet-node (approved):
LGTM but let’s confirm that Fer is also aware of this, as updating a 200 HTTP response to a 400 could be a breaking change.

PR #3698, opened 2025-04-16
feat: improves eth_getCode caching

### LINKED ISSUE #3634: Improve `eth_getCode` caching strategy when bytecode contains "prohibited opcode"
The current implementation of `eth_getCode` does not cache the result when bytecode contains "prohibited opcodes"

https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/7955dae8c5819413236240ed0e89e167dc5b08fe/packages/relay/src/lib/eth.ts#L1206-L1218

This was initially put in place because some of these opcodes might change indeed the bytecode at a given address. However, the cache label already takes into account the `blockNumber`

https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/7955dae8c5819413236240ed0e89e167dc5b08fe/packages/relay/src/lib/eth.ts#L1179

so these opcodes shouldn't affect the result.

Improve caching strategy for `eth_getCode` when `blockNumber` is not `latest`.

simzzz:
**Description**:
This PR improves the caching for `eth_getCode` by caching even with prohibited opcodes, and removing caching for blockNumber === latest

**Related issue(s)**:

Fixes #3634 

**Checklist**

### DIFF
--- a/packages/relay/src/lib/services/ethService/contractService/ContractService.ts
+++ b/packages/relay/src/lib/services/ethService/contractService/ContractService.ts
@@ -1,5 +1,4 @@
 // SPDX-License-Identifier: Apache-2.0
-import { disassemble } from '@ethersproject/asm';
 import { ConfigService } from '@hashgraph/json-rpc-config-service/dist/services';
 import crypto from 'crypto';
 import { Logger } from 'pino';
--- a/packages/relay/src/lib/services/ethService/contractService/ContractService.ts
+++ b/packages/relay/src/lib/services/ethService/contractService/ContractService.ts
@@ -255,11 +255,7 @@ export class ContractService implements IContractService {
           return CommonService.redirectBytecodeAddressReplace(address);
         } else if (result.type === constants.TYPE_CONTRACT) {
           if (result.entity.runtime_bytecode !== constants.EMPTY_HEX) {
-            const prohibitedOpcodes = ['CALLCODE', 'DELEGATECALL', 'SELFDESTRUCT', 'SUICIDE'];
-            const opcodes = disassemble(result.entity.runtime_bytecode);
-            const hasProhibitedOpcode =
-              opcodes.filter((opcode) => prohibitedOpcodes.indexOf(opcode.opcode.mnemonic) > -1).length > 0;
-            if (!hasProhibitedOpcode) {
+            if (blockNumber !== 'latest') {
               await this.cacheService.set(
                 cachedLabel,
                 result.entity.runtime_bytecode,
--- a/packages/relay/tests/lib/eth/eth_getCode.spec.ts
+++ b/packages/relay/tests/lib/eth/eth_getCode.spec.ts
@@ -15,6 +15,7 @@ import {
   DEFAULT_CONTRACT,
   DEFAULT_HTS_TOKEN,
   DEFAULT_NETWORK_FEES,
+  DEPLOYED_BYTECODE,
   HTS_TOKEN_ADDRESS,
   MIRROR_NODE_DEPLOYED_BYTECODE,
   NO_TRANSACTIONS,
@@ -340,7 +341,7 @@ describe('@ethGetCode using MirrorNode', async function () {
       expect(firstCallHistory.length).to.equal(2);
       expect(firstCallHistory[0].url).to.include(`contracts/${CONTRACT_ADDRESS_1}`); // Contract lookup
       expect(firstCallHistory[1].url).to.include('blocks/0'); // Earliest block lookup
-      expect(firstResult).to.equal(EthImpl.emptyHex); // Should return empty for SELFDESTRUCT
+      expect(firstResult).to.equal(constants.EMPTY_HEX); // Should return empty for SELFDESTRUCT
 
       // Reset history and change mock responses to verify both entity and bytecode caching
       restMock.resetHistory();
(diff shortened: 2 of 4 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1537,19 +1537,15 @@ export class EthImpl implements Eth {
           return EthImpl.redirectBytecodeAddressReplace(address);
         } else if (result?.type === constants.TYPE_CONTRACT) {
           if (result?.entity.runtime_bytecode !== EthImpl.emptyHex) {
-            const prohibitedOpcodes = ['CALLCODE', 'DELEGATECALL', 'SELFDESTRUCT', 'SUICIDE'];
-            const opcodes = disassemble(result?.entity.runtime_bytecode);
-            const hasProhibitedOpcode =
-              opcodes.filter((opcode) => prohibitedOpcodes.indexOf(opcode.opcode.mnemonic) > -1).length > 0;
-            if (!hasProhibitedOpcode) {
+            if (blockNumber !== 'latest') {
```
konstantinabl, line 1540: If we don't have tests for this, maybe we can add some?

konstantinabl (changes requested):

packages/relay/src/lib/eth.ts
quiet-node: If possible we should test if the bytecode is cached even when prohibited opcodes are found.

simzzz: Done @quiet-node

packages/relay/tests/lib/eth/eth_getCode.spec.ts
```diff
@@ -281,5 +281,54 @@ describe('@ethGetCode using MirrorNode', async function () {
...
       expect(res).to.equal(EthImpl.emptyHex);
     });
+
+    it('should not cache bytecode when using latest block parameter', async () => {
+      // Record initial number of GET requests
+      const initialGetCount = restMock.history.get.length;
+      // First call with 'latest' should make network calls
+      await ethImpl.getCode(CONTRACT_ADDRESS_1, 'latest', requestDetails);
+      const afterFirstGetCount = restMock.history.get.length;
+      expect(afterFirstGetCount).to.be.greaterThan(initialGetCount);
```
konstantinabl, line 291: @simzzz Can we make this test slightly more specific. e.g instead of checking if its greater than we may expect an exact count, or maybe just check if the cache service is not called at all.

packages/relay/tests/lib/eth/eth_getCode.spec.ts
```diff
@@ -281,5 +281,54 @@ describe('@ethGetCode using MirrorNode', async function () {
...
+        }),
+      );
+
+      // Record initial number of GET requests
+      const initialGetCount = restMock.history.get.length;
+
+      // First call with 'earliest' should make network calls
+      const firstResult = await ethImpl.getCode(CONTRACT_ADDRESS_1, 'earliest', requestDetails);
+      const afterFirstGetCount = restMock.history.get.length;
+      expect(afterFirstGetCount).to.be.greaterThan(initialGetCount);
```
konstantinabl, line 324: Maybe make this more specific as well

packages/relay/tests/lib/eth/eth_getCode.spec.ts
simzzz: I made an improvement that I think makes it more clear now. You can take a look @konstantinabl

quiet-node (dismissed):
LGTM

quiet-node (approved):

Ferparishuertas (approved):
LGTM

konstantinabl (dismissed):

quiet-node (approved):
LGTM

packages/relay/src/lib/services/ethService/contractService/ContractService.ts
```diff
@@ -255,11 +255,7 @@ export class ContractService implements IContractService {
           return CommonService.redirectBytecodeAddressReplace(address);
         } else if (result.type === constants.TYPE_CONTRACT) {
           if (result.entity.runtime_bytecode !== constants.EMPTY_HEX) {
-            const prohibitedOpcodes = ['CALLCODE', 'DELEGATECALL', 'SELFDESTRUCT', 'SUICIDE'];
-            const opcodes = disassemble(result.entity.runtime_bytecode);
-            const hasProhibitedOpcode =
-              opcodes.filter((opcode) => prohibitedOpcodes.indexOf(opcode.opcode.mnemonic) > -1).length > 0;
-            if (!hasProhibitedOpcode) {
+            if (blockNumber !== 'latest') {
```
acuarica, line 258: Does it make sense to move the cache set right after the `HTS_ADDRESS` comparison above?

Skimming through the logs in testnet, around 50% of `eth_getCode` calls pass an HTS token address. The main reason to cache HTS bytecode calls is to avoid calls to `resolveEntityType` and `getHistoricalBlockResponse`.

However, both `resolveEntityType` and `getHistoricalBlockResponse` are cached as well. That means we save only 1 cache request in the best scenario (when all required data is cached).

acuarica (changes requested):

acuarica:
Even if we don't apply code caching at this level, we should still remove the `prohibitedOpcodes` related code and the `asm` dependencies if they are not relevant anymore.

PR #3714, opened 2025-04-22
fix: eth_sendRawTransaction now handles malformed RLP data correctly

### LINKED ISSUE #3652: Fix eth_sendRawTransaction 500 error when handling malformed RLP payloads
## Issue Description

The `eth_sendRawTransaction` endpoint returns a 500 error instead of a proper 400 client error when receiving transactions with malformed RLP payloads. The error occurs during transaction parsing in `Precheck.parseTxIfNeeded` when ethers.js encounters unexpected data after what should be a complete RLP transaction.

## Steps to Reproduce

1. Send a malformed transaction to the `eth_sendRawTransaction` endpoint
2. Example problematic transaction:

```bash
0xf8748201280585800e8dfc0085800e8dfc00832dc6c094aca85ef7e1fce27079bbf99b60fcf6fd19b99b248502540be40080c001a0210446cfb671c3174392410d52fa3cd58723d8417e40cc67c6225b8f7e3ff693a02674b392846c59f783ea96655d39560956dd987051972064a5d853cea0b6f6d711
```

## Current Behavior

- Error: `unexpected junk after rlp payload (argument="data", value=<transaction-hex>, code=INVALID_ARGUMENT, version=6.13.5)`
- Returns HTTP 500 status code
- Error is not properly mapped to a client error

## Expected Behavior

- Return HTTP 400 status code
- Map to a proper client error like `INVALID_ARGUMENTS`
- Include helpful error information for client debugging

## Root Cause

The error occurs in `Precheck.parseTxIfNeeded` when calling `Transaction.from(transaction)` from ethers.js. The specific error (INVALID_ARGUMENT) isn't handled in the error mapping logic, so it's treated as an unknown internal error (500).

## Proposed Solution

1. Add specific error handling in `parseRawTxAndPrecheck` or `genericErrorHandler` to detect ethers.js "unexpected junk after rlp payload" errors, or more generic to check for INVALID_ARGUMENTS code
2. Map these errors to the proper `predefined.INVALID_ARGUMENTS` JsonRpcError
3. Add tests to verify proper error handling for malformed transactions

simzzz:
**Description**:
In this PR, the `eth_sendRawTransaction` logic now handles malformed RLP data correctly by checking for INVALID_ARGUMENT error and returning a predifined error and 400 status code.

**Related issue(s)**:

Fixes #3652 

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1935,8 +1935,9 @@ export class EthImpl implements Eth {
     networkGasPriceInWeiBars: number,
     requestDetails: RequestDetails,
   ): Promise<EthersTransaction> {
-    const parsedTx = Precheck.parseTxIfNeeded(transaction);
+    let parsedTx;
     try {
+      parsedTx = Precheck.parseTxIfNeeded(transaction);
       if (this.logger.isLevelEnabled('debug')) {
         this.logger.debug(
           `${requestDetails.formattedRequestId} Transaction undergoing prechecks: transaction=${JSON.stringify(
@@ -1949,6 +1950,9 @@ export class EthImpl implements Eth {
       await this.precheck.sendRawTransactionCheck(parsedTx, networkGasPriceInWeiBars, requestDetails);
       return parsedTx;
     } catch (e: any) {
+      if (e.code === 'INVALID_ARGUMENT') {
+        throw predefined.INVALID_ARGUMENTS(e.message.toString());
+      }
       this.logger.error(
         `${requestDetails.formattedRequestId} Precheck failed: transaction=${JSON.stringify(parsedTx)}`,
       );
--- a/packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
+++ b/packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
@@ -125,8 +125,8 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
       sdkClientStub = sinon.createStubInstance(SDKClient);
       sinon.stub(hapiServiceInstance, 'getSDKClient').returns(sdkClientStub);
       restMock.onGet(accountEndpoint).reply(200, JSON.stringify(ACCOUNT_RES));
-   JSON.stringify(   restMock.onGet(receiverAccountEndpoint).reply(200, JSON.stringify(RECEIVER_ACCOUNT_RES)));
-   JSON.stringify(   restMock.onGet(networkExchangeRateEndpoint).reply(200, JSON.stringify(mockedExchangeRate)));
+      JSON.stringify(restMock.onGet(receiverAccountEndpoint).reply(200, JSON.stringify(RECEIVER_ACCOUNT_RES)));
+      JSON.stringify(restMock.onGet(networkExchangeRateEndpoint).reply(200, JSON.stringify(mockedExchangeRate)));
     });
 
     afterEach(() => {
@@ -192,6 +192,18 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
         false,
         ethImpl,
         [txHash, requestDetails],
+      );
+    });
+
+    it('should return a predefined INVALID_ARGUMENTS when transaction has invalid format', async function () {
+      const invalidTx =
+        '0xf8748201280585800e8dfc0085800e8dfc00832dc6c094aca85ef7e1fce27079bbf99b60fcf6fd19b99b248502540be40080c001a0210446cfb671c3174392410d52fa3cd58723d8417e40cc67c6225b8f7e3ff693a02674b392846c59f783ea96655d39560956dd987051972064a5d853cea0b6f6d711';
+      await RelayAssertions.assertRejection(
+        predefined.INVALID_ARGUMENTS('invalid hex string'),
+        ethImpl.sendRawTransaction,
+        false,
+        ethImpl,
+        [invalidTx, requestDetails],
       );
     });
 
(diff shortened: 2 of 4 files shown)


packages/ws-server/tests/acceptance/sendRawTransaction.spec.ts
```diff
@@ -178,6 +186,18 @@ describe('@web-socket-batch-2 eth_sendRawTransaction', async function () {
       expect(fromAccountInfo.evm_address).to.eq(accounts[1].address.toLowerCase());
     });
 
+    it(`Should fail eth_sendRawTransaction on Ethers Web Socket Provider when transaction has invalid format`, async () => {
+      const invalidTx = '0xinvalidtransactionformat';
```
quiet-node, line 190: Just curious why don't we use the invlaid real txHash the other tests are using but decided to use an invalid txHash for this case?

quiet-node, line 190: Just curious why don't we use the invlaid real txHash the other tests are using but decided to use an invalid txHash for this case?

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -1002,7 +1002,7 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
         };
 
         const signedTx = await accounts[2].wallet.signTransaction(transaction);
-        const error = predefined.INTERNAL_ERROR();
+        const error = predefined.INVALID_ARGUMENTS('unexpected junk after rlp payload');
```
quiet-node, line 1005: So it did happen in our test before but no one has caught it. Nice one!

quiet-node (changes requested):

packages/relay/src/lib/eth.ts
```diff
@@ -1949,6 +1950,9 @@ export class EthImpl implements Eth {
       await this.precheck.sendRawTransactionCheck(parsedTx, networkGasPriceInWeiBars, requestDetails);
       return parsedTx;
     } catch (e: any) {
+      if (e.code === 'INVALID_ARGUMENT') {
+        throw predefined.INVALID_ARGUMENTS(e.message.toString());
+      }
```
quiet-node, line 1955: Hm it feels a bit cumbersome here. The `parsedTx` is refactored solely to reference it in the error log within the catch block, which seems unnecessary. Additionally, the try/catch is meant to focus on the precheck but ends up handling parsing errors as well, making it feel like a hack or a band-aid solution.

Moreover, the `if (e.code === "INVALID_ARGUMENT")` block seems somewhat arbitrary, given that the error is already expected to be handled by `this.common.genericErrorHandler(e)`.

I'm thinking it might be cleaner to update `Precheck.parseTxIfNeeded` instead, since the error originates from `Transaction.from()` inside that method. Handling the error there would simplify things and make it a bit cleaner of an approach.

Something simple like:

```typescript
public static parseTxIfNeeded(transaction: string | Transaction): Transaction {
  try {
    return typeof transaction === 'string' ? Transaction.from(transaction) : transaction;
  } catch (error: any) {
    throw predefined.INVALID_ARGUMENTS(error.message);
  }
}
```

quiet-node (changes requested):

packages/ws-server/tests/acceptance/sendRawTransaction.spec.ts
simzzz: Both return the same error, but I agree a real invalid hash would be better. I'll change it

packages/relay/src/lib/precheck.ts
```diff
@@ -36,7 +36,11 @@ export class Precheck {
    * @returns {Transaction} The parsed transaction.
    */
   public static parseTxIfNeeded(transaction: string | Transaction): Transaction {
-    return typeof transaction === 'string' ? Transaction.from(transaction) : transaction;
+    try {
+      return typeof transaction === 'string' ? Transaction.from(transaction) : transaction;
+    } catch (e: any) {
+      throw predefined.INVALID_ARGUMENTS(e.message.toString());
+    }
```
quiet-node, line 43: Nice please also don't forget to add a couple of unit tests for this guy in precheck.spec.ts

packages/relay/src/lib/eth.ts
simzzz: Yes, I agree. I didn't think it belonged to `genericErrorHandler(...)` so I did it here but `parseTxIfNeeded (...)` is a better choice. Moved.

packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
```diff
@@ -195,6 +195,18 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
       );
     });
 
+    it('should return a predefined INVALID_ARGUMENTS when transaction has invalid format', async function () {
+      const invalidTx =
+        '0xf8748201280585800e8dfc0085800e8dfc00832dc6c094aca85ef7e1fce27079bbf99b60fcf6fd19b99b248502540be40080c001a0210446cfb671c3174392410d52fa3cd58723d8417e40cc67c6225b8f7e3ff693a02674b392846c59f783ea96655d39560956dd987051972064a5d853cea0b6f6d711';
```
konstantinabl, line 200: May we add a comment on exactly how is this tx invalid

quiet-node (approved):
Nice one! Thanks for the good work! LGTM

packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
simzzz: done

PR #3718, opened 2025-04-23
feat: support non-zero shard/realm

### LINKED ISSUE #3614: Ensure Relay is compatible with accounts, contracts and tokens with non-zero shard/realm numbers
Accounts, contracts and tokens the network might have non-zero shard/realm numbers. See https://docs.hedera.com/hedera/core-concepts/accounts/account-properties for an overview of account IDs in the network.

We need to ensure the Relay supports entities with non-zero shard/realm numbers. Identify all places where this assumption is made and remove them with a more general solution.

Some places to start

- The Relay uses the following constant to identify long-zero addresses
https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/37359837a0beb72da91659bdabba241e187a6715/packages/relay/src/lib/constants.ts#L119

- For example, the following might not longer be valid with non-zero shards/realm https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/e3859b7b145d57a9cc61516c7ea3f53b656c4b78/packages/relay/src/lib/clients/mirrorNodeClient.ts#L1367-L1377

> [!NOTE]
> This bit is from `resolveEntityType` method, which in turn is used in `resolveEvmAddress`. There two methods are used in several places.

- ~~In the following the sizes of shard and realm does not seem to match the right sizes (shard, realm and account are 8 bytes each)~~ The docs were wrong, indicating that shard was 4 bytes. The docs are fixed now.
https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/e3859b7b145d57a9cc61516c7ea3f53b656c4b78/packages/server/tests/helpers/utils.ts#L43-L53

acuarica:
**Description**:

This PR adds preliminary support for non-zero shard/realm numbers. As described in #3614, it was mostly related to `eth_getCode` and the redirect bytecode. This was refactored to avoid duplication.

In addition it now includes `0x` hex prefix in the HTS redirect bytecode. The spec indicated that `eth_getCode` result response should be a hex encoded value starting with `0x`.

**Related issue(s)**:

Fixes #3614

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -1398,18 +1398,23 @@ export class MirrorNodeClient {
           : Promise.reject(),
       ];
 
-      // only add long zero evm addresses for tokens as they do not refer to actual contract addresses but rather encoded entity nums
-      if (entityIdentifier.startsWith(constants.LONG_ZERO_PREFIX)) {
-        promises.push(
-          searchableTypes.find((t) => t === constants.TYPE_TOKEN)
-            ? buildPromise(
-                this.getTokenById(`0.0.${parseInt(entityIdentifier, 16)}`, requestDetails, retries).catch(() => {
-                  return null;
-                }),
-              )
-            : Promise.reject(),
-        );
-      }
+      const toEntityId = (eid: string) => {
+        eid = eid.slice(2);
+        const shard = eid.slice(0, 4 * 2);
+        const realm = eid.slice(4 * 2, 12 * 2);
+        const num = eid.slice(12 * 2);
+        return `${parseInt(shard, 16)}.${parseInt(realm, 16)}.${parseInt(num, 16)}`;
+      };
+
+      promises.push(
+        searchableTypes.find((t) => t === constants.TYPE_TOKEN)
+          ? buildPromise(
+              this.getTokenById(toEntityId(entityIdentifier), requestDetails, retries).catch(() => {
+                return null;
+              }),
+            )
+          : Promise.reject(),
+      );
 
       // maps the promises with indices of the promises array
       // because there is no such method as Promise.anyWithIndex in js
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -1398,12 +1398,12 @@ export class MirrorNodeClient {
           : Promise.reject(),
       ];
 
-      const toEntityId = (eid: string) => {
-        eid = eid.slice(2);
-        const shard = eid.slice(0, 4 * 2);
-        const realm = eid.slice(4 * 2, 12 * 2);
-        const num = eid.slice(12 * 2);
-        return `${parseInt(shard, 16)}.${parseInt(realm, 16)}.${parseInt(num, 16)}`;
+      const toEntityId = (address: string) => {
+        address = address.slice(2);
+        const shardNum = address.slice(0, 8);
+        const realmNum = address.slice(8, 24);
+        const accountNum = address.slice(24);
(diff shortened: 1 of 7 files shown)


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1398,18 +1398,23 @@ export class MirrorNodeClient {
...
-            ? buildPromise(
-                this.getTokenById(`0.0.${parseInt(entityIdentifier, 16)}`, requestDetails, retries).catch(() => {
-                  return null;
-                }),
-              )
-            : Promise.reject(),
-        );
-      }
+      const toEntityId = (address: string) => {
+        address = address.slice(2);
```
Ferparishuertas, line 1402: Its ok by now. but order is still under discussion being able to swap to realm.shard.
We will adress it in other PR if needed

packages/relay/tests/lib/eth/eth_getCode.spec.ts
```diff
@@ -21,6 +21,16 @@ import { generateEthTestEnv } from './eth-helpers';
 
 use(chaiAsPromised);
 
+function entityIdToEvmAddress(entityId: string): string {
+  const pad = (num: string, n: number) =>
+    Number(num)
+      .toString(16)
+      .padStart(n * 2, '0');
+  const [shardNum, realmNum, accountNum] = entityId.split('.');
```
Ferparishuertas, line 29: same here. To be addressed if ids swap position

Ferparishuertas (approved):
LGTM

packages/relay/src/lib/clients/mirrorNodeClient.ts
acuarica: Sounds good, thanks for the heads up.

quiet-node (approved):
LGTM! Good work!

packages/relay/src/lib/clients/mirrorNodeClient.ts
acuarica: > Addresses are sometimes passed with '0x' and sometimes without.

Oh, should/are we be lenient in how we take `address` arguments? The spec says `address` starts with `0x`, from https://ethereum.github.io/execution-apis/api-documentation/

image

If we are lenient, then yes to your original question.

packages/server/tests/acceptance/erc20.spec.ts
```diff
@@ -111,14 +110,9 @@ describe('@erc20 Acceptance Tests', async function () {
...
-          EthImpl.redirectBytecodePostfix
-        }`;
-        if (testTitles[i].testName !== HTS) {
-          expect(res).to.eq(testTitles[i].expectedBytecode);
-        } else {
-          expect(res).to.eq(expectedBytecode);
-        }
+        expect(res).to.be.equal(
+          testTitles[i].expectedBytecode ?? EthImpl.redirectBytecodeAddressReplace(contract.target),
+        );
```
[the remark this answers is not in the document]
quiet-node: oh my IDE just prompted it out 
image

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: oh yeah just saying it in general can't tell where but sometimes in the code I see hex string passes with and without '0x' so I thought .replace() will make it safer. Just a nitpick not a big deal

packages/server/tests/acceptance/erc20.spec.ts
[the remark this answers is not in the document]
acuarica: Mmm, did you clean and recompile from scratch? Changes in one package, _e.g._, `relay`, won't be visible in other packages unless you recompile.

PR #3722, opened 2025-04-24
feat: implement Jumbo Transaction support and enhance pre-requisite validation for eth_sendRawTransaction

### LINKED ISSUE #3573: Jumbo transaction logic +  feature flag
### Problem

TBD

### Solution

When i create a new smart contract via sendRawTransaction, I want to be able to directly deploy the contract via a normal Ethereum Transaction, without needing the Json RPC internally to use HFS, and custom logic, due the. 6kb restrition.

The max size has been raised to 128kb, enabling to deploy smart contract directly ( max 24kb + Call params).

We need to add a new logic behind a feature flag, activated by default, that directly pings smart contract deployment to the consensus node, If deactivated, the HFS logic will still be triggered

Acceptance Criteria

- When i deploy a smart contract, lower than the max SC Size 24kb,  via JSON RPC relay, that has the JUMBO_TX_ENABLED= true, the code should succesfully call the consensus node, and succeed, if total transaction size is 128kb

- When i deploy a smart contract, lower than the max sc Size 24kb,  via JSON RPC relay, that has the JUMBO_TX_ENABLED= true, the code should not call the consensus node, if total transaction size is >128kb, showing a validation error

- When i deploy a smart contract, bigger than the max SC Size 24kb,  via JSON RPC relay, that has the JUMBO_TX_ENABLED= true, the code should not call the consensus node, and fail, showing a validation error

- When i deploy a smart contract, smaller than the max SC Size 24kb,  via JSON RPC relay, that has the JUMBO_TX_ENABLED= false, the code should not call the consensus node durectly, using the HFS logic, and succeed.

- When i deploy a smart contract, bugger than the max SC Size 24kb,  via JSON RPC relay, that has the JUMBO_TX_ENABLED= false, the code should not call the consensus node directly, failing

#Implementation Details

- Add a new precheck for th 24kb size ( EIp-170)
- We need sync with Services, to see their implementation staus

### Alternatives

_No response_

quiet-node:
## Description

This PR introduces a feature flag to allow the Relay to handle jumbo transactions, making it possible to process large transaction payloads directly rather than splitting them across the Hedera File System. Additionally, the PR includes several code improvements for better clarity and added validation.

### Key Changes

- Added `JUMBO_TX_ENABLED` feature flag (currently set to false as default value) that allows transactions to bypass size limit checks
- Improved transaction validation process with the following changes:
  - Reworked transaction size precheck logic for simpler implementation, also bumped the total transaction size limit to 130KB
  - Added `callDataSize` precheck to validate callData field against 128KB limit.
  - Added `contractCodeSize` precheck to validate smart contract deployments against EIP-170 limit (24KB)
  - Renamed functions for better clarity and separation of concerns:
    - `parseRawTxAndPrecheck` → `validateRawTransaction`
    - `parseTxIfNeeded` → `parseRawTransaction`
- Added a new acceptance test file, `sendRawTransactionExtension.spec.ts`, to isolate eth_sendRawTransaction-feature-related tests and reduce the load on the rpc_batch1 suite.

Fixes #3573


packages/config-service/src/services/globalConfig.ts
```diff
@@ -405,6 +405,12 @@ const _CONFIG = {
     required: false,
     defaultValue: 1,
   },
+  JUMBO_TX_ENABLED: {
+    envName: 'JUMBO_TX_ENABLED',
+    type: 'boolean',
+    required: false,
+    defaultValue: false,
```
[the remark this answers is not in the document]
Ferparishuertas: Also add the following two max values and check them individually at code and tests
128kb for the callData payload max and the 130kb max for the actual EthereumTransaction max.

These are the new Ethereumtransaction limits

Ferparishuertas (commented):
Pls review the comments and adress them to be able to be in sync with the funtional needs

packages/config-service/src/services/globalConfig.ts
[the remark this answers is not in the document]
Ferparishuertas: Pls also add tests that ensure that with JUMBO tx activated we are not using HFS. Unit and regression tests. ensuring latest local node version enables jumbotx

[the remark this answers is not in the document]
quiet-node: >128kb for the callData payload max

@Ferparishuertas needs some clarification on this bit. To me, `callData payload` means the `data` field in the RLP-encoded transaction object, included alongside fields like `nonce`, `value`, `gasPrice`, `signature`, etc.. Are you suggesting the 128KB limit applies only to the `data` field of the RLP transaction object? I believe that the 128KB limit should  encompass the total size of the serialized, RLP-encoded transaction, including all fields, not just the `data` field. 

> Pls also add tests that ensure that with JUMBO tx activated we are not using HFS. Unit and regression tests. ensuring latest local node version enables jumbotx

There are already unit tests to ensure the HFS flow isn’t called. Adding acceptance tests to verify HFS flow is tricky, since from the client side, we only get back the transaction hash. We can check the validity of the transaction by verifying the hash against the Mirror Node, but I don’t believe it’s possible to determine if HFS was involved from that perspective. wdyt?

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+
+import MirrorClient from '../clients/mirrorClient';
+import RelayClient from '../clients/relayClient';
+import ServicesClient from '../clients/servicesClient';
+// Assertions from local resources
+import Assertions from '../helpers/assertions';
+import { Utils } from '../helpers/utils';
+import { AliasAccount } from '../types/AliasAccount';
+
+describe('@sendRawTransactionExtension Acceptance Tests', function () {
```
quiet-node, line 21: Note for reviewers: Most eth_sendRawTransaction acceptance tests are located in the rpc_batch1 suite. However, that suite has grown quite large, and adding more tests in this PR would significantly increase the load on the local-node during CI. To avoid this, I’ve created this new sendRawTransactionExtension suite specifically for the additional tests in this PR.

In the future, existing eth_sendRawTransaction precheck tests from rpc_batch1 can also be migrated to this new suite for better organization. That migration will be handled in a separate PR to keep this one focused.

packages/config-service/src/services/globalConfig.ts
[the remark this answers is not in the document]
Ferparishuertas: We can add a check that consensus node never gets called, there are interceptors for that. Do you know what i mean?

If we have junit tests that ensure calls = 0, i am ok with it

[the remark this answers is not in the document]
quiet-node: > We can add a check that consensus node never gets called, there are interceptors for that. Do you know what i mean?

I am not aware of these interceptors and not sure how it's possible to do so from the POV of a client.

> If we have junit tests that ensure calls = 0, i am ok with it

Yeah there should be decent unit tests that ensure CN calls to HFS = 0.

andrewb1269hg (dismissed):
Review and approve `.github/workflows/acceptance.yml`

packages/config-service/src/services/globalConfig.ts
```diff
@@ -405,6 +417,12 @@ const _CONFIG = {
     required: false,
     defaultValue: 1,
   },
+  JUMBO_TX_ENABLED: {
```
Ferparishuertas, line 420: Default value should be true, as Consensus node is doing.

If an environment has CN flag false. or feature does not exist, we set it to false

packages/config-service/src/services/globalConfig.ts
```diff
@@ -117,6 +117,12 @@ const _CONFIG = {
     required: false,
     defaultValue: 3600000,
   },
+  CALL_DATA_SIZE_LIMIT: {
+    envName: 'CALL_DATA_SIZE_LIMIT',
```
konstantinabl, line 121: @quiet-node So that means a whole transaction can be maximum 130KB and the call data can be at most 128kb of the whole transaction?

packages/relay/tests/lib/precheck.spec.ts
```diff
@@ -659,37 +658,199 @@ describe('Precheck', async function () {
...
+
+      const signedTx = await signTransaction(oversizedTx);
+      const tx = ethers.Transaction.from(signedTx);
+
       try {
-        precheck.hexToBytes(hexString);
-      } catch (e) {
-        error = e;
+        precheck.transactionSize(tx);
+        expect('Transaction should have been rejected');
```
konstantinabl, line 694: I am not sure adding expect in a try/catch is the best way and generally accepted way to use chai and mocha for testing. I think its better to do:
```
expect(() => precheck.transactionSize(tx))
        .to.throw(JsonRpcError)
        .and.to.deep.equal(expectedError)
```

packages/relay/tests/lib/precheck.spec.ts
```diff
@@ -659,37 +658,199 @@ describe('Precheck', async function () {
...
+      const tx = await createTransaction(data);
 
-    it('should fail if passed empty string', () => {
-      let error;
       try {
-        precheck.hexToBytes('');
-      } catch (e) {
-        error = e;
+        precheck.callDataSize(tx);
+        expect('Transaction should have been rejected').to.be.false;
```
konstantinabl, line 755: This could be improved as I suggested for the **_should reject transactions exceeding size limit_** test

konstantinabl (changes requested):

packages/relay/tests/lib/precheck.spec.ts
```diff
@@ -659,37 +658,199 @@ describe('Precheck', async function () {
...
+    it('should reject contract creation with code size exceeding limit', async () => {
+      // Create data that exceeds the limit
+      const dataSize = constants.CONTRACT_CODE_SIZE_LIMIT + 1;
+      const data = '0x' + '00'.repeat(dataSize);
+
+      const tx = await createTransaction(data);
+
+      try {
+        precheck.contractCodeSize(tx);
+        expect('Transaction should have been rejected');
```
konstantinabl, line 828: Same as https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3722#discussion_r2086672638

packages/config-service/src/services/globalConfig.ts
quiet-node: Gotcha that makes sense! updated to true by default.

packages/config-service/src/services/globalConfig.ts
quiet-node: Yes exactly. Those are the new throttles coming along with Jumbo Transaction.

packages/relay/src/lib/precheck.ts
```diff
@@ -317,35 +319,34 @@ export class Precheck {
...
+   * by subtracting the '0x' prefix (2 characters) and dividing by 2 (since each byte is represented by 2 hex characters).
+   *
+   * @param {Transaction} tx - The transaction to validate.
+   * @throws {JsonRpcError} If the transaction size exceeds the configured limit.
    */
-  hexToBytes(hex: string): Uint8Array {
-    if (hex === '') {
-      throw predefined.INTERNAL_ERROR('Passed hex an empty string');
+  transactionSize(tx: Transaction): void {
+    const totalRawTransactionSizeInBytes = tx.serialized.replace('0x', '').length / 2;
```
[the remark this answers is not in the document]
quiet-node: Yeah good point! However, I believe this wouldn’t be necessary. In this case, the `tx` object passed to this precheck is already a valid RLP-parsed transaction, which means it was already parsed by the ethers.Transaction() module. 

If the raw signed transaction had an invalid length like an odd length, `Precheck.parseRawTransaction(transaction)` already catches it. Currently, it'd throw `Error invoking RPC: invalid BytesLike value`, but this PR https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3714 by Simo enhances the handling of invalid RLP-encoded transactions and makes it an expected error if rlp encoded transaction is invalid. Therefore, adding an explicit even-length check here would likely be redundant.

Would there be an edge case?

packages/relay/src/lib/precheck.ts
```diff
@@ -317,35 +319,34 @@ export class Precheck {
...
    */
-  checkSize(transaction: string): void {
-    const transactionToBytes: Uint8Array = this.hexToBytes(transaction);
-    const transactionSize: number = transactionToBytes.length;
-    const transactionSizeLimit: number = ConfigService.get('SEND_RAW_TRANSACTION_SIZE_LIMIT');
-
-    if (transactionSize > transactionSizeLimit) {
-      throw predefined.TRANSACTION_SIZE_TOO_BIG(String(transactionSize), String(transactionSizeLimit));
+  callDataSize(tx: Transaction): void {
+    const totalCallDataSizeInBytes = tx.data.replace('0x', '').length / 2;
```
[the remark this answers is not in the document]
quiet-node: https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3722#discussion_r2086977752

packages/relay/tests/lib/precheck.spec.ts
quiet-node: Hi Kosi yeah valid point. I’ve come across this before, and the reason we can’t use the built-in assertion for this check is that JsonRpcError is a custom class that does not extend the built-in `Error` class. Chai’s `.to.throw()` matcher specifically checks for instances of `Error`, so it won’t recognize `JsonRpcError` unless it inherits from `Error`. Because of this, your suggested approach isn’t applicable in this case, at least for now.

We could extend JsonRpcError from the `Error` class to make it compatible with standard assertions, but that is not in the scope of this PR.

I had previously noted this limitation while working on the error mapper in this ticket https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3617.

Until this is addressed, we’ll need to continue using the try-catch approach.

packages/relay/tests/lib/precheck.spec.ts
quiet-node: https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3722#discussion_r2087045342

packages/relay/tests/lib/precheck.spec.ts
quiet-node: https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3722#discussion_r2087045342

konstantinabl (approved):

andrewb1269hg (approved):
Review and approve `.github/workflows/acceptance.yml`

docs/configuration.md
```diff
@@ -26,79 +26,82 @@ These properties are noted below and should be custom set per deployment.
...
+| `MULTI_SET`                                 | "false"                               | Switch between different implementation of setting multiple K/V pairs in the shared cache. True is mSet, false is pipeline                                                                                                                                                                                                                                                                                                                                                                 |
+| `JUMBO_TX_ENABLED`                          | "true"                                | Controls how large transactions are handled during `eth_sendRawTransaction`. When set to `true`, transactions up to 128KB can be sent directly to consensus nodes without using Hedera File Service (HFS), as long as contract bytecode doesn't exceed 24KB. When set to `false`, all transactions containing contract deployments use the traditional HFS approach. This feature leverages the increased transaction size limit to simplify processing of standard Ethereum transactions. |
```
acuarica, line 90: nit: should we include these alphabetically?

acuarica, line 90: not sure how the `as long as contract bytecode doesn't exceed 24KB` is relevant here. Maybe some clarification like `contract bytecode size restriction of 24KB still applies regardless of this flag`?

packages/relay/src/lib/precheck.ts
```diff
@@ -60,6 +59,9 @@ export class Precheck {
     networkGasPriceInWeiBars: number,
     requestDetails: RequestDetails,
   ): Promise<void> {
+    this.contractCodeSize(parsedTx);
+    this.callDataSize(parsedTx);
+    this.transactionSize(parsedTx);
```
acuarica, line 64: maybe we can merge these 3 checks into one method so the restrictions are more clear, instead of having to take a look in 3 different places.

docs/configuration.md
quiet-node: Perfect! Since non-critical will address this in follow PR. Noted it down here https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3753#issuecomment-2877667472

packages/relay/src/lib/precheck.ts
quiet-node: Perfect! Since non-critical will address this in follow PR. Noted it down here https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3753#issuecomment-2877667472

PR #3725, opened 2025-04-24
chore: add dependabot.yml

### LINKED ISSUE #3640: Configure dependabot.yml
> We need to set up dependabot here @PavelSBorisov I'll create an issue on the repo. 

 _Originally posted by @rbarker-dev in [#2732](https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/2732#issuecomment-2783824166)_

dependabot.yml goes in the .github folder; it should be configured for npm, docker, and github actions

PavelSBorisov:
**Description**:

Adding dependabot.yml

**Related issue(s)**:

Fixes #3640 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,42 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directories: 
+      - "/"
+      - "/dapp-example"
+      - "/docs/examples"
+      - "/k6"
+      - "/packages/*"
+      - "/tools/hardhat-example"
+      - "/tools/hardhat-viem-example"
+      - "/tools/solidity-coverage-example"
+      - "/tools/subgraph-example"
+      - "/tools/truffle-example"
+      - "/tools/waffle-example"
+      - "/tools/wagmi-example"
+      - "/tools/web3js-example"
+      - "/tools/whbar-hardhat-example"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 20
+    labels:
+      - "dependencies"
+      - "npm"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+      - "github-actions"
+    
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+      - "docker"


.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
```
quiet-node, line 3: Just curious—how do you resolve security alerts for projects that aren’t NPM-based? The tools/ folder also contains a couple of Go and Rust projects.

quiet-node:
Hey @PavelSBorisov, I noticed this PR hasn’t been updated in a while. Is it still relevant?

.github/dependabot.yml
PavelSBorisov: Good point! I've added the additional ecosystems to the dependabot config to account for those projects.

.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+      - "/"
+      - "/dapp-example"
+      - "/docs/examples"
+      - "/k6"
+      - "/packages/*"
+      - "/tools/hardhat-example"
+      - "/tools/hardhat-viem-example"
+      - "/tools/solidity-coverage-example"
+      - "/tools/subgraph-example"
+      - "/tools/truffle-example"
```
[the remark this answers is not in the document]
PavelSBorisov: Removed.

.github/dependabot.yml
```diff
@@ -0,0 +1,60 @@
...
+      - "/tools/hardhat-viem-example"
+      - "/tools/solidity-coverage-example"
+      - "/tools/subgraph-example"
+      - "/tools/waffle-example"
+      - "/tools/wagmi-example"
+      - "/tools/web3js-example"
+      - "/tools/whbar-hardhat-example"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 20
```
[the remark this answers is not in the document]
acuarica: > Should we consider grouping Dependabot updates into a single PR with multiple upgrades?

I'd say better not to, as this might make harder to trace changes when a single PR contains multiple upgrades.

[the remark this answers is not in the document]
acuarica: On the other hand, should we distinguish between `/packages/*` and the rest of project? We could set `/packages/*` to `daily` and the rest to `weekly` to avoid cluttering as you mentioned @quiet-node. The rationale is that everything under `tools/` is not critical, but `/packages/*` are.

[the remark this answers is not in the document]
quiet-node: Yeah that sounds like a good plan!

quiet-node:
Hey @PavelSBorisov are there any updates on this?

acuarica (dismissed):

rbarker-dev (dismissed):

.github/dependabot.yml
```diff
@@ -0,0 +1,70 @@
...
+    labels:
+      - "dependencies"
+      - "npm"
+
+  - package-ecosystem: "gomod"
+    directories: 
+      - "/tools/golang-example"
+      - "/tools/golang-json-rpc-tests"
+    schedule:
+      interval: "daily"
```
quiet-node, line 40: Should all of the intervals be set to "weekly" or this is fine?

acuarica: Good catch! I believe we want `weekly` here as well.

.github/dependabot.yml
```diff
@@ -0,0 +1,70 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directories:
+      - "/"
+      - "/packages/*"
```
andrewb1269hg, line 6: Do we need to explicitly declare this? Can we do `- "/*"` instead?

.github/dependabot.yml
```diff
@@ -0,0 +1,70 @@
...
+  - package-ecosystem: "docker"
+    directories: 
+      - "/"
+      - "/tools/whbar-hardhat-example"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+      - "docker"
```
andrewb1269hg, line 70:
```suggestion
      - "docker"

```

andrewb1269hg (commented):
Couple minor things.

.github/dependabot.yml
PavelSBorisov: `/` is resolved as just the root dir, `/*` would be resolved as all subdirs of the root dir which would invalidate the directory separation that we are looking for in this repo.

andrewb1269hg (approved):

acuarica:
Hi @PavelSBorisov, we started to see the effects of this config file. We noticed that many, many PRs were created. The problem is that these many PRs degrade GHA performance substantially. Right after the PRs are created, workflows stop running for while. Is there is a way to create PRs in batches? Or at least for now, maybe even decrease the number of PRs created so we can continue working while dependabot PR workflows are running.

I noticed this article https://docs.github.com/en/enterprise-cloud@latest/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates, maybe we can use the `cooldown` setting? Any other ideas?

acuarica:
Hi Pavel, we discussed this internally with the team. We would like to apply your suggestions, that is

- 1 day cooldown
- and lower the amount of open PRs (from 20 to 5)

Over time we can increase this limit once most of the repo's dependencies are up to date. Would this be possible?

PR #3731, opened 2025-04-28
fix: fixes check to remove duplicates from transactionArray

### LINKED ISSUE #3724: Duplicate transactions in eth_getBlockByNumber responses
### Description

Description:
The Axelar team has discovered an issue with the JSON-RPC API where transactions are being duplicated in block responses. When calling eth_getBlockByNumber, the same transaction hash appears multiple times (exactly 10 times in our test case) within the same block response.

Example request:
curl --silent --location 'https://testnet.hashio.io/api' --header 'Content-Type: application/json' --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x11B9DBA",true],"id":1}'

Current behavior:
- The response shows transaction hash 0x3af2ba7854d61a9075330aa7436139cfda8e5dc185465330dac9547912c36e66 repeated 10 times
- The response shows transaction hash 0x40d7ca806019a2b67379e327399a5c242a06f1be5c13321aa31d53480448cf9a repeated 10 times
- Transaction indices are 0xe (14) and 0x10 (16), which are non-sequential

Expected behavior:
- Each transaction should appear exactly once in the block
- Transaction indices should be sequential starting from 0

Impact:
This issue breaks components that assume unique transactions per block, which is standard in Ethereum. The duplication causes processing errors in our application and may impact other services that rely on standard EVM behavior.

Additional information:
The non-sequential transaction indices aren't directly breaking Axelar components, but they deviate from the standard EVM behavior and might cause issues for other tools.

### Steps to reproduce

1. Call eth_getBlockByNumber for block 0x11B9DBA with transactions=true
2. Observe that transaction hashes are duplicated in the response

### Additional context

_No response_

### Hedera network

testnet

### Version

v0.67.0

### Operating system

None

simzzz:
**Description**:
This PR fixes a problem in the `getBlock()` method which caused duplicate transactions to be returned. 

The old code `transactionArray = showDetails ? transactionArray : _.uniq(transactionArray);` didn't account for the situation where duplicate transaction could be returned in some cases where one transaction emits multiple events (e.g. minting multiple NFT serials in one transaction). In that case, duplicate transactions would be generated from the logs. I have changed this line to account for that and exclude duplicates.

**Related issue(s)**:

Fixes #3724

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -2930,8 +2930,7 @@ export class EthImpl implements Eth {
     }
 
     transactionArray = this.populateSyntheticTransactions(showDetails, logs, transactionArray, requestDetails);
-    transactionArray = showDetails ? transactionArray : _.uniq(transactionArray);
-
+    transactionArray = _.uniqWith(transactionArray, _.isEqual);
     const formattedReceipts: IReceiptRootHash[] = ReceiptsRootUtils.buildReceiptRootHashes(
       transactionArray.map((tx) => (showDetails ? tx.hash : tx)),
       contractResults,
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -2931,7 +2931,6 @@ export class EthImpl implements Eth {
     }
 
     transactionArray = this.populateSyntheticTransactions(showDetails, logs, transactionArray, requestDetails);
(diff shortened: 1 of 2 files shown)


Ferparishuertas (approved):
LGTM

quiet-node (approved):
LGTM

PR #3733, opened 2025-04-29
feat: HTSConnector fuzz testing

### LINKED ISSUE #3734: HTS connector add fuzzing to ensure no edge cases with uint to int64
### Problem

Hedera Token Service only works with int64 values, this means that the max number of tokens which can be represented ( in 8 decimals ) is about 92 million tokens:
- int64.max = 9,223,372,036,854,775,807 atomic units
- decimals 8
- Max tokens: 92,233,720.36854775
On other chains, for example ethereum, service usually use uint256 which means much higher values can be presented. This can lead to issues in cases where protocols are porting over high liquidity from evm chains to hedera. We need to:

We need to Fuzz the HTSConnector _credit function for the following cases

A- Will always revert on sending higher values which should wrap around to a negative number
B- Make sure there are not cases where there are silent failures and the wrapping around leads to silent failures where the users receives less amount than what they sent. This would be because the casting logic didn't result in a negative number (which should revert) but in a small number.
uint64 value = uint64(2^64 + 1);  //1
int64 result = int64(value);   //1
Here is a link to the "_credit" function. 

https://github.com/hashgraph/hedera-smart-contracts/blob/f092d859201b015517d79dd3f60e98af9dc4254d/lib/layer-zero/HTSConnector.sol

### Solution

Add test.

### Alternatives

_No response_

natanasow:
**Description**:

Hedera Token Service only works with int64 values, which means that the max number of tokens that can be represented (in 8 decimals) is about 92 million tokens:
- int64.max = 9,223,372,036,854,775,807 atomic units
- decimals 8
- max tokens: 92,233,720.36854775

On other chains, for example, Ethereum, services usually use uint256, which means much higher values can be presented. This can lead to issues in cases where protocols are porting over high liquidity from EVM chains to Hedera. We need to fuzz the HTSConnector `_credit` function.

**Related issue(s)**:

Fixes #3734

**Notes for reviewer**:

**Checklist**


simzzz (approved):

Ferparishuertas (approved):
LGTM

PR #3744, opened 2025-05-06
feat: added support for debug_traceBlockByNumber

### LINKED ISSUE #3742: Implement debug_traceBlockByNumber
### Problem

debug_traceBlockByNumber is currently not supported

### Solution

Add new implementation to support debug_traceBlockByNumber

### Alternatives

_No response_

quiet-node:
**Description**:
This PR adds a new debug method `debug_traceBlockByNumber` to the RPC interface. The method enables tracing of all transactions within a specified block using one of the supported tracers.
* **Parameters**:

  * `blockNumber` (required): Block number or tag (e.g., `latest`, `pending`).
  * `tracerOptions` (optional):

    * `tracer`: Type of tracer (`callTracer` or `prestateTracer`).
    * `tracerConfig`: Additional configuration (e.g., `onlyTopCall`).

* **Response**

  * **callTracer**: Returns a call frame structure with fields like `type`, `from`, `to`, `value`, `gas`, `gasUsed`, `input`, `output`, and any sub-calls.
  * **prestateTracer**: Returns a mapping of addresses to account states, including `balance`, `nonce`, `code`, and `storage`.

Fixes #3742


Ferparishuertas (dismissed):
LGTM
Curious to see Acceptance test were not covering debug

quiet-node:
> LGTM Curious to see Acceptance test were not covering debug

@Ferparishuertas yes I was surprised too. We only have debug_traceTransaction and debug_traceBlockByNumber atm. For some reason, e2e tests for debug_traceTransaction was missed. There are quite some improvements we can make for debug_traceTransaction as well so I added the e2e test requirement to this issue https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3739

Ferparishuertas (dismissed):
LGTM

natanasow (dismissed):
LGTM.

natanasow (approved):

Ferparishuertas (dismissed):
LGTM

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1198,6 +1220,31 @@ export class MirrorNodeClient {
...
   }
 
+  public async getContractState(address: string, requestDetails: RequestDetails, timestamp?: string) {
+    const limitOrderParams: ILimitOrderParams = this.getLimitOrderQueryParam(
+      constants.MIRROR_NODE_QUERY_LIMIT,
+      constants.ORDER.DESC,
+    );
+    const queryParamObject = {};
+
+    if (timestamp) {
```
acuarica, line 1230: isn't this unnecessary given `setQueryParam` already perform this check?

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -1372,10 +1422,11 @@ export class MirrorNodeClient {
         }),
       );
 
-    if (searchableTypes.find((t) => t === constants.TYPE_CONTRACT)) {
-      const contract = await this.getContract(entityIdentifier, requestDetails, retries).catch(() => {
+    if (searchableTypes.find((t) => t.toLowerCase() === constants.TYPE_CONTRACT.toLowerCase())) {
```
acuarica, line 1425: why lowercase? this param is always provided by, why don't we just change the call sites?

packages/relay/src/lib/types/mirrorNode.ts
```diff
@@ -202,3 +202,58 @@ export class MirrorNodeTransactionRecord {
...
+  type: number;
+  v: number;
+  nonce: number;
+}
+
+export interface ContractAction {
+  call_depth: number;
+  call_operation_type: 'CALL' | string;
+  call_type: 'CALL' | string;
+  caller: string; // e.g., '0.0.2661815'
```
acuarica, line 245: should we include these as TSdocs so they are visible on usage?

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: Oh that's a good catch! Thanks for pointing it out!

quiet-node: Removed the unnecessary checker.

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: This change would make the function more flexible and robust when handling the `type` input arguments. Since inputs can come in different cases, like /contract/results/action returning types in uppercase, and other times the input args can be lowercase. So, normalizing them within the function allow consistent behavior without requiring manual case adjustments at each call site, which I believe adding a certain level of convenience and we don't need to care what types to pass in.

packages/relay/src/lib/types/mirrorNode.ts
quiet-node: Yeah true it'd make it more official and cleaner within TSDoc. Updated.

packages/relay/src/lib/clients/mirrorNodeClient.ts
acuarica: Mmm, we could, for example, change the value in `constants` to uppercase to match those coming from `ContractAction`. But more importantly, we could also refine the type of `searchableTypes` to

```typescript
searchableTypes: ('contract' | 'account' | 'token')[] = [constants.TYPE_CONTRACT, constants.TYPE_ACCOUNT, constants.TYPE_TOKEN],
```

We would also need to update those constant definitions

```typescript
  TYPE_CONTRACT: 'contract' as const,
  TYPE_ACCOUNT: 'account' as const,
  TYPE_TOKEN: 'token' as const,
```

or using uppercase as mentioned above. This way, I believe, the intent is more clear, and we avoid the conversion. Moreover, we could simplify this even further, from

```typescript
searchableTypes.find((t) => t.toLowerCase() === constants.TYPE_CONTRACT.toLowerCase())
```

to

```typescript
searchableTypes.includes(constants.TYPE_CONTRACT)
```

which again, I believe is more readable.

In any case, feel free to take it as optional. Going to resolve this.

konstantinabl (changes requested):
Few changes needed

packages/relay/src/lib/types/mirrorNode.ts
```diff
@@ -202,3 +202,115 @@ export class MirrorNodeTransactionRecord {
     this.valid_start_timestamp = transactionRecord.valid_start_timestamp;
   }
 }
+
+/**
+ * Represents the result of a contract call or transaction as returned by the Hedera Mirror Node.
+ *
+ * @property address - The address involved in the contract call.
```
acuarica, line 209: I'd inline this doc comments, that way it's easier to read and follow, and avoid duplicating the whole struct.

packages/relay/tests/lib/mirrorNodeClient.spec.ts
```diff
@@ -1768,4 +1768,141 @@ describe('MirrorNodeClient', async function () {
...
+
+    it('should throw error for invalid contract address', async () => {
+      const invalidAddress = '0x123';
+      mock.onGet(`contracts/${invalidAddress}/state?limit=100&order=desc`).reply(400, JSON.stringify(null));
+      let errorRaised = false;
+      try {
+        await mirrorNodeInstance.getContractState(invalidAddress, requestDetails);
+      } catch (error) {
+        errorRaised = true;
+        expect(error.message).to.equal('Request failed with status code 400');
```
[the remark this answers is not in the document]
quiet-node: Very good point! Updated with the suggestion.

packages/relay/tests/lib/mirrorNodeClient.spec.ts
```diff
@@ -1768,4 +1768,141 @@ describe('MirrorNodeClient', async function () {
...
+    });
+
+    it('should throw error for server error', async () => {
+      mock.onGet(contractStatePath).reply(500, JSON.stringify({ error: 'Server error' }));
+      let errorRaised = false;
+      try {
+        await mirrorNodeInstance.getContractState(contractAddress, requestDetails);
+      } catch (error) {
+        errorRaised = true;
+        expect(error.message).to.equal('Request failed with status code 500');
```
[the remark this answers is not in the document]
quiet-node: Very good point! Updated with the suggestion.

packages/server/tests/integration/server.spec.ts
```diff
@@ -2784,6 +2785,432 @@ describe('RPC Server', function () {
...
+        try {
+          await testClient.post('/', {
+            jsonrpc: '2.0',
+            method: 'debug_traceBlockByNumber',
+            params: ['0x999999999999'], // block number that doesn't exist
+            id: '2',
+          });
+
+          Assertions.expectedError();
+        } catch (error: any) {
```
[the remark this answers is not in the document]
quiet-node: Very good point! Updated with the suggestion.

packages/server/tests/integration/server.spec.ts
```diff
@@ -2784,6 +2785,432 @@ describe('RPC Server', function () {
...
+          BaseTest.invalidParamError(error.response, Validator.ERROR_CODE, MISSING_PARAM_ERROR + ' 0');
+        }
+      });
+
+      it('should fail with invalid block number parameter', async () => {
+        try {
+          await testClient.post('/', {
```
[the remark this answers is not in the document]
quiet-node: Very good point! Updated with the suggestion.

packages/relay/tests/lib/debug.spec.ts
```diff
@@ -523,4 +526,585 @@ describe('Debug API Test Suite', async function () {
...
+        const getHistoricalBlockResponseStub = sinon.stub().resolves(null);
+        sinon.stub(CommonService.prototype, 'getHistoricalBlockResponse').callsFake(getHistoricalBlockResponseStub);
+
+        try {
+          await debugService.traceBlockByNumber(
+            blockNumber,
+            { tracer: TracerType.CallTracer, tracerConfig: { onlyTopCall: false } },
+            requestDetails,
+          );
+          expect.fail('Expected the traceBlockByNumber to throw an error but it did not');
```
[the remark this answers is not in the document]
quiet-node: Very good point! Updated with the suggestion.

packages/relay/src/lib/types/ITracerConfigWrapper.ts
```diff
@@ -1,9 +1,14 @@
...
 import { TracerType } from '../constants';
-import { ITracerConfig } from './ITracerConfig';
+import { ICallTracerConfig, ITracerConfig } from './ITracerConfig';
 
 export interface ITracerConfigWrapper {
   tracer?: TracerType;
   tracerConfig?: ITracerConfig;
 }
+
+export interface IBlockTracerConfig {
```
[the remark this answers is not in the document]
quiet-node: Yeah true I thought since the current flow still use I-prefix interfaces so I thought I'd make it consistent for now but you're right let's not introduce more of this. Updated to remove I prefix for this new interface.

packages/relay/src/lib/debug.ts
```diff
@@ -338,11 +473,132 @@ export class DebugImpl implements Debug {
...
+   *
+   * @async
+   * @param {string} transactionHash - The hash of the transaction to trace.
+   * @param {boolean} onlyTopCall - When true, only includes accounts involved in top-level calls.
+   * @param {RequestDetails} requestDetails - Details for request tracking and logging.
+   * @returns {Promise<object>} A Promise that resolves to an object containing the pre-state information.
+   *                           The object keys are EVM addresses, and values contain balance, nonce, code, and storage data.
+   * @throws {Error} Throws a RESOURCE_NOT_FOUND error if contract results cannot be retrieved.
+   */
+  async prestateTracer(
```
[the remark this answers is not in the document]
quiet-node: Cool yeah thanks for the solid concern. The method is indeed somewhat lengthy, but I wouldn’t say it is unnecessarily or unusally complex. Essentially, it just orchestrates some asynchronous calls, handles errors, and aggregates data.

The logic follows a mostly sequential and clear flow, with each step being well-commented like cache check → fetch actions → filter actions, and so on. I belive it’s not particularly difficult to follow, it just requires a careful read,

The main reason I’m not in favor of splitting it further is that doing so would add additional layers of indirection without significantly improving clarity, nor would it necessarily make the code cleaner, as it would primarily involve moving the existing logic into separate methods without meaningful simplification. It would certainly be helpful if these methods were reused as sub-steps elsewhere, but otherwise, I believe it’s reasonable to keep it as is.

Let me know you thoughts on this. Happy to discuss.

packages/relay/src/lib/types/mirrorNode.ts
quiet-node: yeah definitely it's indeed the good practice for this case. Good point. Updated.

packages/relay/src/lib/debug.ts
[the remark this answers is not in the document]
konstantinabl: Its fine by me, not too complex for now, so we may leave it as it is

packages/server/tests/acceptance/debug.spec.ts
```diff
@@ -0,0 +1,401 @@
...
+      const invalidTracerConfig = { tracer: 'InvalidTracer', onlyTopCall: false };
+
+      try {
+        await relay.call(
+          DEBUG_TRACE_BLOCK_BY_NUMBER,
+          [numberTo0x(deploymentBlockNumber), invalidTracerConfig],
+          requestId,
+        );
+        // If we get here, the test should fail because we expect an error
+        expect.fail('Expected an error to be thrown for invalid tracer type');
```
konstantinabl, line 386: Can we improve this and remove the try/catch as well :?

quiet-node: Yeah, good catch, not sure why I left that try/catch there. Anyway, to avoid using try/catch, it goes back to the .to.throw() issue we discussed and how it’s currently not working with our JsonRpcError.

For now, just to keep things more consistent, I’ve switched to using relay.callFailing, though it also uses try/catch under the hood.

package.json
```diff
@@ -70,6 +70,7 @@
     "acceptancetest:rpc_api_schema_conformity": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@api-conformity' --exit",
     "acceptancetest:serverconfig": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@server-config' --exit",
     "acceptancetest:send_raw_transaction_extension": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@sendRawTransactionExtension' --exit",
+    "acceptancetest:debug": "nyc ts-mocha packages/server/tests/acceptance/index.spec.ts  -g '@debug' --exit",
```
acuarica, line 73: qq: where are we running these scripts with `nyc`? where is the coverage output used?

quiet-node: The workflow uses it https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3744/files#diff-20178bfee541d27c8c2296d927b63f9c6fec922b9ead4965f2e35958204baebbR25. 

I believe the coverage is used as part of the workflow as well.

andrewb1269hg (approved):
Review and approve `.github/workflows/acceptance.yml`

PR #3750, opened 2025-05-08
feat: improve cache layer

### LINKED ISSUE #1058: Improve cache support for APIs
### Problem

The relay support caching for some endpoints and not others. We should assess which would no benefit from it and which may need TTL times updated based on usage

### Solution

Add cache support to heavily utilized methods in which the response does not change.

2025 - MARCh `PLEASE CONDUCT a new review, and design doc in notion. detailing:

- which methods have cache which not, and if yes, the TTL. and if its configurable by config or not
- Distinguish where we cache, at relay level, or close by the MN logic

### Extras

- Group all cache config at our DOC; configuracion sectionm making it understandable, as nowadays it is mixed with other values, and for example MAX_CACHE is only for LRU.
- Group inside cache config, by local and shared cache

_LEGACY_

- [ ] Assess endpoint metrics, what are the heaviest utilized
- [ ] Assess endpoint responses, which are static based on params which do not yet have caching
- [ ] Add caching logic with consideration for mirror node refresh times to ensure details are still fresh
  - [x] #1059
  - [x] #1210
  - [x] #1057  
  - [ ] eth_chainId
  - [ ] eth_feeHistory
  - [ ] net_version
  - [ ] eth_estimateGas
  - [x] #1212
  - [x] eth_getBlockByNumber
  - [x] #1221
- [ ] Permanently cache Earliest block
- [ ] Add cache support for some consensus node calls
  - [x] Add caching logic to getTinyBarGasFee() 

### Alternatives

_No response_

natanasow:
**Description**:

The first problem, in the current codebase, we’ve tightly coupled the cache layer within the eth.ts, which leads to the spaghetti code and violates the SOLID principles. Also, makes it difficult to introduce new people to it and slows down the development process. 

The second problem is related to the missing cache coverage. There are some routes where cache support can be added, reducing the load to the MN and improving overall JSON-RPC relay performance.

Tasks:
- [x] - Add the decorator
- [x] - Deprecate the usage of `CacheService` and use the new cache decorator instead 
- [x] - Add doc
- [x] - Fix tests
- [x] - Remove old implementations within the methods
- [x] - Add caching where it's applicable

**Related issue(s)**:

Fixes #1058 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,157 @@
...
+ *    index: '0',
+ *    value: 'pending|safe'
+ * }]
+ *
+ * @param args
+ * @param params
+ */
+const shouldSkipCachingForSingleParams = (args: any, params: any = []) => {
+  for (const item of params) {
+    const values = item.value.split('|');
```
quiet-node, line 41: So with this .split('|'), are we saying that the item.value will always be block tags like latest|finanlized|pending|etc.? If so should we update the tpe of item.value to be more explicit instead of string? If it's not always block tags with `|` delimeter should update this line to be more robust?

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,157 @@
...
+  for (const [, value] of Object.entries(args)) {
+    if (value?.constructor?.name == 'RequestDetails') {
+      requestId = value['requestId'];
+      ipAddress = value['ipAddress'];
+      connectionId = value['connectionId'];
+      break;
+    }
+  }
+
+  return new RequestDetails({ requestId, ipAddress, connectionId });
```
quiet-node, line 123: QQ: why do we need to reconstruct RequestDetails? Since the RequestDetails object in the arguments list is already an instance of RequestDetails, can’t we simply return that object without reconstructing it?

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,157 @@
...
+      const values = (value as string).split('|');
+      if (values.indexOf(input[key]) > -1) {
+        return true;
+      }
+    }
+  }
+
+  return false;
+};
+
```
quiet-node, line 82: Same for this one—let’s make it more descriptive to avoid guesswork during future revisits.

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,157 @@
...
+      ipAddress = value['ipAddress'];
+      connectionId = value['connectionId'];
+      break;
+    }
+  }
+
+  return new RequestDetails({ requestId, ipAddress, connectionId });
+};
+
+export function cache(cacheService: CacheService, options: CacheOptions = {}) {
```
quiet-node, line 126: Let’s add TSDoc for this main decorator method, explaining how it works and how it should be applied to methods. Including an @example would also help illustrate its usage.

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -102,6 +107,29 @@ export class CacheService {
...
+  /**
+   * Get a cache service instance
+   *
+   * @param type
+   * @param registry
+   * @param reservedKeys
+   */
+  public static getInstance(type: CACHE_LEVEL, registry: Registry = new Registry(), reservedKeys: Set<string> = new Set()): CacheService {
+    if (!this.instances[type]) {
+      const logger = pino({ level: 'silent' });
```
quiet-node, line 126: Should we use global LOG_LEVEL for this guy instead of hardcode it? Also why do we want it to be `silent` which can omit debugging logs?

packages/relay/src/lib/eth.ts
```diff
@@ -173,6 +173,10 @@ export class EthImpl implements Eth {
     2: { type: ParamType.ARRAY, required: false },
   })
   @rpcParamLayoutConfig(RPC_LAYOUT.custom((params) => [Number(params[0]), params[1], params[2]]))
+  @cache(CacheService.getInstance(CACHE_LEVEL.L1), {
+    skipParams: [{ index: '2', value: constants.NON_CACHABLE_BLOCK_PARAMS }],
```
quiet-node, line 177: Looks like this hould be index '1' instead, correct? Param at index 1 is the block number or block tag param.

packages/relay/src/lib/eth.ts
```diff
@@ -173,6 +173,10 @@ export class EthImpl implements Eth {
...
+    ttl: constants.CACHE_TTL.FIFTEEN_MINUTES,
+  })
```
quiet-node, line 179: Also I see that many endpoints like estimateGas, filter endpoints, getStorageAt, getTransactionCount, call, sendRawTransaction are not cached. What's the reason behind it?

packages/relay/tests/lib/eth/eth-helpers.ts
```diff
@@ -15,7 +15,7 @@ import { IPAddressHbarSpendingPlanRepository } from '../../../src/lib/db/reposit
 import { EthImpl } from '../../../src/lib/eth';
 import { ContractService } from '../../../src/lib/services';
```
quiet-node, line 16: unused imports

packages/relay/tests/lib/eth/eth_getBlockReceipts.spec.ts
```diff
@@ -7,8 +7,8 @@ import sinon from 'sinon';
 
 import { numberTo0x } from '../../../dist/formatters';
 import { SDKClient } from '../../../src/lib/clients';
-import constants from '../../../src/lib/constants';
 import { EthImpl } from '../../../src/lib/eth';
+import { ContractService } from '../../../src/lib/services';
```
quiet-node, line 11: unused imports

packages/relay/tests/lib/eth/eth_getBlockReceipts.spec.ts
```diff
@@ -53,9 +53,9 @@ describe('@ethGetBlockReceipts using MirrorNode', async function () {
   this.beforeEach(async () => {
     // reset cache and restMock
     await cacheService.clear(requestDetails);
-    currentGasPriceStub = sinon.stub(ethImpl['common'], 'getCurrentGasPriceForBlock').resolves('0x25');
+    currentGasPriceStub = sinon.stub(ethImpl.common, 'getCurrentGasPriceForBlock').resolves('0x25');
```
quiet-node, line 56: why do we swith back to .common instead of ['common']? It was used to avoid lint errors for accessing private methods, right?

Anywho we should have a more robust solution for this maybe in tsconfig or something. This is just a discussion point not for this PR.

packages/relay/tests/lib/eth/eth_getBlockReceipts.spec.ts
```diff
@@ -53,9 +53,9 @@ describe('@ethGetBlockReceipts using MirrorNode', async function () {
...
     extractBlockNumberOrTagStub = sinon
-      .stub(ethImpl['contractService'], 'extractBlockNumberOrTag')
+      .stub(ethImpl.contractService, 'extractBlockNumberOrTag')
```
quiet-node, line 58: same here and below let's switch it back to avoid unnecessary changes.

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,157 @@
...
+      if (cachedResponse) {
+        return cachedResponse;
+      }
+
+      const result = await method.apply(this, arguments);
+      if (
+        result &&
+        !shouldSkipCachingForSingleParams(arguments, options?.skipParams) &&
+        !shouldSkipCachingForNamedParams(arguments, options?.skipNamedParams)
+      ) {
```
quiet-node, line 144: I see we skipped caching requests with block tags. Should we involve CommonService and make a quick call to mirror node to translate the block tag to block num and add it to cache key so we don't have to skip caching for those scenarios?

quiet-node (commented):
Great work! Much more unified now! Left some Qs and suggestions

packages/relay/src/lib/eth.ts
quiet-node: Oh also for debug_traceTransaction as well.

natanasow: - `eth_estimateGas` - Always should use the latest block IMO, so we can not cache it, or at least it doesn't make sense to me. If you insist, I can add it.
- filter - Endpoints have dynamic responses, could we cache them at all?
- `eth_getStorageAt` - Added.
- `eth_getTransactionCount` - Added.
- `eth_call` - Added.
- `eth_sendRawTransaction` - I don't think it's needed, it's almost impossible to generate the same tx hashes twice, but if you insist, I will add it.

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+interface CacheOptions {
+  skipParams?: CacheSingleParam[],
+  skipNamedParams?: CacheNamedParams[],
+  ttl?: number,
+}
+
+/**
+ * Skip single params in the following format
```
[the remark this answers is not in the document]
natanasow: Fixed.

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * [{
+ *    index: '0',
+ *    value: 'pending|safe'
+ * }]
+ *
+ * @param args
+ * @param params
+ */
+const shouldSkipCachingForSingleParams = (args: any, params: any = []) => {
```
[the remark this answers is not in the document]
natanasow: Added.

packages/relay/src/lib/decorators/cache.decorator.ts
natanasow: Fixed.

packages/relay/src/lib/decorators/cache.decorator.ts
natanasow: TS doc added.

packages/relay/src/lib/decorators/cache.decorator.ts
natanasow: Added.

packages/relay/src/lib/services/cacheService/cacheService.ts
natanasow: Switch to the global `LOG_LEVEL`

packages/relay/src/lib/eth.ts
natanasow: Yep, fixed.

packages/relay/tests/lib/eth/eth-helpers.ts
natanasow: Removed.

packages/relay/tests/lib/eth/eth_getBlockReceipts.spec.ts
natanasow: Same.

packages/relay/tests/lib/eth/eth_getBlockReceipts.spec.ts
natanasow: Rolled back to the previous version.

packages/relay/tests/lib/eth/eth_getBlockReceipts.spec.ts
natanasow: Rolled back to the previous version.

packages/relay/src/lib/decorators/cache.decorator.ts
natanasow: To use the common service here, we should pass it as a dependency or create a new instance here, which requires an MN instance.

This is quite a big effort, but if this is the better solution, I'll implement it.

packages/relay/src/lib/services/ethService/blockService/BlockService.ts
```diff
@@ -94,16 +94,9 @@ export class BlockService implements IBlockService {
     const requestIdPrefix = requestDetails.formattedRequestId;
     this.logger.trace(`${requestIdPrefix} getBlockByHash(hash=${hash}, showDetails=${showDetails})`);
 
-    const cacheKey = `${constants.CACHE_KEY.ETH_GET_BLOCK_BY_HASH}_${hash}_${showDetails}`;
-    let block = await this.cacheService.getAsync(cacheKey, BlockService.ethGetBlockByHash, requestDetails);
```
konstantinabl, line 98: after these changes there are 3 unused variables in the class
cacheService, eth_getBlockByHash and eth_getBlockByNumber

packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
```diff
@@ -296,39 +296,10 @@ export class CommonService implements ICommonService {
...
     const blocks = blocksResponse !== null ? blocksResponse.blocks : null;
     if (Array.isArray(blocks) && blocks.length > 0) {
-      const currentBlock = numberTo0x(blocks[0].number);
-      // save the latest block number in cache
-      await this.cacheService.set(
-        cacheKey,
-        currentBlock,
-        CommonService.latestBlockNumber,
-        requestDetails,
-        this.ethBlockNumberCacheTtlMs,
```
konstantinabl, line 328: This variable is unused now

packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
```diff
@@ -525,24 +496,9 @@ export class CommonService implements ICommonService {
...
-
-      if (!gasPrice) {
-        gasPrice = Utils.addPercentageBufferToGasPrice(await this.getGasPriceInWeibars(requestDetails));
 
-        await this.cacheService.set(
-          constants.CACHE_KEY.GAS_PRICE,
-          gasPrice,
-          constants.ETH_GAS_PRICE,
-          requestDetails,
-          this.ethGasPriceCacheTtlMs,
```
konstantinabl, line 543: this one is also unused

packages/relay/src/lib/services/ethService/feeService/FeeService.ts
```diff
@@ -121,32 +121,13 @@ export class FeeService implements IFeeService {
...
-          feeHistory = await this.getFeeHistory(
-            blockCount,
-            newestBlockNumber,
-            latestBlockNumber,
-            rewardPercentiles,
-            requestDetails,
-          );
-        }
-        if (newestBlock != constants.BLOCK_LATEST && newestBlock != constants.BLOCK_PENDING) {
-          await this.cacheService.set(
```
konstantinabl, line 142: Cache service class property is now unused

konstantinabl (changes requested):
Few nits

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,193 @@
...
+ *     }, {
+ *       name: 'toBlock', value: 'safe|finalized'
+ *     }],
+ *   }]
+ */
+const shouldSkipCachingForNamedParams = (args: IArguments, params: CacheNamedParams[] = []): boolean => {
+  for (const item of params) {
+    const input = args[item.index];
+    // @ts-ignore
+    const skipList = Object.assign({}, ...params.filter(el => el.index == item.index)[0].fields.map(el => {
```
konstantinabl, line 77: Can we add some comment on how this is working and what exactly it does, for me its hard to grasp immediately

packages/relay/src/lib/eth.ts
quiet-node: Ah yeah definitely it totally makes sense! Thanks for confirming!

packages/relay/src/lib/decorators/cache.decorator.ts
quiet-node: Yeah I think it'd be a nice-to-have feature however let's create an enhancement ticktet and address it there. I think it'd be out of this PR's scope.

quiet-node: Yeah I think it'd be a nice-to-have feature however let's create an enhancement ticktet and address it there. I think it'd be out of this PR's scope.

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * If no such instance is found, it returns a new `RequestDetails` object with empty defaults.
+ *
+ * @param args - The arguments object from a function (typically `IArguments`).
+ * @returns The first found `RequestDetails` instance, or a new one with default values if none is found.
+ */
+const extractRequestDetails = (args: IArguments): RequestDetails => {
+  for (const [, value] of Object.entries(args)) {
+    if (value?.constructor?.name == 'RequestDetails') {
+      // @ts-ignore
```
quiet-node, line 137: Let's add an explanation on why we need ts-ignore here

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,193 @@
...
+          method,
+          requestDetails,
+          options?.ttl ?? ConfigService.get('CACHE_TTL')
+        );
+      }
+
+      return result;
+    };
+  };
+}
```
quiet-node, line 193: Ah meant to include this in the previous round of review but forgot. Please add test suites for this decorator I don't think there's one, correct?

quiet-node (commented):
Great work! Only two items.

packages/relay/src/lib/decorators/cache.decorator.ts
natanasow: Got it, here is the follow-up improvement task https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3756.

packages/relay/src/lib/services/ethService/blockService/BlockService.ts
natanasow: Removed.

packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
natanasow: Removed.

packages/relay/src/lib/services/ethService/feeService/FeeService.ts
natanasow: Removed.

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -0,0 +1,193 @@
...
+ *       name: 'fromBlock', value: 'pending|safe'
+ *     }, {
+ *       name: 'toBlock', value: 'safe|finalized'
+ *     }],
+ *   }]
+ */
+const shouldSkipCachingForNamedParams = (args: IArguments, params: CacheNamedParams[] = []): boolean => {
+  for (const item of params) {
+    const input = args[item.index];
+    // @ts-ignore
```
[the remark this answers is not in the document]
natanasow: Optimized for ts.

packages/relay/src/lib/decorators/cache.decorator.ts
natanasow: Optimized for ts.

packages/relay/src/lib/decorators/cache.decorator.ts
natanasow: Added.

konstantinabl (approved):
LGTM

packages/relay/tests/lib/decorators/cache.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+import { ConfigService } from '@hashgraph/json-rpc-config-service/dist/services';
+import { expect } from 'chai';
+import sinon from 'sinon';
+
+import { cache } from '../../../dist/lib/decorators';
+import { CacheService } from '../../../dist/lib/services/cacheService/cacheService';
+import { RequestDetails } from '../../../src/lib/types';
+
+describe('cache decorator', () => {
```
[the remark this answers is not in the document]
natanasow: I thought about several solutions, but as far as we're exporting a function and not a module/class, we have to export explicitly all private methods as well. I did that under `__test__` namespace, wrapped into `__private` methods constant.

Wrapping the `cache` function into a class will result in a changed annotation for caching - `@Cache.cache(...)`, e.g., an endpoint in `eth.ts` would look like this :point_down: which is not so cool IMO.
```ts
  @Cache.cache(CacheService.getInstance(CACHE_LEVEL.L1), {
    skipParams: [{ index: '1', value: constants.NON_CACHABLE_BLOCK_PARAMS }],
  })
  async getBalance(
    account: string,
    blockNumberOrTagOrHash: string | null,
  ): Promise<string> {
    return this.accountService.getBalance(account, blockNumberOrTagOrHash, requestDetails);
  }
```

Do you agree with the solution I already implemented? If you have another proposal, feel free to ping me.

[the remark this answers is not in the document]
quiet-node: Ah, I completely get what you mean. It’s definitely a dilemma since we’re exporting a functional component instead of a class. I initially misunderstood that. And yeah, using @Cache.cache isn’t quite cool loool

The __test__ namespace does look clean, but it feels like a new, somewhat ad-hoc concept for the codebase, which we haven’t used before. I’m personally okay with it, but I’d prefer not to introduce a new pattern just yet.

Would it make sense to wrap only the private methods in a helper class instead of exporting them under the __test__ namespace? That way, we can still keep the cache as a regular decorator on the functional component, while encapsulating the private methods inside a helper class.

We could place the helper class in the same file to avoid extra indirection, and maybe move the actual cache decorator method to the top for easier visibility.

This way, we can test those methods without introducing new concepts. And speaking of new concepts, when the time is right (i.e. at some point in future, outside of this PR), we could explore introducing a testable decorator for private methods, so we have a more unified and consistent approach.

quiet-node (commented):
Great one! Just a couple of items. Also I notice that the unit tests now generating more logs than usual. If you run the unit tests from main, no logs would be printed out but if running the tests from this branch I can see lots of logs. Maybe the log level got tweaked in your branch or some sort but let's make sure we don't have unnecesary changes. 

Another small thing is the files don't seem to prettier-formatted.

natanasow:
@quiet-node regarding this https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3750#discussion_r2091657940.

I already tried that way but it doesn't seem better :disappointed:

```ts
export class CacheHelper {
  private static shouldSkipCachingForSingleParams() {
    // body
  }
  private static shouldSkipCachingForNamedParams() {
    // body
  }
  private static generateCacheKey() {
    // body
  }
  private static extractRequestDetails() {
    // body
  }
}

export function cache(cacheService: CacheService, options: CacheOptions = {}) {
  // here, we can not use it like that because it's a private method out of this function scope
  CacheHelper.shouldSkipCachingForSingleParams();
  
  // using it like that should work, but it breaks the ts convention
  CacheHelper['shouldSkipCachingForSingleParams']
}
```

The only "workaround" that comes to my mind is something like that :point_down: 
```ts
const shouldSkipCachingForSingleParams = () => {
  // body
}

const shouldSkipCachingForNamedParams = () => {
  // body
}

const generateCacheKey = () => {
  // body
}
const extractRequestDetails = () => {
  // body
}

// this class will be used as a proxy to expose the private functions within this file
export class CacheHelper {
  private static shouldSkipCachingForSingleParams() {
    return shouldSkipCachingForSingleParams();
  }
  private static shouldSkipCachingForNamedParams() {
    return shouldSkipCachingForNamedParams();
  }
  private static generateCacheKey() {
    return generateCacheKey();
  }
  private static extractRequestDetails() {
    return extractRequestDetails();
  }
}

export function cache(cacheService: CacheService, options: CacheOptions = {}) {
  // here, we will use the global function instead of the class private method
  // (CacheHelper['shouldSkipCachingForSingleParams']) to avoid calling a private method
  shouldSkipCachingForSingleParams();
}
```

And yep, mixing up components with classes is a mess here, also we should choose the approach that suits our codebase instead of introducing a new one IMO.

packages/relay/src/lib/services/ethService/contractService/ContractService.ts
```diff
@@ -255,18 +245,6 @@ export class ContractService implements IContractService {
           return CommonService.redirectBytecodeAddressReplace(address);
         } else if (result.type === constants.TYPE_CONTRACT) {
           if (result.entity.runtime_bytecode !== constants.EMPTY_HEX) {
-            const prohibitedOpcodes = ['CALLCODE', 'DELEGATECALL', 'SELFDESTRUCT', 'SUICIDE'];
-            const opcodes = disassemble(result.entity.runtime_bytecode);
```
simzzz, line 259: The `disassemble` constant and the `@ethersproject/asm` package could be removed altogether since this has been removed

quiet-node:
> @quiet-node regarding this [#3750 (comment)](https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3750#discussion_r2091657940).
> 
> I already tried that way but it doesn't seem better 😞
> 
> ```ts
> export class CacheHelper {
>   private static shouldSkipCachingForSingleParams() {
>     // body
>   }
>   private static shouldSkipCachingForNamedParams() {
>     // body
>   }
>   private static generateCacheKey() {
>     // body
>   }
>   private static extractRequestDetails() {
>     // body
>   }
> }
> 
> export function cache(cacheService: CacheService, options: CacheOptions = {}) {
>   // here, we can not use it like that because it's a private method out of this function scope
>   CacheHelper.shouldSkipCachingForSingleParams();
>   
>   // using it like that should work, but it breaks the ts convention
>   CacheHelper['shouldSkipCachingForSingleParams']
> }
> ```
> 
> The only "workaround" that comes to my mind is something like that 👇
> 
> ```ts
> const shouldSkipCachingForSingleParams = () => {
>   // body
> }
> 
> const shouldSkipCachingForNamedParams = () => {
>   // body
> }
> 
> const generateCacheKey = () => {
>   // body
> }
> const extractRequestDetails = () => {
>   // body
> }
> 
> // this class will be used as a proxy to expose the private functions within this file
> export class CacheHelper {
>   private static shouldSkipCachingForSingleParams() {
>     return shouldSkipCachingForSingleParams();
>   }
>   private static shouldSkipCachingForNamedParams() {
>     return shouldSkipCachingForNamedParams();
>   }
>   private static generateCacheKey() {
>     return generateCacheKey();
>   }
>   private static extractRequestDetails() {
>     return extractRequestDetails();
>   }
> }
> 
> export function cache(cacheService: CacheService, options: CacheOptions = {}) {
>   // here, we will use the global function instead of the class private method
>   // (CacheHelper['shouldSkipCachingForSingleParams']) to avoid calling a private method
>   shouldSkipCachingForSingleParams();
> }
> ```
> 
> And yep, mixing up components with classes is a mess here, also we should choose the approach that suits our codebase instead of introducing a new one IMO.
> 
> Introducing a testable decorator is a must-have, we definitely should create a task for that, but I think is out of scope for that PR.

@natanasow Yeah good point yeah I agree it's just not clean to use it this way. We can have a more official PR for testing private methods. I also have captured one here https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3659. 

For now, let's revert back to __test__ namespace and go with it I think it's the cleanest way for now. Sorry for the inconvenience but it's all for the best! 

Great work as always!

quiet-node (approved):
LGTM! Great work as always!

packages/relay/package.json
```diff
@@ -49,7 +49,6 @@
     "test:eth-get-logs": "nyc ts-mocha --recursive './tests/**/*.spec.ts' './tests/**/**/*.spec.ts' -g '@ethGetLogs' --exit"
   },
   "dependencies": {
-    "@ethersproject/asm": "^5.7.0",
```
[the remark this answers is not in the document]
acuarica: Related to `eth_getCode` implementation, see https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3698#discussion_r2082171338 for context.

simzzz (approved):

PR #3768, opened 2025-05-15
fix: allowed callTracer to return empty array if no call actions found

### LINKED ISSUE #3751: `debug_traceTransaction` fails with some specific tx hash
For some transaction hashes, _e.g._, `0x47797217418d267efc5afbe2a670d1fdfd0a1ad5cc098c014a8ee5386b4243da`, the RPC method `debug_traceTransaction` fails with the following error

```
Cannot destructure property 'call_type' of 'actionsResponse.actions[0]' as it is undefined.
```

#### Steps to Reproduce

```console
$ curl --request POST \
  --url https://testnet.hashio.io/api \
  --header 'content-type: application/json' \
  --data '{ "method": "debug_traceTransaction", "id": 2, "jsonrpc": "2.0", "params": ["0x47797217418d267efc5afbe2a670d1fdfd0a1ad5cc098c014a8ee5386b4243da",{"tracer": "callTracer"}] }'

{"error":{"code":-32603,"message":"[Request ID: f798f652-4ebc-49c5-8ba1-0923ac2ae223] Error invoking RPC: Cannot destructure property 'call_type' of 'actionsResponse.actions[0]' as it is undefined."},"jsonrpc":"2.0","id":2}%
```

Investigate why this happens and fix the issue.

quiet-node:
**Description**:
This PR allows callTracer to return an empty array if no call actions are found, effectively preventing the deconstruction of an undefined object.

**Related issue(s)**:

Fixes #3751

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/types/debug.ts
+++ b/packages/relay/src/lib/types/debug.ts
@@ -73,5 +73,5 @@ export interface TraceBlockByNumberTxResult {
   /**
    * The result of the trace, which can be either a {@link CallTracerResult} or an {@link EntityTraceStateMap}.
    */
-  result: CallTracerResult | EntityTraceStateMap;
+  result: CallTracerResult | EntityTraceStateMap | null;
 }
--- a/packages/relay/src/lib/debug.ts
+++ b/packages/relay/src/lib/debug.ts
@@ -444,6 +444,9 @@ export class DebugImpl implements Debug {
         throw predefined.RESOURCE_NOT_FOUND(`Failed to retrieve contract results for transaction ${transactionHash}`);
       }
 
+      // return empty array if no actions
+      if (actionsResponse.length === 0) return [];
+
       const { call_type: type } = actionsResponse[0];
       const formattedActions = await this.formatActionsResult(actionsResponse, requestDetails);
 
--- a/packages/relay/src/lib/debug.ts
+++ b/packages/relay/src/lib/debug.ts
@@ -429,7 +429,7 @@ export class DebugImpl implements Debug {
     transactionHash: string,
     tracerConfig: ICallTracerConfig,
     requestDetails: RequestDetails,
-  ): Promise<CallTracerResult> {
+  ): Promise<CallTracerResult | null> {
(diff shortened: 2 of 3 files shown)


acuarica (approved):

Ferparishuertas (dismissed):
LGTM

packages/relay/src/lib/types/debug.ts
```diff
@@ -73,5 +73,5 @@ export interface TraceBlockByNumberTxResult {
   /**
    * The result of the trace, which can be either a {@link CallTracerResult} or an {@link EntityTraceStateMap}.
    */
-  result: CallTracerResult | EntityTraceStateMap;
+  result: CallTracerResult | EntityTraceStateMap | null;
```
acuarica, line 76: Do we have a link of which spec we are following? There is no [official spec](https://ethereum.github.io/execution-apis/api-documentation) for these `debug_*` methods. So we probably want to make it clear which spec we are following.

For example, to answer things like "why is `null` a valid return value here?".

acuarica (approved):
Approved, but left a comment.

natanasow (approved):

Ferparishuertas (approved):
LGTM

packages/relay/src/lib/types/debug.ts
acuarica: Yes, totally. As discussed offline, it's not easy to find a proper spec for these `debug_*` endpoints. QuickNode includes them in their free tier, so we can try them out, but not Alchemy for example.

I think for now returning `null` is fine, but we'd need to find a better spec to adhere to. For example, in the case of "transaction not found", Anvil (Foundry's development EVM) returns an "empty" result (instead of `null`)

```
  "result": {
    "failed": false,
    "gas": 0,
    "returnValue": "",
    "structLogs": []
  }
```

It would be interesting to see what `geth` or Hardhat return in this case as well. With this information we might be able to have a pseudo-spec to adhere to.

PR #3777, opened 2025-05-20
fix: fixed flaky/failing acceptance tests

### LINKED ISSUE #3665: Fix web server release test due to mismatch in `Request ID`
The following release test for Web Server consistently fails (you can run it using `acceptancetest:release_ws`)

```console
$ npm run acceptancetest:release_ws
[...]
       Acceptance tests
         @web-socket-batch-request Batch Requests
           given WS_BATCH_REQUESTS_ENABLED=true are set
             @release Should submit batch requests to WS server using Standard Web Socket and retrieve batch responses:

      AssertionError: expected [ { result: '0x129', …(2) }, …(4) ] to deeply equal [ { result: '0x129', …(2) }, …(4) ]
      + expected - actual

         }
         {
           "error": {
             "code": -32602
      -      "message": "[Request ID: d79be947-5fcb-44e8-8ee9-5c04847cb26c] Invalid parameter 0: Expected 0x prefixed string representing the address (20 bytes), value: 0xhedera"
      +      "message": "[Request ID: 7c3b0dc6-f7b7-4219-9f64-672488ec10e8] Invalid parameter 0: Expected 0x prefixed string representing the address (20 bytes), value: 0xhedera"
           }
           "id": 1
           "jsonrpc": "2.0"
         }
      
      at Suite.<anonymous> (packages/ws-server/tests/acceptance/batchRequest.spec.ts:80:38)
      at fulfilled (packages/ws-server/tests/acceptance/batchRequest.spec.ts:6:58)
[...]
```

The problem seems to be just the comparison. Note that the messages only differ in the `Request ID` (which makes sense) given how the test work

https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/69bf5102b3ebbbc841f4d417664a58938b3bc272/packages/ws-server/tests/acceptance/batchRequest.spec.ts#L69-L81

A possible solution would just to mask the `Request ID` when a given response is an error.

> [!NOTE]
> I believe this test is not being run on CI, that's why we didn't detect it earlier.

### LINKED ISSUE #3666: Fix flaky test `"eth_getBalance" with block number in the last 15 minutes for account that has performed contract deploys/calls`
The following server release test fails intermittently causing flakiness in our test suite.

```console
$
[...]
  0 passing (1m)
  1 failing

       Acceptance tests
         @api-batch-2 RPC Server Acceptance Tests
           eth_getBalance
             @release should execute "eth_getBalance" with block number in the last 15 minutes for account that has performed contract deploys/calls:

      AssertionError: expected '0x2b4f651f00ed82400' to equal '0x2a6f0f4064b89c800'
      + expected - actual

      -0x2b4f651f00ed82400
      +0x2a6f0f4064b89c800
      
      at Context.<anonymous> (packages/server/tests/acceptance/rpc_batch2.spec.ts:552:22)
      at fulfilled (packages/server/tests/acceptance/rpc_batch2.spec.ts:6:58)
      at processTicksAndRejections (node:internal/process/task_queues:105:5)
[...]
```

> [!NOTE]
> This happens against both previewnet and testnet

For reference, here's the failing test

https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/69bf5102b3ebbbc841f4d417664a58938b3bc272/packages/server/tests/acceptance/rpc_batch2.spec.ts#L540-L548

The issue seems to be a race condition when `accounts0StartBalance` is set at `blockNumberAtStartOfTests`. PR https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3495 refactors these tests. However, it seems it also changed its semantics. See

https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3495/files#diff-e91ffd80132cb7a1d8cfed9a53ebc797b745e3113c4cd7004b02a0679566d2d4R138-R148

in particular note how before said PR the line `pollForValidTransactionReceipt(createChildTx.hash)` was before setting the value for `accounts0StartBalance`. Moreover, the line `[childTx, blockNumber, balance] = await Promise.all[...]` can be causing additional issues because if `childTx` was included in `blockNumber` then the reported `balance` would be incorrect.

Another thing to take into account could be the cache on `eth_blockNumber`. If there is another call to `eth_blockNumber` (on a public network this can easily happen) while the test is being setup, then `blockNumberAtStartOfTests` could refer to a cached block number. This could also be the source of the issue described above.

https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/69bf5102b3ebbbc841f4d417664a58938b3bc272/packages/relay/src/lib/services/ethService/ethCommonService/index.ts#L283-L299

quiet-node:
**Description**:
 - fixed batch request WS release test due to mismatch request ID
 - fixed flaky `"eth_getBalance" with block number in the last 15 minutes for account that has performed contract deploys/calls` test in batch2 due to race condition during initialization

**Related issue(s)**:

Fixes #3665 
Fixes #3666 

**Notes for reviewer**:

**Checklist**


simzzz (dismissed):

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -118,15 +118,17 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
     });
     await relay.pollForValidTransactionReceipt(initialFundsTx.hash);
 
-    const [childTx, blockNumber, balance] = await Promise.all([
-      parentContract.createChild(1),
+    createChildTx = await parentContract.createChild(1);
+    await relay.pollForValidTransactionReceipt(createChildTx.hash);
+
+    // added delay to wait for balances properly settled on MN
+    await Utils.wait(1500);
```
acuarica, line 125: Isn't this a double wait? If `createChildTx` is finished, doesn't it mean that balances are already settled on the MN?

quiet-node: Hey Luis, thanks for the question but without the extra wait there still occasional balance mismatches on production networks. I believe that once a transaction reaches consensus, the balances are updated in the Merkle tree almost instantly however for the Mirror Node to reflect those updates, there’s still a slight delay. Without the wait it should work fine, but given the busy traffic on production networks, it seems the Mirror Node sometimes takes longer to catch up, leading the flakyness we're dealing with. That's why I added the extra wait to stablize the results. Not ideal but flexible for different production environemnt.

acuarica: Thanks for the clarification, really helpful. Maybe we should expand the comment to include a bit of this explanation?

Also, we don't remove the flakiness entirely, this is just a mitigation. What if we poll the MN to ensure the mentioned transaction is visible from it? This way, we are 100% sure to have solved the problem. I remember seeing this mechanism in tests in the Smart Contract repo.

acuarica (approved):
In any case, going to approve it.

packages/server/tests/acceptance/rpc_batch2.spec.ts
quiet-node: Yeah, you’re right—a comment would definitely help.

As for the extra wait doesn’t eliminate the flakiness entirely, You’re correct that it’s more of a mitigation to improve stability. If the production environment needs more than 1500 ms, the issue can still occur, though that’s rare—but still possible.

Regarding polling the Mirror Node to confirm the transaction is visible, that’s actually what’s happening under the hood with `await relay.pollForValidTransactionReceipt(initialFundsTx.hash)`, which essentially calls `eth_getTransactionReceipt` to fetch the receipt from the Mirror Node.

Maybe we should consider polling the Mirror Node for updated balances instead? Kinda feel like I have done this before but it was quite cumbersome somehow but let me see what I can do.

You're right we should aim for a more robust fix rather than just a mitigation. Will push in an update for this.

quiet-node: Synced offline with @acuarica and discovered that the flakiness was caused by a cached value of the latest block number. The new solution implements a polling mechanism to ensure the block is correctly updated after contract execution, resulting in stable and consistent outcomes.

acuarica (commented):
nice stuff!

packages/server/tests/acceptance/rpc_batch2.spec.ts
acuarica: Awesome!

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -118,17 +118,27 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
...
-    await relay.pollForValidTransactionReceipt(createChildTx.hash);
+    // Note: There is currently a caching solution for eth_blockNumber that stores the block number.
+    // This loop is designed to poll for the latest block number until it is correctly updated.
+    for (let i = 0; i < 5; i++) {
+      if (blockNumAfterCreateChildTx > blockNumBeforeCreateChildTx) {
+        break;
+      }
+      await Utils.wait(1500);
+      blockNumAfterCreateChildTx = await await relay.call(RelayCalls.ETH_ENDPOINTS.ETH_BLOCK_NUMBER, [], requestId);
+    }
```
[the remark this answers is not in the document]
quiet-node: ha funny - `await await`! Good catch thanks for pointing it out! Implemented your suggestion!

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -72,7 +72,7 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
   const FEE_SCHEDULE_FILE_CONTENT_UPDATED =
     '0a280a0a08541a061a0440a8953a0a0a08061a061a0440889d2d0a0a08071a061a0440b0b63c120208011200'; // Eth gas = 953000
 
-  let blockNumberAtStartOfTests = 0;
+  let blockNumAfterCreateChildTx = 0;
```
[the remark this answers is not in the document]
quiet-node: This is actually the closest I could get lol it’s a global variable, and the polling happens within the beforeAll block, so I couldn’t really place it near the polling.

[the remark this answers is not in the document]
acuarica: Ah, I see it now.

acuarica (approved):
Cool stuff!

natanasow (approved):

PR #3789, opened 2025-05-23
fix: Aligns debug_traceTransaction validation with industry standards

### LINKED ISSUE #3739: [NEED INVESTIGATION] Incorrect param validation for debug_traceTransaction
## Description
The current implementation of `debug_traceTransaction` has incorrect parameter validation and requires custom parameter arrangement in the `arrangeRpcParams()` method. This creates unnecessary complexity and deviates from industry standards.

## Current implementation issues
- Second parameter validation accepts either a string or object, rather than only an object
- Special-case handling in `arrangeRpcParams()` method for `traceTransaction`
- Validation logic expects three parameters when the method should have two
- Needs Investigation – The onlyTopCall option in tracerConfig doesn’t appear to work correctly with callTracer. This is a preliminary observation and requires further investigation to confirm and understand the issue.

## Expected behavior
- Parameter validation should align with industry standards (Alchemy, Quicknode, Infura)
- Second parameter should be an object with standardized fields:
  - `tracer`
  - `tracerConfig` (containing `onlyTopCall`)
- Fix would eliminate need for custom parameter arrangement in `arrangeRpcParams()`
- Standardized approach would allow method parameters to be normalized without special cases

## Params Discrepancy
- For the second parameter, other providers like Infura, QuickNode, Alchemy, etc., appear to support two options for tracer: callTracer and prestateTracer. However, the current implementation of debug_traceTransaction only supports callTracer and opcodeLogger, with no support for prestateTracer. Further investigation is needed to determine the reason for this discrepancy.

## Benefits
- Improved code maintainability
- Consistent interface with other RPC providers
- Simplified parameter handling logic

## References

- https://docs.alchemy.com/reference/debug-tracetransaction
- https://docs.metamask.io/services/reference/mantle/json-rpc-methods/debug/debug_tracetransaction/
- https://www.quicknode.com/docs/ethereum/debug_traceTransaction

konstantinabl:
**Description**:

Changes validation for debug_traceTransaction to align with industry standards (Alchemy, QuickNode)

**Related issue(s)**:

Fixes #3739


acuarica:
Hey @konstantinabl, left a comment in the respective issue https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3739#issuecomment-2904518934.

simzzz (approved):

simzzz (dismissed):
LGTM

packages/relay/src/lib/validators/objectTypes.ts
```diff
@@ -305,4 +306,25 @@ export class TracerConfigWrapper extends DefaultValidation<ITracerConfigWrapper>
...
   }
+
+  validate() {
+    const valid = super.validate();
+
+    const { tracer, tracerConfig } = this.object;
+
+    // we want to accept ICallTracerConfig only if the tracer is callTracer
+    // this config is not valid for opcodeLogger
+    if (
```
quiet-node, line 317: I’m not aware of this. Last time I worked on eth_traceBlockByNumber, onlyTopCall worked with prestate tracer as well. Could you share the documentation that states otherwise?

packages/relay/src/lib/debug.ts
```diff
@@ -106,25 +107,35 @@ export class DebugImpl implements Debug {
...
+    if (tracerObject?.tracer === TracerType.PrestateTracer) {
+      throw predefined.INTERNAL_ERROR('Prestate tracer is not yet supported on debug_traceTransaction');
+    }
+
+    //we use a wrapper since we accept a transaction where a second param with tracer/tracerConfig may not be provided
+    //and we will still default to opcodeLogger
+    const wrapper = tracerObject ?? {};
+    const tracer = wrapper.tracer ?? TracerType.OpcodeLogger;
+    const tracerConfig = wrapper.tracerConfig ?? {};
+
```
quiet-node, line 132: Below should work, right?

```suggestion
const tracer = tracerObject?.tracer ?? TracerType.OpcodeLogger;
const tracerConfig = tracerObject?.tracerConfig ?? {};
```

packages/relay/src/lib/validators/objectTypes.ts
konstantinabl: I think my comment says that it shouldnt work with opcode logger, thats what i meant

quiet-node: I’m also curious why this issue is only coming to light now and hasn’t been reported earlier. Could you double-check whether the original validator already covered these scenarios, or did we really miss them?

If they were overlooked, it might make sense to take a more robust and flexible approach—one that validates config options based on tracer types more generically. That would include not just callTracer and opcodeLogger, but also prestate and any future tracer types, rather than just handling a single case like this method currently does.

If that’s the direction we want to take, I imagine it would require a bit more effort and could go beyond the scope of this PR, which mainly focuses on aligning the parameters. In that case, it might be better to address it in a follow-up PR.

Let me know what you think—I’m okay either way, but I’m just a bit hesitant about leaving a validation method that only handles one specific scenario.

packages/relay/src/lib/debug.ts
konstantinabl: I would assume it works yes, it is essentially the same without the extra variable

packages/relay/src/lib/validators/objectTypes.ts
konstantinabl: > I’m also curious why this issue is only coming to light now and hasn’t been reported earlier. Could you double-check whether the original validator already covered these scenarios, or did we really miss them?
> 
> If they were overlooked, it might make sense to take a more robust and flexible approach—one that validates config options based on tracer types more generically. That would include not just callTracer and opcodeLogger, but also prestate and any future tracer types, rather than just handling a single case like this method currently does.
> 
> If that’s the direction we want to take, I imagine it would require a bit more effort and could go beyond the scope of this PR, which mainly focuses on aligning the parameters. In that case, it might be better to address it in a follow-up PR.
> 
> Let me know what you think—I’m okay either way, but I’m just a bit hesitant about leaving a validation method that only handles one specific scenario.

Good question! I double-checked, and it looks like the original validator didn’t explicitly handle this scenario—it was probably just overlooked at the time. As stated in the issue you opened other things were not quite aligned with other providers (or back then format of parameters accepted was different) That said, I agree that a more generic validation approach could be useful in the future. I improved the current validate method further to take into consideration the reversed case where opcodeLogger config may be used with callTracer tracer

packages/relay/src/lib/validators/objectTypes.ts
```diff
@@ -305,4 +306,44 @@ export class TracerConfigWrapper extends DefaultValidation<ITracerConfigWrapper>
...
+    }
+
+    const hasOpcodeLoggerKeys = keys.some((k) => opcodeLoggerKeys.includes(k));
+    if (hasOpcodeLoggerKeys && tracer !== TracerType.OpcodeLogger) {
+      throw predefined.INVALID_PARAMETER(
+        1,
+        `opcodeLogger 'tracerConfig' for ${this.name()} is only valid when tracer=${TracerType.OpcodeLogger}`,
+      );
+    }
+    return valid;
```
quiet-node, line 347: Sorry, still not convinced with the new approach due to the introduction of `callTracerKeys` and `opcodeLoggerKeys`. There are already multiple key definitions for these, such as `OBJECTS_VALIDATIONS.opcodeLoggerConfig` and `OBJECTS_VALIDATIONS.callTracerConfig`. How about we leverage what already exists instead of introducing new definitions? This would provide a single source of truth rather than duplicated definitions.

New definitions means additional maintenance overhead—future modifications would require updates in multiple places, then it leaves the chance of missing one of the spots and lead to the risk of inconsistencies and unsynced variable values. I believe leveraging the existing validation schema would be a more robust approach.

How about something like this:

```suggestion
try {
  if (tracer === TracerType.CallTracer) {
    new CallTracerConfig(tracerConfig).validate(); // use existed CallTracerConfig class
  } else if (tracer === TracerType.OpcodeLogger) {
    new OpcodeLoggerConfig(tracerConfig).validate(); // use existed OpcodeLoggerConfig class
  }
} catch (error: any) {
  // handle the error down here
  throw predefined.INVALID_PARAMETER(...)
}
```

Since `.validate()` from the base class calls `Validator.validateObject()`, which currently does not check for invalid keys in the config, we can enhance it by introducing a generic key validation method:

```typescript
export function isValidPropertyName(object: any, filters: IObjectSchema): boolean {
  const properties = Object.keys(filters.properties);
  const objectKeys = Object.keys(object);
  return objectKeys.every((key) => properties.includes(key));
}
```

We can then invoke this method within `Validator.validateObject()` to validate keys. Eventually, we just need to do proper error handling, then this will lead to a more cohesive design and better reuse of existing definitions.

Note: The code above is a quick scaffold for suggestion purposes—if you decide you want to use them, please ensure it’s properly validated.

packages/relay/src/lib/validators/objectTypes.ts
quiet-node: Discussion moved to https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3789#discussion_r2135924332

packages/relay/src/lib/validators/objectTypes.ts
konstantinabl: My concern is that when we call super.validate() we already call the test function on the _opcodeLoggerConfig_ and _callTracerConfig_ , where we execute` new CallTracerConfig(param).validate();` and `new OpcodeLoggerConfig(param).validate();` , so doing it again doesn't seem optimal to me, because it means we initialize the _OpcodeLoggerConfig_ and _CallTracerConfig_ classes twice, meaning more memory and CPU is used. 
Maybe we can stick to my approach, but instead of initializing the key arrays we use OBJECT_VALIDATIONS
```
const callTracerKeys = Object.keys(OBJECTS_VALIDATIONS.callTracerConfig.properties);
    const opcodeLoggerKeys = Object.keys(OBJECTS_VALIDATIONS.opcodeLoggerConfig.properties);
    
    const configKeys = Object.keys(tracerConfig);
    const hasCallTracerKeys = configKeys.some((k) => callTracerKeys.includes(k));
    const hasOpcodeLoggerKeys = configKeys.some((k) => opcodeLoggerKeys.includes(k));
```
Let me know what you think

packages/relay/src/lib/types/debug.ts
```diff
@@ -13,6 +13,13 @@ export interface BlockTracerConfig {
   tracerConfig?: ICallTracerConfig;
 }
 
+export interface TransactionTracerConfig {
```
acuarica, line 16: this looks quiet similar to `interface BlockTracerConfig`, can we/should we reuse the same type?

packages/relay/src/lib/types/debug.ts
```diff
@@ -13,6 +13,13 @@ export interface BlockTracerConfig {
...
+  /** The type of tracer to use for block tracing. */
+  tracer: TracerType;
+  /** Optional configuration for the call tracer. */
+  tracerConfig?: ICallTracerConfig | IOpcodeLoggerConfig;
```
acuarica, line 20: this looks like `type ITracerConfig`, I don't mind inlining (actually might make more sense here). But if `ITracerConfig` is not used or duplicated we should unify the types used to avoid confusion.

packages/relay/src/lib/types/debug.ts
konstantinabl: Its similar, however the tracerConfig in BlockTracerConfig is of another type

packages/relay/src/lib/debug.ts
```diff
@@ -106,26 +106,34 @@ export class DebugImpl implements Debug {
...
   @cache(CacheService.getInstance(CACHE_LEVEL.L1))
   async traceTransaction(
     transactionIdOrHash: string,
-    tracer: TracerType,
-    tracerConfig: ITracerConfig,
+    tracerObject: TransactionTracerConfig,
     requestDetails: RequestDetails,
   ): Promise<any> {
+    if (tracerObject?.tracer === TracerType.PrestateTracer) {
+      throw predefined.INTERNAL_ERROR('Prestate tracer is not yet supported on debug_traceTransaction');
```
quiet-node, line 119: It shouldn’t throw an internal error, as that would return a 500 status code, which is misleading and contributes to the failed request alert. I believe it makes more sense to use INVALID_PARAMETER which is mapped to 400 instead.

quiet-node (changes requested):
LG one small item left

acuarica (approved):

PR #3792, opened 2025-05-26
chore: add `eth_getBlockReceipts` method to OpenRPC definitions

### LINKED ISSUE #3793: Add missing eth_getBlockReceipts to openrpc.json
**Description:**
The `eth_getBlockReceipts` is already implemented in the application but is currently missing from the `openrpc.json` specification.

Related to #1837

mwb-al:
**Description**:

Added `eth_getBlockReceipts` to `openrpc.json` to reflect the method already implemented in the application.

**Related issue(s)**:

Fixes #3793

**Checklist**


mwb-al:
`blobGasUsed` and `blobGasPrice` fields have been removed. These fields are not used (ITransactionReceipt)

docs/openrpc.json
```diff
@@ -1518,7 +1552,7 @@
         ]
       },
       "BlockNumberOrTagOrHash": {
-        "title": "Block number or tag",
+        "title": "Block number, tag, or block hash",
```
arianejasuwienas, line 1555: I believe we can only use a tag or a number, so this change proposal needs to be reverted.

docs/openrpc.json
```diff
@@ -1932,7 +1971,16 @@
           "to": {
             "title": "to",
             "description": "Address of the receiver or null in a contract creation transaction.",
-            "$ref": "#/components/schemas/address"
+            "oneOf": [
+              {
+                "title": "Contract Creation (null)",
```
arianejasuwienas, line 1976: In the receipt information, this field is populated for contract creation receipt when I test it against the testnet. So the current documentation appears to be correct in this regard.

docs/openrpc.json
arianejasuwienas: We are only changing the title here so this proposal is ok, my bad!

docs/openrpc.json
mwb-al: According to the implementation of `getBlockReceipts` (see https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/main/packages/relay/src/lib/services/ethService/blockService/BlockService.ts#L181 ), the `"to"` field cannot be null and is always a string.

The method

```ts
resolveEvmAddress(address: string, requestDetails: RequestDetails, types?: string[]): Promise<string>;
```

assumes this behavior.

We need to decide whether this is intentional or if the method should be updated to match the standard specification: [https://github.com/ethereum/execution-apis](https://github.com/ethereum/execution-apis)

I suggest:

1. Update `openrpc.json` to match the current implementation.
2. Create a new GitHub issue describing the differences, so we can decide whether this behavior is intentional or needs to be fixed.

Ferparishuertas (dismissed):
LGTM

acuarica (approved):

acuarica:
Is this PR still draft? Otherwise, we should mark it as "Ready for review" for better visibility.

mwb-al:
I added a case here where `'to'` is `null` – which is compliant with the Ethereum spec.

As soon as this is merged:
[https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3825](https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3825)
I'll switch this PR from Draft to Ready for Review.

simzzz (approved):

PR #3810, opened 2025-06-03
feat: add OpenRPC JSON updater tool (#1837)

### LINKED ISSUE #1837: Make the `openrpc.json` schema adhere to the Ethereum execution API specs.
### Problem

The [`openrpc.json` schema file](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/docs/openrpc.json) provided in the repo is not aligned with the one generated using the [official Ethereum execution API specs repo](https://github.com/ethereum/execution-apis.git).

We noticed the differences while working [on some updates about the `v` parameter](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1834#discussion_r1361361955).

See [here for an example of a compiled `openrpc.json` schema file](https://github.com/InternetOfPeers/execution-apis/blob/main/openrpc.json).

### Solution

- Adhere to the OpenRPC schema definition generated by the official Ethereum execution API specs repo for all those functions we already implemented in the JSON-RPC Relay

### Alternatives

_No response_

mwb-al:
**Description**:

Tool to keep the spec in sync in case of updates.

**Related issue(s)**:

Fixes #1837 

**Checklist**


.github/workflows/openrpc-updater.yml
```diff
@@ -0,0 +1,135 @@
...
+        run: |
+          cd scripts/openrpc-json-updater
+          REPORT_OUTPUT=$(node cli.js)
+          echo "REPORT_OUTPUT<<EOF" >> $GITHUB_ENV
+          echo "$REPORT_OUTPUT" >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
+
+      - name: Perform merge
+        id: merge
+        run: |
```
arianejasuwienas, line 82: This job seems to work ok, but:
- In the issue description we did not mention anything about creating new PRs. Perhaps we should explain why are we creating them and when we do it? 
- Perhaps this job should be optional and executed manually?

.github/workflows/openrpc-updater.yml
```diff
@@ -0,0 +1,135 @@
...
+          TIMESTAMP=$(date +%Y%m%d%H%M%S)
+          UNIQUE_BRANCH="update-openrpc-${TIMESTAMP}"
+          echo "UNIQUE_BRANCH=${UNIQUE_BRANCH}" >> $GITHUB_ENV
+          echo "Generated unique branch name: ${UNIQUE_BRANCH}"
+
+      - name: Create Pull Request
+        if: env.SKIP_PR != 'true'
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+        with:
+          token: ${{ secrets.PAT_TOKEN }}
```
arianejasuwienas, line 117: Maybe rename it to PAT or PERSONAL_ACCESS_TOKEN?

PA**T_TOKEN** = PERSONAL ACCESS **TOKEN TOKEN**

scripts/openrpc-json-updater/original-openrpc.json
```diff
@@ -0,0 +1 @@
+{}
```
arianejasuwienas, line 1: Do we even need a separate file for an empty object?

scripts/openrpc-json-updater/config.js
```diff
@@ -0,0 +1,132 @@
+// SPDX-License-Identifier: Apache-2.0
+
+export const SKIPPED_KEYS = [
+  "examples",
+  "baseFeePerBlobGas",
+  "blobGasUsedRatio",
```
arianejasuwienas, line 6: nit, inconsistent intendations (2 spaces here 4 in 10th line)

scripts/openrpc-json-updater/config.js
```diff
@@ -0,0 +1,132 @@
...
+
+    "eth_maxPriorityFeePerGas.summary",
+    "eth_maxPriorityFeePerGas.result.schema.description",
+    "eth_maxPriorityFeePerGas.result.schema.title",
+
+    "eth_newBlockFilter.summary",
+    "eth_newBlockFilter.result.name",
+];
+
+export const DISCARDED_METHODS = [
```
arianejasuwienas, line 52: Maybe we should include a short explanation (in the comments, or in readme what methods /fields are we skipping and why exactly).

It actually seems important for developers to know about these exclusions.

scripts/openrpc-json-updater/operations/merge.js
```diff
@@ -0,0 +1,462 @@
...
+      // If method doesn't exist in modified, add it
+      if (!modifiedMap.has(name)) {
+        filteredModified.methods.push(origMethod);
+        continue;
+      }
+      const modMethod = modifiedMap.get(name);
+
+      // Process $ref fields
+      this.processRefFields(origMethod, modMethod);
+      // Process differing keys
```
arianejasuwienas, line 75: nit Some of the comments are truly redundant - this one is a good example. We should consider removing comments that:
 -  don't explain anything,
 - don't add any context to the code

scripts/openrpc-json-updater/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "type": "module",
```
arianejasuwienas, line 2: perhaps we should add author, name, description, **license** fields here ?

scripts/openrpc-json-updater/package.json
```diff
@@ -0,0 +1,6 @@
...
+  "dependencies": {
```
arianejasuwienas, line 3:
```suggestion
  "devDependencies": {
```

scripts/openrpc-json-updater/utils/file.utils.js
```diff
@@ -0,0 +1,216 @@
...
+  try {
+    if (!originalContent) {
+      fs.writeFileSync(filePath, JSON.stringify(data, null, 2), 'utf-8');
+      return true;
+    }
+
+    const eol = originalContent.includes('\r\n') ? '\r\n' : '\n';
+    const formatted = formatJson(data);
+    const output = formatted.replace(/\n/g, eol);
+    fs.writeFileSync(filePath, output, 'utf-8');
```
arianejasuwienas, line 211: Perhaps it's worth noting that the script needs to be privileged to write ijn certain directories / files. And add output results to the gitignore file(?)

scripts/openrpc-json-updater/original-openrpc.json
konstantinabl: @mwb-al is this comment actually addressed, i see you have marked it as resolved?

.github/workflows/openrpc-updater.yml
```diff
@@ -0,0 +1,138 @@
...
+          cp ./downloaded-artifacts/refs-openrpc.json scripts/openrpc-json-updater/original-openrpc.json
+
+      - name: Setup Node.js
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+        with:
+          node-version: '22'
+
+      - name: Install dependencies
+        run: |
+          cd scripts/openrpc-json-updater
```
acuarica, line 70: we can use `working-directory` here as well

scripts/openrpc-json-updater/cli.js
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { mergeDocuments } from './operations/merge.js';
+import { compareIgnoringFormatting, prepareDocuments } from './operations/prepare.js';
+import { generateReport } from './operations/report.js';
+import { readJson, writeJson } from './utils/file.utils.js';
+
+const originalFilePath = './original-openrpc.json';
+const modifiedFilePath = '../../docs/openrpc.json';
```
acuarica, line 9: Instead of hardcoding these values, we can pass them as CLI args. This way it would be easier to try this locally without copying files around.

If this falls out of scope, feel free to discard.

.github/workflows/openrpc-updater.yml
```diff
@@ -0,0 +1,138 @@
...
+
+      - name: Build project
+        run: npm run build
+        working-directory: ./execution-apis
+
+      - name: Upload openrpc.json as an artifact
+        uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
+        with:
+          name: openrpc
+          path: ./execution-apis/refs-openrpc.json
```
acuarica, line 40: We might include a brief description somewhere on the difference between `refs-openrpc.json` and `openrpc.json`. The execution-apis repo doesn't clarify what `refs-openrpc.json` is. Perhaps in the `README.md`?

.github/workflows/openrpc-updater.yml
```diff
@@ -0,0 +1,138 @@
...
+        if: env.SKIP_PR != 'true'
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7.0.8
+        with:
+          token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+          commit-message: Update OpenRPC JSON
+          title: 'Update OpenRPC JSON'
+          body: |
+            # OpenRPC JSON Update
+            
+            This PR updates the OpenRPC JSON file with the latest changes.
```
acuarica, line 129: maybe we can expand a little the description, something like

```suggestion
            This PR updates the OpenRPC JSON specification with the latest changes from Ethereum JSON-RPC specification.
```

acuarica (commented):
looking good, great stuff to have this

scripts/openrpc-json-updater/cli.js
mwb-al: fixed: b2478fe2

scripts/openrpc-json-updater/cli.js
```diff
@@ -0,0 +1,72 @@
...
+    const merged = mergeDocuments(normalizedOriginal, normalizedModified);
+
+    if (!hasDifferences(normalizedModified, merged)) {
+      console.log(`\nNo differences found after merge. No changes needed.\n`);
+      process.exit(0);
+    }
+
+    writeJson(modifiedFilePath, merged, modifiedContent);
+    console.log(`\nMerge completed. Updated file: '${modifiedFilePath}'.\n`);
+    return;
```
acuarica, line 65: Hi @mwb-al, quick question, are we missing a `process.exit(1)` here?

I made some changes to our `openrpc.json` spec and wanted to see the differences with the execution-apis. I ran it locally and I got the message `Merge completed. Updated file: '../../docs/openrpc.json'.`, but the exit code was `0`. This won't trigger the PR creation in CI, right?

mwb-al: Actual behaviour and exit codes:
1. Workflow succeeds (green): When there are no differences detected in openrpc.json, the updater exits with code 0. The workflow completes successfully and does not create a PR.
2. Workflow succeeds (green): When differences are detected and changes are merged, the updater exits with `return`. The workflow can proceed to the step that creates a PR with the updated spec.
3. Workflow fails (red): If any error occurs during processing, the updater exits with a non-zero code (process.exit(1)). In this case, the workflow fails and PR creation is skipped.

In this code, `return` is the last statement, there are no more asynchronous operations, and nothing keeps the Node.js event loop running. This means Node.js will exit naturally with code 0, just like with `process.exit(0)`.

Link to official documentation:
https://nodejs.org/api/process.html#process_event_exit

For clarity and certainty, I can create a GitHub issue and PR to address this - should I do it?

PR #3825, opened 2025-06-05
feat: fix contract creation tx receipt compatibility with Ethereum spec

### LINKED ISSUE #3814: `to` field is not `null` for contract creation transactions in `eth_getBlockReceipts`
### Problem

In Ethereum, when a transaction creates a smart contract, the `to` field in the transaction receipt must be `null`. This is reflected in both the Ethereum JSON-RPC spec and the [OpenRPC schema](https://ethereum.github.io/execution-apis/docs/reference/eth_getblockreceipts), where the `to` field is marked as nullable.

In Hedera’s current implementation, the `to` field is never `null`, even for contract creation transactions. Instead, it is populated with the same value as `contractAddress`, which is not compatible with Ethereum behaviour and tooling.

Hedera's implementation would fail this Execution-APIs test:
**Test:** `eth_getBlockReceipts` - "to" field should be `null` for contract creation
**Test location:** https://github.com/ethereum/execution-apis/blob/main/tests/eth_getBlockReceipts/get-block-receipts-latest.io

#### Sample commands

1. Sepolia (correct behavior), with alchemy:
    ```bash
    curl -X POST https://eth-sepolia.g.alchemy.com/v2/$ALCHEMY_API_KEY \
         -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["0x54d857659b4f0de22e13b09543c90a021a829c8316a25dd6e3fc82d6fa03358e"]}'
    ```

    Response:
    ```json
    {
      "jsonrpc": "2.0",
      "id": 1,
      "result": [
        {
          (...)
          "from": "0x41653c7d61609d856f29355e404f310ec4142cfb",
          "to": null, // NULL !!!!!!!!
          "contractAddress": "0x1f9840a85d5af5bf1d1762f925bdaddc4201f984"
        },
        (....)
      ]
    }
    ```
2. Sepolia (correct behaviour), with infura:
    ```bash
    curl -X POST https://sepolia.infura.io/v3/$INFURA_API_KEY \
         -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["0x54d857659b4f0de22e13b09543c90a021a829c8316a25dd6e3fc82d6fa03358e"]}'
    ```
    Respone: The "to" field of the transaction creating the contract (0x1f9840a85d5af5bf1d1762f925bdaddc4201f984) is `null`.

3. Sepolia (correct behaviour), with quicknode:
    ```bash
    curl -X POST https://newest-morning-fire.ethereum-sepolia.quiknode.pro/$QUICKNODE_API_KEY/ \
         -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["0x54d857659b4f0de22e13b09543c90a021a829c8316a25dd6e3fc82d6fa03358e"]}'
    ```
    Respone: The "to" field of the transaction creating the contract (0x1f9840a85d5af5bf1d1762f925bdaddc4201f984) is `null`.
4. Hedera (incorrect behaviour):
    ```bash
         -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["0x134C79E"]}'
    ```

    Returns:
    ```json
    {
      "result": [
        {
          "blockHash": "0x06d0392689b6e587b0557118988810fbb6e20d2e4bf3c19d7cfb9b6ca730fa67",
          (...)
          "to": "0x5fe2753149ba20351ee08f95cbfa2abe10f2bfff", // The same value as in the contractAddress field!
          "contractAddress": "0x5fe2753149ba20351ee08f95cbfa2abe10f2bfff",
          (...)
          "transactionHash": "0x8d836a03a90ad7595abfa265e2a769f850389684e286ac47e883addba773afd8",
          (...)
        }
      ],
      "jsonrpc": "2.0",
      "id": 1
    }
    ```

### Solution

Update the implementation so that the to field is set to null for contract creation transactions, as defined in the [OpenRPC schema](https://ethereum.github.io/execution-apis/docs/reference/eth_getblockreceipts) and observed on Ethereum clients.

Fix can be applied here: https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/main/packages/relay/src/lib/services/ethService/blockService/BlockService.ts#L181 

### Alternatives

An alternative would be to document this difference and continue returning the `contractAddress` in the `to` field. However, this breaks interoperability with Ethereum tools and test suites. `This is not the recommended approach`.

mwb-al:
**Description**:
There is a discrepancy between Hedera and standard Ethereum behavior regarding the `to` field in transaction receipts for contract creation transactions:

- **Standard Ethereum JSON-RPC**: When a contract is created via a transaction with a null `to` field, the transaction receipt also reports a null `to` field
- **Hedera Mirror Node**: Automatically populates the `to` field with the address of the newly created contract

This inconsistency affects both `eth_getTransactionReceipt` and `eth_getBlockReceipts` endpoints, causing compatibility issues with Ethereum tooling that expects standard transaction receipt formats.

#### Solution
This PR implements a centralized fix in `TransactionReceiptFactory.createRegularReceipt()` to correct the `to` field behavior for contract creation transactions. The solution:

1. **Checks for direct contract creation**: Examines if a contract was directly created by the transaction by verifying if the contract's ID appears in the `created_contract_ids` array
2. **Sets `to` field to null**: When direct contract creation is detected, the `to` field is set to null to match standard Ethereum JSON-RPC behavior
3. **Ensures consistent behavior**: Since both affected endpoints (`eth_getTransactionReceipt` and `eth_getBlockReceipts`) use the same factory method, this change ensures consistent behavior across both endpoints

This change affects:
- `eth_getTransactionReceipt` - Individual transaction receipt queries
- `eth_getBlockReceipts` - Batch transaction receipt queries for entire blocks

**Related issue(s)**:

Fixes #3814 

**Checklist**


packages/relay/src/lib/services/ethService/blockService/BlockService.ts
```diff
@@ -175,11 +175,41 @@ export class BlockService implements IBlockService {
...
+      if (contractResult.created_contract_ids.includes(contractResult.contract_id)) {
+        contractResult.to = null;
+      }
+
       contractResult.logs = logsByHash.get(contractResult.hash) || [];
       const [from, to] = await Promise.all([
         this.common.resolveEvmAddress(contractResult.from, requestDetails),
-        this.common.resolveEvmAddress(contractResult.to, requestDetails),
+        contractResult.to === null ? null : this.common.resolveEvmAddress(contractResult.to, requestDetails),
       ]);
```
quiet-node, line 211: resolveEvmAddress can handle null checks, correct? How about ensuring that resolveEvmAddress performs proper null checking so we don’t have to duplicate it at the call sites?

packages/relay/tests/lib/services/ethService/blockService/BlockService.spec.ts
```diff
@@ -0,0 +1,236 @@
+// SPDX-License-Identifier: Apache-2.0
```
quiet-node, line 1: How about we incoporate this test into the test/lib/eth/eth_getBlockReceipts.spec.ts file? Also I believe this new update affects eth_getTransactionReceipts as well, right? Let's also add test to the corressponded test file for getTransactionReceipts.

quiet-node: Also, let’s add some acceptance tests to validate the end-to-end flows for those two endpoints. In fact, let’s review all affected endpoints and ensure we have test coverage for each of them.

acuarica (commented):
Does this change only impact `eth_getBlockReceipts`, or `eth_getTransactionReceipt` as well? Because there is test change for  `eth_getTransactionReceipt`, but I would have assumed that `TransactionService.ts::handleRegularTransactionReceipt` needs to be changed as well.

In any case, we should update both this PR and issue descriptions to match the actual scope of this PR.

quiet-node (approved):
LGTM

mwb-al:
@quiet-node I have updated the branch with the latest main and made sure all tests pass.

quiet-node:
> @quiet-node I have updated the branch with the latest main and made sure all tests pass.

@mwb-al hey thanks looks like the DCO is still failing. Please update the branch and will approve again

acuarica (approved):

Ferparishuertas (approved):
LGTM

PR #3833, opened 2025-06-10
feat: update `getBlockReceipts` to return `null` when block is not found

### LINKED ISSUE #3813: `eth_getBlockReceipts` returns an error instead of `null` for unknown block hashes
### Problem

The `eth_getBlockReceipts` method in Hedera’s JSON-RPC API throws an error when a valid, but nonexistent block hash is queried. This behavior is inconsistent with Ethereum-compatible JSON-RPC endpoints like Sepolia, which return `null`.

#### Why does it matter
1. According to the [OpenRPC specification for Ethereum JSON-RPC](https://ethereum.github.io/execution-apis/docs/reference/eth_getblockreceipts), the result field for `eth_getBlockReceipts` is defined as `nullable`. The current implementation of the Hedera JSON-RPC relay never returns `null` for this field - even when it should.

2. This behavior causes Hedera to fail conformance tests such as:
   Test: `eth_getBlockReceipts` - Future or nonexistent block
   Test location: https://github.com/ethereum/execution-apis/tree/main/tests/eth_getBlockReceipts

#### Examples:
1. Sepolia (expected behavior), with alchemy:
    ```bash
    curl -X POST https://eth-sepolia.g.alchemy.com/v2/$ALCHEMY_API_KEY \
     -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["0x1000000000000000000000000000000000000000000000000000000000000001"]}'
    ```
    
    Response:
    ```json
    {"jsonrpc":"2.0","id":1,"result":null}
    ```
2. Sepolia (expected behavior), with infura:

    ```bash
    curl -X POST https://sepolia.infura.io/v3/$INFURA_KEY -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["0x54d857659b4f0de22e13b09543c90a021a829c8316a25dd6e3fc82d6fa03358e"]}'
    ```

    Response:
    ```json
    {"jsonrpc":"2.0","id":1,"result":null}
    ```

3. Sepolia (expected behavior), with quicknode:

    ```bash
    curl -X POST https://newest-morning-fire.ethereum-sepolia.quiknode.pro/KEY/ \
    -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["0x2331342342342342"]}'
    ```
    Response:
    ```json
    {"jsonrpc":"2.0","id":1,"result":null}
    ```
4. Hedera (current behaviour)

    ```bash
    -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["0x1000000000000000000000000000000000000000000000000000000000000001"]}'
    ```
    Response:
    
    ```json
    {
      "error": {
        "code": -32603,
        "message": "[Request ID: ...] Error invoking RPC: Cannot read properties of null (reading 'number')"
      },
      "jsonrpc": "2.0",
      "id": 1
    }
    ```

### Solution

Update the `eth_getBlockReceipts` method in the relay to check if the block hash exists, and if it doesn't, return `null` as the result, instead of throwing an internal error.

This behavior is more consistent with both:
 - Ethereum JSON-RPC expectations
 - OpenRPC specification

### Alternatives

An alternative would be to retain the current behavior and document the difference clearly in the API documentation. However, this would place a burden on client developers and break compatibility with existing Ethereum tooling. `This is not the preferred solution.`

mwb-al:
**Description**:
When querying the eth_getBlockReceipts method in Hedera’s JSON-RPC API with a valid but nonexistent block hash, the endpoint throws an error. This behavior is inconsistent with Ethereum-compatible JSON-RPC endpoints (such as Sepolia), which return null for nonexistent blocks, as per the standard JSON-RPC specification.

**Related issue(s)**:

Fixes #3813 

**Checklist**

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -1108,7 +1108,7 @@ export class EthImpl implements Eth {
    *
    * @param {string } blockHashOrBlockNumber The block hash, block number, or block tag
    * @param {RequestDetails} requestDetails The request details for logging and tracking
-   * @returns {Promise<Receipt[]>} Array of transaction receipts for the block
+   * @returns {Promise<Receipt[] | null>} Array of transaction receipts for the block or null if block not found
    */
   @rpcMethod
   @rpcParamValidationRules({
@@ -1120,7 +1120,7 @@ export class EthImpl implements Eth {
   public async getBlockReceipts(
     blockHashOrBlockNumber: string,
     requestDetails: RequestDetails,
-  ): Promise<ITransactionReceipt[]> {
+  ): Promise<ITransactionReceipt[] | null> {
     return await this.blockService.getBlockReceipts(blockHashOrBlockNumber, requestDetails);
   }
 
--- a/packages/relay/src/index.ts
+++ b/packages/relay/src/index.ts
@@ -71,7 +71,7 @@ export interface Eth {
 
   getBalance(account: string, blockNumber: string | null, requestDetails: RequestDetails): Promise<string>;
 
-  getBlockReceipts(blockHashOrNumber: string, requestDetails: RequestDetails): Promise<ITransactionReceipt[]>;
+  getBlockReceipts(blockHashOrNumber: string, requestDetails: RequestDetails): Promise<ITransactionReceipt[] | null>;
 
   getBlockByHash(hash: string, showDetails: boolean, requestDetails: RequestDetails): Promise<Block | null>;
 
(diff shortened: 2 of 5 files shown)


packages/relay/src/lib/eth.ts
```diff
@@ -1108,7 +1108,7 @@ export class EthImpl implements Eth {
    *
    * @param {string } blockHashOrBlockNumber The block hash, block number, or block tag
    * @param {RequestDetails} requestDetails The request details for logging and tracking
-   * @returns {Promise<Receipt[]>} Array of transaction receipts for the block
+   * @returns {Promise<Receipt[] | null>} Array of transaction receipts for the block or null if block not found
```
acuarica, line 1111: Why the difference between `Receipt` and `ITransactionReceipt` in the return type? (I see it was already like this before the change, but if it's a trivial fix we should do it).

acuarica (approved):
nice

acuarica (approved):

simzzz (approved):
LGTM

quiet-node (approved):
LGTM

konstantinabl (approved):

PR #3865, opened 2025-06-18
feat: add MAX_GAS_ALLOWANCE_HBAR configuration

### LINKED ISSUE #3880: MAX_GAS_ALLOWANCE tracking issue

Neurone:
**Description**:

Some providers need to fully subsidise their users' transactions. This option is already available on the network and in the SDK by setting the [`max_gas_allowance`](https://github.com/hashgraph/hedera-protobufs/blob/8c27786cec93abab974309074feaef9b48a695b7/services/ethereum_transaction.proto#L110) parameter accordingly.

This PR enables this parameter to be set via the JSON-RPC Relay configuration as well.

**Related issue(s)**:

closes #3880 

**Notes for reviewer**:

Now both the brand new `MAX_GAS_ALLOWANCE_HBAR` and the existing `GAS_PRICE_TINY_BAR_BUFFER` env variables are loaded dynamically via ConfigService.

**Checklist**


packages/relay/src/lib/constants.ts
```diff
@@ -236,6 +236,7 @@ export default {
   },
 
   MAX_TRANSACTION_FEE_THRESHOLD: ConfigService.get('MAX_TRANSACTION_FEE_THRESHOLD'),
+  MAX_GAS_ALLOWANCE_HBAR: ConfigService.get('MAX_GAS_ALLOWANCE_HBAR'),
```
natanasow, line 239: Q: why do we add it here when we use it straight from the ConfigService everywhere via `ConfigService.get('MAX_GAS_ALLOWANCE_HBAR')`?

natanasow (approved):
LGTM, one question.

simzzz (approved):

packages/relay/src/lib/constants.ts
Neurone: I've noticed that other constant parameters (e.g. `GAS_PRICE_TINY_BAR_BUFFER`) are used from both the constants object and the current live env values. I therefore assumed that this was the best practice to save the startup value in case someone wants to access it instead of the current value.

As you correctly noted, at the moment `MAX_GAS_ALLOWANCE_HBAR` is never read from the constants object, so if you prefer I can remove it from there.

Neurone: Hi @natanasow I removed `MAX_GAS_ALLOWANCE_HBAR` from the constants because at the moment it is only used via the ConfigService.

docs/configuration.md
```diff
@@ -103,6 +103,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
+| `MAX_GAS_ALLOWANCE_HBAR`                    | "0"                            | The maximum amount, in hbars, that the relay is willing to pay to complete the transaction in case the senders don't provide enough funds to pay for their transactions. Please note, in case of full subsidized transactions, the sender must set the gas price to `0` and the relay must configure both the `GAS_PRICE_TINY_BAR_BUFFER`, to a number big enough to cover the discrepancy between the base gas price and `0`, and the `MAX_GAS_ALLOWANCE_HBAR` parameters, with the max amount of hbars the relay is willing to spend to execute the transaction.      |
```
acuarica, line 106: It would be better if you can insert this entry in alphabetic order.

.env.http.example
```diff
@@ -43,6 +43,7 @@ OPERATOR_KEY_MAIN=               # Operator private key used to sign transaction
 # FEE_HISTORY_MAX_RESULTS=10     # Maximum number of results to return for eth_feeHistory
 # TX_DEFAULT_GAS=400000          # Default gas for transactions that don't specify gas
 # MAX_TRANSACTION_FEE_THRESHOLD=15000000 # Max transaction fee paid by relay operator account
+# MAX_GAS_ALLOWANCE_HBAR=0       # The maximum amount, in hbars, that the relay is willing to pay to complete the transaction in case the senders don't provide enough funds.
```
acuarica, line 46: See comment in `docs/configuration.md` about notation.

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -1511,6 +1512,57 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+            const info = await mirrorNode.get(`/contracts/results/${transactionHash}`, requestId);
+            expect(info).to.have.property('contract_id');
+            expect(info.contract_id).to.not.be.null;
+            expect(info).to.have.property('created_contract_ids');
+            expect(info.created_contract_ids.length).to.be.equal(1);
+
+            const balanceAfter = await relay.getBalance(accounts[2].wallet.address, 'latest', requestId);
+            expect(balanceAfter).to.be.equal(balanceBefore);
+
+            const transaction2 = {
```
acuarica, line 1543: isn't this a different test case? If yes, we should split it.

docs/configuration.md
acuarica, line 106: I'm not sure about setting `GAS_PRICE_TINY_BAR_BUFFER` in a fully subsidized JSON-RPC Relay. It seems `GAS_PRICE_TINY_BAR_BUFFER` has a different purpose. We might need to change the precheck to account for the `MAX_GAS_ALLOWANCE_HBAR` variable.

acuarica (commented):
Thanks @Neurone for sending this. It looks good so far. Have a couple of questions (we can tackle them in following PRs if needed)

- Should we include a tool example (in the `tools/` folder) showcasing this feature?
- What happens when the sender account does not exist? Before this feature, this wasn't possible because the sender need to have HBARs. Now that's not the case anymore. We should add a test for this scenario.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -172,6 +172,8 @@ export class SDKClient {
       ),
     );
 
+    ethereumTransaction.setMaxGasAllowanceHbar(ConfigService.get('MAX_GAS_ALLOWANCE_HBAR'));
```
acuarica, line 175: I'd like to include a comment to the docs for reference, _i.e._, https://docs.hedera.com/hedera/sdks-and-apis/sdks/smart-contracts/ethereum-transaction

.env.http.example
Neurone: Hi @acuarica , the amount is actually in hbar. I noticed the js doc was wrong while working on this PR, so I fixed it (https://github.com/hiero-ledger/hiero-sdk-js/commit/b666e2ad37c2b38178ead46bf8e4de4d18b2fb28) :D

packages/server/tests/acceptance/rpc_batch1.spec.ts
Neurone: The test case seems the same to me, I'm just testing both legacy (type 0) and modern (type 1/2) transactions work correctly with this configuration. Do you prefer I split this in two?

docs/configuration.md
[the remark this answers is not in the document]
Neurone: Updated the doc

Neurone: Done. I moved also the other variables.

Neurone:
> * Should we include a tool example (in the `tools/` folder) showcasing this feature?

I prefer to merge this and then thinking about specific tools, but at least the new test cases already show how you can play with this parameter and obtain a fully subsidised transaction.

> * What happens when the sender account does not exist? Before this feature, this wasn't possible because the sender need to have HBARs. Now that's not the case anymore. We should add a test for this scenario.

I'm not sure I get the question. This parameter does not introduce a new feature, it just give the relay operators the ability to set that value as they prefer. In case they set a big enough value, and the sender set gasPrice to 0, the total cost of the transaction (network fees and gas consumption) will be on the operator, and nothing will be charged to the sender ([here an example](https://hashscan.io/testnet/transaction/1750866140.036758000)).

There's no other change in the transaction, the sender (the transaction's signer) is still there, untouched as usual.

docs/configuration.md
Neurone: The `MAX_GAS_ALLOWANCE_HBAR` is set by the operator, so I'm not sure it needs a dedicated pre-check.

About `GAS_PRICE_TINY_BAR_BUFFER`, you are right, that serve a different purpose, and that's fine.

That parameter is mentioned here only as a reference, because in case the user set gasPrice=0, a low buffer value (that is probably what you want in a classic JSON-RPC relay configuration) will block those transactions during the pre-check.

So, in case an operator want to support gasPrice=0 transactions, it must update the `GAS_PRICE_TINY_BAR_BUFFER` parameter also accordingly.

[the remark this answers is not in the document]
acuarica: Nice! So that means that the docs https://docs.hedera.com/hedera/sdks-and-apis/sdks/smart-contracts/ethereum-transaction are outdated: "Max Allowance: The maximum amount of HBAR (specified in tinybars) the payer is willing to cover for the gas consumed during transaction execution. [...]"

acuarica: > So, in case an operator want to support gasPrice=0 transactions, it must update the GAS_PRICE_TINY_BAR_BUFFER parameter also accordingly.

Exactly. I believe this is not the best UX for operators, as we are requiring them to update the unrelated `GAS_PRICE_TINY_BAR_BUFFER` variable.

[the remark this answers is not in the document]
Neurone: yep, correct. I sent a PR also there :D

Neurone: We can think about improving the operators UX in general, but probably just providing a preset of config values for the different use cases they want to achieve is enough.

The `GAS_PRICE_TINY_BAR_BUFFER` is mentioned in the comments because it's involved in the process, but it's not a constraint to set it to a specific value. Operators can still decide to keep that value small as they prefer, and pay only for txs which costs is in that specific cost range.

BTW, the scope of this PR is very limited: it's just enable operators setting the `MAX_GAS_ALLOWANCE_HBAR` value as they prefer, like they do for other parameters, so I would suggest to postpone further improvements for another PR.

packages/server/tests/acceptance/rpc_batch1.spec.ts
Neurone: Done.

packages/relay/src/lib/precheck.ts
```diff
@@ -169,10 +170,10 @@ export class Precheck {
     const passes = txGasPrice >= networkGasPrice || Precheck.isDeterministicDeploymentTransaction(tx);
```
acuarica, line 170: This was discussed offline with @Neurone to not force the Relay provider to set `GAS_PRICE_TINY_BAR_BUFFER` when `MAX_ALLOWANCE` is set.

```suggestion
    const passes = txGasPrice >= networkGasPrice || Precheck.isDeterministicDeploymentTransaction(tx) || (txGasPrice === 0n && ConfigService.get('MAX_ALLOWANCE') > 0;
```

docs/configuration.md
Neurone: I modified the comments to adhere to the new precheck behaviour.

simzzz (approved):

Neurone:
I needed to squash the commits in a single signed commit (the rebase via GitHub removed the signatures of the commits and the merge was blocked because of that).

Neurone:
Hi @natanasow @acuarica @simzzz , can you tell me when do you plan to merge this PR? 

I continue to update the feature because the main branch progresses while this PR remains open, but the new update lasts for few days and then new new conflicts arrive and the need for another update.

Please warn me when you want to merge, and I'll update the branch only once.

acuarica:
Hi @Neurone, I'm taking a look at this, ensuring CI passes successfully. Will keep you posted.

acuarica:
@Neurone CI is failing here https://github.com/hiero-ledger/hiero-json-rpc-relay/actions/runs/16263200851/job/46028880961?pr=3865#step:17:11603.

```
       Acceptance tests
         @api-conformity
           @api-batch-1 RPC Server Acceptance Tests
             RPC Server Acceptance Tests
               Transaction related RPC Calls
                 Check subsidizing gas fees
                   given MAX_GAS_ALLOWANCE_HBAR=100 are set
                     should execute a pre EIP-1559 transaction with "eth_sendRawTransaction" and pays the total amount of the fees on behalf of the sender:
     Error: server response 500 Internal Server Error (request={  }, response={  }, error=null, info={ "requestUrl": "http://localhost:7546", "responseBody": "{\"error\":{\"code\":-32603,\"message\":\"[Request ID: 6f588d46-fd70-4fe4-b61b-771983f592d7] Error invoking RPC: No matching transaction record retrieved: transactionId=0.0.1002@1752598042.343293633\"},\"jsonrpc\":\"2.0\",\"id\":185}", "responseStatus": "500 Internal Server Error" }, code=SERVER_ERROR, version=6.13.5)
      at makeError (packages/server/node_modules/ethers/src.ts/utils/errors.ts:694:21)
      at assert (packages/server/node_modules/ethers/src.ts/utils/errors.ts:715:25)
      at FetchResponse.assertOk (packages/server/node_modules/ethers/src.ts/utils/fetch.ts:950:15)
      at JsonRpcProvider._send (packages/server/node_modules/ethers/src.ts/providers/provider-jsonrpc.ts:1245:18)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at /home/runner/_work/hiero-json-rpc-relay/hiero-json-rpc-relay/packages/server/node_modules/ethers/src.ts/providers/provider-jsonrpc.ts:537:40
```

Neurone:
> @Neurone CI is failing here https://github.com/hiero-ledger/hiero-json-rpc-relay/actions/runs/16263200851/job/46028880961?pr=3865#step:17:11603.
> 
> ```
>   1) RPC Server Acceptance Tests
>        Acceptance tests
>          @api-conformity
>            @api-batch-1 RPC Server Acceptance Tests
>              RPC Server Acceptance Tests
>                Transaction related RPC Calls
>                  Check subsidizing gas fees
>                    given MAX_GAS_ALLOWANCE_HBAR=100 are set
>                      should execute a pre EIP-1559 transaction with "eth_sendRawTransaction" and pays the total amount of the fees on behalf of the sender:
>      Error: server response 500 Internal Server Error (request={  }, response={  }, error=null, info={ "requestUrl": "http://localhost:7546", "responseBody": "{\"error\":{\"code\":-32603,\"message\":\"[Request ID: 6f588d46-fd70-4fe4-b61b-771983f592d7] Error invoking RPC: No matching transaction record retrieved: transactionId=0.0.1002@1752598042.343293633\"},\"jsonrpc\":\"2.0\",\"id\":185}", "responseStatus": "500 Internal Server Error" }, code=SERVER_ERROR, version=6.13.5)
>       at makeError (packages/server/node_modules/ethers/src.ts/utils/errors.ts:694:21)
>       at assert (packages/server/node_modules/ethers/src.ts/utils/errors.ts:715:25)
>       at FetchResponse.assertOk (packages/server/node_modules/ethers/src.ts/utils/fetch.ts:950:15)
>       at JsonRpcProvider._send (packages/server/node_modules/ethers/src.ts/providers/provider-jsonrpc.ts:1245:18)
>       at processTicksAndRejections (node:internal/process/task_queues:95:5)
>       at /home/runner/_work/hiero-json-rpc-relay/hiero-json-rpc-relay/packages/server/node_modules/ethers/src.ts/providers/provider-jsonrpc.ts:537:40
> ```

Hi @acuarica , I spot the error. With the last merge the `Constants.CONTRACT_CODE_SIZE_LIMIT` is not there anymore. I updated the tests, now they should work again.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -170,6 +170,10 @@ export class SDKClient {
       ),
     );
 
+    // If the authorized fee from the Ethereum sender is insufficient, the payer of the transaction is charged up to the maxGasAllowance.
+    // see "Max Allowance" in the docs for more details https://docs.hedera.com/hedera/sdks-and-apis/sdks/smart-contracts/ethereum-transaction
+    ethereumTransaction.setMaxGasAllowanceHbar(ConfigService.get('MAX_GAS_ALLOWANCE_HBAR'));
```
quiet-node, line 175: Ha nice the POC of Paymaster does the same thing https://github.com/hiero-ledger/hiero-json-rpc-relay/compare/poc-Paymaster-functionality-at-relay-level#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR184.

quiet-node (commented):
I believe this is another variant of what we refer to as the Paymaster feature https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3917. The goal is essentially the same, but we’ve formalized the approach and named it Paymaster.

The mode is designed to be configurable and offers broader coverage by allowing a set of whitelisted transaction signers to benefit from it. Functionally, Paymaster operates similarly to your solution—just with a bit more structure and flexibility.

Since Paymaster is currently in the planning and design phase and is expected to be delivered soon, do you think it makes sense to hold off on this and wait for the more official solution? Open to your thoughts let me know what you think.

quiet-node:
> Hi @quiet-node this PR does not introduce a new feature, it enables an existing SDK/network configuration to be set by the relay operator, like any other parameter. In specific cases, if configured accordingly by the relay operator, this configuration can be used to fully subsidize user transactions.
> 
> The paymaster is a nice feature to have, but it does much more than this, it adds the concept of paymaster, it adds specific relay-only configurations, a whitelist, etc. and potentially it can evolve in additional features.
> 
> So, to me, they should remain different topics.

Hey @Neurone thanks for the information yes that makes sense!

quiet-node (approved):

PR #3876, opened 2025-06-25
feat: edit `rpc_relay_consensusnode_*` metrics

### LINKED ISSUE #3853: Audit and Refactor Metrics to Avoid High Cardinality from Unique Labels
###  Description

During routine observability reviews, we've identified an opportunity to optimize the performance and cost-efficiency of our metrics system. Specifically, we observed patterns where certain labels may have too many possible values, leading to increased cardinality in Prometheus metrics.

In Prometheus, each unique combination of metric name and label values generates a separate **timeseries**. When label values are highly variable—especially when tied to request-specific or user-specific data—the number of timeseries can grow significantly. This can impact:

* Memory and storage overhead
* Query latency and dashboard performance
* Long-term cost and scalability

To ensure long-term observability efficiency, we propose an audit and refactor of metrics across the Relay codebase to eliminate or minimize high-cardinality label usage.

---

###  Solution

We will undertake the following actions:

1. **Review** all Prometheus metric definitions for labels that may produce excessive variability.
2. **Refactor** metrics to:
   * Remove non-essential high-cardinality labels
   * Replace with broader categorical labels where appropriate (e.g., `status`, `region`, `type`)

---

###  Acceptance Criteria

* [ ] Comprehensive audit of all Relay metrics and associated labels
* [ ] Refactored metrics with reduced or mitigated cardinality impact

---

### Resources

* [Prometheus Docs – Metric and Label Naming](https://prometheus.io/docs/practices/naming/)
* [Prometheus Docs – Avoiding High Cardinality](https://prometheus.io/docs/practices/instrumentation/#dont-overuse-labels)

natanasow:
**Description**:

Recently, we audited all metrics (private page due to security reasons) and how they can be improved more efficiently. In this PR, we remove some of the `rpc_relay_consensusnode_*` labels.

**Related issue(s)**:

Fixes #3853 

**Notes for reviewer**:

**Checklist**


packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -209,7 +208,6 @@ export default class MetricService {
     callerName,
```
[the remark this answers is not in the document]
acuarica: I'd say remove from both. We usually log variables that are used later, but in this case `callerName` is not used later (only for logging). Moreover, doing a quick search in our logs, the only value logged for `callerName` is `eth_sendRawTransaction`. So we don't lose information even if we remove it.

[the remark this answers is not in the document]
natanasow: Sounds good, thank you!

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -206,22 +203,20 @@ export default class MetricService {
     executionMode,
     transactionId,
     txConstructorName,
-    callerName,
     cost,
     gasUsed,
-    interactingEntity,
```
acuarica, line 212: should we remove these properties from `IExecuteQueryEventPayload` as well? what do you think?

acuarica (approved):

konstantinabl (dismissed):

packages/relay/src/lib/services/metricService/metricService.ts
acuarica: You're right. Sorry I wasn't clear enough. Let me clarify. What I meant is to remove both `callerName` and `interactingEntity` wherever we use `IExecuteQueryEventPayload`. Like in

https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/ff0dd5b3d0a60a491947f2521fb80bdd4762cb96/packages/relay/src/lib/clients/sdkClient.ts#L386-L397

wdyt?

natanasow: Oh, totally makes sense. Removing them.

simzzz (approved):

PR #3896, opened 2025-07-03
feat: introduce `READ_ONLY` mode

### LINKED ISSUE #3878: Json RPC Relay  READ ONLY OR READ-WRITE mode
**Context**

When I run the json rpc-relay I want to be able to start either on read-only mode or read-write mode, specifying the mode on the config file.
When I configured it on read-only mode, no CN call can be done, and any rpc call that is CN related should raise an error as "RPC method not allowed on read-only mode".  When I have read only mode Hbar limiter should be deactivated
When i run on write-mode, operator account should specified, and if not, relay shoul be unable to  startup.
When i run on write-mode,  Hbar Rate limiter can be enabled

**Acceptance criteria**

- A new  Config entry should be added called JSON:RPC_RELAY_MODE, which values  read-only|  read-write ( by default)
- Config endpoint should reflect the new mode entry
- Startup log should clearly specify " Starting Json Rpc Relay  in  ? mode
- If read-only

-- I operator account is set we will warn that is set, but will not be used, a it is on read mode.

-- Hbar rate limiter cannot be enabled and  a Warning message will indicate that Hbar limiter will be not enabled, if it is enabled by config

-- If eth_sendtransaction or other rpc method that invoke the CN  are requestrf. relay should  give and error response  should be given with the associated code, saying "RPC method not allowed on read-only mode"

-- ETH_CALL_DEFAULT_TO_CONSENSUS_NODE cannot be enabled and  a Warning message will indicate that ethcall will be not be sent to CN, an fail on every request

-- ETH_CALL_CONSENSUS_SELECTORS will not be applicable 

- If read-write

-- If no operator account is set, Relay should halt on startup, with a clear message "Read-Write mode enforces to have an operator account set"
-- On read write mode, HBAR rate limiter can be enabled, if its enabled

**Pending Items**

-Error Codes definitions

**Tech Detaiils**

- ConfigService
-- add the new flags
-- ValidationService in the ConfigService: Check the config and override, warn/fail based on values and mode 
-- ? Server 

--- Fail  on sendRawTx if o

- Documentation

acuarica:
**Description**:

**Related issue(s)**:

Fixes #3878.

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/config-service/src/services/globalConfig.ts
+++ b/packages/config-service/src/services/globalConfig.ts
@@ -602,7 +602,7 @@ const _CONFIG = {
   OPERATOR_ID_MAIN: {
     envName: 'OPERATOR_ID_MAIN',
     type: 'string',
-    required: true,
+    required: false,
     defaultValue: null,
   },
   OPERATOR_KEY_ETH_SENDRAWTRANSACTION: {
@@ -620,7 +620,7 @@ const _CONFIG = {
   OPERATOR_KEY_MAIN: {
     envName: 'OPERATOR_KEY_MAIN',
     type: 'string',
-    required: true,
+    required: false,
     defaultValue: null,
   },
   RATE_LIMIT_DISABLED: {
@@ -640,6 +640,12 @@ const _CONFIG = {
     type: 'string',
     required: false,
     defaultValue: null,
+  },
+  READ_ONLY: {
+    envName: 'READ_ONLY',
+    type: 'boolean',
+    required: false,
+    defaultValue: false,
   },
   REDIS_RECONNECT_DELAY_MS: {
     envName: 'REDIS_RECONNECT_DELAY_MS',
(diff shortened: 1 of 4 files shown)


acuarica:
We are considering to include one additional requirement as part of this PR. That is, to ensure the operator has positive balance before starting the JSON-RPC Relay/WebSocket. This requirement can be seen as new validation, which is independent of the Read-Only mode feature.

This simply change, however, has some implications. Currently, the start of the Relay is sync, _i.e._, there are no promises running until Koa server starts listening. This means that the validation logic needs to be placed in both Relay and WebSocket's `main`. But we don't currently have tests at `main` level. Only at `server.ts` and `webSocketServer.ts` levels respectively.

Therefore, we can either

- **1.** drop this requirement altogether, or
-  **2.** we can implement this in a following PR given the amount of changes that this may require.

PR #3902, opened 2025-07-08
feat: remove `ETH_CALL_CONSENSUS_SELECTORS` env

### LINKED ISSUE #2984: Revisit ETH_CALL_CONSENSUS_SELECTORS to make sure if it's safe to remove contained selectors
### Problem

In some cases, there are features at a moment supported by the consensus node but not yet on mirror node. Therefore, the Relay implemented the force-to-consensus-node-by-selector logic to whitelist a certain array of selectors and allow them to be called against consensus node instead of mirror node. However, in the later version updates when these features are fully supported in mirror node, the selectors should be removed from the array.

### Solution

Revisit the ETH_CALL_CONSENSUS_SELECTORS env variable to remove fully supported selectors:

Current selectors included:
- [ ] IHRC719.isAssociated(address) - "0xd55fe582"

### Alternatives

_No response_

natanasow:
**Description**:

At times, certain features are supported by the consensus node but not yet by the mirror node. To address this, the Relay implemented the force-to-consensus-by-selector logic, which whitelists specific selectors and allows them to be called against the consensus node instead of the mirror node. However, as these features become fully supported by the mirror node in later updates, the corresponding selectors should be removed from the array.

**Related issue(s)**:

Fixes #2984 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/services/ethService/contractService/ContractService.ts
+++ b/packages/relay/src/lib/services/ethService/contractService/ContractService.ts
@@ -806,7 +806,7 @@ export class ContractService implements IContractService {
   }
 
   /**
-   * Routes the call to either consensus or mirror node based on configuration and selector.
+   * Routes the call to either consensus or mirror node based on configuration.
    *
    * @param {IContractCallRequest} call - The call request
    * @param {number | null} gas - The gas limit


simzzz (commented):
Looks good, just one JSDoc comment left to be updated

simzzz (approved):

acuarica (approved):

PR #3928, opened 2025-07-11
chore: creates tests for connectionLimiter file

konstantinabl:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


quiet-node (dismissed):
LGTM! Nice set of tests! Great work as always!

packages/ws-server/tests/unit/wsMetricRegistry.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+        'totalClosedConnections',
+      ];
+
+      counterMetrics.forEach((metric) => {
+        sinon.assert.calledWith(removeSingleMetricStub, WS_CONSTANTS[metric].name);
+      });
+    });
+
+    it('should initialize all histogram metrics', function () {
+      const histogramMetrics = ['connectionDuration', 'messageDuration'];
```
acuarica, line 41: should we use parameterized tests here so each `calledWith` assert is independent?

acuarica (approved):

packages/ws-server/tests/unit/wsMetricRegistry.spec.ts
konstantinabl: Yes, actually in this case if first assert fails the whole test would fail, so my approach is not optimal. Will change it now

acuarica (approved):

PR #3930, opened 2025-07-11
feat: validate that operator account has positive balance

### LINKED ISSUE #3929: Ensure the operator account has positive balance in Read-Write mode
When the Relay/WebSocket is initiated in non Read-Only mode https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3896, we should validate that the operator account has positive balance.

In case the operator account does not have positive balance (or does not exist), then the Relay/WebSocket must immediately terminate with an error exit code.

acuarica:
**Description**:

**Related issue(s)**:

Fixes #3929.

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/relay/src/lib/relay.ts
+++ b/packages/relay/src/lib/relay.ts
@@ -329,4 +329,14 @@ export class Relay {
   mirrorClient(): MirrorNodeClient {
     return this.mirrorNodeClient;
   }
+
+  async ensureOperatorHasBalance() {
+    if (!ConfigService.get('READ_ONLY')) {
+      const operator = this.clientMain.operatorAccountId!.toString();
+      const balance = await this.ethImpl.getBalance(operator, 'latest', {} as RequestDetails);
+      if (BigInt(balance) === BigInt(0)) {
+        throw new Error(`Operator account \`${operator}\` has no balance`);
+      }
+    }
+  }
 }
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -1,10 +1,18 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import app from './server';
 import { ConfigService } from '@hashgraph/json-rpc-config-service/dist/services';
+
 import { setServerTimeout } from './koaJsonRpc/lib/utils'; // Import the 'setServerTimeout' function from the correct location
+import app, { logger, relay } from './server';
 
 async function main() {
+  try {
+    await relay.ensureOperatorHasBalance();
+  } catch (error) {
+    logger.fatal(error);
+    process.exit(1);
+  }
+
   const server = app.listen({ port: ConfigService.get('SERVER_PORT'), host: ConfigService.get('SERVER_HOST') });
 
   // set request timeout to ensure sockets are closed after specified time of inactivity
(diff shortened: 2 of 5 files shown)


packages/relay/src/lib/relay.ts
```diff
@@ -329,4 +329,16 @@ export class Relay {
...
+
+    const operator = this.clientMain.operatorAccountId!.toString();
+    const balance = BigInt(await this.ethImpl.getBalance(operator, 'latest', {} as RequestDetails));
+    if (balance === BigInt(0)) {
+      throw new Error(`Operator account '${operator}' has no balance`);
+    } else {
+      this.logger.info(`Operator account '${operator}' has balance: ${balance}`);
+    }
+  }
 }
```
quiet-node, line 344: I feel like the ensureOperatorHasBalance() function could be done a little differently as it seems to pull in quite a few components. Like relying on both clientMain and ethImpl makes it feel somewhat fragile, since it depends on both being properly set up. On top of that, using ethImpl.getBalance, an app-level API that itself relies on MirrorNodeClient, cacheService, Logger, eventEmitter, etc. being correctly initialized — just to validate a config value feels a bit circular.

So would it make more sense, or be possible, to:
	•	Get the operator ID directly from ConfigService via ConfigService.get('OPERATOR_ID')
	•	Fetch the balance straight from the Mirror Node using the /balances endpoint via ConfigService.get('MIRROR_NODE_URL')

That way, we could skip involving ethImpl altogether and avoid any extra logic or validation tied to eth_getBalance. I know it’s a lightweight call, but since it’s an app-level API and could be programmatically updated, it makes the function feel a bit less stable for this job.

At the end I just think the fewer moving parts we rely on, the better at this stage of validation.

quiet-node: Also when I first heard about validating the operator balance, I initially thought it could be done inside hapiService during client initialization or some sort since that’s where we set up the operator keys and IDs. It felt like a natural place to include a balance check as part of that process.

Maybe it’s not feasible because it would require the function to be async or maybe we could just use .then() or something along those lines. Either way, it feels like the balance validation belongs there. We’re initializing the SDK client, and if the operator doesn’t have enough balance, it makes sense to catch that early and fail right at that level.

quiet-node (commented):
LG left some thought

acuarica:
After an offline discussion with the team, we came to the conclusion that the start up validation **shouldn't** be in the

- `Precheck` class, as it deals with "prechecks before sending a raw transaction",
- `validators` module, as it deals with validating request payloads, 
- `config-service` package because the dependencies needed and the `async` check.

Moreover, there is already an `async` start up check in the Relay

https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/c1224f4dc0c972562133ddb03a0095e5dd182b0a/packages/relay/src/lib/relay.ts#L210

https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/c1224f4dc0c972562133ddb03a0095e5dd182b0a/packages/relay/src/lib/relay.ts#L250-L259

so having `ensureOperatorHasBalance` in the Relay might not be such a bad idea after all.

PR #3951, opened 2025-07-17
feat: adds new decorators following the typescript v5 standard

### LINKED ISSUE #3593: Update Decorators to Align with TypeScript 5 Changes - P3
Description:

After PR https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3581, the implementation will be using decorators with TypeScript 4.x, requiring the `experimentalDecorators` flag in tsconfig.json. However, TypeScript 5 introduces changes to decorator behavior, as outlined in [this resource](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#differences-with-experimental-legacy-decorators).

To ensure compatibility and future-proof our codebase, we need to research these changes and modify our implementation accordingly.

Solution:
- Research the resource and update the code to align with new behaviors

simzzz:
**Description**:
Adds new decorators with the same functionality as our existing ones following the typescript v5 standard. The new decorators are suffixed with `Standard`.  When we update the repo version, the old ones should be deleted (or suffixed with `Legacy` for example if we deem them to be necessary), and the new ones should be renamed e.g. `cacheStandard` -> `cache`. No further changes to the code will be required

**Related issue(s)**:

Fixes #3593 

**Notes for reviewer**:
It was tricky to set up typescript v5 for an automated test, but I have manually tested and verified the four new decorators work in a local mini sandbox environment that I have set up. 

There was one ugly way to achieve this task without creating new decorators. As of v4, the decorators take 3 params - `target`, `propertyKey`, `descriptor`. As of v5 it's only 2 `target` and `context`. So if we rename the 2nd param (which we don't use) to `context` and we check conditionally if it's a context object in order to execute the v5 logic. If not, we proceed with handling the `descriptor` object which would have been ignored in v5. However I decided against this approach since it would create confusion

**Checklist**

### DIFF
--- a/packages/relay/src/lib/decorators/cache.decorator.ts
+++ b/packages/relay/src/lib/decorators/cache.decorator.ts
@@ -27,23 +27,139 @@ interface CacheOptions {
 }
 
 /**
- * Iterates through the provided 'params' array and checks if any argument in 'args' at the specified 'index'
- * matches one of the pipe-separated values in 'value'. If a match is found, caching should be skipped.
+ * This decorator uses a `CacheService` to attempt to retrieve a cached result before executing the original method. If
+ * no cached response exists, the method is executed and its result may be stored in the cache depending on configurable
+ * options. Caching can be conditionally skipped based on runtime arguments via `skipParams` (for positional args)
+ * and `skipNamedParams` (for object args).
  *
- * @param args - The arguments passed to the method in an array
- * @param params - An array of CacheSingleParam caching rules
- * @returns 'true' if any argument matches a rule and caching should be skipped; otherwise, 'false'.
+ * Legacy decorator for TypeScript 4.x with --experimentalDecorators.
+ * For TypeScript 5+ standard decorators, use @cacheStandard instead.
+ *
+ * @param cacheService - The caching service used to store and retrieve cache entries.
+ * @param options - Optional configuration for caching behavior.
+ *   @property skipParams - An array of rules for skipping caching based on specific argument values.
+ *   @property skipNamedParams - An array of rules for skipping caching based on fields within argument objects.
+ *   @property ttl - Optional time-to-live for the cache entry; falls back to global config if not provided.
+ *
+ * @returns A method decorator function that wraps the original method with caching logic.
  *
  * @example
- *   [{
- *     index: '0',
- *     value: 'pending|safe'
+ *   @cache(CacheService, { skipParams: [...], skipNamesParams: [...], ttl: 300 })
+ */
+export function cache(cacheService: CacheService, options: CacheOptions = {}) {
+  return function (_target: any, _propertyKey: string, descriptor: PropertyDescriptor) {
+    const method = descriptor.value;
+
+    descriptor.value = async function (...args: unknown[]) {
+      const requestDetails = extractRequestDetails(args);
+      const cacheKey = generateCacheKey(method.name, args);
+
+      const cachedResponse = await cacheService.getAsync(cacheKey, method, requestDetails);
+      if (cachedResponse) {
+        return cachedResponse;
+      }
+
+      const result = await method.apply(this, args);
+      if (
+        result &&
+        !shouldSkipCachingForSingleParams(args, options?.skipParams) &&
+        !shouldSkipCachingForNamedParams(args, options?.skipNamedParams)
+      ) {
+        await cacheService.set(
+          cacheKey,
+          result,
+          method,
+          requestDetails,
+          options?.ttl ?? ConfigService.get('CACHE_TTL'),
+        );
+      }
+
+      return result;
+    };
+  };
+}
+
+/**
+ * TypeScript 5+ standard cache decorator.
+ * This is the clean, modern version for TypeScript 5+ without legacy compatibility.
+ *
+ * Uses a `CacheService` to attempt to retrieve a cached result before executing the original method. If
+ * no cached response exists, the method is executed and its result may be stored in the cache depending on configurable
+ * options. Caching can be conditionally skipped based on runtime arguments via `skipParams` (for positional args)
+ * and `skipNamedParams` (for object args).
+ *
+ * @param cacheService - The caching service used to store and retrieve cache entries.
+ * @param options - Optional configuration for caching behavior.
+ *   @property skipParams - An array of rules for skipping caching based on specific argument values.
+ *   @property skipNamedParams - An array of rules for skipping caching based on fields within argument objects.
+ *   @property ttl - Optional time-to-live for the cache entry; falls back to global config if not provided.
+ *
+ * @returns A method decorator function that wraps the original method with caching logic.
+ *
+ * @example
+ *   @cacheStandard(CacheService, { skipParams: [...], skipNamesParams: [...], ttl: 300 })
+ */
+export function cacheStandard(cacheService: CacheService, options: CacheOptions = {}) {
+  return function (target: any, context: ClassMethodDecoratorContext): void {
+    if (context.kind !== 'method') {
+      throw new Error(`@cacheStandard can only be applied to methods, received: ${context.kind}`);
+    }
+
+    const methodName = String(context.name);
+
+    // Use addInitializer to replace the method implementation with cached version
+    context.addInitializer(function (this: any) {
+      const originalMethod = this[context.name as string | symbol];
+
+      // Replace the method with the cached version
+      this[context.name as string | symbol] = async function (this: any, ...args: unknown[]) {
+        const requestDetails = extractRequestDetails(args);
+        const cacheKey = generateCacheKey(methodName, args);
+
+        const cachedResponse = await cacheService.getAsync(cacheKey, methodName, requestDetails);
+        if (cachedResponse) {
+          return cachedResponse;
+        }
+
+        const result = await originalMethod.apply(this, args);
+        if (
+          result &&
+          !shouldSkipCachingForSingleParams(args, options?.skipParams) &&
+          !shouldSkipCachingForNamedParams(args, options?.skipNamedParams)
+        ) {
+          await cacheService.set(
+            cacheKey,
+            result,
+            methodName,
+            requestDetails,
+            options?.ttl ?? ConfigService.get('CACHE_TTL'),
+          );
+        }
+
+        return result;
+      };
+    });
+  };
+}
+
+/**
+ * This is a predicate function that takes a list of arguments and parameters,
+ * and it checks whether the given function should skip caching based on specific positional argument values.
+ *
+ * @param args - A list of arguments passed to a function
+ * @param params - An array of rules for skipping caching
+ * @returns A boolean indicating whether caching should be skipped for the provided arguments
+ *
+ * @example
+ *   skipParams: [{
+ *     index: "0",
+ *     value: "latest|pending"
  *   }]
  */
 const shouldSkipCachingForSingleParams = (args: unknown[], params: CacheSingleParam[] = []): boolean => {
   for (const item of params) {
     const values = item.value.split('|');
-    if (values.indexOf(args[item.index]) > -1) {
+    if (values.includes(args[item.index])) {
       return true;
     }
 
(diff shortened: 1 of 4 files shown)


packages/relay/src/lib/decorators/rpcParamLayoutConfig.decorator.ts
```diff
@@ -57,3 +57,38 @@ export function rpcParamLayoutConfig(layout: string | ParamTransformFn) {
...
+ * @param layout - Parameter layout specification
+ */
+export function rpcParamLayoutConfigStandard(layout: string | ParamTransformFn) {
+  return function (target: any, context: any /* ClassMethodDecoratorContext - requires TS5+ */): void {
+    if (context.kind !== 'method') {
+      throw new Error(`@rpcParamLayoutConfigStandard can only be applied to methods, received: ${context.kind}`);
+    }
+
+    // Attach parameter layout configuration to the method
+    (target as any)[RPC_PARAM_LAYOUT_KEY] = layout;
```
quiet-node, line 92: Shouldn't need to assert type to `target` again right it was already declared as any from the args list, correct?

packages/relay/src/lib/decorators/rpcParamLayoutConfig.decorator.ts
```diff
@@ -57,3 +57,38 @@ export function rpcParamLayoutConfig(layout: string | ParamTransformFn) {
...
+ * ```
+ *
+ * @param layout - Parameter layout specification
+ */
+export function rpcParamLayoutConfigStandard(layout: string | ParamTransformFn) {
+  return function (target: any, context: any /* ClassMethodDecoratorContext - requires TS5+ */): void {
+    if (context.kind !== 'method') {
+      throw new Error(`@rpcParamLayoutConfigStandard can only be applied to methods, received: ${context.kind}`);
+    }
+
```
quiet-node, line 90: Do we need this checker now or should it be in a different PR? Seems a little out of scope.

packages/relay/src/lib/decorators/rpcParamLayoutConfig.decorator.ts
```diff
@@ -28,7 +28,7 @@ export const RPC_LAYOUT = {
 };
 
 /**
- * Decorator for specifying the parameter layout of an RPC method which is different from the standard layout
+ * Legacy decorator for specifying the parameter layout of an RPC method which is different from the standard layout
```
quiet-node, line 31: Should we not explicitly say `legacy` cuz the project is still on TS 4?

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -27,23 +27,139 @@ interface CacheOptions {
...
+  return function (target: any, context: any /* ClassMethodDecoratorContext - requires TS5+ */): void {
+    if (context.kind !== 'method') {
+      throw new Error(`@cacheStandard can only be applied to methods, received: ${context.kind}`);
+    }
+
+    const methodName = String(context.name);
+
+    // Use addInitializer to replace the method implementation with cached version
+    context.addInitializer(function (this: any) {
+      const originalMethod = this[context.name as string | symbol];
```
quiet-node, line 112: Isn't this `target`? Per the [docs](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#decorators), isn't the first parameter the original method?

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -27,23 +27,139 @@ interface CacheOptions {
...
+            requestDetails,
+            options?.ttl ?? ConfigService.get('CACHE_TTL'),
+          );
+        }
+
+        return result;
+      };
+    });
+  };
+}
```
quiet-node, line 143: Nice I saw you said manually tested this did you make sure all the caching works as expected?

quiet-node (commented):
LG left some questions.

quiet-node:
Also since this require TS v5 should we create a follow up ticket to bump TS to v5?

packages/relay/src/lib/decorators/rpcParamLayoutConfig.decorator.ts
simzzz: I thought it would be a good addition but you're right it's out of scope. We can do it in a follow-up PR

packages/relay/src/lib/decorators/cache.decorator.ts
simzzz: yes, I did that

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -27,23 +27,75 @@ interface CacheOptions {
 }
 
 /**
- * Iterates through the provided 'params' array and checks if any argument in 'args' at the specified 'index'
- * matches one of the pipe-separated values in 'value'. If a match is found, caching should be skipped.
+ * TypeScript 5+ standard cache decorator.
+ * This is the clean, modern version for TypeScript 5+ without legacy compatibility.
  *
```
quiet-node, line 32: Probably we don't need to add this piece of information given that we fully migrated to TS v5 and it's now the new default

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -27,23 +27,75 @@ interface CacheOptions {
...
+ * @returns A boolean indicating whether caching should be skipped for the provided arguments
  *
  * @example
- *   [{
- *     index: '0',
- *     value: 'pending|safe'
+ *   skipParams: [{
+ *     index: "0",
+ *     value: "latest|pending"
  *   }]
```
quiet-node, line 93: Seems out of the scope of the PR. It seems like a functional update. Why such change and how comes upgrading to TS 5 requires this change?

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -27,23 +27,75 @@ interface CacheOptions {
...
  */
 const shouldSkipCachingForSingleParams = (args: unknown[], params: CacheSingleParam[] = []): boolean => {
   for (const item of params) {
     const values = item.value.split('|');
-    if (values.indexOf(args[item.index]) > -1) {
+    if (values.includes(args[item.index])) {
```
quiet-node, line 98: Seems out of the scope of the PR. Does the TS 5 migration necessitate this?

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -58,23 +110,20 @@ const shouldSkipCachingForSingleParams = (args: unknown[], params: CacheSinglePa
...
+ *     index: "0",
  *     fields: [{
- *       name: 'fromBlock', value: 'pending|safe'
- *     }, {
- *       name: 'toBlock', value: 'safe|finalized'
- *     }],
+ *       name: "blockTag",
+ *       value: "latest|pending"
+ *     }]
  *   }]
```
quiet-node, line 127: Same for these guys are they out of scope of indeed neccessary for the TS v5 migration? These updates could've been mentioned in the PR description for clearer indication.

packages/relay/src/lib/decorators/rpcMethod.decorator.ts
```diff
@@ -7,8 +7,8 @@
 export const RPC_METHOD_KEY = 'hedera-rpc-method';
 
 /**
- * Decorator that marks a class method as an RPC method.
- * When applied to a method, it marks that method as available for RPC invocation.
+ * TypeScript 5+ standard decorator that marks a class method as an RPC method.
+ * This is the clean, modern version for TypeScript 5+ without legacy compatibility.
```
quiet-node, line 11: Yeah, same here probably we can remove these comments as they don’t really add any value or context.

Instead, let’s restore the original comments that actually explain what this method does and how it’s used.

packages/relay/src/lib/decorators/rpcMethod.decorator.ts
```diff
@@ -20,12 +20,17 @@ export const RPC_METHOD_KEY = 'hedera-rpc-method';
...
-  return descriptor;
+export function rpcMethod(target: any, context: ClassMethodDecoratorContext): any {
+  context.addInitializer(function (this: any) {
+    const methodName = String(context.name);
+    if (this[methodName]) {
+      this[methodName][RPC_METHOD_KEY] = true;
+    }
+  });
+
+  return target;
```
quiet-node, line 35: From the doc, `addInitializer` appears to hook into the constructor of the target method. Do we actually need that in this case? I remember seeing a simpler approach used previously did that not work for some reason?

packages/relay/src/lib/decorators/rpcParamLayoutConfig.decorator.ts
```diff
@@ -28,22 +28,22 @@ export const RPC_LAYOUT = {
 };
 
 /**
- * Decorator for specifying the parameter layout of an RPC method which is different from the standard layout
+ * TypeScript 5+ standard decorator for specifying the parameter layout of an RPC method.
  *
```
quiet-node, line 32: Let's revert this

packages/relay/src/lib/decorators/rpcMethod.decorator.ts
```diff
@@ -20,12 +20,17 @@ export const RPC_METHOD_KEY = 'hedera-rpc-method';
...
-export function rpcMethod(_target: any, _propertyKey: string, descriptor: PropertyDescriptor) {
-  descriptor.value[RPC_METHOD_KEY] = true;
-  return descriptor;
+export function rpcMethod(target: any, context: ClassMethodDecoratorContext): any {
+  context.addInitializer(function (this: any) {
+    const methodName = String(context.name);
+    if (this[methodName]) {
+      this[methodName][RPC_METHOD_KEY] = true;
+    }
+  });
```
quiet-node, line 33: Curious I see it uses the method name in the `this` object to add the RPC_METHOD_KEY value, would it be able to use `target` original method? Or is this a more recommended approach?

packages/relay/src/lib/validators/index.ts
```diff
@@ -31,7 +31,8 @@ export interface IParamValidation {
 export const RPC_PARAM_VALIDATION_RULES_KEY = 'hedera-rpc-param-validation-rules';
 
 /**
- * Decorator that defines a schema for validating RPC method parameters
+ * TypeScript 5+ standard decorator that defines a schema for validating RPC method parameters.
+ * This is the clean, modern version for TypeScript 5+ without legacy compatibility.
  *
```
quiet-node, line 36: Let's revert this

quiet-node (commented):
Left some comments and questions

packages/relay/src/lib/decorators/cache.decorator.ts
acuarica: Unless strictly needed, I'd address this in a follow up PR. Including these functional changes in a migration like this will make the PR harder to review, and harder to track in the future.

packages/relay/src/lib/decorators/rpcMethod.decorator.ts
acuarica: From what I can tell from the docs https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html, `addInitializer` can be used when the method `"might be declared as a property initialized to an arrow function."`

```js
    greet = () => { ... };
```

This doesn't apply to us given decorated methods are declared as proper functions within a class.

acuarica, line 33: is it possible to write it like this instead?

```suggestion
  target[RPC_METHOD_KEY] = true;  
```

packages/relay/src/lib/decorators/cache.decorator.ts
simzzz: reverted

packages/relay/src/lib/decorators/cache.decorator.ts
simzzz: reverted

packages/relay/src/lib/decorators/rpcMethod.decorator.ts
simzzz: I didn't do it originally because the cache decorator used it, which replaced the original method and broke the other decorators. However your comment is correct and I can also remove it from there safely. It's removed from all decorators and all tests pass now

packages/relay/src/lib/validators/index.ts
```diff
@@ -49,10 +49,10 @@ export const RPC_PARAM_VALIDATION_RULES_KEY = 'hedera-rpc-param-validation-rules
  * @returns Method decorator function
  */
 export function rpcParamValidationRules(validationRules: Record<number, IParamValidation>) {
-  return function (_target: any, _propertyKey: string, descriptor: PropertyDescriptor) {
-    // Store validation rules directly on the function as a property
```
acuarica, line 53: better to keep this comment, right?

acuarica (approved):
nice stuff! excited to have TS 5 now

acuarica (approved):

quiet-node (commented):
Nice work! Thanks for opting back to the less overengineered approach! It makes it easier to maintain. Just one tiny comment I think we should resolve before merging.

packages/relay/src/lib/decorators/cache.decorator.ts
```diff
@@ -30,6 +30,56 @@ interface CacheOptions {
...
+ *   @property ttl - Optional time-to-live for the cache entry; falls back to global config if not provided.
+ *
+ * @returns A method decorator function that wraps the original method with caching logic.
+ *
+ * @example
+ *   @cache(CacheService, { skipParams: [...], skipNamesParams: [...], ttl: 300 })
+ */
+export function cache(cacheService: CacheService, options: CacheOptions = {}) {
+  return function (target: any, context: ClassMethodDecoratorContext) {
+    const methodName = String(context.name);
```
acuarica, line 51: This fixes https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3976.

konstantinabl (approved):

PR #3962, opened 2025-07-18
feat: gracefully handle HTML error responses from remote Mirror Node server

### LINKED ISSUE #3711: Intermittent Request Failures on Previewnet with Unknown Server Error Code 567
## Description
Intermittent request failures have been observed on the Previewnet environment (6%-9% of the requests). These requests fail with a status code of 567 (the fallback code for unknown server errors in MirrorNodeClient), indicating the error likely originates in the Relay service rather than the Mirror Node remote server.

## Error Symptoms
The error logs show vague and inconsistent messages:
- `Error invoking RPC: Unexpected 'S'`
- `Error invoking RPC: Unexpected '<'`
These errors:
- Occur in rapid succession during short timeframes
- Are exclusive to the Previewnet environment
- Cannot be reproduced in local development environments
- Do not contain sufficient context in logs for immediate diagnosis

## Root Cause
The error is caused by HTML content being returned instead of expected JSON responses. At the time of the issue, MN was returning 502 Server Error with the below content:
```html
<html><head>
<meta http-equiv="content-type" content="text/html;charset=utf-8">
<title>502 Server Error</title>
</head>
<body text=#000000 bgcolor=#ffffff>
<h1>Error: Server Error</h1>
<h2>The server encountered a temporary error and could not complete your request.<p>Please try again in 30 seconds.</h2>
</body></html>
```

The error messages `Unexpected 'S'` and `Unexpected '<'` occur when the MirrorNodeClient attempts to parse HTML with `JSONBigInt.parse()` in the Axios response interceptor:

```typescript
axiosRequestConfig['transformResponse'] = [
  (data) => {
    // if the data is not valid, just return it to stick to the current behaviour
    if (data) {
      return JSONBigInt.parse(data);
    }
    return data;
  },
];
```

This is a classic case of receiving HTML error pages (starting with `<!DOCTYPE html>` or `<html>`) when JSON is expected. In the Previewnet environment, one of the following is occurring:
- The Mirror Node service is returning HTML error pages during service degradation
- A proxy/WAF in the request path is intercepting and returning HTML error pages
- Network configurations specific to Previewnet are causing redirects to HTML status pages

## Suggested Fix
- Enhance error logging to capture the initial portion of failing responses to confirm HTML content
- Add logic for the parsing to handle the case when the response is an http  page instead of valid JSON object. If response is not a valid JSON object, skip parsing so the error can be thrown properly.

quiet-node:
### Description

This PR adds an HTML error response checker to the MirrorNodeClient class, enabling it to detect when a response is HTML rather than a valid JSON object. This prevents parsing errors in the transformResponse interceptor. If the response is identified as HTML, the interceptor returns the raw data, allowing the MirrorNodeClient to handle the response appropriately.

### Related issue(s)

Fixes #3711 

### Testing Guide

1. New unit test was added in mirrorNodeClient.spec.ts, under `should gracefully handle HTML error responses` unit
2. Run Relay unit test
3. Observe the test passing indicating the new logic works as expected.

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/clients/mirrorNodeClient.ts
+++ b/packages/relay/src/lib/clients/mirrorNodeClient.ts
@@ -378,19 +378,16 @@ export class MirrorNodeClient {
             (data) => {
               // if the data is not valid, just return it to stick to the current behaviour
               if (data) {
-                if (
-                  typeof data === 'string' &&
-                  (data.trim().startsWith('<html') ||
-                    data.trim().startsWith('<!DOCTYPE html') ||
-                    data.trim().startsWith('<'))
-                ) {
-                  // Most likely HTML error response from the remote server.
-                  // Return raw data so response can be processed properly by subsequent operations.
-                  return data;
+                try {
+                  return JSONBigInt.parse(data);
+                } catch (error) {
+                  this.logger.warn(
+                    `${requestDetails.formattedRequestId} Failed to parse response data from Mirror Node: ${error}`,
+                  );
                 }
-                return JSONBigInt.parse(data);
               }
 
+              // Return raw data so response can be processed properly by subsequent operations.
               return data;
             },
           ];
--- a/packages/relay/tests/lib/mirrorNodeClient.spec.ts
+++ b/packages/relay/tests/lib/mirrorNodeClient.spec.ts
@@ -217,6 +217,14 @@ describe('MirrorNodeClient', async function () {
         .to.eventually.be.rejectedWith('Request failed with status code 502')
         .and.have.property('statusCode', 502);
     });
+
+    it('should gracefully handle empty error responses', async () => {
+      // Simulate Mirror Node returning HTML error page
+      mock.onGet('accounts').reply(503, ``);
+      await expect(mirrorNodeInstance.get('accounts', 'accounts', requestDetails))
+        .to.eventually.be.rejectedWith('Request failed with status code 503')
+        .and.have.property('statusCode', 503);
+    });
   });
 
   it('Can extract the account number out of an account pagination next link url', async () => {


packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -378,6 +378,16 @@ export class MirrorNodeClient {
             (data) => {
               // if the data is not valid, just return it to stick to the current behaviour
               if (data) {
+                if (
+                  typeof data === 'string' &&
+                  (data.trim().startsWith('<html') ||
+                    data.trim().startsWith('<!DOCTYPE html') ||
+                    data.trim().startsWith('<'))
```
acuarica, line 385: Wouldn't be more resilient to just put `JSONBigInt.parse(data)` in a `try/catch` block?

This comparison is case sensitive, so for example `<HTML` would produce an error as well.

Note: this repeated `data.trim()` might cause performance problems. Of course, don't have hard data to back this, but this bit here is a hot spot, so having such comparison might cause problems.

packages/relay/tests/lib/mirrorNodeClient.spec.ts
```diff
@@ -204,6 +204,19 @@ describe('MirrorNodeClient', async function () {
...
       });
     }
+
+    it('should gracefully handle HTML error responses', async () => {
+      // Simulate Mirror Node returning HTML error page
+      mock
+        .onGet('accounts')
+        .reply(
+          502,
+          `<!DOCTYPE html><html><head><title>502 Server Error</title></head><body>Error: Server Error</body></html>`,
```
acuarica, line 214: what would be happen if it's empty string `""` for example? should we care about this case?

acuarica (changes requested):
Thanks for taking a look, this is a nice improvement.

packages/relay/src/lib/clients/mirrorNodeClient.ts
quiet-node: Hey yeah I did try using try/catch but somehow I was clinging to the idea of throwing the parsing error which would discard the original error response from the Mirror Node. That’s why I opted to detect the issue before attempting to parse.

But looking back at the logic, the interceptor doesn’t need to throw a parsing error but just return the raw response so it can be handled later. Just updated it to follow that approach.

Thanks for raising this! It definitely makes more sense and is more robust than what I had before.

packages/relay/tests/lib/mirrorNodeClient.spec.ts
quiet-node: Yessir good point! Added a new unit test to cover that case

acuarica (approved):

PR #3979, opened 2025-07-23
feat: simplify `_CONFIG` object structur

### LINKED ISSUE #3895: Simplify `_CONFIG` object structure
The [`_CONFIG`](https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/3684f3cf3be8e4a521b57c7277f0e816c0f86030/packages/config-service/src/services/globalConfig.ts#L81) object that holds the environment variables configuration requires each entry to have an `envName` property. This property is actually equal to its outer key, for example

```js
  BATCH_REQUESTS_ENABLED: {
    envName: 'BATCH_REQUESTS_ENABLED',
    type: 'boolean',
    required: false,
    defaultValue: true,
  },
```

This is redundant, and may confuse newcomers wanting to contribute to the Relay.

Moreover, it seems this property is not used anywhere besides the `loggerService`. However, this can be easily replaced by either using the key directly (and making it type safe), or by including an optional field that declares if the entry should be masked or not. 

Once `envName` had been replaced in `loggerService`, we should remove it from `_CONFIG` object.

Additionally, in the same `loggerService`, the `KNOWN_SECRET_PREFIXES` field is not used and it should be removed as well.

natanasow:
### Description

The [`_CONFIG`](https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/3684f3cf3be8e4a521b57c7277f0e816c0f86030/packages/config-service/src/services/globalConfig.ts#L81) object that holds the environment variables configuration requires each entry to have an `envName` property. This property is actually equal to its outer key, for example

```js
  BATCH_REQUESTS_ENABLED: {
    envName: 'BATCH_REQUESTS_ENABLED',
    type: 'boolean',
    required: false,
    defaultValue: true,
  },
```

This is redundant, and may confuse newcomers wanting to contribute to the Relay.

Moreover, it seems this property is not used anywhere besides the `loggerService`. However, this can be easily replaced by either using the key directly (and making it type safe), or by including an optional field that declares if the entry should be masked or not. 

Once `envName` had been replaced in `loggerService`, we should remove it from `_CONFIG` object.

Additionally, in the same `loggerService`, the `KNOWN_SECRET_PREFIXES` field is not used and it should be removed as well.

### Related issue(s)

Fixes #3895 

### Testing Guide

1.
2.
3.

N/A

N/A

### Checklist


konstantinabl (dismissed):

packages/config-service/tests/src/services/validationService.spec.ts
```diff
@@ -121,79 +121,74 @@ describe('ValidationService tests', async function () {
...
-      expect(castedEnvs).to.not.haveOwnProperty(GlobalConfig.ENTRIES.GH_ACCESS_TOKEN.envName);
-      expect(castedEnvs[GlobalConfig.ENTRIES.GH_ACCESS_TOKEN.envName]).to.be.undefined;
+      expect(castedEnvs).to.not.haveOwnProperty('GH_ACCESS_TOKEN');
+      expect(castedEnvs['GH_ACCESS_TOKEN']).to.be.undefined;
     });
 
     it('should to cast string type', async () => {
       const castedEnvs = ValidationService.typeCasting({
-        [GlobalConfig.ENTRIES.CHAIN_ID.envName]: '0x160c',
+        ['CHAIN_ID']: '0x160c',
```
acuarica, line 136: no need to use bracket notation now

```suggestion
        CHAIN_ID: '0x160c',
```

acuarica (commented):
Good to have this.

packages/config-service/tests/src/services/validationService.spec.ts
natanasow: Fixed.

acuarica (approved):

packages/config-service/src/services/loggerService.ts
```diff
@@ -27,7 +15,7 @@ export class LoggerService {
    * @param envValue
    */
   static maskUpEnv(envName: string, envValue: string | undefined): string {
-    const isSensitiveField: boolean = this.SENSITIVE_FIELDS.indexOf(envName) > -1;
+    const isSensitiveField: boolean = (this.SENSITIVE_FIELDS as string[]).indexOf(envName) > -1;
```
quiet-node, line 18: So this.SENSITIVE_FIELDS is typed as ConfigKey but is now inferred as string[] so it'd work with envName: string. Should we instead explicitly declare envName as ConfigKey to make it less awkward?

quiet-node (commented):
LG just one item

packages/config-service/src/services/loggerService.ts
quiet-node: Ah yeah I see right it makes sense and also because the `envName` is indeed from the process.env object so type string makes more sense. Yeah good point thanks!

quiet-node (approved):
LGTM! Great work!

PR #3984, opened 2025-07-25
chore: use Node's native `fetch` in PR check workflow

### LINKED ISSUE #3924: Removing axios from the PR check
The [`PR Label and Milestone Check`](https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/0dc6105329eb44a3621e5ff60182d0db915d1753/.github/workflows/pr-label-milestone-check.yml) workflow help us to enforce certain rules for newly created PRs, _e.g._, that it has an associated milestone.

One of these rules is that every PR **must** have an associated issue. This rule works fine for members of the [hiero-json-rpc-relay-maintainers](https://github.com/orgs/hiero-ledger/teams/hiero-json-rpc-relay-maintainers) team (associated issues are part of our internal project board). However, it does not play well when other contributors submit a PR because a dummy issue needs to be created just to satisfy this constrain. See for example

- https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3865
- https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3922

Ideally, we should make this workflow more lenient by allowing non-members of hiero-json-rpc-relay-maintainers to submit PRs without an associated issue. If this is not possible, we should evaluate to drop this rule altogether.

Additionally, there is room for improvement in the [`check_pr.js`](https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/0dc6105329eb44a3621e5ff60182d0db915d1753/.github/scripts/check-pr.js) script. It uses `axios` which takes ~2min to install in every run.

Image

This is unnecessary given that we can use Node's built-in [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) API instead of `axios`.

acuarica:
### Description

This PR uses Node's native `fetch` in the PR check workflow script and removes the `axios` dependency. It includes other improvements such as output with colors to highlight log messages, and collect all errors at once for better feedback.

In the workflow itself, the `axios` dependency is no longer installed. Additionally, the workflow uses Node 22 given it's the active version.

### Related issue(s)

Fixes #3924.

N/A

N/A

### Checklist

### DIFF
--- a/.github/scripts/check-pr.js
+++ b/.github/scripts/check-pr.js
@@ -1,56 +1,46 @@
-const axios = require('axios');
 
-const githubToken = process.env.GITHUB_TOKEN;
-const { GITHUB_REPOSITORY, GITHUB_PR_NUMBER } = process.env;
+const { GITHUB_TOKEN, GITHUB_REPOSITORY, GITHUB_PR_NUMBER } = process.env;
 
 const [owner, repo] = GITHUB_REPOSITORY.split('/');
 
+/**
+ * @param {string} endpoint 
+ * @returns 
+ */
+async function _fetch(endpoint) {
+  return await fetch('https://api.github.com/repos/' + endpoint, {
+    // headers: {
+    //   Authorization: `token ${githubToken}`,
+    // },
+  });
+}
+
+/**
+ * 
+ * @param {string} prNumber 
+ * @returns 
+ */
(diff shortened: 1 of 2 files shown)


quiet-node (approved):
LGTM there seem to be a lint error tho

andrewb1269hg (approved):

Issue #3707, opened 2025-04-17
refactor: create new eventEmitter instances in classes instead of passing the same instance


konstantinabl:
Currently, we pass around one instance of the eventEmitter in classes that need it. @acuarica suggested we shouldn't do this, but instead create a new instance in the different classes.

Discussion - > https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3646#discussion_r2039745626

PR #4081, opened 2025-07-31
refactor: improve execution event handling

### LINKED ISSUE #3706: refactor: add event types
This topic arised from the following discussion https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3646#discussion_r2039743506

**Problem**
Currently, the events we emit, are not typed, so type safety is not ensured and this may bugs and harder maintainability, since when you change an event you cannot just change the type, but need to hunt down all such events in the repo and edit them

**Solution**
Introduce a typed event API so that emit/on calls are checked by TypeScript. Two approaches are possible:

1. Use a third party package called typed-emitter, which allows to wrap an eventemitter and provide a type of the event

```typescript
interface TransactionServiceEvents {
     [constants.EVENTS.ETH_EXECUTION]: {
       method: typeof constants.ETH_SEND_RAW_TRANSACTION
       requestDetails: RequestDetails
     }
     // …other events
   }

class TransactionService {
       private eventEmitter: TypedEmitter<TransactionServiceEvents>

       constructor() {
         this.eventEmitter = new EventEmitter() as TypedEmitter<TransactionServiceEvents>
       }
}
```

### Pros

- Already used by the community (npm community) and tested
- No extra code for us to maintain

### Cons

- Extra dependency for our repo
- If the package is deprecated it would affect us

2. Use our own EventEmitter wrapper
```typescript
class TypedEmitter<E extends Record<string, any>> extends EventEmitter {
       emit<K extends keyof E>(event: K, payload: E[K]): boolean {
         return super.emit(event as string, payload)
       }
       on<K extends keyof E>(event: K, listener: (payload: E[K]) => void) {
         return super.on(event as string, listener)
       }
     }
```
### Pros

- No third-party dependencies—pure in-house code 
- Customize our behaviour

### Cons

- Maintain and write it ourselves (possibly more resources towards something existing)
- Need to keep it in sync with eventEmitter API
- Test it ourselves

natanasow:
### Description

Currently, the events we emit, are not typed, so type safety is not ensured, and this may lead to bugs and harder maintainability, since when you change an event, you cannot just change the type, but need to hunt down all such events in the repo and edit them.

### Solution
Introduce a typed event API so that emit/on calls are checked by TypeScript. Use the native event types introduced in `@types/node` >= 20.14.x.

### Related issue(s)

Fixes #3706 #3707
 
### Testing Guide

1.
2.
3.

N/A

N/A

### Checklist


packages/relay/src/lib/eth.ts
```diff
@@ -80,7 +82,7 @@ export class EthImpl implements Eth {
    * Event emitter for publishing and subscribing to events.
    * @private
    */
-  private readonly eventEmitter: EventEmitter;
+  private readonly eventEmitter: TypedEmitter<TypedEvents>;
```
quiet-node, line 85: LEt's remove unused EventEmitter import

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -13,6 +13,7 @@ import {
   IExecuteTransactionEventPayload,
   ITransactionRecordMetric,
   RequestDetails,
+  TypedEvents,
```
quiet-node, line 16: IEthExecutionEventPayload could be removed

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -108,16 +109,56 @@ export default class MetricService {
...
+          transactionId,
+          txConstructorName,
+          cost,
+          gasUsed,
+          status,
+          requestDetails,
+          originalCallerAddress,
+        });
+      },
+    );
```
quiet-node, line 159: hmmm this doesn't seem too clean with what we had before. Can we do soemthing with the addExpenseAndCaptureMetrics or captureTransactionMetrics or something so it'd be cleaner?

quiet-node (commented):
Good work left some qs and comments around.

package.json
```diff
@@ -100,7 +100,8 @@
     "pino-pretty": "^7.6.1",
     "pnpm": "^10.0.0",
     "prom-client": "^14.0.1",
-    "redis": "^4.7.0"
+    "redis": "^4.7.0",
+    "typed-emitter": "^2.1.0"
```
acuarica, line 104: It looks like this package `typed-emitter` is quite old (and probably unmaintained).

Maybe it's better to use Node's native typings. We are using an old version of `@types/node` (`17.x.x`) which doesn't include support for typed events. However, a newer version https://www.npmjs.com/package/@types/node/v/20.14.8?activeTab=code does include typed events (see `events.d.ts` for event definitions).

packages/relay/src/lib/constants.ts
```diff
@@ -224,9 +224,9 @@ export default {
   },
 
   EVENTS: {
```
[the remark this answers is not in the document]
acuarica: Yeah, ideally this should be removed and used the keys of the event map definition instead.

packages/relay/src/typedEmitter.ts
```diff
@@ -0,0 +1,17 @@
...
+ * A strongly-typed event emitter based on the native node Event Emitter
+ *
+ * @extends EventEmitter
+ * @template TypedEvents - Defines event names and payload structures.
+ */
+export class TypedEmitter extends EventEmitter<TypedEvents> {
+  constructor() {
+    super();
+  }
+}
```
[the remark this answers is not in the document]
natanasow: Fixed.

packages/relay/src/lib/types/events.ts
```diff
@@ -23,7 +23,25 @@ export interface IExecuteQueryEventPayload {
...
+  EXECUTE_QUERY: [
+    executionMode: string,
+    transactionId: string,
+    txConstructorName: string,
+    cost: number,
+    gasUsed: number,
+    status: string,
+    requestDetails: RequestDetails,
+    originalCallerAddress: string | undefined,
+  ];
```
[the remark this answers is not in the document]
natanasow: Fixed.

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -338,10 +338,7 @@ export class TransactionService implements ITransactionService {
    * @param requestDetails The request details for logging and tracking
    */
   private emitEthExecutionEvent(requestDetails: RequestDetails): void {
-    this.eventEmitter.emit(constants.EVENTS.ETH_EXECUTION, {
-      method: constants.ETH_SEND_RAW_TRANSACTION,
-      requestDetails: requestDetails,
-    });
+    this.eventEmitter.emit(constants.EVENTS.ETH_EXECUTION, constants.ETH_SEND_RAW_TRANSACTION, requestDetails);
```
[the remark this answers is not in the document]
natanasow: Fixed.

packages/relay/src/lib/types/events.ts
```diff
@@ -23,7 +20,20 @@ export interface IExecuteQueryEventPayload {
...
+  originalCallerAddress: string;
 }
+
+export interface TypedEvents {
+  [constants.EVENTS.ETH_EXECUTION]: [IEthExecutionEventPayload];
+  [constants.EVENTS.EXECUTE_QUERY]: [IExecuteQueryEventPayload];
+  [constants.EVENTS.EXECUTE_TRANSACTION]: [IExecuteTransactionEventPayload];
+}
+
+export type CustomEventEmitter = EventEmitter<TypedEvents>;
```
quiet-node, line 39: Looks like this throw the warning below

```
Type 'EventEmitter' is not generic.ts(2315)
(alias) class EventEmitter
(alias) interface EventEmitter
(alias) namespace EventEmitter
import EventEmitter
The EventEmitter class is defined and exposed by the events module:

const EventEmitter = require('events');
All EventEmitters emit the event 'newListener' when new listeners are added and 'removeListener' when existing listeners are removed.

It supports the following option:

@since — v0.1.26
```

Strange! EventEmitter is not genetic but I saw it accepted type param in the previous version, right?

packages/relay/src/lib/relay.ts
```diff
@@ -136,7 +136,7 @@ export class Relay {
     const chainId = ConfigService.get('CHAIN_ID');
     const duration = constants.HBAR_RATE_LIMIT_DURATION;
 
-    this.eventEmitter = new EventEmitter();
+    this.eventEmitter = new EventEmitter<TypedEvents>();
```
quiet-node, line 139: This also gives

```
Expected 0 type arguments, but got 1.ts(2558)
(alias) interface TypedEvents
import TypedEvents
```

packages/relay/src/lib/types/events.ts
natanasow: Upgrading to node v22.x and running `npm ci` should resolve this issue as @acuarica suggested.

packages/relay/src/lib/relay.ts
natanasow: Same as above.

packages/relay/src/lib/constants.ts
```diff
@@ -273,4 +273,4 @@ export default {
   ETH_GET_TRANSACTION_RECEIPT: 'eth_GetTransactionReceipt',
   ETH_SEND_RAW_TRANSACTION: 'eth_sendRawTransaction',
   NON_CACHABLE_BLOCK_PARAMS: 'latest|pending|finalized|safe',
-};
+} as const;
```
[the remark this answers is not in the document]
natanasow: Fixed.

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -187,8 +187,8 @@ export class MirrorNodeClient {
     });
 
     if (useCacheableDnsLookup) {
-      betterLookupInstall(httpAgent);
-      betterLookupInstall(httpsAgent);
+      betterLookupInstall(httpAgent as any);
+      betterLookupInstall(httpsAgent as any);
```
acuarica, line 191: why do we need these `as any`? AFAICS if we remove them there are no compilation or linter errors, right?

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -339,13 +328,15 @@ export class SDKClient {
       return transactionResponse;
     } finally {
       if (transactionId?.length) {
-        this.eventEmitter.emit(constants.EVENTS.EXECUTE_TRANSACTION, {
+        this.eventEmitter.emit('execute_transaction', {
           transactionId,
-          requestDetails,
+          callerName,
```
acuarica, line 333: We removed `callerName` and `interactingEntity` from this metric, we should remove them in `IExecuteTransactionEventPayload` interface instead.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -13,28 +13,21 @@ import {
...
   Query,
   Status,
   Transaction,
-  TransactionId,
   TransactionRecord,
   TransactionRecordQuery,
   TransactionResponse,
 } from '@hashgraph/sdk';
-import { EventEmitter } from 'events';
+import EventEmitter from 'events';
```
acuarica, line 23: nit: use `{ EventEmitter }` for consistency

packages/relay/src/lib/types/events.ts
```diff
@@ -1,29 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { RequestDetails } from './RequestDetails';
+import { IEthExecutionEventPayload, IExecuteQueryEventPayload, IExecuteTransactionEventPayload } from './sdkClient';
```
acuarica, line 3: If we had two different event types (one for the SDK Client and one for EthImp) then these event definitions can be placed there. But given we use only one event definition, `TypedEvents`, I'd inline this type definitions here.

Also because it can be confusing to declare `IEthExecutionEventPayload` in a file sdkClient but being emitted by `EthImpl` class.

packages/relay/src/lib/eth.ts
```diff
@@ -114,16 +115,15 @@ export class EthImpl implements Eth {
    * @param {MirrorNodeClient} mirrorNodeClient - Client for querying the Hedera mirror node.
    * @param {Logger} logger - Logger instance for logging system messages.
    * @param {string} chain - The chain identifier for the current blockchain environment.
-   * @param {Registry} registry - Registry instance for registering metrics.
    * @param {CacheService} cacheService - Service for managing cached data.
+   * @param eventEmitter
```
quiet-node, line 119: looks like it's a left-over

packages/relay/src/lib/types/events.ts
quiet-node: Agreed. Quite confusing seeing the new sdkClient.ts file in the types folder.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -89,16 +73,21 @@ export class SDKClient {
    * @param {Client} clientMain - The primary Hedera client instance used for executing transactions and queries.
    * @param {Logger} logger - The logger instance for logging information, warnings, and errors.
    * @param {EventEmitter} eventEmitter - The eventEmitter used for emitting and handling events within the class.
+   * @param hbarLimitService
```
quiet-node, line 76: Let's add some explanation for consistency

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -89,16 +73,21 @@ export class SDKClient {
...
    */
-  constructor(clientMain: Client, logger: Logger, eventEmitter: EventEmitter, hbarLimitService: HbarLimitService) {
+  constructor(
+    clientMain: Client,
+    logger: Logger,
+    readonly eventEmitter: EventEmitter<TypedEvents>,
```
quiet-node, line 81: Why is eventEmitter moved to param property instead of explicit class property? I know this technically works the same but it seems to break consistency. If we're doing this should we also move logger and hbarLimitService to param property as well? If afraid of it being out of scope, I think we should revert and use explicit property to remain consistency.

packages/relay/src/lib/types/events.ts
quiet-node: Indeed! Thanks much guys!

packages/relay/src/lib/clients/mirrorNodeClient.ts
natanasow: Removing `as any` results in a compilation error:
```
      src/lib/clients/mirrorNodeClient.ts(190,27): error TS2345: Argument of type 'Agent' is not assignable to parameter of type 'Agent & { createConnection?: ((options: any, callback: Function) => Socket) | undefined; }'.
        Type 'Agent' is not assignable to type '{ createConnection?: ((options: any, callback: Function) => Socket) | undefined; }'.
          Types of property 'createConnection' are incompatible.
            Type '(options: ClientRequestArgs, callback?: ((err: Error | null, stream: Duplex) => void) | undefined) => Duplex' is not assignable to type '(options: any, callback: Function) => Socket'.
              Types of parameters 'callback' and 'callback' are incompatible.
                Type 'Function' is not assignable to type '(err: Error | null, stream: Duplex) => void'.
                  Type 'Function' provides no match for the signature '(err: Error | null, stream: Duplex): void'.
```

packages/relay/src/lib/clients/sdkClient.ts
natanasow: Fixed.

packages/relay/src/lib/clients/sdkClient.ts
natanasow: Added.

acuarica (commented):
Might be better if you can rebase/merge with `main` to avoid errors when we merge and for an easier review.

packages/relay/tests/lib/sdkClient.spec.ts
```diff
@@ -96,6 +97,7 @@ describe('SdkClient', async function () {
       duration,
     );
 
+    // @ts-ignore
```
acuarica, line 100: if this is needed because error is `EventEmitter` is not generic, we might need to update also server's and ws-server's `@types/node` dependency.

acuarica (approved):

packages/relay/src/lib/clients/sdkClient.ts
quiet-node: Yes agree. We should turn logger and hbarLimitService into param property as well

quiet-node (approved):
LGTM. One item about the constructor of the SDKClient but it's not a blocking issue.

PR #4169, opened 2025-08-06
feat: implement traffic-weighted stress test framework and scenario improvements

quiet-node:
### Description

This PR introduces a new k6-based stress test framework for the Hedera JSON RPC Relay, simulating realistic traffic patterns using 90 days of production data. Key changes include:

- Added stress test scenarios and traffic weights based on real endpoint usage.
- Refactored test logic to support both sequential and concurrent (stress) execution.
- Improved test preparation scripts with configurable wallet balances and artifact generation.
- Updated documentation with setup, environment variables, and recommended settings for stress testing.
- Bug fixes and reliability improvements for several test scenarios.

### Related issue(s)

Fixes https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3989

### Testing Guide

Follow the instruction in the new `Stress Test` section in k6/README.md for a comprehensive test guide

N/A

N/A

### Checklist


.eslintrc.js
```diff
@@ -10,6 +10,9 @@ module.exports = {
         "plugin:@typescript-eslint/recommended",
         "prettier",
     ],
+    "globals": {
+        "__ENV": "readonly",
+    },
```
quiet-node, line 15: fix lint errors

k6/src/lib/traffic-weights.js
```diff
@@ -0,0 +1,83 @@
...
+ * @param {number} totalVUs - Total number of VUs to distribute
+ * @returns {Object} VU allocation per endpoint
+ */
+export function calculateVUAllocation(totalVUs = 10) {
+  const allocation = {};
+
+  for (const [endpoint, weight] of Object.entries(trafficWeights)) {
+    // Simple proportional allocation: VUs = totalVUs * percentage
+    const vus = Math.round(weight * totalVUs);
+    // Ensure minimum 1 VU only if the calculated value would be 0
```
simzzz, line 75: If we have 1VU per endpoint wouldn't we always exceed the DEFAULT_VUS?

k6/src/lib/common.js
```diff
@@ -265,13 +283,12 @@ function TestScenarioBuilder() {
   this._testDuration = undefined;
   this._maxDuration = undefined;
 
-  this.build = function () {
-    const that = this;
```
konstantinabl, line 269: Just curious, any idea why we had this?

k6/src/lib/common.js
```diff
@@ -1,10 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
+import { randomIntBetween } from 'https://jslib.k6.io/k6-utils/1.4.0/index.js';
```
natanasow, line 3: q: Do we need to use and load an external lib for a random number?

natanasow (approved):
LGTM

k6/src/lib/common.js
quiet-node: Yeah, good question. I believe we don’t have to. I’m just leveraging it since it’s convenient, already used throughout the project, and remain consistency. We can consider removing it in a different PR just not a prio at the moment.

k6/src/lib/common.js
quiet-node: const that = this is typically used to preseve the original this context as it could be different depends on the calling site or nested function or callback etc. It could happen with regular functions. However, when switch to an arrow function, the this context inherits the this value from the enclosing class scope, which means the this in this arrow class will always refer to this in TestScenarioBuilder(). Eslint was complaining about it that's why.

quiet-node: Resolving this for now to proceed with the merge. Happy to discuss offline or follow up in future PRs.

PR #4171, opened 2025-08-07
chore: extend Relay's `JsonRpcError` from `Error`

### LINKED ISSUE #3617: Properly extend JsonRpcError from Error class while maintaining JSON-RPC 2.0 specification compliance
## Description:

**Current Behavior:**
When `JsonRpcError` doesn't extend `Error`:
- Error objects have the correct serialized format for JSON-RPC responses
- Stack traces are not captured, making debugging harder
- `error instanceof Error` checks fail
- `this.name` is undefined, which accidentally passes schema validation

When `JsonRpcError` extends `Error`:
- We get proper stack traces and type checking
- The `message` property becomes non-enumerable and is lost during JSON serialization
- `this.name` is automatically set to "Error", causing schema validation to fail as the OpenRPC spec expects "Method not found"

**Proposed Solution:**
1. Modify `JsonRpcError` to properly extend `Error` with:
   - Explicit prototype setting (`Object.setPrototypeOf(this, JsonRpcError.prototype)`)
   - Setting `this.name = 'JsonRpcError'` for better error identification
   - Implementing `toJSON()` to ensure all required fields are serialized properly

2. Update the OpenRPC schema to:
   - Remove the `name` field pattern requirement from `unsupportedError` schema, OR
   - Update it to match our actual implementation

**Technical Details:**
- The issue stems from how JavaScript handles inheritance from built-in classes like `Error`
- Properties inherited from `Error` like `message` are non-enumerable by default
- `JSON.stringify()` only includes enumerable properties
- OpenRPC test schema validations are failing due to the `name` property mismatch

**Benefits:**
- Better error handling and debugging with proper stack traces
- Consistent error objects that follow JavaScript best practices
- Maintained compliance with JSON-RPC 2.0 specification
- Better type checking with `instanceof`

**Additional Considerations:**
- We should review other error classes in the codebase for similar issues
- May need to adjust server-side code in `koaJsonRpc/index.ts` that handles error serialization
- Consider adding tests to verify proper error serialization and schema validation

acuarica:
### Related issue(s)

Fixes #3617.

N/A

N/A

### Checklist

### DIFF
--- a/packages/server/src/koaJsonRpc/lib/RpcError.ts
+++ b/packages/server/src/koaJsonRpc/lib/RpcError.ts
@@ -1,77 +1,28 @@
 // SPDX-License-Identifier: Apache-2.0
 
-export class JsonRpcError extends Error {
-  code: any;
-  data: any;
-  constructor(message, code, data) {
-    super();
-
-    this.message = message;
-    this.code = code;
-
-    if (typeof data !== 'undefined') {
-      this.data = data;
-    }
-  }
+export interface IJsonRpcError {
+  readonly code: number;
+  readonly message: string;
+  readonly data?: unknown;
 }
 
-export class ParseError extends JsonRpcError {
-  constructor() {
-    super('Parse error', -32700, undefined);
-  }
-}
+/**
+ * Standard JSON-RPC 2.0 error codes and messages.
+ *
+ * See _5.1 Error object_ https://www.jsonrpc.org/specification#error_object.
+ */
+export const spec = {
+  ParseError: { code: -32700, message: 'Parse error' },
+  InvalidRequest: { code: -32600, message: 'Invalid Request' },
+  MethodNotFound: (methodName: string) => ({ code: -32601, message: `Method ${methodName} not found` }),
 
-export class InvalidRequest extends JsonRpcError {
-  constructor() {
-    super('Invalid Request', -32600, undefined);
-  }
-}
-
-export class MethodNotFound extends JsonRpcError {
-  constructor(methodName: string) {
-    super(`Method ${methodName} not found`, -32601, undefined);
-  }
-}
-
-export class InvalidParams extends JsonRpcError {
-  constructor() {
-    super('Invalid params', -32602, undefined);
-  }
-}
-
-export class InternalError extends JsonRpcError {
   /**
    * @param err The error object that caused this `InternalError`.
    */
-  constructor(err: unknown) {
-    const message = err && typeof err === 'object' && 'message' in err ? err.message : 'Internal error';
-    super(message, -32603, undefined);
-  }
-}
+  InternalError: (err: unknown) => ({
+    code: -32603,
+    message: err && typeof err === 'object' && 'message' in err ? String(err.message) : 'Internal error',
+  }),
 
-export class Unauthorized extends JsonRpcError {
-  constructor() {
-    super('Unauthorized', -32604, undefined);
-  }
-}
-
-export class ServerError extends JsonRpcError {
-  constructor(code) {
-    if (code < -32099 || code > -32000) {
-      throw new Error('Invalid error code');
-    }
-    super('Server error', code, undefined);
-  }
-}
-
-export class IPRateLimitExceeded extends JsonRpcError {
-  constructor(methodName) {
-    super(`IP Rate limit exceeded on ${methodName}`, -32605, undefined);
-  }
-}
-
-export class HBARRateLimitExceeded extends JsonRpcError {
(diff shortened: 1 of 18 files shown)


packages/server/src/koaJsonRpc/lib/RpcError.ts
```diff
@@ -1,77 +1,28 @@
 // SPDX-License-Identifier: Apache-2.0
 
-export class JsonRpcError extends Error {
-  code: any;
-  data: any;
-  constructor(message, code, data) {
-    super();
+export interface IJsonRpcError {
```
quiet-node, line 3: Since this is not implemented by any class should we consider dropping the `I` prefix?

packages/server/src/koaJsonRpc/lib/RpcResponse.ts
```diff
@@ -1,8 +1,18 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { JsonRpcError } from '@hashgraph/json-rpc-relay';
+import { type IJsonRpcError } from './RpcError';
 
-import { IJsonRpcResponse } from './IJsonRpcResponse';
+export type IJsonRpcResponse<Result = unknown> = {
```
quiet-node, line 5: Since this is not implemented by any class should we consider dropping the I prefix?

quiet-node (approved):
LGTM! Great work! Looks so much cleaner! Just two nits

packages/server/src/koaJsonRpc/lib/RpcError.ts
acuarica: We use the `I` prefix to denote the shape of objects, regardless of having any concrete implementation. For example, the `ICacheClient` interface is used in the more "traditional" way (being implemented by two classes). But on the other hand, we have many `I`-prefixed interfaces that denote the shape of objects. See  for example `IJsonRpcRequest` or the interfaces in `types/mirrorNode.ts`.

I believe this use of the `I` prefix is consistent with the above.

packages/server/src/koaJsonRpc/lib/RpcResponse.ts
acuarica: The use of `type` here (as opposed to `interface`) is just a technicality. This is because interfaces cannot represent unions. However, the use of `I` prefix is still consistent with https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4171#discussion_r2260639955.

simzzz (dismissed):
LGTM, left just one nit

packages/server/src/koaJsonRpc/lib/httpErrorMapper.ts
```diff
@@ -54,7 +52,7 @@ export function translateRpcErrorToHttpStatus(jsonRpcError: JsonRpcError | JsonR
   // Handle Mirror Node errors (-32020)
   // Note: -32020 corresponds to predefined.MIRROR_NODE_UPSTREAM_FAILURE,
   // where `jsonRpcError.data` represents the actual HTTP status code returned from the Mirror Node upstream server.
-  if (jsonRpcError.code === -32020 && jsonRpcError.data) {
+  if (jsonRpcError.code === -32020 && typeof jsonRpcError.data === 'number') {
```
natanasow, line 55: q: What else besides HTTP status code can be the `.data` field? Can we rename it to `.statusCode`?

natanasow (approved):
LGTM, one question.

simzzz (approved):

packages/server/src/koaJsonRpc/lib/httpErrorMapper.ts
acuarica: At the moment, nothing. However, the idea is that this `.data` can be a generic field to include any additional information.

More practically though, the issue is that `JsonRpcError` class in the Relay needs to match the type `IJsonRpcError` interface in server. Otherwise the following `jsonRespError` will fail

https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/539ed01672908706422cd83919957eb14fabe785/packages/server/src/koaJsonRpc/index.ts#L181-L187

If we rename it, we need to do some conversion just for this case.

packages/server/src/koaJsonRpc/lib/RpcResponse.ts
acuarica: Same as https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4171#discussion_r2262834237.

konstantinabl (approved):

PR #4196, opened 2025-08-13
feat: moves formatContractResult from CommonServices to formatters

### LINKED ISSUE #3747: Move formatContractResult back to formatters
This isssue is a followup task after the refactor
https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3720#discussion_r2060745626 

**Problem** 

Currently, the formatContractResult is in the CommonService. However, it makes sense for it to be moved to formatters.ts , where we have our methods related to formatting

simzzz:
### Description

Currently, the formatContractResult is in the CommonService. However, it makes sense for it to be moved to formatters.ts, where we have our methods related to formatting. This PR does that

### Related issue(s)

Fixes #3747 

### Testing Guide

1. Verify formatContractResult is in formatters and not in CommonService and everywhere it's used has been updated

### DIFF
--- a/packages/relay/src/lib/services/ethService/blockService/BlockService.ts
+++ b/packages/relay/src/lib/services/ethService/blockService/BlockService.ts
@@ -19,7 +19,7 @@ import { Block, Log, Transaction } from '../../../model';
 import { IContractResultsParams, ITransactionReceipt, MirrorNodeBlock, RequestDetails } from '../../../types';
 import { CacheService } from '../../cacheService/cacheService';
 import { IBlockService, ICommonService } from '../../index';
-import { CommonService } from '../ethCommonService/CommonService';
+
 export class BlockService implements IBlockService {
   /**
    * The cache service used for caching all responses.


packages/relay/src/formatters.ts
```diff
@@ -261,6 +263,47 @@ const tinybarsToWeibars = (value: number | null, allowNegativeValues: boolean =
   return value == null ? null : value * constants.TINYBAR_TO_WEIBAR_COEF;
 };
 
+const formatContractResult = (cr: any): Transaction | null => {
```
acuarica, line 266: This function does not perform any actual formatting, right?. It's more like a factory function, as a such, what if we put it in the `transactionFactory.ts` file? (not necessary inside the class though).

acuarica (dismissed):
lg, left a suggestion.

packages/relay/src/formatters.ts
quiet-node: yes makes sense and I agree! Let's move it to the factory class and probably rename it to something else that's less misleading.

acuarica (approved):

quiet-node (commented):
LGTM however Codacy seems to fail. I'll approve after the fix.

PR #4210, opened 2025-08-18
chore: refactor `SDKClient` and `HAPIService`

### LINKED ISSUE #2804: Refactor SdkClient
### Problem

Currently, we are passing client and service classes around within methods and this is creating unnecessary coupling between the SdkClient and other classes, which we should address.

### Solution

To avoid this and simplify the dependency structure, we can take the following steps:

- ~~Follow the dependency tree: EthImpl -> HapiService -> AccountService, TransactionService, ContractService, FeeService -> SdkClient, MirrorNodeClient. We can consider opening separate issues to extract logic for accounts, contracts, and fees into distinct services~~.

- Keep the logic for initializing SdkClient into HapiService. This way, HapiService will manage the SdkClient and be the main point of interaction.

- ~~Remove the direct dependency on TransactionService from EthImpl. Instead, initialize TransactionService inside HapiService, and have HapiService pass the initialized SdkClient to it~~.

- Add methods such as createFile, deleteFile, appendFile, etc., directly in HapiService. This should be our main entry point for operations (e.g., instead of calling hapiService.getSdkClient().createFile(), we should directly call hapiService.createFile()).

- ~~The wrapper logic for executeQuery and executeTransaction should also reside within HapiService. HapiService can then use these wrappers and delegate the actual execution to the SdkClient as needed.~~

- ~~With HapiService calling TransactionService to fetch and extract transaction details, it remains agnostic to whether SdkClient or MirrorNodeClient is being used.~~

This approach untangles the dependencies between classes, resulting in a clear and linear dependency tree.

### Alternatives

_No response_

acuarica:
### Description

This PR ensures that `HAPIService` properly encapsulates the `SDKClient`. In this particular case, proper encapsulation is important because `HAPIService` recreates the `SDKClient` to avoid having a long-lived `SDKClient` https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/1197. That's why the `SDKClient` reference cannot escape from `HAPIService`.

Also with this change, our `SDKClient` fully contains the `Client` provided by the SDK. This in turn makes `HAPIService` simpler by only being responsible for creating `SDKClient`.

In addition, some public methods, usually getters, were unused in application code (only in tests). These methods were effectively removed. They were replaced by their `private` property directly.

### Related issue(s)

Fixes #2804.

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/services/hapiService/hapiService.ts
+++ b/packages/relay/src/lib/services/hapiService/hapiService.ts
@@ -230,6 +230,11 @@ export default class HAPIService {
     return this.client;
   }
 
+  /**
+   * Wrapper around the SDK client's `submitEthereumTransaction` method.
+   *
+   * See {@link SDKClient.submitEthereumTransaction} for more details.
+   */
   public async submitEthereumTransaction(
     transactionBuffer: Uint8Array,
     callerName: string,
@@ -248,6 +253,11 @@ export default class HAPIService {
     );
   }
 
+  /**
+   * Wrapper around the SDK client's `deleteFile` method.
+   *
+   * See {@link SDKClient.deleteFile} for more details.
+   */
   public async deleteFile(
     fileId: FileId,
     requestDetails: RequestDetails,
@@ -257,6 +267,11 @@ export default class HAPIService {
     return this.getSDKClient().deleteFile(fileId, requestDetails, callerName, originalCallerAddress);
   }
 
+  /**
+   * Wrapper around the SDK client's `getTransactionRecordMetrics` method.
+   *
+   * See {@link SDKClient.getTransactionRecordMetrics} for more details.
+   */
   public async getTransactionRecordMetrics(
     transactionId: string,
     txConstructorName: string,
--- a/packages/relay/src/lib/services/metricService/metricService.ts
+++ b/packages/relay/src/lib/services/metricService/metricService.ts
@@ -54,6 +54,7 @@ export default class MetricService {
    * @param logger - Logger instance for logging system messages.
    * @param metricsCollector - `SDKClient` or `HAPIService` for fetching transaction record metrics from the Consensus Node.
    * @param register - Registry instance for registering metrics.
+   * @param hbarLimitService - An instance of the HbarLimitService that tracks hbar expenses and limits.
    */
   constructor(
     logger: Logger,


packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -278,39 +278,6 @@ export class TransactionService implements ITransactionService {
...
-      this.logger.trace('sendTransaction()');
-    }
-    return predefined.UNSUPPORTED_METHOD;
-  }
-
-  /**
-   * Sign transaction - not supported
-   * @returns A JsonRpcError indicating that the method is not supported
-   */
-  public signTransaction(): JsonRpcError {
```
konstantinabl, line 296: I wonder why do we need to delete this, in case in the future we implement those, we would need to create them again? Having them in the transaction service, does it affect performance or anything else?

packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -321,34 +230,38 @@ export default class HAPIService {
...
-    return this.isReinitEnabled;
-  }
-
-  /**
-   * Return transaction count with current sdk instance.
-   * @returns transactionCount
-   */
-  public getTransactionCount() {
-    return this.transactionCount;
+  public async submitEthereumTransaction(
```
konstantinabl, line 233: Do you think it would be suitable to add ts doc here? I know we have it in the sdkClient as well, just wondering, not sure if it propagates upwards

packages/relay/src/lib/relay.ts
```diff
@@ -288,9 +280,9 @@ export class Relay {
         // Invoked when the registry collects its metrics' values.
         // Allows for updated account balance tracking
         try {
-          const operatorAccountId = clientMain.operatorAccountId!.toString();
+          const accountId = operatorAccountId!.toString();
```
simzzz, line 283: Why are we using the ! operator here when operatorAccountId is `AccountId | null`?

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
acuarica: IMO having this much indirection just to return a constant or error seems unnecessary overhead. I don't see the point of having to implement a method in a service that's going to be called only from one place just to return a constant (or error).

A good example is this PR change. We need to just update the signature of an `eth_` method, but we also need to modify the service and interface files to accomplish this.
https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4211/files#diff-8b5281eaec3750c0c951dc4aa88d62b77f87d296f78df2a20c9e38e07864aaef

packages/relay/src/lib/services/hapiService/hapiService.ts
acuarica: Good point, I don't think either, I'll add it.

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -17,78 +17,51 @@ import { HbarLimitService } from '../hbarLimitService';
...
-   * @param {Registry} register - Registry instance for registering metrics.
+   * @param logger - Logger instance for logging system messages.
+   * @param sdkClient - Client for interacting with the Hedera SDK.
+   * @param mirrorNodeClient - Client for querying the Hedera mirror node.
+   * @param register - Registry instance for registering metrics.
    */
   constructor(
     logger: Logger,
-    sdkClient: SDKClient,
+    private readonly cnMetrics: Pick<SDKClient, 'getTransactionRecordMetrics'>,
```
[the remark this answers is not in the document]
quiet-node: Yeah, “proxy” was a poor word choice. What I meant is that cnMetrics is essentially a slice of SDKClient an object that only exposes getTransactionRecordMetrics, which I like because it keeps only what’s needed.

By “inconsistency” I meant what Kosi mentioned MirrorNodeClient can also be passed as a slice of MNClient, containing only the required methods instead of the full object. That way, both would follow the same pattern.

So if we decide to go this route, I think we should apply it consistently to both parameters. Otherwise, we can just stick with passing the full SDKClient as before. My main point is just that it feels more consistent this way.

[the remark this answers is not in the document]
acuarica: > Otherwise, we can just stick with passing the full SDKClient as before.

Please notice that we can't, because we also pass a `HAPIService` in some sites.

packages/relay/src/lib/eth.ts
```diff
@@ -563,7 +564,7 @@ export class EthImpl implements Eth {
   @rpcMethod
   @rpcParamLayoutConfig(RPC_LAYOUT.REQUEST_DETAILS_ONLY)
   signTransaction(): JsonRpcError {
-    return this.transactionService.signTransaction();
+    return predefined.UNSUPPORTED_METHOD;
   }
```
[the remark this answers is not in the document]
quiet-node: I kindly disagree. In my opinion, PRs should remain focused and scoped. No matter how the technical debts or updates, mixing in with them distracts from the main goal of the PRs, makes reviews harder, and complicates tracing issues during future revisits. I know this particular update seems to be minor and convenient but it can open opportunities for others, in other PRs, to drift from the core purposes and will erode focus over time.

As for technical debts piling up, I don’t see what's wrong or anything bad with it. We can track and group related items under an umbrella issue as part of triaging and backlog grooming. That’s exactly what those processes are for keeping things organized and manageable anyway. In my view, the idea is “see something, report something.” Create a small issue and link it to the umbrella issue, or note it directly there. This way we can focus on the current priority and do not need to spend time to think about addressing technical debts at that moment. This could make sure we keep PRs on target while still making sure technical debt is acknowledged and queued for resolution.

packages/relay/src/lib/services/hapiService/hapiService.ts
```diff
@@ -321,34 +230,38 @@ export default class HAPIService {
...
+    );
   }
 
-  /**
-   * Return error codes which can trigger a sdk instance reinitialization.
-   * @returns errorCodes
-   */
-  public getErrorCodes() {
-    return this.errorCodes;
+  public async deleteFile(
```
konstantinabl, line 251: Does the same about ts doc apply here?

packages/relay/src/lib/services/metricService/metricService.ts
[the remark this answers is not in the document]
quiet-node: It would use either mirrorNodeClient or sdkClient (or HapiService now) based on GET_RECORD_DEFAULT_TO_CONSENSUS_NODE, right? How would that work out if we merge them into one, and then there's some decisions in the future make us switch the value of GET_RECORD_DEFAULT_TO_CONSENSUS_NODE to use the SDK instead?

packages/relay/src/lib/services/hapiService/hapiService.ts
acuarica: I've included a lightweight TSdoc comment to link them to their corresponding `SDKClient` method. I wanted to avoid duplicating the whole TSdoc from SDKClient.

konstantinabl (approved):

packages/relay/src/lib/services/metricService/metricService.ts
```diff
@@ -14,82 +12,57 @@ import {
 } from '../../types';
 import { HbarLimitService } from '../hbarLimitService';
 
+interface ITransactionMetricsCollector {
```
natanasow, line 15: q: should we place this interface in `packages/relay/src/lib/types/metrics.ts`?

packages/relay/src/lib/relay.ts
```diff
@@ -35,11 +35,9 @@ import { Web3Impl } from './web3';
 
 export class Relay {
   /**
-   * @private
-   * @readonly
-   * @property {Client} clientMain - The primary Hedera client used for interacting with the Hedera network.
+   * The primary Hedera client used for interacting with the Hedera network.
    */
-  private readonly clientMain: Client;
+  private readonly operatorAccountId: AccountId | null;
```
natanasow, line 40: q: can we transform the `AccountId` SDK type to a string, use only the string representation, and get rid of the SDK-specific types? Or are we using methods from `AccountId` object? 

It can be applicable for the `PublicKey` [here](https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4210/files#diff-4b0bcc39d9efb3e923f041c7459f1cc42d34597d4200839df6764c2a345a1c1bR123).

natanasow (approved):

packages/relay/src/lib/relay.ts
acuarica: That's a good point. I'll open a new issue as a follow up for this.

PR #4211, opened 2025-08-18
fix: align uncle-related RPC endpoints with Ethereum OpenRPC Spec

### LINKED ISSUE #3841: Fix eth_getUncleCountByBlockNumber to comply with Ethereum OpenRPC schema
### Problem

The current implementation of the `eth_getUncleCountByBlockNumber` method does not accept any parameters and always returns `'0x0'`, which is hardcoded as `constants.ZERO_HEX`.

```ts
async getUncleCountByBlockNumber(requestDetails: RequestDetails): Promise<string> {
  return constants.ZERO_HEX; // '0x0'
}
```

However, according to the Ethereum OpenRPC specification, the method **requires a `Block` parameter** with the following schema:

```json
{
  "name": "Block hash",
  "schema": {
    "$ref": "#/components/schemas/hash32"
  }
}
```

The `hash32` schema requires exactly 64 hexadecimal characters after the `0x` prefix. The currently returned `'0x0'` does **not** meet this requirement and is invalid if treated as a `hash32`.

### Solution

* Update the method to accept the required `Block` parameter in accordance with the Ethereum schema:

  ```ts
  async getUncleCountByBlockNumber(blockHash: string, requestDetails: RequestDetails): Promise<string>
  ```
* Add `@rpcParamValidationRules` for the `blockHash` parameter using `ParamType.HASH32` and mark it as required:

  ```ts
  @rpcParamValidationRules({
    0: { type: ParamType.HASH32, required: true },
  })
  ```
* Update `@rpcParamLayoutConfig` to include the parameter layout.
* Even though the method always returns `'0x0'` (since Hedera does not support uncles), it must validate the input per Ethereum’s expectations.
* Add tests to verify that missing or malformed hashes are rejected, and valid `hash32` inputs are accepted.

### Alternatives

_No response_

quiet-node:
### Description

This PR makes the following uncle-related endpoints conform to the Ethereum OpenRPC/JSON-RPC schema by accepting and validating the required parameters:

- `eth_getUncleCountByBlockNumber` - now accepts `blockNumOrTag` parameter
- `eth_getUncleCountByBlockHash` - now accepts `blockHash` parameter
- `eth_getUncleByBlockNumberAndIndex` - now accepts `blockNumOrTag` and `index` parameters
- `eth_getUncleByBlockHashAndIndex` - now accepts `blockHash` and `index` parameters

While Hedera still returns `0x0` for uncle counts and `null` for uncle blocks (no uncles exist), the endpoints now properly validate inputs and reject malformed/missing parameters per the OpenRPC specification.

### Related issue(s)

Fixes #3841

### Checklist

### DIFF
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -622,7 +622,6 @@
       "params": [
         {
           "name": "Block hash",
-          "description": "Hash of a block",
           "required": false,
           "schema": {
             "$ref": "#/components/schemas/hash32"
@@ -630,10 +629,9 @@
         },
         {
           "name": "Index",
-          "description": "Uncle index position",
           "required": false,
           "schema": {
-            "$ref": "#/components/schemas/hex"
+            "$ref": "#/components/schemas/uint"
           }
         }
       ],
@@ -652,7 +650,6 @@
       "params": [
         {
           "name": "Block",
-          "description": "Block number or tag",
           "required": false,
           "schema": {
             "$ref": "#/components/schemas/BlockNumberOrTag"
@@ -660,10 +657,9 @@
         },
         {
           "name": "Index",
-          "description": "Uncle index position",
           "required": false,
           "schema": {
-            "$ref": "#/components/schemas/hex"
+            "$ref": "#/components/schemas/uint"
           }
         }
       ],
@@ -682,7 +678,6 @@
       "params": [
         {
           "name": "Block hash",
-          "description": "Hash of a block",
           "required": false,
           "schema": {
             "$ref": "#/components/schemas/hash32"
@@ -706,7 +701,6 @@
       "params": [
         {
           "name": "Block",
-          "description": "Block number or tag",
           "required": false,
           "schema": {
             "$ref": "#/components/schemas/BlockNumberOrTag"


packages/relay/src/lib/services/ethService/blockService/BlockService.ts
```diff
@@ -241,44 +241,54 @@ export class BlockService implements IBlockService {
...
   /**
    * Always returns null. There are no uncles in Hedera.
-   * @returns null
+   *
+   * @param blockHash - The block hash
+   * @param index - The uncle index
+   * @returns null as Hedera does not support uncle blocks
    */
-  async getUncleByBlockHashAndIndex(): Promise<null> {
+  async getUncleByBlockHashAndIndex(blockHash: string, index: string): Promise<null> {
```
natanasow, line 249: This method returns only `null`, and it's not `async`. I think we can remove the `async` and change the return to simple `null` instead of `Promise<null>`. What do you think?

packages/relay/src/lib/services/ethService/blockService/BlockService.ts
```diff
@@ -241,44 +241,54 @@ export class BlockService implements IBlockService {
...
 
   /**
    * Always returns '0x0'. There are no uncles in Hedera.
-   * @returns '0x0'
+   *
+   * @param blockHash - The block hash
+   * @returns '0x0' as Hedera does not support uncle blocks
    */
-  async getUncleCountByBlockHash(): Promise<string> {
+  async getUncleCountByBlockHash(blockHash: string): Promise<string> {
```
natanasow, line 276: Same, we can remove `async` and change the return type to `string`.

packages/relay/src/lib/services/ethService/blockService/IBlockService.ts
```diff
@@ -13,8 +13,12 @@ export interface IBlockService {
   getBlockTransactionCountByHash: (hash: string, requestDetails: RequestDetails) => Promise<string | null>;
   getBlockTransactionCountByNumber: (blockNum: string, requestDetails: RequestDetails) => Promise<string | null>;
   getBlockReceipts: (blockHash: string, requestDetails: RequestDetails) => Promise<ITransactionReceipt[] | null>;
-  getUncleByBlockHashAndIndex: (requestDetails: RequestDetails) => Promise<null>;
-  getUncleByBlockNumberAndIndex: (requestDetails: RequestDetails) => Promise<null>;
-  getUncleCountByBlockHash: (requestDetails: RequestDetails) => Promise<string>;
-  getUncleCountByBlockNumber: (requestDetails: RequestDetails) => Promise<string>;
+  getUncleByBlockHashAndIndex: (blockHash: string, index: string, requestDetails: RequestDetails) => Promise<null>;
```
natanasow, line 16: Here, the return type should be changed as well.

packages/relay/src/lib/eth.ts
```diff
@@ -452,13 +452,18 @@ export class EthImpl implements Eth {
...
    */
   @rpcMethod
-  @rpcParamLayoutConfig(RPC_LAYOUT.REQUEST_DETAILS_ONLY)
-  async getUncleByBlockHashAndIndex(requestDetails: RequestDetails): Promise<null> {
-    return this.blockService.getUncleByBlockHashAndIndex(requestDetails);
+  @rpcParamValidationRules({
+    0: { type: 'blockHash', required: true },
+    1: { type: 'hex', required: true },
+  })
+  async getUncleByBlockHashAndIndex(blockHash: string, index: string, requestDetails: RequestDetails): Promise<null> {
```
natanasow, line 465: Same as the `BlockService` proposal, we can remove `async` and change the return type to `null`.

packages/relay/src/index.ts
```diff
@@ -113,13 +113,13 @@ export interface Eth {
 
   getTransactionReceipt(hash: string, requestDetails: RequestDetails): Promise<Receipt | null>;
 
-  getUncleByBlockHashAndIndex(requestDetails: RequestDetails): Promise<any>;
+  getUncleByBlockHashAndIndex(blockHash: string, index: string, requestDetails: RequestDetails): Promise<any>;
```
natanasow, line 116: If we agree on https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4211/files#r2288006882, we must fix these as well.

packages/relay/src/lib/services/ethService/blockService/IBlockService.ts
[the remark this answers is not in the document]
quiet-node: good catch! Fixed

packages/relay/src/lib/services/ethService/blockService/BlockService.ts
quiet-node: Yes that's true and I like it! Fixed

packages/relay/src/lib/services/ethService/blockService/BlockService.ts
quiet-node: Indeed! Fixed

packages/relay/src/lib/eth.ts
quiet-node: Indeed! Fixed

packages/relay/src/index.ts
quiet-node: Indeed! Fixed

packages/relay/src/lib/eth.ts
```diff
@@ -482,13 +496,16 @@ export class EthImpl implements Eth {
...
+   * @param blockHash - The block hash
    * @param {RequestDetails} requestDetails - Details about the request for logging and tracking
    * @returns {Promise<string>} Always returns '0x0'
    */
   @rpcMethod
-  @rpcParamLayoutConfig(RPC_LAYOUT.REQUEST_DETAILS_ONLY)
-  async getUncleCountByBlockHash(requestDetails: RequestDetails): Promise<string> {
-    return this.blockService.getUncleCountByBlockHash(requestDetails);
+  @rpcParamValidationRules({
+    0: { type: 'blockHash', required: true },
```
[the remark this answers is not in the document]
quiet-node: Yeah nice catch! Ha that's bazzard how it is not required while it's the only indicator to find the uncle counts. I think we should keep them as required as they make sense this way. 

Also it seems like other providers make these params required as well e.g. [Metamask](https://docs.metamask.io/services/reference/ethereum/json-rpc-methods/eth_getunclecountbyblockhash/), [Alchemy](https://www.alchemy.com/docs/node/base/base-api-endpoints/eth-get-uncle-count-by-block-hash),  [QuickNode](https://www.quicknode.com/docs/ethereum/eth_getUncleCountByBlockHash), [CoinAPI](https://docs.coinapi.io/naas-api/ethereum/Methods/eth_getUncleCountByBlockHash), etc. 

Wdyt?

packages/relay/tests/lib/openrpc.spec.ts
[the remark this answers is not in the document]
quiet-node: Right great catch! Definitely missed this! Fixed

packages/server/tests/acceptance/rpc_batch2.spec.ts
```diff
@@ -664,49 +664,95 @@ describe('@api-batch-2 RPC Server Acceptance Tests', function () {
...
 
-    it('should execute "eth_getUncleByBlockHashAndIndex" for non-existing block hash and index=0', async function () {
-      const res = await relay.call(
-        RelayCalls.ETH_ENDPOINTS.ETH_GET_UNCLE_BY_BLOCK_HASH_AND_INDEX,
-        [Address.NON_EXISTING_BLOCK_HASH, 0],
-        requestId,
-      );
-      expect(res).to.be.null;
+    it('should fail to execute "eth_getUncleByBlockHashAndIndex" with invalid parameters', async function () {
+      for (const invalidBlockHash of [
```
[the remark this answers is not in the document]
quiet-node: yeah good point. Updated.

natanasow (dismissed):

konstantinabl (dismissed):

packages/relay/tests/lib/openrpc.spec.ts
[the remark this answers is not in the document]
acuarica: @quiet-node if you feel brave, you can try to update `openrpc.json` once more, and ensure the openrpc updater doesn't kick-off (of course, this has to be done after rebasing onto `main` now we have merged https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4299)

packages/relay/src/lib/eth.ts
```diff
@@ -453,13 +453,17 @@ export class EthImpl implements Eth {
    * @rpcMethod Exposed as eth_getUncleByBlockHashAndIndex RPC endpoint
    * @rpcParamLayoutConfig decorated method parameter layout
    *
-   * @param {RequestDetails} requestDetails - Details about the request for logging and tracking
+   * @param blockHash - The block hash
+   * @param index - The uncle index
    * @returns {Promise<null>} Always returns null
```
acuarica, line 458: super nit 😄 : remove or update the return type so it matches the TypeScript definition. (better if we remove it IMO)

packages/relay/src/lib/eth.ts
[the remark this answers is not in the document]
acuarica: We should stick to the execution-apis as much when possible. In this case won't matter much because these endpoints are not used, but in general it's a good practice. Sticking to the execution-apis will save us some headaches when comparing against it.

acuarica (approved):

packages/relay/src/lib/eth.ts
quiet-node: Ah yeah missed that during clean up. Yeah I agree we can remove it. Doesn't show up in the hover tooltip window any. Updated

natanasow (approved):

packages/relay/tests/lib/openrpc.spec.ts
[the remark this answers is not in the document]
acuarica: We can sync up offline to discuss this.

PR #4360, opened 2025-09-05
refactor: optimize precheckSendRawTransactionCheck for improved efficiency and code cleanliness

### LINKED ISSUE #3721: Refactor precheck.sendRawTransactionCheck for improved efficiency and code cleanliness
The current implementation of `precheck.sendRawTransactionCheck()` has several inefficiencies:

1. Redundant logging in individual precheck methods that's redundant with the catch block in `validateRawTransaction`
2. The `balance()` precheck method unnecessarily re-validates the account object which was already validated by `verifyAccount()`
3. The `balance()` method contains redundant code: `const result = {passes: false, error: predefined.INSUFFICIENT_ACCOUNT_BALANCE}`
4. Somehow, the `mirrorAccountInfo` object retrival for `nonce()` and `balance()` precheck feels awkward and sort of cumbersome. Not sure what makes it feel like so but just wanted to put it here for later consideration. Same for `receiverAccount()` precheck

Proposed changes:

- Remove unnecessary logging in individual precheck methods
- Refactor the `balance()` method to accept just the account balance rather than the whole account object
- Remove redundant account validation in the `balance()` method
- Remove the unused `result` object in the `balance()` method

quiet-node:
### Description

This PR refactors the `precheckSendRawTransactionCheck` method in the Precheck class for improved efficiency and code cleanliness:

- Reorders prechecks to perform cheaper validations first (before network calls)
- Simplifies the balance check logic by removing unnecessary object wrapping
- Removes unused parameters and excessive trace logging
- Renames variables for better clarity (e.g., `mirrorAccountInfo` to `signerAccountInfo`)
- Enhances error messages with more specific details (e.g., including transaction type in unsupported transaction errors)

### Related issue(s)

Fixes #3721

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/precheck.ts
+++ b/packages/relay/src/lib/precheck.ts
@@ -7,7 +7,7 @@ import { Logger } from 'pino';
 import { prepend0x } from '../formatters';
 import { MirrorNodeClient } from './clients';
 import constants from './constants';
-import { JsonRpcError, predefined } from './errors/JsonRpcError';
+import { predefined } from './errors/JsonRpcError';
 import { CommonService } from './services';
 import { RequestDetails } from './types';
 
@@ -69,12 +69,12 @@ export class Precheck {
     this.transactionSize(parsedTx);
     this.transactionType(parsedTx);
     this.gasLimit(parsedTx);
-    const mirrorAccountInfo = await this.verifyAccount(parsedTx, requestDetails);
-    this.nonce(parsedTx, mirrorAccountInfo.ethereum_nonce);
     this.chainId(parsedTx);
     this.value(parsedTx);
     this.gasPrice(parsedTx, networkGasPriceInWeiBars);
-    this.balance(parsedTx, mirrorAccountInfo);
+    const signerAccountInfo = await this.verifyAccount(parsedTx, requestDetails);
+    this.nonce(parsedTx, signerAccountInfo.ethereum_nonce);
+    this.balance(parsedTx, signerAccountInfo.balance.balance);
     await this.receiverAccount(parsedTx, requestDetails);
   }
 
@@ -87,15 +87,6 @@ export class Precheck {
   async verifyAccount(tx: Transaction, requestDetails: RequestDetails): Promise<any> {
     const accountInfo = await this.mirrorNodeClient.getAccount(tx.from!, requestDetails);
     if (accountInfo == null) {
-      if (this.logger.isLevelEnabled('trace')) {
-        this.logger.trace(
-          `Failed to retrieve address '${
-            tx.from
-          }' account details from mirror node on verify account precheck for sendRawTransaction(transaction=${JSON.stringify(
-            tx,
-          )})`,
-        );
-      }
       throw predefined.RESOURCE_NOT_FOUND(`address '${tx.from}'.`);
     }
 
@@ -108,32 +99,22 @@ export class Precheck {
    * @param accountInfoNonce - The nonce of the account.
    */
   nonce(tx: Transaction, accountInfoNonce: number): void {
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(
-        `Nonce precheck for sendRawTransaction(tx.nonce=${tx.nonce}, accountInfoNonce=${accountInfoNonce})`,
-      );
-    }
-
     if (accountInfoNonce > tx.nonce) {
       throw predefined.NONCE_TOO_LOW(tx.nonce, accountInfoNonce);
     }
   }
 
   /**
-   * Checks the chain ID of the transaction.
-   * @param tx - The transaction.
+   * Validates that the transaction's chain ID matches the network's chain ID.
+   * Legacy unprotected transactions (pre-EIP155) are exempt from this check.
+   *
+   * @param {Transaction} tx - The transaction to validate.
+   * @throws {JsonRpcError} If the transaction's chain ID doesn't match the network's chain ID.
    */
   chainId(tx: Transaction): void {
     const txChainId = prepend0x(Number(tx.chainId).toString(16));
     const passes = this.isLegacyUnprotectedEtx(tx) || txChainId === this.chain;
     if (!passes) {
-      if (this.logger.isLevelEnabled('trace')) {
-        this.logger.trace(
-          `Failed chainId precheck for sendRawTransaction(transaction=%s, chainId=%s)`,
-          JSON.stringify(tx),
-          txChainId,
-        );
-      }
       throw predefined.UNSUPPORTED_CHAIN_ID(txChainId, this.chain);
     }
   }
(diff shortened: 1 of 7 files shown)


packages/relay/src/lib/precheck.ts
```diff
@@ -204,59 +177,15 @@ export class Precheck {
 
   /**
    * Checks the balance of the sender account.
-   * @param tx - The transaction.
-   * @param account - The account information.
+   * @param {Transaction} tx - The transaction.
```
acuarica, line 180: I'd remove the JSdoc type.

packages/relay/src/lib/precheck.ts
```diff
@@ -204,59 +177,15 @@ export class Precheck {
...
-  balance(tx: Transaction, account: any): void {
-    const result = {
-      passes: false,
-      error: predefined.INSUFFICIENT_ACCOUNT_BALANCE,
-    };
-
+  balance(tx: Transaction, accountBalance: number): void {
     const txGasPrice = BigInt(tx.gasPrice || tx.maxFeePerGas! + tx.maxPriorityFeePerGas!);
     const txTotalValue = tx.value + txGasPrice * tx.gasLimit;
+    const accountBalanceInWeiBars = BigInt(accountBalance.toString()) * BigInt(constants.TINYBAR_TO_WEIBAR_COEF);
```
acuarica, line 186: no need to convert `accountBalance` to string to convert it to `BigInt`.

packages/relay/src/lib/precheck.ts
```diff
@@ -204,59 +177,15 @@ export class Precheck {
...
-            tx,
-          )}, totalValue=${txTotalValue})`,
-        );
-      }
-      throw predefined.RESOURCE_NOT_FOUND(`tx.from '${tx.from}'.`);
-    }
-
-    let tinybars: bigint;
-    try {
-      tinybars = BigInt(account.balance.balance.toString()) * BigInt(constants.TINYBAR_TO_WEIBAR_COEF);
```
acuarica, line 232: I can see 2 things can failing here, retrieving of `account.balance.balance` and the actual parsing of the value. How can we ensure that no error happens here? Because there is no check in `sendRawTransactionCheck` neither. Or this `try/catch` was redundant all along?

packages/relay/src/lib/precheck.ts
```diff
@@ -356,12 +267,7 @@ export class Precheck {
   transactionType(tx: Transaction) {
     // Blob transactions are not supported as per HIP 866
     if (tx.type === 3) {
-      if (this.logger.isLevelEnabled('trace')) {
-        this.logger.trace(
-          `Transaction with type=${tx.type} is unsupported for sendRawTransaction(transaction=${JSON.stringify(tx)})`,
-        );
-      }
-      throw predefined.UNSUPPORTED_TRANSACTION_TYPE;
+      throw predefined.UNSUPPORTED_TRANSACTION_TYPE(tx.type);
```
acuarica, line 270: is there a case where we don't use `3` as `txType` for `UNSUPPORTED_TRANSACTION_TYPE`? It seems it's always `3`, if that's the case we should remove it.

packages/relay/src/lib/precheck.ts
quiet-node: ha nice catch you're right don't know what I was thinking lol

packages/relay/src/lib/precheck.ts
quiet-node: Not quite clear on this one. Isn't the try/catch handle the error if any of the 2 failing cases you mentioned happened? wdym by `Because there is no check in sendRawTransactionCheck neither. Or this try/catch was redundant all along?`?

packages/relay/src/lib/precheck.ts
acuarica: > and easier to extend for different txType values in the future.

The problem I see with this approach is that when I encounter `predefined.UNSUPPORTED_TRANSACTION_TYPE(3)` it makes me believe that there are other unsupported types. So that's why I'd go with the latter. `UNSUPPORTED_TRANSACTION_TYPE_3` makes more explicit that the only unsupported type is `3`.

packages/relay/src/lib/precheck.ts
```diff
@@ -205,58 +177,14 @@ export class Precheck {
...
-   * @param account - The account information.
+   * @param accountBalance - The account balance in tinybars.
    */
-  balance(tx: Transaction, account: any): void {
-    const result = {
-      passes: false,
-      error: predefined.INSUFFICIENT_ACCOUNT_BALANCE,
-    };
-
+  balance(tx: Transaction, accountBalance: number): void {
```
acuarica, line 182: I saw some tests converting to `number` probably just to satisfy this `accountBalance`'s type. Maybe better if we use

```suggestion
  balance(tx: Transaction, accountBalance: number | string): void {
```

to allow both. And it also it's safer to use `string` than `number`.

packages/relay/src/lib/precheck.ts
acuarica: So I see the `try/catch` was removed here in the `balance` method (and it wasn't moved either to the `sendRawTransactionCheck` method). How can we ensure that

```typescript
const accountBalanceInWeiBars = BigInt(accountBalance) * BigInt(constants.TINYBAR_TO_WEIBAR_COEF);
```

doesn't fail?

If it never fails, then, can we assume that the `try/catch` was redundant? Hope it's a bit more clear now.

quiet-node: Ah, totally I get it now. Good catch indeed! I think the try/catch was added to prevent errors like “Cannot read properties of null” since the official MN getAccount endpoint can return a nullable account.balance. 

I see two possible approaches:
	1.	Check account.balance inside verifyAccount() right after got the account back from MN.
	2.	Pass `account.balance` object to `Precheck.balance()` instead of `account.balance.balance`, and then validate it inside `balance()`. That keeps the check more localized and makes sense since it’s directly related to balance handling.

What do you think?

packages/relay/src/lib/precheck.ts
quiet-node: Agree! Updated to use UNSUPPORTED_TRANSACTION_TYPE_3

packages/relay/src/lib/precheck.ts
acuarica: I'd go with option 2., as you point out, it looks to me a better solution.

packages/relay/src/lib/precheck.ts
```diff
@@ -69,33 +69,23 @@ export class Precheck {
...
     this.gasLimit(parsedTx);
-    const mirrorAccountInfo = await this.verifyAccount(parsedTx, requestDetails);
-    this.nonce(parsedTx, mirrorAccountInfo.ethereum_nonce);
     this.chainId(parsedTx);
     this.value(parsedTx);
     this.gasPrice(parsedTx, networkGasPriceInWeiBars);
-    this.balance(parsedTx, mirrorAccountInfo);
+    const signerAccountInfo = await this.verifyAccount(parsedTx, requestDetails);
+    this.nonce(parsedTx, signerAccountInfo.ethereum_nonce);
+    this.balance(parsedTx, signerAccountInfo.balance.balance);
```
simzzz, line 77: Are we always guaranteed to have balance in `signerAccountInfo`?

packages/relay/src/lib/precheck.ts
quiet-node: Coolio okay went with option 2 and also add this validity check for nonce() method as well.

simzzz (approved):
LGTM

PR #4404, opened 2025-09-23
fix: debug_traceTransaction validation for default tracer

### LINKED ISSUE #4368: `debug_traceTransaction` throws error in Remix IDE
### Description

Trying to debug a transaction in Remix IDE fails with the following error:
```
Error while querying the provider: [Request ID: c43567bd-9d5e-431a-bd91-f72bf3b8ecbc] Invalid parameter 'disableStorage' for TracerConfigWrapper: Unknown parameter
server response 400 Bad Request (request={  }, response={  }, error=null, info={ "requestUrl": "http://127.0.0.1:7546", "responseBody": "{\"error\":{\"code\":-32602,\"message\":\"[Request ID: c43567bd-9d5e-431a-bd91-f72bf3b8ecbc] Invalid parameter 'disableStorage' for TracerConfigWrapper: Unknown parameter\"},\"jsonrpc\":\"2.0\",\"id\":13}", "responseStatus": "400 Bad Request" }, code=SERVER_ERROR, version=6.14.0)
```
In previous versions of the relay, this worked.

### Steps to reproduce

1. Execute a transaction
2. Open Remix IDE and connect via external provider
3. Paste the transaction hash in the debugger and click `Start debugging`

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.71.0

### Operating system

macOS

simzzz:
### Description

When tracer is not specified, the second parameter has more fields that are specific to the [opcode logger](https://geth.ethereum.org/docs/developers/evm-tracing/built-in-tracers#structopcode-logger) and which will be ignored when tracer field is set to any value, such as `disableStorage`, `enableMemory`, `disableStack`, `fullStorage`. The case when tracer is not specified was missed during implementation, and after implementing stricter validation in 0.70 has resulted in an error. This is the request that Remix makes, hence the bug.

### Related issue(s)

Fixes #4368 

### Testing Guide

1. Run the relay locally from this branch
2. Open Remix IDE, install the debugger plugin, and use custom external HTTP provider pointing to the local relay URL
3. Pass a transaction hash to the debugger and click start debugging

There is a config property `fullStorage` which is not explicitly specified by Geth, but it is passed by Remix, however it is always set to false when passed. In order not to break Remix compatibility, I have added it to the validation, however it will be ignored after that.

N/A

### Checklist

### DIFF
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -1176,7 +1176,7 @@
           "schema": {
             "$ref": "#/components/schemas/TracerConfigWrapper"
           },
-          "description": "Configuration object for the tracer."
+          "description": "Configuration object for the tracer. Supports two formats: 1) Standard nested format with 'tracer' and 'tracerConfig' properties, 2) Simplified format with opcodeLogger config properties directly at top level (when no explicit 'tracer' is specified, defaults to opcodeLogger)."
         }
       ],
       "result": {
@@ -1665,6 +1665,10 @@
           },
           "disableStorage": {
             "type": "boolean"
+          },
+          "fullStorage": {
+            "type": "boolean",
+            "description": "Non-standard parameter sometimes sent by Remix - accepted for compatibility but ignored in implementation"
           }
         }
       },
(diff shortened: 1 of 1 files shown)


quiet-node (approved):
LGTM nice work! debug endpoints are always complicating. Just wonder if we need to do this for debug_traceBlockByNumber as well? If yes we should extend this PR or maybe better to open a new ticket for it?

PR #4409, opened 2025-09-24
feat: Мove creation of redis client outside of cache service

### LINKED ISSUE #4387: Move creation of redis client outside of cache service
Currently, we create the Redis client in the cache service, because thats where we pass it to the RedisCache class. However, since other classes may need the client to connect to Redis its best to move it out of the cacheService class.

konstantinabl:
### Description

This PR's goal to move the redis client creation out of the RedisCache class, since the client will be reused to connect to Redis for out Transaction Pool feature. On top of this, RedisCache's main purpose is to handle operations for get,set, delete etc from Redis, not maintaining the client.

### Related issue(s)

Fixes #4387

In our original design, the creation of the redis client and the maintenance of the connection was handled in the Redis Cache class, which breaks the first SOLID principle. Now the creation of the client and the connection is moved to a RedisClientManager class, which will manage the client and any operations on it. All client related methods are removed from RedisCache and CacheService classes.

### Notes for Reviewer

1. We always pass a connected client, so there is no need to re-check here
https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4409/files#diff-244a262ac886fc247f5483123909ff160f65a2acfe6957f1a5b35d58453b725eL307

2. Redis may not be enabled, then the redisClient will be undefined and we need to switch to LRU cache
https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4409/files#diff-6885de8b57b52c8ab19662969cbcf2d242b9944777fe284d756168a2af093d0dR93

3. These methods have been deleted, since they were only ever used in tests, which is not a good practice
https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4409/files#diff-6885de8b57b52c8ab19662969cbcf2d242b9944777fe284d756168a2af093d0dL127

4. These tests have been deleted because they belong in the test file for the new RedisClientManager
https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4409/files#diff-7aaa58f003fccbb2493cf321726c1878a5532f99de9148f2876aae60978722ceL226
https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4409/files#diff-7aaa58f003fccbb2493cf321726c1878a5532f99de9148f2876aae60978722ceL499

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/eth.ts
+++ b/packages/relay/src/lib/eth.ts
@@ -968,6 +968,7 @@ export class EthImpl implements Eth {
   @rpcParamValidationRules({
     0: { type: 'transaction', required: true },
     1: { type: 'blockParams', required: true },
+    2: { type: 'stateOverride', required: false },
   })
   @cache({
     skipParams: [{ index: '1', value: constants.NON_CACHABLE_BLOCK_PARAMS }],
--- a/packages/relay/src/lib/services/index.ts
+++ b/packages/relay/src/lib/services/index.ts
@@ -15,3 +15,4 @@ export * from '../types/rateLimiter';
 export * from './rateLimiterService/LruRateLimitStore';
 export * from './rateLimiterService/RedisRateLimitStore';
 export * from './rateLimiterService/rateLimiterService';
+export * from './transactionPoolService/LocalPendingTransactionStorage';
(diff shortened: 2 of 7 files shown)


packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -107,73 +113,6 @@ export class CacheService {
...
-    }
-    return false;
-  }
-
-  /**
-   * Connects the Redis client if it is not already connected.
-   *
-   * @returns {Promise<void>} A Promise that resolves when the client is connected.
-   */
-  async connectRedisClient(): Promise<void> {
```
konstantinabl, line 127: These methods are deleted, since they were only used in tests

natanasow (approved):

Issue #3574, opened 2025-03-21
Wrap Logger with Custom Logger Service


Ferparishuertas:
### Problem

Our logging logic is quite verbose, making the code quite long, due the need of:

1. Check if trace or debug is enabled, to avoid the internal pino serialization perhaps not activated
2. We are using either RequestDetails, request, to extract requestid at log level, making either overloading methods with unnecesary params, or making de log string a bit longer

### Solution

When I read the JSON RPC code I want to be able to read it easily without being overwhelmed with dirty logging checks and logic

- Create a LoggerService the wraps Pino Logger ( making also the code Agnostic the the used logger, pino or Other)
- Include all levels methods, with internal checks for debug and info enabled, making code where more readable not needinf to check on every log as it is checked internally
- Overload methods with traceWithRequestDetails, traceWithRequest, to encapsulate the logic of extracting requestid, making code way more cleaner

> [!NOTE]
> Problem **2.** was moved here https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/4080.

### Alternatives

TBD

PR #4421, opened 2025-09-27
feat: replace isLevelEnabled guards with Pino interpolation values

### LINKED ISSUE #4080: Use "interpolation values" instead of string interpolation when logging
Our log messages are guarded by the `isLevelEnabled` method. This guard was introduced in https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/1118 to avoid building the `message` argument when the corresponding log level is not active. This is because, in most cases, we use JavaScript's string interpolation to build the `message` argument. In turn this causes performance issues because the `message` is built regardless of the active log level.

However, as mentioned in https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3574, this clutters the code making harder to read and it is error prone when introducing new logging statements.

An alternative approach can be to use Pino's [`interpolationValues`](https://getpino.io/#/docs/api?id=interpolationvalues). In a nutshell, interpolation values allow us to log using a literal string (not a JavaScript template string) and additional values using a `printf`-like format. In this scenario, no actual `message` is being built, so any overhead should be negligible.

Therefore, we should remove the log level enabled check. That is, we should revert PR https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/3154 and any other instance where we use `isLevelEnabled`. And replace it replace it with Pino's interpolation values.

Ideally, we want to ensure that this change does not affect performance. We should run K6 tests to compare alternatives, and only introduce this change if it does not affect performance.

### Acceptance Criteria

- [ ] `warn`, `info`, `debug` and `trace` calls should receive as first argument a string literal (not a template literal) using interpolation values. The rest of the arguments **must** be variable expressions. That is, no additional computation, _e.g._, function call, string concat, object creation, _etc._, can be passed to a logger method. If an actual computation is needed, the guard must be kept. See below.
- [ ] A few log entries effectively need some preprocessing before doing the actual logging. In those cases, the `isLevelEnabled` guard should be kept. See for example https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/08827c8700808e8d9cf5f51b24355f5526b748a7/packages/relay/src/lib/clients/cache/redisCache.ts#L117-L121
- [ ] Many methods decorated with `@rpcMethod` include a `trace` at the beginning. These can traces (and their guards) can be safely removed. The dispatcher already logs when an RPC method is being executed.
- [ ] The removal of `isLevelEnabled` guards **should not** affect performance.

> This is a follow up of https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/3574. PR #3990 only dealt with removing the `Request ID` prefix from logging calls.

belloibrahv:
Based on the changes we've implemented for issue #4080, here's the PR description using the template:

---

### Description

This PR replaces `isLevelEnabled` guards with Pino interpolation values to improve logging performance and code readability. The current implementation uses JavaScript string interpolation within conditional guards, which causes performance overhead by building log messages even when the corresponding log level is disabled. By leveraging Pino's interpolation values (printf-style formatting), we eliminate this overhead while maintaining the same logging functionality.

The changes follow the approach outlined in issue #3574, addressing the cluttered code and error-prone nature of manually adding `isLevelEnabled` guards throughout the codebase.

### Related issue(s)

Fixes #4080

### Testing Guide

1. **Verify logging still works correctly**: Run the application and check that log messages appear at appropriate levels (trace, debug, info, warn)
2. **Test performance**: Compare log performance before and after changes - interpolation values should have negligible overhead compared to string building
3. **Check RPC method execution**: Verify that RPC method execution is still properly logged by the dispatcher (removed redundant trace logs from @rpcMethod methods)
4. **Validate sensitive data handling**: Ensure that cases requiring data preprocessing (censoring, JSON.stringify) still have proper guards in place

N/A

- The remaining compilation errors (BigNumber type conflicts, logger method signature issues) are pre-existing and should be addressed in separate PRs
- Consider running K6 performance tests to validate the performance improvements as mentioned in the original issue

### Checklist

---

**Summary of Changes:**
- **16 files modified** with logging improvements
- **Replaced simple guards** with Pino interpolation values (`%s`, `%o`)
- **Removed redundant @rpcMethod trace logs** (dispatcher handles this)
- **Preserved guards** for cases requiring preprocessing
- **Improved code readability** by eliminating conditional logging clutter
- **Maintained functionality** while improving performance

### DIFF
--- a/packages/relay/src/lib/clients/cache/redisCache.ts
+++ b/packages/relay/src/lib/clients/cache/redisCache.ts
@@ -149,9 +149,7 @@ export class RedisCache implements IRedisCacheClient {
     const message = `Caching ${censoredKey}:${censoredValue} on ${callingMethod} for ${
       resolvedTtl > 0 ? `${resolvedTtl} ms` : 'indefinite time'
     }`;
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(`${message}`);
-    }
+    this.logger.trace('%s', message);
     // TODO: add metrics
   }
 
--- a/packages/relay/src/lib/clients/sdkClient.ts
+++ b/packages/relay/src/lib/clients/sdkClient.ts
@@ -557,11 +557,11 @@ export class SDKClient {
     let transactionFee: number = 0;
     let txRecordChargeAmount: number = 0;
     try {
-      if (this.logger.isLevelEnabled('debug')) {
-        this.logger.debug(
-          `Get transaction record via consensus node: transactionId=${transactionId}, txConstructorName=${txConstructorName}`,
-        );
-      }
+      this.logger.debug(
+        'Get transaction record via consensus node: transactionId=%s, txConstructorName=%s',
+        transactionId,
+        txConstructorName,
+      );
 
       const transactionRecord = await new TransactionRecordQuery()
         .setTransactionId(transactionId)
--- a/packages/relay/src/lib/debug.ts
+++ b/packages/relay/src/lib/debug.ts
@@ -115,10 +115,6 @@ export class DebugImpl implements Debug {
     tracerObject: TransactionTracerConfig,
     requestDetails: RequestDetails,
   ): Promise<any> {
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(`traceTransaction(${transactionIdOrHash})`);
-    }
-
     //we use a wrapper since we accept a transaction where a second param with tracer/tracerConfig may not be provided
     //and we will still default to opcodeLogger
     const tracer = tracerObject?.tracer ?? TracerType.OpcodeLogger;
--- a/packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
+++ b/packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
@@ -79,9 +79,7 @@ export class RpcMethodDispatcher {
     const operationHandler = this.methodRegistry.get(rpcMethodName);
 
     if (!operationHandler) {
-      if (this.logger.isLevelEnabled('debug')) {
-        this.logger.debug(`RPC method not found in registry: rpcMethodName=${rpcMethodName}`);
-      }
+      this.logger.debug('RPC method not found in registry: rpcMethodName=%s', rpcMethodName);
 
       throw this.throwUnregisteredRpcMethods(rpcMethodName);
     }
@@ -90,11 +88,11 @@ export class RpcMethodDispatcher {
     const methodParamSchemas = operationHandler[RPC_PARAM_VALIDATION_RULES_KEY];
 
     if (methodParamSchemas) {
-      if (this.logger.isLevelEnabled('info')) {
-        this.logger.info(
-          `Validating method parameters for ${rpcMethodName}, params: ${JSON.stringify(rpcMethodParams)}`,
-        );
-      }
+      this.logger.info(
+        'Validating method parameters for %s, params: %s',
+        rpcMethodName,
+        JSON.stringify(rpcMethodParams),
+      );
       validateParams(rpcMethodParams, methodParamSchemas);
     }
 
--- a/packages/relay/src/lib/services/ethService/accountService/AccountService.ts
+++ b/packages/relay/src/lib/services/ethService/accountService/AccountService.ts
@@ -99,9 +99,7 @@ export class AccountService implements IAccountService {
     blockNumberOrTagOrHash: string,
     requestDetails: RequestDetails,
   ): Promise<string> {
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(`getBalance(account=${account}, blockNumberOrTag=${blockNumberOrTagOrHash})`);
-    }
+    this.logger.trace('getBalance(account=%s, blockNumberOrTag=%s)', account, blockNumberOrTagOrHash);
 
     let latestBlock: LatestBlockNumberTimestamp | null | undefined;
     // this check is required, because some tools like Metamask pass for parameter latest block, with a number (ex 0x30ea)
--- a/packages/relay/src/lib/services/ethService/blockService/BlockService.ts
+++ b/packages/relay/src/lib/services/ethService/blockService/BlockService.ts
@@ -155,11 +155,11 @@ export class BlockService implements IBlockService {
 
     const receiptPromises = contractResults.map(async (contractResult) => {
       if (Utils.isRevertedDueToHederaSpecificValidation(contractResult)) {
-        if (this.logger.isLevelEnabled('debug')) {
-          this.logger.debug(
-            `Transaction with hash ${contractResult.hash} is skipped due to hedera-specific validation failure (${contractResult.result})`,
-          );
-        }
+        this.logger.debug(
+          'Transaction with hash %s is skipped due to hedera-specific validation failure (%s)',
+          contractResult.hash,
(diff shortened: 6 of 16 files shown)


packages/relay/src/lib/debug.ts
```diff
@@ -115,10 +115,6 @@ export class DebugImpl implements Debug {
     tracerObject: TransactionTracerConfig,
     requestDetails: RequestDetails,
   ): Promise<any> {
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(`traceTransaction(${transactionIdOrHash})`);
-    }
-
```
quiet-node: Should we also remove the same log for the traceBlockByNumber() method below?

packages/relay/src/lib/services/ethService/feeService/FeeService.ts
```diff
@@ -139,9 +139,7 @@ export class FeeService implements IFeeService {
    * @param requestDetails
    */
   public async maxPriorityFeePerGas(): Promise<string> {
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace('maxPriorityFeePerGas()');
-    }
+    this.logger.trace('maxPriorityFeePerGas()');
```
quiet-node: We can safely remove this as well

packages/relay/src/lib/services/ethService/contractService/ContractService.ts
```diff
@@ -100,9 +100,7 @@ export class ContractService implements IContractService {
    * @returns An empty array of addresses
    */
   public accounts(): [] {
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(`accounts()`);
-    }
+    this.logger.trace('accounts()');
```
quiet-node: Let's remove these as well. Don't add much information anyway.

packages/relay/src/lib/services/ethService/blockService/BlockService.ts
```diff
@@ -155,11 +155,11 @@ export class BlockService implements IBlockService {
...
       if (Utils.isRevertedDueToHederaSpecificValidation(contractResult)) {
-        if (this.logger.isLevelEnabled('debug')) {
-          this.logger.debug(
-            `Transaction with hash ${contractResult.hash} is skipped due to hedera-specific validation failure (${contractResult.result})`,
-          );
-        }
+        this.logger.debug(
+          'Transaction with hash %s is skipped due to hedera-specific validation failure (%s)',
+          contractResult.hash,
+          contractResult.result,
```
quiet-node: Looks like lines 209, 229, 250, 264, 277, 290, 430, 455 can be updated as well

packages/relay/src/lib/services/ethService/accountService/AccountService.ts
```diff
@@ -99,9 +99,7 @@ export class AccountService implements IAccountService {
     blockNumberOrTagOrHash: string,
     requestDetails: RequestDetails,
   ): Promise<string> {
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(`getBalance(account=${account}, blockNumberOrTag=${blockNumberOrTagOrHash})`);
-    }
+    this.logger.trace('getBalance(account=%s, blockNumberOrTag=%s)', account, blockNumberOrTagOrHash);
```
quiet-node: line 297 can be updated, 305 same as doesn't need JSON.stringify(), 345 can be removed,

packages/relay/src/lib/clients/cache/redisCache.ts
```diff
@@ -149,9 +149,7 @@ export class RedisCache implements IRedisCacheClient {
     const message = `Caching ${censoredKey}:${censoredValue} on ${callingMethod} for ${
       resolvedTtl > 0 ? `${resolvedTtl} ms` : 'indefinite time'
     }`;
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(`${message}`);
-    }
+    this.logger.trace('%s', message);
```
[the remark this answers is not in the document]
quiet-node: Also looks line line 178, 203, 220, 289, 307, 325, 340 can be updated, correct?

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -148,7 +148,7 @@ export class MirrorNodeClient {
    */
   private readonly cacheService: CacheService;
 
-  static readonly EVM_ADDRESS_REGEX: RegExp = /\/accounts\/([\d\.]+)/;
+  static readonly EVM_ADDRESS_REGEX: RegExp = /\/accounts\/([\d.]+)/;
```
quiet-node: Is this change related to this PR's goal? The dot there would work with \ as well, correct?

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -557,11 +557,11 @@ export class SDKClient {
...
     try {
-      if (this.logger.isLevelEnabled('debug')) {
-        this.logger.debug(
-          `Get transaction record via consensus node: transactionId=${transactionId}, txConstructorName=${txConstructorName}`,
-        );
-      }
+      this.logger.debug(
+        'Get transaction record via consensus node: transactionId=%s, txConstructorName=%s',
+        transactionId,
+        txConstructorName,
```
quiet-node: Should lines 238, 491, 531 be updated?

quiet-node (commented):
Hello @belloibrahv, sorry for the late review and appreciate much your effort! Looking great so far but looks like there are still several places can drop the guards. I left them and along with other items down below. Please let me know what you think.

packages/relay/src/lib/dispatcher/rpcMethodDispatcher.ts
```diff
@@ -90,11 +88,11 @@ export class RpcMethodDispatcher {
     const methodParamSchemas = operationHandler[RPC_PARAM_VALIDATION_RULES_KEY];
 
     if (methodParamSchemas) {
-      if (this.logger.isLevelEnabled('info')) {
-        this.logger.info(
-          `Validating method parameters for ${rpcMethodName}, params: ${JSON.stringify(rpcMethodParams)}`,
-        );
-      }
+      this.logger.info(
```
simzzz: Should we revert this change? `JSON.stringify` can get slow for large objects

packages/relay/src/lib/services/ethService/contractService/ContractService.ts
simzzz: And 503

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -106,9 +106,7 @@ export class TransactionService implements ITransactionService {
     transactionIndex: string,
     requestDetails: RequestDetails,
   ): Promise<Transaction | null> {
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(`getTransactionByBlockHashAndIndex(hash=${blockHash}, index=${transactionIndex})`);
-    }
+    this.logger.trace('getTransactionByBlockHashAndIndex(hash=%s, index=%s)', blockHash, transactionIndex);
```
simzzz: I believe we can remove most of the other guards in this file as well except the one in `validateRawTransaction` and `sendRawTransactionErrorHandler`

simzzz (commented):
LG overall, left a few comments.

PR #4440, opened 2025-10-02
fix: fixes debug_traceBlockByNumber response to not be null for a transaction with no internal calls

### LINKED ISSUE #4386: Wrong response when calling debug_traceBlockByNumber and a transaction from the block doesn't contain any calls
### Description

When calling `debug_traceBlockByNumber` for a block, and some transaction from the block hasn't made any calls, then null would be returned for that transaction. For example:
```
{
    "method": "debug_traceBlockByNumber",
    "params": [
        "0x5067CB1",
        {
            "tracer":"callTracer"
        }
    ],
    "id": 1,
    "jsonrpc": "2.0"
}
```
returns
`{"result":[{"txHash":"0xa74c140aedec484c5c6cb559aecd75386d15cbafcc3e55c90d5302adbea86337","result":null}],"jsonrpc":"2.0","id":27}`

The result be the root transaction object with an empty calls array instead of null, for example:

```
{"result: 
  [{
    "txHash":"0xa74c140aedec484c5c6cb559aecd75386d15cbafcc3e55c90d5302adbea86337",
    "result":{
        "type": "CALL",
        "from": "0x...",
        "to": "0x...",
        "input": "0x",
        "output": "0x",
        "gas": "0x0",
        "gasUsed": "0x0",
        "value": "0x0",
        "calls": []
    }}],"jsonrpc":"2.0","id":27}
```

An example response from Quicknode for such a transaction (Geth/Alchemy also adhere):

Image

### Steps to reproduce

1. Try to call `debug_traceBlockByNumber` for a block containing such a transaction, for example `0x5067CB1`
2. Observe the result

### Additional context

_No response_

### Hedera network

_No response_

### Version

0.71.0

### Operating system

None

simzzz:
### Description

Ensures that `debug_traceBlockByNumber` and `debug_traceTransaction` with `callTracer` always return a `calls` field as an array, never `null` or `undefined`, even when transactions have no internal calls.

### Related issue(s)

Fixes #4386 

### Testing Guide

1. Call `debug_traceBlockByNumber` for a block that contains a transaction with no calls such as 0x522ba41 on mainnet
2. Make sure result is an object with an empty calls array instead of null

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/tests/lib/debug.spec.ts
+++ b/packages/relay/tests/lib/debug.spec.ts
@@ -418,8 +418,7 @@ describe('Debug API Test Suite', async function () {
 
           expect(result).to.deep.equal(expectedResult);
         });
-
-        it('Test call tracer with single action (no internal calls) returns empty calls array', async function () {
+        it('should return empty calls array when using callTracer with single action (no internal calls)', async function () {
           const singleActionResponse = {
             actions: [contractsResultsActionsResult.actions[0]], // Only the root action
           };
--- a/packages/relay/src/lib/debug.ts
+++ b/packages/relay/src/lib/debug.ts
@@ -459,7 +459,7 @@ export class DebugImpl implements Debug {
     transactionHash: string,
     tracerConfig: ICallTracerConfig,
     requestDetails: RequestDetails,
-  ): Promise<CallTracerResult | null> {
+  ): Promise<CallTracerResult> {
     try {
       const [actionsResponse, transactionsResponse] = await Promise.all([
         this.mirrorNodeClient.getContractsResultsActions(transactionHash, requestDetails),


quiet-node:
> @quiet-node hmm, I also can't reproduce it along with other examples I had written down. (perhaps it was a bug on MN side due to modularization, that is now fixed?). Nevertheless, I still think we should fix the response format for this particular case. I will try to find another example.

Yeah let's bring this up to a parking lot session and discuss as a team

packages/relay/tests/lib/debug.spec.ts
```diff
@@ -419,16 +419,31 @@ describe('Debug API Test Suite', async function () {
           expect(result).to.deep.equal(expectedResult);
         });
 
-        it('Should return empty array if no actions found', async function () {
-          restMock.onGet(CONTARCTS_RESULTS_ACTIONS).reply(200, JSON.stringify({ actions: [] }));
+        it('Test call tracer with single action (no internal calls) returns empty calls array', async function () {
```
konstantinabl: the format for it() statements is It('should return an empty call array...) etc. , please rename

quiet-node (approved):
LGTM

konstantinabl (approved):

PR #4471, opened 2025-10-14
feat: adds factory pattern for relay class, to support future redis client …

### LINKED ISSUE #4470: Refactor Relay initialization to use Static Factory Method pattern
### Problem
The `Relay` class has a two-step initialization pattern (`new Relay()` + `await relay.init()`) that is error-prone and leads to services being initialized with `undefined` Redis client and not propery ensure operator has balance

**Current flow:**
1. In server/index.ts we import relay from server/server.ts, which runs all the initialization code for the Relay class
2. Services within the class initialize with `undefined` Redis client (`CacheService`, `EthImpl`)
3. Later, `init()` connects Redis and sets `this.redisClient`
4. But services already have `undefined` - updating the field doesn't help

**Issues found:**
- Acceptance tests don't call `relay.init()` (discovered in PR review)
- Services get `undefined` Redis client during construction
- Easy to forget calling `init()` after construction
- Split initialization logic between constructor and `init()` method

### Proposed Solution
Implement the **Static Factory Method** pattern by:

1. Make constructor `private` and accept `redisClient` parameter
2. Create `static async init(logger, register): Promise<Relay>` method that:
   - Connects to Redis **first**, if redis is enable
   - Constructs Relay with connected Redis client
   - Performs operator balance check
   - Returns fully initialized instance

**New usage:**
```typescript
// Old (error-prone)
const relay = new Relay(logger, register);
await relay.init();

// New (foolproof)
const relay = await Relay.init(logger, register);
```

### Benefits
- **Enforces correct initialization** - Private constructor prevents `new Relay()`
-  **Services get working Redis client** - Connected before construction
- **Single initialization method** - All async setup in one place
- **Catches test bugs** - Tests must use proper initialization
- **Type-safe** - Returns `Promise<Relay>` with guarantees

### Implementation Checklist
- [ ] Update `Relay` class in `packages/relay/src/lib/relay.ts`
  - [ ] Make constructor private with `redisClient` parameter
  - [ ] Create static `init()` factory method
  - [ ] Move Redis connection logic into factory
  - [ ] Make `isRedisEnabled()` static
- [ ] Update `packages/server/src/index.ts` to use `Relay.init()`
- [ ] Update `packages/ws-server/src/index.ts` to use `Relay.init()`
- [ ] Update acceptance tests in `packages/server/tests/acceptance/`
- [ ] Update unit tests that instantiate Relay
- [ ] Update documentation/examples

### Breaking Change
This is a breaking change for any code that directly instantiates `Relay` with `new Relay()`. All consumers must update to use the async factory method.

---

**Related:** Addresses issue found in PR #4454 review

konstantinabl:
### Description

This PR introduces the usage of an init method for initialization of the Relay class instead of using the constructor. 
This allows us doing asynchronous operations, like creating a redis client on creation
### Related issue(s)

Fixes #4470 

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/config/methodConfiguration.ts
+++ b/packages/relay/src/lib/config/methodConfiguration.ts
@@ -188,6 +188,8 @@ export function getMethodConfiguration(): MethodRateLimitConfiguration {
 
 // Backwards compatibility: Export a Proxy that lazily calls getMethodConfiguration()
 // This allows existing code to use `methodConfiguration.eth_chainId` without changes
+// A Proxy is used for lazy evaluation to ensure test configuration overrides work correctly
+// (ConfigService is read at access time, not module load time).
 export const methodConfiguration = new Proxy({} as MethodRateLimitConfiguration, {
   get(target, prop) {
     return getMethodConfiguration()[prop as keyof MethodRateLimitConfiguration];


packages/server/tests/acceptance/index.spec.ts
```diff
@@ -94,7 +94,7 @@ describe('RPC Server Acceptance Tests', function () {
     logger.info(`E2E_RELAY_HOST: ${ConfigService.get('E2E_RELAY_HOST')}`);
 
     if (global.relayIsLocal) {
-      runLocalRelay();
+      await runLocalRelay();
```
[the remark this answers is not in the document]
konstantinabl: yes, true, fixed

packages/server/tests/integration/server.spec.ts
```diff
@@ -2572,97 +2575,89 @@ describe('RPC Server', function () {
...
       it('should execute with CallTracer type and valid CallTracerConfig', async () => {
-        expect(
-          await testClient.post('/', {
-            jsonrpc: '2.0',
-            method: 'debug_traceTransaction',
-            params: [contractHash1, { tracer: TracerType.CallTracer, tracerConfig: { onlyTopCall: true } }],
-            id: 1,
-          }),
-        ).to.not.throw;
+        const response = await testClient.post('/', {
```
quiet-node: Why these changes? Seems like doing the same thing to me?

packages/ws-server/src/index.ts
```diff
@@ -3,19 +3,21 @@
...
+    const { app, httpApp } = await initializeWsServer();
+
+    const host = ConfigService.get('SERVER_HOST');
+    app.listen({ port: constants.WEB_SOCKET_PORT, host });
+    httpApp.listen({ port: constants.WEB_SOCKET_HTTP_PORT, host });
   } catch (error) {
-    logger.fatal(error);
+    // If initialization fails, logger might not be available
+    console.error('Failed to initialize WebSocket server:', error);
     process.exit(1);
```
quiet-node: If that’s the case, we can move the logger outside the initializeWsServer() method so it’s not bound to it and will always be initialized. Same as how it's done in http. Let's try to keep thing consistent and less moving parts.

Moreover, we should keep this as logger.fatal instead of console.error for better visibility and filtering options in the observability dashboard.

packages/server/tests/integration/proxyHeaders.spec.ts
```diff
@@ -86,11 +88,21 @@ describe('Proxy Headers Integration Tests', function () {
...
-      },
-    });
+    try {
+      return await testClient.post('/', createRequestWithIP(id, ip), {
+        headers: {
+          'X-Forwarded-For': ip,
+        },
+      });
+    } catch (error: any) {
+      console.error('[TEST] Request failed:', {
```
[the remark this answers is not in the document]
konstantinabl: fixed, its leftover, code i thought i deleted

packages/server/tests/integration/server.spec.ts
konstantinabl: There are actually a few things wrong with this assertion:

1. **Awaiting before expect() defeats the purpose**: When you await the promise before passing it to `expect()`, the promise has already resolved/rejected by the time the assertion runs.
 - If the promise **rejects** → The error is thrown immediately, and `expect()` never runs
 - If the promise **resolves** → You're checking if a plain response object "throws", which doesn't make sense
You can check for all these tests that even changing the assertion to "to.throw" will result in pass. Thats a huge regression we missed.
2. Using to.not.throw on a promise doesn't make sense in this test framework. For asynchronous code we use to.be.rejected or to.be.fulfilled. 
However, my current solution is also not optimal, because rejection will not be handled gracefully, i will change it to use chaining to handle error gracefully

packages/ws-server/tests/unit/webSocketServer.spec.ts
```diff
@@ -122,10 +128,26 @@ describe('webSocketServer websocket handling', () => {
     return ws;
   }
 
-  beforeEach((done) => {
-    mirrorStubServer = createMirrorStubServer();
```
konstantinabl: This is removed, because these are unit tests and we shouldn't get to the point of calling MN server at all. A thing like this could be done in integration tests

packages/ws-server/tests/unit/webSocketServer.spec.ts
```diff
@@ -24,42 +31,41 @@ async function httpGet(server: http.Server, path: string): Promise<{ status: num
...
-    mirrorStubServer.listen(5551, '127.0.0.1', () => {
-      server = http.createServer(httpApp.callback());
-      server.listen(0, '127.0.0.1', done);
+  let httpApp: any;
+  let mockRelay: any;
+  let relayInitStub: sinon.SinonStub;
+
+  beforeEach(async function () {
+    // Create a mock relay object
+    mockRelay = {
```
konstantinabl: Mocking relay, since these are unit tests testing the web socket server and relay isn't needed

packages/server/tests/integration/server.spec.ts
quiet-node: Nice! Thanks for the explanation!

packages/server/tests/integration/proxyHeaders.spec.ts
```diff
@@ -32,7 +33,7 @@ describe('Proxy Headers Integration Tests', function () {
 
   let testServer: Server;
   let testClient: AxiosInstance;
-  let app: Koa<Koa.DefaultState, Koa.DefaultContext>;
+  let app: Koa;
```
quiet-node: unused variable

packages/server/tests/integration/server.spec.ts
```diff
@@ -161,13 +162,12 @@ describe('RPC Server', function () {
...
     const CUSTOMIZE_HOST = '127.0.0.1';
     const configuredServer = app.listen({ port: CUSTOMIZE_PORT, host: CUSTOMIZE_HOST });
-
     return new Promise<void>((resolve, reject) => {
       configuredServer.on('listening', () => {
         const address = configuredServer.address();
 
         try {
-          expect(address).to.not.be.null;
+          expect(address).to.not.equal(null);
```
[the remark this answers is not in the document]
konstantinabl: Because, otherwise eslint complains about unused expression and you mentioned earlier you don't think we should disable it

konstantinabl:
> Should we also replace `relay = new Relay(logger, new Registry());` everywhere in the tests with `relay = await Relay.init(logger, new Registry());`?

@simzzz fixed

packages/ws-server/tests/unit/webSocketServer.spec.ts
konstantinabl: Because otherwise this is an integration and not a unit test
https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4471/files/a8ee86e9074044627f6fc8689f9143a81daa2055#r2436064075

packages/server/tests/integration/proxyHeaders.spec.ts
konstantinabl: fixed

packages/relay/src/lib/relay.ts
```diff
@@ -341,4 +341,28 @@ export class Relay {
...
+   * ```typescript
+   * const relay = await Relay.init(logger, register);
+   * ```
+   */
+  static async init(logger: Logger, register: Registry): Promise<Relay> {
+    // Create Relay instance
+    const relay = new Relay(logger, register);
+
+    // Check operator balance if not in read-only mode
+    await relay.ensureOperatorHasBalance();
```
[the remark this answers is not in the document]
konstantinabl: Help in what sense?

packages/server/tests/integration/server.spec.ts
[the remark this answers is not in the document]
quiet-node: The question is, were the ESLint warnings already there before the update?

If yes, we can leave them as they are and address them in separate PRs to avoid confusion during review and to keep unrelated changes out of this PR’s main goal. This will also help you stay focused on the core purpose of the PR rather than fixing unrelated issues.

If no, then I’d like to understand why the server refactor introduced these warnings and why they’re appearing now when they didn’t before.

packages/server/tests/integration/proxyHeaders.spec.ts
```diff
@@ -60,6 +61,8 @@ describe('Proxy Headers Integration Tests', function () {
         console.error(err);
       }
     });
+    // Clear the Prometheus registry to avoid conflicts with other test files
+    register.clear();
```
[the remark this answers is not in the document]
konstantinabl: Well essentially we are changing the way we initialize the server, we are not just importing a file now and statically loading. In this case we need to clear out the registry before the next file, because the registry is separate and not cleared on server close. Maybe it should be?

packages/relay/src/lib/relay.ts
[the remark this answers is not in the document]
simzzz: I was thinking about a startup race where the transaction pool might be used before Redis is ready, so I suggested awaiting a connect to guarantee readiness. On a second thought now, I don't think this will be necessary and it would only slow down startup . Resolving

PR #4499, opened 2025-10-23
docs: document Paymaster for gasless transactions

Reccetech:
Added Paymaster functionality for gasless transactions, including configuration details, use cases, and best practices.

Fixes: https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/4477

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,80 @@ Expected response:
     ]
 }
 ```
+---
+
+## Paymaster (Gasless Transactions)
+
+Starting in v0.71, the relay supports a Paymaster-style mode that allows operators to sponsor gas fees for selected users or contracts — enabling gasless transactions and simpler onboarding.
+
+### Overview (Paymaster)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -304,6 +304,80 @@ Expected response:
     ]
 }
 ```
+---
+
+## Paymaster (Gasless Transactions)
```
quiet-node: Since Paymaster is still in the experimental stage, could you please add a short disclaimer section to mention that?

quiet-node: Hey Keith sorry for the late response. So by experimental I meant that the Paymaster was implemented based on a POC and hasn’t gone through a full acceptance criteria review yet. It was built for a specific experimental use case requested previously, and only validated against successful (“happy path”) scenarios, not failure cases. So it might be unstable or prone to failure for now. That's why we should put a disclaimer for the section.

PR #4521, opened 2025-10-27
feat: adds RLP hex txn storage to tx pool service

### LINKED ISSUE #4500: Save the RLP decoded transaction in the storage of transaction pool instead of just transaction hash
### Problem

Currently, we're saving only the transaction hash in the transaction pool storage but for `txpool_*` related methods, we need the entire transaction object.

### Solution

Save the RLP decoded transaction in the transaction pool.

### Note

- We should use https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4454 as a base till it's not merged.
- Affected classes: `LocalPendingTransactionStorage` and `RedisPendingTransactionStorage`.

### Alternatives

_No response_

simzzz:
### Description

Currently, we're saving only the transaction hash in the transaction pool storage but for txpool_* related methods, we need the entire transaction object. This PR adds storage of the entire txn as an RLP hex in both LRU and Redis.

### Related issue(s)

Fixes #4500

### DIFF
--- a/packages/relay/src/lib/services/transactionPoolService/transactionPoolService.ts
+++ b/packages/relay/src/lib/services/transactionPoolService/transactionPoolService.ts
@@ -47,11 +47,18 @@ export class TransactionPoolService implements ITransactionPoolService {
    */
   async saveTransaction(address: string, tx: Transaction): Promise<void> {
     const addressLowerCased = address.toLowerCase();
-
     const rlpHex = tx.serialized;
-    await this.storage.addToList(addressLowerCased, rlpHex);
 
-    this.logger.debug({ address, rlpHex: rlpHex.substring(0, 20) + '...' }, 'Transaction saved to pool');
+    try {
+      await this.storage.addToList(addressLowerCased, rlpHex);
+      this.logger.debug({ address, rlpHex: rlpHex.substring(0, 20) + '...' }, 'Transaction saved to pool');
+    } catch (error) {
+      this.logger.error(
+        { address, error: (error as Error).message, rlpHex: rlpHex.substring(0, 20) + '...' },
+        'Failed to save transaction to pool',
+      );
+      throw error;
+    }
   }
 
   /**
@@ -65,9 +72,16 @@ export class TransactionPoolService implements ITransactionPoolService {
   async removeTransaction(address: string, rlpHex: string): Promise<void> {
     const addressLowerCased = address.toLowerCase();
 
-    await this.storage.removeFromList(addressLowerCased, rlpHex);
-
-    this.logger.debug({ address, rlpHex: rlpHex.substring(0, 20) + '...' }, 'Transaction removed from pool');
+    try {
+      await this.storage.removeFromList(addressLowerCased, rlpHex);
+      this.logger.debug({ address, rlpHex: rlpHex.substring(0, 20) + '...' }, 'Transaction removed from pool');
+    } catch (error) {
+      this.logger.error(
+        { address, error: (error as Error).message, rlpHex: rlpHex.substring(0, 20) + '...' },
+        'Failed to remove transaction from pool',
+      );
+      throw error;
+    }
   }
 
   /**
--- a/packages/relay/tests/lib/services/transactionPoolService/transactionPoolService.spec.ts
+++ b/packages/relay/tests/lib/services/transactionPoolService/transactionPoolService.spec.ts
@@ -68,9 +68,11 @@ describe('TransactionPoolService Test Suite', function () {
       expect(mockStorage.addToList.calledWith(testAddress.toLowerCase(), testRlpHex)).to.be.true;
     });
 
-    it('should propagate storage errors', async () => {
+    it('should log error and rethrow when storage fails', async () => {
       const storageError = new Error('Storage connection failed');
       mockStorage.addToList.rejects(storageError);
+
+      const loggerSpy = sinon.spy(transactionPoolService['logger'], 'error');
 
       try {
         await transactionPoolService.saveTransaction(testAddress, testTransaction);
@@ -81,6 +83,8 @@ describe('TransactionPoolService Test Suite', function () {
 
       expect(mockStorage.addToList.calledOnce).to.be.true;
       expect(mockStorage.addToList.calledWith(testAddress.toLowerCase(), testRlpHex)).to.be.true;
+      expect(loggerSpy.calledOnce).to.be.true;
+      expect(loggerSpy.firstCall.args[0]).to.have.property('error', 'Storage connection failed');
     });
   });
 
@@ -93,9 +97,11 @@ describe('TransactionPoolService Test Suite', function () {
       expect(mockStorage.removeFromList.calledOnceWith(testAddress.toLowerCase(), testRlpHex)).to.be.true;
     });
 
-    it('should propagate storage errors', async () => {
+    it('should log error and rethrow when storage fails', async () => {
       const storageError = new Error('Storage removal failed');
       mockStorage.removeFromList.rejects(storageError);
+
+      const loggerSpy = sinon.spy(transactionPoolService['logger'], 'error');
 
       try {
         await transactionPoolService.removeTransaction(testAddress, testRlpHex);
@@ -105,6 +111,8 @@ describe('TransactionPoolService Test Suite', function () {
       }
 
       expect(mockStorage.removeFromList.calledOnceWith(testAddress.toLowerCase(), testRlpHex)).to.be.true;
+      expect(loggerSpy.calledOnce).to.be.true;
+      expect(loggerSpy.firstCall.args[0]).to.have.property('error', 'Storage removal failed');
     });
   });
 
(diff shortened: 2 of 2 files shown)


quiet-node:
Haven’t gone through this in depth yet, but it looks like we’re changing the design of the txPool implementation. To avoid adding new changes on top of moving parts, I suggest holding off on merging this into the feature branch until the feature branch is finalized and merged into main first.

packages/config-service/src/services/globalConfig.ts
```diff
@@ -614,6 +614,16 @@ const _CONFIG = {
     required: false,
     defaultValue: false,
   },
+  TRANSACTION_POOL_MGET_BATCH_SIZE: {
```
konstantinabl: Shouldn't we add these to configuration.md file

packages/relay/src/lib/services/transactionPoolService/LocalPendingTransactionStorage.ts
```diff
@@ -34,14 +42,19 @@ export class LocalPendingTransactionStorage implements PendingTransactionStorage
    * @param addr - The account address
    * @param txHash - The transaction hash to add to the pending list
    */
-  async addToList(addr: string, txHash: string): Promise<void> {
+  async addToList(addr: string, txHash: string, rlpHex?: string): Promise<void> {
```
konstantinabl: Why is rlpHex optional?

konstantinabl (dismissed):
A comment and a few questions

quiet-node (commented):
I think we should have a sync session to discuss this as a team

packages/relay/src/lib/services/transactionPoolService/LocalPendingTransactionStorage.ts
```diff
@@ -13,8 +13,16 @@ export class LocalPendingTransactionStorage implements PendingTransactionStorage
...
 
+  // Global set of all pending transaction hashes
+  private readonly globalTransactionIndex: Set<string>;
+
+  // Maps transaction hash to RLP hex payload
+  private readonly transactionPayloads: Map<string, string>;
+
   constructor() {
     this.pendingTransactions = new Map();
+    this.globalTransactionIndex = new Set();
```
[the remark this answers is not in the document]
quiet-node: Ha and also if we make `pendingTransactions` into  `Map{sender: [rlpHex]}`, this seems to help with the LockService to make it have access to the list of pending transactions which is a list of rlpHexes. They then can then be decoded by the LockService and reordered based on nonces. Pretty good point for the collaboration between LockService x TxPool.

packages/relay/src/lib/services/transactionPoolService/RedisPendingTransactionStorage.ts
```diff
@@ -38,46 +50,43 @@ export class RedisPendingTransactionStorage implements PendingTransactionStorage
...
-    // doing this to be able to atomically add the transaction hash
-    // and set the expiration time
-    await this.redisClient.multi().sAdd(key, txHash).expire(key, this.storageTtl).execAsPipeline();
+    await this.redisClient
+      .multi()
+      .sAdd(addressKey, rlpHex)
+      .expire(addressKey, this.storageTtl)
+      .sAdd(this.globalIndexKey, rlpHex)
+      .expire(this.globalIndexKey, this.storageTtl)
+      .exec();
```
quiet-node: The previous implementation had it as .execAsPipeline(), this implementation has .exec(). Why the update?

packages/relay/src/lib/services/transactionPoolService/RedisPendingTransactionStorage.ts
```diff
@@ -15,17 +16,28 @@ export class RedisPendingTransactionStorage implements PendingTransactionStorage
    */
   private readonly keyPrefix = 'txpool:pending:';
 
+  /**
+   * Key for the global pending transactions index.
+   */
+  private readonly globalIndexKey = 'txpool:pending:txns';
```
quiet-node: Should we make this a dependant of keyPrefix?

```suggestion
private readonly globalIndexKey = `${keyPrefix}txns`;
```

Also should we make it more clear? globalIndexKey is quite confusing. Could it be globalPendingTxsKey or something? and the value could be `${keyPrefix}global`?

quiet-node (commented):
Left some comments and questions

packages/relay/src/lib/services/transactionPoolService/RedisPendingTransactionStorage.ts
quiet-node: can you elaborate on your explanation as it doesn’t quite make sense to me. What exactly do you mean by “inconsistent state,” and what does it have to do with .exec() vs .execAsPipeline()?

As far as I understand, neither .exec() nor .execAsPipeline() guarantees rollback if one of the commands fails, right? Each command executes in order, but a failure in one doesn’t undo the others. So I’m not seeing how switching to .exec() here would prevent the situation you’re describing. 

Could you clarify what you mean and how it applies in this case?

packages/relay/src/lib/services/transactionPoolService/RedisPendingTransactionStorage.ts
simzzz: done

packages/relay/src/lib/services/transactionPoolService/RedisPendingTransactionStorage.ts
```diff
@@ -38,46 +50,43 @@ export class RedisPendingTransactionStorage implements PendingTransactionStorage
...
    * @param address - Account address whose pending list should be modified.
-   * @param txHash - Transaction hash to remove from the list.
+   * @param rlpHex - The RLP-encoded transaction as a hex string.
    */
-  async removeFromList(address: string, txHash: string): Promise<void> {
+  async removeFromList(address: string, rlpHex: string): Promise<void> {
     const key = this.keyFor(address);
-
-    await this.redisClient.sRem(key, txHash);
+    await this.redisClient.multi().sRem(key, rlpHex).sRem(this.globalIndexKey, rlpHex).execAsPipeline();
```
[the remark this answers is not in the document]
simzzz: fixed

packages/relay/src/lib/services/transactionPoolService/LocalPendingTransactionStorage.ts
```diff
@@ -70,5 +81,26 @@ export class LocalPendingTransactionStorage implements PendingTransactionStorage
...
+  /**
+   * Retrieves pending transaction payloads (RLP hex) for a specific address.
+   *
+   * @param address - The account address to query
+   * @returns Array of transaction RLP hex strings for the address
+   */
+  async getTransactionPayloads(address: string): Promise<string[]> {
+    const addressTransactions = this.pendingTransactions.get(address);
+    return addressTransactions ? Array.from(addressTransactions) : [];
   }
```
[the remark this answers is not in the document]
simzzz: Yes, I think it would be good to simplify the storage. I changed it

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -499,7 +499,7 @@ export class TransactionService implements ITransactionService {
 
     // Remove the transaction from the transaction pool after submission
     if (ConfigService.get('ENABLE_TX_POOL')) {
-      await this.transactionPoolService.removeTransaction(originalCallerAddress, parsedTx.hash!);
+      await this.transactionPoolService.removeTransaction(originalCallerAddress, parsedTx.serialized);
```
konstantinabl: Why do we switch to serialized?

packages/relay/src/lib/services/transactionPoolService/RedisPendingTransactionStorage.ts
quiet-node: Okay cool so .exec() is true atomic but .execAsPipeline is just a batch transaction. Okay that make sense thanks for the clarification.

packages/config-service/src/services/globalConfig.ts
```diff
@@ -614,6 +609,11 @@ const _CONFIG = {
     required: false,
     defaultValue: false,
   },
+  TRANSACTION_POOL_STORAGE_TTL_SECONDS: {
```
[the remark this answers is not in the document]
simzzz: I started working on this PR before Kosi introduced PENDING_TRANSACTION_STORAGE_TTL in the tx pool PR, so I implemented it with naming the variable TRANSACTION_POOL_STORAGE_TTL_SECONDS. Knowing we've already shipped PENDING_TRANSACTION_STORAGE_TTL makes it logical to revert it. Will fix

quiet-node (approved):
Coolio LGTM!

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
simzzz: We agreed to switch the unique identifier to the RLP payload. parsedTx.serialize are exactly the RLP-encoded signed transaction bytes

PR #4559, opened 2025-10-30
feat: adds `cache:` prefix to all CacheService keys

### LINKED ISSUE #4501: Add cache: prefix to Redis Cache class
## Problem

`CacheService` stores Redis keys without a prefix, while other services use prefixes (`pending:*`, `hbar-limit:*`, etc.). This causes:

1. **Unsafe clearing**: `CacheService.clear()` can't distinguish its keys from others, so we hardcode a filter:
   ```typescript
   const keysToDelete = allKeys.filter((key) => !key.startsWith('pending:'));
   ```
   This couples RedisCache to TransactionPoolService and doesn't scale.

2. **No ownership**: Can't tell which keys belong to which service.

3. **Maintenance burden**: Every new Redis user must update the filter.

## Solution

Add `cache:` prefix to all CacheService keys:

```typescript
// Before: eth_blockNumber
// After:  cache:eth_blockNumber
```

Then `clear()` can safely delete only `cache:*` keys without knowing about other services.

## Implementation

1. Add prefix constant and helper method in CacheService
2. Update all cache operations to prefix keys
3. Update `clear()` to only delete `cache:*` pattern
4. Remove hardcoded filter in RedisCache
5. Update tests

## Migration

⚠️ Existing cache will be invalidated (keys renamed). Acceptable since cache is ephemeral and will rebuild naturally.

## Acceptance Criteria

- [ ] All CacheService keys prefixed with `cache:`
- [ ] `clear()` only deletes `cache:*` keys
- [ ] No hardcoded filters in RedisCache
- [ ] Tests passing

simzzz:
### Description

`CacheService` stores Redis keys without a prefix, while other services use prefixes (`pending:*`, `hbar-limit:*`, etc.). This causes:

1. **Unsafe clearing**: `CacheService.clear()` can't distinguish its keys from others, so we hardcode a filter:
   ```typescript
   const keysToDelete = allKeys.filter((key) => !key.startsWith('pending:'));
   ```
   This couples RedisCache to TransactionPoolService and doesn't scale.

2. **No ownership**: Can't tell which keys belong to which service.

3. **Maintenance burden**: Every new Redis user must update the filter.

This PR solves that by adding the prefix to all CacheService keys. The `pending:` key was also changed to `txpool:pending`. All tests have been added to reflect that.

### Related issue(s)

Fixes #4501 

### Testing Guide

1. Call any method using cache
2. Check keys in Redis

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/clients/cache/localLRUCache.ts
+++ b/packages/relay/src/lib/clients/cache/localLRUCache.ts
@@ -16,6 +16,8 @@ type LRUCacheOptions = LRUCache.OptionsMaxLimit<string, any, unknown> & LRUCache
  * @implements {ICacheClient}
  */
 export class LocalLRUCache implements ICacheClient {
+  private static readonly CACHE_KEY_PREFIX = 'cache:';
+
   /**
    * Configurable options used when initializing the cache.
    *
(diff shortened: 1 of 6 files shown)


quiet-node (approved):
LGTM

PR #4574, opened 2025-11-04
fix: non existing from triggers validation error now

### LINKED ISSUE #4575: Failing pipelines on eth_calls with non-existing `from` address
### Description

Github actions fail on the tests containing eth_calls with non-existing account in from field. Those calls now result in validation error, and the tests still requires empty responses there...

### Steps to reproduce

Run the pipelines (ci/cd).

### Additional context

_No response_

### Hedera network

other

### Version

v0.1.0

### Operating system

None

arianejasuwienas:
### Description

The eth_call request will now trigger a validation error if the from field contains a non-existent account address. We need to update the tests to accept this new behavior.

### Related issue(s)

Fixes #4575

### Testing Guide

Github actions output check.


arianejasuwienas:
Closing it, we will fix the app behaviour, not test (tests are actually fine, checking for eth conformity ).

PR #4605, opened 2025-11-11
feat: create LockService class + relevant interfaces

### LINKED ISSUE #4592: Create LockService class
Create base lock service architecture with strategy pattern.

- Create interface for Lock Service
- Create interface for Lock Strategy
- Implement strategy pattern using Lock Strategy interface
- Define LockState interface with session key, mutex, timing properties

simzzz:
### Description

This PR creates a base lock service class with strategy pattern. It also includes: 
- Interface for Lock Service
- Interface for Lock Strategy
- Strategy pattern using Lock Strategy interface
- Defines LockState interface with session key, mutex, timing properties

### Related issue(s)

Fixes #4592 

### Testing Guide

Verify relevant classes and interfaces are implemented and align with the design doc. Since this subtask is implementing just the interfaces and the initial wrapper class, I believe it makes sense to include the actual test coverage once the strategies are implemented.

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/services/lockService/LockService.ts
+++ b/packages/relay/src/lib/services/lockService/LockService.ts
@@ -1,7 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
 
-import { Logger } from 'pino';
-
 import { LockStrategy } from '../../types/lock';
 
 /**
@@ -15,19 +13,12 @@ export class LockService {
   private readonly strategy: LockStrategy;
 
   /**
-   * Logger instance for lock service operations.
-   */
-  private readonly logger: Logger;
-
-  /**
    * Creates a new LockService instance.
    *
    * @param strategy - The lock strategy implementation to use.
-   * @param logger - Logger instance for tracking lock operations.
    */
-  constructor(strategy: LockStrategy, logger: Logger) {
+  constructor(strategy: LockStrategy) {
     this.strategy = strategy;
-    this.logger = logger.child({ name: 'lock-service' });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/relay/src/lib/services/lockService/LockService.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * Acquires a lock for the specified address.
+   * Blocks until the lock is available (no timeout on waiting).
+   *
+   * @param address - The sender address to acquire the lock for.
+   * @returns A promise that resolves to a unique session key.
+   */
+  async acquireLock(address: string): Promise<string> {
```
quiet-node: Would it make sense to rename it to lockId for more general use?

packages/relay/src/lib/types/lockService.ts
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { Mutex } from 'async-mutex';
+
+/**
+ * Represents the state of a lock for a specific address.
+ * Contains mutex, ownership information, and timing metadata.
+ */
+export interface LockState {
```
quiet-node: Looks like this will only be used in LocalLock only should we hold off and not add it to general lockService type file? We can have this in the local lock strategy class to avoid unnecessary indirection

quiet-node (commented):
Some questions and suggestions. Let me know what you think

simzzz:
@quiet-node changes are applied

packages/relay/src/lib/services/lockService/LockService.ts
quiet-node: Hm yeah back in the POC phase Fer wanted the class to be more generic so it can be used with other stuff if needed to. But you're right we should align with the design doc and we can always change it in future if needed.

PR #4610, opened 2025-11-13
feat: create `LocalLockStrategy` using LRU and async-mutex

### LINKED ISSUE #4593: Create Local Lock Strategy
Implement in-memory locking for single-instance deployments

- Implement LocalLockStrategy with async-mutex wrapper
- Add LRU cache for lock states (add number of maximum transactions)
- Implement acquireLock()
- Implement releaseLock()
- Add force release mechanism with 30s execution timer

natanasow:
### Description

Implement in-memory locking for single-instance deployments

- Implement LocalLockStrategy with async-mutex wrapper
- Add LRU cache for lock states (add number of maximum transactions)
- Implement acquireLock()
- Implement releaseLock()
- Add force release mechanism with 30s execution timer

### Related issue(s)

Fixes #4593 

### Testing Guide

1.
2.
3.

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
+++ b/packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
@@ -84,7 +84,8 @@ export class LocalLockStrategy {
    */
   async releaseLock(address: string, sessionKey: string): Promise<void> {
     if (this.logger.isLevelEnabled('debug')) {
-      this.logger.debug(`Releasing lock for address ${address} and session key ${sessionKey}.`);
+      const holdTime = Date.now() - state.acquiredAt!;
+      this.logger.debug(`Releasing lock for address ${address} and session key ${sessionKey} held for ${holdTime}ms.`);
     }
 
     const state = this.localLockStates.get(address);


packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
```diff
@@ -0,0 +1,168 @@
...
+  /**
+   * Maximum number of lock entries stored in memory.
+   * Prevents unbounded memory growth.
+   */
+  public static LOCAL_LOCK_MAX_ENTRIES: number = 1_000; // Max 1000 addresses
+
+  /**
+   * Time-to-live for each lock entry in the cache (in milliseconds).
+   */
+  public static LOCAL_LOCK_TTL: number = 300_000; // 5 minutes
```
konstantinabl: I think we discussed that if we have LOCAL_LOCK_MAX_ENTRIES we wont need this TTL

packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
```diff
@@ -0,0 +1,168 @@
...
+   * Retrieve an existing lock state for the given address, or create a new one if it doesn't exist.
+   *
+   * @param address - Unique identifier for the lock
+   * @returns The LockState object associated with the address
+   */
+  private getOrCreateState(address: string): LockState {
+    if (!this.localLockStates.has(address)) {
+      this.localLockStates.set(address, {
+        mutex: new Mutex(),
+        sessionKey: null,
```
konstantinabl: Why not pass the session key here when creating the state?

packages/relay/src/lib/services/lockService/LockStrategyFactory.ts
```diff
@@ -20,13 +21,13 @@ export class LockStrategyFactory {
    * @param logger - Logger instance for the lock strategy.
    * @returns A LockStrategy implementation.
    */
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+
   static create(redisClient: RedisClientType | undefined, logger: Logger): LockStrategy {
     // TODO: Remove placeholder errors once strategies are implemented
     if (redisClient) {
-      throw new Error('Redis lock strategy not yet implemented');
+      // throw new Error('Redis lock strategy not yet implemented');
```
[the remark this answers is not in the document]
natanasow: I think it's better to leave it as that because the next one who will implement Redis will have better visibility on where to hook in.

packages/relay/tests/lib/services/lockService/LocalLockStrategy.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+  beforeEach(() => {
+    lockStrategy = new LocalLockStrategy(pino({ level: 'silent' }));
+  });
+
+  afterEach(() => {
+    sinon.restore();
+  });
+
+  function getStateEntry(address) {
+    // @ts-ignore
```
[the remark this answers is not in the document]
natanasow: Because `localLockStates` is private, and if we don't want to add another method just for testing purposes, we should add `@ts-ignore` here.

packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * Retrieve an existing lock state for the given address, or create a new one if it doesn't exist.
+   *
+   * @param address - Unique identifier for the lock
+   * @returns The LockState object associated with the address
+   */
+  private getOrCreateState(address: string): LockState {
+    if (!this.localLockStates.has(address)) {
```
quiet-node: Should we have a simple private method to build the lockKey to normalize the id? This way we can make sure no issue with the case sensitivity and also we can use the same consistent key for both read and write?

packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
quiet-node: If we pass the sessionKey here and only set it when address doesn’t exist yet, then how would we assign the sessionKey when the address already exists in localLockStates? Keep in mind we never delete entries from localLockStates we only reset their fields.

So to set the sessionKey for an existing address, we’d have to assign it again anyway. That feels unnecessary, and I think assigning the sessionKey after mutex.acquire() as we currently do makes more sense.

packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
```diff
@@ -0,0 +1,162 @@
...
+import { Logger } from 'pino';
+
+/**
+ * Represents the internal state for a lock associated with a given address.
+ */
+interface LockState {
+  mutex: Mutex;
+  sessionKey: string | null;
+  acquiredAt: number | null;
+  maxLockTime: NodeJS.Timeout | null;
```
quiet-node: Should we rename this as lockTimeOutId for better clarity? It holds the id so we can remove it later, right?

```suggestion
lockTimeoutId: NodeJS.Timeout | null;
```

packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
```diff
@@ -0,0 +1,162 @@
...
+    const state = this.localLockStates.get(address);
+
+    // Ensure only the lock owner can release
+    if (state?.sessionKey !== sessionKey) {
+      return; // Not the owner — safely ignore
+    }
+
+    // Perform cleanup and release
+    await this.doRelease(state);
+  }
```
quiet-node: How about we do it like below so it's cleaner? We ignore if sessionKeys are not the same anyway.

```suggestion
// Ensure only the lock owner can release
if (state?.sessionKey === sessionKey) {
    await this.doRelease(state);
}
```

packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
```diff
@@ -0,0 +1,162 @@
...
+      return; // Already released or lock reassigned
+    }
+
+    if (this.logger.isLevelEnabled('debug')) {
+      const holdTime = Date.now() - state.acquiredAt!;
+      this.logger.debug(`Force releasing expired local lock for address ${address} held for ${holdTime}ms.`);
+    }
+
+    await this.doRelease(state);
+  }
```
quiet-node: Same here we can just care for state.sessionKey === sessionKey

```suggestion
if (state?.sessionKey === sessionKey) {
    await this.doRelease(state);

    if (this.logger.isLevelEnabled('debug')) {
      const holdTime = Date.now() - state.acquiredAt!;
      this.logger.debug(`Force releasing expired local lock for address ${address} held for ${holdTime}ms.`);
    }
}
```

quiet-node (changes requested):
Cool left couple of suggestions and comment, and one more thing we need to fix before merging.

*Update: seems like you reverted changes in TransactionService that's why my comment on that class is not showing.

packages/relay/src/lib/services/lockService/LockService.ts
```diff
@@ -29,6 +40,10 @@ export class LockService {
    * @returns A promise that resolves to a unique session key.
    */
   async acquireLock(address: string): Promise<string> {
+    if (this.logger.isLevelEnabled('debug')) {
+      this.logger.debug(`Acquiring lock for address ${address}.`);
```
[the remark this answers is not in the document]
natanasow: Fixed.

packages/relay/tests/lib/services/lockService/LocalLockStrategy.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+    const stateA = (lockStrategy as any).getOrCreateState('a');
+    const stateB = (lockStrategy as any).getOrCreateState('b');
+
+    expect(stateA).to.not.equal(stateB);
+  });
+
+  it('should clear timeout and reset state on release', async () => {
+    const address = 'test-reset';
+    const sessionKey = await lockStrategy.acquireLock(address);
+    const state = (lockStrategy as any).localLockStates.get(address);
```
konstantinabl: Apply bracket notation everywhere

PR #4623, opened 2025-11-18
feat: add cache factory, unify client interface, remove fallback (#4558)

### LINKED ISSUE #4558: Create factory for CacheClient
## Problem

Currently, we pass the redisClient to our CacheService, which may be undefined. Based on this value the CacheService decides if it needs to initialize LRUCache or RedisCache, however passing redisClient as undefined breaks the Interface segregation principle.

```typescript
class CacheService implements ICacheClient {
  //.....
  public constructor(
    logger: Logger,
    register: Registry = new Registry(),
    reservedKeys: Set<string> = new Set(),
    redisClient?: RedisClientType,
  ) {
    this.logger = logger;
    this.register = register;

    this.internalCache = new LocalLRUCache(logger.child({ name: 'localLRUCache' }), register, reservedKeys);
    this.sharedCache = this.internalCache;
    this.isSharedCacheEnabled = !ConfigService.get('TEST') && redisClient !== undefined;
    this.shouldMultiSet = ConfigService.get('MULTI_SET');

      this.sharedCache = new RedisCache(logger.child({ name: 'redisCache' }), register, redisClient!);
    }
    //.......
}
```
## Solution
Create a factory class creating either LRUCache or RedisCache and pass them to CacheService

```typescript
export class CacheFactory {
  static create(redisClient?: RedisClientType): CacheClient {
    return redisClient 
      ? new RedisCache(redisClient)
      : new LRUCache();
  }
}

// In Relay
const cache = CacheFactory.create(this.redisClient);
new CacheService(cache, ...)
```
## Benefits

- Encapsulates cache selection logic
- Cache service does not deal with creating cache classes
- Easier to add new cache types
- Stick to the interface segregation principle

jasuwienas:
### Description

1. The Cache Client creation process has been moved into a dedicated factory.
2. A single, unified client interface is now used instead of separate interfaces for each client (IRedisClient and IClient).
3. The fallback mechanism has been removed from the cache service, along with all conditions determining which client to use - the service now relies on a single client.
4. Most of the changes are in the tests. Since the cache service creation method has been updated, all test cases that instantiate it now follow the new logic.

### Related issue(s)

Fixes #4558 

### Testing Guide

1. Run all tests with debug mode on WITHOUT this tasks changes applied.
2. Make sure they all pass.

### Additional work needed

The Redis cache service still implements an additional, unnecessary interface and only measures logs and delegates calls to the underlying client. It will be removed entirely in [this](https://github.com/jasuwienas/hiero-json-rpc-relay/pull/1/files#diff-8cd9baaa5174bced654881e8fe6065c132dd4d22a48b6f9c7a3a2464fe8f09d7) PR.

### Checklist

### DIFF
--- a/packages/relay/tests/lib/clients/localLRUCache.spec.ts
+++ b/packages/relay/tests/lib/clients/localLRUCache.spec.ts
@@ -2,7 +2,6 @@
 
 import chai, { expect } from 'chai';
 import chaiAsPromised from 'chai-as-promised';
-import pino from 'pino';
 import { Registry } from 'prom-client';
 import sinon from 'sinon';
 
@@ -13,8 +12,11 @@ chai.use(chaiAsPromised);
 
 describe('LocalLRUCache Test Suite', async function () {
   this.timeout(10000);
-
-  const logger = pino({ level: 'silent' });
+  const logger = {
+    child: sinon.stub().returnsThis(),
+    trace: sinon.stub(),
+    isLevelEnabled: sinon.stub().returns(true),
+  };
   const registry = new Registry();
   const callingMethod = 'localLRUCacheTest';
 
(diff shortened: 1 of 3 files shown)


jasuwienas:
Dropping the fallback mechanism is a breaking change! During review, please take a look at:
_packages/relay/tests/lib/services/cacheService/cacheService.spec.ts_
to see what has changed

packages/relay/src/lib/clients/cache/redisCache/redisCache.ts
```diff
@@ -129,9 +122,11 @@ export class RedisCache implements IRedisCacheClient {
    *
    * @param keyValuePairs - An object where each property is a key and its value is the value to be cached.
    * @param callingMethod - The name of the calling method.
+   * @param [ttl] - The time-to-live (expiration) of the cache item in milliseconds. Used in fallback to pipelineSet.
    * @returns A Promise that resolves when the values are cached.
    */
-  async multiSet(keyValuePairs: Record<string, any>, callingMethod: string): Promise<void> {
+  async multiSet(keyValuePairs: Record<string, any>, callingMethod: string, ttl?: number): Promise<void> {
+    if (!this.options.multiSetEnabled) return this.pipelineSet(keyValuePairs, callingMethod, ttl);
```
quiet-node: Wasn't this logic already handled at cacheService.ts layer? Also how does this relate to the goal of the PR, which is about add cache factory, unify interface and remove fallback?

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -278,31 +181,16 @@ export class CacheService {
...
-        this.logger.error(redisError, `Error occurred while incrementing cache in Redis.`);
-      }
+    if (this.isSharedCacheEnabled) {
+      this.cacheMethodsCounter
+        .labels(callingMethod, CacheService.cacheTypes.REDIS, CacheService.methods.INCR_BY)
+        .inc(1);
+    } else {
+      this.cacheMethodsCounter.labels(callingMethod, CacheService.cacheTypes.LRU, CacheService.methods.GET).inc(1);
+      this.cacheMethodsCounter.labels(callingMethod, CacheService.cacheTypes.LRU, CacheService.methods.SET).inc(1);
     }
```
quiet-node: Duplicated?

packages/relay/src/lib/services/cacheService/cacheService.ts
```diff
@@ -161,66 +112,40 @@ export class CacheService {
   private async getFromInternalCache(key: string, callingMethod: string): Promise<any> {
     this.cacheMethodsCounter.labels(callingMethod, CacheService.cacheTypes.LRU, CacheService.methods.GET).inc(1);
 
-    return await this.internalCache.get(key, callingMethod);
+    return await this.client.get(key, callingMethod);
```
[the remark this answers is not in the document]
jasuwienas: Not to mention, I’d like to remove getAsync altogether and simply replace it with get, as proposed here:
https://github.com/jasuwienas/hiero-json-rpc-relay/pull/2/files

packages/relay/src/lib/services/cacheService/cacheService.ts
jasuwienas: Then we could remove this config from the PR:
https://github.com/jasuwienas/hiero-json-rpc-relay/pull/1/files#diff-a4de9b026ae2f3a282e86029a29dd350918077bc420139fa3b8dc35411bcc458R27
…and get rid of all the prometheus mapping logic there.

quiet-node (approved):
LGTM great work!

packages/relay/src/lib/clients/cache/localLRUCache.ts
```diff
@@ -280,6 +282,70 @@ export class LocalLRUCache implements ICacheClient {
...
 
+  /**
+   * Increments a value in the cache.
+   *
+   * @param key The key to increment
+   * @param amount The amount to increment by
+   * @param callingMethod The name of the calling method
+   * @returns The value of the key after incrementing
+   */
+  public async incrBy(key: string, amount: number, callingMethod: string): Promise<number> {
```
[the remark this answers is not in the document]
jasuwienas: This logic isn’t new, this code was previously used as a fallback when Redis failed (in CacheService). I moved it here to maintain a consistent interface across all clients. Removing it might be a good idea, but we should probably handle that in a separate PR. What do you think? @konstantinabl

PR #4642, opened 2025-11-24
feat: create acceptance tests nonce ordering

### LINKED ISSUE #4596: Create acceptance tests

konstantinabl:
### Description

This PR adds comprehensive acceptance tests in sendRawTransactionExtension.spec.ts to validate the lock service integration for nonce ordering in eth_sendRawTransaction.

#### Test Coverage

- FIFO ordering: 3 and 10 rapid transactions from same sender process in arrival order
- Parallel processing: Transactions from different senders don't block each other
- Lock release: Lock properly releases on validation errors (gas price too low)
- Async/Sync modes: Both processing modes maintain correct nonce ordering
- Mixed load: Multiple senders with multiple transactions each

### Related issue(s)

Fixes #4596 

### Testing Guide

1.
2.
3.

Not all of the accpetance criteria scenarios are covered with tests since some of them touch upon internal workings and not user experience

N/A

### Checklist

### DIFF
--- a/packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
+++ b/packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
@@ -433,25 +433,6 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
       expect(result2.nonce).to.equal(startNonce + 1);
     });
 
-    it('should handle rapid burst in async mode without nonce errors', async function () {
-      const sender = accounts[1];
-      const startNonce = await relay.getAccountNonce(sender.address);
-      const gasPrice = await relay.gasPrice();
-
-      // Send 5 transactions in rapid succession
-      const txHashes = await Promise.all(sendTransactionWithoutWaiting(sender, startNonce, 5, gasPrice));
-
-      // All should return tx hashes immediately
-      expect(txHashes).to.have.length(5);
-      txHashes.forEach((hash) => expect(hash).to.exist);
-
-      // All should eventually succeed
-      for (const txHash of txHashes) {
-        const receipt = await relay.pollForValidTransactionReceipt(txHash);
-        expect(receipt.status).to.equal('0x1');
-      }
-    });
-
     withOverriddenEnvsInMochaTest({ USE_ASYNC_TX_PROCESSING: false }, () => {
       it('should release lock after full processing in sync mode', async function () {
         const sender = accounts[0];


packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -297,4 +308,273 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
       expect(info.result).to.equal('INSUFFICIENT_TX_FEE');
     });
   });
+
+  describe('@nonce-ordering Lock Service Tests', function () {
+    this.timeout(240 * 1000); // 240 seconds
+    overrideEnvsInMochaDescribe({ ENABLE_NONCE_ORDERING: true });
+    const sendTransactionWithoutWaiting = async (signer: any, nonce: number, gasPrice: number) => {
```
simzzz: `gasPrice` remains unused here

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -41,10 +51,11 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
 
   this.beforeAll(async () => {
     const initialAccount: AliasAccount = global.accounts[0];
-    const neededAccounts: number = 3;
+    const neededAccounts: number = 5;
     accounts.push(
       ...(await Utils.createMultipleAliasAccounts(mirrorNode, initialAccount, neededAccounts, initialBalance)),
     );
+    console.log('accounts', accounts);
```
quiet-node: left over right?

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -297,4 +308,273 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
...
+      finalNonces.forEach((nonce, i) => {
+        expect(nonce).to.equal(startNonces[i] + 1);
+      });
+
+      // Submission should be fast (not blocking each other)
+      // Even with network latency, parallel submission should be < 5 seconds
+      expect(submitTime).to.be.lessThan(5000);
+    });
+
+    it('should handle mixed load: 5 txs each from 3 different senders', async function () {
```
quiet-node: Same for this we can use this test which technically covers `should process three transactions from different senders concurrently` unit, right?

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -297,4 +308,273 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
...
+        value: ONE_TINYBAR,
+        nonce: startNonce + 1, // This nonce is ahead of the account's actual nonce
+      };
+      const signedSecondTx = await sender.wallet.signTransaction(secondTx);
+
+      // Submit both transactions immediately to test lock release
+      const invalidTxPromise = relay.call('eth_sendRawTransaction', [signedInvalidTx]).catch((error: any) => error);
+
+      // Small delay to ensure first tx acquires lock first
+      await new Promise((resolve) => setTimeout(resolve, 10));
```
quiet-node: Do we need this timer here? As long as the submitted transactions are from the same sender, it will always allow the first one to acquire lock first, right?

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -297,4 +308,273 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
...
+
+      const secondTxPromise = relay.sendRawTransaction(signedSecondTx);
+
+      // Wait for both to complete
+      const [invalidResult, txHash] = await Promise.all([invalidTxPromise, secondTxPromise]);
+
+      // Verify first tx failed with validation error
+      expect(invalidResult).to.be.instanceOf(Error);
+      expect(invalidResult.message).to.include('gas price');
+
```
quiet-node: Should we override the env config, USE_TX_ASYNC_PROCESS to false here so no matter what value set for this config in .env, this test will pass with this assertion?

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -297,4 +308,273 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
...
+      // Verify lock was released (second tx was allowed to proceed)
+      expect(txHash).to.exist;
+
+      // Wait for second tx to be processed
+      await new Promise((r) => setTimeout(r, 2100));
```
quiet-node: Should we be consistent and use relay.pollForValidTransactionReceipt(txHash);?

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -297,4 +308,273 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
...
+      // Submit both transactions immediately to test lock release
+      const invalidTxPromise = relay.call('eth_sendRawTransaction', [signedInvalidTx]).catch((error: any) => error);
+
+      // Small delay to ensure first tx acquires lock first
+      await new Promise((resolve) => setTimeout(resolve, 10));
+
+      const secondTxPromise = relay.sendRawTransaction(signedSecondTx);
+
+      // Wait for both to complete
+      const [invalidResult, txHash] = await Promise.all([invalidTxPromise, secondTxPromise]);
```
quiet-node: so if USE_ASYNC_TX_PROCESSING is false like in localAcceptance.env, then wouldn't txHash in this case also an error object with wrong nonce status, but not a string txHash, right? Because later on I see you use this txHash as a param for MN call, which wouldn't be accurate.

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -297,4 +308,273 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
...
+        // Send two transactions - in sync mode, first must complete before second starts
+        const tx1Hash = await sendTransactionWithoutWaiting(sender, startNonce, gasPrice);
+        const tx1Time = Date.now() - startTime;
+
+        const tx2Hash = await sendTransactionWithoutWaiting(sender, startNonce + 1, gasPrice);
+        const tx2Time = Date.now() - startTime;
+
+        // In sync mode, tx2 should start after tx1 completes
+        expect(tx2Time).to.be.greaterThan(tx1Time);
+
```
quiet-node: What is the intended purpose of this block here? Since `const tx1Hash = await sendTransactionWithoutWaiting(sender, startNonce, gasPrice);` uses `await`, the execution will naturally pause until the request completes before proceeding to the next line. The same applies to `const tx2Hash = await sendTransactionWithoutWaiting(sender, startNonce + 1, gasPrice);`. As a result, `tx2Time` will always be greater than `tx1Time`. Given this deterministic behavior, what value does this test provide, considering the outcome is inherently guaranteed?

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
konstantinabl: yeah you are right, removed

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -297,4 +308,273 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
...
+      // Second tx should result in WRONG_NONCE (filtered out by mirror node)
+      // Similar to rpc_batch1.spec.ts pattern - check for 404
+      try {
+        await mirrorNode.get(`/contracts/results/${txHash}`);
+        expect.fail('Expected transaction to not be found (WRONG_NONCE)');
+      } catch (error: any) {
+        expect(error.response.status).to.equal(404);
+      }
+
```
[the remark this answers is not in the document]
konstantinabl: Yes, i removed the try/catch

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -297,4 +308,273 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
...
+        to: accounts[2].address,
+        value: ONE_TINYBAR,
+        nonce,
+        maxPriorityFeePerGas: gasPrice,
+        maxFeePerGas: gasPrice,
+      };
+      const signedTx = await signer.wallet.signTransaction(tx);
+      return relay.sendRawTransaction(signedTx);
+    };
+
```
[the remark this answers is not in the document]
konstantinabl: edited it, now its cleaner

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -297,4 +308,273 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
...
+      expect(finalNonce).to.equal(startNonce + 3);
+
+      // Verify transactions were processed in order by checking mirror node
+      for (let i = 0; i < 3; i++) {
+        const result = await mirrorNode.get(`/contracts/results/${txHashes[i]}`);
+        expect(result.nonce).to.equal(startNonce + i);
+      }
+    });
+
+    it('should handle rapid burst of 10 transactions from same sender', async function () {
```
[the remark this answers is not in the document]
konstantinabl: Honestly, i thought we may have on ensuring fifo order and one for stress testing with more transactions, but the stress test one would also test FIFO order so i will merge them

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
konstantinabl: Yeah, in this case the first one idea was to check if we process them correctly and fast withoug blocking the second one is stress testing to ensure locking works and keeps order, should i merge these?

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
konstantinabl: On second thought after running the tests, no we shouldnt poll, because then the test will be stuck. We filter out wrong_nonce error so currently the receipt wont be able to be retrieved and the polling will execute "forever". I addded the timeout so we just have some time until data is exported to MN

simzzz (approved):

natanasow (approved):

PR #4650, opened 2025-11-26
feat: review the 200 and 400 http code

### LINKED ISSUE #4630: Review the 200 and 400 http error codes and make them configurable
### Problem

When there is a consensus node error, for a wrong nonce, or similar, we are raising 400 instead of 200, which is the de facto standard (there is no standard for this).

We need to check which functional errors are generating 400 vs 200, compare them with Alchemy/QuickNode, and make the HTTP code configurable to allow relay providers to choose to be a de facto or no de facto standard.

### Solution

- list all errors
- add a config var for selected HTTP code
- return the HTTP code (200 or 400) based on the config var

### Alternatives

_No response_

jasuwienas:
### Description

Compliance layer checking if the api response format is correct - fixing it when needed.

### Related issue(s)

Fixes #4630

### Testing Guide

1. Run tests
2. Check if pass

N/A

### Checklist


jasuwienas:
I should gain access to run GitHub Actions soon. I’ll remove the draft status afterward. It would be good to run them just to make sure all the tests still work correctly with the default configuration.

jasuwienas:
Tests were prepared in accordance with the AC described above.

simzzz (approved):
LGTM

.github/workflows/acceptance.yml
```diff
@@ -38,10 +38,15 @@ jobs:
           - { name: 'Websocket Batch 3', testfilter: 'ws_batch3', test_ws_server: true }
           - { name: 'Cache Service', testfilter: 'cache-service' }
           - { name: 'Server Config', testfilter: 'serverconfig' }
+          - { name: 'Semantics and varying response status', testfilter: 'json_rpc_compliance' }
+          - name: "Semantics and default OK response status"
+            testfilter: "json_rpc_compliance"
+            on_valid_json_rpc_http_response_status_code: 200
```
quiet-node: Curious why we don't make it as an object to remain consistent with others?

jasuwienas: @quiet-node This actually is an object in YAML- just written in a different notation. I would’ve exceeded the 120, character line limit if I’d kept it on a single line. (other input params have shorter names...)

PR #4651, opened 2025-11-26
feat: fail when sending a transaction with non empty acess list

### LINKED ISSUE #4430: Fail fast when sending a transaction with a non-empty `accessList`
Transactions that send non-empty access lists should not be accepted by the Relay. That is, when a transaction sent with `eth_sendRawTransaction` contains a non-empty access list, the transaction should be treated as invalid and `eth_sendRawTransaction` should fail.

The CN does accept access lists in `EthereumTransacion`. However, given it is not possible at the moment to create them (via the equivalent of `eth_createAccessList`), it's better to reject them altogether.

---

_Original issue_

When sending a transaction with a non-empty `accessList`, the JSON-RPC Relay returns a transaction hash for which there is no valid receipt. This usually leads to timeout errors on the client side waiting for the transaction to be confirmed. For example sending a transaction with Foundry's `cast` _(replace `$PRIVATE_KEY` with your private key for the selected network)_

```sh
cast send -v --rpc-url https://testnet.hashio.io/api --private-key $PRIVATE_KEY --gas-limit 100000 \
        --access-list '[{
                "address": "0xa02457e5dfd32bda5fc7e1f1b008aa5979568150",
                "storageKeys": ["0x0000000000000000000000000000000000000000000000000000000000000081"]
        }]' 0x31074Ce044F5500bB4D729D6d92eb84fa908595C
```

it will send the transaction but it will timeout trying to get receipts.

The JSON-RPC Relay does not fully support access list. However, when sending such a transaction the Relay should either fail fast (with a 4xx error) or sending the transaction altogether but ignoring the `accessList` field.

### Additional Information

Upon log inspection, the transaction was actually never sent to the consensus node. The error was produced in the SDK

```
ERROR (relay-eth/61 on relay-node1-87f6bf79-dftrv): [Request ID: d351c8ff-32ba-4f7b-ba0a-c74b17dda609]
Failed to successfully submit sendRawTransaction: transaction={"type":2,"to":"0x31074Ce044F5500bB4D729D6d92eb84fa908595C","data":"0x","nonce":23,"gasLimit":"100000","gasPrice":null,"maxPriorityFeePerGas":"1","maxFeePerGas":"1420000000001","value":"0","chainId":"298","sig":{"_type":"signature","networkV":null,"r":"0x74df4b3128c0fdd376bbd72825059351542708748e803846e2e4146541b69d21","s":"0x2db253edebde114eb4c4b04e05b2300002ee0ec87eee590271161ded6314e8ce","v":28},"accessList":[{"address":"0xa02457E5Dfd32Bda5Fc7e1f1b008Aa5979568150","storageKeys":["0x0000000000000000000000000000000000000000000000000000000000000081"]}]}
    err: {
      "type": "TypeError",
      "message": "text.startsWith is not a function",
      "stack":
          TypeError: text.startsWith is not a function
              at Object.decode (/home/node/app/packages/relay/node_modules/@hashgraph/sdk/lib/encoding/hex.cjs:24:20)
              at /home/node/app/packages/relay/node_modules/@hashgraph/sdk/lib/EthereumTransactionDataEip1559.cjs:91:66
              at Array.map (<anonymous>)
              at EthereumTransactionDataEip1559.fromBytes (/home/node/app/packages/relay/node_modules/@hashgraph/sdk/lib/EthereumTransactionDataEip1559.cjs:91:53)
              at Cache._ethereumTransactionDataEip1559FromBytes (/home/node/app/packages/relay/node_modules/@hashgraph/sdk/lib/EthereumTransactionDataEip1559.cjs:134:99)
              at EthereumTransactionData.fromBytes (/home/node/app/packages/relay/node_modules/@hashgraph/sdk/lib/EthereumTransactionData.cjs:36:31)
              at SDKClient.<anonymous> (/home/node/app/packages/relay/dist/lib/clients/sdkClient.js:67:75)
              at /home/node/app/packages/relay/dist/lib/clients/sdkClient.js:9:71
              at new Promise (<anonymous>)
    }
```

The error comes from here when trying to send an EIP1559 transaction

https://github.com/hiero-ledger/hiero-sdk-js/blob/559db5e88a9a2eb4d513eb26481f8a58d2797889/src/EthereumTransactionDataEip1559.js#L86-L88

jasuwienas:
### Description

Failing the method sendRawTransaction call with non-empty access list

### Related issue(s)

Fixes #4430

### Testing Guide

1. Run conformity tests or check manually

To check manually:
```
npx hedera restart -d --verbose=trace
docker stop json-rpc-relay json-rpc-relay-ws
npm run build && npm run start
```
THis should work:
```
cast send -v --rpc-url http://127.0.0.1:7546 --private-key [PK] --gas-limit 100000 \
0x31074Ce044F5500bB4D729D6d92eb84fa908595C
```

This should fail:

```
─ cast send -v --rpc-url http://127.0.0.1:7546 --private-key [PK] --gas-limit 100000 \
        --access-list '[{                 
                "address": "0xa02457e5dfd32bda5fc7e1f1b008aa5979568150",
                "storageKeys": ["0x0000000000000000000000000000000000000000000000000000000000000081"]
        }]' 0x31074Ce044F5500bB4D729D6d92eb84fa908595
```

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/precheck.ts
+++ b/packages/relay/src/lib/precheck.ts
@@ -220,7 +220,7 @@ export class Precheck {
    * @param tx - The transaction.
    */
   accessList(tx: Transaction): void {
-    if (tx.accessList && tx.accessList.length > 0) throw predefined.NOT_YET_IMPLEMENTED;
+    if (tx.accessList?.length) throw predefined.NOT_YET_IMPLEMENTED;
   }
 
   /**


packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -97,6 +97,30 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
       });
     });
 
+    describe('accessList', function () {
```
konstantinabl: Should we add a unit test in the precheck.spec.ts, wdyt?

acuarica: +1

acuarica (commented):
lg, left some small comments

packages/server/tests/acceptance/sendRawTransactionExtension.spec.ts
jasuwienas: > Moreover, can we add another test case where the access list is an empty array?

We send an empty array in all of our sendRawTransaction tests :) .

But yeah, not a bad idea to add the dedicated one, so we can see the difference right away (thanks!). Just commited it, recheck please. @acuarica

packages/relay/src/lib/precheck.ts
```diff
@@ -214,6 +215,14 @@ export class Precheck {
     }
   }
 
+  /**
+   * Checks if the value of the access was not set.
+   * @param tx - The transaction.
+   */
+  accessList(tx: Transaction): void {
+    if (tx.accessList && tx.accessList.length > 0) throw predefined.NOT_YET_IMPLEMENTED;
```
natanasow: nit: a shorter version
```typescript
if (tx.accessList?.length) throw predefined.NOT_YET_IMPLEMENTED;
```

konstantinabl (approved):

PR #4678, opened 2025-12-03
feat: removed fallback predefined gas value for eth_estimateGas when failed to retrieve gas estimate from Mirror Node

### LINKED ISSUE #4459: Inconsistent Mirror Node responses for `contracts/call` when `value` field is omitted in HTS transfer estimation
### Description

This ticket tracks an inconsistency observed when calling `eth_estimateGas` through the Relay (`https://testnet.hashio.io/api`). The behavior changes depending on whether the `value` field is included in the request payload. This discrepancy originates from different responses returned by the Mirror Node, but results in inconsistent handling within the Relay.

---

### Steps to Reproduce

**Request without `value`:**

```bash
curl https://testnet.hashio.io/api/ \
  -X POST \
  --data '{
    "method": "eth_estimateGas",
    "params": [{
      "from": "0x743b4e1010671579CEC58f0932Bdac1f94F99847",
      "to": "0x0000000000000000000000000000000000644921",
      "data": "0xa9059cbb000000000000000000000000b99bdacec66c5cea3448ac561dfb498a26ffac150000000000000000000000000000000000000000000000000000000000000001"
    }],
    "id": 1,
    "jsonrpc": "2.0"
  }'
```

**Behavior:**

* The Relay receives a `BAD_REQUEST` from the Mirror Node with message `CONTRACT_REVERT_EXECUTED`.
* The Relay interprets this as a revert and returns a JSON-RPC error with code 3.

**Request with `value`:**

```bash
curl https://testnet.hashio.io/api/ \
  -X POST \
  --data '{
    "method": "eth_estimateGas",
    "params": [{
      "from": "0x743b4e1010671579CEC58f0932Bdac1f94F99847",
      "to": "0x0000000000000000000000000000000000644921",
      "data": "0xa9059cbb000000000000000000000000b99bdacec66c5cea3448ac561dfb498a26ffac150000000000000000000000000000000000000000000000000000000000000001",
      "value": "0x1"
    }],
    "id": 1,
    "jsonrpc": "2.0"
  }'
```

**Behavior:**

* The Mirror Node returns a different error message: `LOCAL_CALL_MODIFICATION_EXCEPTION`.
* The Relay falls back to `predefinedGasForTransaction()` and returns an estimated gas value instead of an error.

---

### Impact

The Relay produces inconsistent results for otherwise equivalent `eth_estimateGas` requests depending on the presence of the `value` field. Although this behavior stems from Mirror Node differences, it affects downstream clients relying on consistent gas estimation results.

---

### Tracking Actions

* [x] Track this issue until the Mirror Node resolves the equivalent issue https://github.com/hiero-ledger/hiero-mirror-node/issues/12156
* [ ] Consider normalizing handling within the Relay to ensure predictable behavior for `eth_estimateGas` calls regardless of Mirror Node message type.
* [ ] Re-test once the Mirror Node fix is confirmed.

quiet-node:
### Description

This PR fixes inconsistent error handling in `eth_estimateGas` by removing the fallback gas behavior that masked legitimate errors.

**Problem:** Previously, when the Mirror Node returned any error other than `CONTRACT_REVERT_EXECUTED`, the Relay would silently return a predefined fallback gas estimate. This hid real errors from users, who would then submit transactions that were guaranteed to fail on-chain.

**Solution:** This PR enables the Relay now properly propagates errors according to industry standards (aligned with Geth, Besu, and major providers):

- Contract reverts → code `3` (`CONTRACT_REVERT`)
- 400 errors (non-revert, validation failures) → code `-32000` (`COULD_NOT_SIMULATE_TRANSACTION`)
- 429, 5xx errors → re-thrown to preserve original error context and handled at dispatch level

### Related issue(s)

Fixes #4459

### Testing Guide

1. **Contract revert scenario:**
   - Call `eth_estimateGas` on a contract function that reverts
   - Expected: Error with code `3` and revert message

2. **Non-existent sender account:**
   - Call `eth_estimateGas` with a `from` address that doesn't exist on the network
   - Expected: Error with code `-32000` and message containing "Error occurred during transaction simulation"

3. **Valid gas estimation:**
   - Call `eth_estimateGas` with valid from/to addresses and valid contract call data
   - Expected: Returns gas estimate as before

- Combined MirrorNodeClientError isContractReverted() & isContractRevertOpcodeExecuted() into isContractRevert(), make sure it checks for CONTRACT_REVERT_EXECUTED message explicitly
- With the new logic of CONTRACT_REVERT for eth_estimateGas, eth_call errors are also updated to align:
   - If truly CONTRACT_REVERT -> return code 3
   - If 400 from MN -> return code -32000
   - If 429, 5xx -> return re-thrown to preserve original error context and handled at dispatch level

N/A

### Checklist

### DIFF
--- a/packages/relay/tests/lib/eth/eth_estimateGas.spec.ts
+++ b/packages/relay/tests/lib/eth/eth_estimateGas.spec.ts
@@ -582,21 +582,6 @@ describe('@ethEstimateGas Estimate Gas spec', async function () {
     contractService.estimateGas = originalEstimateGas;
   });
 
-  it('should handle mirror node response with no gas estimate', async function () {
-    const originalEstimateGas = contractService.estimateGas;
-    contractService.estimateGas = async () => {
-      return predefined.INTERNAL_ERROR('Test error for estimateGas');
-    };
-
-    const result = await ethImpl.estimateGas(transaction, null, requestDetails);
-
-    expect(result).to.be.an('error');
-    expect((result as JsonRpcError).code).to.equal(-32603);
-    expect((result as JsonRpcError).message).to.contain('Test error for estimateGas');
-
-    contractService.estimateGas = originalEstimateGas;
-  });
-
   it('should accepts both input and data fields but copy value of input field to data field', async () => {
     const inputValue = 'input value';
     const dataValue = 'data value';


packages/relay/src/lib/errors/MirrorNodeClientError.ts
```diff
@@ -45,11 +44,7 @@ export class MirrorNodeClientError extends Error {
     return this.statusCode === MirrorNodeClientError.ErrorCodes.ECONNABORTED;
   }
 
-  public isContractReverted(): boolean {
-    return this.statusCode === MirrorNodeClientError.ErrorCodes.CONTRACT_REVERT_EXECUTED;
-  }
-
-  public isContractRevertOpcodeExecuted() {
```
quiet-node: *Note for reviewers:* These two are now consolidated into a single method, `isContractRevert()`, which explicitly checks for the error.message ===  MirrorNodeClientError.messages.CONTRACT_REVERT_EXECUTED instead of falling back to a 400 status code as before, which wasn't accurate since not all 400 errors indicate a contract revert.

packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
```diff
@@ -458,7 +458,7 @@ export class CommonService implements ICommonService {
       }
     }
 
-    throw predefined.COULD_NOT_ESTIMATE_GAS_PRICE;
+    throw predefined.INTERNAL_ERROR('Failed to retrieve gas price from network fees');
```
quiet-node: *Note for reviewers:* Previously, this was thrown as `COULD_NOT_ESTIMATE_GAS_PRICE`, which was inaccurate since this method is used for `eth_gasPrice`, not estimation. I’ve updated it to use `INTERNAL_ERROR` instead, as reaching this point indicates an unexpected code issue rather than a gas-related problem from the Mirror Node.

PR #4679, opened 2025-12-04
feat: replace `isLevelEnabled` guards with Pino interpolation values

### LINKED ISSUE #4080: Use "interpolation values" instead of string interpolation when logging

natanasow:
### Description

This PR implements a comprehensive logging performance optimization by replacing `isLevelEnabled` guards with Pino's interpolation values (printf-style formatting) across the codebase. The current implementation uses JavaScript string interpolation within conditional guards, which causes performance overhead by building log messages even when the corresponding log level is disabled.

**Key improvements:**
- **Performance**: Eliminates string building overhead
- **Readability**: Removes cluttered conditional logging statements, making code cleaner and more maintainable
- **Consistency**: Standardizes logging approach using `%s` for string interpolation and `%o` for object interpolation
- **RPC Method Cleanup**: Removes redundant trace logging

### Related issue(s)

Fixes #4080 

### Testing Guide

1.
2.
3.

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
+++ b/packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
@@ -226,17 +226,18 @@ export class CommonService implements ICommonService {
   ): Promise<any> {
     if (!returnLatest && this.blockTagIsLatestOrPending(blockNumberOrTagOrHash)) {
       this.logger.debug(
-        `Detected a contradiction between blockNumberOrTagOrHash and returnLatest. The request does not target the latest block, yet blockNumberOrTagOrHash representing latest or pending: returnLatest=${returnLatest}, blockNumberOrTagOrHash=${blockNumberOrTagOrHash}`,
+        `Detected a contradiction between blockNumberOrTagOrHash and returnLatest. The request does not target the latest block, yet blockNumberOrTagOrHash representing latest or pending: returnLatest=%s, blockNumberOrTagOrHash=%s`,
+        returnLatest,
+        blockNumberOrTagOrHash,
       );
       return null;
     }
 
     if (blockNumberOrTagOrHash === constants.EMPTY_HEX) {
-      if (this.logger.isLevelEnabled('debug')) {
-        this.logger.debug(
-          `Invalid input detected in getHistoricalBlockResponse(): blockNumberOrTagOrHash=${blockNumberOrTagOrHash}.`,
-        );
-      }
+      this.logger.debug(
+        `Invalid input detected in getHistoricalBlockResponse(): blockNumberOrTagOrHash=%s.`,
+        blockNumberOrTagOrHash,
+      );
       return null;
     }
 


jasuwienas (approved):
That's pretty straightforward! I wasn't sure whether we should use %s for an array, but it matches the current behavior where we inject the values directly into the string, so it's fine (`${['']}` === [''].toString()).

That said, these changes will **heavily** conflict with the work I did for the Redis refactor. (and which I didn't merge because of the newman issue): https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4623

Perhaps we can leave the updates to RedisCache.ts out of this PR and handle them in a separate task? wdyt?

simzzz (approved):

simzzz (approved):

packages/relay/src/lib/db/repositories/hbarLimiter/ipAddressHbarSpendingPlanRepository.ts
```diff
@@ -122,7 +118,7 @@ export class IPAddressHbarSpendingPlanRepository {
       ? `Removed IP address from HbarSpendingPlan with ID ${ipAddressSpendingPlan.planId}`
       : `Trying to remove an IP address, which is not linked to a spending plan`;
     if (this.logger.isLevelEnabled('trace')) {
```
quiet-node: Can potentially be removed

packages/relay/src/lib/services/ethService/blockService/BlockService.ts
```diff
@@ -229,7 +229,7 @@ export class BlockService implements IBlockService {
     requestDetails: RequestDetails,
   ): Promise<string | null> {
     if (this.logger.isLevelEnabled('trace')) {
```
quiet-node: can potentially be removed

packages/relay/src/lib/eth.ts
```diff
@@ -219,9 +219,7 @@ export class EthImpl implements Eth {
   @rpcMethod
   @rpcParamLayoutConfig(RPC_LAYOUT.REQUEST_DETAILS_ONLY)
   async blockNumber(requestDetails: RequestDetails): Promise<string> {
-    if (this.logger.isLevelEnabled('trace')) {
-      this.logger.trace(`blockNumber()`);
-    }
+    this.logger.trace(`blockNumber()`);
```
quiet-node: Looks like most of the logs in this file can be removed

quiet-node (commented):
Great work so far but I think many more guards can be removed or updated

simzzz (approved):

PR #4753, opened 2025-12-29
Ddeskov task

ddeskov-limechain:
The task was to optimize the time it took for `eth_getBlockByNumber` , `eth_getBlockByHash`, and `eth_getBlockReceipts` endpoints to resolve the response. There are 2 major optimizations implemented:
1. Requests to Mirror Node are running in parallel. 
2. Responses received form Mirror Node are handled with native NodeJS functions, instead of CPU intensive JavaScript ones (Lodash).

**NOTE:** The code was tested only by comparing the length of the responses on original code and current version for params:
`"params": [
        "0x392EF1B",
        true
    ]`


Ferparishuertas:
I do have several questions

- what is the overall repsonse time decrease?
- wha is the overall cpu lower down?

Functional questions

- Can you explain the page algorittm you build, the reasoning?
- Why the paging algorithm is configurable for the user?
- What are the risks taken with this algorithm?
- Why caching was removed?

Ferparishuertas:
Thanks for the detailed answer @ddeskov-limechain 
From where did you forked the repo? this paging algorithm is not part of the repo. Thats why i asked for algorithm details

Ferparishuertas:
Thanks for answer.

How do you ensure you are not loosing any transaction on a given slice, if the number of transactions in the response is max 25 by default and up to 100 if configured as parameter?

packages/relay/src/lib/clients/mirrorNodeClient.ts
```diff
@@ -148,12 +148,35 @@ export class MirrorNodeClient {
...
 
+  /**
+   * Circuit breaker properties for timestamp slicing
+   * @private
+   */
+  private sliceFailureCount = 0;
+  private lastSliceFailureTime = 0;
+  private readonly SLICE_FAILURE_THRESHOLD = 3;
+  private readonly SLICE_FAILURE_WINDOW_MS = 60000;
+
```
quiet-node: What's the rationale behind these numbers? Should we make them configurable?

PR #4813, opened 2026-01-19
feat: eth_getTransactionReceipt now throws -32003 on consensus errors

### LINKED ISSUE #4769: Map Ethereum eth_sendRawTransaction and eth_getTransactionReceipt experience [Phase 1]
**PROBLEM**

Ethereum users receive immediate responses for errors related to nonce or any issue that prevents a transaction from being forwarded to the network, thanks to the existence of the mempool/transaction pool.

On Hedera, we cannot raise these kinds of errors right away — we do so only at consensus time.

On the other hand, we *do* have custom errors related to throttles, which are new to Ethereum developers.

Both of these differences alter the usual EVM experience for users because:
• They will not receive an error immediately if the RPC has async transaction processing enabled
• When they query the transaction receipt, they may loop indefinitely as no receipt will ever appear

Given this, we have both a responsibility and a challenge: to provide as close to a 1:1 experience as possible with Ethereum.

**Solution**

#### 1. Sync Processing Mode (`async = false`)
If the relay is in sync processing mode:
• It waits for consensus
• It can return consensus-time errors directly in the `sendRawTransaction` response

This has been implemented in : "Ensure errors from consensus node are not silently swallowed"
[#4697](https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/4697)

#### 2. Async Processing Mode (`async = true`)
If the relay is in async mode:
• It does *not* wait for consensus
• It cannot return consensus-time errors in `sendRawTransaction`

**Proposal**: Introduce new logic in `getTransactionReceipt` that fails with a proper RPC error (instead of returning `null`) when we know the transaction has failed due to nonce issues or throttles. Use a custom RPC error code within the reserved range.º

-Review the error defintion proposal: -32003 Message = given error. 32003 is transaction rejected and the most meaningful
- Identify how to check if tx failed at consensus time in MN, and fail at rpc level. if it implies some extra MN calls, share with the team for discussion, and possible review with in MN
- Test with web3js, and ethers, to check if tx.wait and other scenarios are capturing the error

simzzz:
### Description

This PR implements transaction failure detection for `eth_getTransactionReceipt` when the relay operates in async processing mode (`USE_ASYNC_TX_PROCESSING=true`), solving the infinite polling issue when transactions fail at consensus.

**Problem:**
When a transaction fails at consensus (e.g., `WRONG_NONCE`, `THROTTLED_AT_CONSENSUS`, `CONSENSUS_GAS_EXHAUSTED`), the Mirror Node does not index a contract result, causing `eth_getTransactionReceipt` to return `null` indefinitely. This leads to infinite polling by Ethereum clients like ethers.js and web3.js.

**Solution:**
- Added a new feature flag `ENABLE_STANDARIZE_HEDERA_SPECIAL_CONSENSUS_ERRORS ` (disabled by default since it's a breaking change)
- When enabled, `eth_getTransactionReceipt` detects Hedera-specific validation failures and throws a JSON-RPC error `-32003` (Transaction rejected)
- Uses existing Mirror Node contract result response (no additional API calls)
- All consensus failure statuses are mapped to error code `-32003` with descriptive messages

**Implementation Details:**
- **Decision 1**: Uses Option A (Single Request) - no extra Mirror Node calls, works with existing `getContractResult` response
- **Decision 2**: Uses Option A (Unified Error) - all failures return `-32003` with specific status in the message
- Error message format: `"Transaction rejected: {STATUS}"` (e.g., `"Transaction rejected: WRONG_NONCE"`)

### Related issue(s)

Fixes #4769 

#### Prerequisites
- Build the project: `npm run build`
- Set environment variable: `ENABLE_STANDARIZE_HEDERA_SPECIAL_CONSENSUS_ERRORS =true`

#### Test 1: Verify feature is disabled by default
1. Start the relay without setting `ENABLE_STANDARIZE_HEDERA_SPECIAL_CONSENSUS_ERRORS `
2. Submit a transaction that will fail at consensus (e.g., with wrong nonce)
3. Call `eth_getTransactionReceipt` with the transaction hash
4. **Expected**: Returns the receipt normally (old behavior maintained for backward compatibility)

#### Test 2: Test WRONG_NONCE failure detection
1. Start the relay with `ENABLE_STANDARIZE_HEDERA_SPECIAL_CONSENSUS_ERRORS =true`
2. Submit a transaction with an incorrect nonce
3. Call `eth_getTransactionReceipt` with the transaction hash
4. **Expected**: Returns JSON-RPC error with code `-32003` and message containing `"WRONG_NONCE"`

#### Test 3: Test THROTTLED_AT_CONSENSUS failure detection
1. Start the relay with `ENABLE_STANDARIZE_HEDERA_SPECIAL_CONSENSUS_ERRORS =true`
2. Submit enough transactions to trigger throttling
3. Call `eth_getTransactionReceipt` for a throttled transaction
4. **Expected**: Returns JSON-RPC error with code `-32003` and message containing `"THROTTLED_AT_CONSENSUS"`

#### Test 4: Test successful transactions still work
1. Start the relay with `ENABLE_STANDARIZE_HEDERA_SPECIAL_CONSENSUS_ERRORS =true`
2. Submit a valid transaction that succeeds
3. Call `eth_getTransactionReceipt` with the transaction hash
4. **Expected**: Returns normal receipt with status `0x1`

#### Test 5: Run unit tests
bash
cd packages/relay
npm test -- --grep "eth_getTransactionReceipt"

**Expected**: All 19 tests pass, including 7 new "Transaction Failure Detection" tests

N/A

N/A 

### Checklist


Ferparishuertas (commented):
Its looking good.
I added some comments

packages/relay/tests/lib/eth/eth_getTransactionReceipt.spec.ts
```diff
@@ -372,4 +372,87 @@ describe('@ethGetTransactionReceipt eth_getTransactionReceipt tests', async func
     expect(receipt).to.exist;
     expect(receipt?.to).to.be.null;
   });
+
+  describe('Transaction Failure Detection', () => {
```
quiet-node: Would it possible to add some acceptance test and also add it to the @release suite as well?

Ferparishuertas (dismissed):
LGTM.  BTW Is suggest to add the acceptance tests, and include them in release

quiet-node (commented):
LG one item

Ferparishuertas (dismissed):
LGTM

quiet-node (approved):
LGTM nice work!

jasuwienas (approved):
lg!

konstantinabl (approved):

PR #4828, opened 2026-01-24
feat: optimize block retrieval with parallel timestamp slicing for mirror node api calls

### LINKED ISSUE #4806: Optimize block retrieval with parallel timestamp slicing for Mirror Node API calls
## Problem

Block retrieval for `eth_getBlockByNumber`, `eth_getBlockByHash`, and related endpoints becomes significantly slow when blocks contain thousands of transactions. The current implementation fetches data from the Mirror Node API sequentially, making multiple paginated requests (100 results per page) one after another. For blocks with 11,000+ transactions, this results in 45-50 second response times, creating poor user experience and potential timeout issues.

The sequential approach creates a network I/O bottleneck where each API call must wait for the previous one to complete, even though the Mirror Node can handle concurrent requests.

## Proposed Solution

This POC PR https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4753 introduces an efficicent parallel timestamp slicing strategy that significantly improves block retrieval performance.

**How it works:**

Instead of fetching transaction data sequentially (one page at a time), the approach divides the block's timestamp range into multiple non-overlapping time slices and fetches them all concurrently from the Mirror Node API. As results arrive, progressive deduplication merges them efficiently using Set-based lookups.

**Key features:**

- **Smart activation**: Only triggers for large blocks (configurable threshold)
- **Circuit breaker**: Automatically falls back to sequential fetching after repeated failures (TBD on if neccessary or overkill?)
- **Backward compatible**: Zero breaking changes, existing behavior preserved as fallback

**Impact:**

Response times improve from 45-50 seconds to 1-2 seconds for blocks with 11,000+ transactions, a 95%+ improvement while maintaining correctness.

quiet-node:
### Description

This PR introduces **parallel timestamp slicing** for Mirror Node API calls (originally proposed in [PR#4753](https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4753)), significantly reducing the latency of block retrieval in `eth_getBlockByNumber`, `eth_getBlockByHash`, `eth_getBlockReceipts`, and `eth_getLogs`.

By splitting the block's timestamp range into concurrent requests, we bypass the sequential pagination bottleneck inherent in the Mirror Node log retrieval process.

---

#### The Sequential Bottleneck

Retrieving large blocks (for example blocks with 10,000+ transactions) currently requires 100+ sequential API calls due to the Mirror Node's 100-record limit per response. This results in response times of ~14–15 seconds (best case) on production networks because each request must wait for the "next" link from the previous response.

#### The Parallel Approach

Instead of sequential pagination, this PR introduces **Parallel Timestamp Slicing**. We divide the block's timestamp range into $N$ non-overlapping segments and fetch them concurrently. This approach eliminates the waiting time between requests, allowing us to leverage the full throughput of the Mirror Node API.

### Key Technical Details

- **Dynamic Scaling**: The number of slices is calculated dynamically based on the total logs/transactions in the block, ensuring optimal resource allocation.
- **Sliding Window Concurrency**: Optimized request handling using `Promise.race()` to ensure maximum throughput with a configurable concurrency limit (default: 30).
- **Boundary Precision**: Uses `lt` (exclusive) for intermediate slices and `lte` (inclusive) for the final slice to ensure zero-gap coverage without duplicates.
- **Resilient Fallback**: Includes an automatic fallback to sequential pagination if timestamp metadata is missing, malformed, or if the slicing execution encounters recoverable API errors. This ensures performance when possible and stability always.

### Key Changes

- **`mirrorNodeClient.ts`**: Implemented slicing logic, concurrency management, and result merging.
- **`blockWorker.ts`**: Integrated slice calculation into the block retrieval workflow.
- **`CommonService.ts`**: Updated log retrieval signatures to support slicing parameters.
- **`mirrorNodeClient.spec.ts`**: Added comprehensive unit tests for boundary conditions and deduplication.

---

### Performance Impact

| Metric                            | Before     | After           | Improvement        |
| :-------------------------------- | :--------- | :-------------- | :----------------- |
| `eth_getBlockByNumber` (11k+ txs) | ~14,000 ms | ~3,500 ms       | **~75% reduction** |
| `eth_getBlockByHash` (11k+ txs)   | ~15,000 ms | ~3,500-4,000 ms | **~75% reduction** |
| `eth_getBlockReceipts` (11k+ txs)                     | ~10,000 ms | ~3,500 ms       | **~75% reduction** |
| `eth_getLogs` (w/ blockHash (11k+ txs))                     | ~10,000 ms | ~2,000 ms       | **~80% reduction** |

---

> [!important]
> In this PR, `parallel timestamp slicing` logic is intentionally applied only to synthetic (HAPI native) transactions and not to regular Ethereum contract result transactions. Ethereum transactions are comparatively slow and sparse, while high-volume blocks are typically ~99% synthetic; since slice counts are dynamically derived from total block transaction volume, applying this algorithm to Ethereum transaction retrieval would unnecessarily create hundreds of timestamp slices and issue excessive Mirror Node requests with little or no data returned. For this reason, Ethereum transaction retrieval remains sequential, and slicing is limited to synthetic transaction flows where it provides real performance benefits.

### Related issue(s)

Fixes #4806

### Testing Guide

1. Start the relay locally connected to Mainnet with this branch and with `main`.
2. Request a high-transaction block:
   ```bash
   curl -X POST http://localhost:7546 \
     -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x392ef1b", true],"id":1}'
   ```
3. Verify the response time of this branch is significantly lower than when run with `main`.

---

The work for eth_getLogs with block range has been isolated to another ticket because the team needs more discussion on the design. More information is reported here: https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/4830

### Checklist

### DIFF
--- a/packages/relay/src/lib/debug.ts
+++ b/packages/relay/src/lib/debug.ts
@@ -25,7 +25,7 @@ import {
   TransactionTracerConfig,
   TxHashToContractResultOrActionsMap,
 } from './types';
-import type { ContractAction, MirrorNodeContractResult } from './types/mirrorNode';
+import type { ContractAction, MirrorNodeBlock, MirrorNodeContractResult } from './types/mirrorNode';
 import { rpcParamValidationRules } from './validators';
 
 /**
@@ -640,20 +640,25 @@ export class DebugImpl implements Debug {
 
   /**
    * Retrieves all transaction hashes in a block (EVM + synthetic) along with pre-fetched data.
+   * Optimizes performance by using parallel timestamp slicing for large result sets.
    *
    * @private
-   * @param blockResponse - Block metadata with timestamp range
+   * @param blockResponse - Block metadata including timestamp range and transaction count
    * @param requestDetails - Request tracking details
    * @returns Object containing transaction hashes and pre-fetched data (contract results and actions)
    */
   private async getBlockTransactionDetails(
-    blockResponse: { timestamp: { from: string; to: string } },
+    blockResponse: MirrorNodeBlock,
     requestDetails: RequestDetails,
   ): Promise<{
     transactionHashes: string[];
     preFetchedData: TxHashToContractResultOrActionsMap;
   }> {
     const timestampRange = [`gte:${blockResponse.timestamp.from}`, `lte:${blockResponse.timestamp.to}`];
+
+    // Calculate slice count based on actual transaction count in the block
+    const maxLogsPerSlice = ConfigService.get('MIRROR_NODE_TIMESTAMP_SLICING_MAX_LOGS_PER_SLICE');
+    const sliceCount = Math.ceil(blockResponse.count / maxLogsPerSlice);
 
     // Fetch both contract results and all logs in the block in parallel
     const [contractResults, allLogs] = await Promise.all([
@@ -662,7 +667,7 @@ export class DebugImpl implements Debug {
         { timestamp: timestampRange },
         undefined,
       ]),
-      this.mirrorNodeClient.getContractResultsLogsWithRetry(requestDetails, 1, {
+      this.mirrorNodeClient.getContractResultsLogsWithRetry(requestDetails, sliceCount, {
         timestamp: timestampRange,
       }),
     ]);


jasuwienas (approved):
LGTM, great work!

PR #4847, opened 2026-01-28
feat: creates metrics service for lock service

### LINKED ISSUE #4712: Add relevant metrics for lock service
Add metrics that would allow for monitoring of the lock service's behaviour

- **relay_lock_wait_time_seconds - average wait time of a transaction to acquire lock**
- **relay_lock_hold_duration_seconds - average time a transaction holds a lock**
- **relay_wrong_nonce_errors_total - help business know how many such errors we have**
- **redis_queue_depth - amount of transactions waiting on average in redis**

konstantinabl:
### Description

Added metrics for the Lock Service and Transaction Pool Service to improve observability and monitoring capabilities.

### Related issue(s)

Fixes #4712 

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/relay.ts
+++ b/packages/relay/src/lib/relay.ts
@@ -22,7 +22,7 @@ import { RpcMethodDispatcher } from './dispatcher';
 import { EthImpl } from './eth';
 import { CacheClientFactory } from './factories/cacheClientFactory';
 import { NetImpl } from './net';
-import { LockService, LockStrategyFactory } from './services';
+import { LockService, LockStrategyFactory, TransactionPoolService } from './services';
 import HAPIService from './services/hapiService/hapiService';
 import { HbarLimitService } from './services/hbarLimitService';
 import MetricService from './services/metricService/metricService';
@@ -341,6 +341,11 @@ export class Relay {
     this.metricService = new MetricService(this.logger, metricsCollector, this.register, hbarLimitService);
 
     const storage = PendingTransactionStorageFactory.create(this.redisClient);
+    const transactionPoolService = new TransactionPoolService(
+      storage,
+      this.logger.child({ name: 'transaction-pool-service' }),
+      this.register,
+    );
 
     // Create Eth implementation with connected Redis client
     this.ethImpl = new EthImpl(
@@ -349,7 +354,7 @@ export class Relay {
       this.logger.child({ name: 'relay-eth' }),
       chainId,
       this.cacheService,
-      storage,
+      transactionPoolService,
       lockService,
       this.register,
     );
@@ -367,7 +372,7 @@ export class Relay {
       this.metricService.addExpenseAndCaptureMetrics(args);
     });
 
-    this.txpoolImpl = new TxPoolImpl(storage, this.logger.child({ name: 'relay-txpool' }), this.register);
+    this.txpoolImpl = new TxPoolImpl(transactionPoolService);
 
     // Create Debug and Admin implementations
     this.debugImpl = new DebugImpl(this.mirrorNodeClient, this.logger, this.cacheService);
--- a/packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
+++ b/packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
@@ -81,7 +81,7 @@ export class LocalLockStrategy implements LockStrategy {
     }
     const state = this.getOrCreateState(address);
 
-    this.lockMetricsService.incrementWaitingTxns('local');
+    this.lockMetricsService.incrementWaitingTxns(this.type);
 
     try {
       // Acquire the mutex (this will block until available)
@@ -90,7 +90,7 @@ export class LocalLockStrategy implements LockStrategy {
       const acquiredAt = process.hrtime.bigint();
       // Record wait time
       const waitTimeSeconds = Number(acquiredAt - startTime) / 1e9;
-      this.lockMetricsService.recordWaitTime('local', waitTimeSeconds);
+      this.lockMetricsService.recordWaitTime(this.type, waitTimeSeconds);
 
       // Record lock ownership metadata
       state.sessionKey = sessionKey;
@@ -101,15 +101,15 @@ export class LocalLockStrategy implements LockStrategy {
       }, ConfigService.get('LOCK_MAX_HOLD_MS'));
 
       // Record successful acquisition
-      this.lockMetricsService.recordAcquisition('local', 'success');
-      this.lockMetricsService.incrementActiveCount('local');
+      this.lockMetricsService.recordAcquisition(this.type, 'success');
+      this.lockMetricsService.incrementActiveCount(this.type);
 
       return { sessionKey, acquiredAt };
     } catch (error) {
-      this.lockMetricsService.recordAcquisition('local', 'fail');
+      this.lockMetricsService.recordAcquisition(this.type, 'fail');
       throw error;
     } finally {
-      this.lockMetricsService.decrementWaitingTxns('local');
+      this.lockMetricsService.decrementWaitingTxns(this.type);
     }
   }
 
@@ -118,9 +118,9 @@ export class LocalLockStrategy implements LockStrategy {
    *
    * @param address - The locked resource key
    * @param sessionKey - The session key of the lock holder
-   * @param acquiredAt - The timestamp when the lock was acquired (optional, uses internal state if not provided)
+   * @param acquiredAt - High-resolution timestamp (nanoseconds) when the lock was acquired
    */
-  async releaseLock(address: string, sessionKey: string, acquiredAt?: bigint): Promise<void> {
+  async releaseLock(address: string, sessionKey: string, acquiredAt: bigint): Promise<void> {
     const normalizedAddress = LockService.normalizeAddress(address);
     const state = this.localLockStates.get(normalizedAddress);
     if (state) {
@@ -128,11 +128,10 @@ export class LocalLockStrategy implements LockStrategy {
       if (state.sessionKey === sessionKey) {
         await this.doRelease(state);
 
-        // get hold time in seconds
-        const holdTimeNs = process.hrtime.bigint() - acquiredAt!;
+        const holdTimeNs = process.hrtime.bigint() - acquiredAt;
         const holdTimeMs = Number(holdTimeNs) / 1e6;
-        this.lockMetricsService.recordHoldDuration('local', Number(holdTimeNs) / 1e9);
-        this.lockMetricsService.decrementActiveCount('local');
+        this.lockMetricsService.recordHoldDuration(this.type, Number(holdTimeNs) / 1e9);
+        this.lockMetricsService.decrementActiveCount(this.type);
 
         if (this.logger.isLevelEnabled('debug')) {
           this.logger.debug(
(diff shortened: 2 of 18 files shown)


packages/relay/src/lib/services/lockService/LockService.ts
```diff
@@ -14,13 +15,16 @@ export class LockService {
...
+  private readonly strategyType: string;
+
   /**
    * Creates a new LockService instance.
    *
    * @param strategy - The lock strategy implementation to use.
    */
   constructor(strategy: LockStrategy) {
     this.strategy = strategy;
+    this.strategyType = strategy instanceof RedisLockStrategy ? 'redis' : 'local';
```
jasuwienas: With the current implementation, whenever we add a new lock strategy we have to remember to update this line. It would be better to store this value within the LockStrategy itself and depend on the abstraction here.
```
this.strategyType = strategy.type;
```

packages/relay/src/lib/services/lockService/LockService.ts
```diff
@@ -14,13 +15,16 @@ export class LockService {
    */
   private readonly strategy: LockStrategy;
 
+  private readonly strategyType: string;
```
jasuwienas: I think we have already type defined for that:

https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4847/changes#diff-1c9901cf108c03a501f7935426c4f2c5123bcffb4e4923085a132fc126550b90R8

```
/**
 * Strategy type label values for lock metrics.
 */
export type LockStrategyLabel = 'local' | 'redis';
```

packages/relay/src/lib/services/transactionPoolService/transactionPoolService.ts
```diff
@@ -132,8 +222,27 @@ export class TransactionPoolService implements ITransactionPoolService {
...
+    }
+  }
 
-    return payloads;
+  /**
+   * Retrieves the number of unique addresses with pending transactions.
+   *
+   * @returns A promise that resolves to the count of unique addresses.
+   */
+  async getUniqueAddresses(): Promise<number> {
```
jasuwienas: Maybe we should include Count in the method name, just to make it clear from the call site that the result is a number rather than an array.

packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
```diff
@@ -90,9 +119,14 @@ export class LocalLockStrategy implements LockStrategy {
     if (state) {
       // Ensure only the lock owner can release
       if (state.sessionKey === sessionKey) {
+        const holdTime = Date.now() - state.acquiredAt!;
+
         await this.doRelease(state);
```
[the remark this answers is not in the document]
konstantinabl: Edit: actually we cannot do it after doRelease currently, because doRelease resets state.acquiredAt and we would get an invalid calculation for holdTime. I can move the metric update in doRelease but then I will have to change it and I'm not sure if I should move the reset of state after the release
So instead of having it like this
```
private async doRelease(state: LockState): Promise<void> {
    // Clear timeout first
    clearTimeout(state.lockTimeoutId!);

    // Reset state
    state.sessionKey = null;
    state.lockTimeoutId = null;
    state.acquiredAt = null;

    // Release the mutex lock
    state.mutex.release();
  }
```

I can change it to
```
private async doRelease(state: LockState): Promise<void> {
    // Clear timeout first
    clearTimeout(state.lockTimeoutId!);

    // Release the mutex lock
    state.mutex.release();
     
   update metric here
    // Reset state
    state.sessionKey = null;
    state.lockTimeoutId = null;
    state.acquiredAt = null;
  }
```

packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
```diff
@@ -33,20 +34,31 @@ export class LocalLockStrategy implements LockStrategy {
     max: ConfigService.get('LOCAL_LOCK_MAX_ENTRIES'),
   });
 
+  readonly type: string = 'local';
```
quiet-node: Looks like mising TsDoc like other propoerties

packages/relay/src/lib/services/lockService/LockMetricsService.ts
```diff
@@ -0,0 +1,235 @@
...
+ * Strategy type label values for lock metrics.
+ */
+export type LockStrategyLabel = 'local' | 'redis';
+
+/**
+ * Status label values for lock acquisition metrics.
+ */
+export type LockAcquisitionStatus = 'success' | 'fail';
+
+export type RedisOperationLabel = 'acquire' | 'release' | 'heartbeat';
```
quiet-node: let's be consistent and add tsdoc for this

packages/relay/src/lib/services/lockService/RedisLockStrategy.ts
```diff
@@ -23,16 +24,25 @@ import { LockService } from './LockService';
...
   private readonly heartbeatTtlMs: number;
   private readonly keyPrefix = 'lock';
 
-  constructor(redisClient: RedisClientType, logger: Logger) {
+  /**
+   * In-memory map to track lock acquisition times by session key.
+   * Used for calculating hold duration metrics.
+   */
+  private readonly acquisitionTimes: Map<string, number> = new Map();
+
```
quiet-node: I'm not in favor of using a local unbounded Map in a distributed Redis-backed service. I see a few problems with this approach.

First, it could cause a memory leak. The unbounded Map adds an entry during acquire() but doesn't delete it until release(). If anything goes wrong between these two operations, the entry stays in the Map forever and could eventually cause out-of-memory errors.

Second, since this service uses Redis and should be as stateless as possible, splitting the state between Redis (the lock itself) and local variables (the timing metadata) makes the system fragile and error-prone.

Instead, would it make sense to store a compound value like `sessionKey:acquisitionTime` in Redis? 

This would let us retrieve the timestamp during the release operation effectively, and Relay would be as stateless as it was before the update.

Could look like below:

Acquisition:
```
// Concatenate key and time
const val = `${sessionKey}:${Date.now()}`;
const acquired = await this.redisClient.set(lockKey, val, {
  NX: true,
  PX: this.maxLockHoldMs,
});
```

Release:
```
// Lua script to find the value of the lockKey and then extract the timestamp base on sessionKey
local val = redis.call("get", KEYS[1])
if val then
    -- Extract sessionKey (everything before the colon)
    local storedKey = string.match(val, "([^:]+)")
    if storedKey == ARGV[1] then
        redis.call("del", KEYS[1])
        return val -- Return the full "key:timestamp" string
    end
end
return 0
```

With this approach, the acquisition timestamp always stays closely and side-by-side with the sessionKey. The system becomes resilient to pod restarts or any issues between acquire and release. It's simpler, remain stateless, eliminates the possibility of memory leaks, and we can remove the acquisitionTimes map entirely.

Let me know what you think or any question, but I wouldn't vote for an in-memory solution in Redis backed service.

packages/relay/src/lib/services/transactionPoolService/RedisPendingTransactionStorage.ts
```diff
@@ -102,6 +102,15 @@ export class RedisPendingTransactionStorage implements PendingTransactionStorage
...
 
+  /**
+   * Retrieves the size of the global set, giving all unique addresses in the pool
+   *
+   * @returns Number of unique addresses in the pool
+   */
+  async getSetSize(): Promise<number> {
+    return await this.redisClient.sCard(this.globalPendingTxsKey);
+  }
+
```
quiet-node: Same with this, the TSDoc says unique addresses, but then this.redisClient.sCard(this.globalPendingTxsKey) seems to counts total transactions, not addresses, right?

quiet-node: Also, why does this getSetSize() and the getSetSize() on LocalPendingTransaction return different things? It looks like these methods should serve as data retrieval points for the `rpc_relay_txpool_active_addresses` metrics, which should `track the number of unique addresses with pending transactions`, but these two are not consistent and seems to return the wrong thing after all.

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -67,12 +68,16 @@ export class TransactionService implements ITransactionService {
    */
   private readonly mirrorNodeClient: MirrorNodeClient;
 
+  private readonly wrongNonceMetric: Counter;
```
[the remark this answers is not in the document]
konstantinabl: fixed

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -67,12 +68,16 @@ export class TransactionService implements ITransactionService {
...
+
   /**
    * The precheck class used for checking the fields like nonce before the tx execution.
    * @private
    */
   private readonly precheck: Precheck;
 
+  private readonly registry: Registry;
```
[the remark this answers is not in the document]
konstantinabl: yes, you are right

packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
```diff
@@ -151,11 +188,17 @@ export class LocalLockStrategy implements LockStrategy {
...
 
+      // Record metrics for timeout release
+      this.lockMetricsService.recordHoldDuration('local', holdTimeS);
+      this.lockMetricsService.recordTimeoutRelease('local');
+      this.lockMetricsService.decrementActiveCount('local');
+
       if (this.logger.isLevelEnabled('debug')) {
-        const holdTime = Date.now() - state.acquiredAt!;
-        this.logger.debug(`Force releasing expired local lock for address ${address} held for ${holdTime}ms.`);
+        this.logger.debug(`Force releasing expired local lock for address ${address} held for ${holdTimeS}ms.`);
```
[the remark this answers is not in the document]
konstantinabl: In my opinion, i think this is out of scope. The PR is already big, we may open a new issue for this

packages/relay/src/lib/services/lockService/LockService.ts
```diff
@@ -58,4 +61,8 @@ export class LockService {
   static normalizeAddress(address: string): string {
     return address.toLowerCase();
   }
+
+  getStrategyType(): string {
```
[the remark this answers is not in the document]
konstantinabl: added

packages/relay/src/lib/services/lockService/LocalLockStrategy.ts
```diff
@@ -58,24 +68,43 @@ export class LocalLockStrategy implements LockStrategy {
...
+      this.lockMetricsService.recordAcquisition('local', 'success');
+      this.lockMetricsService.incrementActiveCount('local');
 
-    return sessionKey;
+      return sessionKey;
+    } catch (error) {
+      this.lockMetricsService.recordAcquisition('local', 'fail');
+      throw error;
+    } finally {
+      this.lockMetricsService.decrementWaitingTxns('local');
```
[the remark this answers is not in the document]
jasuwienas: For example. I know we’ve had issues with zombie processes before, and they started appearing because a single request that acquired a lock went down and never released or processed it. So I guess this can happen not only on the whole-app/pod level. @konstantinabl

packages/relay/src/lib/relay.ts
```diff
@@ -341,6 +341,11 @@ export class Relay {
     this.metricService = new MetricService(this.logger, metricsCollector, this.register, hbarLimitService);
 
     const storage = PendingTransactionStorageFactory.create(this.redisClient);
+    const transactionPoolService = new TransactionPoolService(
+      storage,
+      this.logger.child({ name: 'transaction-pool-service' }),
+      this.register,
+    );
```
quiet-node: **Notes for reviewers**: The reason `TransactionPoolService` initialization is pulled out to the global level and injected into `EthImpl` and `TxPoolImpl`, instead of being initialized inside those services themselves, is that there was a critical bug. Previously, `EthImpl` would create one `TransactionPoolService` instance, then `TxPoolImpl` would create another instance, and both would register the same 4 metrics to the same `Registry`. The `removeSingleMetric` call meant the second instance's metrics silently replaced the first instance's metrics. The impact was that `EthImpl's TransactionPoolService` metrics, including the `collect()` callbacks, were silently discarded, leaving only `TxPoolImpl`'s metrics active. The fix is to create TransactionPoolService once in relay.ts and inject the same instance into both EthImpl and TxPoolImpl.

simzzz (approved):
LGTM

PR #4856, opened 2026-01-29
fix: ensure SDK consensus errors are properly surfaced to clients for eth_sendRawTransaction

### LINKED ISSUE #4848: Fix Error Masking in sendRawTransactionProcessor: SDK Errors with TransactionIDs Incorrectly Treated as Successful
## Problem Description

The `sendRawTransactionProcessor()` method in `TransactionService` has a critical error handling flaw that causes SDK errors with valid transaction IDs to be masked and treated as successful transactions.

### Current Flawed Logic Flow

1. `submitTransaction()` is called and returns `{ txSubmitted, submittedTransactionId, error }`
2. When SDK errors occur (e.g., `WRONG_NONCE`, `INVALID_ACCOUNT_ID`, `CONSENSUS_GAS_EXHAUSTED`, `BUSY`), the SDK still populates `submittedTransactionId` because the transaction reached consensus before failing
3. The code checks `if (submittedTransactionId)` **before** checking the error
4. This causes the method to query the Mirror Node for contract results and return a transaction hash, **masking the actual error**
5. Only if the Mirror Node query fails does the error handler get invoked

### Root Cause

In `submitTransaction()`:

```typescript
if (e instanceof SDKClientError) {
  submittedTransactionId = e.transactionId || '';
}
```

When SDK throws errors like `WRONG_NONCE`, `INVALID_ACCOUNT_ID`, etc., the transaction has already been submitted to consensus and received a transaction ID. The SDK correctly populates this ID in the `SDKClientError`, which then gets assigned to `submittedTransactionId`.

The issue is in `sendRawTransactionProcessor()`:

```typescript
if (submittedTransactionId) {
  // Query Mirror Node and return transaction hash
  // This executes even when there's a legitimate error!
}
```

This check happens **before** evaluating the error, so SDK errors with transaction IDs are incorrectly treated as successful.

### Impact

1. **Silent Failures**: Clients receive transaction hashes for failed transactions instead of proper error responses
2. **Debugging Difficulty**: Actual error details are lost, making troubleshooting nearly impossible
3. **API Compliance**: RPC behavior diverges from expected JSON-RPC error handling standards

## Proposed Solution

Reorder the logic in `sendRawTransactionProcessor()` to check errors **before** processing transaction IDs.

### New Logic Flow

1. After `submitTransaction()` returns, check if `error` is present and is **not** an SDK timeout error
2. If there's a non-timeout error, invoke `sendRawTransactionErrorHandler()` immediately
3. Only if there's no error (or only a timeout error), proceed with the `if (submittedTransactionId)` block

### SDK Timeout Detection

SDK timeout errors need special handling because they may have been successful at consensus level:

- `e.isTimeoutExceeded()` - timeout exceeded
- `e.isConnectionDropped()` - connection dropped
- `e.isGrpcTimeout()` - GRPC timeout

These methods are already available on `SDKClientError` ([SDKClientError.ts:37-48](packages/relay/src/lib/errors/SDKClientError.ts#L37-L48)).

quiet-node:
### Description

This PR fixes a critical error handling flaw in the `sendRawTransactionProcessor` method where SDK errors with valid transaction IDs were incorrectly masked and treated as successful transactions.

Previously, the logic checked for the existence of a `submittedTransactionId` _before_ checking for an error. Since the SDK populates the transaction ID even for consensus-level failures (like `WRONG_NONCE`, `INVALID_ACCOUNT_ID`, `CONSENSUS_GAS_EXHAUSTED`), the Relay would proceed to poll the Mirror Node for a transaction record that would never exist, masking the actual error.

**Key Changes:**

1. **Reordered Error Logic:** The processor now evaluates the `error` state before attempting to process the `submittedTransactionId`.

2. **Smart Verification with `shouldVerifyWithMirrorNode()`:** Introduced intelligent error classification to distinguish between errors requiring different handling:
   - **Pre-execution Failures:** Errors in `HEDERA_SPECIFIC_REVERT_STATUSES` config (e.g., `WRONG_NONCE`, `INVALID_ACCOUNT_ID`, `BUSY`, `THROTTLED_AT_CONSENSUS`, `CONSENSUS_GAS_EXHAUSTED`) represent failures that occurred before transaction execution at consensus. These are thrown immediately without Mirror Node polling for txHash.
   - **Post-execution Failures:** Other SDK errors (e.g., `INVALID_CONTRACT_ID`, `CONTRACT_REVERT_EXECUTED`, etc.) represent transactions that executed but reverted. These require Mirror Node polling to validate & retrieve the transaction hash.
   - **Timeout Failures:** SDK timeout errors (`timeout exceeded`, `Connection dropped`, gRPC timeout) are now treated as definitive failures. These are thrown immediately without Mirror Node polling to reduce unnecessary load and improve response time.

3. **Enhanced Error Handling:**
   - Consensus errors are now properly surfaced to clients as `CONSENSUS_NODE_ERROR` (code `-32000`) instead of generic `INTERNAL_ERROR` (code `-32603`)
   - Error messages preserve the original SDK status and details

4. **Configuration-Driven Design:**
   - Pre-execution failure statuses are defined in `HEDERA_SPECIFIC_REVERT_STATUSES` config
   - `INVALID_SOLIDITY_ADDRESS` is now included in this list because it fails before execution and has no valid transaction record.
   - Unit tests validate synchronization between test cases and configuration
   - Easy to maintain and extend as new error types are discovered

### Related issue(s)

Fixes #4848

### Testing Guide

- Send a valid transaction (should return hash).
- Send a transaction with an incorrect nonce:
- **Expected:** Immediate `Nonce too high` or `Nonce too low` error
- **Previous:** Long timeout followed by `INTERNAL_ERROR`
- **SDK Consensus Error Test (e.g. `CONSENSUS_GAS_EXHAUSTED`):**
  - To verify generic SDK errors are surfaced correctly, simulate a `CONSENSUS_GAS_EXHAUSTED` error by configuring the local node with extreme throttling:

    ```json
    {"key": "contracts.throttle.throttleByOpsDuration", "value": "true"},
    {"key": "contracts.opsDurationThrottleCapacity", "value": "10"},
    {"key": "contracts.opsDurationThrottleUnitsFreedPerSecond", "value": "0"}
    ```

  - Send any `eth_sendRawTransaction`
  - **Expected:** Immediate`CONSENSUS_NODE_ERROR`with`CONSENSUS_GAS_EXHAUSTED` details (demonstrating the fix applies to all pre-execution consensus errors)

### Checklist

### DIFF
--- a/packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
+++ b/packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
@@ -513,59 +513,102 @@ export class TransactionService implements ITransactionService {
     // Remove the transaction from the transaction pool after submission
     await this.transactionPoolService.removeTransaction(originalCallerAddress, parsedTx.serialized);
 
-    // Determine if Mirror Node verification is needed
-    const shouldVerifyWithMirrorNode = submittedTransactionId && this.shouldVerifyWithMirrorNode(error);
+    // Handle submission errors - throws for definitive failures, returns for MN polling cases
+    await this.handleSubmissionError(error, parsedTx, requestDetails);
 
-    if (!shouldVerifyWithMirrorNode) {
-      // Definitive failure: handle error immediately without MN polling
-      await this.sendRawTransactionErrorHandler(error, parsedTx, requestDetails);
-    }
-
-    // Poll Mirror Node to validate transaction & retrieve transaction hash
+    // At this point, either no error or a post-execution failure that needs MN polling
     return this.getTransactionHashFromMirrorNode(submittedTransactionId, error, requestDetails);
   }
 
   /**
-   * Determines whether to poll the Mirror Node to validate a transaction after submission.
+   * Handles transaction submission errors by classifying and routing them appropriately.
    *
-   * Polling is skipped for errors that definitively indicate transaction failure without execution:
-   * - Non-SDK errors (application-level failures)
-   * - Pre-execution failures (status codes in `HEDERA_SPECIFIC_REVERT_STATUSES`)
-   * - SDK timeout errors (network timeouts or connection failures)
+   * This method serves as the single decision point for error handling after transaction submission.
+   * It evaluates the error type and either throws immediately or returns to allow Mirror Node polling.
    *
-   * Polling is performed for:
-   * - Successful submissions (to retrieve the transaction hash)
-   * - Post-execution failures (transactions that executed on-chain but reverted)
+   * Error handling flow:
+   * 1. No error → return (proceed to MN polling for tx hash)
+   * 2. Non-SDK error → throw as-is (application-level failure)
+   * 3. SDK timeout error → throw as-is (network failure)
+   * 4. Pre-execution failure (in HEDERA_SPECIFIC_REVERT_STATUSES):
+   *    - WRONG_NONCE: polls MN for account nonce, throws NONCE_TOO_HIGH/NONCE_TOO_LOW
+   *    - Others: throws TRANSACTION_REJECTED with status details
+   * 5. Post-execution failure → return (proceed to MN polling for tx hash)
    *
-   * @param error - The error from transaction submission, or `null` if successful
-   * @returns `true` if the Mirror Node should be polled, `false` otherwise
+   * @param error - The error from transaction submission, or null/undefined if successful
+   * @param parsedTx - The parsed transaction for nonce comparison (used for WRONG_NONCE handling)
+   * @param requestDetails - Request details for logging and tracking
+   * @throws {JsonRpcError} NONCE_TOO_HIGH or NONCE_TOO_LOW for wrong nonce errors
+   * @throws {JsonRpcError} TRANSACTION_REJECTED for pre-execution failures
+   * @throws {Error} Original error for non-SDK or timeout errors
    */
-  private shouldVerifyWithMirrorNode(error: any): boolean {
-    // Success case - always poll to get transaction hash
+  private async handleSubmissionError(
+    error: any,
+    parsedTx: EthersTransaction,
+    requestDetails: RequestDetails,
+  ): Promise<void> {
+    // No error - proceed to MN polling for transaction validation and txhash retrieval
     if (!error) {
-      return true;
+      return;
     }
 
-    // Non-SDK errors are application-level failures - don't poll
+    // Non-SDK errors are definitive failures - propagate as-is
     if (!(error instanceof SDKClientError)) {
-      return false;
+      throw error;
     }
 
     // Update metrics for SDK errors
     this.hapiService.decrementErrorCounter(error.statusCode);
 
-    // SDK timeout errors indicate consensus node did not process the transaction
+    // SDK timeout errors - propagate as-is
     if (error.isTimeoutExceeded() || error.isConnectionDropped() || error.isGrpcTimeout()) {
-      return false;
+      throw error;
     }
 
     // Check if this is a pre-execution failure
     const preExecutionFailures: string[] = ConfigService.get('HEDERA_SPECIFIC_REVERT_STATUSES');
-    const isPreExecutionFailure = preExecutionFailures.includes(error.status.toString());
+    if (preExecutionFailures.includes(error.status.toString())) {
+      // WRONG_NONCE requires special handling to determine if nonce is too high or too low
+      // TODO: should be removed in https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/4860
+      if (error.status.toString() === constants.TRANSACTION_RESULT_STATUS.WRONG_NONCE) {
+        const mirrorNodeGetContractResultRetries = this.mirrorNodeClient.getMirrorNodeRequestRetryCount();
 
-    // Return true if error is NOT in pre-execution failures list -> poll MN
-    // Return false if error IS in pre-execution failures list -> throw immediately
-    return !isPreExecutionFailure;
+        // Poll Mirror Node to get updated account nonce for comparison
+        let accountNonce: number | null = null;
+        for (let i = 0; i < mirrorNodeGetContractResultRetries; i++) {
+          const accountInfo = await this.mirrorNodeClient.getAccount(parsedTx.from!, requestDetails);
+          if (accountInfo.ethereum_nonce !== parsedTx.nonce) {
+            accountNonce = accountInfo.ethereum_nonce;
+            break;
+          }
+
+          this.logger.trace(
+            `Repeating retry to poll for updated account nonce. Count %d of %d. Waiting %d ms before initiating a new request`,
+            i,
+            mirrorNodeGetContractResultRetries,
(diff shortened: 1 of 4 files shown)


packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -515,96 +513,128 @@ export class TransactionService implements ITransactionService {
...
+    if (!(error instanceof SDKClientError)) {
+      return false;
+    }
 
-        return contractResult.hash;
-      } catch (e: any) {
-        sendRawTransactionError = e;
-      }
+    // Update metrics for SDK errors
+    this.hapiService.decrementErrorCounter(error.statusCode);
```
Ferparishuertas: Why this counter metric is within this method
why are we decrementing?

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -515,96 +513,128 @@ export class TransactionService implements ITransactionService {
...
+   * @param parsedTx - The parsed transaction for nonce comparison
+   * @param requestDetails - Request details for logging and tracking
+   * @throws {JsonRpcError} Throws NONCE_TOO_HIGH or NONCE_TOO_LOW for wrong nonce errors
+   * @throws {Error} Re-throws all other errors as-is
    */
   private async sendRawTransactionErrorHandler(
-    e: any,
-    transactionBuffer: Buffer,
-    txSubmitted: boolean,
+    error: any,
```
Ferparishuertas: Whats the reason to keep this value, having known that MN will never be accurate on nonces, and we implemented a txpool for this.
Giving nonce to low or high, wirh the changes to have MN errors, is really adding value?

packages/config-service/src/services/globalConfig.ts
```diff
@@ -230,6 +230,7 @@ const _CONFIG = {
       'EVM_HOOK_GAS_THROTTLED',
       'BUSY',
       'CONSENSUS_GAS_EXHAUSTED',
+      'INVALID_SOLIDITY_ADDRESS',
```
Ferparishuertas: This list is patch overall. This needs to be classified at CN level, because if not we will loose the track what is rejection and what not

Ferparishuertas (commented):
It looking good
That said i do have key functional observations

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
quiet-node: This metric is in this method because at this point only SDK errors are possible, making it a good place to call this metric. We decrement it because the method documentation explains that it decrements the error encountered counter, and when it reaches 0, the client is reset.

This is still included to preserve the old logic that was previously there.

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
quiet-node: Yes should be addressed in the remove WRONG_NONCE special handler ticket https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/4860

packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
```diff
@@ -906,5 +920,308 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
...
+        });
+
+        /**
+         * Post-execution errors (not in HEDERA_SPECIFIC_REVERT_STATUSES) should trigger
+         * Mirror Node polling because the transaction executed on the network and has a
+         * valid transaction record, even though it failed.
+         *
+         * Examples: CONTRACT_REVERT_EXECUTED, INVALID_CONTRACT_ID, INVALID_ALIAS_KEY
+         */
+        const POST_EXECUTION_ERRORS: Array<{ statusName: string; statusCode: number }> = [
```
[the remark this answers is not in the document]
quiet-node: Are you suggesting we should call it CONSENSUS_TIME_ERRORS instead of POST_EXECUTION_ERRORS? I think POST_EXECUTION_ERRORS makes more sense because this category is for errors like CONTRACT_REVERT_EXECUTED and INVALID_ALIAS_KEY, where transactions were executed but reverted, and they have valid records in Mirror Node. This is different from the list we included in HEDERA_SPECIFIC_REVERT_STATUSES, such as BUSY or CONSENSUS_GAS_EXHAUSTED.

CONSENSUS_TIME_ERRORS is a but misleading to me for this case

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -515,96 +513,128 @@ export class TransactionService implements ITransactionService {
...
-          this.mirrorNodeClient.getMirrorNodeRequestRetryCount(),
-        );
+    // Determine if Mirror Node verification is needed
+    const shouldVerifyWithMirrorNode = submittedTransactionId && this.shouldVerifyWithMirrorNode(error);
 
-        if (!contractResult) {
-          if (sendRawTransactionError instanceof SDKClientError) {
-            throw sendRawTransactionError;
-          }
+    if (!shouldVerifyWithMirrorNode) {
```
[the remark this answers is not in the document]
quiet-node: After reviewing it makes more sense to throw TRANSACTION_REJECT for these errors instead of CONSENSUS_NODE_ERROR, will fix

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -1495,8 +1495,8 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
           Assertions.expectedError();
         } catch (e: any) {
           const { error } = e?.response ? e.response.bodyJson : e;
-          expect(error.code).to.eq(predefined.INTERNAL_ERROR().code);
-          expect(error.message).to.contain(`failed precheck with status INVALID_SOLIDITY_ADDRESS against node account`);
+          expect(error.code).to.eq(-32000); // CONSENSUS_NODE_ERROR
```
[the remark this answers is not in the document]
quiet-node: fixed in https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4856/changes/ae2995193a9cf56c0ebaea78fb47e41a81944672 where I use TRANSACTION_REJECT (-32003) instead of generic CONSENSUS_SDK_ERROR.

Thanks for the nice catch

packages/server/tests/acceptance/rpc_batch1.spec.ts
```diff
@@ -1495,8 +1495,8 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+          expect(error.message).to.contain(`INVALID_SOLIDITY_ADDRESS`);
```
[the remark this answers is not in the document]
Ferparishuertas: Gotcha. Was confused by other precheck error

Ferparishuertas (approved):
LGTM

natanasow (approved):

PR #4872, opened 2026-02-03
fix: reduced Mirror Node load for WRONG_NONCE handling

### LINKED ISSUE #4864: Refactor WRONG_NONCE error handling in TransactionService to reduce Mirror Node load
### Summary

The current implementation of `handleSubmissionError` in eth_sendRawTransaction flow handles the `WRONG_NONCE` error by repeatedly polling the Mirror Node to find an updated account nonce. This approach increases load on the Mirror Node and returns a non-standard `INTERNAL_ERROR` if the nonce cannot be retrieved, which is confusing for end users.

### Motivation

- **Performance**: Repeated requests to the Mirror Node for a single transaction failure are inefficient and increase latency and traffic.
- **Error Clarity**: The current `predefined.INTERNAL_ERROR("Cannot find updated account nonce for WRONG_NONCE error.")` is not standard for JSON-RPC clients and doesn't accurately reflect why the transaction failed from the client's perspective.
- **Robustness**: If the Mirror Node is lagging or the account info is temporarily unavailable, we should fail gracefully with a standard error.

### Proposed Changes

Refactor the `WRONG_NONCE` handling logic:

1.  **Reduce Requests**: Replace the retry loop for `getAccount` with a single request to the Mirror Node.
2.  **Graceful Error Handling**: Wrap the `getAccount` call in a `try/catch` block.
3.  **Refactor Nonce Logic**:
    - If the request is successful and `accountInfo` contains a valid `ethereum_nonce`:
      - If `parsedTx.nonce > accountNonce`, throw `NONCE_TOO_HIGH`.
      - Else, throw `NONCE_TOO_LOW`.
    - If the request fails for any reason (Mirror Node error, account not found, or `accountInfo.ethereum_nonce` is missing/null), immediately throw `predefined.TRANSACTION_REJECTED` instead of `INTERNAL_ERROR`.
4.  Remove the existing retry loop and the `accountNonce === null` check.

quiet-node:
### Description

This PR refactors the `WRONG_NONCE` error handling in `TransactionService.handleSubmissionError` to reduce Mirror Node load and improve error clarity for end users.

Previously, `WRONG_NONCE` errors triggered a retry loop polling the Mirror Node for account nonce updates. When nonce determination failed, an `INTERNAL_ERROR` was thrown with a non-standard message.

This change replaces the polling mechanism with a single Mirror Node request. When nonce comparison succeeds, `NONCE_TOO_HIGH` or `NONCE_TOO_LOW` errors are thrown as before. When the request fails or nonce cannot be determined, the error gracefully falls through to `TRANSACTION_REJECTED` with the original `WRONG_NONCE` status, providing a standard JSON-RPC error response.

### Related issue(s)

Fixes #4864

### Checklist

### DIFF
--- a/packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
+++ b/packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
@@ -531,7 +531,10 @@ export class TransactionService implements ITransactionService {
    * 2. Non-SDK error → throw as-is (application-level failure)
    * 3. SDK timeout error → throw as-is (network failure)
    * 4. Pre-execution failure (in HEDERA_SPECIFIC_REVERT_STATUSES):
-   *    - WRONG_NONCE: polls MN for account nonce, throws NONCE_TOO_HIGH/NONCE_TOO_LOW
+   *    - WRONG_NONCE: fetch account nonce from MN
+   *      - If nonce too high → throw NONCE_TOO_HIGH
+   *      - If nonce too low → throw NONCE_TOO_LOW
+   *      - If unable to determine → fallback to original status
    *    - Others: throws TRANSACTION_REJECTED with status details
    * 5. Post-execution failure → return (proceed to MN polling for tx hash)
    *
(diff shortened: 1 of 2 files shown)


packages/relay/tests/lib/eth/eth_sendRawTransaction.spec.ts
```diff
@@ -1153,11 +1153,53 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
...
+          };
+          const signed = await signTransaction(txWithNonce);
+
+          // SDK throws WRONG_NONCE error with proper Status object
+          const wrongNonceError = new SDKClientError(
+            { status: Status.WrongNonce, message: 'WRONG_NONCE' },
+            'WRONG_NONCE',
+            transactionIdServicesFormat,
+          );
+          sdkClientStub.submitEthereumTransaction.throws(wrongNonceError);
```
jasuwienas: Would it make sense to include an acceptance test as well?

jasuwienas: No, it makes sense now. I get why we had it.

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -571,34 +574,22 @@ export class TransactionService implements ITransactionService {
...
-        if (accountNonce === null) {
-          throw predefined.INTERNAL_ERROR(`Cannot find updated account nonce for WRONG_NONCE error.`);
-        }
-
-        if (parsedTx.nonce > accountNonce) {
-          throw predefined.NONCE_TOO_HIGH(parsedTx.nonce, accountNonce);
-        } else {
-          throw predefined.NONCE_TOO_LOW(parsedTx.nonce, accountNonce);
+        // Determine if nonce is too high or too low
+        if (accountNonce != null && accountNonce !== parsedTx.nonce) {
```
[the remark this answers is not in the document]
jasuwienas: well, ok :)

PR #4883, opened 2026-02-06
feat: reduce unnecessary redis calls

### LINKED ISSUE #4882: Unnecessary Redis calls and missing error handling cause `eth_sendRawTransaction` failures
# Problem

## Context

On mainnet, `eth_sendRawTransaction` fails intermittently with `Socket closed unexpectedly` errors due to Redis connectivity issues (Redis memory limit was 1GB).

The error originates from the `node-redis` client (`SocketClosedUnexpectedlyError`) and propagates unhandled through prechecks, causing the entire transaction to fail.

## Issues

1. **Unnecessary Redis call when TX pool is disabled**
   - `TransactionPoolService.getPendingCount()` always calls `storage.getList()` (Redis `sCard`) even when `ENABLE_TX_POOL=false`
   - Unlike `saveTransaction` and `removeTransaction`, it's missing the `isEnabled()` early-return check
   - This is called during prechecks in `Precheck.sendRawTransactionCheck()` (line 77)

2. **No graceful degradation on Redis failure**
   - A transient Redis connection failure during prechecks should not fail the transaction outright
   - `getPendingCount` could safely return `0` if Redis is unavailable, since the pending count is supplementary to the mirror node nonce

3. **Audit other call sites for unnecessary Redis calls**
   - Verify whether other `TransactionPoolService` methods (`getTransactions`, `getAllTransactions`) or `CacheService` operations have similar gaps where Redis is hit unnecessarily or without proper error handling

# Solution

## Action Items

- [ ] Add `isEnabled()` check to `getPendingCount` (and `getTransactions` / `getAllTransactions`)
- [ ] Add error handling in `getPendingCount` to return `0` on Redis failure
- [ ] Audit remaining Redis call sites for unnecessary calls when features are disabled
- [ ] Review Redis memory/connection configuration for mainnet

## References

- `packages/relay/src/lib/services/transactionPoolService/transactionPoolService.ts` - missing guard
- `packages/relay/src/lib/precheck.ts:77` - call site
- `packages/relay/src/lib/errors/RedisCacheError.ts` - `SocketClosedUnexpectedlyError`
- `packages/relay/src/lib/clients/redisClientManager.ts` - reconnect strategy

### Alternatives

_No response_

simzzz:
### Description

This PR reduces unnecessary Redis calls when TX_POOL is disabled, avoiding issues where transactions fail to send if redis is unavailable.

### Related issue(s)

Fixes #4882 

### Testing Guide

1. Run the relay with TX_POOL disabled
2. Send a transaction
3. Verify in the logs that no calls to Redis are made

N/A

N/A

### Checklist

### Additional Investigation
For the "Audit other call sites for unnecessary Redis calls" part:

For CacheService and HBARLimiter there are no redundant calls and we fallback to LRU
For TransactionPoolService, I've addressed them with guards
For LockService, there is no fallback and it's not redundant. If Redis dies mid-flight, acquireLock o releaseLock can throw
For RateLimiterService, there is no fallback and it's not redundant. If Redis dies, rate limiting fails
For the graceful degradation issue, it could be problematic and I believe we didn't include it initially on purpose:
If getPendingCount() catches Redis errors and returns 0 internally, it breaks eth_getTransactionCount('pending'), because wallets will receive stale nonces (mnNonce + 0 instead of mnNonce + actualPendingCount), causing them to submit transactions with a wrong nonce.

If only the precheck.ts call site catches the Redis error and returns 0, while AccountService lets it throw, you get context-dependent behavior, the same method call either throws or returns 0 depending on the caller. This violates function purity/predictability: callers can't reason about the method's behavior without knowing the entire call stack. It also creates maintenance burden since error handling is scattered across call sites instead of centralized.

### DIFF
--- a/packages/relay/src/lib/services/transactionPoolService/transactionPoolService.ts
+++ b/packages/relay/src/lib/services/transactionPoolService/transactionPoolService.ts
@@ -3,11 +3,13 @@
 import { ConfigService } from '@hashgraph/json-rpc-config-service/dist/services';
 import { Transaction } from 'ethers';
 import { Logger } from 'pino';
+import { Counter, Gauge, Registry } from 'prom-client';
 
 import {
   PendingTransactionStorage,
   TransactionPoolService as ITransactionPoolService,
 } from '../../types/transactionPool';
+import { RedisPendingTransactionStorage } from './RedisPendingTransactionStorage';
 
 /**
  * Service implementation that orchestrates pending transaction management.
@@ -15,11 +17,9 @@ import {
  */
 export class TransactionPoolService implements ITransactionPoolService {
   /**
-   * Return if the transaction pool is enabled based on ENABLE_TX_POOL env
+   * Gauge tracking number of unique addresses with pending transactions.
    */
-  public static isEnabled(): boolean {
-    return ConfigService.get('ENABLE_TX_POOL');
-  }
+  private readonly activeAddressesGauge: Gauge;
 
   /**
    * The logger used for logging transaction pool operations.
@@ -29,6 +29,21 @@ export class TransactionPoolService implements ITransactionPoolService {
   private readonly logger: Logger;
 
   /**
+   * Counter tracking pool operations (add, remove).
+   */
+  private readonly operationsCounter: Counter;
+
+  /**
+   * Gauge tracking current total pending transactions across all addresses.
+   */
+  private readonly pendingCountGauge: Gauge;
+
+  /**
+   * Counter tracking storage operation failures by operation and backend type.
+   */
+  private readonly storageErrorsCounter: Counter;
+
+  /**
    * The storage implementation for managing pending transactions.
    *
    * @private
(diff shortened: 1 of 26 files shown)


packages/relay/src/lib/services/transactionPoolService/transactionPoolService.ts
```diff
@@ -116,8 +122,13 @@ export class TransactionPoolService implements ITransactionPoolService {
...
    * @param address - The account address to query.
    * @returns A promise that resolves to a Set of RLP hex strings.
+   *          Returns an empty set if the transaction pool is disabled.
    */
   async getTransactions(address: string): Promise<Set<string>> {
+    if (!TransactionPoolService.isEnabled()) {
+      return new Set();
+    }
+
     const addressLowerCased = address.toLowerCase();
```
quiet-node: Same here I don't think we have a test for this. Let's add some

PR #4919, opened 2026-02-13
feat: add `debug_getRawReceipts` method

### LINKED ISSUE #4892: Add `debug_getRawReceipts` method
### Problem

Currently, we're missing some `debug_*` methods.

### Solution

- Implement `debug_getRawReceipts` method - we should use `eth_getBlockReceipts` as a base
- Check conformity tests
- Edit **openrpc.json** and **rpc-api.md**
-  Add tests (if needed)

RPC method:
```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "debug_getRawReceipts",
  "params": [
    "latest" // block number or block hash
  ]
}
```
Response:
```json
{
  "jsonrpc": "2.0",
  "id": "1",
  "result": [
    "0xf901a80...",   // RLP encoded receipt
    "0xf901a70...",   // RLP encoded receipt
    "0xf901a22..."   // RLP encoded receipt
  ]
}
```

**Note: for not found receipts (e.g. future block) return `"results": []`**

### Alternatives

_No response_

BartoszSolkaBD:
### Description

This PR adds support for the debug_getRawReceipts JSON-RPC method. It returns an array of EIP-2718 binary-encoded transaction receipts for a given block. 

Changes:
- Relay: New getRawReceipts(blockNumber, requestDetails) on the Debug interface.
- Docs: debug_getRawReceipts added to docs/openrpc.json and docs/rpc-api.md.
- Tests:
  - Unit (packages/relay/tests/lib/debug.spec.ts): empty array when block not found, empty array when block has no receipts, multiple receipts, single receipt with logs; UNSUPPORTED_METHOD when debug API is disabled; invalid block number → INVALID_PARAMETER.
  - Acceptance (packages/server/tests/acceptance/debug.spec.ts): happy path for a block with transactions (array of hex strings), invalid block number.
  - Conformity: Overwrite packages/server/tests/acceptance/data/conformity/overwrites/debug_getRawReceipts/get-block-n.io.
  - [debug_getRawReceipts.spec.ts](https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4919/changes#diff-b5e5c15e4000c32042bc03280696959559acd6b3a7dd42021ede1fb7d0924f85) - round-trip decode assertions (RLP decode and compare to stubbed receipt fields)

### Related issue(s)

Fixes #4892 

### Testing Guide

1. Run the relay against mainnet
Start the JSON-RPC relay with mainnet config (and DEBUG_API_ENABLED=true if required). Ensure it is connected and serving requests.
2. Call debug_getRawReceipts for block with non-empty receipts.
``` json
   {"jsonrpc":"2.0","id":1,"method":"debug_getRawReceipts","params":["0x56e86ab"]}
```
3. Validate against eth_getBlockReceipts for the same block
``` json
   {"jsonrpc":"2.0","id":1,"method":"eth_getBlockReceipts","params":["0x56e86ab"]}
```

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/factories/transactionReceiptFactory.ts
+++ b/packages/relay/src/lib/factories/transactionReceiptFactory.ts
@@ -183,11 +183,17 @@ class TransactionReceiptFactory {
         ? new Uint8Array(0)
         : hexToBytes(prepend0x(BigInt(cumulativeGasUsed).toString(16))); // canonical RLP encoding (no leading zeros)
 
+    const receiptLogsParam = receipt.logs.map((log) => [
+      hexToBytes(log.address),
+      log.topics.map((t) => hexToBytes(t)),
+      hexToBytes(log.data),
+    ]);
+
     const encodedList = RLP.encode([
       receiptRootOrStatus,
       cumulativeGasUsedBytes,
       hexToBytes(receipt.logsBloom),
-      this.encodeLogsForReceipt(receipt.logs),
+      receiptLogsParam,
     ]);
 
     if (txType === 0) {
@@ -195,19 +201,6 @@ class TransactionReceiptFactory {
     }
     const withPrefix = concatBytes(intToBytes(txType), encodedList);
     return prepend0x(toHexString(withPrefix));
-  }
-
-  /**
-   * Converts receipt logs into the RLP encoded log structure.
-   *
-   * Each log becomes a 3-tuple [address, topics[], data] per the Yellow Paper
-   * (address and data as bytes; topics as array of 32-byte topic hashes).
-   *
-   * @param logs - The logs array from the transaction receipt (see {@link Log}).
-   * @returns Array of [address, topics, data] as Uint8Arrays for RLP encoding.
-   */
-  private static encodeLogsForReceipt(logs: Log[]): [Uint8Array, Uint8Array[], Uint8Array][] {
-    return logs.map((log) => [hexToBytes(log.address), log.topics.map((t) => hexToBytes(t)), hexToBytes(log.data)]);
   }
 }
 


packages/server/tests/acceptance/debug.spec.ts
```diff
@@ -1233,4 +1234,41 @@ describe('@debug API Acceptance Tests', function () {
       });
     });
   });
+
+  describe('debug_getRawReceipts', () => {
```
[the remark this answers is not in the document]
BartoszSolkaBD: we have a test here https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4919/changes#diff-a0d7caed18547ec84f2f466ff3e64154ff1eae3f7d6c6f8ae25c9a3153f5e52cR2030 that validates the fields individually. Please let me know if you think this is sufficient, or if we should also add similar checks to the acceptance tests.

packages/relay/src/lib/debug.ts
```diff
@@ -65,19 +66,27 @@ export class DebugImpl implements Debug {
...
    * @param {ICacheClient} cacheService - Service for managing cached data.
+   * @param {Eth} eth - The Eth implementation used for handling Ethereum-specific JSON-RPC requests.
    */
-  constructor(mirrorNodeClient: MirrorNodeClient, logger: Logger, cacheService: ICacheClient) {
+  constructor(mirrorNodeClient: MirrorNodeClient, logger: Logger, cacheService: ICacheClient, eth: Eth) {
     this.logger = logger;
     this.common = new CommonService(mirrorNodeClient, logger, cacheService);
     this.mirrorNodeClient = mirrorNodeClient;
     this.cacheService = cacheService;
+    this.eth = eth;
```
quiet-node: I have made a comment about circular dependency issue here https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4909#discussion_r2818735143. I think there's a better way to deal with this.

packages/relay/src/lib/debug.ts
```diff
@@ -252,6 +261,36 @@ export class DebugImpl implements Debug {
...
+  })
+  @cache()
+  async getRawReceipts(blockNumber: string, requestDetails: RequestDetails): Promise<string[] | null> {
+    DebugImpl.requireDebugAPIEnabled();
+    const receipts = await this.eth.getBlockReceipts(blockNumber, requestDetails);
+    if (!receipts) {
+      return null;
+    }
+    return receipts.map((receipt) => encodeReceiptToHex(receipt));
+  }
```
quiet-node: I think we're taking advantage of the convenience of reusing `this.eth.getBlockReceipts(blockNumber, requestDetails)`, but we're trading that convenience for a significant performance hit that could look really bad on blocks with tens of thousands of transactions.

The core issue is that `debug_getRawReceipts` routes through the full `eth_getBlockReceipts` pipeline, which is designed to return rich, user-facing JSON receipts with resolved addresses. But raw receipts only need 5 primitive fields that are already available in the mirror node response, no address resolution required. So for a block with 10,000 transactions, that means roughly 20,000 unnecessary `resolveEvmAddress` calls, plus all the wasted object construction for fields like `from`, `to`, `effectiveGasPrice`, and others that `encodeReceiptToHex` might never touch.

There is also a double-loop problem with the current implementation. After `this.eth.getBlockReceipts()` fully resolves and returns the receipt list, it has already looped through each receipt once. Then `getRawReceipts()` in `debug.ts` has to loop through the entire array a second time to encode each receipt. Given that this second loop runs on the main thread, it can easily crash the Relay under heavy network load with large blocks, which has actually happened to us before.

I think a better approach would be to add a dedicated `getRawReceipts` function directly in the block worker. This function would reuse the existing lightweight methods, strip out everything we do not need such as `resolveEvmAddress` calls or unnecessary object construction, and handle the receipt encoding in the same loop to keep it to a single pass. From there, `BlockService` can expose a `getRawReceipts()` method that delegates to the worker thread, and `debug.ts` can simply call `blockService.getRawReceipts()`. This bypasses `eth.getBlockReceipts` entirely and avoids all the overhead that comes with it.

Let me know what you think.

packages/relay/src/lib/types/IReceiptRlpInput.ts
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { Log } from '../model';
+
+export interface IReceiptRlpInput {
```
quiet-node: lets add tsdoc

packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -420,6 +399,101 @@ export async function getBlockReceipts(
...
+    throw WorkersPool.wrapError(e);
+  }
+}
+
+async function loadBlockExecutionData(
+  blockHashOrBlockNumber: string,
+  requestDetails: RequestDetails,
+): Promise<{
+  block: MirrorNodeBlock | null;
+  contractResults: any[];
```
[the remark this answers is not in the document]
BartoszSolkaBD: Yes, `MirrorNodeContractResult`. I would propose to create a follow up PR that will add this type everywhere it's needed.

packages/relay/src/lib/receiptSerialization.ts
```diff
@@ -0,0 +1,76 @@
...
+  return logs.map((log) => [hexToBytes(log.address), log.topics.map((t) => hexToBytes(t)), hexToBytes(log.data)]);
+}
+
+/**
+ * Encodes a single transaction receipt to EIP-2718 binary form.
+ *
+ * Produces the RLP-encoded 4-tuple (receipt_root_or_status, cumulative_gas_used,
+ * logs_bloom, logs) per the Ethereum Yellow Paper. For typed transactions (type !== 0),
+ * the output is the single-byte type prefix followed by that RLP payload (EIP-2718).
+ *
```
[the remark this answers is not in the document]
BartoszSolkaBD: I've added a link to https://ethereum.github.io/yellowpaper/paper.pdf

packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -420,6 +399,101 @@ export async function getBlockReceipts(
...
+      contractResult.logs = logsByHash.get(contractResult.hash) || [];
+
+      const transactionReceiptParams: IRegularTransactionReceiptRlpInputParams = {
+        logs: contractResult.logs,
+        receiptResponse: contractResult,
+        blockGasUsedBeforeTransaction,
+      };
+      const receiptRlpInput = TransactionReceiptFactory.createReceiptRlpInput(
+        transactionReceiptParams,
+      ) as IReceiptRlpInput;
```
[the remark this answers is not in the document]
BartoszSolkaBD: removed

packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -420,6 +398,169 @@ export async function getBlockReceipts(
...
+
+      contractResult.logs = logsByHash.get(contractResult.hash) || [];
+
+      const receiptRlpInput = createReceiptRlpInput(contractResult.logs, contractResult, blockGasUsedBeforeTransaction);
+      blockGasUsedBeforeTransaction += contractResult.gas_used;
+      return encodeReceiptToHex(receiptRlpInput);
+    });
+
+    const resolvedEncodedReceipts = await Promise.all(encodedReceiptPromises);
+    encodedReceipts.push(
```
quiet-node: Looks like the `encodedReceiptPromises` implementation has a misleading pattern. The callbacks in this method are marked as async but contain no await, meaning no actual asynchronous work is being done. Wrapping them in Promise.all provides no concurrency benefit since there is no I/O happening. We should replace this with a regular synchronous loop.

packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -420,6 +398,169 @@ export async function getBlockReceipts(
...
+    cumulativeGasUsed: constants.ZERO_HEX,
+    logs: syntheticLogs,
+    logsBloom: LogsBloomUtils.buildLogsBloom(syntheticLogs[0].address, syntheticLogs[0].topics),
+    root: constants.DEFAULT_ROOT_HASH,
+    status: constants.ONE_HEX,
+    transactionIndex: syntheticLogs[0].transactionIndex,
+    type: constants.ZERO_HEX, // fallback to 0x0 from HAPI transactions
+  };
+}
+
```
quiet-node: Can these two be unified something like

```typescript
function createReceiptRlpInput(
  logs: Log[],
  receiptResponse?: any,
  blockGasUsedBeforeTransaction?: number,
  isSynthetic: boolean
): IReceiptRlpInput {
  return {
    cumulativeGasUsed: isSynthetic ? ...
    logs,
    logsBloom: isSynthetic ? ...
    root: isSynthetic ? ...
    status: isSynthetic ? ...
    transactionIndex: isSynthetic ? ...
    type: isSynthetic ? ...
  };
}
```

Or should we keep it separate for readibility?

quiet-node (approved):
LGTM Great work!

PR #4936, opened 2026-02-17
fix: fixed cumulativeGasUsed calculations in eth_getBlockReceipts and eth_getTransactionReceipt (#4921)

### LINKED ISSUE #4921: Fix cumulativeGasUsed: use per-transaction cumulative gas, not block total
## Problem

`cumulativeGasUsed` in transaction receipts is currently set to the total gas used in the block (block_gas_used) for every receipt in that block. It should instead be the cumulative gas used in the block up to and including that transaction, as defined in the Ethereum Yellow Paper.

So for a block with 3 transactions using 1000, 2000, and 3000 gas:
- Tx 0: cumulativeGasUsed = 1000
- Tx 1: cumulativeGasUsed = 3000 (1000 + 2000)
- Tx 2: cumulativeGasUsed = 6000 (1000 + 2000 + 3000)

## Current behavior
Every receipt in the block gets the same value: the block’s total gas (e.g. 6000 for all three in the example above). This comes from using `receiptResponse.block_gas_used` (mirror node: “The total gas used in the block”) for `cumulativeGasUsed`.

### Steps to reproduce

1. Pick a block with multiple receipts, e.g. 0x56e7e72.
2. Call `eth_getBlockReceipts` for that block.
3. Observe that every receipt in the response has the same `cumulativeGasUsed` value, which equals the block’s total gas used, instead of increasing per transaction as defined in the Ethereum Yellow Paper.

### Additional context

_No response_

### Hedera network

_No response_

### Version

0.74.1 and main branch

### Operating system

None

BartoszSolkaBD:
### Description

This PR fixes an issue where cumulativeGasUsed in transaction receipts was incorrectly set to the block’s total gas used (block_gas_used) for every transaction in the block.

It updates the receipt generation logic so that cumulativeGasUsed reflects the cumulative gas consumed in the block up to and including each transaction, matching the Ethereum Yellow Paper semantics.

This change has been implemented for both eth_getBlockReceipts and eth_getTransactionReceipt.

It also updates the receipts root hash computation (and its expected test value) to use the corrected cumulative gas and adds targeted tests to validate multi‑transaction block behavior and OpenRPC contract for these endpoints.

### Related issue(s)

Fixes #4921 

### Testing Guide

1. Pick a block with multiple receipts, e.g. 0x56e7e72.
2. Call eth_getBlockReceipts for that block.
3. Observe that every receipt in the response has the correct cumulativeGasUsed value, which is increasing per transaction as defined in the Ethereum Yellow Paper.

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
+++ b/packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
@@ -195,7 +195,7 @@ function buildReceiptRootHashes(txHashes: string[], contractResults: any[], logs
 
     receipts.push({
       transactionIndex: transactionIndexHex,
-type: crPerTx && crPerTx.type != undefined ? intToHex(crPerTx.type) : null,
+      type: crPerTx && crPerTx.type != undefined ? intToHex(crPerTx.type) : null,
       root: crPerTx ? crPerTx.root : constants.ZERO_HEX_32_BYTE,
       status: crPerTx ? crPerTx.status : constants.ONE_HEX,
       cumulativeGasUsed: intToHex(cumulativeGas),


packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -132,27 +132,50 @@ function populateSyntheticTransactions(
...
-    let transactionIndex: any = null;
+  const items: {
+    transactionIndex: number;
+    logsPerTx: Log[];
+    crPerTx: any[];
+  }[] = [];
+
+  for (const txHash of txHashes) {
+    const logsPerTx: Log[] = logs.filter((log) => log.transactionHash === txHash);
+    const crPerTx: any[] = contractResults.filter((cr) => cr.hash === txHash);
```
jasuwienas: Don't we really know what is the type of crPerTx?

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -487,12 +493,39 @@ export class TransactionService implements ITransactionService {
       this.common.resolveEvmAddress(receiptResponse.to, requestDetails),
     ]);
 
+    let blockGasUsedBeforeTransaction = 0;
+    if (receiptResponse.transaction_index > 0) {
+      const params: IContractResultsParams = {
+        blockNumber: receiptResponse.block_number,
+      };
+
+      const blockContractResults = await this.mirrorNodeClient.getContractResults(requestDetails, params);
```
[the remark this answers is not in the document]
BartoszSolkaBD: https://docs.hedera.com/api-reference/contracts/list-contract-results-from-all-contracts-on-the-network#parameter-transaction-index - there is an option to even query by transaction index `x >= 0`. A thing to discuss is, if we want to ask mirror node team to calculate this value on their side, or we do it ourselves and optimize what we can

packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
jasuwienas: Sure, it’s not that important, it just crossed my mind while I was checking the PR. We can address it in a separate issue, we already have this problem in the code.

packages/relay/src/lib/services/ethService/transactionService/TransactionService.ts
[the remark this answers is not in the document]
jasuwienas: Honestly, I didn’t anticipate all of these implications when we were preparing the initial issue. What I do know is that fetching the block can be a very heavy operation (and I think it happens far less frequently than fetching a transaction receipt).
Including this could have significant performance implications with potentially limited benefit. I think we should all consider the options:
1. Ignore this value for this particular endpoint - worst solution imo, since it would make the same transaction look different depending on where it’s queried.
2. Accept the performance impact (investigate how significant it actually is) and proceed with the full implementation. We should also try to filter the results as much as possible. If filtering by transaction index works (assuming it’s the same transaction.index we’re referring to: I’m not entirely sure, as it sometimes appears to come from logs) or timestamps, we should do it deffinitely.
3. (Keep the current “fake” value for cumulativeGasUsed - but clearly document that behavior and do this operation on the MN side.)

packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -372,7 +395,15 @@ export async function getBlockReceipts(
       logsByHash.set(log.transactionHash, existingLogs);
     }
 
-    const receiptPromises = contractResults.map(async (contractResult) => {
+    // Ensure contract results are processed in transaction_index (block) order
+    const sortedContractResults = [...contractResults].sort((a, b) => {
+      const aIdx = a.transaction_index ?? Number.MAX_SAFE_INTEGER;
```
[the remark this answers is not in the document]
jasuwienas: Ok, I get it now, synthetics dont ever influnce this value since they always use 0 gas.In this case I don't think we should take them into consideration here.

BUT we need to set this cumulativeGasPrice for the synthetic receipts generated here:
https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4936/changes/BASE..586dec6a6ba4dd355f07abae9d2425755e928503#diff-b8e1972336c5bbc108e68c01afa6bf2999c059bf3ea9ee67642d8a906f57a45eR444

as well. Which we aren't doing now, right?

packages/relay/src/lib/factories/transactionReceiptFactory.ts
```diff
@@ -104,7 +105,10 @@ class TransactionReceiptFactory {
       blockNumber: numberTo0x(receiptResponse.block_number),
       from: from,
       to: to,
-      cumulativeGasUsed: numberTo0x(receiptResponse.block_gas_used),
+      cumulativeGasUsed:
+        blockGasUsedBeforeTransaction != null
+          ? numberTo0x(blockGasUsedBeforeTransaction + receiptResponse.gas_used)
+          : constants.ZERO_HEX,
       gasUsed: nanOrNumberTo0x(receiptResponse.gas_used),
```
quiet-node: In what scenario a contract result receipt has zero hex as cumulativeGasUsed? Should we make it the new required field as we're fixing the issue?

packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -132,27 +132,50 @@ function populateSyntheticTransactions(
...
+    const { transactionIndex, logsPerTx, crPerTx } = item;
+
+    const gasUsed = crPerTx[0]?.gas_used ?? 0;
+    cumulativeGas += gasUsed;
+    const transactionIndexHex = intToHex(transactionIndex);
+
     receipts.push({
-      transactionIndex,
+      transactionIndex: transactionIndexHex,
       type: crPerTx.length && crPerTx[0].type ? intToHex(crPerTx[0].type) : null,
```
quiet-node: Hmmm I know this is not introduced in this PR but this seems like a bug, type could be 0, and 0 is falsy in Javascript, so the condition evaluates to `null` instead of the correct type, right?

quiet-node (commented):
LG couple of items

packages/relay/src/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -131,28 +131,81 @@ function populateSyntheticTransactions(
...
+    contractResultsByHash.set(cr.hash, list);
+  }
 
-  for (const i in txHashes) {
-    const txHash: string = txHashes[i];
-    const logsPerTx: Log[] = logs.filter((log) => log.transactionHash == txHash);
-    const crPerTx: any[] = contractResults.filter((cr) => cr.hash == txHash);
+  for (const txHash of txHashes) {
+    const logsPerTx: Log[] = logsByTxHash.get(txHash) ?? [];
+    const crPerTx: any[] = contractResultsByHash.get(txHash) ?? [];
```
quiet-node: nit: should not need the extra casts but should be able to auto inferred, correct?

quiet-node (dismissed):
LGTM

jasuwienas (dismissed):

PR #4957, opened 2026-02-20
feat: prepare initial version of openjson rpc file (#4915)

### LINKED ISSUE #4915: [CONFORMITY] Align our openrpc.json updating scriptt the latest Ethereum Spec + Extend the existing conformity check process
**Problem**

Our conformity tests are not being able to capture all the new updated to the Ethereum Json RPC Spec, and neither distinguish what changes are related to methods that we will no impllement as they are not Hiero compatible or which ones are updates that perhaps we have a customization, we need to be aware of.

Detected Details

- We are pinning to a given commit, which makes our foto static
- We do have a unique skip list

**Solution**

 (*) Point to the latest commit possible

- Check differences, and identify actions needed to be 1to1 in sync. Could be new User stories to add any new changes needed.

 (*) Create several skip lists.

- NON_SUPPORTED_ SKIP_LIST = methods that we will not support. Skip always

- OVERRIDEN_SKIP_LIST = methods that we do support , but we hardcore/adjust behavior. Any new change should be marked as a comment

- NOT_YET_SUPPORTED_ SKIP_LIST = methods that we will support, but we do not yet, due prioritization. skip always

- FORK_NOT_YET_SUPPORTED_ SKIP_LIST = methods that we will support, but we do not yet, due fork is not yet supported. skip always

 (*) Change GitHub action to start once a week
- Remove current on main merge behavior to a cronjob based action, that executed every friday, and opens PR on change detection

 (*) Add a Support Role action to check every week the Github action execution
- Add as section for conforimity tests, to review weekly the action execution and possible PR
- Review weekly if we need to change the pinned commit

jasuwienas:
### Description

Prepare initial version of the openrpc json scheam file. I've compared our openjson rpc with the one generated by openapi and extracted only the suggestion which I think we should apply...

### Related issue(s)

Fixes #4915

### Testing Guide

1. Go through PR
2. Let me know your thoughts about proposed approach for each of the changes.
3. If we all agree we can merge.

N/A

N/A

### Checklist

### DIFF
--- a/docs/openrpc.json
+++ b/docs/openrpc.json
@@ -894,7 +894,7 @@
       "result": {
         "name": "Filter Identifier",
         "schema": {
-          "$ref": "#/components/schemas/uint"
+          "$ref": "#/components/schemas/hash16"
         }
       },
       "tags": [
@@ -918,7 +918,7 @@
       "result": {
         "name": "Filter Identifier",
         "schema": {
-          "$ref": "#/components/schemas/uint"
+          "$ref": "#/components/schemas/hash16"
         }
       },
       "tags": [
@@ -935,7 +935,7 @@
         {
           "name": "Filter Identifier",
           "schema": {
-            "$ref": "#/components/schemas/uint"
+            "$ref": "#/components/schemas/hash16"
           }
         }
       ],
@@ -974,7 +974,7 @@
         {
           "name": "Filter Identifier",
           "schema": {
-            "$ref": "#/components/schemas/uint"
+            "$ref": "#/components/schemas/hash16"
           }
         }
       ],


docs/openrpc.json
```diff
@@ -57,9 +57,12 @@
         "name": "Accounts",
         "description": "Always returns an empty array",
         "schema": {
-          "title": "empty array",
+          "title": "Accounts - always empty",
           "type": "array",
-          "pattern": "^\\[\\s*\\]$"
+          "pattern": "^\\[\\s*\\]$",
+          "items": {
```
jasuwienas: Note for reviewers , I think we should either update the pattern or include the items component here. Note about this always being empty is included in otger field.

docs/openrpc.json
```diff
@@ -217,9 +220,7 @@
               "description": "An array of gas used ratio.",
               "type": "array",
               "items": {
-                "title": "floating point number",
-                "type": "number",
-                "pattern": "^([0-9].[0-9]*|0)$"
+                "$ref": "#/components/schemas/ratio"
```
jasuwienas: Instead of float pattern we have:

```
        "type": "number",
        "minimum": 0,
        "maximum": 1
```
It's essentially the same. With the difference that our previous version used to indicate this is the value between 0 and 9.9(9), not 0 and 1.

We don't really seem to be able return value bigger than 1 here, no matter for which block I query I always get 0.5:

```
curl -s https://testnet.hashio.io/api \
  -d '{
    "jsonrpc":"2.0",
    "method":"eth_feeHistory",
    "params":[
      "0x5",
      "latest",
      [10,20,30]
    ],
    "id":1
  }'
```
```json
{"result":{"baseFeePerGas":["0xd18c2e2800","0xd18c2e2800"],"gasUsedRatio":[0.5],"oldestBlock":"0x1e4a9e7","reward":[["0x0","0x0","0x0"]]},"jsonrpc":"2.0","id":1}%                                           

```

Which seems to be confirmed in code:
```
      gasUsedRatio: Array(blockCount).fill(constants.DEFAULT_GAS_USED_RATIO),
```
tldr, we can just add note that it's always 0.5.

docs/openrpc.json
```diff
@@ -245,13 +246,14 @@
                 }
               }
             }
-          }
+          },
+          "additionalProperties": false
```
jasuwienas: When you include additional properties we return:

```
{"error":{"code":-32602,"message":"[Request ID: f6a2b507-b9c2-40bc-9759-3312400fc719] Invalid params"},"jsonrpc":"2.0","id":1}%                                                                              
```
So  we can include this in our schema too.

docs/openrpc.json
```diff
@@ -847,12 +865,11 @@
...
       "result": {
-        "name": "eth_maxPriorityFeePerGas result",
+        "name": "Max priority fee per gas",
         "schema": {
           "description": "Always returns '0x0'. Hedera doesn't have a concept of tipping nodes to promote any behavior",
-          "title": "hex encoded unsigned integer",
-          "type": "string",
-          "pattern": "0x0"
+          "title": "Max priority fee per gas",
+          "$ref": "#/components/schemas/uint"
```
jasuwienas: Same as in other places - we already stated that the fee is always 0x0. And 0x0 is compatible with uint too, why not use default docs.

natanasow (approved):

docs/openrpc.json
```diff
@@ -473,10 +474,10 @@
           }
         },
         {
-          "name": "Position",
+          "name": "Storage slot",
           "required": true,
           "schema": {
-            "$ref": "#/components/schemas/uint"
+            "$ref": "#/components/schemas/uint256"
```
Neurone: 'uint' is an alias for 'uint256' and they should not be defined with different syntaxes. We should have only `uint` or only `uint256` definitions in this OpenJSON RPC spec, not both. I always prefer to be explicit, so I'm in favour of using `uint256` for everything. If we don't want to change that, we can use `uint` here and change the regEx definition to support syntaxes like `0x0001`, etc.

docs/openrpc.json
```diff
@@ -892,16 +909,16 @@
...
             "$ref": "#/components/schemas/LogFilter"
           }
         }
       ],
       "result": {
-        "name": "Filter ID",
+        "name": "Filter Identifier",
         "schema": {
-          "$ref": "#/components/schemas/hash16"
+          "$ref": "#/components/schemas/uint"
```
Neurone: `hash16` or `bytes16` seems more appropriate. `uint` means `uint256`, that does not seem to apply in this case.

docs/openrpc.json
jasuwienas: Yes, this is a good example of what I meant  and why I was insisting on changing the standard as little as possible.

We have a script that fetches the latest changes from `ethereum/execution-apis` and compares them with our version (for now I want us to stay aligned with `ethereum/execution-apis@v1.0.0-beta.4`). If we keep standardopenrpc schema  definition here, the comparison script will detect a difference and automatically create a new PR proposing to replace it with `bytesMax32` (assuming everything works as expected).

The alternative would be to apply our own documentation/standard for this field and mark it as “ignore differences” in the comparison script (see PR: [https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4963](https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/4963)). But then we would miss this update in the standard assuming that we are just incompatible here anyway (which is not the case).That’s one of the main reasons I’d prefer to stick to the official standard as closely as possible.

PR #4998, opened 2026-02-27
feat: allow mutliple paymaster accounts, and smart contracts per paymaster

### LINKED ISSUE #4977: [FEATURE] Extended Paymaster: Allow mutliple paymaster accounts, and smart contracts per paymaster
**Problem**

As a relay provider, to support paymaster for two separate projects, an instance per project needs to be deployed, not being this very  cost effective, and introduces more management.

As as relay provider i want to be able to configure multiple paymaster accounts, and smart contract whitellst per account

**Solution**

We will not fully depend on the relay - `OPERATOR_ID_MAIN` `OPERATOR_KEY_MAIN` values for paymaster

Payer can be either operator or a new accounts set under. 

- PAYMASTER_ACCOUNTS= [[Account, type, key. allowance]. [Account, type, key, allowance]]
- PAYMASTER_WHITELIST is for the default operator only
- PAYMASTER_ACCOUNTS_WHITELISTS =[[Account:[contractA, contractB]],[Account:[contractA, contractB]]]

**DOD**

- Operator account based paymaster still works as expected
- New accounts can be configured and work in parallel to operator account based paymaster
- Providers can configure only non operator accounts paymaster, and  the functionality works
- Existing documentations is updated and Experimental is removed from the feature description
- A smart contract cannot be included in more than 1 paymaster. In case of duplicates the last paymaster valie configured, is the paymaster

natanasow:
### Description

As a relay provider, to support paymaster for two separate projects, an instance per project needs to be deployed, not being this very  cost effective, and introduces more management.

As a relay provider, I want to be able to configure multiple paymaster accounts, and smart contract whitelists per account

### Solution

We will not fully depend on the relay - `OPERATOR_ID_MAIN` `OPERATOR_KEY_MAIN` values for paymaster

Payer can be either an operator or a new account set under: 
- `PAYMASTER_ACCOUNTS`= [[account, type, key, allowance], [account, type, key, allowance]]
- `PAYMASTER_WHITELIST` is for the default operator only
- `PAYMASTER_ACCOUNTS_WHITELISTS` = [[account, [contractA, contractB]], [account, [contractA, contractB]]]

**DOD**

- Operator account-based paymaster still works as expected
- New accounts can be configured and work in parallel to the operator account-based paymaster
- Providers can configure only non-operator accounts paymaster, and  the functionality works
- Existing documentation is updated ,and Experimental is removed from the feature description
- A smart contract cannot be included in more than 1 paymaster. In case of duplicates, the last paymaster value configured is the paymaster

### Related issue(s)

Fixes #4977 

### Testing Guide

1.
2.
3.

N/A

N/A

### Checklist

### DIFF
--- a/packages/relay/tests/lib/services/ethService/ethCommonService/commonService.spec.ts
+++ b/packages/relay/tests/lib/services/ethService/ethCommonService/commonService.spec.ts
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { ConfigService } from '@hashgraph/json-rpc-config-service/dist/services';
+import { expect } from 'chai';
+import sinon from 'sinon';
+
+import { prepend0x } from '../../../../../dist/formatters';
+import { CommonService } from '../../../../../dist/lib/services';
+
+describe('CommonService', () => {
+  describe('getPaymasterIfTxCanBeSubsidized', async () => {
+    let configStub: sinon.SinonStub;
+
+    beforeEach(() => {
+      // reset maps before each test
+      (CommonService as any).PAYMASTER_ACCOUNTS_WHITELISTS_MAP = new Map();
+      (CommonService as any).PAYMASTER_ACCOUNTS_MAP = new Map();
+
+      configStub = sinon.stub(ConfigService, 'get');
+    });
+
+    afterEach(() => {
+      sinon.restore();
+    });
+
+    it('should return paymaster account when toAddress is whitelisted in PAYMASTER_ACCOUNTS_WHITELISTS_MAP', () => {
+      const toAddress = '0x0000000000000000000000000000000000000000';
+      const normalized = prepend0x(toAddress.toLowerCase());
+
+      (CommonService as any).PAYMASTER_ACCOUNTS_WHITELISTS_MAP.set(normalized, '0.0.9303');
+      (CommonService as any).PAYMASTER_ACCOUNTS_MAP.set('0.0.9303', ['0.0.9303', null, null, 100]);
+
+      const result = CommonService.getPaymasterIfTxCanBeSubsidized(toAddress);
+
+      expect(result).to.deep.equal({
+        accountId: '0.0.9303',
+        gasAllowance: 100,
+      });
+    });
+
+    it('should return null when whitelisted accountId exists but PAYMASTER_ACCOUNTS_MAP has no entry', () => {
+      const toAddress = '0x0000000000000000000000000000000000000000';
+      const normalized = prepend0x(toAddress.toLowerCase());
+
+      (CommonService as any).PAYMASTER_ACCOUNTS_WHITELISTS_MAP.set(normalized, '0.0.9303');
+
+      configStub.withArgs('PAYMASTER_ENABLED').returns(false);
+
+      const result = CommonService.getPaymasterIfTxCanBeSubsidized(toAddress);
+
+      expect(result).to.equal(null);
+    });
+
+    it('should return default paymaster when PAYMASTER_ENABLED and whitelist contains *', () => {
+      configStub.withArgs('PAYMASTER_ENABLED').returns(true);
+      configStub.withArgs('PAYMASTER_WHITELIST').returns(['*']);
+      configStub.withArgs('OPERATOR_ID_MAIN').returns('0.0.1000');
+      configStub.withArgs('MAX_GAS_ALLOWANCE_HBAR').returns(120);
+
+      const result = CommonService.getPaymasterIfTxCanBeSubsidized('0xdef');
+
+      expect(result).to.deep.equal({
+        accountId: '0.0.1000',
+        gasAllowance: 120,
+      });
+    });
+
+    it('should return default paymaster when address is in PAYMASTER_WHITELIST', () => {
+      const toAddress = '0x0000000000000000000000000000000000000000';
+      const normalized = prepend0x(toAddress.toLowerCase());
+
+      configStub.withArgs('PAYMASTER_ENABLED').returns(true);
+      configStub.withArgs('PAYMASTER_WHITELIST').returns([normalized]);
+      configStub.withArgs('OPERATOR_ID_MAIN').returns('0.0.1000');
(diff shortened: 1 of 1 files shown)


packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
```diff
@@ -627,20 +657,61 @@ export class CommonService implements ICommonService {
...
+   * Returns `null` if the transaction is not eligible for paymaster subsidization.
    */
-  public static isSubsidizedTransaction(toAddress: string | null): boolean {
-    const payMasterWhiteList = ConfigService.get('PAYMASTER_WHITELIST').map((e) => e.toLowerCase());
+  public static getPaymasterIfTxCanBeSubsidized(
+    toAddress: string | null,
+  ): { accountId: string; gasAllowance: number } | null {
+    // handle default paymaster functionality
+    if (ConfigService.get('PAYMASTER_ENABLED')) {
+      const paymasterWhitelist = ConfigService.get('PAYMASTER_WHITELIST').map((e) => e.toLowerCase());
```
jasuwienas: I know it's inherited so we can keep it as is, but I don't think that we should be formatting this whitelist for each transaction anew. We should initialize it once and keep it as is.

Ferparishuertas (dismissed):
LGTM

packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
```diff
@@ -627,20 +657,61 @@ export class CommonService implements ICommonService {
...
-      (payMasterWhiteList.includes('*') ||
-        (toAddress && payMasterWhiteList.includes(prepend0x(toAddress.toLowerCase()))))
-    );
+    // handle contract deployment
+    if (!toAddress) {
+      return null;
+    }
+
+    // handle paymaster accounts
+    const paymasterAccountId = CommonService.PAYMASTER_ACCOUNTS_WHITELISTS_MAP.get(prepend0x(toAddress.toLowerCase()));
```
jasuwienas: We will use this code even if PAYMASTER_ENABLED = false (previous implementation). It's a bit confusing...

packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
```diff
@@ -627,20 +657,61 @@ export class CommonService implements ICommonService {
...
+   *
+   * @returns An object containing:
+   * - `accountId`: The paymaster account ID to be used for subsidizing gas.
+   * - `gasAllowance`: The maximum gas allowance (in HBAR) provided by the paymaster.
+   *
+   * Returns `null` if the transaction is not eligible for paymaster subsidization.
    */
-  public static isSubsidizedTransaction(toAddress: string | null): boolean {
-    const payMasterWhiteList = ConfigService.get('PAYMASTER_WHITELIST').map((e) => e.toLowerCase());
+  public static getPaymasterIfTxCanBeSubsidized(
```
simzzz: Maybe it would be good to add a unit test for this method in particular as it seems to be the core routing logic with multiple code paths.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -77,6 +88,45 @@ export class SDKClient {
...
+    this.initPaymastersClients();
+  }
+
+  /**
+   * Init paymaster clients
+   *
+   * @private
+   */
+  private initPaymastersClients(): void {
+    this.paymasterClients = new Map(
```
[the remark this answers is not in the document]
jasuwienas: Why Im asking: I know we already had some issues with unclosed conecctions, now we will ahve as many of them as paymaster configurations... If there is a rarely used paymaster maybe it would be good to close it's client when not used and create it on-demand.

packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
```diff
@@ -627,20 +657,61 @@ export class CommonService implements ICommonService {
...
+    }
+    const paymasterAccount = CommonService.PAYMASTER_ACCOUNTS_MAP.get(paymasterAccountId);
+    if (!paymasterAccount) {
+      return null;
+    }
+    const [accountId, , , gasAllowance] = paymasterAccount;
+
+    return {
+      accountId,
+      gasAllowance: Number(gasAllowance),
```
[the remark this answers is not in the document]
jasuwienas: Also, there’s no validation or startup configuration initialization. Misconfigurations resulting in NaN's here will happen runtime when processing the trasnaction... It will be difficult to debug.

packages/relay/src/lib/clients/sdkClient.ts
```diff
@@ -63,6 +63,17 @@ export class SDKClient {
    */
   private readonly hbarLimitService: HbarLimitService;
 
+  /**
+   * Map of accountId - Client for each paymaster defined in PAYMASTER_ACCOUNTS,
+   * and the default operator client if the default paymaster functionality is enabled
+   */
+  private paymasterClients!: Map<string, Client>;
```
jasuwienas: We are never reassinging a value here, and we are creating it in constructor, why arent we staying with :
```
private readonly paymasterClients: Map<string, Client>;
```
like in the other params?

packages/relay/src/lib/clients/sdkClient.ts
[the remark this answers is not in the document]
natanasow: Keeping the connection alive is the SDK's responsibility, and I think they fixed all of the reported issues about that in the latest version.

Also, the whole idea behind adding a map was to skip on-demand and dynamic clients initilization for gaining maximum performance.

packages/relay/src/lib/services/ethService/ethCommonService/CommonService.ts
quiet-node: I agree this should be added in the follow-up as a hot path priority. Currently this method is used for both prechecks and later for executeTransaction(), so it needs to be kept as lean as possible.

Ferparishuertas (approved):
LGTM

PR #5044, opened 2026-03-05
feat: return complete logs for synthetic transactions in eth getblockreceiptseth gettransactionreceipt (#5043)

### LINKED ISSUE #5043: Return complete logs for synthetic transactions in eth_getBlockReceipts/eth_getTransactionReceipt
### Problem

We received a report of an apparent discrepancy between the number of log events visible via Mirror Node APIs / Hashscan vs what the JSON-RPC Relay returns from eth_getBlockReceipts for block 0x578A104.

Example:

 - Block: 0x578A104
 - Example transaction in the block (Hashscan): https://hashscan.io/mainnet/transaction/1772296225.720230615
 - Mirror Node logs query around the block timestamp shows 3 events: https://mainnet.mirrornode.hedera.com/api/v1/contracts/results/logs?timestamp=lte:1772296225.765865240&timestamp=gte:1772296224.716414000&limit=100&order=asc

JSON-RPC Relay call:
```
curl -s -X POST -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":"2","method":"eth_getBlockReceipts","params":["0x578A104"]}' \
  https://mainnet.hashio.io/api | jq .
```
The relay returned a receipt with only 1 log in result[].logs (logIndex 0x0) while MN/Hashscan indicated 3.

In the synthetic receipt creation logic we appear to only include the first log (https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/d6a36d324a65453d0e569e66caf529ce0ccf1be3/packages/relay/src/lib/factories/transactionReceiptFactory.ts#L57)

we currently grab only the first log:

logs: [syntheticLogs[0]]

This is because we assumed that for single HTS transfer there is only 1 log emitted always (like ERC20 transfer). But here we have cryptoTransfer call which result with multiple transfers triggered (and events).

### Solution

Do not exclude non-first log received from MN from the response.

### Alternatives

_No response_

jasuwienas:
### Description

Keep all of the synthetic logs in the transction response.

### Related issue(s)

Fixes #5043

### Testing Guide

1. Take a look at tests, verify them yourself (you can curl mainnet examples to confirm)

N/A

N/A

### Checklist

### DIFF
--- a/packages/server/tests/acceptance/fixtures/multiLogReceiptFixture.ts
+++ b/packages/server/tests/acceptance/fixtures/multiLogReceiptFixture.ts
@@ -47,6 +47,10 @@ export class MultiLogReceiptFixture {
       this.createToken('MultiLog2 Testing Token', 'MLTT2'),
     ]);
     await this.associate(tokenIds, recipientId, recipientKey);
+
+    // Wait for all the previously queued transactions to be already mined (removing flakiness)
+    await new Promise((r) => setTimeout(r, 3000));
+
     const record = await this.executeCryptoTransfer(tokenIds, recipientId);
     const blockId = await this.fetchBlockNumberWithRetries(record.consensusTimestamp.toString());
     return `0x${Number(blockId).toString(16)}`;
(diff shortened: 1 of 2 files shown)


packages/relay/tests/lib/eth/eth_getTransactionReceipt.spec.ts
```diff
@@ -256,6 +256,27 @@ describe('@ethGetTransactionReceipt eth_getTransactionReceipt tests', async func
     expect(receipt.logsBloom).to.eq(emptyBloom);
   });
 
+  it('handles bloom for transaction with multiple synthetic transaction logs', async function () {
```
quiet-node: Would it possible to add an acceptance test for this as well?

PR #5059, opened 2026-03-09
chore: optimize Dockerfile using standalone multi-stage build and Node File Trace technique

### LINKED ISSUE #4986: Dockerfile Optimization for Memory Efficiency and Runtime Reliability
## Problem Summary

The current Docker deployment of the Hedera JSON-RPC Relay is inefficient in both memory utilization and signal handling due to its execution model. Specifically, the use of `npm` as the entrypoint creates a process chain that introduces significant memory overhead and prevents the application from responding correctly to container orchestration signals.

---

## Current Behavior

The existing `Dockerfile` relies on an execution chain managed by `npm`:

```dockerfile
# Current Entrypoint
ENTRYPOINT ["npm", "run"]
CMD ["start"]
```

When the container starts, it spawns a hierarchy of processes:

1. `npm run start` (from root and PID 1)
2. `sh -c npx lerna exec --scope @hashgraph/json-rpc-server -- npm run start`
3. `npm exec lerna exec ...`
4. `sh -c lerna exec ...`
5. `node /home/node/app/node_modules/.bin/lerna exec ...`
6. `/bin/sh -c npm run start`
7. `npm run start` (inside packages/server)
8. `sh -c node dist/index.js`
9. `node dist/index.js` (The actual Relay app at PID 65)

### 1. The "Wrapper Tax" (Memory Overhead)

This chain of idle manager processes (`npm`, `sh`, `lerna`) consumes approximately **~65MB of Resident Set Size (RSS)** before the application logic even begins to execute. In high-density environments or when running with strict resource quotas (e.g., **128MB** limits), this overhead accounts for **>50% of the total budget**.

This frequently leads to:

- **Premature OOMKills**: The container is killed during startup because the "wrapper tax" leaves insufficient headroom for Node.js to initialize its heap.
- **Wasted Infrastructure Cost**: Requiring 256MB+ RSS just to account for "manager overhead" increases the total cost of ownership for large-scale relay deployments.

### 2. Signal Handling and PID 1 Issues

In Docker, the process defined in the `ENTRYPOINT` becomes **PID 1**. PID 1 has special responsibilities for handling OS signals (e.g., `SIGTERM`, `SIGINT`).

- Because `npm` is PID 1, the actual `node` application is a "grandchild" process.
- `npm` does not proactively forward signals to its children.
- **The Result**: When Kubernetes tries to stop the container, the application never receives the `SIGTERM`. It fails to perform graceful shutdowns (closing database connections, flushing logs, etc.) and is eventually hard-killed (`SIGKILL`) after the termination grace period, leading to potential state inconsistency.

---

### 1. Direct Node Execution (PID 1 Fix)

Refactor the entrypoint to bypass the `npm` wrapper and call the built JavaScript directly:

```dockerfile
ENTRYPOINT ["node"]
CMD ["packages/server/dist/index.js"]
```

By making `node` the PID 1 process, we instantly reclaim a significant chunk of resources and ensure the application natively handles lifecycle signals for graceful termination.

### 2. Multi-Stage Build and Image Resizing

The current image size is approximately **~900MB** because it retains build-time dependencies, the `.npm` cache, and raw source code.

- **Recommendation**: Implement a multi-stage `Dockerfile`.
- **Strategy**: Perform the build in a `build` stage, and then copy only the `dist/` folders and production `node_modules` into a `slim` or `alpine` runtime image.
- **Goal**: Reduce image size, speeding up deployment pull times and reducing the security attack surface.

quiet-node:
### Description

This PR optimizes the Docker deployment of the Hedera JSON-RPC Relay to improve memory efficiency and ensure reliable signal handling during container orchestration. Key changes include:

- Transitioned to a multi-stage Dockerfile (Node 22-alpine) to separate the build environment from the runtime image. This excludes development dependencies and build tools from the final image, making it significantly smaller and more secure.

- Introduced a new `scripts/build-standalone.js` script using `@vercel/nft` to isolate only the exact runtime dependencies needed. This generates the most optimal standalone release build and, combined with the multi-stage build, reduces the image size from 900MB to around 175MB, an 80% reduction.

- Updated the `ENTRYPOINT` to run the application directly via `node`. This removes the `npm`, `learn`, and `sh` wrapper overhead, reclaims idle RSS, and ensures `SIGTERM` signals are handled correctly for graceful shutdowns.

- Updated `docker-compose.yml` and the WebSocket Helm chart to work with the new direct-node entrypoint. The `.github/workflows/image-build.yml` workflow has also been enhanced to run automated liveness tests for both HTTP and WebSocket servers using the new Docker image structure.

### Related issue(s)

Fixes #4986

### Testing Guide

1. **Build the Docker Image:**
   Run `docker build . -t hedera-json-rpc-relay:latest` to ensure the new multi-stage build completes successfully.
2. **Verify Image Size:**
   Check the size of the generated image using `docker images hedera-json-rpc-relay:latest`. It should be significantly smaller than previous versions.
3. **Run via Docker Compose:**
   Start the relay and WebSocket server using `docker-compose up -d`. Verify that both services start correctly and respond to health checks.
4. **CI Validation:**
   The updated `.github/workflows/image-build.yml` automatically validates the build, starts both HTTP and WS servers, and performs liveness checks.

### Checklist


Dockerfile
```diff
@@ -1,30 +1,75 @@
...
+COPY scripts/ scripts/
+
+# Copy OpenRPC document for runtime access by /openrpc endpoint.
+COPY docs/openrpc.json docs/
+
+# Compile packages.
+RUN npm run build
+
+# Stores npm_package_version required by ConfigService.
+RUN node -p "'npm_package_version=' + require('./package.json').version" > .env.release
```
quiet-node: **notes**: This is necessary because `npm_package_version` is no longer automatically populated by npm. Since the new version runs Node directly as PID 1, the `npm_package_version` variable must now be set manually.

charts/hedera-json-rpc-relay-websocket/templates/deployment.yaml
```diff
@@ -29,7 +29,8 @@ spec:
       {{- end }}
       containers:
       - name: {{ .Chart.Name }}
-        args: ["start:ws"]
+        command: ["node", "--env-file=/home/node/app/.env.release"]
```
simzzz: If a user mounts a custom .env, this could conflict. Also the --env-file flag will fail if the file doesn't exist. Do you think that could be an issue? If so maybe we can make it configurable?

Issue #5091, opened 2026-03-13
Make sure FilterId validation check the uint type instead of bytes


jasuwienas:
Follow-up for an issue: https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/4995

Ensure filter identifier is always proper uint format

We correctly made it return uint variable instead of bytes32. But this filter id is later used as an input param in future requests. We have to correctly validate its NEW type.

Sinc eup till now we used to return 0 prefixed filter id (to always have 16 bytes), we should still be able to return a filter which used to be prefixed with these 0s (backward compatibility).

PR #5094, opened 2026-03-16
feat: add support for correct 16bytes 0prefixed filterid format (#5091)

### LINKED ISSUE #5023: baseFeePerGas should be always 0
According to https://hips.hedera.com/hip/hip-415, baseFeePerGas should be always 0. The reasoning for this, is that transactions within the same block can have different gasPrice set.

There is actually a current and next in the exchangerate with an expiration time that indicates when it expires. But a block could span that expiration time.

In any case, on Hedera we are not burning fees, and we should no represent this

**Solution**

Ensure that contains always baseFeePerGas = 0, on a harcoded way, and not calculating it on the fly, which is BTW, totally off.

**Solution**
baseFeePerGas is always = 0x0
Extend the conformity tests to ensure that at block level we always comply to hip-415 ( ReceiptsRoot will be updated on a new  HIP. We need to consider that we are doing it right in the relay now for receiptroot)
- identify the methods where this value are exposed
- propose the HIP spec testing
- implement it

jasuwienas:
### Description

After switching to the new solution with uint filter id (compatible with json rpc starndard behaviour) we should still allow eople to submit 0 prefixed value to make the solution backward-compatible.

### Related issue(s)

Fixes #5023

### Testing Guide

1. Run new acceptance test checking if both formats of the submitted value works correctly.

N/A

N/A

### Checklist

### DIFF
--- a/packages/server/tests/acceptance/rpc_batch1.spec.ts
+++ b/packages/server/tests/acceptance/rpc_batch1.spec.ts
@@ -2605,7 +2605,7 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
         const filter = { fromBlock: 'latest', toBlock: 'latest' };
         const createUintFilterIdWithLessThan16Bytes = async () => {
           for (let attempt = 0; attempt < 200; attempt++) {
-            // Each attempt has ~10% of success rate.
+            // Each attempt has 10% of success rate.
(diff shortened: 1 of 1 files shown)


quiet-node (approved):
LG. But PR title seems to have a typo, `currect`. Let's fix it before merging.

PR #5095, opened 2026-03-16
feat: determine blockGasLimit based on hapi version

### LINKED ISSUE #5065: BlockGasLimit per Block and Release
### **PROBLEM**

At release 0.69, we introduced opsDuration which allows higher gaslimit per block, estimated at 150M

As we change implicitly the gaslimit , we should set the blockgaslimit based on releae

### **SOLUTION**

Define a configurable logic, that is able to set blockgaslimit per release range. This configuration is hardcoded, and not overridable by config

Something like this

0.0.0 = 30000000
0.69.0 = 15000000

Which internally converts it to lower and upper boundaried

0.0.0-0.69.0 
0.69.0 -*

The defined value will be automaticalluy set  to eth_getBlockByNumber and eth_getBlockByHash removing the hardcoded vakue to 30M

Mirror node block endpoint shared the release where the block was generated

### **DOD**

- All tests pass
- eth_getBlockByNumber and eth_getBlockByHash return 30 o 150M based on the blocknumber/release number
- We need to be prepapred for 1, 2 . o more than 2 values in the range calculation, but in reality we have only 2 values now.

0.0.0-0.69.0  30M
0.69.0 -* 150M

example

0.0.0-0.69.0 
0.69.0 - 0.70.0

this leads to

<0.69.0 30M
0.69.0 -<0..70.0 150M
>=0.70.0

bootcodes:
### Description

This PR fixes the issue to return the correct block `gasLimit` depending on the hapi version.

### Related issue(s)

Fixes #5065

### Testing Guide

1. `npx lerna exec --scope @hashgraph/json-rpc-relay -- npx ts-mocha tests/lib/config/blockGasLimit.spec.ts --exit`
2.
3.

N/A

N/A

### Checklist


jasuwienas (commented):
Looks good (should work as intended) , just want to make sure about one thing.

PR #5134, opened 2026-03-23
feat: align relay to mirror node returing weibars (#4901)

### LINKED ISSUE #4901: maxFeePerGas and maxPriorityFeePerGas returned in tinybars instead of weibars for type 0x2 transactions
### Description

`eth_getTransactionByHash` (and related methods) returns `maxFeePerGas` and `maxPriorityFeePerGas` in **tinybars** for type 0x2 (EIP-1559) transactions, while `effectiveGasPrice` and `baseFeePerGas` are correctly returned in **weibars**. This unit mismatch breaks all Ethereum tooling that assumes consistent wei units across gas price fields.

This is the same class of bug that #3053 reported for the `value` field. PR #3080 fixed `value` by wrapping it with `tinybarsToWeibars()`, but `maxFeePerGas` and `maxPriorityFeePerGas` were not converted.

### Steps to reproduce

Query any type 0x2 transaction and compare fields:

```bash
# Transaction object — maxFeePerGas is in tinybars
curl -s -X POST https://testnet.hashio.io/api \
  -d '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["<any_type2_tx_hash>"],"id":1}'

# Receipt — effectiveGasPrice is in weibars
curl -s -X POST https://testnet.hashio.io/api \
  -d '{"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":["<any_type2_tx_hash>"],"id":1}'
```

### Example (from a private Hedera network, chain ID 8901)

**`eth_getTransactionByHash`:**
```json
{
  "type": "0x2",
  "maxFeePerGas": "0x656",
  "maxPriorityFeePerGas": "0x656",
  "gasPrice": "0x0"
}
```

`0x656` = 1,622 — this is in **tinybars**.

**`eth_getTransactionReceipt`:**
```json
{
  "type": "0x2",
  "effectiveGasPrice": "0xeb98732ec00"
}
```

`0xeb98732ec00` = 16,190,000,000,000 wei = 16,190 gwei — this is in **weibars**.

The correct `maxFeePerGas` in weibars should be `1622 * 10_000_000_000 = 16,220,000,000,000` (same order of magnitude as `effectiveGasPrice`). Instead it's returned as raw `1622`.

**`eth_getBlockByNumber` (latest block):**
```json
{
  "baseFeePerGas": "0xeb98732ec00"
}
```

Also in weibars (16,190 gwei), consistent with receipts but inconsistent with tx object fields.

### Impact

Any tooling that computes priority fee as `maxFeePerGas - baseFeePerGas` gets a massively negative result:

```
priority_fee = (1,622 - 16,190,000,000,000) * 21,000 = -339,989,999,965,938,000
```

**Affected downstream tools:**
- **Blockscout** gas price oracle displays negative gas prices (e.g. `-1,977 gwei`)
- **TheGraph** and other indexers that compute priority fees
- Any EVM client library that cross-validates gas fields

**Affected RPC methods:**
- `eth_getTransactionByHash`
- `eth_getTransactionByBlockHashAndIndex`
- `eth_getTransactionByBlockNumberAndIndex`
- `eth_getBlockByHash` (with full transactions)
- `eth_getBlockByNumber` (with full transactions)

### Root cause

In `packages/relay/src/lib/factories/transactionFactory.ts` lines 123-124, `max_fee_per_gas` and `max_priority_fee_per_gas` from the mirror node API are passed through without conversion:

```typescript
return TransactionFactory.createTransactionByType(cr.type, {
    ...commonFields,
    maxPriorityFeePerGas: cr.max_priority_fee_per_gas,  // raw tinybars — NOT converted
    maxFeePerGas: cr.max_fee_per_gas,                   // raw tinybars — NOT converted
});
```

Compare with the `value` field on line 114, which was correctly fixed in PR #3080:

```typescript
value: nanOrNumberInt64To0x(tinybarsToWeibars(cr.amount, true)),  // correctly converted
```

### Suggested fix

Apply `tinybarsToWeibars()` (already exists in `formatters.ts` since PR #3080) to both fields:

```typescript
return TransactionFactory.createTransactionByType(cr.type, {
    ...commonFields,
    maxPriorityFeePerGas: tinybarsToWeibars(cr.max_priority_fee_per_gas),
    maxFeePerGas: tinybarsToWeibars(cr.max_fee_per_gas),
});
```

Tests in `eth_getTransactionByHash.spec.ts` that assert the raw tinybar values will need updating to expect weibar values.

### Hedera network

All networks (mainnet, testnet, previewnet, private networks)

### Version

v0.74.1 (latest release), also confirmed on `main` branch (v0.75.0-SNAPSHOT)

### Operating system

N/A (server-side relay bug)

BartoszSolkaBD:
### Description

This PR adds support for the mirror node hbar query parameter on contract results APIs so the relay can request raw weibar values (hbar=false) instead of tinybar-truncated monetary fields (hbar=true, default on mirror). That aligns with mirror node behavior where storing converted tinybars made it impossible to reconstruct original RLP for gasPrice, maxFeePerGas, maxPriorityFeePerGas, and value.

### Related issue(s)

Fixes #4901 

### Testing Guide

1. Unit tests - From repo root, run relay tests that cover transaction factory and eth_getTransactionByHash / eth_getTransactionByBlockHashAndIndex (e.g. the updated *.spec.ts files).
2. Use a mirror node that supports hbar=false on contract results. Confirm value, gasPrice, and EIP-1559 fee fields match raw RLP / expected wei values without double conversion.

N/A

N/A

### Checklist


jasuwienas (commented):
Overall it looks great, just want to clarify few things.

docs/configuration.md
```diff
@@ -87,6 +87,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
+| `MIRROR_NODE_CONTRACT_RESULTS_HBAR`                    | "true"                                                                                                                                                                                                   | Flag to set if gasPrice, maxFeePerGas, maxPriorityFeePerGas, and value should be returned in tinybars.                                                                                                                                                                                                                                                                                                                                                                                                             |
```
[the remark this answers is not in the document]
BartoszSolkaBD: nope, we can control it using `hbar` queyr parameter in mirror node requests. Here are the details of the new implementation: https://github.com/hiero-ledger/hiero-mirror-node/issues/12972

Ferparishuertas (commented):
I think we should remove the config as it will be always weibars

packages/relay/tests/lib/eth/eth_getTransactionByBlockHashAndIndex.spec.ts
```diff
@@ -227,6 +227,10 @@ describe('@ethGetTransactionByBlockHashAndIndex using MirrorNode', async functio
       requestDetails,
     );
     expect(result).to.be.an.instanceOf(Transaction1559);
+    if (result) {
```
jasuwienas: We already checked if this value is set, no need to include this condition here.

packages/relay/src/lib/factories/transactionFactory.ts
```diff
@@ -155,17 +145,21 @@ export const createTransactionFromContractResult = (cr: any): Transaction | null
...
+  const maxPriorityFeePerGas = formatGasFee(cr.max_priority_fee_per_gas);
+
+  const maxFeePerGas = formatGasFee(cr.max_fee_per_gas);
+
   return TransactionFactory.createTransactionByType(cr.type, {
     ...commonFields,
-    maxPriorityFeePerGas: cr.max_priority_fee_per_gas,
-    maxFeePerGas: cr.max_fee_per_gas,
+    maxPriorityFeePerGas: maxPriorityFeePerGas,
+    maxFeePerGas: maxFeePerGas,
```
jasuwienas:
```suggestion
    maxPriorityFeePerGas,
    maxFeePerGas,
```

packages/relay/src/lib/factories/transactionFactory.ts
```diff
@@ -133,12 +127,8 @@ export const createTransactionFromContractResult = (cr: any): Transaction | null
...
   }
 
-  const gasPrice =
-    cr.gas_price === null || cr.gas_price === '0x'
-      ? '0x0'
-      : isHex(cr.gas_price)
-        ? numberTo0x(BigInt(cr.gas_price) * BigInt(constants.TINYBAR_TO_WEIBAR_COEF))
-        : nanOrNumberTo0x(cr.gas_price);
+  const gasPrice = formatGasFee(cr.gas_price);
+  const valueHex = nanOrNumberInt64To0x(cr.amount == null ? null : BigInt(cr.amount));
```
jasuwienas: nit: we used to check if this value is not greater than the TOTAL_SUPPLY_TINYBARS (5_000_000_000_000_000_000) and throw Exception otherwise. But I don't think it was ever good approach, so it's good we stopped doing so (we have this value from mirrornode anwyay).

jasuwienas (dismissed):
Looks great now and works as expected. I don’t have any major comments, so I’m happy to approve.

That said, I’m not entirely convinced that enforcing the use of this query parameter and assuming the precision we will receive will be the one we're expecting is the best approach. Even if we ensure it’s not released before MN on preview/test/mainnet, we could still break network instances for users who haven’t updated their solo node but do update their relay version. Wouldnt be much of a problem if the whole thing would just start throwing exceptions, but we will just start returning unexpected values instead... (but if we want it then ok)>

jasuwienas (dismissed):

jasuwienas (dismissed):

src/relay/lib/clients/mirrorNodeClient.ts
```diff
@@ -840,8 +845,9 @@ export class MirrorNodeClient {
       return cachedResponse;
     }
 
+    const resourcePath = `${MirrorNodeClient.GET_CONTRACT_RESULT_ENDPOINT}${transactionIdOrHash}`;
     const response = await this.get(
-      `${MirrorNodeClient.GET_CONTRACT_RESULT_ENDPOINT}${transactionIdOrHash}`,
+      MirrorNodeClient.withHbarDisabled(resourcePath),
```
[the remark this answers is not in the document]
quiet-node: Also worth verifying first whether the `/contracts/results/{hash}/actions` endpoint is also affected by `hbar=false`. If `/actions` does NOT respect `hbar=false` and still returns tinybars, then only line 711 should be changed. This needs to be confirmed against Mirror Node API documentation.

quiet-node (commented):
Clean work but still have some feedback and concerns. Also the PR description still references `MIRROR_NODE_CONTRACT_RESULTS_HBAR` config flag and cache key suffixing by unit mode but neither exists in the final code. Let's update the PR description to reflect the actual implementation.

Ferparishuertas (dismissed):
LGTM. Lets pass all the checks and merge it

src/relay/lib/clients/mirrorNodeClient.ts
```diff
@@ -166,6 +166,11 @@ export class MirrorNodeClient {
     'MIRROR_NODE_CONTRACT_RESULTS_LOGS_BLOCK_RANGE_PG_MAX',
   );
 
+  private static withHbarDisabled(path: string): string {
+    const sep = path.includes('?') ? '&' : '?';
+    return `${path}${sep}hbar=false`;
+  }
+
```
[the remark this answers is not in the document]
BartoszSolkaBD: 1. Will do
2. Will do
3. There is only one paginated endpoint with `hbar` parameter - `GET /api/v1/contracts/results`. I tested it on v0.151.0 - it returns `hbar` parameter in `next` url
4. There was a flag to toggle between behaviours, but we decided to enforce it. https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5134#pullrequestreview-3997613515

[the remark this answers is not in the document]
quiet-node: Thanks for addressing the other points. For the MN version concern. I understand the decision to enforce `hbar=false` and the config flag removal. but would it make sense to open a follow-up issue to add a startup probe or at least a log warning about the minimum required mirror node version? The concern is that operators running private networks with an older MN would get silently wrong monetary values (10^10x too small) with no errors. Even a single log line during relay startup like "Warning: hbar=false requires mirror node >= v0.XXX" would help operators catch this.

Not blocking this PR on it, just want to make sure we track it.

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -133,12 +127,7 @@ export const createTransactionFromContractResult = (cr: any): Transaction | null
     return null;
   }
 
-  const gasPrice =
-    cr.gas_price === null || cr.gas_price === '0x'
-      ? '0x0'
-      : isHex(cr.gas_price)
-        ? numberTo0x(BigInt(cr.gas_price) * BigInt(constants.TINYBAR_TO_WEIBAR_COEF))
-        : nanOrNumberTo0x(cr.gas_price);
+  const gasPrice = formatGasFee(cr.gas_price);
```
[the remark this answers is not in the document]
quiet-node: oh yeah that's a valid point. You're right per MN spec those fields are returned in hex strings. Indeed removing it is a simplification. Thanks for pointing it out!

quiet-node (dismissed):
LGTM! Great work indeed! ~Just one non-blocking item https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5134#discussion_r3060389997~

src/relay/lib/clients/mirrorNodeClient.ts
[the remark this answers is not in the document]
BartoszSolkaBD: Issue created: https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/5238

jasuwienas (dismissed):
LGTM!

src/relay/lib/debug.ts
```diff
@@ -725,7 +725,7 @@ export class DebugImpl implements Debug {
 
       const { resolvedFrom, resolvedTo } = await this.resolveMultipleAddresses(from, to, requestDetails);
 
-      const value = numberTo0x(tinybarsToWeibars(amount) ?? 0);
+      const value = numberTo0x(amount != null ? BigInt(amount.toString()) : BigInt(0));
```
bootcodes: nit: you can change this here to : 

```suggestion
      const value = nanOrNumberTo0x(amount);
```

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -155,7 +144,7 @@ export const createTransactionFromContractResult = (cr: any): Transaction | null
     transactionIndex: nullableNumberTo0x(cr.transaction_index),
     type: cr.type === null ? '0x0' : nanOrNumberTo0x(cr.type),
     v: cr.v === null ? '0x0' : nanOrNumberTo0x(cr.v),
-    value: nanOrNumberInt64To0x(tinybarsToWeibars(cr.amount, true)),
+    value: nanOrNumberInt64To0x(cr.amount == null ? null : BigInt(cr.amount)),
```
bootcodes:
```suggestion
    value: nanOrNumberInt64To0x(cr.amount),
```

BartoszSolkaBD: this doesn't work as expected. There was a similar cast in `tinybarsToWeibars`

```
return value == null ? null : BigInt(value) * BigInt(constants.TINYBAR_TO_WEIBAR_COEF);
```

Without it, this gets treated as !Number and the `numberTo0x` doesn't work as expected
image

bootcodes: Ok fair - a suggestion can be to move the casting to BigInt if typeof string inside nanOrNumberInt64To0x so we can get ride of those ternary operations and handle string and null properly

quiet-node (dismissed):
LGTM Great work!

jasuwienas (dismissed):
LGTM!

bootcodes (approved):

quiet-node (dismissed):
LGTM

jasuwienas (approved):

natanasow (approved):

PR #5217, opened 2026-04-01
feat: enable same HTTP methods on websocket

### LINKED ISSUE #4991: WebSocket server should offer the same features as the HTTP server
### Problem

When developers or apps connect to the WebSocket port of the Relay, they expect to leverage at least the same functionalities offered by the HTTP port, in addition to the WS specific ones (i.e. subscription).

At the moment, that expectation is not fulfilled and developers/applications can discover later in the process that a specific method is not supported by the WS connection. Because the application cannot use a dynamic failover to the HTTP version (subscription should become polling, etc.) this forces the applications to always use the HTTP server if they want to be sure everything will work.

This is an example with the `eth_feeHistory` method:

```shell
➜ echo '{"method":"eth_feeHistory","params":["0x4", "latest", [25, 75]],"id":1,"jsonrpc":"2.0"}' | websocat wss://testnet.hashio.io/ws

{"error":{"code":-32601,"message":"[Request ID: 1056199a-6e67-4b36-bbe9-7b4cfb48a7ad] Method eth_feeHistory not found"},"jsonrpc":"2.0","id":1}


➜  curl https://testnet.hashio.io/api -X POST -H "Content-Type: application/json" --data '{"method":"eth_feeHistory","params":["0x4", "latest", [25, 75]],"id":1,"jsonrpc":"2.0"}'

{"result":{"baseFeePerGas":["0xe680992c00","0xe680992c00","0xe680992c00","0xe680992c00","0xe680992c00"],"gasUsedRatio":[0.5,0.5,0.5,0.5],"oldestBlock":"0x1da5d11","reward":[["0x0","0x0"],["0x0","0x0"],["0x0","0x0"],["0x0","0x0"]]},"jsonrpc":"2.0","id":1}
```

### Solution

- The WebSocket connection should be just another way to access the same functionalities offered by the HTTP server.
- All methods provided by the Relay and not strictly related to the HTTP/WS protocol (i.e. subscribe, unsubscribe) should be offered on both HTTP and WS connections.

### Alternatives

_No response_

bootcodes:
### Description

The WebSocket server previously gated method dispatch in `WS_CONSTANTS.METHODS`. Any method not in that list (i.e. including eth_feeHistory), returned `-32601 Method not found` even though the relay implemented them.

### Related issue(s)

Fixes #4991

### Testing Guide

1. Run your collection of eth_* methods on http relay server and ws relay server and check outcome is the same
2.
3.

**The bug**
In `main` run : `npm run start:ws`
Run this script :
```node -e "
  const WebSocket = require('ws');

  async function sendRequest(ws, id) {
    return new Promise((resolve) => {
      ws.once('message', (data) => resolve(JSON.parse(data)));
      ws.send(JSON.stringify({ id, jsonrpc: '2.0', method: 'eth_chainId', params: [] }));
    });
  }

  async function connectAs(spoofedIp) {
    return new Promise((resolve, reject) => {
      const ws = new WebSocket('ws://localhost:8546/ws', {
        headers: { 'X-Forwarded-For': spoofedIp }
      });
      ws.on('open', () => resolve(ws));
      ws.on('error', reject);
    });
  }

  async function run() {
    const LIMIT = 3;

    // Connect as two different fake IPs
    const wsA = await connectAs('10.0.0.1');
    const wsB = await connectAs('10.0.0.2');

    // Exhaust the rate limit for IP A
    console.log('--- Exhausting rate limit for 10.0.0.1 ---');
    for (let i = 1; i <= LIMIT; i++) {
      const r = await sendRequest(wsA, i);
      console.log(`IP A req ${i}: ${r.result ?? JSON.stringify(r.error)}`);
    }

    // IP A's next request should be rate-limited
    const rA = await sendRequest(wsA, 99);
    console.log(`IP A req ${LIMIT+1}: ${rA.result ?? JSON.stringify(rA.error)}`);

    // KEY TEST: Does IP B get rate-limited too?
    console.log('\n--- Testing IP B (should be independent) ---');
    const rB = await sendRequest(wsB, 100);
    console.log(`IP B req 1: ${rB.result ?? JSON.stringify(rB.error)}`);

    if (rB.error?.code === -32605) {
      console.log('\nBUG CONFIRMED: IP B is rate-limited despite never exceeding its quota.');
      console.log('   app.proxy = true is missing — both IPs appear as 127.0.0.1');
    } else {
      console.log('\n IPs are independent — app.proxy = true is working correctly.');
    }

    wsA.close();
    wsB.close();
  }

  run().catch(console.error);"
```

**Fix the bug** 

`webSocketServer.ts` was not setting app.proxy = true on its Koa instance. 
 Two features silently broke as a result:
  - WS_CONNECTION_LIMIT_PER_IP  every WebSocket client appeared to originate from the same IP (the proxy's), so the per-IP connection limit either blocked all clients simultaneously once the proxy's bucket was full, or was trivially bypassed by any single client.
  - IP rate limiting on WS RPC methods — all clients shared one rate-limit counter, allowing one heavy client to exhaust the window for everyone else, and preventing any meaningful per-client throttling.

The HTTP server (server.ts) already had app.proxy = true and the RFC 7239 Forwarded→X-Forwarded-For translation middleware.

**The fix**

Extracted the proxy setup into `src/server/utils/proxyUtils.ts` and called it from both servers. The extracted utility:
  1. Sets app.proxy = true
  2. Translates RFC 7239 Forwarded headers to X-Forwarded-For

**Verification**

`tests/ws-server/unit/proxyIpIsolation.spec.ts` contains two tests that directly prove the behavior:
  - proxy=false (baseline): with app.proxy manually reverted to false, two clients with different X-Forwarded-For headers collapse to 127.0.0.1 and share one rate-limit bucket — IP B gets a -32605 error despite never making a request
  - proxy=true (fix): with app.proxy = true (set by applyProxyMiddleware), each spoofed IP has its own independent counter — IP A's exhausted bucket does not affect IP B

Migrate duplicate `eth_*` acceptance tests to `tests/protocol/`. Methods tested independently in both
`tests/server/acceptance/` and `tests/ws-server/acceptance/ `with identical assertions are candidates:
`eth_blockNumber, eth_gasPrice, eth_getLogs, eth_getBalance, eth_call, eth_estimateGas, etc.`
  
 When migrating, delete both originals and replace with a single parameterized test in `tests/protocol/`
 
 Related issue : #5223

### Checklist


BartoszSolkaBD (changes requested):
Looks nice 😄 
I like the idea of creating `RpcProtocolClient` for unified test cases.

tests/protocol/index.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+import { initializeServer } from '../../../src/server/server';
+import { initializeWsServer } from '../../../src/ws-server/webSocketServer';
+import MirrorClient from '../../server/clients/mirrorClient';
+import RelayClient from '../../server/clients/relayClient';
+import ServicesClient from '../../server/clients/servicesClient';
+import { AliasAccount } from '../../server/types/AliasAccount';
+
+chai.use(chaiAsPromised);
+
+describe('RPC Server Acceptance Tests', function () {
```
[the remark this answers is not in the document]
bootcodes: Agreed

Ferparishuertas (commented):
Good work! Some nits

- you mention that there could be some duplicated acceptance tests. Shall we open a follow up issue?
 - I am not seeing  rate limiting test for WebSocket methods. As we have added all, Can you implement some tests?

jasuwienas (dismissed):
LGTM!

src/ws-server/webSocketServer.ts
```diff
@@ -99,6 +100,9 @@ export async function initializeWsServer(
 
   const app = websockify(new Koa());
 
+  // Enable proxy support and RFC 7239 Forwarded header translation
+  applyProxyMiddleware(app);
```
quiet-node: Should we add the `proxyIpIsolation.spec.ts` unit tests mentioned in the PR description? The proxy fix is the correctness-critical change here, and the Codecov report shows `proxyUtils.ts` at 27% coverage. Having the two dedicated tests (proxy=false baseline, proxy=true fix) would directly prove the per-IP isolation behavior and guard against regression.

tests/ws-server/acceptance/index.spec.ts
```diff
@@ -5,159 +5,6 @@ import dotenv from 'dotenv';
 import path from 'path';
 dotenv.config({ path: path.resolve(__dirname, '../../../.env') });
```
quiet-node, line 6: Should this be `'../../.env'` instead? The file lives at `tests/protocol/index.spec.ts` so `__dirname` is two levels deep, not three. Three `../` jumps would land outside the repo root, so the `.env` file would silently not load for local development.

src/server/utils/proxyUtils.ts
```diff
@@ -0,0 +1,110 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import Koa from 'koa';
+
+/**
+ * Parse RFC 7239 Forwarded header to extract the original client IP
+ *
+ * This function safely parses the Forwarded header without using regex to avoid
+ * ReDoS (Regular Expression Denial of Service) vulnerabilities. It includes
+ * input length limits and basic validation to prevent malicious input from
```
quiet-node: The JSDoc says "without using regex to avoid ReDoS" but there's a regex `/^[a-fA-F0-9:.]+$/` at line 76. The regex itself is fine (no backtracking risk), but the comment is inaccurate. Would it make sense to update the JSDoc to so it's not misleading?

tests/ws-server/unit/validations.spec.ts
```diff
@@ -74,19 +74,38 @@ describe('validations unit test', async function () {
...
+
+      expect(verifySupportedMethod(mockRelay, WS_CONSTANTS.METHODS.ETH_SUBSCRIBE)).to.be.true;
+      expect(verifySupportedMethod(mockRelay, WS_CONSTANTS.METHODS.ETH_UNSUBSCRIBE)).to.be.true;
+    });
 
-    UNSUPPORTED_METHODS.forEach((method) => {
-      expect(verifySupportedMethod(method)).to.be.false;
+    it('should return false for unknown method names', () => {
+      const mockRelay = { rpcMethodRegistry: new Map() } as any;
+      const GARBAGE_METHODS = [
```
quiet-node, line 98: Should we add a couple of entries like `debug_unknownOp`, `net_unknownMethod`, `web3_unknownMethod` to the garbage list? Right now the false-return path is only tested with eth\_\* junk. Adding cross-namespace garbage would make sure the registry lookup doesn't accidentally match partial namespace prefixes.

tests/ws-server/helper/rpcMethods.ts
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: Apache-2.0
+
+const REGISTRY_METHODS = [
```
quiet-node, line 7: Would it make sense to add a comment noting that `REGISTRY_METHODS` is a test-only snapshot, not an exhaustive or authoritative list? If the relay adds or removes methods, this list will silently drift, and future contributors might treat it as a canonical registry. Something like "This list is used only for building test mocks; relay.rpcMethodRegistry is the source of truth."

src/ws-server/webSocketServer.ts
quiet-node, line 104: Ohh this proxy fix is a genuinely good catch! However, it seems out of scope of the PR tho, right? The `app.proxy = true` change and the method-routing fix are solving two separate problems. Bundling them makes sense when you're already in the file, but it does make the git history a little harder to read later or for future revisit. If someone bisects a rate-limiting issue down the line and lands on a commit titled "enable same HTTP methods on websocket", they might not immediately connect the dots. 

I think in future if we flag an issue we should raise a ticket and have a dedicated PR for it. That way each concern has its own clear entry point in the history, and if either change ever needs to be reverted independently, it's clean. Not asking to split this PR now, just flagging it as a good habit to build for changes like this. 

Anyhoo still good work tho

quiet-node: Also should we add the proxyIpIsolation.spec.ts unit tests mentioned in the PR description? The proxy fix is the correctness-critical change here, and the Codecov report shows proxyUtils.ts at 27% coverage. Having the two dedicated tests (proxy=false baseline, proxy=true fix) would directly prove the per-IP isolation behavior and guard against regression.

quiet-node (commented):
Nice work overall. Left some items and concerns. Also looks like codacy is complaining about many files https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5217/checks?check_run_id=70629267661.

tests/ws-server/helper/rpcMethods.ts
bootcodes: Agreed - It would be nice to have that in another PR where we check every existing method on the relay and check they all work with latest up-to-date status

src/server/utils/proxyUtils.ts
bootcodes: Copy-pasted the previous JSDoc - will update it accordingly

src/ws-server/webSocketServer.ts
bootcodes: Yes totally agree ! should have split the PRs - forgot that we release per batch - so pushing to main (without proxy fix) wouldn't have caused any issues as we would have patched it straight after.

+1 for test coverage

tests/ws-server/acceptance/index.spec.ts
bootcodes: Pointing to wrong file - but you are correct

tests/protocol/rateLimiter.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+import { expect } from 'chai';
+
+import { WsTestHelper } from '../ws-server/helper';
+import { ALL_PROTOCOL_CLIENTS } from './helpers/protocolClient';
+
+/**
+ * Protocol-level rate limiting tests.
+ */
+describe('@protocol-acceptance @ratelimiter Rate Limiting', () => {
+  this.timeout(30_000);
```
quiet-node, line 12: Looks like the `describe` callback in this file uses an arrow function `() => { ... }`, but `this.timeout(30_000)` on the next line depends on Mocha's `this` context, which only binds correctly with `function() { ... }`. With an arrow function, `this` refers to the enclosing module scope, so `this.timeout()` silently does nothing and all tests default to 2000ms.

The same applies to `this.skip()` in the `before()` hook of `rateLimiterRedis.spec.ts`. It will throw a `TypeError` when Redis is not available instead of gracefully skipping. 

Would it make sense to switch the `describe` (and `before`) callbacks to regular `function()` syntax? Also same idea for tests/protocol/rateLimiterRedis.spec.ts as well

tests/ws-server/helper/index.ts
```diff
@@ -74,6 +75,26 @@ export class WsTestHelper {
...
+  /**
+   * Required: `app.proxy = true` on the WS server.
+   */
+  static sendRequestWithIp(method: string, params: any[], ip: string): Promise<any> {
+    const ws = new WebSocket(WsTestConstant.WS_RELAY_URL, undefined, {
+      headers: { 'X-Forwarded-For': ip },
+    });
+
+    return new Promise((resolve, reject) => {
+      ws.on('error', reject);
```
quiet-node: Small thing: when `ws.on('error', reject)` fires, the WebSocket is never closed. Would it make sense to call `ws.close()` before rejecting? Something like:
```ts
 ws.on('error', (err) => {
   ws.close();
   reject(err);
 });
 ```

This would prevent leaking open connections in test runs, especially for the rate limiter tests that fire many requests.

**Review by quiet-node — COMMENTED** [review]
Looks great! Only two items left

**Review by quiet-node — COMMENTED** [review]
Looking great but looks like Codacy and codecov are complaining quite hard. can you please take a look?

File: tests/ws-server/unit/validations.spec.ts [hunk 9]
```diff
@@ -74,19 +75,41 @@ describe('validations unit test', async function () {
...
+      const mockRelay = { rpcMethodRegistry: new Map() } as unknown as Relay;
+
+      expect(verifySupportedMethod(mockRelay, WS_CONSTANTS.METHODS.ETH_SUBSCRIBE)).to.be.true;
+      expect(verifySupportedMethod(mockRelay, WS_CONSTANTS.METHODS.ETH_UNSUBSCRIBE)).to.be.true;
+    });
 
-    UNSUPPORTED_METHODS.forEach((method) => {
-      expect(verifySupportedMethod(method)).to.be.false;
+    it('should return false for unknown method names', () => {
+      const mockRelay = { rpcMethodRegistry: new Map() } as any;
```
bootcodes:
```suggestion
      const mockRelay = { rpcMethodRegistry: new Map() } as unknown as Relay;
```

bootcodes:
Seems like the codecov yields false positives here on codecov/project

quiet-node (approved):
LGTM! Great work!

jasuwienas (approved):
LGTM!

PR #5224, opened 2026-04-06
perf: optimize getBlockTransactionDetails()

### LINKED ISSUE #4904: [perf] O(N^2) complexity in getBlockTransactionDetails()
**Problem:**
The `getBlockTransactionDetails()` method in `packages/relay/src/lib/debug.ts` has a performance bottleneck with O(N^2) complexity.

Inside the `txHashArray.forEach` loop, the code performs a `.find()` on `actionsResults`:

```typescript
const actionsResult = actionsResults.find((ar) => ar.txHash === txHash);
```

Since `actionsResults` is an array and this search is executed for every transaction in the block, the total complexity is quadratic relative to the number of transactions in the block.

**Impact:**
Tracing blocks with a high transaction count (e.g., several hundred transactions) will experience significant CPU latency on the relay.

**Recommended Solution:**

1. Convert `actionsResults` to a `Map<string, ContractAction[]>` before the final loop to enable O(1) lookups.
2. Replace the `.find()` with a direct map lookup.

bootcodes:
### Description

This PR introduces a fix to improve the performance bottleneck from O(N^2) complexity to O(N) for `getBlockTransactionDetails()` method in `src/relay/lib/debug.ts`

### Related issue(s)

Fixes #4904

### Testing Guide

1. Run tests
2.
3.

N/A

N/A

### Checklist


BartoszSolkaBD (approved):

src/relay/lib/debug.ts
```diff
@@ -956,12 +956,14 @@ export class DebugImpl implements Debug {
     // Build pre-fetched data map
     const preFetchedData: TxHashToContractResultOrActionsMap = {};
 
+    const actionsMap = new Map<string, ContractAction[]>(actionsResults.map((ar) => [ar.txHash, ar.actions]));
```
jasuwienas: nit: I don't like the param name 'ar' and we can avoid having it entirerly by simply destructuring those fields:


```suggestion
    const actionsMap = new Map<string, ContractAction[]>(actionsResults.map(({txHash, actions}) => [txHash, actions]));
```

jasuwienas (dismissed):
LGTM!

tests/relay/lib/debug.spec.ts
```diff
@@ -1605,6 +1605,42 @@ describe('Debug API Test Suite', async function () {
           expect(result).to.be.an('array').with.lengthOf(1);
           expect(result[0]).to.deep.equal({ txHash: contractResult1.hash, result: callTracerResult1 });
         });
+
+        it("should pass each transaction its own actions, not another transaction's", async function () {
```
quiet-node, line 1609: Nice test! Would it make sense to also add a case where one transaction has actions and the other has an empty array `[]`? That would lock in the `actions.length > 0` guard behavior and make sure `callTracer` gets `undefined` (not `[]`) for the empty-actions transaction, preventing a subtle bypass of its internal fetch logic.

tests/relay/lib/debug.spec.ts
```diff
@@ -1605,6 +1605,42 @@ describe('Debug API Test Suite', async function () {
...
+            blockNumber,
+            { tracer: TracerType.CallTracer, tracerConfig: { onlyTopCall: false } },
+            requestDetails,
+          );
+
+          const callForHash1 = callTracerSpy.getCalls().find((c) => c.args[0] === contractResult1.hash);
+          const callForHash2 = callTracerSpy.getCalls().find((c) => c.args[0] === contractResult2.hash);
+
+          expect(callForHash1?.args[4]).to.deep.equal(actionsForHash1);
+          expect(callForHash2?.args[4]).to.deep.equal(actionsForHash2);
```
quiet-node, line 1640: Should we also cover the case where `getContractsResultsActions` throws for one transaction? The catch block returns `{ txHash, actions: [] }`, and the `length > 0` guard should exclude it from preFetchedData. A quick test would make sure the Map optimization handles this path correctly too.

quiet-node (commented):
LG just two small items

BartoszSolkaBD (approved):

quiet-node (dismissed):
LGTM

jasuwienas (dismissed):
LGTM!

natanasow (dismissed):

jasuwienas (approved):

quiet-node (approved):
LGTM! Great work!

PR #5225, opened 2026-04-06
perf: optimize redundant log query

### LINKED ISSUE #4905: [perf] Redundant log fetching (N+1) in debug_traceBlockByNumber
**Problem:**

The `debug_traceBlockByNumber` method in `packages/relay/src/lib/debug.ts` performs redundant network requests for logs.

1. In `getBlockTransactionDetails()`, the relay already fetches all logs for the block timestamp range to identify HTS transaction hashes.
2. However, these logs are discarded after extracting the hashes.
3. Later, when processing the block, `handleSyntheticTransaction` is called for each HTS transaction, which triggers a new API call to fetch logs for that specific transaction hash.

**Impact:**
This creates an N+1 query pattern for logs. If a block contains thousands of HTS transactions, the relay makes N + 1 requests for logs instead of just 1. This significantly increases latency and puts unnecessary load on the Mirror Node.

**Recommended Solution:**

1. In `getBlockTransactionDetails()`, group the existing `allLogs` by transaction hash.
2. Include these logs in the `preFetchedData` map.
3. Update `handleSyntheticTransaction` (and the tracer methods) to accept and use pre-fetched logs if available, avoiding the redundant network call.

bootcodes:
### Description

This PR introduces a fix to store prefetched logs in an object and pass it down into child functions to avoid any redundant API calls. 

### Related issue(s)

Fixes #4905

### Testing Guide

1. Run tests
2.
3.

N/A

N/A

### Checklist


src/relay/lib/debug.ts
```diff
@@ -53,6 +54,11 @@ import {
 import type { ContractAction, MirrorNodeBlock, MirrorNodeContractResult } from './types/mirrorNode';
 import { rpcParamValidationRules } from './validators';
 
+interface PreFetchedData {
+  logs: MirrorNodeContractLog[];
```
jasuwienas: Wouldn’t it make more sense to store logs as a map keyed by txHash? We’re always filtering them by txHash before using them anyway. I’m fine with either approach, but since we’re already using a map for contract results, it would be more consistent to do the same here.

src/relay/lib/debug.ts
```diff
@@ -684,10 +691,12 @@ export class DebugImpl implements Debug {
       }
 
       if (!actionsResponse?.[0]?.call_operation_type || !transactionsResponse) {
+        const mirrorNodeContractLog = preFetchedLogs?.find((log) => log.transaction_hash === transactionHash);
         return (await this.handleSyntheticTransaction(
```
quiet-node: This is O(N) per synthetic transaction. For a block with K synthetic transactions out of N total logs, total cost is O(K * N). In a worst case where most transactions are synthetic HTS operations (common on Hedera), K ~ N, giving O(N^2). We should make preFetchedLogs a map for O(1) look up

src/relay/lib/debug.ts
quiet-node: I aggree With Mariusz. We should use map for this to avoid O(N*N). Right now each tracer does a .find() scan over the entire array, which is O(N) per lookup. For blocks with many synthetic transactions, this adds up to O(N^2). Building the map once in getBlockTransactionDetails() would give us O(1) lookups

quiet-node: Let's also add TSDoc

tests/relay/lib/debug.spec.ts
```diff
@@ -1805,6 +1805,34 @@ describe('Debug API Test Suite', async function () {
             .filter((call) => call.args[0] === contractResultWrongNonce.hash);
           expect(wrongNonceActionsCalls).to.be.empty;
         });
+
+        it('should not make per-transaction log API calls when logs are pre-fetched at block level', async function () {
```
quiet-node, line 1847: This test covers the CallTracer path nicely. Should we add an equivalent test for PrestateTracer? It received the same `preFetchedLogs` parameter and `.find()` logic, so it would be good to have coverage for both paths

src/relay/lib/debug.ts
```diff
@@ -953,13 +966,17 @@ export class DebugImpl implements Debug {
...
     const actionsResults = await Promise.all(actionsPromises);
 
-    // Build pre-fetched data map
-    const preFetchedData: TxHashToContractResultOrActionsMap = {};
+    // Build pre-fetched data object
+    const preFetchedData: PreFetchedData = {
+      logs: [],
+      txHashMap: {},
+    };
+    preFetchedData.logs = allLogs || [];
```
quiet-node, line 980: Nit: `logs` is initialized to `[]` and then immediately overwritten with `allLogs || []` on the next line. Would it make sense to just initialize it directly in the object literal?

quiet-node (commented):
Clean work left some feedback

src/relay/lib/model.ts
```diff
@@ -385,4 +388,19 @@ export class Log {
...
+      blockHash: toHash32(log.block_hash),
+      blockNumber: numberTo0x(log.block_number),
+      data: log.data,
+      logIndex: numberTo0x(log.index),
+      removed: false,
+      topics: log.topics,
+      transactionHash: toHash32(log.transaction_hash),
+      transactionIndex: numberTo0x(log.transaction_index),
+    });
+  }
```
quiet-node, line 411: Nice extraction! Let's make sure we have TSDoc for it as well

quiet-node (dismissed):
LGTM great work! Couple of codacy warnings tho.

src/relay/lib/model.ts
```diff
@@ -3,6 +3,9 @@
 // Used for fake implementation of block history
 import { Status, TransactionRecord } from '@hashgraph/sdk';
 
+import { numberTo0x, toHash32 } from '../formatters';
+import { MirrorNodeContractLog } from './types';
```
bootcodes:
```suggestion
import type { MirrorNodeContractLog } from './types';
```

src/relay/lib/debug.ts
```diff
@@ -26,7 +26,7 @@ import constants, { CallType, TracerType } from './constants';
 import { cache, RPC_LAYOUT, rpcMethod, rpcParamLayoutConfig } from './decorators';
 import { predefined } from './errors/JsonRpcError';
 import { BlockFactory } from './factories/blockFactory';
-import { Block } from './model';
+import { Block, Log } from './model';
```
bootcodes:
```suggestion
import { Block, type Log } from './model';
```

quiet-node (approved):
LGTM! Great work!

natanasow (approved):

PR #5228, opened 2026-04-07
fix: resolve lint errors across repo and strengthen CI enforcement (1/3)

### LINKED ISSUE #4700: Resolve Lint Errors Across Packages and Strengthen CI Enforcement
## Problem

All four packages in the repository currently have numerous lint errors. Running `npm run lint` in any package reveals extensive ESLint violations, which hinders the development process. Developers encounter these errors even when their changes do not touch the affected areas, leading to distractions and potential clutter in code reviews.

## Goals

- Resolve all lint errors across the entire codebase.
- Ensure successful TypeScript typechecking in all packages.
- Apply `eslint --fix` where possible to auto-correct fixable issues.
- Strengthen the CI pipeline to catch and fail on lint issues for entire repo for every pull request.

## Acceptance Criteria

- `npm run lint` passes in all packages without errors.
- `npm run typecheck` (or equivalent) succeeds in all packages.
- CI jobs include lint checks that block PR merges on failures.
- No regressions in existing functionality or tests.

bootcodes:
### Description

This PR strengthens the CI pipeline by adding automated lint checks and resolves all lint errors across the codebase.

**1. Lint Error Resolution:**
- Fixed all lint errors

**2. Eslint config:**
- Scoped the eslint to `src/ `and `test/` folder

**3. CI Pipeline Enhancement**
- Created new .github/workflows/lint.yml workflow that runs on every PR
- Lint checks now block PR merges if any errors are found
- Adds a separate lint-type-aware CI job that runs no-floating-promises and no-misused-promises on PRs only via eslint.config.ci.mjs. Keeps local linting fast by deferring expensive type-aware rules to CI.

**4. Husky hook**
- Added pre-push hook to run npm build

### Related issue(s)

Fixes #4700

### Testing Guide

1. Verify Lint Status: `npm run lint`
2. Verify Type Checking: `npx tsc --noEmit`
3. Verify Build: `npm run build`

Scoped eslint config to `src` and `test` folder - see `eslint.config.mjs`

N/A

### Checklist


src/relay/lib/clients/mirrorNodeClient.ts
```diff
@@ -156,6 +156,7 @@ export class MirrorNodeClient {
    */
   private readonly cacheService: ICacheClient;
 
+  // eslint-disable-next-line no-useless-escape
   static readonly EVM_ADDRESS_REGEX: RegExp = /\/accounts\/([\d\.]+)/;
```
jasuwienas: @bootcodes are you sure we can't just apply
```suggestion
static readonly EVM_ADDRESS_REGEX: RegExp = /accounts\/([\d.]+)/;
```
this suggestion?

src/relay/lib/clients/mirrorNodeClient.ts
```diff
@@ -637,7 +637,7 @@ export class MirrorNodeClient {
     const queryParamsFiltered = Object.fromEntries(
       Object.entries(queryParamObject).filter(([key, value]) => {
         if (key === MirrorNodeClient.ACCOUNT_TRANSACTIONS_PROPERTY && value) return false;
-        return value !== undefined && value !== '';
+        return value != null && value !== '';
```
jasuwienas: Are you sure, @bootcodes we used to keep nulls here. (but it actually makes sense not to submit them to mirrornode, that's right...)

src/relay/lib/validators/utils.ts
```diff
@@ -38,7 +38,7 @@ export function validateObject<T extends object = any>(object: T, filters: IObje
     }
   }
 
-  const paramsMatchingFilters = Object.keys(filters.properties).filter((key) => object[key] !== undefined);
```
jasuwienas, line 41: I'm not quite sure if we want to remove nulls in each of these places...

jasuwienas (commented):
Overall, it looks great. I’m just not sure whether it’s correct to treat null the same as undefined in places where we previously distinguished between the two. I don’t think this PR should introduce business-level changes to the application, even if they make sense.

src/relay/lib/clients/mirrorNodeClient.ts
bootcodes: So as a general rule `!= null` checks if the value is **nor undefined nor null.** 
We can keep `!== undefined` if it was already in place in the code. However changing `!= undefined` to `!== undefined` can result in regressions.

eslint.config.ci.mjs
```diff
@@ -0,0 +1,29 @@
...
+// and are too expensive to run on every save locally.
+// Run via: npx eslint . --config eslint.config.ci.mjs
+
+import { defineConfig } from 'eslint/config';
+import tsParser from '@typescript-eslint/parser';
+import baseConfig from './eslint.config.mjs';
+
+export default defineConfig([
+  ...baseConfig,
+
```
jasuwienas:
```suggestion
```

tests/server/acceptance/htsPrecompile/precompileCalls.spec.ts
```diff
@@ -188,7 +179,6 @@ describe('@precompile-calls Tests for eth_call with HTS', async function () {
       adminPrivateKey: accounts[0].privateKey,
     };
     const mintResult0 = await servicesNode.mintNFT({ ...mintArgs, tokenId: nftTokenId0 });
-    const mintResult1 = await servicesNode.mintNFT({ ...mintArgs, tokenId: nftTokenId1 });
```
jasuwienas, line 191: Are you sure we don't need this second token minted as well? Maybe we just don't care about the result?

src/relay/lib/clients/mirrorNodeClient.ts
jasuwienas: But the only reason I'm pointing this out is because we had
```
!== undefined
```

here, not:
```
!= undefined
```
If you replace
```
const value = null;
if (value !== undefined) console.log('ABC');
```

with:

```
const value = null;
if (value != null) console.log('ABC');
```
You alter app behaviour

bootcodes: Yes I saw that - will revert and keep all previous `!== undefined`

src/relay/lib/clients/mirrorNodeClient.ts
```diff
@@ -637,7 +636,7 @@ export class MirrorNodeClient {
...
       }),
```
quiet-node: looks like the revert for this line didn't make it into the final diff. Query parameters with `null` values (e.g., `limit: null`) that were previously forwarded to Mirror Node will now be silently dropped. We shouldn't take that chance and potentially introduce any regression.

src/relay/lib/clients/mirrorNodeClient.ts
```diff
@@ -828,7 +827,7 @@ export class MirrorNodeClient {
   public async getContractId(contractIdOrAddress: string, requestDetails: RequestDetails, retries?: number) {
     const cachedLabel = `${constants.CACHE_KEY.GET_CONTRACT}.id.${contractIdOrAddress}`;
     const cachedResponse: any = await this.cacheService.getAsync(cachedLabel, MirrorNodeClient.GET_CONTRACT_ENDPOINT);
-    if (cachedResponse != undefined) {
+    if (cachedResponse) {
```
quiet-node: Would it make sense to use `!= null` here instead of a truthiness check, for consistency with the `isValidContract` cache check at line 809? Truthiness checks could miss falsy-but-valid cached values

src/relay/lib/clients/mirrorNodeClient.ts
```diff
@@ -863,10 +862,10 @@ export class MirrorNodeClient {
...
     if (
-      response != undefined &&
-      response.transaction_index != undefined &&
-      response.block_number != undefined &&
-      response.block_hash != constants.EMPTY_HEX &&
+      response != null &&
+      response.transaction_index != null &&
+      response.block_number != null &&
+      response.block_hash !== constants.EMPTY_HEX &&
       response.result === 'SUCCESS'
```
quiet-node, line 883: I supposed these are changed because the linting is complaining?

src/relay/lib/db/types/hbarLimiter/errors.ts
```diff
@@ -23,7 +23,7 @@ export class EvmAddressHbarSpendingPlanNotFoundError extends Error {
 
 export class IPAddressHbarSpendingPlanNotFoundError extends Error {
   constructor(ipAddress: string) {
-    super(`IPAddressHbarSpendingPlan not found`);
+    super(`IPAddressHbarSpendingPlan not found for IP address ${ipAddress}`);
```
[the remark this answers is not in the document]
quiet-node: Should we just prefix with `_ipAddress` instead? Including IP addresses in error messages could expose PII in logs/stack traces. Since forever we never log any IP addresses.

src/config-service/commands/printEnvs.ts
```diff
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
-// eslint-disable-next-line @typescript-eslint/no-require-imports
-const { ConfigService } = require('../../dist/services');
+import { ConfigService } from '../services';
```
quiet-node: Let's make sure `npm run print-env` still works after this change. The old `require('../../dist/services')` was loading from compiled output, and the new import changes the resolution path.

.husky/pre-push
```diff
@@ -0,0 +1 @@
+npm run build
```
quiet-node: Would it make sense to run `npm run lint` here instead of the full `npm run build`? A full build on every push might slow down the local development workflow significantly. The build check could be deferred to CI where it already runs.

quiet-node: Well actually doesn't the pre-commit hook already handles linting via lint-staged for each commit? If it doesn't work should we just fix it instead of adding a new hook?

src/relay/lib/constants.ts
```diff
@@ -112,6 +112,7 @@ export default {
   ETH_GET_GAS_PRICE_CACHE_TTL_MS_DEFAULT: `${CACHE_TTL.FIFTEEN_MINUTES}`,
   ETH_GAS_FEE_TTL: `${CACHE_TTL.HALF_HOUR}`,
   ETH_FEE_HISTORY_TTL: `${CACHE_TTL.HALF_HOUR}`,
+  // eslint-disable-next-line no-useless-escape
   TRANSACTION_ID_REGEX: /\d{1}\.\d{1}\.\d{1,10}\@\d{1,10}\.\d{1,9}/,
```
quiet-node: Should we just remove the `\` before `@` instead of suppressing the lint rule? The `@` character has no special meaning in regex, so `\@` and `@` are identical. That way we don't need the disable comment.

quiet-node (commented):
Great work overall! Left some items

src/relay/lib/clients/mirrorNodeClient.ts
bootcodes: I guess we could - I never expect contract.contract_id's value to be false or "" - so that is the reason I removed it here. But will keep it for consistency

src/relay/lib/clients/mirrorNodeClient.ts
bootcodes: yes they are : It's not a problem here though
`> null != undefined
false`
It won't pass the condition if null or undefined values are encountered

src/relay/lib/constants.ts
bootcodes: Tested it - all works fine without `\`

src/config-service/commands/printEnvs.ts
bootcodes: Good catch - it didn't seem to work in the first place either

.husky/pre-push
bootcodes: `npm run lint` on pre-push would be redundant with lint-staged on pre-commit hook. But `npm run build` isn't the right gate either. I'm more inclined to have `tsc --noEmit` on pre-push hook, as currently it doesn't invoke `tsc` so type errors slip through and only appear in CI. We could catch those errors early without significant bottleneck on local development

src/relay/lib/clients/mirrorNodeClient.ts
```diff
@@ -637,7 +636,8 @@ export class MirrorNodeClient {
     const queryParamsFiltered = Object.fromEntries(
       Object.entries(queryParamObject).filter(([key, value]) => {
         if (key === MirrorNodeClient.ACCOUNT_TRANSACTIONS_PROPERTY && value) return false;
-        return value !== undefined && value !== '';
+        // eslint-disable-next-line eqeqeq
+        return value != undefined && value !== '';
```
quiet-node: looks like the revert here might still have a subtle issue. The current code uses `value != undefined` (loose equality), which requires the `eslint-disable-next-line eqeqeq` suppress comment. The original was `value !== undefined` (strict equality), which doesn't need a suppress. Since the project's `eqeqeq` rule with `{ null: ignore }` allows strict checks, would it make sense to just restore `return value !== undefined && value !== ''` here? That would match the original behavior exactly and drop the disable comment.

src/config-service/commands/printEnvs.ts
quiet-node: Nice but could we quickly confirm that `npm run print-env` actually works end-to-end after this change? It was noted that the old path was already broken, so it would be good to verify the new path resolves correctly against the compiled output. If out of scope let's create a ticket to report this separately. Non-blocking.

.husky/pre-push
quiet-node: Coolio yeah since its a hook for pre-push so tsc --noEmit might not be so slow. Well hopefully it's not so slow LOL. Anywho thanks for the explanation

quiet-node (commented):
LG but only one item left

src/relay/lib/clients/mirrorNodeClient.ts
bootcodes: yes probably missed this one - will fix it

.husky/pre-push
```diff
@@ -0,0 +1 @@
+tsc --noEmit
```
jasuwienas, line 1: @bootcodes should we use the same script as in ci  or npm package tsc --noEmit -p tsconfig.eslint.json ?

nit: Btw won't this pipeline check be enough? I'm fine with keeping this pre-push, but I would personally vote for not including it since it might cause more issues than solutions..

tests/ws-server/acceptance/subscribe.spec.ts
```diff
@@ -230,7 +230,7 @@ describe('@web-socket-batch-3 eth_subscribe', async function () {
           webSocket.on('message', function incoming(data) {
             const parsed = JSON.parse(data);
             if (parsed.id !== null || parsed.method) {
-              if (subscriptionId == '') {
+              if (subscriptionId === '') {
```
jasuwienas: Here we changed app behaviour.. subscription id comes from parsed json (data). And might as well be 0 (which compares to '' ). When its 0 we enter here. And we won't do it anymore.

I guess current approach is ok - I guess 0 should never be a generated subscriiton id, but I'm really not sure. Let's double check it before merging.

jasuwienas (dismissed):
Looks great ! thanks for resolving my previous concerns.

.husky/pre-push
bootcodes: Will open a thread on slack to discuss this

quiet-node (dismissed):
LGTM great work!

.husky/pre-push
jasuwienas: Discused internally, let's include this!

jasuwienas (dismissed):
LGTM

simzzz (dismissed):

jasuwienas (approved):

quiet-node (approved):
LGTM great work as always!

PR #5234, opened 2026-04-09
chore: use solo node instead of local node (#5060)

### LINKED ISSUE #5060: [A5 SOLO] Migrate Hiero Local Node to Solo (Once metrics are equivalent)
At the EVM workstream we have switched to solo on the following projects

- https://github.com/hashgraph/hedera-evm-testing
- https://github.com/hiero-ledger/hiero-contracts

We have left a couple of repos, to be able to compare, and only merge when overall solo is on similar speed, performance as local node. Within the following repos we ha several pipelines to update

- https://github.com/hashgraph/hedera-forking
- https://github.com/hiero-ledger/hiero-json-rpc-relay
- https://github.com/hashgraph/hedera-smart-contracts ( We need to evaluate if a local network is still needed, as we are migrating system-contracts to hiero-ledger/hiero-oontracts

Repos that will be deprectated
- https://github.com/hashgraph/hedera-sourcify. Will me migrated to use the global sourcify instance

jasuwienas:
### Description

Replaces hedera local node with hiero solo node.

Updates the CN and MN versions and fixes the tests to work correctly with them: 
 - Gas limit changes are required because the gas limit is now dependent on the CN version. For CN versions below 0.69, it differed from the version we switched to (0.72.0). It applies to conformity tests and debug tests.
 - Two tests (api_batch1) that depended on the transaction failing were written in a way that caused them to fail due to gas limits. I’ve updated them to fail with a wrong_nonce error instead, so they no longer depend on a configurable or changeable value.
 - We had an incorrect assumption that logs emitted by a transaction would always include a fragment field. After updating the MN (change between 0.144.0 and v0.151.0) and CN versions, additional logs for token transfers were introduced, which caused tests related to token operations to fail. I’ve updated the tests to correctly handle this case, expecting correct object interface.
 -  MN no longer support long zero address flag. All the tests which depended on it are no longer executable.

### Related issue(s)

Fixes #5060

### Testing Guide

1. Run the tests
2. Make sure they pass...

N/A

### Checklist


.github/scripts/forward-ports.sh
quiet-node, line 1: So this script works around Solo's port-forwarding, but hiero-ledger/solo#3833 was fixed and seems like it still failed? I think if there's still a bug specifically with falcon deploy, we should file it on solo repo so it gets fixed upstream, rather than a workaround to fix at the consumer level like this. Or maybe we can go with this workaround now, create a ticket on Solo, then add a TODO here to track. If it's fixed upstream, it can avoid unnecessary complexity for our CI.

.github/scripts/forward-ports.sh
```diff
@@ -0,0 +1,36 @@
...
+ps aux | grep "port-forward" | grep kubectl | awk '{print $2}' | xargs -r kill -9
+NS="$(kubectl get ns -o name | sed 's|^namespace/||' | grep 'hiero' | head -n1)"
+
+listen() {
+  local pod="$1"
+  local ports="$2"
+  (
+    while true; do
+      if ! ps aux | grep -F kubectl | grep -F port-forward | grep -F " ${ports}" | grep -v grep >/dev/null; then
+        kubectl port-forward --address 0.0.0.0 "$pod" -n "$NS" "${ports}" >/dev/null 2>&1 &
```
quiet-node, line 30: on line 34, `kubectl get pods -A` searches all namespaces but the port-forward uses `$NS` -- should we use `kubectl get pods -n "$NS"` to ensure they match?

.github/workflows/acceptance-workflow.yml
```diff
@@ -97,14 +97,30 @@ jobs:
...
+          install_only: true
+          node_image: kindest/node:v1.32.5@sha256:e3b2327e3a5ab8c76f5ece68936e4cafaa82edf58486b769727ab0b3b97a5b0d
+          version: v0.29.0
+          kubectl_version: v1.32.5
+          verbosity: 3
+          wait: 120s
+
+      - name: Start the solo node
+        run: |
+          npx @hashgraph/solo@${{ inputs.soloVersion || '0.68.0' }} one-shot falcon deploy \
```
quiet-node, line 133: It looks like `soloVersion` isn't defined as an input in the `workflow_call` section. This means the `|| '0.68.0'` fallback will always be used and callers like `manual-testing.yml` can't override it. Should we add `soloVersion` as a proper input to all the reusable workflows?

.github/workflows/acceptance-workflow.yml
quiet-node, line 1: I see `release-acceptance.yml` still references hedera local node. It has a "Do not use! WIP" comment so it's probably fine for now, but should we track updating it as a follow-up?

.github/workflows/acceptance-workflow.yml
quiet-node: With this update, the networkTag and mirrorTag will no longer be used (here and in other workflows)

I think a nice way to handle this is to add a new setup-value-file step. This step can load the default falcon.yml, then dynamically loads the networkTag and mirrorTag. That way, anyone who still wants to run the test using specific tags can continue to do so.

quiet-node: Also about the flags,
- Falcon alreayd have --force-port-forward set to true, right? I think we don't need to set it explicitly. In fact, should we make --force-port-forward=false as we have our own portforward setup anyway?
- We should add --deploy-explorer=false to save time

.github/falcon.yml
```diff
@@ -0,0 +1,15 @@
+network:
+  --release-tag: 'v0.72.0'
+  --node-aliases: 'node1'
+setup:
+  --release-tag: 'v0.72.0'
+  --node-aliases: 'node1'
+consensusNode:
+  --node-aliases: 'node1'
+  --force-port-forward: true
```
quiet-node: Same question here do we need this flag (same for mirrorNode) as we would have customized portforward anyway?

k6/src/lib/parameters.js
```diff
@@ -86,17 +86,17 @@ export const computeLatestLogParameters = (configuration) =>
...
 export const setDefaultValuesForEnvParameters = () => {
-  __ENV['MIRROR_BASE_URL'] = __ENV['MIRROR_BASE_URL'] || 'http://localhost:5551';
-  __ENV['RELAY_BASE_URL'] = __ENV['RELAY_BASE_URL'] || 'http://localhost:7546';
-  __ENV['DEFAULT_DURATION'] = __ENV['DEFAULT_DURATION'] || '120s';
-  __ENV['DEFAULT_VUS'] = __ENV['DEFAULT_VUS'] || 10;
-  __ENV['DEFAULT_LIMIT'] = __ENV['DEFAULT_LIMIT'] || 100;
-  __ENV['DEFAULT_PASS_RATE'] = __ENV['DEFAULT_PASS_RATE'] || 0.95;
-  __ENV['DEFAULT_MAX_DURATION'] = __ENV['DEFAULT_MAX_DURATION'] || 500;
-  __ENV['STRESS_TEST_TARGET_TOTAL_RPS'] = __ENV['STRESS_TEST_TARGET_TOTAL_RPS'] || 100;
+  __ENV.MIRROR_BASE_URL ||= 'http://localhost:5551';
```
quiet-node, line 89: Curious why needed these updates?

tests/relay/lib/debug.spec.ts
```diff
@@ -320,7 +320,7 @@ describe('Debug API Test Suite', async function () {
     timestamp: '0x698afa66',
     difficulty: '0x0',
     extraData: '0x',
-    gasLimit: '0x1c9c380',
+    gasLimit: '0x8f0d180',
```
quiet-node, line 323: Why this change? New CN versions? If so let's kindly update the PR description so it's clear why we need new values. Like from which CN version or MN version and whatnot. It would help for future revisit what's happening in this PR

.github/falcon.yml
quiet-node, line 1: With this update, we seem to be running into the same challenge we had with LocalNode. We don't have an automated pipeline to keep the CN and MN versions up to date. What used to happen is we would keep running tests against the same versions for however long we need, sometimes spanning several releases, until we eventually needed to upgrade. And when we finally did upgrade, lot of stuff break 😅

I think we should look into a better approach here. For example, we could set up a cron job that runs weekly, checks for new CN and MN versions, and automatically creates a PR to bump them. That way we can catch breaking changes early and stay in sync with CN and MN as they evolve.

quiet-node: Created this ticket for tracking https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/5235

tests/server/acceptance/htsPrecompile/precompileCalls.spec.ts
```diff
@@ -283,11 +283,6 @@ describe('@precompile-calls Tests for eth_call with HTS', async function () {
       expect(totalSupply).to.eq(BigInt(INITIAL_SUPPLY));
     });
 
-    it('Function with IERC20(token).balanceOf(account) - using long zero address', async () => {
```
quiet-node, line 286: Could you explain why we remove these tests? I understand since the new versions it can behave different but let's drop some expalantion in the PR description so we're clear what's happening, why many tests are simply removed but not updated. Just don't want to simply lose coverage

tests/server/acceptance/htsPrecompile/tokenCreate.spec.ts
```diff
@@ -554,7 +554,7 @@ describe('@tokencreate HTS Precompile Token Create Acceptance Tests', async func
           Constants.GAS.LIMIT_1_000_000,
         );
         const responseCodeRevokeKyc = (await revokeKycTx.wait()).logs.filter(
-          (e) => e.fragment.name === Constants.HTS_CONTRACT_EVENTS.ResponseCode,
+          (e) => e.fragment?.name === Constants.HTS_CONTRACT_EVENTS.ResponseCode,
```
quiet-node, line 557: Why need ? now?

tests/server/acceptance/rpc_batch1.spec.ts
```diff
@@ -1002,18 +1002,16 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
-              gasLimit: 41484,
+              nonce: (await relay.getAccountNonce(accounts[1].address)) + 2,
             };
             const signedTx = await accounts[1].wallet.signTransaction(tx);
-            const txHash = await relay.sendRawTransaction(signedTx);
-            await relay.pollForValidTransactionReceipt(txHash);
-            const mnResult = await mirrorNode.get(`/contracts/results/${txHash}`);
+            await relay.sendRawTransaction(signedTx);
+
+            await new Promise((r) => setTimeout(r, 5000));
```
quiet-node, line 1011: Why switch to hard sleeper now instead of dynamic polling?

tests/server/acceptance/rpc_batch1.spec.ts
```diff
@@ -1036,36 +1033,26 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+
+            const mnResult = await mirrorNode.get(`/contracts/results/${txHash2}`);
 
             const nonceLatest = await relay.getAccountNonce(accounts[1].address);
             const noncePending = await relay.getAccountNonce(accounts[1].address, 'pending');
 
-            expect(mnResult1.result).to.equal('INSUFFICIENT_GAS');
-            expect(mnResult2.result).to.equal('SUCCESS');
-            expect(mnResult3.result).to.equal('INSUFFICIENT_GAS');
```
quiet-node, line 1068: Why remove these coverage for the different transaction records?

quiet-node (commented):
Thanks for taking on the challenge of upgrading the versions! I’ve done it once so I know how tough it can be. Great work! Left some Qs and concerns

.github/scripts/forward-ports.sh
jasuwienas: I’ll update the description to better clarify what’s currently missing in the solo implementation.

What I'm actually missing in solo is flag like:

* `--external-address`

so  I can bind solo to the address: `--address 0.0.0.0`

Also, I believe it was your suggestion to remove HAProxy and call the underlying services directly.

.github/workflows/acceptance-workflow.yml
jasuwienas: Yes, right, we have to create the config file dynamically (and nice idea with the explorer, we DON'T need it)

k6/src/lib/parameters.js
jasuwienas: It's just a codacy proposal applied

tests/relay/lib/debug.spec.ts
jasuwienas: CN version change related udpate, I'll include it in the PR description

tests/server/acceptance/htsPrecompile/precompileCalls.spec.ts
jasuwienas: Long zero addresses flag is no longer supported by mn. The tsts which depend on them had to be removed, I'll include in the PR descirpiton.

tests/server/acceptance/htsPrecompile/tokenCreate.spec.ts
jasuwienas: Extra logs are being provided by mirrornode for transfers. They do not have the fragment part included. ALso.. Log returned by ethers here can be of type:

```
    get logs(): Array<EventLog | Log>;
```
So by definition we should expect it not to always have the fragment propoerty.

.github/workflows/acceptance-workflow.yml
jasuwienas: I've updated this yml as well.

tests/server/acceptance/rpc_batch1.spec.ts
jasuwienas: I've added reasoning behind editing this test into the PR description.  Let me know if that is sufficient ;) . (also, reverted status check for failures, there was no need to remove that).

tests/server/acceptance/rpc_batch1.spec.ts
jasuwienas: I've added reasoning behind editing this test into the PR description.  Let me know if that is sufficient ;) . (also brought back dynamic polling, there was no need to remove that).

tests/server/acceptance/htsPrecompile/tokenCreate.spec.ts
jasuwienas: I've added reasoning behind editing this test into the PR description.  Let me know if that is sufficient.

tests/server/acceptance/htsPrecompile/precompileCalls.spec.ts
jasuwienas: I've added reasoning behind editing this test into the PR description.  Let me know if that is sufficient.

.github/scripts/forward-ports.sh
jasuwienas: done

.github/scripts/forward-ports.sh
jasuwienas: If you know any workaround for that let me know please. I'm keeping it as is for now, since some of the tests (dapp for example) require the port to be available from the container level as well, not only on the localhost.

quiet-node: Thanks for updating the description. It's much clearer now. Would it make sense to file a ticket on the Solo repo for the missing `--external-address` flag, or any issue, and add `# TODO(solo#XXXX)` comments at the top of this script? That way we have a clear path to remove this workaround once it's fixed upstream.

.github/workflows/release-acceptance.yml
```diff
@@ -68,8 +79,49 @@ jobs:
...
+            --dev \
+            --deploy-explorer=false \
+            --deploy-relay=false
+        timeout-minutes: 15
+
+      - name: Forward solo ports
+        run: .github/scripts/forward-ports.sh
+
+      - name: Start Redis
+        run: docker run -d --name redis -p 6379:6379 redis:7-alpine
```
quiet-node, line 123: Nice one! However it looks like there are still a couple of leftover local-node commands at line 129 and 132. Let's remove them

.github/workflows/acceptance-workflow.yml
quiet-node: Thanks for addressing these other two.  A few things I noticed about the flags:
1. The dynamically generated `falcon.yml` still has `--force-port-forward: true` under `consensusNode` in all workflows. Since we have our own port-forward script, should we remove this or set it to false?

2. `dapp.yml` is missing `--deploy-explorer=false` that the other workflows have.

3. In `subgraph.yml`, the CLI has `--force-port-forward=false` but the config file has `--force-port-forward: true`. These might conflict.

quiet-node (commented):
Looking great left some extra item on the new updates

.github/workflows/release-acceptance.yml
jasuwienas: Right, missed those.

BartoszSolkaBD (changes requested):
1. Update READMEs references to local node where possible.
2. Update package.json files, as they also have local node references:
    - `"local-node": "npx @hashgraph/hedera-local restart -d"`
    - `"@hashgraph/hedera-local": "^2.39.0"`
3. We also have `LOCAL_NODE` environment variable - not sure if it should be removed or just aligned with Solo naming

.github/workflows/subgraph.yml
```diff
@@ -39,12 +39,71 @@ jobs:
...
+          verbosity: 3
+          wait: 120s
+
+      - name: Start the solo node
+        working-directory: .
+        run: |
+          npx @hashgraph/solo@0.68.0 one-shot falcon deploy \
+            --values-file .github/falcon.yml \
+            --dev \
+            --force-port-forward=false \
```
quiet-node: Let's remove this guy as well

quiet-node (commented):
Nice work only one small item and the TODO comment and should be good on my end now

.github/scripts/forward-ports.sh
jasuwienas: Sure,I'll create an issue with this proposal on their board.

jasuwienas: https://github.com/hiero-ledger/solo/issues/3995

quiet-node (dismissed):
LGTM! Awesome work!

natanasow (dismissed):

.github/workflows/hoppscotch.yml
```diff
@@ -53,11 +53,48 @@ jobs:
...
+          install_only: true
+          node_image: kindest/node:v1.32.5@sha256:e3b2327e3a5ab8c76f5ece68936e4cafaa82edf58486b769727ab0b3b97a5b0d
+          version: v0.29.0
+          kubectl_version: v1.32.5
+          verbosity: 3
+          wait: 120s
+
+      - name: Start the solo node
+        run: |
+          npx @hashgraph/solo@0.68.0 one-shot falcon deploy \
```
rbarker-dev:
```suggestion
          @hashgraph/solo@${{ inputs.soloVersion || '0.68.0' }} one-shot falcon deploy \
```

Recommend adding the workflow_dispatch inputs and using that.

tools/hardhat-viem-example/README.md
```diff
@@ -1,6 +1,6 @@
 # Hedera Hardhat Viem Example Project
 
-This Hedera Hardhat Viem Example Project offers boilerplate code for testing and deploying smart contracts via Hardhat with Viem. It includes configuration for both community-hosted and local ([Hedera Local Node](https://github.com/hashgraph/hedera-local-node)) instances of the [Hedera JSON RPC Relay](https://github.com/hiero-ledger/hiero-json-rpc-relay). 
+This Hedera Hardhat Viem Example Project offers boilerplate code for testing and deploying smart contracts via Hardhat with Viem. It includes configuration for both community-hosted and local ([Hiero Solo Node](https://github.com/hiero-ledger/solo)) instances of the [Hedera JSON RPC Relay](https://github.com/hiero-ledger/hiero-json-rpc-relay).
```
rbarker-dev:
```suggestion
This Hedera Hardhat Viem Example Project offers boilerplate code for testing and deploying smart contracts via Hardhat with Viem. It includes configuration for both community-hosted and local ([Solo](https://github.com/hiero-ledger/solo)) instances of the [Hedera JSON RPC Relay](https://github.com/hiero-ledger/hiero-json-rpc-relay). 
```

tools/subgraph-example/README.md
```diff
@@ -122,9 +126,9 @@ To deploy the subgraph:
 
 _NOTE: At this time the whole test workflow can't be proficiently automated, so you'll need to perform some manual steps:_
 
-1. Be sure to start a clean local hedera node. If the node is currently running stop it.
-   1. Run `npx hedera stop` to be sure that all containers are stopped and the temp files and volumes have been removed.
-   2. Run `npx hedera start --network local-test` to start a new clean node.
+1. Be sure to start a clean local node using Hiero Solo. If a previous local node is running, stop it.
```
rbarker-dev:
```suggestion
1. Be sure to start a clean Solo instance. If a previous Solo instance is running, stop it.
```

tools/subgraph-example/README.md
```diff
@@ -122,9 +126,9 @@ To deploy the subgraph:
...
+   1. Run `npx @hashgraph/solo falcon destroy` to ensure any previous local network is torn down.
```
rbarker-dev:
```suggestion
   1. Run `npx @hashgraph/solo falcon destroy` to ensure any previous solo network is torn down.
```

tools/subgraph-example/README.md
```diff
@@ -122,9 +126,9 @@ To deploy the subgraph:
...
+   2. Run `npx @hashgraph/solo one-shot falcon deploy --dev --deploy-relay=false` to start a clean single-node local network.
```
rbarker-dev:
```suggestion
   2. Run `npx @hashgraph/solo one-shot falcon deploy --dev --deploy-relay=false` to start a clean single-node instance of solo.
```

tools/waffle-example/.env.example
```diff
@@ -1,5 +1,5 @@
 # The operator private keys start with "0x" and are in hexadecimal format
 # Your account ECDSA hex-encoded private key (Ex: 0xb46751179bc8aa9e129d34463e46cd...)
-# All available local node private keys can be found here https://github.com/hashgraph/hedera-local-node#commands
+# All available local node private keys can be found here https://solo.hiero.org/docs/advanced-solo-setup/using-environment-variables/#operator-and-key-configuration
```
rbarker-dev:
```suggestion
# All available Solo private keys can be found here https://solo.hiero.org/docs/advanced-solo-setup/using-environment-variables/#operator-and-key-configuration
```

tools/waffle-example/README.md
```diff
@@ -12,7 +12,7 @@ RELAY_ENDPOINT=
 ```
 
  - `OPERATOR_PRIVATE_KEY` is your account ECDSA hex-encoded private key.
- - `RELAY_ENDPOINT` is a path to your JSON RPC Api. `https://testnet.hashio.io/api` for testnet. Remember to start your Hedera local node if you want to use the http://localhost:7546 endpoint.
+ - `RELAY_ENDPOINT` is a path to your JSON RPC Api. `https://testnet.hashio.io/api` for testnet. Remember to start your local node if you want to use the http://localhost:7546 endpoint.
```
rbarker-dev:
```suggestion
 - `RELAY_ENDPOINT` is a path to your JSON RPC Api. `https://testnet.hashio.io/api` for testnet. Remember to start your Solo instance if you want to use the http://localhost:7546 endpoint.
```

rbarker-dev (dismissed):

quiet-node (approved):
LGTM!

natanasow (approved):

PavelSBorisov (approved):

BartoszSolkaBD (approved):

PR #5261, opened 2026-04-14
feat: Add authorisationlist on type 4 tx

### LINKED ISSUE #5208: Pectra: AuthorizationList on eth_call eth_estimateGas
### Problem

With the inclusion of EIP-7702, a new type4 transaction exists and this implies changes to several rpc methods

We already have a support for auth list in the methods mentioned in this issue:
https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/4863

We need to:

- Send the authorization lis tot the  MN call endpoint (both for estimation and when doing the call)
- Implement the needed changes within the mentioned methods mentioned here https://ethereum.github.io/execution-apis/api/methods/eth_call
- https://ethereum.github.io/execution-apis/api/methods/eth_estimateGas
- Add the needed integration test for this

```
{
  "type": "0x4",
  "hash": "0x...example...",
  "from": "0x...the_authorizing_EOA...",
  "to": "0x...often_the_EOA_itself_or_delegator...",
  "authorizationList": [
    {
      "chainId": "0x1",                  // Mainnet
      "address": "0x63c0c19a282a1b52b07dd5a65b58948a07dae32b",  // Delegated-to contract (e.g. MetaMask delegator)
      "nonce": "0x5",                    // Signer's nonce at authorization time
      "yParity": "0x0",
      "r": "0xabc...long hex...",
      "s": "0xdef...long hex..."
    }
    // Possibly more entries for batch/multi-account delegation
  ],
  // ... other standard fields (nonce, gas, value, data, etc.)
}
```

### Solution

- Create a PR to the pectra Branch we have at relay ( we will merge all features at once into the main branch)
- New acceptance tests 

### Alternatives

_No response_

bootcodes:
### Description

This PR introduces support for EIP-7702 (authorizationList) in `eth_call`, `eth_estimateGas` as part of the Pectra upgrade.
A new `authorizationListEntry` validator schema is added with custom `yParityHex` types for per-entry validation.

### Related issue(s)

Fixes #5208

### Testing Guide

1. **Unit tests** : run the relay unit suite
2.
3.

- Per-entry relay-level validation of `authorizationListEntry` fields (address, chainId, nonce, r, s, yParity) was added for consistency with how the validator schema works for other array types - the relay validates format only.
- A dedicated test verifies that when a type 4 raw transaction string is parsed by `Precheck.parseRawTransaction`, ethers deserialises `authorizationList` entries into its internal `Authorization` format (bigint chainId/nonce, nested Signature object), which differs from the relay's `AuthorizationListEntry` shape. This format difference is harmless for `eth_sendRawTransaction`.
  
Acceptance tests for `eth_call/eth_estimateGas` with `authorizationList` are skipped pending Mirror Node support for the field in `/api/v1/contracts/call` - tracked in [hiero-ledger/hiero-mirror-node#12379](https://github.com/hiero-ledger/hiero-mirror-node/issues/12379). 
Un-skip the `describe.skip` block in `sendRawTransactionExtension.spec.ts` once that is merged.

### Checklist


src/relay/lib/precheck.ts
```diff
@@ -251,6 +252,19 @@ export class Precheck {
...
   }
 
+  /**
+   * Validates the authorization list entries for EIP-7702 (type 4) transactions.
+   *
+   * @param tx - The transaction to validate.
+   * @throws {JsonRpcError} If any entry contains an invalid address.
+   */
+  authorizationList(tx: Transaction): void {
+    if (Number(tx.type) === 0 && tx.authorizationList && tx.authorizationList.length > 0) {
```
BartoszSolkaBD: why not `tx.type !== 4`? What about other types of transactions? Maybe we should keep a list of not supported tx types? This would allow us to future proof this condition.

src/relay/lib/precheck.ts
```diff
@@ -251,6 +252,19 @@ export class Precheck {
     }
   }
 
+  /**
+   * Validates the authorization list entries for EIP-7702 (type 4) transactions.
+   *
+   * @param tx - The transaction to validate.
+   * @throws {JsonRpcError} If any entry contains an invalid address.
+   */
+  authorizationList(tx: Transaction): void {
```
BartoszSolkaBD, line 262: As per EIP-7702: "The transaction is considered invalid if the length of authorization_list is zero."

BartoszSolkaBD (changes requested):
Looks good! I'm not sure about the precheck for authorizationList and the validation being split between validators/objectTypes.ts and precheck.ts. Let me know what you think

src/relay/lib/precheck.ts
bootcodes: Correct only type 4 supports authorizationList - will fix this

bootcodes:
@BartoszSolkaBD - The split is intentional and follows the existing pattern in the codebase.

`objectTypes.ts` handles structural validation, it validates the shape of the authorizationList field (entry fields like address, chainId, nonce, r, s, yParity). This runs for all methods that accept a transaction object, including `eth_call` and `eth_estimateGas`.

eth_call/eth_estimateGas are decorated with @rpcParamValidationRules and `0: { type: 'transaction', required: true }`-> so it runs :
```
transaction: {
    test: (param: any) => {
      if (Object.prototype.toString.call(param) === '[object Object]') {
        return validateSchema(OBJECTS_VALIDATIONS.transaction, param);
      }

      return false;
    },
    error: 'Expected TransactionObject',
  },
```
and validates the structural shape of the transaction. 

`precheck.ts` handles business-rule validation. This runs exclusively in the sendRawTransaction path, which is the only path where a transaction is actually submitted and these invariants must hold before hitting the network.

So yes both paths do different things - and we don't (currently) enforce strict validation like in `precheck.ts` on @rpcParams decorated functions. Might be worth strengthening and/or moving that into a single shared harness.
Will see if I can enforce it here in objectTypes.ts

src/relay/lib/precheck.ts
```diff
@@ -251,6 +253,21 @@ export class Precheck {
...
+    validateAuthorizationList(Number(tx.type), tx.authorizationList);
+
+    // EIP-7702 mandates that tx.to must not be null
+    if (tx.type === 4 && tx.to == null) {
+      throw predefined.INVALID_PARAMETER('to', 'must not be null for type 4 transaction');
```
jasuwienas: nit: What do you think about using messages similar to the one used by geth or besu:

https://github.com/ethereum/go-ethereum/blob/master/core/error.go#L131
(I’d prefer to stick with a format used elsewhere so we can clearly justify the reasoning behind the chosen message: ))

jasuwienas (approved):
LGTM!

(PS: Type 3 txs also don’t support contract creation, but since we don’t support them, your current implementation is great. Nice catch!)
**I just have one global suggestion when it comes to the messages we used for errors**

src/relay/lib/precheck.ts
bootcodes: yes - that is good practice

bootcodes:
Close and recreated clean branch on #5292

PR #5271, opened 2026-04-15
feat: update sdk version and decode access list (#5139)

### LINKED ISSUE #5139: Support for AccessList in the hiero sdk package
### Problem

Creating EthereumTransactionData from raw transaction bytes currently fails when the transaction contains a non-empty access list.
Although the access list is decoded from the raw transaction, the current implementation of the `fromBytes` method from expects the access list to be an array of strings.

More details are included in this issue:

https://github.com/hiero-ledger/hiero-sdk-js/issues/3885

### Solution

If this issue requires changes in the SDK repository, we will need to:
 - wait for them to be implemented,
 -  update the package version we depend on in order to apply the fix in our code as well.

### Alternatives

_No response_

jasuwienas:
### Description

Updates sdk version to the one where access list works correctly.
Workaround for https://github.com/hiero-ledger/hiero-mirror-node/issues/13343 (mirror node currently returns encoded access list, so we have to decode in our end)

Additionally:
This PR removes the method toNullableBigNumber which was never used outside of its own tests so bignumber.js dependency can be removed entirely. Proposed in: https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5271#discussion_r3109144532

### Related issue(s)

Fixes #5139

### Testing Guide
1. Run tests in workflow - chekc everything is fine. 

N/A

We can't go main with beta version of sdk... We will have to wait for full release anyway.

It would also be great to have this decoding logic handled on the mirrornode side. But up till then we can do it on our end.

### Checklist


src/relay/lib/factories/transactionFactory.ts
```diff
@@ -120,15 +123,23 @@ const formatAuthorizationList = (authorizationList: any): AuthorizationListEntry
...
+    ? (decoded
+        .filter((_value, _index, item) => Array.isArray(item))
+        .map((_value, _index, [address, storageKeys]) => ({
+          address: formatAddress(!Array.isArray(address) ? toHexString(address) : null),
+          storageKeys: (Array.isArray(storageKeys) ? storageKeys : [])
+            .map((key) => (!Array.isArray(key) ? prepend0x(toHexString(key)) : null))
+            .filter(Boolean),
+        })) as AccessListEntry[])
     : [];
+};
```
bootcodes: nit: for readability 

```suggestion
  const decoded = RLP.decode(accessList);
  if (!Array.isArray(decoded)) return [];
  
  return decoded
    .filter(Array.isArray)
    .map(([address, storageKeys]) => {
      const storageKeysRes = (Array.isArray(storageKeys) ? storageKeys : [])
        .filter((key) => !Array.isArray(key))
        .map((key) => prepend0x(toHexString(key as Uint8Array)));

      return {
        address: formatAddress(Array.isArray(address) ? null : toHexString(address as Uint8Array)),
        storageKeys: storageKeysRes,
      };
    }) as AccessListEntry[];
};
```

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -114,19 +117,38 @@ const formatAuthorizationList = (authorizationList: any): AuthorizationListEntry
...
+ * @param hex
+ * @return decoded stream
+ */
+function* decodeStream(hex: string) {
+  let rest: string | Uint8Array = hex;
+  while (rest) {
+    const { data, remainder } = RLP.decode(rest, true);
+    yield data;
+    rest = remainder;
+  }
```
quiet-node: Looks like while (rest) could be risky here. RLP.decode(input, stream: true) always returns { data, remainder: Uint8Array }, and when the last entry is consumed, remainder is new Uint8Array(0). Since every object is truthy in JavaScript, the loop enters one extra unwanted iteration instead of stopping.

It currently doesn't hang only because @ethereumjs/rlp happens to return undefined for empty input, which finally breaks the loop, but that's undocumented behavior. If a future version of the library returns { data: ..., remainder: Uint8Array([]) } instead, this would become a true infinite loop.

Would it make sense to change the condition to while (rest.length > 0)? That checks whether there are actually bytes left to decode, rather than relying on the object's truthiness.

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -114,19 +117,38 @@ const formatAuthorizationList = (authorizationList: any): AuthorizationListEntry
...
-          address: formatAddress(item.address),
-          storageKeys: item.storageKeys ?? [],
+const formatAccessList = (accessList: unknown): AccessListEntry[] =>
+  typeof accessList === 'string' && isHex(accessList)
+    ? [...decodeStream(accessList)]
+        .filter((data) => data && data.length === 2)
+        .map(([address, storageKeys]) => ({
+          address: formatAddress(address instanceof Uint8Array ? toHexString(address) : null),
+          storageKeys: Array.isArray(storageKeys)
+            ? storageKeys.filter((item) => item instanceof Uint8Array).map((key) => prepend0x(toHexString(key)))
```
quiet-node: Should we add `.padStart(64, '0')` to the storage key `toHexString` call? RLP strips leading zeros, so a storage slot like `1` would be returned from RLP decode as a 1-byte `Uint8Array([0x01])`, producing `'0x01'` instead of the required 32-byte. The current acceptance tests use pre-padded 32-byte test values, so this would not be caught by the existing test suite.

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -114,19 +117,38 @@ const formatAuthorizationList = (authorizationList: any): AuthorizationListEntry
...
-  accessList && Array.isArray(accessList)
-    ? accessList
-        .filter((item: any) => item !== null && typeof item === 'object')
-        .map((item: any) => ({
-          address: formatAddress(item.address),
-          storageKeys: item.storageKeys ?? [],
+const formatAccessList = (accessList: unknown): AccessListEntry[] =>
+  typeof accessList === 'string' && isHex(accessList)
+    ? [...decodeStream(accessList)]
+        .filter((data) => data && data.length === 2)
```
quiet-node: Would it make sense to change `data && data.length === 2` to `Array.isArray(data) && data.length === 2`? The intent is to match RLP list items (which are arrays), but a 2-byte `Uint8Array` also satisfies `.length === 2` and is truthy. If MirrorNode ever returns a malformed entry, the check would pass incorrectly, producing a synthetic `{ address: ZERO_ADDRESS, storageKeys: [] }` entry in the response.

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -114,19 +117,38 @@ const formatAuthorizationList = (authorizationList: any): AuthorizationListEntry
...
  */
-const formatAccessList = (accessList: any): AccessListEntry[] =>
-  accessList && Array.isArray(accessList)
-    ? accessList
-        .filter((item: any) => item !== null && typeof item === 'object')
-        .map((item: any) => ({
-          address: formatAddress(item.address),
-          storageKeys: item.storageKeys ?? [],
+const formatAccessList = (accessList: unknown): AccessListEntry[] =>
+  typeof accessList === 'string' && isHex(accessList)
```
quiet-node: Should we keep the `Array.isArray` branch alongside the new hex-decoding path? The FIXME says this code will need to be reverted when MirrorNode fixes `mirror-node#13343`, but when that happens, the relay will silently return `[]` for all access lists (since the new code only handles hex strings) until the revert is deployed. Having both branches active at once avoids that regression window and makes the eventual revert safe to do incrementally.

tests/relay/lib/eth/eth_estimateGas.spec.ts
```diff
@@ -680,18 +680,17 @@ describe('@ethEstimateGas Estimate Gas spec', async function () {
...
-    // @ts-ignore
-    expect((result as JsonRpcError).code).to.equal(-32603);
-    // @ts-ignore
-    expect((result as JsonRpcError).message).to.contain('Test error for estimateGas');
+    expect(ethImpl.estimateGas(transaction, null, requestDetails)).to.eventually.be.rejected.and.satisfy(
+      (error: JsonRpcError) => {
+        expect(error.code).to.equal(-32603);
+        expect(error.message).to.contain('Test error for estimateGas');
+      },
+    );
```
quiet-node: let's make sure we add `await` before this `expect(...)` call. Without it, `chai-as-promised` returns a `Promise` that is never awaited, so the test always passes regardless of whether the rejection actually happens. All the other rejection assertions in this file use `await expect(...)` — this one was missed during the refactor.

tests/server/acceptance/sendRawTransactionExtension.spec.ts
```diff
@@ -133,50 +136,110 @@ describe('@sendRawTransactionExtension Acceptance Tests', function () {
...
+              storageKeys: [`${prepend0x('00'.repeat(31))}01`, `${prepend0x('00'.repeat(31))}02`],
+            },
+            {
+              address: ACCESS_LIST_TEST_ADDRESS_2,
+              storageKeys: [`${prepend0x('00'.repeat(31))}03`, `${prepend0x('00'.repeat(31))}04`],
+            },
+          ],
+        },
+        {
+          label: 'non-empty access list with multiple addresses and multiple storage keys',
```
quiet-node: Duplicated test labels?

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -114,19 +117,38 @@ const formatAuthorizationList = (authorizationList: any): AuthorizationListEntry
         }))
     : [];
 
+/**
+ * Decodes a hex string into an array of RLP-encoded items. Since the MirrorNode does NOT return the full RLP encoding,
+ * with root included, we need to decode the data we get from them, streaming it part by part.
+ * @param hex
+ * @return decoded stream
+ */
+function* decodeStream(hex: string) {
```
quiet-node, line 145: Should we add some tests for this guy and `formatAccessList`? They have some non-obvious behavior (RLP streaming, byte-array-to-hex conversion, leading-zero handling), and since this is a temporary workaround with a FIXME, having unit tests would make the eventual revert much safer to verify.

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -114,19 +117,38 @@ const formatAuthorizationList = (authorizationList: any): AuthorizationListEntry
...
+  while (rest) {
+    const { data, remainder } = RLP.decode(rest, true);
+    yield data;
+    rest = remainder;
+  }
+}
+
 /**
  * Formats an access list by normalizing and sanitizing its fields.
+ * FIXME (mirror-node#13343): this code fragment has to be reverted to the previous version when mirror node
```
quiet-node: Should we open a relay-side tracking issue for the FIXME revert and add it to the comment alongside `mirror-node#13343`? That way the revert doesn't depend on someone watching the external mirror-node issue but it would show up in the relay's own backlog.

package.json
```diff
@@ -107,13 +107,15 @@
     "@ethereumjs/rlp": "^10.0.0",
     "@ethereumjs/trie": "^6.2.1",
     "@ethereumjs/util": "^9.1.0",
-    "@hashgraph/sdk": "^2.80.0",
+    "@ethersproject/keccak256": "^5.8.0",
+    "@hiero-ledger/sdk": "^2.83.0-beta.2",
```
quiet-node, line 111: Do we want to use this beta version? Is there a more reliable GA for this?

jasuwienas: 1. Not on main (we will have to remember to switch to non-beta version before releasing pectra related stuff).
2. Not yet.

src/relay/lib/factories/transactionFactory.ts
jasuwienas: Right. It must've really work by accident, didn't notice. Fixed already.

src/relay/lib/factories/transactionFactory.ts
jasuwienas: Yes, it makes sense actually. Done.

src/relay/lib/factories/transactionFactory.ts
jasuwienas: Done

src/relay/lib/factories/transactionFactory.ts
jasuwienas: Ok, we can keep both versions for safety. - Reworked it , let me know if that is the effect we want to achieve :)

tests/server/acceptance/sendRawTransactionExtension.spec.ts
jasuwienas: yes , one of them was wrong, updated it

src/relay/lib/factories/transactionFactory.ts
jasuwienas: I've included the tests for it (similar to one we have for authorization list)

package.json
```diff
@@ -107,13 +107,15 @@
...
     "@hoppscotch/cli": "0.30.0",
     "@koa/cors": "^5.0.0",
     "async-mutex": "^0.5.0",
     "axios": "^1.4.0",
     "axios-retry": "^4.5.0",
     "better-lookup": "^1.3.0",
+    "bignumber.js": "^10.0.2",
```
natanasow: Q: Do we really need this package? I just saw it's used in 2 files, do you want to remove it and replace the usage with `BigInt`?

jasuwienas: That;s actually a good idea! Thanks, I'll remove big number usage. There might not be a better time to do that.

jasuwienas: >  I just saw it's used in 2 files

it was actually much stranger than that... We used bigNumber in method toNullableBigNumber, which was never used in our application outside of the test case scenarios testing... this method itself :) .

I got rid of it, and dropped the bignumber.js usage.

natanasow (approved):

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -114,30 +118,69 @@ const formatAuthorizationList = (authorizationList: any): AuthorizationListEntry
...
+
 /**
  * Formats an access list by normalizing and sanitizing its fields.
  *
  * @param {any} accessList - The raw access list.
  * @returns {AccessListEntry[]} A normalized access list.
  */
-const formatAccessList = (accessList: any): AccessListEntry[] =>
-  accessList && Array.isArray(accessList)
+const formatAccessList = (accessList: any): AccessListEntry[] => {
```
simzzz: Should we replace any with unknown here?

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -114,30 +118,69 @@ const formatAuthorizationList = (authorizationList: any): AuthorizationListEntry
...
+ */
+function* decodeStream(hex: string) {
+  let rest: string | Uint8Array = hex;
+  while (rest.length > 0) {
+    const { data, remainder } = RLP.decode(rest, true);
+    yield data;
+    rest = remainder;
+  }
+}
+
```
simzzz, line 153: `RLP.decode` throws on malformed or truncated input. If the mirror node ever returns an invalid hex blob, this propagates as an unhandled exception through `eth_getTransactionByHash` / `eth_getBlockByNumber` instead of a graceful empty list. Should we wrap it in a try/catch?

simzzz: Also worth adding a unit test for malformed input

simzzz (commented):
LG, just some small comments

src/relay/lib/factories/transactionFactory.ts
jasuwienas: Included it!

simzzz (approved):

package.json
jasuwienas: I've created an issue for that:
https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/5314

We'll review all of the pectra related changes when merging them to main, we'll have to update this version before doing so.

PR #5275, opened 2026-04-16
fix: clear WebSocket ping interval on connection close (#5233)

BartoszSolkaBD:
### Description

This PR fixes a resource leak where the `setInterval` used for WebSocket ping messages was never cleared when a connection closed. Each closed connection left an orphaned interval running indefinitely, accumulating over time and consuming memory/CPU.

The fix stores the interval ID on `ctx.websocket.pingIntervalId` when created in `initializeWsServer`, then calls `clearInterval` in `handleConnectionClose` before terminating the socket.

### Related issue(s)

Fixes #5233

### Testing Guide

1. Run the WebSocket server unit tests:
``` bash
npm run test:ws-server
```
2. Verify the `timer cleanup` tests in `utils.spec.ts` pass. They assert `clearInterval` is called with the correct interval ID on close, and that missing timers are handled gracefully.
3. Verify the `should set up ping interval when WS_PING_INTERVAL > 0` test in `webSocketServer.spec.ts` passes. It asserts `setInterval` is called during connection setup when `WS_PING_INTERVAL > 0`.

N/A

- `inactivityTTL` (`setTimeout` in `connectionLimiter.ts:startInactivityTTLTimer`) is also not cleared on connection close, only on activity reset. This is a separate potential leak that should be evaluated in a follow-up.

### Checklist

### DIFF
--- a/src/ws-server/utils/utils.ts
+++ b/src/ws-server/utils/utils.ts
@@ -53,7 +53,7 @@ export const handleConnectionClose = async (
     clearInterval(ctx.websocket.pingIntervalId);
   }
   if (ctx.websocket?.inactivityTTL) {
-    clearInterval(ctx.websocket.inactivityTTL);
+    clearTimeout(ctx.websocket.inactivityTTL);
   }
 
   // terminate connection
--- a/src/ws-server/utils/utils.ts
+++ b/src/ws-server/utils/utils.ts
@@ -52,6 +52,9 @@ export const handleConnectionClose = async (
   if (ctx.websocket?.pingIntervalId) {
     clearInterval(ctx.websocket.pingIntervalId);
   }
+  if (ctx.websocket?.inactivityTTL) {
+    clearInterval(ctx.websocket.inactivityTTL);
+  }
 
   // terminate connection
   ctx.websocket.terminate();
--- a/src/ws-server/utils/utils.ts
+++ b/src/ws-server/utils/utils.ts
@@ -48,6 +48,11 @@ export const handleConnectionClose = async (
   // Update the connection duration histogram with the calculated duration
   wsMetricRegistry.getHistogram('connectionDuration').observe(durationInSeconds);
 
+  // clear timers to prevent orphaned resources
+  if (ctx.websocket?.pingIntervalId) {
+    clearInterval(ctx.websocket.pingIntervalId);
+  }
+
   // terminate connection
   ctx.websocket.terminate();
 };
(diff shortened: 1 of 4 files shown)


src/ws-server/utils/utils.ts
```diff
@@ -48,6 +48,11 @@ export const handleConnectionClose = async (
   // Update the connection duration histogram with the calculated duration
   wsMetricRegistry.getHistogram('connectionDuration').observe(durationInSeconds);
 
+  // clear timers to prevent orphaned resources
+  if (ctx.websocket?.pingIntervalId) {
+    clearInterval(ctx.websocket.pingIntervalId);
+  }
+
```
quiet-node, line 58: Nice one! I see the PR mentioned inactivityTTL, should we take the chance and also clear `ctx.websocket.inactivityTTL` here in the same pass? Addressing them in one place might make the memory improvement easier to verify. 

Happy to do it as a follow-up if you prefer to keep this PR tightly scoped.

tests/ws-server/unit/utils.spec.ts
```diff
@@ -183,6 +183,70 @@ describe('Utilities unit tests', async function () {
...
+        clearIntervalSpy = sinon.spy(global, 'clearInterval');
+      });
+
+      afterEach(() => {
+        clearIntervalSpy.restore();
+      });
+
+      it('should clear the ping interval when closing a connection', async () => {
+        const intervalId = setInterval(() => {}, 100000);
+        const ctxWithPing: any = {
```
quiet-node: Let's address these small codacy warnings in this file

quiet-node (commented):
LG but one small item and one question

src/ws-server/utils/utils.ts
```diff
@@ -48,6 +48,14 @@ export const handleConnectionClose = async (
   // Update the connection duration histogram with the calculated duration
   wsMetricRegistry.getHistogram('connectionDuration').observe(durationInSeconds);
 
+  // clear timers to prevent orphaned resources
+  if (ctx.websocket?.pingIntervalId) {
+    clearInterval(ctx.websocket.pingIntervalId);
+  }
+  if (ctx.websocket?.inactivityTTL) {
+    clearInterval(ctx.websocket.inactivityTTL);
```
jasuwienas: are you sure its interval, not timeout? @BartoszSolkaBD

natanasow (dismissed):

jasuwienas (approved):
LGTM!

simzzz (approved):

PR #5276, opened 2026-04-16
fix: debug_transaction can return to: null

### LINKED ISSUE #5244: bug: debug_traceTransaction returns `to: "0x0"` for pre-check–rejected CREATE txs
## Summary

For transactions that are rejected by the Hedera consensus node *before* any EVM execution (e.g. `INSUFFICIENT_TX_FEE`, `WRONG_NONCE`), `debug_traceTransaction` / `debug_traceBlockBy*` with `callTracer` returns `to: "0x0"` instead of `null`.

- `"0x0"` is not a valid 20-byte address — downstream indexers that parse the field as an address choke on it.
- The Ethereum convention for contract-create / no-recipient traces is `to: null`.
- The same bug also means `type` is reported as `"CALL"` for what are actually `CREATE` transactions.

## Affected transactions (Hedera mainnet)

Observed in production with these four records. All four are contract-create (`to: null` on the raw EVM tx) rejected at Hedera pre-check, so the mirror node returns a `contracts/results/{hash}` row but an empty `actions` array.

| Block | Tx hash | Sender | Nonce | Mirror node `result` |
|---|---|---|---|---|
| [92991661](https://hashscan.io/mainnet/block/92991661) | [`0x913e3e8ea8701961e5d62c32ceb469e0d9eebce67df71cf63d284a0aba0494da`](https://hashscan.io/mainnet/transaction/0x913e3e8ea8701961e5d62c32ceb469e0d9eebce67df71cf63d284a0aba0494da) | `0x…008dfa00` (`0.0.9304064`) | 185 | `INSUFFICIENT_TX_FEE` |
| [92991707](https://hashscan.io/mainnet/block/92991707) | [`0x913e3e8ea8701961e5d62c32ceb469e0d9eebce67df71cf63d284a0aba0494da`](https://hashscan.io/mainnet/transaction/0x913e3e8ea8701961e5d62c32ceb469e0d9eebce67df71cf63d284a0aba0494da) *(same)* | `0x…008dfa00` | 185 | `INSUFFICIENT_TX_FEE` |
| [92992948](https://hashscan.io/mainnet/block/92992948) | [`0x913e3e8ea8701961e5d62c32ceb469e0d9eebce67df71cf63d284a0aba0494da`](https://hashscan.io/mainnet/transaction/0x913e3e8ea8701961e5d62c32ceb469e0d9eebce67df71cf63d284a0aba0494da) *(same)* | `0x…008dfa00` | 185 | `INSUFFICIENT_TX_FEE` |
| [93659985](https://hashscan.io/mainnet/block/93659985) | [`0x681f782e745e678c7aa2541877bbd3baa2ed2d409d65f120bf1d45d81ad89b35`](https://hashscan.io/mainnet/transaction/0x681f782e745e678c7aa2541877bbd3baa2ed2d409d65f120bf1d45d81ad89b35) | `0x…009f2ccf` (`0.0.10431695`) | 87 | `WRONG_NONCE` |

Authoritative mirror-node data for each:

- https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/results/0x913e3e8ea8701961e5d62c32ceb469e0d9eebce67df71cf63d284a0aba0494da
- https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/results/0x913e3e8ea8701961e5d62c32ceb469e0d9eebce67df71cf63d284a0aba0494da/actions → `{"actions": []}`
- https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/results/0x681f782e745e678c7aa2541877bbd3baa2ed2d409d65f120bf1d45d81ad89b35
- https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/results/0x681f782e745e678c7aa2541877bbd3baa2ed2d409d65f120bf1d45d81ad89b35/actions

All four records have `to: null`, `address: null`, `contract_id: null`, `created_contract_ids: []`, `gas_used: 0`, `gas_consumed: 0`, `status: "0x0"`, and an empty actions array.

> Side note: blocks 92991661 / 92991707 / 92992948 are the **same signed raw tx resubmitted three times** (same hash, sender, nonce 185). Its `maxFeePerGas` is `0x44` — it can never succeed. Probably worth alerting the operator of `0.0.9304064` separately.

## Root cause

1. Raw EVM CREATE (`to: null`) is wrapped by the relay into a HAPI `EthereumTransaction`/`ContractCreate`.
2. Hedera rejects at pre-check (`INSUFFICIENT_TX_FEE`, `WRONG_NONCE`, etc.) — no EVM execution.
3. Mirror node records the contract-result with `to: null` and returns `{"actions": []}` from the actions endpoint.
4. \`DebugImpl.callTracer\` at [\`src/relay/lib/debug.ts:644\`](../blob/main/src/relay/lib/debug.ts#L644) fetches both and enters the \"contract result but no actions\" branch at [\`debug.ts:666\`](../blob/main/src/relay/lib/debug.ts#L666).
5. \`resolveMultipleAddresses(from, transactionsResponse.to=null, …)\` passes \`null\` through unchanged (\`resolveAddress\` short-circuits on falsy input at [\`debug.ts:544\`](../blob/main/src/relay/lib/debug.ts#L544)).
6. [\`debug.ts:676\`](../blob/main/src/relay/lib/debug.ts#L676) evaluates \`resolvedTo ?? constants.ZERO_HEX\` → \`constants.ZERO_HEX\`.
7. [\`constants.ts:239\`](../blob/main/src/relay/lib/constants.ts#L239) defines \`ZERO_HEX: '0x0'\` — note this is **not** \`ZERO_ADDRESS_HEX\` (40 zeros) defined on the very next line.
8. The string \`'0x0'\` is placed into \`CallTracerResult.to\` and returned.

The same fallback exists in [\`getEmptyTracerObject\`](../blob/main/src/relay/lib/debug.ts#L1003) at \`debug.ts:1002-1003\`, which is used by \`handleSyntheticTransaction\` for HTS-synthetic txs — same bug, different code path.

## Reproduction

```bash
curl -s -X POST -H 'Content-Type: application/json' <your-relay-url> -d '{
  "jsonrpc":"2.0","id":1,"method":"debug_traceTransaction",
  "params":["0x913e3e8ea8701961e5d62c32ceb469e0d9eebce67df71cf63d284a0aba0494da",
            {"tracer":"callTracer"}]
}'
```

Expected: `"to": null`, `"type": "CREATE"`
Actual: `"to": "0x0"`, `"type": "CALL"`

## Suggested fix

In \`src/relay/lib/debug.ts\`:

1. **Line 676** — drop the fallback, let \`null\` flow through:
   ```diff
   - resolvedTo ?? constants.ZERO_HEX,
   + resolvedTo,
   ```
2. **Lines 1002-1003** — same treatment in \`getEmptyTracerObject\`. Don't default \`resolvedFrom\`/\`resolvedTo\` to \`ZERO_HEX\`; accept \`null\`/\`undefined\`.
3. **Line 1001** — \`getEmptyTracerObject\` hard-codes \`type: CallType.CALL\`. When the caller knows the tx was a CREATE (\`transactionsResponse.to == null\`), propagate \`type: CallType.CREATE\` so the resulting trace is shape-correct for the tx class.
4. Add a regression test covering the \"contract-result + empty actions + \`to: null\`\" path using one of the hashes above as fixture data.

Optionally also audit callers of \`constants.ZERO_HEX\` — \`'0x0'\` is the wrong constant for anything that is meant to be an *address*; \`constants.ZERO_ADDRESS_HEX\` (\`'0x' + '0'.repeat(40)\`) is the address-shaped one.

## Impact

Any downstream tool calling \`debug_traceTransaction\` or \`debug_traceBlockBy*\` on a block containing a pre-check-rejected CREATE will receive a non-address string in the \`to\` field. Block-level trace consumers (indexers, explorers) hit this whenever such a tx lands in a block, which is why it surfaced across several recent mainnet blocks.

bootcodes:
### Description

This PR introduces a bug fix for `debug_transaction` method that returned `to: "0x0"` on rejected transactions on Hedera consensus node before EVM execution. 

### Related issue(s)

Fixes #5244

### Testing Guide

1. Run test suite 
2. Spin up a local relay instance with env variable set to mainnet and run the following command :
```
curl -s -X POST -H 'Content-Type: application/json'  http://localhost:7546/api -d '{
  "jsonrpc":"2.0","id":1,"method":"debug_traceTransaction",
  "params":["0x681f782e745e678c7aa2541877bbd3baa2ed2d409d65f120bf1d45d81ad89b35",
            {"tracer":"callTracer"}]
}'
```
3. Verify that all the transactions listed in #5244 return `"type":"CREATE"` and `"to":null`

N/A

N/A

### Checklist


src/relay/lib/types/debug.ts
```diff
@@ -47,9 +47,9 @@ export interface CallTracerResult {
   /** The type of the call (e.g., 'CALL', 'CREATE', etc.). */
   type: string;
   /** The address initiating the call. */
-  from: string;
+  from: string | null | undefined;
```
BartoszSolkaBD: when it can be null or undefined? If there is no such case, maybe we should stick to just `string`

src/relay/lib/types/debug.ts
```diff
@@ -47,9 +47,9 @@ export interface CallTracerResult {
...
   /** The address receiving the call. */
-  to: string;
+  to: string | null | undefined;
```
BartoszSolkaBD: when it can be undefined?

src/relay/lib/debug.ts
BartoszSolkaBD:
```suggestion
async resolveMultipleAddresses(
  from: string,
  to: string | null,
  requestDetails: RequestDetails,
): Promise<{ resolvedFrom: string; resolvedTo: string | null }>
```

BartoszSolkaBD:
```suggestion
async resolveAddress(
  address: string | null,
  requestDetails: RequestDetails,
  types: string[] = [...],
): Promise<string | null>
```

BartoszSolkaBD (changes requested):
Nice job! I think we can tighten the types here easily without any tradeoffs. Please let me know what you think.

If we decide to go this way, please make sure that `to` is typed as `string | null` in all the places. I think `MirrorNodeContractResult.to` should be aligned as well.

src/relay/lib/types/debug.ts
bootcodes: So undefined is there because in 

```
private getEmptyTracerObject(
    tracer: TracerType,
    resolvedFrom?: string,
    resolvedTo?: string,
  )
``` 
`resolvedFrom` and `resolvedTo` can be optional 
And we remove all `?? constants.ZERO_HEX` for `resolvedFrom` and `resolvedTo` so those types propagate through.

I added null type because in [here](https://mainnet-public.mirrornode.hedera.com/api/v1/contracts/results/0x913e3e8ea8701961e5d62c32ceb469e0d9eebce67df71cf63d284a0aba0494da), the "to" field is null. 

Also :
```
  async resolveAddress(
    address: string,
    types: string[] = [constants.TYPE_CONTRACT, constants.TYPE_TOKEN, constants.TYPE_ACCOUNT],
  ): Promise<string> {
    // if the address is null or undefined we return it as is
    if (!address) return address;
```


Will debug this to see what is actually propagated (empty string or null)

jasuwienas: CallTracerResult is the object we are returning as a final output for the reulay method eth_traceTransaction call.

If I'm reading the docs correctly:
https://docs.chainstack.com/reference/ethereum-tracetransaction#calltracer-response

here `undefined` should **not** be allowed for `from` / `to` fields. But `to` should be nullable.
(correct me if I'm wrong 🙈 )

jasuwienas: > So undefined is there because in

emptyTracerObject can return not only CallTracerResult but also other tracers. In that case - undefined is acceptable. But I think we should sanitize / validate it for calltracer

bootcodes: yes I will sanitize `emptyTracerObject` to be more strict on return type for `CallTracerResult`

src/relay/lib/debug.ts
```diff
@@ -551,12 +550,12 @@ export class DebugImpl implements Debug {
    * @returns {Promise<string>} The address returned as an EVM address.
    */
   async resolveAddress(
-    address: string,
+    address: string | null,
```
quiet-node, line 554: Nice one but would it make sense to also update the `to` parameter of `resolveMultipleAddresses` to `string | null` here? Right now the return type correctly carries `resolvedTo: string | null`, but the input parameter still says `to: string`. Since `transactionsResponse.to` can be `null` (the whole scenario this PR is fixing), threading the nullable type all the way through would close the remaining type gap. Same applies to `MirrorNodeContractResult.to` in `mirrorNode.ts`.

src/relay/lib/debug.ts
```diff
@@ -15,8 +15,7 @@ import {
   tinybarsToWeibars,
   toHexString,
 } from '../formatters';
-import { type Debug } from '../index';
-import { JsonRpcError } from '../index';
+import { type Debug, JsonRpcError } from '../index';
```
quiet-node: Should we keep the `type` qualifier on `JsonRpcError` here? It looks like commit `6e820e289` added `type JsonRpcError` as part of a project-wide `import type` cleanup. The consolidation into a single import line dropped it. Since `JsonRpcError` is only used as a return type annotation, `import { type Debug, type JsonRpcError } from '../index'` would preserve that convention.

src/relay/lib/debug.ts
quiet-node: Btw lets update TSDoc as well

src/relay/lib/debug.ts
```diff
@@ -1006,34 +1001,41 @@ export class DebugImpl implements Debug {
    * @param resolvedTo - Optional resolved 'to' address (used by CallTracer).
```
quiet-node, line 1007: Would it make sense to add a note here that `null` means CREATE and non-null means CALL? The null-driven branching is the heart of this fix, and documenting it here would help future maintainers avoid accidentally breaking it.

src/relay/lib/debug.ts
```diff
@@ -590,7 +589,7 @@ export class DebugImpl implements Debug {
       this.resolveAddress(to, requestDetails, [constants.TYPE_CONTRACT, constants.TYPE_TOKEN, constants.TYPE_ACCOUNT]),
     ]);
 
-    return { resolvedFrom, resolvedTo };
+    return { resolvedFrom: resolvedFrom!, resolvedTo };
```
quiet-node: Looks like codacy is complaining.

src/relay/lib/debug.ts
```diff
@@ -551,12 +550,12 @@ export class DebugImpl implements Debug {
...
   async resolveAddress(
-    address: string,
+    address: string | null,
     requestDetails: RequestDetails,
     types: string[] = [constants.TYPE_CONTRACT, constants.TYPE_TOKEN, constants.TYPE_ACCOUNT],
-  ): Promise<string> {
+  ): Promise<string | null> {
     // if the address is null or undefined we return it as is
-    if (!address) return address;
+    if (address == null) return null;
```
quiet-node, line 559: Just a small note: the guard changed from `if (!address)` (which also caught empty strings) to `if (address == null)`. Would it make sense to explicitly handle `address === ''` as well, or was the empty-string case intentionally dropped? Just want to make sure this isn't a latent edge case.

bootcodes: Yes I know, I don't think the comment reflects exactly what happened here. It catched empty strings and returned that, but it has nothing to do with null or undefined. 
I don't think there are scenarios where address is an empty string. Can leave it in place if we decide it

simzzz (dismissed):

src/relay/lib/debug.ts
quiet-node: I think this part might still need one more pass. The non-null assertion is gone now, but it was replaced with `@ts-ignore`, so TypeScript still is not being shown why `resolvedFrom` is safe here. Would it make sense to replace the suppression with an explicit guard instead? Something like:

```ts
if (resolvedFrom == null) {
  throw predefined.INTERNAL_ERROR('Missing resolved sender address');
}

return { resolvedFrom, resolvedTo };
```

That keeps `resolvedTo` nullable for the CREATE case, but makes the `resolvedFrom` invariant explicit to both readers and the type checker.

src/relay/lib/debug.ts
quiet-node: Yeah that's true! Thanks for the explanation!

quiet-node (approved):
LGTM! Great work!

jasuwienas (commented):
LGTM!

jasuwienas (approved):

Issue #101, opened 2022-05-26
Cannot pull container image


rafaelkallis:
### Description

I am trying to pull the container image locally using `docker pull "ghcr.io/hashgraph/hedera-json-rpc-relay:main"`, but I am not authorized to access the resource. Using fedora and macos.

![image](https://user-images.githubusercontent.com/9661903/170455669-d339ff1b-a3fc-4200-a86c-f16987531583.png)

I have tried this both with and without a github access token, as mentioned in your FAQs.

### Steps to reproduce

`docker pull "ghcr.io/hashgraph/hedera-json-rpc-relay:main"`

### Hedera network

other

### Version

na

### Operating system

Linux

Nana-EC:
Hi @rafaelkallis , thanks for bringing this up.
Can you confirm that you've done a `docker login ghcr.io` recently?
Sometimes a token may have expired. 
In which case a `docker logout ghcr.io` followed by a `docker login ghcr.io` may resolve your issue.

I encountered this myself when writing the documentation.

rafaelkallis:
@Nana-EC the access token has not expired yet. Also note that I have tried pulling ghcr images from other repos and it worked just fine.

You recently went public with the repo, could it be that the container image still is not publicly accessible?

Nana-EC:
Thanks @rafaelkallis that helps a bit.

Please try one more time as I did indeed have one more setting to apply.
If that doesn't work i'm look into this more later today/over the weekend.

Nana-EC:
Awesome. Thanks for bringing this up.
Please note the images represent early versions of heavy work in progress items.

Let us know if you find anything or feel free to open up other tickets or PRs with suggestions.

Issue #11, opened 2022-04-15
Env variable for private key & account for local contract call


Kalina-Todorova:
We need an env variable for private key and account for local contract call - and have it be dedicated for this, ex: eth_call - one account, eth_getCode - separate account, etc..

PR #5287, opened 2026-04-17
fix: improve Redis acquireLock logic to address stuck requests (#5260)

### LINKED ISSUE #5260: Fix issues with the rpc_relay_lock_waiting_txns metric
### Problem

In our lock strategies (Redis and LRU), we increment the gauge rpc_relay_lock_waiting_txns by 1 before a process starts waiting for a lock.

Once the process acquires the lock and finishes execution, it should decrement this gauge by 1. This means each task should not leave any long-term impact on the metric.

However, if the process handling a transaction is killed or stopped after incrementing the gauge-but before decrementing it-the metric becomes inaccurate and permanently increased.

### Solution

We should consider whether there is a way to periodically correct or reset this value. If such a solution is possible, we should implement it.

### Alternatives

We may need to reconsider whether this gauge is appropriate.

One option would be to replace it with a counter (e.g., tracking total wait events), or explore a different approach to representing this behavior more reliably?

BartoszSolkaBD:
### Description

Fixes a drift in the `rpc_relay_lock_waiting_txns` Prometheus gauge. The gauge is incremented when a transaction joins the Redis lock queue and decremented in finally, so drift means acquireLock promises are permanently stuck in the poll loop.

Root cause: the poll loop in `RedisLockStrategy.acquireLock` assumed a session remained in the Redis queue once `lPush`ed. When that assumption breaks (Redis restart without persistence, failover to a replica that never saw the lPush, maxmemory eviction, or **another waiter zombie-cleaning this session after an event-loop stall** exceeding pollIntervalMs × heartbeatMissedCount), `firstInQueue === sessionKey` check never succeeds and the code silently falls through both the head-of-queue and zombie-cleanup branches into sleep() forever. `finally` never runs, the gauge stays pinned, and the heartbeat continues to refresh so the session looks alive to peers.

### Related issue(s)

Fixes #5260

### Testing Guide

1. Run Relay with `REDIS_ENABLED=true, ENABLE_NONCE_ORDERING=true, USE_ASYNC_TX_PROCESSING=true`
2. Fire traffic: x concurrent eth_sendRawTransaction from the same sender (at least 30-40)
3. Mid-burst: docker kill <redis>; docker start <redis> to force an empty Redis on reconnect (this simulates the issue, but as well you can simulate event-loop stall and zombie-cleanup)
4. Wait ~30s for traffic to drain

Expected before the fix (main branch): `rpc_relay_lock_waiting_txns` gauge pinned at non-zero indefinitely
Expected after the fix: `rpc_relay_lock_waiting_txns` returns to 0, `rpc_relay_lock_queue_rejoins_total` is non-zero, subsequent transactions continue to succeed.

Initial issue suggested periodically resetting/correcting the gauge. This PR aims to fix the issue that causes the gauge to drift.

N/A

N/A

### Checklist


src/relay/lib/services/lockService/RedisLockStrategy.ts
```diff
@@ -72,10 +74,30 @@ export class RedisLockStrategy implements LockStrategy {
 
       // Poll until first in queue and can acquire lock
       while (true) {
+        const elapsedMs = Number(process.hrtime.bigint() - startTime) / 1e6;
+        if (elapsedMs > this.maxQueueTimeMs) {
+          throw new Error(
```
jasuwienas: Just to confirm: should we really kill such transactions and not process them at all? :)

src/relay/lib/services/lockService/RedisLockStrategy.ts
```diff
@@ -72,10 +74,30 @@ export class RedisLockStrategy implements LockStrategy {
...
+        }
+
         // Refresh own heartbeat of the active waiter (Proof of Life)
         // note: `1` is just a placeholder value and doesn't matter, only TTL matters
         await this.redisClient.set(heartbeatKey, '1', { PX: this.heartbeatTtlMs });
 
+        // Verify membership - do not assume we are still queued. Redis may have
+        // been restarted/failover'd/evicted, or another waiter may have
+        // zombie-cleaned this session after an event-loop stall.
+        const position = await this.redisClient.lPos(queueKey, sessionKey);
```
jasuwienas: (I dont know what impact it will have on performance 🙊 )

If we use this lPos do we still need to call lIndex soon after? If our sessionKey has first index assinged it should mean it is actually first in the queue, rght?

src/relay/lib/services/lockService/RedisLockStrategy.ts
quiet-node: Well it's not going to be entirely discarded, but rather be fail-open, i.e. process without lock right?

But as disuccsed in daily I think we should reconsider whether LOCK_MAX_QUEUE_TIME_MS is needed here. The lPos rejoin mechanism already prevents the stuck loop: an evicted session rejoins the queue and eventually reaches the front, so `finally` always runs and the gauge always recovers. The stuck loop was the only source of gauge drift.

The concern with the hard deadline is that it conflicts with the original design decision for the while loop: a single sender can submit 100+ transactions that all queue up, and the last one has a legitimately long wait. With the 300s default and say there's a ~3s average consensus latency, TX #101 from a busy sender would already hit the deadline and fail-open (proceed without the lock), silently breaking nonce ordering. At worst-case LOCK_MAX_HOLD_MS of 30s, even TX #11 would hit it.

Maybe we can remove LOCK_MAX_QUEUE_TIME_MS entirely and let the lPos rejoin carry the fix. Or if we want a last-resort escape valve, should the default be much higher (e.g., LOCK_MAX_HOLD_MS * max expected queue depth)?

src/relay/lib/services/lockService/RedisLockStrategy.ts
quiet-node: Yeah I think we should call lPos only when we've detected the suspicious stuck pattern rather than on every iteration? In the happy path, lPos always returns non-null and the result is immediately discarded, but it's O(N) every 50ms for every waiting session. The head-of-queue session (the one actively trying toacquire the lock, at the list tail) pays the worst-case full scan each cycle.

One option: only call lPos after N consecutive iterations where neither `firstInQueue === sessionKey` (not first) nor a zombie was found (no progress). That's the exact stuck-loop pattern this check is designed to catch. Or use `lPos(queueKey, sessionKey, { RANK: -1 })` which scans from the right, giving 0 if we are at the FIFO tail, and also covers the rejoin check, reducing two round-trips to one. lmk what you think.

docs/configuration.md
```diff
@@ -117,6 +117,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
+| `LOCK_MAX_QUEUE_TIME_MS`                               | "300000"                                                                                                                                                                                                                                                                          | Maximum time in milliseconds a waiter can spend in the Redis lock queue before aborting the acquisition. Default is 5 minutes.                                                                                                                                                                                                                                                                                               |
```
quiet-node: If we do keep this, should we document that exceeding this deadline causes fail-open behavior where the transaction still processes but without the nonce-ordering guarantee? Right now "aborting the acquisition" could be read as the transaction failing, when it actually proceeds silently without a lock. This is especially important to call out given the design intent of the while loop (supporting long queues for high-volume senders). A note like "fail-open, nonce ordering not guaranteed if triggered, set above max_queue_depth x LOCK_MAX_HOLD_MS" would help operators choose a safe value.

src/relay/lib/services/lockService/RedisLockStrategy.ts
```diff
@@ -76,6 +79,20 @@ export class RedisLockStrategy implements LockStrategy {
...
+            await this.redisClient.lPush(queueKey, sessionKey);
+            this.lockMetricsService.recordQueueRejoin(this.type);
+            if (this.logger.isLevelEnabled('debug')) {
+              this.logger.debug(`Session not in queue; rejoined: address=${address}, sessionKey=${sessionKey}`);
+            }
+          }
+        }
+
         // Check if first in line
         const firstInQueue = await this.redisClient.lIndex(queueKey, -1);
```
jasuwienas, line 95: nit: Cant we do:
```
firstInQueue = position === 0 || await lIndex
```
instead here?

src/relay/lib/services/lockService/RedisLockStrategy.ts
```diff
@@ -76,6 +79,20 @@ export class RedisLockStrategy implements LockStrategy {
...
+        // Periodically verify queue membership.
+        const now = process.hrtime.bigint();
+        if (Number(now - lastMembershipCheckAt) / 1e6 >= this.membershipCheckIntervalMs) {
+          lastMembershipCheckAt = now;
+          const position = await this.redisClient.lPos(queueKey, sessionKey);
+          if (position === null) {
+            await this.redisClient.lPush(queueKey, sessionKey);
+            this.lockMetricsService.recordQueueRejoin(this.type);
+            if (this.logger.isLevelEnabled('debug')) {
+              this.logger.debug(`Session not in queue; rejoined: address=${address}, sessionKey=${sessionKey}`);
```
jasuwienas: @BartoszSolkaBD let's use args here

```
Session not in queue; rejoined: address=%s, sessionKey=%s
```

That way we won't have to build string inline and we won't event need this if here.

Why I think we can remove this condition:
```
this.logger.debug
```
already checks internally whether the log level is enabled, so we don’t need to duplicate that check on our side. The only case where it would make sense is if we were preparing parameters that require extra work before the call (e.g., JSON.stringify), which we’d want to avoid unless the log is actually emitted.

jasuwienas (dismissed):
lgtm, just 2 nits I think we should consider, but non-nblockers for me.

src/relay/lib/services/lockService/RedisLockStrategy.ts
BartoszSolkaBD: New arrivals go to the left (Redis head). The session that has been waiting longest sits at the right (Redis tail). We deliberately read index -1 to pick the oldest waiter as "first in FIFO queue."

To check "am I first in queue?" you'd need position === lLen - 1, which requires an extra LLEN call.

jasuwienas (approved):

docs/configuration.md
BartoszSolkaBD: removed, I've added `LOCK_QUEUE_MEMBERSHIP_CHECK_INTERVAL_MS` instead

simzzz (approved):

quiet-node (approved):

PR #5292, opened 2026-04-17
feat: Add authorisationlist on type 4 tx

### LINKED ISSUE #5208: Pectra: AuthorizationList on eth_call eth_estimateGas

bootcodes:
### Description

This PR introduces support for EIP-7702 (authorizationList) in `eth_call`, `eth_estimateGas` as part of the Pectra upgrade.
A new `authorizationListEntry` validator schema is added with custom `yParityHex` types for per-entry validation.

### Related issue(s)

Fixes #5208

### Testing Guide

1. **Unit tests** : run the relay unit suite
2.
3.

- Per-entry relay-level validation of `authorizationListEntry` fields (address, chainId, nonce, r, s, yParity) was added for consistency with how the validator schema works for other array types - the relay validates format only.
- A dedicated test verifies that when a type 4 raw transaction string is parsed by `Precheck.parseRawTransaction`, ethers deserialises `authorizationList` entries into its internal `Authorization` format (bigint chainId/nonce, nested Signature object), which differs from the relay's `AuthorizationListEntry` shape. This format difference is harmless for `eth_sendRawTransaction`.

Acceptance tests for `eth_call/eth_estimateGas` with `authorizationList` are skipped pending Mirror Node support for the field in `/api/v1/contracts/call` - tracked in
     [hiero-ledger/hiero-mirror-node#12379](https://github.com/hiero-ledger/hiero-mirror-node/issues/12379).
     Un-skip the `describe.skip` block in `sendRawTransactionExtension.spec.ts` once that is merged.

### Checklist


jasuwienas (commented):
It looks great overall, just want to confirm two things ( non-blockers only, let me know if you think we should ignore them :) ):
1. What’s our approach to error messages discussed before :) ? Are we aligning with geth or Besu? From my investigation, other providers seem to use a very similar but not the same message ("EIP-7702 transaction cannot be used to create contract", while its "[type 4 transaction...](https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5292/changes#annotation_51900131325)" in our code). I just want to confirm  what's our approach and if we are consistent with some other client for example.


2. What do you think about applying the [Codacy suggestions](https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5292/changes#diff-91e6bc4f90f7cfa8315ebb45d755d50b0a16513f2c05b01d4aee519fc5a2eb16R13)? They’re quite minor, and I’m fine ignoring them. BUT they’re also VERY easy to apply (using template literals instead of string concatenation)?

src/relay/lib/validators/types.ts
```diff
@@ -158,6 +161,18 @@ export const TYPES = {
     },
     error: 'Expected StateOverride object (currently accepting any object structure)',
   },
+  yParityHex: {
+    test: (param: string) => /^0x([0-9a-fA-F]?){1,2}$/.test(param),
+    error: 'Expected 0x-prefixed yParity value (0 for even, 1 for odd)',
+  },
+  authorizationListEntry: {
+    test: (param: any) => {
+      if (Object.prototype.toString.call(param) !== '[object Object]') return false;
```
jasuwienas: @bootcodes In other tests included here we used :
```
typeof param === 'object'
```
instead of 
```
Object.prototype.toString.call(param) !== '[object Object]'
```
shouldnt we stick to that in order to be more consistent?

bootcodes: We actually have `Object.prototype.toString.call(param) !== '[object Object]'` and `typeof param === 'object'` both present in the file ..

jasuwienas (approved):
LGTM!

src/relay/lib/validators/types.ts
jasuwienas: Right, nice catch. So maybe it was a good moment to clean that up (we're still waiting with pectra release I guess).

jasuwienas (approved):
LGTM!

BartoszSolkaBD (approved):

PR #5312, opened 2026-04-21
test: unify ws and http tests(#5223)

### LINKED ISSUE #5223: Refactor: Move duplicate tests under same unified test case suite for HTTP and WS RPC methods
### Problem

Acceptance tests were initially written under `tests/server/acceptance `and included most `eth_*` methods. 
After [#5217 ](https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5217) is merged, both HTTP and WS RPC methods are enabled, therefore to strengthen the test suite without duplicating the test cases, it is better to have an unified test suite for both HTTP and WS RPC methods.

### Solution

Migrate duplicate `eth_* `acceptance tests to `tests/protocol/`. 
Methods tested independently in both
`tests/server/acceptance/ `and `tests/ws-server/acceptance/ `with identical assertions are candidates:
`eth_blockNumber, eth_gasPrice, eth_getLogs, eth_getBalance, eth_call, eth_estimateGas, etc.
`
When migrating, delete both originals and replace with a single parameterized test in `tests/protocol/`

### Alternatives

_No response_

BartoszSolkaBD:
### Description

This PR unifies acceptance tests for ws and http protocols into common `protocol` codebase

### Related issue(s)

Fixes #5223 

### Testing Guide

1. Check if tests pass

N/A

N/A

### Checklist


tests/protocol/feeHistory.spec.ts
```diff
@@ -15,32 +33,141 @@ describe('@release @protocol-acceptance eth_feeHistory', async () => {
...
+        lastBlockAfterUpdate = (await mirrorNode.get(`/blocks?limit=1&order=desc`)).blocks[0];
+      });
+
+      after(async () => {
+        await servicesNode.updateFileContent(FEE_SCHEDULE_FILE_ID, feeScheduleContentAtStart.toString('hex'));
+        await servicesNode.updateFileContent(EXCHANGE_RATE_FILE_ID, exchangeRateContentAtStart.toString('hex'));
+        await new Promise((resolve) => setTimeout(resolve, 4000));
+      });
+
+      for (const client of ALL_PROTOCOL_CLIENTS) {
```
simzzz, line 126: This loop will create 6 sibling it() blocks - 3 names each appearing twice with no `describe(client.abel, ...)` wrapper. Mocha output will show duplicated test names with no transport label which will make failures ambigous and could break grep filters. I suggest we wrap the with a `describe(client.label, () => { … })`

tests/protocol/helpers/protocolClient.ts
simzzz, line 1: Every original `tests/ws-server/acceptance/*.spec.ts` had two transport sections: STANDARD_WEB_SOCKET and ETHERS_WS_PROVIDER. The migration only carries forward the standardWS path via WsProtocolClient. The Ethers-provider variant is silently dropped for all 15 migrated methods (blockNumber, gasPrice, getBalance, getCode, getLogs, etc.).

These were the only tests that verified a real ethers.js client can interoperate with the relay's WS server. Is this intentional?

simzzz (commented):
Good work, left two suggestions

tests/protocol/helpers/protocolClient.ts
BartoszSolkaBD: This was mentioned in the original issue comment: https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/5223#issuecomment-4281109302

I wonder if we should add 3rd transport and run just a subset of it, or what could be the approach to this matter.

simzzz: I would suggest adding single new spec, e.g. tests/protocol/ethersWsCompat.spec.ts, using a single use ethers.WebSocketProvider and calling a few methods from it without changing ALL_PROTOCOL_CLIENTS so we don't 3x all tests. Wdyt?

tests/protocol/helpers/protocolClient.ts
```diff
@@ -66,13 +67,9 @@ class WsProtocolClient implements RpcProtocolClient {
...
-    method: string,
-    params: unknown[],
-    options?: CallRawOptions,
-  ): Promise<{ result?: unknown; error?: { code: number; message: string } }> {
+  async callRaw(method: string, params: unknown, options?: CallRawOptions): Promise<RpcRawResponse> {
     if (options?.ip) {
-      return WsTestHelper.sendRequestWithIp(method, params, options.ip);
+      return WsTestHelper.sendRequestWithIp(method, Array.isArray(params) ? params : [], options.ip);
     }
     return WsTestHelper.sendRequestToStandardWebSocket(method, params);
```
quiet-node, line 74: The old WS specs ran each test through both a raw `ws` client and an `ethers.WebSocketProvider`. The migration seems to drop the ethers-WS path entirely, right?

Should we do something here so we still catch ethers-specific WS regressions or something?

.github/workflows/manual-testing.yml
```diff
@@ -192,6 +270,15 @@ jobs:
       - websocket-batch-2
       - websocket-batch-3
       - cacheservice
+      - server-config
+      - protocol
```
quiet-node: I don't seem to see a `protocol:` job defined in this workflow. Is this addition intentional?

tests/protocol/getBalance.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+      it('@release should execute "eth_getBalance" with one block behind latest block number', async () => {
+        const latestBlock = (await mirrorNode.get(`/blocks?limit=1&order=desc`)).blocks[0];
+        const res = (await client.call(METHOD_NAME, [
+          getBalanceContractAddress,
+          numberTo0x(latestBlock.number - 1),
+        ])) as string;
+        expect(res).to.eq(ethers.toQuantity(ONE_WEIBAR));
+      });
+
+      it('@release should execute "eth_getBalance" with latest block hash', async () => {
```
quiet-node, line 159: The test is named "with latest block hash" but it actually passes `numberTo0x(latestBlock.number)`, which is identical to the test right above it. 

We should fix this so the block hash branch is actually covered

tests/protocol/chainId.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  after(async () => {
+    if (global?.socketServer) {
+      expect(global.socketServer._connections).to.eq(0);
+    }
+  });
+
+  for (const client of ALL_PROTOCOL_CLIENTS) {
+    describe(client.label, () => {
+      it('should execute "eth_chainId"', async () => {
+        const res = (await client.call(METHOD_NAME, [null])) as string;
```
quiet-node: was passing `[null]` here intentional, or did you mean `[]`? The original HTTP test used `[]`. Either is fine; if `[null]` is intended, would it make sense to rename the test "with null param" so the intent is clear?

.github/workflows/acceptance.yml
quiet-node, line 1: Codecov is reporting head coverage down 25% because the new protocol jobs aren't uploading reports the way the HTTP/WS jobs do. Could we double-check that `acceptance-workflow.yml` collects coverage for the new `protocol-*` matrix entries? The `package.json` scripts already wrap with `c8`, so it looks like a workflow-side fix.

quiet-node (commented):
Great work but some findings

quiet-node (dismissed):
LGTM great work

quiet-node (approved):

simzzz (approved):

Issue #5286, opened 2026-04-17
Disable worker pool in default configuration


jasuwienas:
### Problem

We are currently enabling worker pool by default.

### Solution

WORKERS_POOL_ENABLED value should be set to false by default.

PR #5319, opened 2026-04-22
fix: unit tests are stopped mid processing with success result code (#5318)

### LINKED ISSUE #5318: Unit tests are stopped mid-processing with success result code
### Description

When we are currently starting our unit tests, through

npm run test

or

npm run build-and-test

they are terminated after 10 seconds after the cliHelper.spec.tests. THis is because we are testing graceful stop behaviour which, as a fallback terminates parent process after 10 seconds.

We can';t do that when the parent process is the test executing script.

### Steps to reproduce

Start:

npm run test

Or check the pipelines:

https://github.com/hiero-ledger/hiero-json-rpc-relay/actions/runs/24767619465/job/72465623851?pr=5316

I know unit tests SHOULD fail there . But they are terminated early.

### Additional context

_No response_

### Hedera network

other

### Version

0.77.0-SNAPSHOT

### Operating system

None

jasuwienas:
### Description

1. Making sure all of the unit tests will be executed, by preventing the test running process from terminating itself directly.
2. #5286 follow-up - our unit tests depend on the worker pool, so we should not get rid of it entirely and let them still use it, even when its disabled by default. When I tried to rebase pectra-upgrade branch onto main I've been making sure all of the tests are passing. ANd I noticed that they really SHOULDN't since I forgot about this unit tests config previously...

### Related issue(s)

Fixes #5318

### Testing Guide

1. npm run test
2. Verify process is not stopped.

N/A

N/A

### Checklist


tests/relay/test.env
```diff
@@ -32,3 +32,4 @@ MEMWATCH_ENABLED=true
 WRITE_SNAPSHOT_ON_MEMORY_LEAK=false
 USE_ASYNC_TX_PROCESSING=true
 ENABLE_TX_POOL=true
+WORKERS_POOL_ENABLED=true
```
jasuwienas: note for reviewers: There is a simulation used for unit tests not to really use the workers pool queueing mechanism. But for it to work correctly the value has to be set to true anyway...

natanasow (dismissed):

BartoszSolkaBD (approved):

simzzz (dismissed):

natanasow (approved):

simzzz (approved):

PR #5324, opened 2026-04-23
perf: parallelize address resolution in prepareTransactionArray

### LINKED ISSUE #5263: perf: parallelize address resolution in prepareTransactionArray
## Problem

[prepareTransactionArray](https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/be4abe123e4dd919d2d3e8fe73457ce273c517de/src/relay/lib/services/ethService/blockService/blockWorker.ts#L288) in `blockWorker.ts` resolves `from`/`to` addresses
  sequentially. Each transaction must complete before the next starts:

  ```ts
  for (const contractResult of contractResults) {
    [contractResult.from, contractResult.to] = await Promise.all([
      commonService.resolveEvmAddress(contractResult.from, ...),
      commonService.resolveEvmAddress(contractResult.to, ...),
    ]);
  }
```

  For a block with N transactions this creates 2N serial async operations
  (Redis roundtrips or mirror node calls).

  **Impact**

  - Directly increases worker task duration for blocks with many transactions
  - Holds a Piscina thread for longer, increasing queue depth for other tasks
  (getLogs, getBalance, getBlockReceipts)

  **Proposed Fix**

  Parallelize across all transactions using `Promise.all` like used [here](https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/be4abe123e4dd919d2d3e8fe73457ce273c517de/src/relay/lib/services/ethService/blockService/blockWorker.ts#L409):
```ts
  await Promise.all(contractResults.map(async (contractResult) => {
    [contractResult.from, contractResult.to] = await Promise.all([
      commonService.resolveEvmAddress(contractResult.from, ...),
      commonService.resolveEvmAddress(contractResult.to, ...),
    ]);
  }));
```

Gitjay11:
### Description

This PR addresses a performance bottleneck in `prepareTransactionArray` within `blockWorker.ts` by parallelizing the resolution of `from` and `to` addresses. Previously, these addresses were resolved sequentially inside a `for` loop, meaning each transaction had to finish resolving before the next one could start. 

The updated implementation maps over all transactions and uses `Promise.all` across the array to execute these asynchronous operations concurrently. This significantly reduces the total worker task duration and avoids holding up Piscina threads for blocks with a large number of transactions.

### Related issue(s)

Fixes #5263

### Testing Guide

1. Check out the branch and ensure all dependencies are installed via `npm install`.
2. Build the project using `npm run build` to confirm compilation completes without errors.
3. Run the unit testing suite with `npm run test` to verify no regressions occurred in block generation logic.
4. Run the acceptance tests (e.g. `npm run acceptancetest`) to ensure the RPC relay accurately resolves and formats block requests under the new parallel execution logic.

N/A

N/A

### Checklist


BartoszSolkaBD (approved):
Hey, great job!

While we are refactoring this function already, maybe it's worth optimising also `showDetails = false` path?

``` js
  if (!showDetails) {
    return contractResults.map((contractResult) => contractResult.hash);
  }
```

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -292,13 +292,17 @@ async function prepareTransactionArray(
   chain: string,
   commonService: CommonService,
 ): Promise<Transaction[] | string[]> {
+  await Promise.all(
+    contractResults.map(async (contractResult) => {
+      [contractResult.from, contractResult.to] = await Promise.all([
```
jasuwienas: I think we can skip this entirely when `showDetails === false`.

jasuwienas (dismissed):
Hey, I really like this change, thanks a lot for the contribution! 🙌

It got me thinking that this might actually be a good moment to introduce a slightly broader set of improvements to this method.
1. Slight concern this might fire off a lot of requests at once—maybe batching/splicing could make it a bit safer? Just something to consider.
2. Right now we’re only sorting receipts, not the transactions array itself, so removing the second loop would mess up the order if we keep things parallel. That said, maybe it’d make sense to align transaction order with the receipt order (sorted by index)? That way:
   - txArray stays consistent with the calculated root,
   - we could get rid of the second loop over contract results
 3. With the current approach, we may actually reduce the likelihood of cache hits when multiple contractResults share the same addresses. In the previous implementation, resolving addresses sequentially meant the first lookup would populate the cache, and subsequent calls could reuse it. With the new parallelized Promise.all approach, we may end up issuing multiple identical resolveEvmAddress requests at the same time before the cache is populated.
It might be worth considering deduplicating addresses before resolving them to avoid redundant parallel requests.??

src/relay/lib/services/ethService/blockService/blockWorker.ts
Gitjay11: yes, I did it

Gitjay11:
> Hey, great job!
> 
> While we are refactoring this function already, maybe it's worth optimising also `showDetails = false` path?
> 
> ```js
>   if (!showDetails) {
>     return contractResults.map((contractResult) => contractResult.hash);
>   }
> ```

@BartoszSolkaBD Thanks for the review and the suggestion. Added an early return for the showDetails = false path, it now skips address resolution entirely and returns just the hashes.

Gitjay11:
@jasuwienas Thank you for the thorough review and suggestions, I've addressed all three points and updated the changes. would appreciate your review on these changes, Thanks.

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -46,6 +46,9 @@ const mirrorNodeClient = new MirrorNodeClient(
 );
 const commonService = new CommonService(mirrorNodeClient, logger, cacheService);
 
+/** Maximum number of address resolutions to run concurrently to avoid overwhelming the mirror node. */
+const ADDRESS_RESOLUTION_BATCH_SIZE = 10;
```
BartoszSolkaBD: I think this should be an environment variable with sensible default

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -285,22 +288,80 @@ async function getRootHash(receipts: IReceiptRootHash[]): Promise<string> {
...
+  }
+
+  // Remove from toAddresses any that already appear in fromAddresses to avoid duplicate resolution
+  // (they'll be resolved with the broader default types)
+  const toOnly = new Set<string>();
+  for (const addr of toAddresses) {
+    if (!fromAddresses.has(addr)) {
+      toOnly.add(addr);
+    }
+  }
```
BartoszSolkaBD:
```suggestion
const fromAddresses = new Set<string>();
const toAddresses = new Set<string>();

for (const cr of contractResults) {
  if (cr.from) fromAddresses.add(cr.from);
  if (cr.to) toAddresses.add(cr.to);
}

const toOnly = new Set(Array.from(toAddresses).filter((a) => !fromAddresses.has(a)));
```

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -285,22 +288,80 @@ async function getRootHash(receipts: IReceiptRootHash[]): Promise<string> {
   return prepend0x(Buffer.from(trie.root()).toString('hex'));
 }
 
+async function resolveUniqueAddresses(
+  contractResults: any[],
```
BartoszSolkaBD:
```suggestion
  contractResults: Array<{ from?: string; to?: string }>,
```

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -285,22 +288,80 @@ async function getRootHash(receipts: IReceiptRootHash[]): Promise<string> {
...
+    ...[...fromAddresses].map((addr) => async () => {
+      const resolved = await commonService.resolveEvmAddress(addr, requestDetails, [constants.TYPE_ACCOUNT]);
+      resolvedMap.set(addr, resolved);
+    }),
+    // 'to' addresses use default types (CONTRACT, TOKEN, ACCOUNT)
+    ...[...toOnly].map((addr) => async () => {
+      const resolved = await commonService.resolveEvmAddress(addr, requestDetails);
+      resolvedMap.set(addr, resolved);
+    }),
+  ];
```
BartoszSolkaBD:
```suggestion
    const makeResolveTask = (addr: string, types?: string[]) => async () => {
    resolvedMap.set(addr, await commonService.resolveEvmAddress(addr, requestDetails, types));
  };

  const tasks: Array<() => Promise<void>> = [
    ...Array.from(fromAddresses, (addr) => makeResolveTask(addr, [constants.TYPE_ACCOUNT])),
    ...Array.from(toOnly, (addr) => makeResolveTask(addr)),
  ];

  for (let i = 0; i < tasks.length; i += ADDRESS_RESOLUTION_BATCH_SIZE) {
    await Promise.all(tasks.slice(i, i + ADDRESS_RESOLUTION_BATCH_SIZE).map((task) => task()));
  }
```

BartoszSolkaBD (changes requested):
Hey, thanks for speedy reply :)

Some suggestions, please let me know what you think.

bootcodes:
@Gitjay11 please add some tests as well

src/relay/lib/services/ethService/blockService/blockWorker.ts
bootcodes: You can remove `toAddresses` set entirely and populate the toOnly set directly in your first loop 
```
if (cr.to && !fromAddresses.has(cr.to)) {
    toOnly.add(cr.to);
}
```

Gitjay11:
@BartoszSolkaBD @bootcodes Thanks for the review. I've addressed all your suggestions.

Would appreciate another look when you get a chance, thanks.

src/relay/lib/services/ethService/blockService/blockWorker.ts
Gitjay11: i did it now, thanks

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -46,6 +52,9 @@ const mirrorNodeClient = new MirrorNodeClient(
 );
 const commonService = new CommonService(mirrorNodeClient, logger, cacheService);
 
+/** Maximum number of address resolutions to run concurrently to avoid overwhelming the mirror node. */
+const ADDRESS_RESOLUTION_BATCH_SIZE: number = ConfigService.get('ADDRESS_RESOLUTION_BATCH_SIZE') ?? 10;
```
BartoszSolkaBD: the default value is already set in globalConfig, I think we can safely remove `?? 10`

BartoszSolkaBD (changes requested):
Looks great!

please also add an entry into `configuration.md` with the newly added variable

src/relay/lib/services/ethService/blockService/blockWorker.ts
Gitjay11: okay, I did it

BartoszSolkaBD (approved):

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -285,22 +298,71 @@ async function getRootHash(receipts: IReceiptRootHash[]): Promise<string> {
...
+  const fromAddresses = new Set<string>();
+  const toAddresses = new Set<string>();
+
+  for (const cr of contractResults) {
+    if (cr.from) fromAddresses.add(cr.from);
+    if (cr.to) toAddresses.add(cr.to);
+  }
+
+  const toOnly = new Set(Array.from(toAddresses).filter((a) => !fromAddresses.has(a)));
+
```
bootcodes:
```suggestion
  const toOnly = new Set<string>();
  
  for (const cr of contractResults) {
    if (cr.from) fromAddresses.add(cr.from);
    if (cr.to && !fromAddresses.has(cr.to)) {
        toOnly.add(cr.to);
    }
  }
```

```

Comment (Gitjay11) [reply src/relay/lib/services/ethService/blockService/blockWorker.ts]: I did it

**Comment by Gitjay11** [comment]
@bootcodes the change is done now.

File: src/relay/lib/services/ethService/blockService/blockWorker.ts [hunk 8]
```diff
@@ -286,22 +298,71 @@ async function getRootHash(receipts: IReceiptRootHash[]): Promise<string> {
...
+): Promise<Map<string, string>> {
+  const fromAddresses = new Set<string>();
+  const toOnly = new Set<string>();
+
+  for (const cr of contractResults) {
+    if (cr.from) fromAddresses.add(cr.from);
+    if (cr.to && !fromAddresses.has(cr.to)) {
+      toOnly.add(cr.to);
+    }
+  }
```
BartoszSolkaBD, line 319: This implementation is order-dependent and can produce incorrect results.

In a sequence of records like:
```
[
 {from: "a", to: "b"},
 {from: "b", to: "c"},
 {from: "c", to: "d"},
 {from: "d", to: "e"}
]
```

each to address is evaluated before it appears as a from in a later record.

As a result, "b", "c", and "d" are incorrectly classified as toOnly, even though they are present in from.
The correct result should only include "e".

BartoszSolkaBD (changes requested):
Hey, I think there is a bug in the `toOnly` logic. Please verify, add test cases and make necessary changes to make it work as intended

src/relay/lib/services/ethService/blockService/blockWorker.ts
bootcodes: you're right ! Nice catch 
Even if changing the code above : 
```
for (const cr of c) {
    if (cr.from && !toAddresses.has(cr.from)) {
      fromAddresses.add(cr.from);
    }
    if (cr.to && !fromAddresses.has(cr.to)) {
      toAddresses.add(cr.to);
      }
  }
```

This changes the outcome to : 
`fromAddresses = {a}, toAddresses = {b, c, d, e}`

Whereas we expect : `fromAddresses = {a, b, c, d}, toAddresses (toOnly) = {e}`

If resolveEvmAddress uses the type hint at all - then it's a correctness bug on my end. Otherwise it may stand as final resolved value simply fetches the value from the map. If it doesn't matter we can implement that and remove following toOnly Set allocation entirely.

The initial implementation used : `new Set(Array.from(toAddresses).filter((a) => !fromAddresses.has(a)))`;  Array.from + filter allocates an intermediate array of at most |toAddresses| strings. 
Maybe settle on : 
```
const toOnly = new Set<string>();
  for (const addr of toAddresses) {
    if (!fromAddresses.has(addr)) toOnly.add(addr);
  }
```

that uses less memory here. Wdyt ?

Gitjay11: yeah, i did the changes

BartoszSolkaBD (approved):

tgntr:
Hello @Gitjay11, thank you for your contribution! #5263 will be resolved with #5499

PR #5334, opened 2026-04-24
feat: add retry mechanism when operator balance is 0

### LINKED ISSUE #5317: Add retry mechanism when operator balance is 0
### Problem

As part of an initiative in the Solo project, we are aiming to reduce the time for creating a deployment with all components of a Hiero network. One of the approaches we are using is deploying all components in parallel. However we noticed that if the Relay is deployed after the Mirror Node is operational, and before the Consensus Node is started, it crashes with the following error, which kicks-off a crash backoff loop, which can make the whole deployment process slower:

```
{"level":20,"time":1776848542320,"pid":1,"hostname":"relay-1-6c6b4d9db6-wp59z","msg":"An accepted error occurred while communicating with the mirror node server: method=GET, path=accounts/0.0.2?transactions=false, status=404}"}
{"level":20,"time":1776848542320,"pid":1,"hostname":"relay-1-6c6b4d9db6-wp59z","msg":"Unable to find account 0.0.2 in block null (latest), returning 0x0 balance"}
[09:02:22.320] FATAL (rpc-server/1 on relay-1-6c6b4d9db6-wp59z): Operator account '0.0.2' has no balance
    err: {
      "type": "Error",
      "message": "Operator account '0.0.2' has no balance",
      "stack":
          Error: Operator account '0.0.2' has no balance
              at Relay.<anonymous> (/home/node/app/packages/relay/dist/lib/relay.js:286:23)
              at fulfilled (/home/node/app/packages/relay/dist/lib/relay.js:6:58)
    }
``` 

### Solution

Handle this error without crashing the Relay node process and add a retry mechanism. Ideally, the parameters of the retry, such as maximum attempts, delay, and backoff, should be configurable via Helm chart values.

natanasow:
### Description

As part of an initiative in the Solo project, we are aiming to reduce the time for creating a deployment with all components of a Hiero network. One of the approaches we are using is deploying all components in parallel. However we noticed that if the Relay is deployed after the Mirror Node is operational, and before the Consensus Node is started, it crashes with the following error, which kicks-off a crash backoff loop, which can make the whole deployment process slower:

```
{"level":20,"time":1776848542320,"pid":1,"hostname":"relay-1-6c6b4d9db6-wp59z","msg":"An accepted error occurred while communicating with the mirror node server: method=GET, path=accounts/0.0.2?transactions=false, status=404}"}
{"level":20,"time":1776848542320,"pid":1,"hostname":"relay-1-6c6b4d9db6-wp59z","msg":"Unable to find account 0.0.2 in block null (latest), returning 0x0 balance"}
[09:02:22.320] FATAL (rpc-server/1 on relay-1-6c6b4d9db6-wp59z): Operator account '0.0.2' has no balance
    err: {
      "type": "Error",
      "message": "Operator account '0.0.2' has no balance",
      "stack":
          Error: Operator account '0.0.2' has no balance
              at Relay.<anonymous> (/home/node/app/packages/relay/dist/lib/relay.js:286:23)
              at fulfilled (/home/node/app/packages/relay/dist/lib/relay.js:6:58)
    }
``` 

### Solution

Handle this error without crashing the Relay node process and add a retry mechanism. Ideally, the parameters of the retry, such as maximum attempts, delay, and backoff, should be configurable via Helm chart values.

### Related issue(s)

Fixes #5317 

### Testing Guide

1.
2.
3.

N/A

N/A

### Checklist


src/relay/lib/relay.ts
```diff
@@ -484,13 +484,41 @@ export class Relay {
     }
   }
 
-  private async ensureOperatorHasBalance() {
+  private async ensureOperatorHasBalance(): Promise<void> {
+    const maxAttempts = ConfigService.get('OPERATOR_BALANCE_STARTUP_MAX_ATTEMPTS') || 1;
```
BartoszSolkaBD, line 490: when `OPERATOR_BALANCE_STARTUP_MAX_ATTEMPTS=0`, this will resolve to `1`.


```suggestion
    const maxAttempts = ConfigService.get('MIRROR_NODE_STARTUP_MAX_ATTEMPTS');
    if (maxAttempts === 0) { return; }
```

tests/relay/lib/relay.spec.ts
```diff
@@ -10,11 +10,17 @@ import sinon from 'sinon';
 import { Relay } from '../../../src/relay';
 import { MirrorNodeClient } from '../../../src/relay/lib/clients/mirrorNodeClient';
 import { MirrorNodeClientError } from '../../../src/relay/lib/errors/MirrorNodeClientError';
+import { EthImpl } from '../../../src/relay/lib/eth';
```
BartoszSolkaBD: unused import

tests/relay/lib/relay.spec.ts
```diff
@@ -139,6 +145,49 @@ describe('Relay', () => {
...
+              balance: { balance: 930300564400 },
+              transactions: [],
+              links: {},
+            });
+            await expect(relay.initializeRelay()).to.not.be.rejected;
+            expect(getAccountPageLimitStub.callCount).to.equal(1);
+          });
+
+          it('should be able to get operator balance on third attempt', async function () {
+            const error = new Error('network error');
```
BartoszSolkaBD: unused variable

tests/relay/lib/relay.spec.ts
```diff
@@ -139,6 +145,49 @@ describe('Relay', () => {
...
+              account: operatorId,
+              balance: { balance: 930300564400 },
+              transactions: [],
+              links: {},
+            });
+            await expect(relay.initializeRelay()).to.not.be.rejected;
+            expect(getAccountPageLimitStub.callCount).to.equal(1);
+          });
+
+          it('should be able to get operator balance on third attempt', async function () {
```
BartoszSolkaBD, line 164: missing test for all attempts exhausted

BartoszSolkaBD (changes requested):
Hey, a few remarks below 😄

src/relay/lib/relay.ts
natanasow: Yes, because we don't want to completely skip this check like in the `waitForMirrorNode` method.

We want to check the operator balance at least once.

tests/relay/lib/relay.spec.ts
natanasow: Added.

BartoszSolkaBD (approved):

jasuwienas (dismissed):
LGTM! It serves it's purpose.

Just a thing to have in mind:
We may greatly postpone startup of our application, our healthcheck will not be ready for the whole duration of waiting for operator balance. And we don't really know what is the true reason our balance-check fail.

Things to consider:
**shouldnt we ignore this error, start an app in warmup(readonly) mode for a moment, and query for this balance asynchronously?** - I think it would be better approach.

jasuwienas:
Btw don't we want to add some backoff as well there? @natanasow (it's part of the PR description, but we can't configure it right now).

docs/configuration.md
```diff
@@ -147,6 +147,8 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
 | `OPERATOR_KEY_MAIN`                           | ""                                                                                                                                                                                                   | Operator private key used to sign transactions in hex encoded DER format. This may be either an ED22519 private key or an ECDSA private key. The private key must be associated with/used to derive `OPERATOR_ID_MAIN`.                                                          |
+| `OPERATOR_BALANCE_STARTUP_MAX_ATTEMPTS`       | "10"                                                                                                                                                                                                 | Maximum number of attempts to obtain operator balance on startup before aborting.                                                                                                                                                                                                                                                                                                                                                 |
```
Ferparishuertas: Please extend the comment, that the minimun attempt is 1, and in case of failure, relay will not startup

Ferparishuertas: Added to this the default values seem to me to high, at total time perspective

10 and 1000ms prioritizes a better balance of speed and reliability
Please lests change to 1000ms

Ferparishuertas (changes requested):
All look good. Only DOC and value changes needed

bootcodes (approved):

docs/configuration.md
```diff
@@ -147,6 +147,8 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
+| `OPERATOR_BALANCE_STARTUP_RETRY_DELAY_MS`     | "3000"                                                                                                                                                                                               | Delay in milliseconds between consecutive obtaining operator balance startup attempts.                                                                                                                                                                                                                                                                                                                                                                                                       |
```
quiet-node: Would it make sense to align these new defaults with the Helm `startupProbe` budget? Right now `10` attempts with `3000ms` delay means roughly `27s` of waiting before the last attempt, while the chart only gives the container about `20s` to become ready. In the Solo deployment case from the issue, that can still restart the pod before this retry loop finishes.

src/relay/lib/relay.ts
```diff
@@ -484,13 +484,41 @@ export class Relay {
...
+      try {
+        const balance = BigInt(await this.ethImpl.getBalance(operator, 'latest', {} as RequestDetails));
+        if (balance === BigInt(0)) {
+          throw new Error(`Operator account '${operator}' has no balance`);
+        }
+
+        this.logger.info(`Operator account '%s' has balance: %s`, operator, balance);
+        this.logger.info('Operator balance obtained (startup check passed on attempt %d/%d)', attempt, maxAttempts);
+        return;
+      } catch (error: unknown) {
```
quiet-node, line 503: Should we narrow this retry path to the expected startup condition instead of retrying every exception? `waitForMirrorNode()` already distinguishes retryable vs non-retryable failures, but this loop will currently replay unexpected internal or configuration errors as well.

Maybe something like 

```typescript
const retryable =
    error instanceof Error && error.message === `Operator account '${operator}' has no balance`;

  if (!retryable || isLastAttempt) {
    throw error;
  }
```

docs/configuration.md
natanasow: Fixed.

docs/configuration.md
natanasow: Changed the default values to 10 attempts and 1000ms.

src/relay/lib/relay.ts
natanasow: Fixed.

natanasow:
> Btw don't we want to add some backoff as well there? @natanasow (it's part of the PR description, but we can't configure it right now).

Synced with the issue opener (Ivo from the solo team) and the current implementation satisfies their needs.

Ferparishuertas (approved):
LGTM

jasuwienas (approved):
lgtm!

PR #5356, opened 2026-05-02
feat: two-lock nonce sequencing for concurrent eth_sendRawTransaction (#5349)

### LINKED ISSUE #5349: Transaction pool visibility conflicts with nonce ordering for concurrent transactions
### Description

There's a design dilemma in `sendRawTransaction`.

We want to save a transaction into the pool as early as possible so concurrent transactions from the same sender can be visible through pending `/tx-pool` APIs while later transactions are still waiting on the per-sender lock (nonce-ordering). However, the `saveTransaction` method is an asynchronous method with random response time, which essentially breaks the FIFO order of the transactions coming in and destroys the purpose of the nonce-ordering lock.

### The dilemma

When `saveTransaction` BEFORE `acquireLock`, then if two requests arrive together:

1. Tx-A has nonce `N`
2. Tx-B has nonce `N+1`
3. If both are saved to the pool before Tx-A finishes validation, Tx-A can count Tx-B
4. Tx-A may then be treated as if it should already be using nonce `N+1`, and get rejected as `NONCE_TOO_LOW` by nonce precheck.

If we move `saveTransaction` AFTER `acquireLock`, that nonce problem goes away because the lock protects the pool update. However, that also means transactions waiting behind the lock are not visible in the pool yet, so pending APIs like txpool\_ or eth_getTransactionCount(addr, "pending") can only see one transaction at a time for that sender.

### The root cause

The current pool is doing two different jobs at once:

1. External visibility: "what transactions are currently pending?"
2. Internal nonce accounting: "what nonce should this sender use next?"

Those two goals pull in opposite directions. Saving early helps visibility, but can break nonce validation. Saving late protects nonce validation, but hides lock-waiting transactions from pending views.

### Potential option (needs team discussion)

First thing we can maybe drop the "await" before `saveTransaction` to allow the transaction to be saved in the pool but in the background (non-blocking) so it doesn't affect the transaction FIFO order. Meaning TX-A can hit `saveTransaction` and immediately go acquire the lock. Then potentially we can split the pool into two states for example `queued` and `pending`, so lock-waiting transactions (already in the pool) can be visible without affecting nonce calculation that affect nonce precheck.

This would make the transactions able to be seen in the pool, while still ensuring the nonce ordering works as expected.

quiet-node:
### Description

This PR establishes the two-lock nonce sequencing design agreed with the team to fix FIFO nonce ordering breaking under concurrent `eth_sendRawTransaction` submissions from the same sender.

**Flow:**

1. Stateless precheck (callData, size, type, gasLimit, chainId, value)
2. Lock 1 (ingress, per-sender FIFO gate)

   * Read confirmed/pending transaction count summary
   * Nonce check
   * `await saveTransaction` (on failure: throw)
 
3. Lock 2 (execution gate)

   * Balance check and network stateful checks (gas price, receiver, access list)
   * CN submission
   * Pre-exec failure (e.g. `WRONG_NONCE`, `INSUFFICIENT_PAYER_BALANCE`): remove pool entry, throw
   * Post-exec failure (e.g. `CONTRACT_REVERT_EXECUTED`): leave pool entry for MN-watcher to clean up
   * Timeout / connection error: fire-and-forget MN poll for receipt
4. MN poll: removes pool entry once MN reflects the receipt; zombie cleanup after max attempts

**What's in this PR:**

* **Two-lock structure in `sendRawTransaction`**: Lock 1 (ingress nonce gate) and Lock 2 (execution gate) are now explicit
* **`await saveTransaction` in Lock 1**: replaced fire-and-forget; on save failure an infra error is thrown — keeps the pool accurate for `txpool_*` endpoints and the MN-watcher
* **Rollback split by CN failure type**: only pre-exec failures remove the pool entry directly; post-exec and ambiguous outcomes (timeout, connection error) rely on MN polling
* **`handleSubmissionError` returns `{ error, shouldPollAndCleanup }`**: separates error classification from side-effects; pre-exec cleanup happens inside; caller drives the poll flag
* **Balance check moved to Lock 2**: balance and stateful validation now happen under the execution lock
* **`eth_getTransactionCount("pending")` uses the local transaction pool**: confirmed count + local pending count are used for nonce admission and pending transaction count resolution
* **`CACHED_SENDER_TX_COUNT_TTL` config** (default 5 min): documented in `docs/configuration.md`

Fixes: #5349


src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -275,71 +300,106 @@ export class TransactionService implements ITransactionService {
...
-      // Acquire lock before async operations
-      // This ensures proper nonce ordering for transactions from the same sender
-      if (parsedTx.from) {
-        lockResult = await this.lockService.acquireLock(parsedTx.from);
+      const senderLocalNonceEntry = await this.transactionPoolService.getSenderLocalNonce(senderAddress);
+      if (senderLocalNonceEntry) {
+        // ----- Warm path: fast + without call to MN -----
+        this.precheck.nonce(parsedTx, senderLocalNonceEntry.value);
+        admittedVersion = senderLocalNonceEntry.version;
+        await this.transactionPoolService.setSenderLocalNonce(senderAddress, {
```
jasuwienas: I thought we agreed to store the local nonce once per request series (right after fetching it from the mirror node), and then calculate it for subsequent calls as:
```
cacheValue + count(pending)
```
With this approach, there shouldn’t be any need to manually increment or decrement the nonce for each transaction.

src/relay/lib/services/ethService/accountService/AccountService.ts
```diff
@@ -263,11 +264,21 @@ export class AccountService implements IAccountService {
...
+        // Warm path: cache hit — return directly without hitting MN.
+        const senderLocalNonce = await this.transactionPoolService.getSenderLocalNonce(address);
+        if (senderLocalNonce) return numberTo0x(senderLocalNonce.value);
+
+        // Cold path: compute from MN + pool, then seed the senderLocalNonce
+        const mnNonce = await this.getAccountLatestEthereumNonce(address, requestDetails);
+        const pendingCount = await this.transactionPoolService.getPendingCount(address);
+        const nextNonce = Number(mnNonce) + pendingCount;
+
+        // fire-and-forget the cache update, no need to delay the response if the cache is down or slow
```
jasuwienas: I’m not entirely convinced about the “fire-and-forget” approach here. If this particular call lags significantly, we might add the transaction to the pool and start processing the next one while still not having a cached value available.

src/relay/lib/services/ethService/accountService/AccountService.ts
```diff
@@ -263,11 +264,21 @@ export class AccountService implements IAccountService {
       // previewnet and testnet bug have a genesis blockNumber of 1 but non system account were yet to be created
       return constants.ZERO_HEX;
     } else if (this.common.blockTagIsLatestOrPending(blockNumOrTag)) {
-      const mnNonce = await this.getAccountLatestEthereumNonce(address, requestDetails);
       if (blockNumOrTag === constants.BLOCK_PENDING) {
-        return numberTo0x(Number(mnNonce) + (await this.transactionPoolService.getPendingCount(address)));
+        // Warm path: cache hit — return directly without hitting MN.
+        const senderLocalNonce = await this.transactionPoolService.getSenderLocalNonce(address);
+        if (senderLocalNonce) return numberTo0x(senderLocalNonce.value);
```
jasuwienas: When there are no transactions pending on our end, we can end the 'series' processing and fetch the value from the mirror node again, right?

This might not be strictly necessary since we assume the user submits transactions only through our relay. **However**, even in our own tests we sometimes submit transactions alternately via the SDK (raw transactions) and our relay so when there is an option to make it work I think we should, since there’s no need to wait 30 seconds for the cache to disappear every time. We only need this workaround to handle large bursts of transactions submitted all at once. Once the backlog is cleared, we **can** return to the normal processing mode.

jasuwienas (commented):
Hey! This looks really great! Nice job getting everything implemented so quickly. It also works as expected and addresses the issues we had.

I just want to clarify a couple of things, since I remember we discussed handling some parts a bit differently last time. Let me know if I’m misremembering anything or if something changed in the meantime.

src/relay/lib/services/transactionPoolService/transactionPoolService.ts
```diff
@@ -268,4 +272,69 @@ export class TransactionPoolService implements ITransactionPoolService {
...
+
+  /**
+   * Returns the cached { value, version } entry for the sender's next expected nonce,
+   * or null if the entry is absent or no cache service is configured.
+   *
+   * @param address - The sender's EVM address.
+   */
+  async getSenderLocalNonce(address: string): Promise<{ value: number; version: string } | null> {
+    if (!this.cacheService) return null;
+    const entry = await this.cacheService.getAsync(this.senderLocalNonceCacheKey(address), 'getSenderLocalNonce');
```
bootcodes:
```suggestion
    const entry = await this.cacheService.get(this.senderLocalNonceCacheKey(address), 'getSenderLocalNonce');
```

getAsync seems deprecated

src/config-service/services/globalConfig.ts
```diff
@@ -651,6 +651,11 @@ const _CONFIG = {
     required: false,
     defaultValue: 133120, // 130 KB
   },
+  SENDER_LOCAL_NONCE_TTL_MS: {
+    type: 'number',
+    required: false,
+    defaultValue: 300_000, // 5 minutes
```
bootcodes: did we settle on 5mins for the cache time ?

src/relay/lib/services/ethService/accountService/AccountService.ts
bootcodes: Agree with @jasuwienas, `sendRawTransaction` also calls `setSenderLocalNonce` which can override the fire-and-forget call here if doesn't arrives first (race condition) only if https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/40810f529b14401021849533cc34cc4c64d82b33/src/relay/lib/factories/cacheClientFactory.ts#L11 returns CacheClient as a RedisCache (possibly only in prod ?)

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -261,6 +262,30 @@ export class TransactionService implements ITransactionService {
...
+   *      b. await saveTransaction
+   *         on save failure → decrement cache (gen-matched) → throw INTERNAL_ERROR
+   *   3. release Lock 1
+   *   4. acquire Lock 2 (execution lock)
+   *   5. balance check (reuse senderAccountInfo on cold; fresh verifyAccount on warm)
+   *      validateReceiverAndGasStateful (gas + access list + receiver)
+   *      on failure → decrement cache (gen-matched) → remove pool entry → release Lock 2 → throw
+   *   6. hand off to sendRawTransactionProcessor (CN submission + outcome handling)
+   *
+   */
```
bootcodes: nit: merge both TSDoc for this function

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -275,71 +300,106 @@ export class TransactionService implements ITransactionService {
...
 
-    let lockResult: LockAcquisitionResult | undefined;
-    let networkGasPriceInWeiBars: number;
+    const senderAddress = parsedTx.from!;
+    // Two keys for Lock 1 (short: admit + save) and Lock 2 (long: CN submit)
+    const ingressLockKey = `${senderAddress}:ingress`;
+    const execLockKey = `${senderAddress}:exec`;
+
+    // ===== Lock 1: ingress lock =====
+    const ingressLockResult = await this.lockService.acquireLock(ingressLockKey);
```
bootcodes, line 356: I believe the entire 2 lock + senderLocalNonce flow should be gated behind ENABLE_NONCE_ORDERING ? as `acquireLock` can return `undefined` if ENABLE_NONCE_ORDERING is disabled

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -275,71 +300,106 @@ export class TransactionService implements ITransactionService {
...
-     * wait for all transaction processing logic to complete before returning the transaction hash.
-     */
     return await this.sendRawTransactionProcessor(
       transactionBuffer,
       parsedTx,
       networkGasPriceInWeiBars,
-      lockResult,
+      execLockResult,
+      admittedVersion,
       requestDetails,
```
bootcodes:
```
const processorPromise = this.sendRawTransactionProcessor(
      transactionBuffer,
      parsedTx,
      networkGasPriceInWeiBars,
      execLockResult,
      admittedVersion,
      requestDetails,
    );

    if (useAsyncTxProcessing) {
      void processorPromise;
      return Utils.computeTransactionHash(transactionBuffer);
    }

    return await processorPromise;
```

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -594,121 +654,121 @@ export class TransactionService implements ITransactionService {
...
+      admittedVersion,
+      requestDetails,
+    );
+
+    if (shouldPollAndCleanup) {
+      void this.pollMirrorNodeAndCleanup(parsedTx, requestDetails);
+    }
 
-    // At this point, either no error or a post-execution failure
+    if (submissionError) throw submissionError;
```
bootcodes, line 736: this never throws if `useAsyncTxProcessing` is true - maybe wrap the void call with a .catch(err) => {this.logger.....})

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -594,121 +654,121 @@ export class TransactionService implements ITransactionService {
     transactionBuffer: Buffer,
     parsedTx: EthersTransaction,
     networkGasPriceInWeiBars: number,
-    lockResult: LockAcquisitionResult | undefined,
+    execLockResult: LockAcquisitionResult | undefined,
+    admittedVersion: string,
     requestDetails: RequestDetails,
   ): Promise<string | JsonRpcError> {
-    const originalCallerAddress = parsedTx.from?.toString() || '';
+    const senderAddress = parsedTx.from?.toString() || '';
```
bootcodes: should we allow senderAddress to be empty string ?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -594,121 +654,121 @@ export class TransactionService implements ITransactionService {
...
+    const execLockKey = `${senderAddress}:exec`;
 
     this.eventEmitter.emit('eth_execution', {
       method: constants.ETH_SEND_RAW_TRANSACTION,
     });
 
     const { error } = await this.submitTransaction(
```
bootcodes, line 709: this can probably return `submittedTransactionId` and have a happy path to return early and dissociate errorHandling from successful tx submission

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -594,121 +654,121 @@ export class TransactionService implements ITransactionService {
...
     if (!error) {
-      return;
+      return { error: null, shouldPollAndCleanup: true };
     }
 
-    // Non-SDK errors are definitive failures - propagate as-is
     if (!(error instanceof SDKClientError)) {
-      throw error;
+      return { error, shouldPollAndCleanup: true };
     }
```
bootcodes:
```
if (!(error instanceof SDKClientError)) {
    return { error: error ?? null, shouldPollAndCleanup: true };
  }
```

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -594,121 +654,121 @@ export class TransactionService implements ITransactionService {
...
+      await this.transactionPoolService.removeTransaction(senderAddress, parsedTx.serialized);
+
+      return { error: preExecError, shouldPollAndCleanup: false };
     }
 
-    // Post-execution failure (e.g. CONTRACT_REVERT_EXECUTED, INVALID_ALIAS_KEY, etc.),
-    // proceed anyway: details can be obtained using the getTransactionReceipt method
-    return;
+    // Post-exec: nonce moved, leave pool entry for the MN-poll watcher
+    return { error: null, shouldPollAndCleanup: true };
```
bootcodes, line 808:
```suggestion
      const errorStatus = error.status.toString();
      let preExecError: JsonRpcError = predefined.TRANSACTION_REJECTED(errorStatus, error.message);

// TODO: should be removed in https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/4860
if (errorStatus === constants.TRANSACTION_RESULT_STATUS.WRONG_NONCE) {
this.wrongNonceMetric.labels(ConfigService.get('ENABLE_NONCE_ORDERING') ? this.lockService.getStrategyType() : "none").inc();

try {
const accountNonce = (await this.mirrorNodeClient.getAccount(parsedTx.from!, requestDetails))?.ethereum_nonce;
if (accountNonce != null && accountNonce !== parsedTx.nonce) {
          preExecError =
            parsedTx.nonce > accountNonce
              ? predefined.NONCE_TOO_HIGH(parsedTx.nonce, accountNonce)
              : predefined.NONCE_TOO_LOW(parsedTx.nonce, accountNonce);
   }
} catch (mirrorNodeError) {
this.logger.debug(mirrorNodeError, `Failed to fetch account nonce for WRONG_NONCE error handling`);
}
      // pre-execution failure would not have moved the nonce, safe to roll back and clean up here without needing to poll MN for certainty
      await this.transactionPoolService.decrementSenderLocalNonce(senderAddress, admittedVersion);
      await this.transactionPoolService.removeTransaction(senderAddress, parsedTx.serialized);
      return { error: preExecError, shouldPollAndCleanup: false };
}

    // Post-exec: nonce moved, leave pool entry for the MN-poll watcher
    return { error: null, shouldPollAndCleanup: true };
```

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -772,4 +832,28 @@ export class TransactionService implements ITransactionService {
...
+  // ===== MN poll cleanup (POC) =====
+  // Polls MN for the receipt; removes the pool entry once MN reflects, or after max attempts (zombie cleanup).
+  private async pollMirrorNodeAndCleanup(parsedTx: EthersTransaction, requestDetails: RequestDetails): Promise<void> {
+    const senderAddress = parsedTx.from?.toString() || '';
+    const txHash = parsedTx.hash!;
+    const maxAttempts = 30;
+    const intervalMs = 2000;
+
+    for (let i = 0; i < maxAttempts; i++) {
+      await new Promise((resolve) => setTimeout(resolve, intervalMs));
```
bootcodes, line 882: do we want to wait 2 seconds every time before polling a first time ?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -772,4 +832,28 @@ export class TransactionService implements ITransactionService {
 
     return { submittedTransactionId, error };
   }
+
+  // ===== MN poll cleanup (POC) =====
+  // Polls MN for the receipt; removes the pool entry once MN reflects, or after max attempts (zombie cleanup).
+  private async pollMirrorNodeAndCleanup(parsedTx: EthersTransaction, requestDetails: RequestDetails): Promise<void> {
+    const senderAddress = parsedTx.from?.toString() || '';
+    const txHash = parsedTx.hash!;
+    const maxAttempts = 30;
```
bootcodes: Move to values to ConfigService once POC marker is removed

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -772,4 +832,28 @@ export class TransactionService implements ITransactionService {
...
+        const result = await this.mirrorNodeClient.getContractResult(txHash, requestDetails);
+        if (result && result.hash) {
+          await this.transactionPoolService.removeTransaction(senderAddress, parsedTx.serialized);
+          return;
+        }
+      } catch {
+        // continue polling on transient errors
+      }
+    }
+    // zombie cleanup: max attempts reached, remove anyway
```
bootcodes, line 893: Possibly add a warning for zombie cleanup

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -772,4 +832,28 @@ export class TransactionService implements ITransactionService {
...
+    await this.transactionPoolService.removeTransaction(senderAddress, parsedTx.serialized);
```
bootcodes, line 896: removeTransaction can possibly throw here

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: I don't think it is an issue? If we get undefined here we will simply never run releaseLock for this key?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: Wasn't that always the case with async tx processing? We submit something to the CN, user never waits for the output. User already got this tx hash, so this interruption here is unnecessary. And cleanup was scheduled 🤔 .

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: Yeah, it looks strange, and it should never happen. We validate this before we submit the transaction. Having an empty string here could as well end up with a critical exception. But I'm not sure if that is the right time to address it.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: We made sure error is set in the line 718
```
    if (!error) {
      return;
      return { error: null, shouldPollAndCleanup: true };
    }
    ```

File: src/relay/lib/services/ethService/transactionService/TransactionService.ts (hunk 17 above)
Comment (jasuwienas) [reply src/relay/lib/services/ethService/transactionService/TransactionService.ts:896]: Really nice catch, yeah, this needs to be addressed.

File: docs/configuration.md [hunk 18]
```diff
@@ -106,6 +106,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
+| `SENDER_LOCAL_NONCE_TTL_MS`                            | "300000"                                                                                                                                                                                                                                                                          | How long (in ms) the relay caches a sender's next-expected nonce. While warm, `eth_sendRawTransaction` skips the Mirror Node nonce lookup and `eth_getTransactionCount("pending")` returns the cached value directly. On expiry, the next request re-reads Mirror Node and the pending pool to reseed. Lower values reduce stale-nonce risk; higher values reduce Mirror Node calls.                                                                                                                                                                                                                                                                        |
```
Ferparishuertas: The value stored should be `eth_getTransactionCount() without pending not including pending.
The name should be comprehensive, and default TTL 30seconds

CACHED_SENDER_TX_COUNT_TTL seems to me a better value, and description should change. One thing is nonce another is tx_count

src/relay/lib/services/ethService/accountService/AccountService.ts
Ferparishuertas: We should compute the value locally if ttl has not passed, being the value
x_count + pending

Why

- it introduces less read write to tx_count
- we only update the value on succesful transactions
- pending has its own lifecycle

src/relay/lib/services/ethService/accountService/AccountService.ts
Ferparishuertas: Me neither. We should enforce the value update.

For me all this logic should be hidden in transactionPoolService

src/relay/lib/types/transactionPool.ts
```diff
@@ -48,6 +48,22 @@ export interface TransactionPoolService {
    * @returns A promise that resolves to a Set of RLP hex strings.
    */
   getAllTransactions(): Promise<Set<string>>;
+
+  /**
+   * Returns the cached { value, version } entry for the sender's next expected nonce, or null if absent.
+   */
+  getSenderLocalNonce(address: string): Promise<{ value: number; version: string } | null>;
```
Ferparishuertas: Why we need Sender in the name of the function?

src/config-service/services/globalConfig.ts
Ferparishuertas: 30 seconds

src/relay/lib/services/ethService/transactionService/TransactionService.ts
bootcodes: It was always the case - but we lose observability on that if it throws silently

src/relay/lib/services/ethService/transactionService/TransactionService.ts
Ferparishuertas: Yes. Same comment narrative above

src/relay/lib/services/ethService/transactionService/TransactionService.ts
bootcodes: No that's not the problem here - this PR feature introduces the use of localNonce cache with lock mechanisms. senderLocalNonce requires Lock 1's mutual exclusion to be correct. Without the lock it might corrupt the counter on concurrent requests no ?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
bootcodes: maybe just throw predefined.INTERNAL_ERROR if senderAddress is empty could be a quick win - Agree it should never happen

src/relay/lib/services/ethService/transactionService/TransactionService.ts
bootcodes: I meant you can merge both `if(!error)` and `!(error instanceof SDKClientError) ` into one if loop as they would both evaluate to true if error === null

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -275,71 +300,106 @@ export class TransactionService implements ITransactionService {
...
+        const pendingCount = await this.transactionPoolService.getPendingCount(parsedTx.from!, 0);
+        const signerRemoteNonce = senderAccountInfo.ethereum_nonce + pendingCount;
+        this.precheck.nonce(parsedTx, signerRemoteNonce);
+        admittedVersion = randomUUID();
+        await this.transactionPoolService.setSenderLocalNonce(senderAddress, {
+          value: signerRemoteNonce + 1,
+          version: admittedVersion,
+        });
+      }
+
```
Ferparishuertas, line 374: All the previous should be encaspulated in precheck

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -275,71 +300,106 @@ export class TransactionService implements ITransactionService {
...
+      // save transaction to pool
+      try {
+        await this.transactionPoolService.saveTransaction(senderAddress, parsedTx);
+      } catch (saveError) {
+        // Roll back the increment so the caller's retry isn't blocked until TTL expires.
+        await this.transactionPoolService.decrementSenderLocalNonce(senderAddress, admittedVersion);
```
Ferparishuertas: not needed

src/relay/lib/services/ethService/transactionService/TransactionService.ts
Ferparishuertas: And all this should be encapsulated in precheck

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: Ah,that's goot point, right.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -772,4 +832,28 @@ export class TransactionService implements ITransactionService {
...
+    const txHash = parsedTx.hash!;
+    const maxAttempts = 30;
+    const intervalMs = 2000;
+
+    for (let i = 0; i < maxAttempts; i++) {
+      await new Promise((resolve) => setTimeout(resolve, intervalMs));
+      try {
+        const result = await this.mirrorNodeClient.getContractResult(txHash, requestDetails);
+        if (result && result.hash) {
+          await this.transactionPoolService.removeTransaction(senderAddress, parsedTx.serialized);
```
Ferparishuertas: We need to increment the tx_count

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -594,121 +654,121 @@ export class TransactionService implements ITransactionService {
...
+      return { error, shouldPollAndCleanup: true };
     }
 
-    // Check if this is a pre-execution failure
     const preExecutionFailures: string[] = ConfigService.get('HEDERA_SPECIFIC_REVERT_STATUSES');
     if (preExecutionFailures.includes(error.status.toString())) {
-      // WRONG_NONCE requires special handling to determine if nonce is too high or too low
+      let preExecError: JsonRpcError;
+
       // TODO: should be removed in https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/4860
```
Ferparishuertas: We need to clean this comment and close this issue

Ferparishuertas (commented):
Severall commments

I think transactionService should be made more comprehensive

src/relay/lib/services/ethService/transactionService/TransactionService.ts
Ferparishuertas: Not at all,, We need to change this logic, to something wirh backoff

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: Extended details are already logged  in error handler which returned this error, but I've included additional (non-detail) not to lose the informtation entirely,right.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: Second thoughts: this method does not throw anything, it's entirely wrapped in try catch condition.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: As for now standard behaviour for this case across whole sendTransactionProcess is usae og tx.from!,  I used this too.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: You're righ, with nonce ordering disabled this whole flow makes no sense. I'll check it when querying for transactionCounts.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: I didn't came up with anything better than that for now, But I moved this value into the config in env variables so we can update it if we decide it makes sense without changing the code.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: I  went with an error.... It can reuslt in permanently (for the cache duration) broken nonce calculations (for a given account).

src/config-service/services/globalConfig.ts
```diff
@@ -661,6 +661,21 @@ const _CONFIG = {
...
   },
+  SEND_RAW_TRANSACTION_POLLING_INTERVAL_MS: {
+    type: 'number',
+    required: false,
+    defaultValue: 2000, // 2s
+  },
+  SEND_RAW_TRANSACTION_POLLING_MAX_ATTEMPTS: {
+    type: 'number',
+    required: false,
+    defaultValue: 30, // 30 attempts
```
Ferparishuertas: i suggest 10. 30 seems to much having 2 seconds interval

src/relay/lib/services/ethService/accountService/AccountService.ts
```diff
@@ -251,6 +252,33 @@ export class AccountService implements IAccountService {
     return { balanceFound, weibars };
   }
 
+  /**
+   * Get transaction counts associated with an account.
+   *
+   * @param {string} address The account address
+   * @param {RequestDetails} requestDetails The request details for logging and tracking
+   */
+  public async getTransactionCountSummary(
```
Ferparishuertas: private?

src/relay/lib/services/ethService/accountService/AccountService.ts
```diff
@@ -271,11 +299,11 @@ export class AccountService implements IAccountService {
...
       return constants.ZERO_HEX;
     } else if (this.common.blockTagIsLatestOrPending(blockNumOrTag)) {
-      const mnNonce = await this.getAccountLatestEthereumNonce(address, requestDetails);
       if (blockNumOrTag === constants.BLOCK_PENDING) {
-        return numberTo0x(Number(mnNonce) + (await this.transactionPoolService.getPendingCount(address)));
+        const { confirmedCount, pendingCount } = await this.getTransactionCountSummary(address, requestDetails);
+        return numberTo0x(confirmedCount + pendingCount);
       }
-      return mnNonce;
+      return await this.getAccountLatestEthereumNonce(address, requestDetails);
```
Ferparishuertas: if latest, shouldng we answer from cache if availabl, use getTransactionCountSummary, but only de cached value, not pending?

src/relay/lib/services/ethService/accountService/AccountService.ts
Ferparishuertas: Why summary? isnt there a better name?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -262,7 +264,28 @@ export class TransactionService implements ITransactionService {
...
+   *         cold  → verifyAccount (MN) + getPendingCount → nonce check
+   *                 senderAccountInfo captured for Lock 2 reuse
+   *                 cache := { value: signerNonce + 1, version: randomUUID() }
+   *                 tx.admittedVersion := cache.version
+   *      b. await saveTransaction
+   *         on save failure → decrement cache (gen-matched) → throw INTERNAL_ERROR
+   *   3. release Lock 1
+   *   4. acquire Lock 2 (execution lock)
+   *   5. balance check (reuse senderAccountInfo on cold; fresh verifyAccount on warm)
+   *      validateReceiverAndGasStateful (gas + access list + receiver)
```
Ferparishuertas: Can we move gas and accesslist to first step?
they are statess

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -275,73 +298,141 @@ export class TransactionService implements ITransactionService {
...
         await this.common.getGasPriceInWeibars(requestDetails),
       );
-      if (this.logger.isLevelEnabled('debug')) {
-        this.logger.debug(
-          `Transaction undergoing account and network (stateful) prechecks: transaction=%s`,
-          JSON.stringify(parsedTx),
-        );
-      }
-      await this.precheck.validateAccountAndNetworkStateful(parsedTx, networkGasPriceInWeiBars, requestDetails);
+      await this.precheck.validateReceiverAndGasStateful(parsedTx, networkGasPriceInWeiBars, requestDetails);
```
Ferparishuertas, line 434: I suggest to remove the Receiver request, and move all prehecks to lock 1

Ferparishuertas: its very low percentage of use cases versus the value

src/relay/lib/services/ethService/accountService/AccountService.ts
jasuwienas: It has to be public so I can use it in TransactionService.

I've renamed it to getTransactionCounts. Will that be good enough? If you have something better in mind let me know, please!

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: I'm not sure how to do it. Gas price verificiation can;t be moved to lock 1 because it requires us to make call for the networkGasPriceInWeiBars value first. Even if we remove receiver validation here we will end up with gas, balance and account checks anyway.

Ferparishuertas (dismissed):

src/relay/lib/services/transactionPoolService/LocalPendingTransactionStorage.ts
```diff
@@ -112,4 +129,19 @@ export class LocalPendingTransactionStorage implements PendingTransactionStorage
...
+   * or if no cache service is configured.
+   *
+   * Notes:
+   * - This cache does NOT track the evolving expected nonce; it only stores the initial baseline.
+   * - Callers should derive subsequent expected nonces relative to this value.
+   *
+   * @param _address - The sender's EVM address.
+   */
+  async getConfirmedCount(_address: string): Promise<number | null> {
+    return new Promise((resolve) => resolve(null));
```
bootcodes: @jasuwienas this seems to always return null - is that correct ?

```
await storage.addToListAndSetConfirmedCount(testAddress1, testRlp1, 5);
const result = await storage.getConfirmedCount(testAddress1);
// expected : 5 but returns null
```

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -262,7 +264,28 @@ export class TransactionService implements ITransactionService {
   }
 
   /**
-   * Sends a raw transaction
+   * Sends a row transaction: 2-lock sendRawTransaction
+   *
+   * FLOW:
```
bootcodes, line 269: this needs to be revised and updated accordingly

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -275,73 +298,141 @@ export class TransactionService implements ITransactionService {
...
+      const { confirmedCount, pendingCount, mirrorNodeArtifact } = await this.accountService.getTransactionCounts(
+        senderAddress,
+        requestDetails,
       );
+      verifiedBalance = mirrorNodeArtifact?.balance;
+
+      this.precheck.nonce(parsedTx, confirmedCount + pendingCount);
+
+      // save transaction to pool
+      try {
```
bootcodes: Why having a nested try/catch here and not one outer try/catch/finally ? Currently if `saveTransaction` throws - it's just propagated upwards to the JSON-RPC handler middleware

src/relay/lib/services/transactionPoolService/transactionPoolService.ts
```diff
@@ -137,7 +137,7 @@ export class TransactionPoolService implements ITransactionPoolService {
     const rlpHex = tx.serialized;
 
     try {
-      await this.storage.addToList(addressLowerCased, rlpHex);
+      await this.storage.addToListAndSetConfirmedCount(addressLowerCased, rlpHex, tx.nonce);
```
bootcodes: there is a semantic error here.  Most clients call `eth_getTransactionCount(addr, "pending")` which returns confirmedCount + pendingCount so `tx.nonce = confirmedCount + pendingCount`.
The parameter accepted by `addToListAndSetConfirmedCount` is documented expected to be the **confirmed count**.

bootcodes (changes requested):

bootcodes:
@jasuwienas I got this potential bug picked up by claude :
This **only** happens when CACHED_SENDER_TX_COUNT_TTL < PENDING_TRANSACTION_STORAGE_TTL. Why **default config it does not happen**

```
Concrete example of the real failure

  Config:
  CACHED_SENDER_TX_COUNT_TTL   = 10s   # operator wants fresh MN data
  PENDING_TRANSACTION_STORAGE_TTL = 60s   # keep pending txs alive longer

  State at T=0: ethereum_nonce=5 on MN. Three txs (nonces 5, 6, 7) in the pool — last added 12 seconds ago.

  At T=12s confirmedCountKey has expired (10s TTL). addressKey still alive (48s left, 3 items).

  Nonce 8 arrives:

  AccountService.ts:262  getConfirmedCount(address)  → null   (expired)
  AccountService.ts:263  getPendingCount(address)    → 3      (addressKey alive)

  AccountService.ts:266  confirmedCount == null → falls through to MN
  AccountService.ts:276  accountData.ethereum_nonce  → 5
                         returns { confirmedCount: 5, pendingCount: 3 }

  TransactionService.ts:366  precheck.nonce(parsedTx, 5 + 3)
  precheck.ts:117            if (8 > 8) → false → PASSES

  TransactionService.ts:370  saveTransaction(senderAddress, parsedTx)
  transactionPoolService.ts:140  addToListAndSetConfirmedCount(addr, rlp, tx.nonce)
                                                                 ^^^^^^^^
                                                                 tx.nonce = 8, not confirmedCount = 5

  RedisPendingTransactionStorage.ts:78  SET confirmedCountKey 8 NX
                                        key doesn't exist (expired) → WRITES 8
                                        correct value would be 5

  Nonce 9 arrives:

  AccountService.ts:262  getConfirmedCount(address)  → 8   (poisoned)
  AccountService.ts:263  getPendingCount(address)    → 4

  AccountService.ts:266  confirmedCount != null → returns immediately, never queries MN
                         returns { confirmedCount: 8, pendingCount: 4 }

  TransactionService.ts:366  precheck.nonce(parsedTx, 8 + 4)
  precheck.ts:117            if (12 > 9) → true → NONCE_TOO_LOW thrown
  ```

File: src/relay/lib/services/transactionPoolService/transactionPoolService.ts (hunk 32 above)
Comment (jasuwienas) [reply src/relay/lib/services/transactionPoolService/transactionPoolService.ts]: We are setting the value when it's missing only (so pending pool should be empty). Or refreshing its cache when it's already set. The idea is to hold the confirmed count there, so name of the param is ok. Perhaps method should be renamed to addToListAndSetOrRefreshConfirmedCount, but I'm really not sure about that. WDYT @bootcodes  ?

File: src/relay/lib/services/transactionPoolService/LocalPendingTransactionStorage.ts (hunk 29 above)
Comment (jasuwienas) [reply src/relay/lib/services/transactionPoolService/LocalPendingTransactionStorage.ts]: no, it should return 5.

File: src/relay/lib/services/ethService/transactionService/TransactionService.ts (hunk 31 above)
Comment (jasuwienas) [reply src/relay/lib/services/ethService/transactionService/TransactionService.ts]: (It actually updates the error message as well and throws an error mapped to `predefined.*` instead of the originally thrown Redis-based errors.)

I agree that the error handling could be much simpler here, for example, by throwing dedicated errors at each level and handling them all in a single method.  But maybe we can review/refactor that separately later?
I didn’t intend to refactor this part and just left it as it was.

File: src/relay/lib/services/ethService/transactionService/TransactionService.ts [hunk 33]
```diff
@@ -262,7 +264,28 @@ export class TransactionService implements ITransactionService {
   }
 
   /**
-   * Sends a raw transaction
+   * Sends a row transaction: 2-lock sendRawTransaction
```
quiet-node: typo *row? Should be *raw, right?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -275,73 +298,141 @@ export class TransactionService implements ITransactionService {
...
-      await this.precheck.validateAccountAndNetworkStateful(parsedTx, networkGasPriceInWeiBars, requestDetails);
+      await this.precheck.validateReceiverAndGasStateful(parsedTx, networkGasPriceInWeiBars, requestDetails);
     } catch (error) {
-      // Release lock on any error during validation or prechecks
-      if (lockResult) {
-        await this.lockService.releaseLock(parsedTx.from!, lockResult.sessionKey, lockResult.acquiredAt);
+      await this.transactionPoolService.removeTransaction(senderAddress, parsedTx.serialized);
+      if (execLockResult) {
+        await this.lockService.releaseLock(execLockKey, execLockResult.sessionKey, execLockResult.acquiredAt);
       }
```
quiet-node: This could be an issue in case removeTransaction throws (e.g. Redis connection failure, etc.) `releaseLock` will be skipped and lock 2 is now held forever for this sender and all subsequent txs will be blocked forever.

Let's move this guy in a `finally` block so a pool failure wouldn't swallow the original error. Also it'd make it symetric and consistent with Lock 1.

src/relay/lib/services/transactionPoolService/LocalPendingTransactionStorage.ts
quiet-node, line 110: Would it make sense to also call `this.confirmedCount.clear()` inside `removeAll()`? The new confirmedCount map isn't cleared here, so after a pool reset stale baseline counts survive. The Redis implementation handles this correctly via its key pattern delete, but the local in-memory one doesn't.

src/relay/lib/services/transactionPoolService/RedisPendingTransactionStorage.ts
```diff
@@ -76,9 +89,26 @@ export class RedisPendingTransactionStorage implements PendingTransactionStorage
...
+   * @param rlpHex - The RLP-encoded transaction as a hex string.
+   */
+  async removeFromListAndIncrementConfirmedCount(address: string, rlpHex: string): Promise<void> {
+    await this.redisClient
+      .multi()
+      .sRem(this.keyFor(address), rlpHex)
+      .sRem(this.globalPendingTxsKey, rlpHex)
+      .incr(this.keyFor(address, this.confirmedTxCountKey)) // If the key does not exist, it will be set to 1
+      .exec();
+  }
```
quiet-node, line 127: Let's make sure the TTL is refreshed on the confirmed-count key here, same as `addToListAndSetConfirmedCount` does. Without it, the key is set once and never extended. So for a burst longer than `CACHED_SENDER_TX_COUNT_TTL` (default 30s), the key expires mid-burst and the next `getTransactionCounts` call falls back to MN with a stale baseline. Should we add `.expire(confirmedCountKey, CACHED_SENDER_TX_COUNT_TTL)` to the pipeline here?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -275,73 +298,141 @@ export class TransactionService implements ITransactionService {
...
+        senderAddress,
+        requestDetails,
       );
+      verifiedBalance = mirrorNodeArtifact?.balance;
+
+      this.precheck.nonce(parsedTx, confirmedCount + pendingCount);
+
+      // save transaction to pool
+      try {
+        await this.transactionPoolService.saveTransaction(senderAddress, parsedTx);
```
quiet-node: Should the nonce check and `saveTransaction` here be gated behind an `ENABLE_NONCE_ORDERING` check? 

When the flag is false, `acquireLock` returns undefined so the ingress lock is never held, but `getTransactionCounts` + `precheck.nonce` + `saveTransaction` still run. Two concurrent callers can both read the same pool state, both pass the nonce check, and both write to the pool. Exactly the race this PR fixes for the enabled path. 

Would it make sense to return early from `admitTransaction` when `ENABLE_NONCE_ORDERING=false` and fall back to the previous flow? Or something else?

quiet-node (commented):
LG. Thanks again for taking over this! Great work! 

Left some findings.

src/relay/lib/services/transactionPoolService/transactionPoolService.ts
quiet-node: Should we pass the MN-fetched ethereum_nonce here instead of using `tx.nonce`? The two values are equal only when the tx nonce is exactly the next in sequence. If a future-nonce tx is the first in a burst (nonce=7, MN says 5), the cache gets seeded with 7, and nonces 5 and 6 start getting rejected as NONCE_TOO_LOW until the TTL expires.

src/relay/lib/services/transactionPoolService/RedisPendingTransactionStorage.ts
jasuwienas: Yes, great idea, thanks!

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: yes

src/relay/lib/services/transactionPoolService/transactionPoolService.ts
jasuwienas: Yes, I have second thoughts on that, @bootcodes thanks, nice catch. I didn't take the whole ttl implications here when answering you yesterday. Let's use the nonce from  mirror nodee only here.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: Wait, was the previous implementation actually behaving differently?

We already took the number of pending transactions into account when verifying the nonce, just like we do now. So when receiving many transactions at once,  validation result depended entirely on the order in which they were checked (which was largely random).

What I would agree with is either:

* not validating the nonce at all when nonce ordering is disabled (and letting the CN decide what makes sense and what doesn’t), or
* validating only against the recently cached value from the MN.
 
This is the safest approach. We know the MN value is the absolute minimum. What we do *not* know is the real state of currently pending transactions ( so they should not influence nonce validation when NONCE ordering is disabled).
I'll go with this solution for now, let me know what you think about that.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: Okay, done, it turned out to be a bit more complicated than that.
We want to release the lock only after the transaction has been submitted to the Consensus Node (unless something fails earlier in the process). However, in async transaction processing mode, we do NOT wait for the Consensus Node response before returning the result to the user. Because of that, I don’t think we should keep the user blocked by the lock during that time either.
So when async transaction mode is enabled, I release the lock here as well. I think that makes sense.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
bootcodes: @jasuwienas is this FLOW up to date IYO ?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
bootcodes: @jasuwienas just to confirm - this has been handled right ?

bootcodes (approved):

src/relay/lib/services/ethService/transactionService/TransactionService.ts
quiet-node: >However, in async transaction processing mode, we do NOT wait for the Consensus Node response before returning the result to the user. Because of that, I don’t think we should keep the user blocked by the lock during that time either.

I don't think it works that way. The lock is there to gate transactions from submitting to CN, NOT from transactions to return back from users. Even if USE_ASYNC_TX_PROCESSING=true and the CN submission happened in the background, we still want the transactions to hold the lock and only release after gets back from CN, so subseuquent transactions cannot jump ahead due to race conditioning. 

If we return to users right away without waiting for the response from consensus node, then subsequent transactions will keep on getting submitted to CN out of order -> destroy the whole purpose of the lock, correct?

jasuwienas: Yes, but what would be the point of USE_ASYNC_TX_PROCESSING if every other transaction still has to wait for the previous one to be submitted before the next submission can start?
For a single transaction, it might make sense since we get the response faster. But as soon as we make another call, we would end up queued anyway.

quiet-node: As mentioned, `USE_ASYNC_TX_PROCESSING` enables users to receive the txHash immediately, following the standard Ethereum flow. Users can then poll `eth_getTransactionReceipt` to check the status.

The locking system, however, is separate and exists for transaction processors. Even when `USE_ASYNC_TX_PROCESSING=true`, the system calls `sendRawTransactionProcessorPromise` in the background and returns the txHash right away, but transactions must still go through the lock. The lock should only be released after the transaction is successfully submitted to the CN, ensuring subsequent transactions do not arrive out of order.

**Example scenario:**

- At t=0s, senderA submits tx8, tx9, and tx10.  
- At t=3s, tx8 acquires the lock and finishes `prepareExecution`. 
       - With `USE_ASYNC_TX_PROCESSING=true`, it triggers `sendRawTransactionProcessorPromise` in the background. tx8 must continue holding the lock until it completes at the CN (this should not depend on the flag). tx9 and tx10 wait. 
       - Again with `USE_ASYNC_TX_PROCESSING=true`, the txHash is returned to the user immediately.  
- At t=5s, tx8 finishes at the CN and releases the lock, allowing tx9 and tx10 to proceed.

If tx8 released the lock immediately at t=3s just because `USE_ASYNC_TX_PROCESSING=true` and without waiting for confirmation from CN, then tx9 and tx10 could jump lock 2 and reach the CN before tx8 finishes, causing `WRONG_NONCE` errors and defeating the purpose of the locking system.

Let me know what you think

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -586,129 +691,124 @@ export class TransactionService implements ITransactionService {
...
-    await this.handleSubmissionError(error, parsedTx, requestDetails);
+    const { error: submissionError, shouldPollAndCleanup } = await this.handleSubmissionError(
+      error,
+      parsedTx,
+      senderAddress,
+      requestDetails,
+    );
+
+    if (shouldPollAndCleanup) {
+      if (ConfigService.get('ENABLE_NONCE_ORDERING') && !ConfigService.get('USE_ASYNC_TX_PROCESSING')) {
```
quiet-node: Why do we care about ENABLE_NONCE_ORDERING and USE_ASYNC_TX_PROCESSING to pollMirrorNodeAndCleanup? Should we just care about ENABLE_TX_POOL instead because the whole idea of pollMirrorNodeAndCleanup is to remove the trasnaction from pending pool, right?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: Yes, I think this issue no longer applies. When nonce ordering is disabled we will compare the tx nonce to the value we got from MN (or cached) with no pending pool appended... (and we only check min)

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: ok,I'll  release lock only when the transaction is submitted, recheck please.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: This polling mechanism makes sense only when nonce_ordering is enabled. We don't need to track what happens with the transaction on the mirror node side otherwise. The sooner we remove it from the pool the better (this is current behaviour from main).

Since we agreed here https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5356#discussion_r3232219196 that in async tx processing first tx from an address can block another ones at the submission step the second condition can be removed (!asyncTxProcessing).

quiet-node: I see it's updated to `if (ConfigService.get('ENABLE_NONCE_ORDERING'))`, but I'm still a bit confused.

Since the entire purpose of `pollMirrorNodeAndCleanup()` is to remove transactions from the pool, why do we consider `ENABLE_NONCE_ORDERING` when we are ignoring `ENABLE_TX_POOL`? If `ENABLE_NONCE_ORDERING` is true but `ENABLE_TX_POOL` is false, there are no transactions in the pool to remove, which makes the poll for the Mirror Node unnecessary and wasteful. However, given your point that "This polling mechanism makes sense only when nonce_ordering is enabled. We don't need to track what happens with the transaction on the mirror node side otherwise," I propose updating it to be ``if (ConfigService.get('ENABLE_NONCE_ORDERING') && !ConfigService.get('ENABLE_TX_POOL'))`?

jasuwienas (dismissed):
Since @quiet-node created this PR he can't approve (please leave a note here that you're fine with the current version of the changes)

quiet-node (commented):
LGTM

simzzz (dismissed):

simzzz (dismissed):

Ferparishuertas (dismissed):

simzzz (approved):

jasuwienas (approved):

PR #5357, opened 2026-05-02
refactor: clean up getBalance archival balance resolution logic

### LINKED ISSUE #5193: Refactor: clean up `getBalance` archival balance resolution logic
### Problem

**Context**

During [#5146](https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5146) (migration of `eth_getBalance` to a worker thread), several pre-existing issues in the balance resolution logic were identified but intentionally left unaddressed to keep the migration PR focused.
This issue tracks the follow-up cleanup.

**Problems**

 **1. Opaque control flow in archival balance resolution**

The method that resolves whether a block identifier refers to the chain tip or a historical block (extractBlockNumberAndTimestamp) is difficult to follow. The core logic is:

  ```
if block tag is pending/latest → return null (not archival)
  resolve block identifier → block number
  if block is within tip tolerance → return null (not archival)
  return balance at historical block ?? 0
```

This is buried under nested conditionals and a mutable `blockNumberOrTagOrHash` variable that is promoted to `constants.BLOCK_LATEST` inside the function - a hidden state change that is non-obvious to callers.

**2. Double blockTagIsLatestOrPending check**

`blockTagIsLatestOrPending` is called twice with the same key before and after `extractBlockNumberAndTimestamp`. The second check is necessary because the function can promote `blockNumberOrTagOrHash` to `BLOCK_LATEST` internally - but this is
invisible from the call site. A discriminated return type would make this explicit and eliminate the second check.

**3. Magic number for block tip tolerance**

The tolerance of 1 block used to determine whether a request targets the chain tip is an inline magic number with no explanation. It should be a named constant with documentation explaining why that value was chosen.

### Solution

  - Refactor `extractBlockNumberAndTimestamp` to return a discriminated union ({ isLatest: true, latestBlock } | { isLatest: false, latestBlock, blockNumberOrTagOrHash }) so callers do not need to inspect the returned `blockNumberOrTagOrHash` to
  determine routing
  - Extract the tip tolerance to `constants.LATEST_BLOCK_TOLERANCE` with a comment explaining the value
  - Restructure the nested conditionals into a flat, readable sequence matching the logic description above

### Alternatives

_No response_

### PR DESCRIPTION
**Pull request by prakharsingh-74**
This PR refactors the archival balance resolution logic in eth_getBalance to improve code readability and eliminate opaque state mutations.

**Key improvements include:**

- Refactoring AccountService.extractBlockNumberAndTimestamp to return a discriminated union (isLatest: true/false), making the control flow explicit for callers.
- Eliminating a redundant blockTagIsLatestOrPending check in accountWorker.ts by leveraging the new return type.
- Replacing the hardcoded magic number for block tip tolerance with a documented LATEST_BLOCK_TOLERANCE constant in constants.ts.
- Flattening nested conditionals to simplify the logical path for determining historical vs. latest balances.

Related issue(s)
Fixes #5193

**Testing Guide**
The changes were verified using the existing Relay test suite to ensure no regressions in balance resolution (including 'latest', 'pending', and historical block identifiers).

- Ensure the Relay is configured in a valid state (e.g., READ_ONLY=true if no operator keys are provided).
- Run the eth_getBalance spec suite:
`npx ts-mocha tests/relay/lib/eth/eth_getBalance.spec.ts --exit`
- Expected Outcome: All 27 tests pass, covering various block tags, block numbers within/outside tip tolerance, and mirrored balance lookups.

image

### CODE REVIEW

**Comment by prakharsingh-74** [comment]
@jasuwienas and @natanasow can  you please review this PR

File: src/relay/lib/services/ethService/accountService/AccountService.ts [hunk 1]
```diff
@@ -132,10 +132,12 @@ export class AccountService implements IAccountService {
   public async extractBlockNumberAndTimestamp(
     blockNumberOrTagOrHash: string,
     requestDetails: RequestDetails,
+  ): Promise<
+    | { isLatest: true; latestBlock: LatestBlockNumberTimestamp }
```
simzzz, line 136: The old annotation was not deleted hence your code doesn't build.

src/relay/lib/constants.ts
```diff
@@ -120,6 +120,7 @@ export default {
 
   // block ranges
   MAX_BLOCK_RANGE: 5,
+  LATEST_BLOCK_TOLERANCE: 1,
```
simzzz, line 123: The ticket explicitly asks for the constant "with a comment explaining why that value was chosen", could you also document this?

simzzz (changes requested):
I apologize for the late review, please address the changes I've noted. Thank you !

PR #5368, opened 2026-05-05
feat: use `blockhash` as parameter on `eth_call` against MN

### LINKED ISSUE #5031: Track MN: use `blockHash` as parameter on `eth_call`
### Problem

Currently, if we execute `eth_call` against our relay, we can receive `DEPENDENT_SERVICE_IMMATURE_RECORDS` error, which is strange to the end users.

This is possible because we're transforming `blockHash` to `blockNumber` executing MN call:
- https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/main/packages/relay/src/lib/services/ethService/contractService/ContractService.ts#L123
- https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/main/packages/relay/src/lib/services/ethService/contractService/ContractService.ts#L417

We're doing this transformation because MN POST `/contracts/call` still doesn't support `blockHash` as a parameter.

Example:
- https://testnet.mirrornode.hedera.com/api/v1/blocks/32286750
- curl using block number
```
$ curl --location 'https://testnet.mirrornode.hedera.com/api/v1/contracts/call' \
--data '{
    "block": "32286750",
    "to": "0x7ab63cef6787dd4902202e084fd619bc1de7eee6",
    "data": "0x1687fe8e",
    "estimate": false
}'
```
results in:
```
{"result":"0x000000000000000000000000000000000000000000000000000000000000160e"}
```
- curl using block hash
```
curl --location 'https://testnet.mirrornode.hedera.com/api/v1/contracts/call' \
--data '{
    "block": "0x73c2914d97b0aabb5fa44e366b1e665304b99917009be6c2ab6a28d6b301cc7b077a9818d5d2f470af06708fd17a048c",
    "to": "0x7ab63cef6787dd4902202e084fd619bc1de7eee6",
    "data": "0x1687fe8e",
    "estimate": false
}'
 ```
results in:
```
{"_status":{"messages":[{"message":"Bad Request","detail":"JSON parse error: Cannot construct instance of `org.hiero.mirror.web3.viewmodel.BlockType`, problem: Invalid block value: 0x73c2914d97b0aabb5fa44e366b1e665304b99917009be6c2ab6a28d6b301cc7b077a9818d5d2f470af06708fd17a048c","data":""}]}}
```

### Solution

Once the MN starts to support `blockHash` as a parameter, we can remove this transformation (the MN call), resulting in improved performance.

### Alternatives

_No response_

### PR DESCRIPTION
**Pull request by natanasow**
Fixes inconsistent `eth_call` behavior when using block hashes. Previously, the relay transformed `blockHash` into `blockNumber` before calling Mirror Node, since
  `/contracts/call` did not support block hashes — leading to potential `DEPENDENT_SERVICE_IMMATURE_RECORDS` errors and confusing responses.

  Mirror Node v0.154.0 now supports `blockHash` directly for `/contracts/call`. This PR removes the conversion logic and passes block hashes through unchanged, improving correctness and eliminating the extra `/api/v1/blocks/{hash}`
  lookup.

  **Related issue(s)**

  Fixes #5031

  **Testing Guide**

  1. Deploy a contract to a local or testnet environment.
  2. Retrieve a past block hash (e.g., from `eth_getBlockByNumber`).
  3. Call `eth_call` with the block hash as the `blockParam` (e.g., `{"method":"eth_call","params":[{...}, "0xabc123..."]}`).
  4. Verify the response returns the correct contract state at that block without error, and confirm no extra `/api/v1/blocks/{hash}` lookup is made to Mirror Node.

  **Note**

  Protocol Contract Service CI is expected to fail: MN v0.154.0 has an upstream bug where `POST /api/v1/contracts/call` returns HTTP 500 when `value > 0` and `to` is a contract address. This is unrelated to the blockHash feature. An
  upstream bug report will be filed against hiero-ledger/hiero-mirror-node.

### CODE REVIEW

**Comment by natanasow** [comment]
- Requires MN > `0.154.x`. 
- Blocked until https://github.com/hiero-ledger/hiero-mirror-node/issues/13495 is resolved.

**Comment by simzzz** [comment]
Self-hosted operators on MN < 0.154 will now get HTTP 400 on eth_call with a blockHash, the same error the issue describes, just from the other direction. 

Previously the relay silently converted blockHash → blockNumber, so it worked on any MN version. This is a breaking change for older deployments. We should note the MN ≥ 0.154 requirement in the docs

File: src/relay/lib/services/ethService/contractService/ContractService.ts
Comment (simzzz) [inline src/relay/lib/services/ethService/contractService/ContractService.ts]: I think we can delete this method as it's dead code now, right?

Comment (simzzz) [reply src/relay/lib/services/ethService/contractService/ContractService.ts]: Same goes for isBlockHash

File: src/relay/lib/services/ethService/contractService/ContractService.ts [hunk 1]
```diff
@@ -429,11 +420,7 @@ export class ContractService implements IContractService {
     // if blockParam is a string, could be a blockNumber or blockTag or blockHash
     if (blockParam.length > 0) {
       // if string is a blockHash, we return its corresponding blockNumber
```
simzzz: nit: comment says "if string is a blockHash, we return its corresponding blockNumber" but the code now does the opposite (returns the hash as-is). Remove or update to something like "blockHash strings are passed directly to MN".

src/relay/lib/services/ethService/contractService/ContractService.ts
```diff
@@ -399,14 +394,10 @@ export class ContractService implements IContractService {
...
    * @param {string | object | null} blockParam - The block parameter (string, object, or null)
-   * @param {RequestDetails} requestDetails - The request details for logging and tracking
    * @returns {Promise<string | null>} The extracted block number or tag, or null if not provided
    * @private
    */
-  private async extractBlockNumberOrTag(
-    blockParam: string | object | null,
-    requestDetails: RequestDetails,
-  ): Promise<string | null> {
+  private async extractBlockNumberOrTag(blockParam: string | object | null): Promise<string | null> {
```
simzzz: nit: this method is no longer async, right? we can remove the async

natanasow:
Blocked until https://github.com/hiero-ledger/solo/issues/4288 is resolved.

tests/relay/lib/eth/eth_call.spec.ts
```diff
@@ -667,12 +667,12 @@ describe('@ethCall Eth Call spec', async function () {
     });
 
     it('should return null when blockParam is null in extractBlockNumberOrTag', async function () {
-      const result = await contractService['extractBlockNumberOrTag'](null, requestDetails);
+      const result = await contractService['extractBlockNumberOrTag'](null);
       expect(result).to.be.null;
     });
```
simzzz, line 675: In addition to the acceptance test, I think it would be good to add two unit tests for the following cases: 

1. string-form blockHash is forwarded unchanged in the MN payload - mock postContractCall, call contractService.call(callData, '0xabc...64chars', requestDetails), assert the stub was called with block: '0xabc...64chars'; 
2. object { blockHash: '0xabc...' } likewise.

wdyt?

src/relay/lib/services/ethService/contractService/ContractService.ts
natanasow: Fixed.

src/relay/lib/services/ethService/contractService/ContractService.ts
natanasow: Fixed.

src/relay/lib/services/ethService/contractService/ContractService.ts
natanasow: Fixed.

tests/relay/lib/eth/eth_call.spec.ts
natanasow: Added (didn't mock the `postContractCall` because we can test `extractBlockNumberOrTag` directly).

bootcodes (dismissed):

simzzz (dismissed):

Ferparishuertas (dismissed):
LGTM

src/relay/lib/services/ethService/contractService/ContractService.ts
quiet-node, line 1: Since this raises the hard floor to MN v0.154.0 (blockHash on `/contracts/call` only landed there), should we update the minimum Mirror Node version banner in docs/configuration.md? It currently says v0.151.0, and self-hosted operators on <0.154.0 would start getting HTTP 400 for blockHash-bound eth_calls right after picking up this change. simzzz already flagged this in the thread.

src/relay/lib/services/ethService/ethCommonService/CommonService.ts
```diff
@@ -594,18 +590,14 @@ export class CommonService implements ICommonService {
...
   };
 
-  private isBlockNumValid = (num: string): boolean => {
+  private isBlockNumValid = (num: string) => {
     return /^0[xX]([1-9A-Fa-f]+[0-9A-Fa-f]{0,13}|0)$/.test(num) && Number.MAX_SAFE_INTEGER >= Number(num);
   };
 
-  public isBlockParamValid = (tag: string | null): boolean => {
+  public isBlockParamValid = (tag: string | null) => {
     return tag == null || this.isBlockTagEarliest(tag) || this.isBlockTagFinalized(tag) || this.isBlockNumValid(tag);
```
quiet-node, line 627: Now that `isBlockHash` is gone, should we also drop the `BLOCK_HASH_REGEX` constant in src/relay/lib/constants.ts? It looks like there are no other consumers after this PR.

tests/relay/lib/eth/eth_call.spec.ts
```diff
@@ -667,18 +669,25 @@ describe('@ethCall Eth Call spec', async function () {
...
 
     it('should return null when blockParam is null in extractBlockNumberOrTag', async function () {
```
quiet-node: Tiny nit: now that `extractBlockNumberOrTag` is synchronous, would it make sense to drop `async function`/`await` in these four cases?

src/relay/lib/validators/types.ts
```diff
@@ -54,13 +49,14 @@ export const TYPES = {
       }
       return (
         (/^0[xX]([1-9A-Fa-f]+[0-9A-Fa-f]{0,13}|0)$/.test(param) && Number.MAX_SAFE_INTEGER >= Number(param)) ||
-        ['earliest', 'latest', 'pending', 'finalized', 'safe'].includes(param)
+        ['earliest', 'latest', 'pending', 'finalized', 'safe'].includes(param) ||
+        new RegExp(Constants.BASE_HEX_REGEX + '{64}$').test(param)
```
quiet-node, line 58: We're adding a new accept branch to `blockParams` but I couldn't find any unit test targeting `TYPES.blockParams` in `validators.spec.ts`. Would it make sense to add a small `describe` covering 64-char hex / earliest|latest|... / hex number / negative cases? The acceptance test covers the happy path but only end-to-end.

quiet-node (commented):
some feedback

quiet-node:
Still blocked. More info at https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/5031#issuecomment-4733935234

quiet-node:
Still blocked. More info at https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/5031#issuecomment-4848679031

src/relay/lib/services/ethService/contractService/ContractService.ts
```diff
@@ -399,14 +399,10 @@ export class ContractService implements IContractService {
...
    * @param {string | object | null} blockParam - The block parameter (string, object, or null)
-   * @param {RequestDetails} requestDetails - The request details for logging and tracking
    * @returns {Promise<string | null>} The extracted block number or tag, or null if not provided
    * @private
    */
-  private async extractBlockNumberOrTag(
-    blockParam: string | object | null,
-    requestDetails: RequestDetails,
-  ): Promise<string | null> {
+  private extractBlockNumberOrTag(blockParam: string | object | null): string | null {
```
ValentinVPK: This function now can also return a block hash and not only block number or tag. Let's rename it to something like `extractBlockParam` so it doesn't cause any confusion

src/relay/lib/services/ethService/contractService/ContractService.ts
ValentinVPK: Let's update the TSDoc for this function to be inline with the new functionality. Also on the `returns` statement we must change to `string | null` because it no longer returns a `Promise` and to mention that also a hash could be returned

ValentinVPK (commented):
LGTM

Two minor comments about renaming the helper function and fixing the TSDoc for it.
One other doc that should be updated is:

- In `relay.ts` startup warning still says "requires mirror node >= v0.151.0". We should change it to 154+

src/relay/lib/services/ethService/contractService/ContractService.ts
natanasow: Fixed.

src/relay/lib/services/ethService/contractService/ContractService.ts
natanasow: Fixed.

ValentinVPK (approved):

Ferparishuertas (approved):

jasuwienas (approved):

PR #5371, opened 2026-05-05
feat: async savetransaction break fifo nonce ordering for concurrent transactions in sendrawtransaction adjustments (#5349)

### LINKED ISSUE #5349: Transaction pool visibility conflicts with nonce ordering for concurrent transactions

jasuwienas:
### Description

Lock service rework - improvement proposals. I’ve adjusted the changes originally proposed by @quiet-node to better align with the solution we agreed on (based on my understanding and how I remembered things). This includes:

1. Allow multi-relay/source behavior (default) when our txPool is empty. This indicates that we have already processed all transactions from the burst. At that point, we can safely return to the normal flow and use the mirror node as the source of truth again. This change enables users to use both the SDK and our relay simultaneously without needing to wait 5 minutes between calls. Many of our tests (e.g., in the hiero-contracts and hedera-smart-contracts repositories) rely on this pattern - e.g., one transaction via the relay, another via the SDK and then relay again. Without this change, such tests would fail for up to 5 minutes.
2. Store only the initial value received from the mirror node when the transaction burst begins. This removes the need to continuously synchronize two values (tx pool size and cached value) whenever the pool changes.
3. Do not cache responses for get queries to the txCount endpoint. No signature is required in this call, and we cannot reliably determine the requester.
4. Fixed the tests.
5. Updated the default configuration: reduced the maximum expected mirror node indexing time from 5 minutes to 30 seconds. In practice, indexing should take no more than ~3 seconds, so a 10× buffer should still be safe?

### Related issue(s)

Fixes #5349

### Testing Guide

1.
3.
7.

N/A

N/A

### Checklist


src/relay/lib/services/transactionPoolService/transactionPoolService.ts
```diff
@@ -273,68 +273,58 @@ export class TransactionPoolService implements ITransactionPoolService {
...
+   * as returned by the mirror node for the first transaction in a burst; returns null if absent
+   * or if no cache service is configured.
+   *
+   * Notes:
+   * - This cache does NOT track the evolving expected nonce; it only stores the initial baseline.
+   * - Callers should derive subsequent expected nonces relative to this value.
    *
    * @param address - The sender's EVM address.
    */
   async getSenderLocalNonce(address: string): Promise<{ value: number; version: string } | null> {
```
bootcodes: nit: does SenderLocalNonce reference to sender's nonce in LocalCache ?

bootcodes:
Like the improvement @jasuwienas !

src/relay/lib/services/transactionPoolService/transactionPoolService.ts
jasuwienas: This one had to be renamed :)

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -314,24 +311,28 @@ export class TransactionService implements ITransactionService {
     let senderAccountInfo: any = null;
 
     try {
-      const senderLocalNonceEntry = await this.transactionPoolService.getSenderLocalNonce(senderAddress);
-      if (senderLocalNonceEntry) {
+      const [senderLocalNonceEntry, pendingCount] = await Promise.all([
```
Ferparishuertas, line 314: Why not calling AccountService here to get the value, and not repeat the same logic? you ask with pending and you have it

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -314,24 +311,28 @@ export class TransactionService implements ITransactionService {
...
       } else {
         // ----- Cold path: with call to MN -----
         senderAccountInfo = await this.precheck.verifyAccount(parsedTx, requestDetails);
-        const pendingCount = await this.transactionPoolService.getPendingCount(parsedTx.from!, 0);
         const signerRemoteNonce = senderAccountInfo.ethereum_nonce + pendingCount;
         this.precheck.nonce(parsedTx, signerRemoteNonce);
         admittedVersion = randomUUID();
-        await this.transactionPoolService.setSenderLocalNonce(senderAddress, {
-          value: signerRemoteNonce + 1,
+        await this.transactionPoolService.setSenderInitialNonce(senderAddress, {
```
Ferparishuertas: Isnt this part of accountService too?

src/relay/lib/services/transactionPoolService/transactionPoolService.ts
```diff
@@ -273,68 +273,58 @@ export class TransactionPoolService implements ITransactionPoolService {
...
 
   /**
-   * Returns the cache key for a sender's local nonce entry.
+   * Returns the cache key for a sender's cached initial nonce baseline entry.
    *
    * @param address - The sender's EVM address.
    */
-  private senderLocalNonceCacheKey(address: string): string {
-    return `senderLocalNonce:${address.toLowerCase()}`;
+  private senderInitialNonceCacheKey(address: string): string {
```
Ferparishuertas, line 287: Why call it cache, initial nonce?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
jasuwienas: I'll continue working on that in this PR: https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5356

PR #5399, opened 2026-05-13
feat: add env to disable MN prechecks in `eth_sendRawTransaction`

### LINKED ISSUE #5375: [Feature] Add Configuration Flag to Disable Mirror Node Prechecks in sendRawTransaction
# User Story: Add Configuration Flag to Disable Mirror Node Prechecks in `sendRawTransaction`

## Background / Context

The Hedera JSON-RPC relay currently performs several Mirror Node (MN) calls during the `eth_sendRawTransaction` flow for:

- Gas price estimation
- HBAR exchange rate retrieval
- Account existence, balance, and receiver_sig_required checks
- Latest nonce validation (and detailed error reporting for wrong nonce)

These prechecks add significant latency, especially under load or degraded MN conditions. Environments like performance testing, lab work, High tps uses or failover scenarios may wish to **trade validation for reduced latency**. (with aync processing = true, txpool deactivated etc)

---

### 1. Configuration

- Introduce a new boolean config flag: **`DISABLE_MN_PRECHECKS`**  
- __Default:__ `false`
- __Configurable via:__ environment variable or config file.

### 2. Behavior: If `DISABLE_MN_PRECHECKS=false` (default, current behavior)

- All existing Mirror Node prechecks are performed.
- No behavioral change.

### 3. Behavior: If `DISABLE_MN_PRECHECKS=true`

- **No Mirror Node calls are made during the `sendRawTransaction` precheck phase:**  (beside if aync = false, latest MN poll)
 
  - ❌ Network gas price/exchange rate call  
  - ❌ Account check (existence, balance, receiver signature flag)  
  - ❌ Nonce check  
- Only stateless ('local') checks are performed:
  - ✅ RLP parse  
  - ✅ Chain ID validation  
  - ✅ Transaction type, size, call-data size, gas limit
- After stateless checks, the transaction is submitted to Hedera as usual.
- **If the Relay backend rejects the transaction** (e.g. wrong nonce, insufficient balance), return a generic error (not a custom MN-validated message). 

### 4. Logging and Documentation

- Log a warning when `DISABLE_MN_PRECHECKS=true`:
- Add documentations, and showcase the testing speed data  as example

### 4. Speed/ Response time ( manual testing)

DISABLE_MN_PRECHECKS = true should be several order of magnitude faster in sending transactions than 0= false

natanasow:
### Description

The Hedera JSON-RPC relay currently performs several Mirror Node (MN) calls during the `eth_sendRawTransaction`. These prechecks add significant latency, especially under load or degraded MN conditions. Environments like performance testing, lab work, high TPS uses, or failover scenarios may wish to trade validation for reduced latency. (with async processing = true, txpool deactivated, etc)

### Solution

- introduce a new boolean config `DISABLE_MN_PRECHECKS` (default false)
- if `DISABLE_MN_PRECHECKS=false` - All existing Mirror Node prechecks are performed. No behavioral change.
- if `DISABLE_MN_PRECHECKS=true` - No Mirror Node calls are made during the `eth_sendRawTransaction` prechecks.

### Related issue(s)

Fixes #5375 

### Testing Guide

1.
2.
3.

N/A

N/A

### Checklist


src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -303,7 +303,10 @@ export class TransactionService implements ITransactionService {
           JSON.stringify(parsedTx),
         );
       }
-      await this.precheck.validateAccountAndNetworkStateful(parsedTx, networkGasPriceInWeiBars, requestDetails);
+
+      if (!ConfigService.get('DISABLE_MN_PRECHECKS')) {
```
simzzz: The original issue requires skipping network gas price/exchange rate call too, these still happen unconditionally. This makes the new docs line incorrect: configuration.md claims "gas price estimation, HBAR exchange rate retrieval ... pre-checks are skipped." Are they left intentionally?

simzzz: Also the ticket says  "If the Relay backend rejects the transaction (e.g. wrong nonce), return a generic error (not a custom MN-validated message)." handleSubmissionError is unchanged, so on WRONG_NONCE it still does a MN getAccount() call and returns NONCE_TOO_LOW/NONCE_TOO_HIGH.

docs/configuration.md
```diff
@@ -41,7 +41,8 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
+| `DISABLE_MN_PRECHECKS`                                 | "false"                                                                                                                                                                                                                                                                           | Flag to determine whether the MN pre-checks must be performed. If `true`, gas price estimation, HBAR exchange rate retrieval, accounts (existence, balance, and sig_required), and nonce pre-checks are skipped. Disabling the pre-checks reduces eth_sendRawTransaction response time by approximately 180 ms in local benchmarks.                                                                                                                                                                                                                                                                                                                        |
```
simzzz: nit: DISABLE_MN_PRECHECKS reads as "disable all MN prechecks," but it only disables the account/network stateful group. If gas price and exchange rate MN traffic remain, something like DISABLE_MN_ACCOUNT_PRECHECKS would be more honest.

simzzz (dismissed):
noticed a gap between the ticket and the implementation, also there's no test coverage

I would add these 2 at minimum:
  - A TransactionService unit test asserting precheck.validateAccountAndNetworkStateful is not called when the flag is true, and is called when false.
  - A ConfigService test for the new warnings mechanism (tests/config-service/src/services/configService.spec.ts already has a "should log warning" pattern to follow).

docs/configuration.md
Ferparishuertas: What are are disabling.
Mirror node check on send raw tx

DISABLE_MN_PRECHECKS_ON_TX_SENDING

src/relay/lib/services/ethService/transactionService/TransactionService.ts
Ferparishuertas: Yes we are missing all the checks,
The. objective. is No single MN call executed while sending a transaction to CN

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -287,23 +287,33 @@ export class TransactionService implements ITransactionService {
...
         this.logger.debug(
           `Transaction undergoing account and network (stateful) prechecks: transaction=%s`,
           JSON.stringify(parsedTx),
         );
       }
-      await this.precheck.validateAccountAndNetworkStateful(parsedTx, networkGasPriceInWeiBars, requestDetails);
+
+      if (!disableMnPrechecks) {
+        await this.precheck.validateAccountAndNetworkStateful(parsedTx, networkGasPriceInWeiBars, requestDetails);
+      }
```
quiet-node, line 467: Should we still run `precheck.gasPrice(parsedTx, networkGasPriceInWeiBars)` here when the flag is on? It does not touch Mirror Node — it just compares the user-supplied gas price against the fallback network price we just computed. The ticket's AC3 says stateless checks should still run including gas-price-related basics, and otherwise an under-priced tx will pay a full HAPI round trip just to learn it would be rejected, which is the opposite of the latency win we are selling

docs/configuration.md
```diff
@@ -30,105 +30,108 @@ These properties are noted below and should be custom set per deployment.
...
..._NETWORK_EXCHANGE_RATE_IN_CENTS`              | "12"                                                                                                                                                                                                                                                                              | Static HBAR-to-USD-cents exchange rate used by `eth_sendRawTransaction` when `DISABLE_MN_PRECHECKS_ON_TX_SENDING=true`. Consumed only by the HFS (large call-data) path for HBAR-limit fee estimation; ignored otherwise.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
```
quiet-node: why 12 again? can you put the reference in here so we know where 12 comes from or how it's calculated?

src/config-service/services/index.ts
```diff
@@ -80,6 +91,13 @@ export class ConfigService {
       for (const name in this.envs) {
         logger.info(LoggerService.maskUpEnv(name, this.envs[name]));
       }
+
+      // printing warnings
+      for (const warning of this.warnings) {
+        if (this.envs[warning.envName] === warning.value) {
+          logger.warn(`${warning.envName} is set to '${warning.value}'`);
+        }
+      }
```
quiet-node: What is the purpose of this warning? When an operator runs this server, they will already see the environment name and its value, so the warning seems quite obvious. I don't see the point of the warning because it just prints the variable name and its value without explaining anything else, which is a bit strange. Am i missing anything?

src/config-service/services/index.ts
```diff
@@ -56,6 +56,17 @@ export class ConfigService {
...
+   * List of predefined warnings to be shown
+   * @private
+   */
+  private readonly warnings: Array<{ envName: string; value: any }> = [
+    {
+      envName: 'DISABLE_MN_PRECHECKS_ON_TX_SENDING',
+      value: true,
+    },
+  ];
+
```
quiet-node: We also have other warnings, such as the one regarding the Mirror Node Minimum Version, correct? It prints a good, meaningful warning about Mirror Nodes, indicating that if you use this configuration, it should be at a certain version of MN when a variable is used. 

Now that we are adding these warnings, which feels a bit odd to me because they don't show any extra reasoning. Warnings are generally read as something needing attention or provide an action to take to resolve the warnings, but in this case, they just print the variable name and its value. The operator can be quite confused seeing them

docs/configuration.md
```diff
@@ -30,105 +30,108 @@ These properties are noted below and should be custom set per deployment.
...
..._NETWORK_GAS_PRICE_IN_TINYBARS`               | "71"                                                                                                                                                                                                                                                                              | Static network gas price (in tinybars) used by `eth_sendRawTransaction` when `DISABLE_MN_PRECHECKS_ON_TX_SENDING=true`. Used to compute the Hedera `maxTransactionFee` for the wrapped ethereum transaction. Choose a value at or above the live network gas price; setting it too low may cause `INSUFFICIENT_TX_FEE` rejections.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
```
quiet-node: Same for this why 71? if any resource lets add them

quiet-node (commented):
some feedback

src/relay/lib/services/ethService/transactionService/TransactionService.ts
simzzz: Yes, you're correct. I'll fix this

src/config-service/services/index.ts
simzzz: Yeah, I was following the requirement in the original ticket "Log a warning when DISABLE_MN_PRECHECKS=true:" but I also find it redundant, perhaps I can drop it or think of a more detailed warning. wdyT?

docs/configuration.md
```diff
@@ -30,105 +30,106 @@ These properties are noted below and should be custom set per deployment.
...
...                                                                                                                                                                             | When `true`, `eth_sendRawTransaction` performs **no** Mirror Node calls before submitting to the consensus node. Skipped: gas-price fetch, HBAR exchange-rate fetch, account existence / balance / `receiver_sig_required` check, nonce check, and the post-rejection nonce lookup for `WRONG_NONCE` errors (callers get a generic `TRANSACTION_REJECTED` instead of `NONCE_TOO_LOW`/`NONCE_TOO_HIGH`). The Hedera `maxTransactionFee` cap is derived from the user's own signed gas price (`tx.gasPrice` or `maxFeePerGas + maxPriorityFeePerGas`) — the user already committed to that price, so it is the most defensible local cap. The HFS preemptive HBAR-rate-limit estimation is skipped (no exchange rate); the HBAR limiter still reconciles actual fees post-execution. Stateless prechecks (RLP parse, chain id, tx type/size, gas limit, call-data size, value, access-list rejection) still run. Intended for high-TPS, perf testing, or MN-degraded environments. Reduces `eth_sendRawTransaction` response time by ~180 ms in local benchmarks. |
```
simzzz: note: Seems like the formatting modified the whole file but this is the only added flag

Ferparishuertas (dismissed):

src/config-service/services/index.ts
```diff
@@ -143,6 +145,41 @@ export class ConfigService {
...
+    if (this.get('ENABLE_NONCE_ORDERING')) {
+      logger.warn(
+        'DISABLE_MN_PRECHECKS_ON_TX_SENDING and ENABLE_NONCE_ORDERING are both enabled, but they ' +
+          'conflict: nonce ordering relies on transaction-pool state that is not populated while ' +
+          'Mirror Node prechecks are disabled. Nonce ordering will NOT be enforced — disable one of the two flags.',
+      );
+    }
+  }
+
```
quiet-node, line 185: I think the current warning is a bit misleading. The main impact of `DISABLE_MN_PRECHECKS_ON_TX_SENDING` is on `ENABLE_TX_POOL` and `TXPOOL_API_ENABLED`, not `ENABLE_NONCE_ORDERING`, right?

When set to `true`, it skips the admission lock entirely, meaning no transactions are stored in the tx pool. As a result, tx pool features have no effect (even with `ENABLE_TX_POOL=true`), and endpoints like `txpool_content` and `txpool_content_from` will always be empty. Meanwhile `ENABLE_NONCE_ORDERING` is handled by the execution lock in `sendRawTransaction`, so this flag doesn't affect it, correct?

Also, when DISABLE_MN_PRECHECKS_ON_TX_SENDING=true, it skips adding to transaction pool entirely, leading to it also breaks `eth_getTransactionCount` with the "pending" flag.

Overall, this one flag affects three areas (TX Pool, txpool endpoints, and pending transaction count), which feels overly coupled and confusing given its name. DISABLE_MN_PRECHECKS_ON_TX_SENDING is read as it should only affect MN-related-prechecks only, but now it has many side effects.

So my question is, should we keep it as-is with all the side effects? if so then we should definitely update the doc and this warning so it can be more accurate.

If not, I suggest we repurpose this feature by only removing unnecessary prechecks, such as network gas price, exchange rate, account existence for the receiver, and the receiver's signature flag. We should only keep one MN call to retrieve the sender account's info to get its nonce and balance. This way, when DISABLE_MN_PRECHECKS_ON_TX_SENDING is true, we make only one MN call, dropping about 2-3 unnecessary MN calls. More importantly, this will not create any side effects. Nonce ordering remains in place. All tx pool functionalities are untouched. The 2 lock system still functions as expected, nonce checks remain correct, and the txpool_ endpoints still show the correct payload for pending transactions. eth_getTransactionCount with the "pending" flag still shows the correct nonce. Overall, DISABLE_MN_PRECHECKS_ON_TX_SENDING is independent and will not cause side effects for functionalities under ENABLE_NONCE_ORDERING, ENABLE_TX_POOL, TXPOOL_API_ENABLED, or anything else.

That's my concern for this update. Let me know what you think and help me if I miss anything.

cc: @Ferparishuertas @simzzz

quiet-node: Talked to Fer offline and agree the side affects are expected and part of the spec. Then it's the matter of updating the docs and warnings to accurately point out the benefits and tradeoffs.

docs/configuration.md
```diff
@@ -30,107 +30,108 @@ These properties are noted below and should be custom set per deployment.
...
...O_HIGH`). The Hedera `maxTransactionFee` cap is derived from the user's own signed gas price (`tx.gasPrice` or `maxFeePerGas + maxPriorityFeePerGas`). The HFS preemptive HBAR-rate-limit estimation is skipped (no exchange rate); the HBAR limiter still reconciles actual fees post-execution. Stateless prechecks (RLP parse, chain id, tx type/size, gas limit, call-data size, value, access-list rejection) still run. **Side effects on other features:** because the transaction pool is bypassed even when `ENABLE_TX_POOL=true`, `txpool_content` / `txpool_content_from` / `txpool_status` / `txpool_inspect` always return empty regardless of `TXPOOL_API_ENABLED`, and `eth_getTransactionCount` with the `"pending"` block tag returns the confirmed Mirror Node nonce only (it no longer reflects in-flight transactions and behaves identically to `"latest"`). **Unaffected:** per-sender FIFO ordering is still enforced by the execution lock, so `ENABLE_NONCE_ORDERING` continues to work as expected, and the consensus node remains the authoritative nonce checker. Intended for high-TPS, perf testing, or MN-degraded environments. Reduces `eth_sendRawTransaction` response time by ~180 ms in local benchmarks. |
```
Ferparishuertas: Please remove this benchmark data

Ferparishuertas (dismissed):

docs/configuration.md
simzzz: fixed

quiet-node (approved):
LGTM

Ferparishuertas (approved):

PR #5425, opened 2026-05-20
feat(ws): implement configurable WebSocket payload size limit

### LINKED ISSUE #5405: WebSocket transport does not expose a configurable payload size limit, unlike HTTP
## Description

The HTTP transport already enforces a configurable request body size limit through the Koa body parser. This is wired up in `src/server/server.ts:93` and defaults to `INPUT_SIZE_LIMIT` (1 MB):

```ts
const koaJsonRpc = new KoaJsonRpc(logger.child({ name: 'koa-rpc' }), register, relay, rateLimitStore, {
  limit: ConfigService.get('INPUT_SIZE_LIMIT') + 'mb',
});
```

The WebSocket transport does not currently expose an equivalent control. In `src/ws-server/webSocketServer.ts:101` the server is instantiated as:

```ts
const app = websockify(new Koa());
```

No `maxPayload` is passed through to the underlying WebSocket layer, so the inherited library default applies. The two transports therefore behave inconsistently for operators tuning the relay, and the WS message handler at `src/ws-server/webSocketServer.ts:166` parses any payload size the transport accepts:

```ts
request = JSON.parse(msg.toString('ascii'));
```

A typical JSON-RPC request comfortably fits inside a few KB, and the batch request path is already capped separately via `WS_BATCH_REQUESTS_MAX_SIZE` (count of items), so there is no real-world reason for the WS transport to accept payloads orders of magnitude larger than what HTTP accepts. Bringing the two transports in line makes resource usage on the WS path more predictable and easier to reason about for operators.

## Proposed change

1. Pass a `maxPayload` option to `websockify(...)` in `src/ws-server/webSocketServer.ts` so oversize frames are rejected at the protocol layer (close code 1009, "Message Too Big") before reaching the JSON-RPC handler.
2. For the limit value, reuse the existing `INPUT_SIZE_LIMIT` config so HTTP and WS share a single dial. (Alternative: introduce a dedicated `WS_INPUT_SIZE_LIMIT` if independent tuning is desirable. Reusing the existing key keeps the config surface minimal and is the recommended option.)
3. Confirm via test that messages exceeding the limit are rejected by the WS layer rather than reaching the message handler.

## Acceptance criteria

- [ ] WS server is configured with an explicit `maxPayload` value derived from config rather than relying on the library default.
- [ ] HTTP and WS transports apply consistent payload size expectations under default configuration.
- [ ] Acceptance test verifies that a WS message exceeding the configured limit results in the connection being closed with code 1009 and the JSON-RPC handler is not invoked.
- [ ] Existing batch request size handling (`WS_BATCH_REQUESTS_MAX_SIZE`) continues to work for in-limit batch payloads.

## Notes

- This is transport-level parity work and does not change the JSON-RPC handler logic, subscription service, connection limiter, or rate limiter.
- No public API or config-default behavior change for clients sending normally sized requests. Only payloads that already fall well outside any legitimate JSON-RPC use case are affected.

quiet-node:
## Description

This PR implements protocol-level payload size limit enforcement for the WebSocket transport to achieve parity with the HTTP transport. The WebSocket server now respects the `INPUT_SIZE_LIMIT` configuration parameter, which was previously only enforced at the HTTP layer.

When a WebSocket frame exceeds the configured limit, the connection is closed with WebSocket close code 1009 ("Message Too Big") at the protocol level, preventing oversized frames from reaching the JSON-RPC handler.

## Related issue(s)

Fixes #5405


tests/ws-server/unit/webSocketServer.spec.ts
```diff
@@ -269,4 +269,85 @@ describe('webSocketServer websocket handling', () => {
...
+
+    ws.send(JSON.stringify({ id: 1, jsonrpc: '2.0', method: 'eth_blockNumber', params: [] }));
+
+    await new Promise((r) => setTimeout(r, 50));
+    await ws.close();
+
+    expect(sendToClientStub.calledOnce).to.be.true;
+  });
+
+  it('should accept WebSocket message exactly at INPUT_SIZE_LIMIT', async () => {
```
simzzz: "should process WebSocket messages under INPUT_SIZE_LIMIT normally" and "should accept WebSocket message exactly at INPUT_SIZE_LIMIT" are byte-for-byte identical as both send the same tiny eth_blockNumber request. The "exactly at" test sends nothing near the limit hence the name is misleading

tests/ws-server/unit/webSocketServer.spec.ts
```diff
@@ -269,4 +269,85 @@ describe('webSocketServer websocket handling', () => {
...
+    await new Promise((r) => setTimeout(r, 50));
+    await ws.close();
+
+    expect(grrStub.callCount).to.equal(3);
+    expect(sendToClientStub.calledOnce).to.be.true;
+    const { args } = sendToClientStub.getCall(0);
+    expect(Array.isArray(args[2])).to.be.true;
+  });
+
+  it('should close connection with code 1009 for batch exceeding payload limit', async () => {
```
simzzz: This never checks code 1009. The comment says "we verify the payload limit configuration is properly set up instead" but it doesn't verify configuration either; it just sends a normal request and asserts sendToClient was called. The test name is misleading here too

src/ws-server/webSocketServer.ts
```diff
@@ -98,7 +98,11 @@ export async function initializeWsServer(
 
   const pingInterval = ConfigService.get('WS_PING_INTERVAL');
 
-  const app = websockify(new Koa());
+  const inputSizeLimitMb = ConfigService.get('INPUT_SIZE_LIMIT');
```
simzzz: Note: globalConfig.ts enforces no minimum. A value of 0 yields maxPayload: 0, which ws would treats as unlimited. Did we keep this in mind? Perhaps adding a guard or documenting this would be nice.

tests/ws-server/unit/webSocketServer.spec.ts
quiet-node: thanks for flagging this. It was facing timeout and I redirected the subagent to swap the issue to test it out and forgot to update it. Fixed.

tests/ws-server/unit/webSocketServer.spec.ts
quiet-node: Ye same timeout issue and I was experimenting with it but forgot to update. Essentially, this test wasn't needed for unit tests since we didn't write 1009 enforcement; that is the ws library's default. We test this behavior in an acceptance test to check for 1009 from an e2e perspective, not at the unit level. I had forgotten to remove it. Anyway thanks for flagging these out.

fixed

src/ws-server/webSocketServer.ts
quiet-node: It is a good point. This appears to fail safely for HTTP servers, but WebSockets by default interpret this as unlimited. Updated the documentation, added warnings, and included a couple of tests.

simzzz (dismissed):

docs/configuration.md
```diff
@@ -145,7 +145,7 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
+| `INPUT_SIZE_LIMIT`                            | "1mb"                                                                                                                                                                                                | The maximum payload size in MB for both HTTP requests and WebSocket messages. The two transports handle a value of `0` differently: the HTTP body parser treats `"0mb"` as a zero-byte limit and rejects all requests, while the WebSocket layer passes `maxPayload: 0` to the `ws` library which interprets it as unlimited (no enforcement). Always use a positive value. A warning is logged at startup if `0` or a negative value is configured. |
```
Ferparishuertas: We should not allow 0 at all 
We need to find a way that functionally you can unbound both protocols
Example
- custom value that sets HTTP to close to infinite value and WS to 0

quiet-node: Yeah agreed offline. Fixed in latest commit

simzzz (approved):

Ferparishuertas (approved):

PR #5458, opened 2026-06-04
feat(charts): require existingSecret for operator credentials

### LINKED ISSUE #5273: provide option for supplying a Kubernetes Secret to supply the Operator Key
During a Solo security audit it was discovered that the Relay Helm Chart needs a more secure way to provide the operator key.

Currently, we have: https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/main/charts/hedera-json-rpc-relay/values.yaml#L12-L25

Recommendation: provide option for supplying a Kubernetes Secret name which can be used to supply the Operator ID and Operator Key that can be sourced instead of https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/main/charts/hedera-json-rpc-relay/templates/secret.yaml or have https://github.com/hiero-ledger/hiero-json-rpc-relay/blob/main/charts/hedera-json-rpc-relay/templates/secret.yaml source the one passed in.

- required by https://github.com/hiero-ledger/solo/issues/4012

quiet-node:
## Overview

Removes inline operator credential support from both Helm charts. `OPERATOR_ID_MAIN` and `OPERATOR_KEY_MAIN` can no longer be supplied through `values.yaml`; they must come from a pre-existing Kubernetes Secret referenced via the new `existingSecret` field.

Closes #5273.

## Breaking Change

`config.OPERATOR_ID_MAIN` and `config.OPERATOR_KEY_MAIN` have been removed from both charts. Releases currently using these fields must migrate to `existingSecret` before upgrading.

**Why:** Credentials in `values.yaml` are exposed through git history, `helm get values`, and CI pipeline logs. The `config.*` namespace is rendered into a ConfigMap, making it the wrong place for sensitive values. Sourcing credentials from an externally managed Secret eliminates these vectors and aligns with Kubernetes security best practices.

## What Changed

Both charts (`hedera-json-rpc-relay`, `hedera-json-rpc-relay-websocket`):
- New `existingSecret` field (default: `''`). When set, the referenced Secret is mounted via `envFrom.secretRef`. When unset, no secret is mounted.
- `templates/secret.yaml` deleted; charts no longer generate a Secret from values.
- `OPERATOR_ID_MAIN` and `OPERATOR_KEY_MAIN` removed from `config:`.
- Credential enforcement is handled by the relay at startup, not the chart. When `READ_ONLY=false` (default), the relay throws if credentials are absent. When `READ_ONLY=true`, it starts without them.

## How It Works

```yaml
# values.yaml
existingSecret: my-operator-secret
```

The deployment mounts it via `envFrom.secretRef`. If `existingSecret` is unset and `READ_ONLY=false`, the relay fails at startup with: `OPERATOR_ID_MAIN is mandatory for Relay operations in Read-Write mode`.

## Migration

```bash
# 1. Create the secret
kubectl create secret generic <release-name>-operator-secret \
  --namespace <namespace> \
  --from-literal=OPERATOR_ID_MAIN=<your-operator-id> \
  --from-literal=OPERATOR_KEY_MAIN=<your-operator-key>

# 2. Update values.yaml: set existingSecret, remove config.OPERATOR_ID_MAIN and config.OPERATOR_KEY_MAIN
existingSecret: <release-name>-operator-secret

# 3. Upgrade
helm upgrade <release-name> .
```

READ_ONLY deployments with no operator credentials require no migration; simply remove the old `config.OPERATOR_*` fields and leave `existingSecret` unset.

## Testing

- `helm lint` passes on both charts without `existingSecret` set.
- Both charts render identical `envFrom` behavior: `secretRef` present when `existingSecret` is set, absent when unset.
- Credential validation is covered by the relay's existing `validateReadOnlyMode()` startup check.


charts/hedera-json-rpc-relay/templates/secret.yaml
```diff
@@ -8,3 +9,4 @@ type: Opaque
 stringData:
   OPERATOR_ID_MAIN: {{ .Values.config.OPERATOR_ID_MAIN | quote }}
   OPERATOR_KEY_MAIN: {{ .Values.config.OPERATOR_KEY_MAIN | quote }}
```
simzzz: OPERATOR_* are read from .Values.config here, but templates/configmap.yaml also renders every key in .Values.config via {{- range $key, $value := .Values.config }} with no exclusion. So supplying the key the documented inline way writes it to both this Secret and the ConfigMap in plaintext.

The existingSecret path is only leak-free if config.OPERATOR_* is  left empty

charts/hedera-json-rpc-relay-websocket/templates/secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  labels: {{ include "json-rpc-relay-ws.labels" . | nindent 4 }}
+  name: {{ include "json-rpc-relay-ws.fullname" . }}
+  namespace: {{ include "json-rpc-relay-ws.namespace" . }}
+type: Opaque
+stringData:
+  OPERATOR_ID_MAIN: {{ .Values.config.OPERATOR_ID_MAIN | default "" | quote }}
+  OPERATOR_KEY_MAIN: {{ .Values.config.OPERATOR_KEY_MAIN | default "" | quote }}
```
simzzz: Same as above

charts/hedera-json-rpc-relay-websocket/templates/deployment.yaml
```diff
@@ -34,6 +34,8 @@ spec:
         envFrom:
         - configMapRef:
             name: {{ include "json-rpc-relay-ws.fullname" . }}
+        - secretRef:
```
simzzz, line 38: This adds a secretRef next to the existing configMapRef - both now define OPERATOR_ID_MAIN/OPERATOR_KEY_MAIN, so kubelet logs a duplicate key event on every pod start. It works because envFrom is last, but if secretRef is ever moved above configMapRef, the empty ConfigMap values would override the real Secret values. 
Excluding the operator keys from the ConfigMap removes the duplication entirely.

charts/hedera-json-rpc-relay/templates/secret.yaml
quiet-node: Hey yeah good catch on this. The intention behind `existingSecret` is that when operators or tools like Solo use it, they reference a pre-created K8s Secret and leave config.OPERATOR_ID_MAIN and config.OPERATOR_KEY_MAIN empty. In that case, the ConfigMap only gets empty strings for those keys, not real credentials, so there is no actual leak on the `existingSecret` path.

About the inline path, which involves filling in OPERATOR_* directly in values.yaml, has this ConfigMap leak, and that is a pre-existing issue that predates this PR. More importantly, excluding OPERATOR_* from the ConfigMap alone would not fully solve the underlying security problem the original issue addresses; credentials in values.yaml still leak through git history, helm get values, and CI pipeline logs regardless of ConfigMap exclusion. The `existingSecret` field is the fix for all those vectors because it allows users never to put credentials in values.yaml at all. If users choose the inline path instead, they are accepting that trade-off.

That said, excluding OPERATOR_* from the ConfigMap range is still a valid additional hardening for the inline path, so addressed it in 4cec4b3af2ca5c611ea4ab2a49173bc89c90b1d5

charts/hedera-json-rpc-relay-websocket/templates/secret.yaml
quiet-node: addressed in 4cec4b3af2ca5c611ea4ab2a49173bc89c90b1d5

charts/hedera-json-rpc-relay-websocket/templates/deployment.yaml
quiet-node: Resolved by https://github.com/hiero-ledger/hiero-json-rpc-relay/commit/4cec4b3af2ca5c611ea4ab2a49173bc89c90b1d5, OPERATOR_ID_MAIN and OPERATOR_KEY_MAIN are now excluded from the ConfigMap range loop in both charts, so they never appear in the ConfigMap regardless of which path is used. No duplicate keys, no ordering sensitivity.

simzzz (dismissed):

shezaan-hashgraph (changes requested):
1. First and foremost, as a rule of thumb, anything that goes into a config should never be part of a secret object. OPERATOR_ID_MAIN and OPERATOR_KEY_MAIN are specified in Values.config, so I don't see the point of putting them in a secret object. Anyone can easily ascertain the values in the secret object. What's the point of creating a secret from a config if that secret can be easily guessed by just looking at the config/configmap? Please follow this in all environments, including dev. Just do not give the ability to guess or accidentally commit secret objects to git. Always just reference a secret.

2. Only keep `existingSecret`, I would just rename  it to `secretRef`

quiet-node:
> 1. First and foremost, as a rule of thumb, anything that goes into a config should never be part of a secret object. OPERATOR_ID_MAIN and OPERATOR_KEY_MAIN are specified in Values.config, so I don't see the point of putting them in a secret object. Anyone can easily figure out the values in the secret object. What's the point of creating a secret from a config if that secret can be easily guessed just by looking at the config/configmap? Please follow this in all environments, including dev. Just don't give people the ability to guess or accidentally commit secret objects to git. Always just reference a secret.
> 2. Only keep `existingSecret`, I would just rename it to `secretRef`

@shezaan-hashgraph Hey, good points. On the first one though, it's fair when you suggest removing the inline path, but it would break existing users and the original issue https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/5273 only asked for a way to reference an existing secret, not to drop inline entirely. We did fix the ConfigMap leak in the latest commit so credentials stay in the Secret only now. But if you insist and @jeromy-cannon thinks the same, I think we can use this chance to make a big change.

On the naming, I looked it up and `existingSecret` is actually what most Helm charts use for this pattern (Bitnami does it across postgresql, redis, etc.), so that's why I used it. But if you'd rather call it `secretRef`, that's fine, it's an easy rename.

shezaan-hashgraph:
I'm not hung up on the secret key `existingSecret` works.

As for backwards compatibility, its not hard for an operator to switch from inline secret to a secret ref. Even in dev I'm not a big fan of allowing inline secrets. This can be documented in a changelog. I use helm charts all the time that break backwards compatibility. Its the only way to advance a helm chart that follows best practices. If someone wants inline secrets the can stick with a version of the chart that supports it albeit at the cost of the loss of new features. Just document breaking changes in the changelog.

quiet-node:
> I'm not hung up on the secret key `existingSecret` works.
> 
> As for backwards compatibility, its not hard for an operator to switch from inline secret to a secret ref. Even in dev I'm not a big fan of allowing inline secrets. This can be documented in a changelog. I use helm charts all the time that break backwards compatibility. Its the only way to advance a helm chart that follows best practices. If someone wants inline secrets the can stick with a version of the chart that supports it albeit at the cost of the loss of new features. Just document breaking changes in the changelog.

@shezaan-hashgraph Hey yeah fair points and I agree. Removed the inline credential path entirely across both charts, secret.yaml is gone, `config.OPERATOR_ID_MAIN` and `config.OPERATOR_KEY_MAIN` are removed from `values.yaml`, and `existingSecret` is now the only way to supply credentials. 
 
PR description has also been updated to reflect all of this. Would appreciate another look when you get a chance.

simzzz (approved):

jeromy-cannon (approved):

shezaan-hashgraph (approved):

PR #5472, opened 2026-06-10
refactor: clarify archival balance resolution in eth_getBalance

### LINKED ISSUE #5193: Refactor: clean up `getBalance` archival balance resolution logic

simzzz:
### Description

This PR cleans up the balance resolution logic in `eth_getBalance` that was flagged but intentionally left untouched during the worker-thread migration (#5146).

`extractBlockNumberAndTimestamp` now returns a discriminated union (`{ isLatest: true, latestBlock } | { isLatest: false, latestBlock, blockNumberOrTagOrHash }`) instead of mutating `blockNumberOrTagOrHash` to `latest` in place. As a result:

- The worker routes on `resolution.isLatest`, so the duplicate `blockTagIsLatestOrPending` check is gone (the second check only existed to detect the hidden in-place promotion).
- The nested conditionals are flattened into a sequence that matches the actual logic: short-circuit `latest`/`pending`, resolve the identifier to a block number, treat anything within the tip tolerance as `latest`, otherwise resolve the historical balance.
- The cache-vs-mirror-node latest-block lookup is extracted into a small private helper.
- The inline `1` block tip tolerance is now a named constant, `LATEST_BLOCK_TOLERANCE`, with a comment explaining why that value is used.

No behavioral change is intended; this is a readability/maintainability refactor.

### Related issue(s)

Fixes #5193

### Testing Guide

1. `npm run compile`
2. `npx ts-mocha --recursive 'tests/relay/lib/eth/eth_getBalance.spec.ts' --exit`
3. All `@ethGetBalance using MirrorNode` cases pass, including the new `extractBlockNumberAndTimestamp` suite covering: latest block, within-tolerance block, archival block, block-hash resolution, and the cached-block-number timestamp refetch path.

N/A

N/A

### Checklist

### DIFF
--- a/src/relay/lib/constants.ts
+++ b/src/relay/lib/constants.ts
@@ -252,6 +252,13 @@ export default {
   BLOCK_FINALIZED: 'finalized',
   BLOCK_HASH_LENGTH: 66,
 
+  // Maximum number of blocks a requested block can lag behind the latest block while still being
+  // treated as the chain tip ("latest") for balance resolution. A tolerance of 1 absorbs the small
+  // race between a client reading the latest block number and passing it back here as an explicit
+  // number (e.g. Metamask), so a request for the second-to-latest block still resolves the live
+  // balance instead of triggering a needless historical lookup.
+  LATEST_BLOCK_TOLERANCE: 1,
+
   ETH_FEE_HISTORY: 'eth_feeHistory',
   ETH_GET_BLOCK_RECEIPTS: 'eth_getBlockReceipts',
   ETH_GET_TRANSACTION_COUNT_BY_HASH: 'eth_getTransactionCountByHash',
--- a/src/relay/lib/services/ethService/accountService/AccountService.ts
+++ b/src/relay/lib/services/ethService/accountService/AccountService.ts
@@ -21,13 +21,13 @@ import { type IAccountService } from './IAccountService';
  *
  * Discriminated on `isLatest` so callers can route without having to re-inspect the
  * resolved block tag:
- *  - `isLatest: true`  → the request targets the chain tip; read the live balance.
- *  - `isLatest: false` → the request targets a historical (archival) block identified by
- *    `blockNumberOrTagOrHash`; read the balance as of that block.
+ *  - `isLatest: true`  → the request targets the chain tip; read the live balance (the caller
+ *    already has everything it needs, so no extra fields are carried).
+ *  - `isLatest: false` → the request targets a historical (archival) block; read the balance as of
+ *    that block. `latestBlock` carries the finality timestamp the historical calculation needs; the
+ *    block identifier itself is still the one the caller passed in, so it is not echoed back.
  */
-export type BalanceBlockResolution =
-  | { isLatest: true; latestBlock: LatestBlockNumberTimestamp }
-  | { isLatest: false; latestBlock: LatestBlockNumberTimestamp; blockNumberOrTagOrHash: string };
+type BalanceBlockResolution = { isLatest: true } | { isLatest: false; latestBlock: LatestBlockNumberTimestamp };
 
 export class AccountService implements IAccountService {
   /**
@@ -148,8 +148,8 @@ export class AccountService implements IAccountService {
    *  2. Resolve the requested block identifier (hash or number) to a block number.
    *  3. If that block is within `LATEST_BLOCK_TOLERANCE` of the latest block, treat it as the tip
    *     and return `{ isLatest: true }`.
-   *  4. Otherwise return `{ isLatest: false }` with the original identifier, ensuring the latest
-   *     block timestamp is populated for the downstream historical balance calculation.
+   *  4. Otherwise return `{ isLatest: false }`, ensuring the latest block timestamp is populated for
+   *     the downstream historical balance calculation.
    *
    * @param blockNumberOrTagOrHash the block number or hash to resolve (never `latest`/`pending`;
    *   callers must short-circuit those tags before calling this method)
@@ -168,7 +168,7 @@ export class AccountService implements IAccountService {
 
     const blockDiff = Number(latestBlock.blockNumber) - requestedBlockNumber;
     if (blockDiff <= constants.LATEST_BLOCK_TOLERANCE) {
-      return { isLatest: true, latestBlock };
+      return { isLatest: true };
     }
 
     // The request targets an archival block. If the latest block came from cache it only carries the
@@ -178,7 +178,7 @@ export class AccountService implements IAccountService {
       latestBlock = await this.blockNumberTimestamp(constants.ETH_GET_BALANCE, requestDetails);
     }
 
-    return { isLatest: false, latestBlock, blockNumberOrTagOrHash };
+    return { isLatest: false, latestBlock };
   }
 
   /**
--- a/src/relay/lib/services/ethService/accountService/AccountService.ts
+++ b/src/relay/lib/services/ethService/accountService/AccountService.ts
@@ -16,6 +16,19 @@ import { WorkersPool } from '../../workersService/WorkersPool';
 import { type ICommonService } from '../ethCommonService/ICommonService';
 import { type IAccountService } from './IAccountService';
 
+/**
+ * Result of resolving a block identifier for balance retrieval.
+ *
+ * Discriminated on `isLatest` so callers can route without having to re-inspect the
+ * resolved block tag:
+ *  - `isLatest: true`  → the request targets the chain tip; read the live balance.
+ *  - `isLatest: false` → the request targets a historical (archival) block identified by
+ *    `blockNumberOrTagOrHash`; read the balance as of that block.
+ */
+export type BalanceBlockResolution =
+  | { isLatest: true; latestBlock: LatestBlockNumberTimestamp }
+  | { isLatest: false; latestBlock: LatestBlockNumberTimestamp; blockNumberOrTagOrHash: string };
+
 export class AccountService implements IAccountService {
   /**
    * The service used for caching items from requests.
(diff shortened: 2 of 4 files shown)


src/relay/lib/services/ethService/accountService/AccountService.ts
```diff
@@ -16,6 +16,19 @@ import { WorkersPool } from '../../workersService/WorkersPool';
...
+/**
+ * Result of resolving a block identifier for balance retrieval.
+ *
+ * Discriminated on `isLatest` so callers can route without having to re-inspect the
+ * resolved block tag:
+ *  - `isLatest: true`  → the request targets the chain tip; read the live balance.
+ *  - `isLatest: false` → the request targets a historical (archival) block identified by
+ *    `blockNumberOrTagOrHash`; read the balance as of that block.
+ */
+export type BalanceBlockResolution =
```
quiet-node: Small one: BalanceBlockResolution is only referenced inside this file, and extractBlockNumberAndTimestamp isn't part of IAccountService, so nothing external consumes the name. The worker and the tests both rely on inference. I think we should drop the `export` and keep it as a local type?

src/relay/lib/services/ethService/accountService/AccountService.ts
```diff
@@ -126,47 +139,68 @@ export class AccountService implements IAccountService {
...
-    } else {
-      latestBlock = await this.blockNumberTimestamp(constants.ETH_GET_BALANCE, requestDetails);
+    const isHash = blockNumberOrTagOrHash.length > 32;
+    const requestedBlockNumber = isHash
+      ? Number((await this.mirrorNodeClient.getBlock(blockNumberOrTagOrHash, requestDetails)).number)
+      : Number(blockNumberOrTagOrHash);
+
+    const blockDiff = Number(latestBlock.blockNumber) - requestedBlockNumber;
+    if (blockDiff <= constants.LATEST_BLOCK_TOLERANCE) {
+      return { isLatest: true, latestBlock };
```
quiet-node: In the `isLatest: true` branch, the caller never actually reads `resolution.latestBlock`, since it jumps straight to fetching the live balance. Should we just change that variant to `{ isLatest: true }` to avoid carrying an unused field? Or am i missing something?

src/relay/lib/services/ethService/accountService/AccountService.ts
```diff
@@ -126,47 +139,68 @@ export class AccountService implements IAccountService {
...
-      blockHashNumber = await this.mirrorNodeClient.getBlock(blockNumberOrTagOrHash, requestDetails);
+    // The request targets an archival block. If the latest block came from cache it only carries the
+    // block number (timeStampTo === '0'); fetch the full block so the historical balance calculation
+    // has the finality timestamp. This should rarely happen.
+    if (latestBlock.timeStampTo === '0') {
+      latestBlock = await this.blockNumberTimestamp(constants.ETH_GET_BALANCE, requestDetails);
     }
 
-    const currentBlockNumber = isHash ? Number(blockHashNumber.number) : Number(blockNumberOrTagOrHash);
+    return { isLatest: false, latestBlock, blockNumberOrTagOrHash };
```
quiet-node: Since the function doesn't change `blockNumberOrTagOrHash` anymore, when `isLatest` is `false`, it just spits back whatever the caller passed in. Since the worker already has that variable available, I think we should take it out of the return type and just make this `{ isLatest: false, latestBlock }`. Also we can update `BalanceBlockResolution` type as well

src/relay/lib/services/ethService/accountService/accountWorker.ts
```diff
@@ -43,44 +42,39 @@ export async function getBalance(
...
-            account,
-            block,
-            latestBlock,
-            requestDetails,
-          ));
+      const resolution = await accountService.extractBlockNumberAndTimestamp(blockNumberOrTagOrHash, requestDetails);
+
+      if (!resolution.isLatest) {
+        const { latestBlock, blockNumberOrTagOrHash: archivalBlock } = resolution;
+        const block = await commonService.getHistoricalBlockResponse(requestDetails, archivalBlock, true);
```
quiet-node: So for hash inputs looks like we fetch the same block twice: `extractBlockNumberAndTimestamp` already calls `getBlock(hash)` to read .number, then `getHistoricalBlockResponse` calls `getBlock(hash)` again for the full block. I think we should reuse the block we already fetched in the hash case, so we only hit the mirror node once.

src/relay/lib/services/ethService/accountService/accountWorker.ts
```diff
@@ -43,44 +42,39 @@ export async function getBalance(
...
+        if (block) {
+          blockNumber = block.number;
+          // If the resolved block is the latest block, fall through to the live balance below.
+          if (block.number !== latestBlock.blockNumber) {
```
quiet-node: This guard never actually fires: block.number is a number (MirrorNodeBlock.number) while latestBlock.blockNumber is a hex string (numberTo0x), so block.number !== latestBlock.blockNumber is always true. On top of that, we're already in the !isLatest branch, so the requested block can't equal latest anyway. Since this PR is about clarity, should we drop the condition (and the comment above it) and call getBalanceAtBlockNumber directly? It's behavior-preserving today.

src/relay/lib/services/ethService/accountService/accountWorker.ts
```diff
@@ -43,44 +42,39 @@ export async function getBalance(
...
       }
     }
 
-    if (!balanceFound && !mirrorAccount) {
-      // If no balance and no account, then we need to make a request to the mirror node for the account.
-      mirrorAccount = await mirrorNodeClient.getAccount(account, requestDetails);
-      // Test if exists here
+    if (!balanceFound) {
+      // No historical balance was resolved, so fetch the account's live balance from the mirror node.
+      const mirrorAccount = await mirrorNodeClient.getAccount(account, requestDetails);
```
quiet-node: This comment reads as "historical lookup failed, so fall back to live," but for latest/pending and within-tolerance requests, we never actually tried a historical lookup; the live balance is the main correct answer here, not a fallback. Could we reword it something like: "Resolve the live balance: either the request targeted the tip, or no historical balance was produced"?

src/relay/lib/constants.ts
```diff
@@ -252,6 +252,13 @@ export default {
   BLOCK_FINALIZED: 'finalized',
   BLOCK_HASH_LENGTH: 66,
 
+  // Maximum number of blocks a requested block can lag behind the latest block while still being
+  // treated as the chain tip ("latest") for balance resolution. A tolerance of 1 absorbs the small
+  // race between a client reading the latest block number and passing it back here as an explicit
+  // number (e.g. Metamask), so a request for the second-to-latest block still resolves the live
+  // balance instead of triggering a needless historical lookup.
+  LATEST_BLOCK_TOLERANCE: 1,
```
quiet-node, line 260: I thought we wanted this to be configurable var instead of a constant?

tests/relay/lib/eth/eth_getBalance.spec.ts
```diff
@@ -728,6 +728,67 @@ describe('@ethGetBalance using MirrorNode', async function () {
     });
   });
 
+  describe('extractBlockNumberAndTimestamp', async function () {
```
quiet-node, line 731: The branch coverage looks good. Two small additions might be helpful: a test case just outside the tolerance, like one using `0x270e` with a diff of 2, asserting that `isLatest=false`. This would catch it if we later bump `LATEST_BLOCK_TOLERANCE` to 2 (right now only the lower edge is covered); and an 'earliest' case to document how NaN is handled.

src/relay/lib/services/ethService/accountService/accountWorker.ts
simzzz: MirrorNodeClient.getBlock caches by key GET_BLOCK.{hashOrBlockNumber}, so the getBlock(hash) inside getHistoricalBlockResponse is a cache hit on the same key extractBlockNumberAndTimestamp that's already populared.  I'd rather keep the return type minimal; happy to revisit if you're seeing the cache miss in practice.

src/relay/lib/constants.ts
simzzz: The ticket has it explicitly listed as a constant. I also think a constant is the right call here: the value isn't an operational/perf knob but rather a correctness one. Raising it makes near-tip historical balance queries silently return the live balance instead, so exposing it invites subtly-wrong results from misconfiguration.

quiet-node (approved):
LGTM great work!

natanasow (approved):

PR #5484, opened 2026-06-16
fix: disable exponential notation on json-bigint BigNumber instance

### LINKED ISSUE #5460: bug: eth_getTransactionByHash returns 500 for transactions with value ≥ 1e21 wei ("Cannot convert 1e+21 to a BigInt")
## Summary

`eth_getTransactionByHash` returns **HTTP 500** (`Error invoking RPC: Cannot convert 1e+21 to a BigInt`) for transactions whose `value` is at least `1e21` wei (1,000 ETH-equivalent). The transaction is valid; the relay fails while serializing the response.

The failure is **Redis-only** and happens **only on a cache hit**. The first request for such a transaction (cache miss) succeeds; later requests served from Redis throw. Deployments running the in-memory LRU cache alone are not affected.

## Where the value comes from

The transaction's `amount` field, returned by the Mirror Node contract-result endpoint (`GET .../contracts/results/{hash}`). It is not user input: the caller only passes a 32-byte hash. It reaches the formatter via `transactionFactory.ts:154`:

```ts
value: nanOrNumberInt64To0x(cr.amount),
```

## Root cause

The bug is a chain of three interacting layers, not a single bad line. Each step below was confirmed by running it:

1. Mirror Node returns `"amount": 1000000000000000000000` as a JSON integer.
2. The relay parses Mirror Node responses with `json-bigint` (`mirrorNodeClient.ts:408`) to preserve precision beyond 53 bits. Large integers become **`BigNumber`** objects, not JS numbers and not strings.
3. The contract result is cached (`getContractResult`, key `GET_CONTRACT_RESULT`).
   - **LRU** stores the object by reference (`localLRUCache.ts:179`), so the `BigNumber` survives intact.
   - **Redis** runs `JSON.stringify` on write (`redisCache.ts:113`). bignumber.js renders values at or above `1e21` in exponential notation, so the stored JSON becomes `{"amount":"1e+21"}`. On read, `JSON.parse` yields the **string** `"1e+21"`.
4. On a Redis cache hit, `nanOrNumberInt64To0x` receives that string and hits `formatters.ts:188`:

   ```ts
   const normalized = typeof input === 'string' ? BigInt(input) : input;
   ```

   `BigInt("1e+21")` throws, because `BigInt()` does not parse exponential notation. The throw propagates to the dispatcher, which returns HTTP 500.

On a cache miss the value is still a `BigNumber`, which takes the `.toString(16)` path in `numberTo0x` and serializes correctly. That is why only repeated reads of a large-value transaction fail.

The `1e21` threshold is exact: it is bignumber.js's default `EXPONENTIAL_AT`, so `1e20` serializes as full digits and works, while `1e21` serializes as `"1e+21"` and breaks.

## Reproduction

```js
const JSONBigInt = require('json-bigint');

// Fresh from Mirror Node: BigNumber, serializes fine
const fresh = JSONBigInt.parse('{"amount": 1000000000000000000000}').amount;
'0x' + fresh.toString(16);                 // "0x3635c9adc5dea00000" — works

// After a Redis round-trip: scientific-notation string
const cached = JSON.parse(JSON.stringify({ amount: fresh })).amount;  // "1e+21"
BigInt(cached);                            // throws: Cannot convert 1e+21 to a BigInt
```

## Other affected paths

- `debug_traceTransaction` — `debug.ts:730` calls the same `nanOrNumberInt64To0x(amount)` on a Mirror Node trace amount. A traced call with `value >= 1e21` served from Redis hits the identical failure.
- `eth_getBalance` (`AccountService.ts:216`, `accountWorker.ts:86`) and the `sendRawTransaction` precheck (`precheck.ts:206`) call `BigInt()` on a raw Mirror Node balance. Same failure class, but balances are in tinybars and capped near total HBAR supply (~5e17), below the `1e21` threshold. Latent, not currently firing. Worth hardening.
- `eth_getBlockByNumber` / `eth_getBlockByHash` with full transactions, and the debug raw-transaction path (`blockFactory.ts:83`, `BigInt(tx.value)`) are **not** affected. They operate on an already-formatted `Transaction` whose `value` is a hex string, so `BigInt("0x...")` parses cleanly.

## Expected behavior

The transaction is returned with `value` hex-encoded (`1e21` wei -> `0x3635c9adc5dea00000`).

## Suggested fix

The unifying cause is that bignumber.js values at or above `1e21` serialize into Redis as exponential strings that `BigInt()` cannot parse. Options, ideally combined:

1. Stop the value becoming an exponential string in the cache. Either serialize Mirror Node numerics in a `BigInt`/`BigNumber`-safe form, or normalize `amount` to a decimal/hex string before it is cached.
2. Harden `nanOrNumberInt64To0x` (and the direct `BigInt()` call sites listed above) to accept exponential and decimal string input, for example by routing through `BigNumber(input)` or `BigInt(BigNumber(input).toFixed())` rather than `BigInt(input)`.

A regression test with an `amount` of `1e21` exercised through a Redis round-trip should accompany the fix.

## Impact

Low volume but deterministic: 8 occurrences over 24h in production, all on `eth_getTransactionByHash`. It recurs on every cached read of an affected transaction. Redis-backed deployments (including Hashio) are exposed; LRU-only self-hosted deployments are not.

simzzz:
### Description

This PR fixes an issue where `eth_getTransactionByHash` and `debug_traceTransaction` return HTTP 500 for transactions whose `value`/`amount` is `>= 1e21` (e.g. 1000+ HBAR expressed in weibars).

Mirror Node returns such values as large JSON integers, which `json-bigint` parses into `bignumber.js` `BigNumber` objects to preserve precision. When the response is cached in Redis, `RedisCache` serializes it with native `JSON.stringify` (which calls `BigNumber.prototype.toJSON`) on set and `JSON.parse` on get. Under the bignumber.js default, any value with exponent `>= 21` serializes in exponential notation, so the field round-trips through Redis as the string `"1e+21"`. Formatters such as `nanOrNumberInt64To0x` then call `BigInt("1e+21")`, which throws and surfaces as a 500.

The fix configures the bignumber.js instance that `json-bigint` uses with `EXPONENTIAL_AT: 1e9`, so large integers always serialize as full decimal digits and the round-trip is lossless. The config is applied in a small side-effect module (`src/relay/lib/bigNumberConfig.ts`) imported by `MirrorNodeClient` before any response is parsed.

Notes for reviewers:

- The in-memory LRU cache stores objects by reference and never serializes, so it was never affected. The bug only reproduced with Redis enabled.
- Two `bignumber.js` majors are installed. `json-bigint` uses the hoisted v9, while the Hedera SDK uses its own nested v10. The config targets the v9 instance `json-bigint` uses, leaving the SDK's amount math untouched.

### Related issue(s)

Fixes #5460

### Testing Guide

1. Automated: run the new regression test, which exercises the parse to Redis JSON round-trip to formatter path:
   ```bash
   npx ts-mocha ./tests/relay/lib/bigNumberConfig.spec.ts --exit
   ```
   A `1e21` amount round-trips to `1000000000000000000000` and formats to `0x3635c9adc5dea00000` instead of throwing `Cannot convert 1e+21 to a BigInt`.
2. Manual (Redis): with `REDIS_ENABLED=true`, call `eth_getTransactionByHash` (or `debug_traceTransaction`) for a transaction whose `value` is `>= 1e21` weibars (1000 HBAR or more). Call it twice so the second request is served from the Redis cache.
3. Before this change the cache-hit call returns HTTP 500; after it, the call returns the correct value.

N/A

N/A

### Checklist

### DIFF
--- a/src/relay/lib/bigNumberConfig.ts
+++ b/src/relay/lib/bigNumberConfig.ts
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import BigNumber from 'bignumber.js';
+
+/**
+ * Mirror Node REST responses can contain integers larger than JavaScript's safe
+ * range (e.g. the wei `value`/`amount` fields). `mirrorNodeClient` parses those
+ * with `json-bigint`, which turns them into `bignumber.js` `BigNumber` instances
+ * to preserve precision.
+ *
+ * By default bignumber.js renders a value whose exponent is >= 21 in exponential
+ * notation, so `new BigNumber('1e21').toString()` is `"1e+21"`. That bites us on
+ * the Redis cache path: `RedisCache` serializes cached Mirror Node objects with
+ * native `JSON.stringify` (which invokes `BigNumber.prototype.toJSON`) on set and
+ * `JSON.parse` on get, so the field comes back as the string `"1e+21"`. Formatters
+ * such as `nanOrNumberInt64To0x` then call `BigInt("1e+21")`, which throws and
+ * surfaces as an HTTP 500 on `eth_getTransactionByHash` / `debug_traceTransaction`.
+ * (The in-memory LRU cache stores objects by reference and never serializes, so it
+ * sidesteps this — the bug is Redis-only.)
+ *
+ * Setting `EXPONENTIAL_AT` to its maximum (1e9, the documented "never use
+ * exponential" value) makes `toString`/`toJSON` always emit full decimal digits, so
+ * the round-trip yields `"1000000000000000000000"` and `BigInt(...)` succeeds. EVM
+ * `value` is a uint256 (max ~1.16e77), so any threshold below ~78 would still break
+ * real values; 1e9 clears that range with a wide margin.
+ *
+ * IMPORTANT: this configures the bignumber.js instance that `json-bigint` uses. Two
+ * major versions are installed — `json-bigint` resolves the hoisted top-level
+ * `bignumber.js` (v9), while the Hedera SDK carries its own nested v10. `import
+ * BigNumber from 'bignumber.js'` resolves to the same v9 instance json-bigint
+ * `require`s, so the config takes effect; the SDK's v10 (used for Hbar/amount math)
+ * is left untouched, keeping the blast radius to Mirror Node response parsing.
+ *
+ * This module is imported for its side effect by `mirrorNodeClient`, guaranteeing
+ * the configuration is applied before the first response is parsed, on every
+ * entrypoint (HTTP, WS) and in tests.
+ */
+BigNumber.config({ EXPONENTIAL_AT: 1e9 });
--- a/src/relay/lib/clients/mirrorNodeClient.ts
+++ b/src/relay/lib/clients/mirrorNodeClient.ts
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
+// Side-effect import: configures the json-bigint bignumber.js instance before any response is parsed.
+import '../bigNumberConfig';
+
 import Axios, { type AxiosInstance, type AxiosRequestConfig, type AxiosResponse } from 'axios';
 import axiosRetry from 'axios-retry';
 import { install as betterLookupInstall } from 'better-lookup';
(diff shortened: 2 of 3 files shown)


quiet-node (approved):
LGTM

Ferparishuertas (approved):

PR #5499, opened 2026-06-22
perf: deduplicate resolveEvmAddress calls in block worker

### LINKED ISSUE #5264: perf: resolveEvmAddress deduplication
## Problem

  Within a single block, many transactions share the same `from`/`to` addresses
  (e.g. multiple calls to the same contract, or the same EOA sending multiple txs).

  `resolveEvmAddress` is called once per address per transaction with no
  deduplication. For a block with 50 transactions sharing 5 unique addresses,
  this results in up to 100 lookups for 5 distinct values.

  Although Redis caching prevents duplicate mirror node calls, the Redis roundtrips
  themselves are still serialized in `prepareTransactionArray` and add up.

  ## Impact

  - Redundant cache lookups proportional to transaction count, not unique address count
  - Compounds with the serialization issue in `prepareTransactionArray` (#5263)

  ## Proposed Fix

  - Collect unique addresses across all contract results
  - Resolve them once in parallel 
  - Map results back.

  Reduces lookups from O(2N transactions) to O(unique addresses)

ValentinVPK:
### Description
  
This PR reduces redundant `resolveEvmAddress` calls in `eth_getBlockByNumber`, `eth_getBlockByHash`, and `eth_getBlockReceipts`. Previously,
  `resolveEvmAddress` was called once per transaction for both `from` and `to` addresses. Since `resolveEvmAddress` internally calls `resolveEntityType` which is
  Redis-cached, duplicate addresses within the same block would not hit the mirror node again — but would still incur a Redis roundtrip per duplicate. This PR
  collects unique addresses into Sets first, resolves each unique address once in parallel via `Promise.all`, then performs synchronous Map lookups — eliminating
  redundant Redis roundtrips for duplicate addresses within the same block, and also reducing mirror node HTTP calls for addresses not yet cached. The shared
  deduplication logic is extracted into a `resolveContractResultAddresses` helper to avoid duplication across the two call sites.

  Additionally, `eth_getBlockReceipts` was missing the `[TYPE_ACCOUNT]` search type filter for `from` address resolution that
  `eth_getBlockByNumber`/`eth_getBlockByHash` already had. Since `from` in a signed transaction is always an EOA, restricting the lookup to account types avoids
  unnecessary contract and token lookups. This PR aligns `eth_getBlockReceipts` with the existing behavior.

  ### Related issue(s)

  Fixes #5264

  ### Testing Guide

  1. Run `npx ts-mocha --timeout 15000 'tests/relay/lib/eth/eth_getBlockByNumber.spec.ts' 'tests/relay/lib/eth/eth_getBlockByHash.spec.ts'
  'tests/relay/lib/eth/eth_getBlockReceipts.spec.ts' --exit` and verify all tests pass.
  2. For a block with N transactions sharing duplicate `from` or `to` addresses, verify the number of `resolveEvmAddress` calls equals the number of unique
  addresses rather than the total transaction count.
  3. Verify `eth_getBlockByNumber`, `eth_getBlockByHash`, and `eth_getBlockReceipts` return correct results with no change in response structure.

  N/A

  N/A


src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -294,20 +294,46 @@ async function getRootHash(receipts: IReceiptRootHash[]): Promise<string> {
   return prepend0x(Buffer.from(trie.root()).toString('hex'));
 }
 
+async function resolveContractResultAddresses(
```
quiet-node, line 316: let's add a tsdoc here so we can explain the method

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -294,20 +294,46 @@ async function getRootHash(receipts: IReceiptRootHash[]): Promise<string> {
...
-    [contractResult.from, contractResult.to] = await Promise.all([
-      commonService.resolveEvmAddress(contractResult.from, requestDetails, [constants.TYPE_ACCOUNT]),
-      commonService.resolveEvmAddress(contractResult.to, requestDetails),
-    ]);
-
+    contractResult.from = fromAddressMap.get(contractResult.from) ?? contractResult.from;
+    contractResult.to = contractResult.to
+      ? (toAddressMap.get(contractResult.to) ?? contractResult.to)
+      : contractResult.to;
     contractResult.chain_id = contractResult.chain_id || chain;
```
quiet-node: looks like the ternary is redundant. I think below works fine and cleaner

```suggestion
contractResult.to = toAddressMap.get(contractResult.to) ?? contractResult.to;
```

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -415,17 +440,14 @@ export async function getBlockReceipts(
...
-        ]);
+    const [fromAddressMap, toAddressMap] = await resolveContractResultAddresses(contractResults, requestDetails);
 
-        return { contractResult, logs, from, to };
-      }),
-    );
+    const resolved = contractResults.map((contractResult) => {
+      const logs = logsByHash.get(contractResult.hash) || [];
+      const from = fromAddressMap.get(contractResult.from) ?? contractResult.from;
+      const to = contractResult.to === null ? null : (toAddressMap.get(contractResult.to) ?? contractResult.to);
```
quiet-node: Same here ternary seems to be redundant.

```suggestion
const to = toAddressMap.get(contractResult.to) ?? contractResult.to;
```

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -294,20 +294,46 @@ async function getRootHash(receipts: IReceiptRootHash[]): Promise<string> {
...
+  for (const contractResult of contractResults) {
+    if (contractResult.from) uniqueFromAddresses.add(contractResult.from);
+    if (contractResult.to) uniqueToAddresses.add(contractResult.to);
+  }
+
+  return Promise.all([
+    Promise.all(
+      [...uniqueFromAddresses].map(
+        async (address) =>
+          [address, await commonService.resolveEvmAddress(address, requestDetails, [constants.TYPE_ACCOUNT])] as const,
```
quiet-node: Should we add a test assertion that specifically verifies [constants.TYPE_ACCOUNT] is passed for from address lookups in eth_getBlockReceipts? The current dedup test uses sinon.match.any for the third argument, so it does not confirm the new TYPE_ACCOUNT alignment. That would help lock in the behavioral change you documented in the PR description.

tests/relay/lib/eth/eth_getBlockByNumber.spec.ts
```diff
@@ -441,7 +441,7 @@ describe('@ethGetBlockByNumber using MirrorNode', async function () {
   });
 
   describe('eth_getBlockByNumber with tag', async function () {
-    const TOTAL_GET_CALLS_EXECUTED = 12;
+    const TOTAL_GET_CALLS_EXECUTED = 11;
```
quiet-node: Why change here?

quiet-node (commented):
Looking clean some suggestions and Qs

tests/relay/lib/eth/eth_getBlockByNumber.spec.ts
ValentinVPK: Because both transactions present in the `defaultContractResults` that are used in the test, share the same `from` address (LONG_ZERO_ADDRESS) reducing the total count of calls by 1

src/relay/lib/services/ethService/blockService/blockWorker.ts
ValentinVPK: Yes, good catch.

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -294,20 +294,56 @@ async function getRootHash(receipts: IReceiptRootHash[]): Promise<string> {
...
+  requestDetails: RequestDetails,
+): Promise<[Map<string, string>, Map<string, string>]> {
+  const uniqueFromAddresses = new Set<string>();
+  const uniqueToAddresses = new Set<string>();
+  for (const contractResult of contractResults) {
+    if (contractResult.from) uniqueFromAddresses.add(contractResult.from);
+    if (contractResult.to) uniqueToAddresses.add(contractResult.to);
+  }
+
+  return Promise.all([
```
simzzz: This fans out a Promise.all over every unique from and to address at once. In the getBlock path that replaces the old per-tx await loop, which kept it to about 2 lookups at a time. With showDetails=false there's no block size cap, so a big block on a cold cache could fire hundreds of resolveEntityType calls in parallel and overload the mirror node. Can we maybe chunk it? 

cc: @quiet-node

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -294,20 +294,56 @@ async function getRootHash(receipts: IReceiptRootHash[]): Promise<string> {
...
+}
+
 async function prepareTransactionArray(
   contractResults: any[],
   showDetails: boolean,
   requestDetails: RequestDetails,
   chain: string,
-  commonService: CommonService,
 ): Promise<Transaction[] | string[]> {
+  const [fromAddressMap, toAddressMap] = await resolveContractResultAddresses(contractResults, requestDetails);
```
simzzz: When showDetails is false we still resolve every from/to here, but downstream only contractResult.hash gets used, so these lookups are wasted on the most common eth_getBlockByNumber call. Can we skip resolveContractResultAddresses unless showDetails is true? That also removes the worst case of the parallel fan-out in the helper.

tests/relay/lib/eth/eth_getBlockReceipts.spec.ts
```diff
@@ -343,6 +350,47 @@ describe('@ethGetBlockReceipts using MirrorNode', async function () {
...
+        links: { next: null },
+      };
+
+      setupStandardResponses({
+        [CONTRACT_RESULTS_WITH_FILTER_URL_2]: threeTransactionResults,
+      });
+
+      const resolveEvmAddressStub = sinon
+        .stub(commonService, 'resolveEvmAddress')
+        .callsFake((address) => Promise.resolve(address));
```
simzzz, line 377: Since the stub just echoes the address back, this proves we dedupe the calls but not that the right resolved address ends up on the right tx. Could we return a distinct value per address (e.g. address + '-resolved') and assert from/to on the returned receipts? That would catch a map-keying mistake.

simzzz (commented):
Clean work overall, a couple of comments and suggestions

tests/relay/lib/eth/eth_getBlockReceipts.spec.ts
ValentinVPK: Done, added second test case.

src/relay/lib/services/ethService/blockService/blockWorker.ts
ValentinVPK: Done - the check is added. That also reduced the number of calls in the `eth_getBlockByNumber` from 11 to 4

src/relay/lib/services/ethService/blockService/blockWorker.ts
ValentinVPK: Good catch. I implemented a concurrency limiter capped at `MIRROR_NODE_HTTP_MAX_SOCKETS`, I think this is better than chunking since the next address is picked up immediately as slots free up, with no gaps between batches. The cap matches the maximum number of concurrent mirror node connections so we never exceed what the mirror node can handle.

simzzz (approved):

quiet-node (approved):
lgtm

PR #5504, opened 2026-06-23
feat: add typing for contract results

### LINKED ISSUE #4994: Add typing for MirrorNodeClient.getContractResults contract results response
### Problem

IContractResult in mirrorNode.ts is currently a very minimal interface that only exposes type and a loosely-typed entity object. This does not reflect the actual /contracts/results response from mirror node and provides no type-safety for fields like transaction_index, gas_used, hash, etc. 

As a result, MirrorNodeClient.getContractResults is effectively untyped (inferred as any/any[] via getPaginatedResults), and call sites such as blockWorker.loadBlockExecutionData and TransactionService have to work with any and raw property strings. 

This makes the code harder to understand and refactor, increases the risk of runtime errors when mirror node responses change, and is inconsistent with the stronger typing we use for other mirror-node client methods.

### Solution

Introduce a fully-typed IContractResult interface that models the real mirror node /contracts/results payload (including fields like transaction_index, gas_used, hash/transaction_hash, status, logs, contract ID, etc.), using optional properties where appropriate. 

Update MirrorNodeClient.getContractResults (and getPaginatedResults generics if needed) so it returns a properly typed Promise<IContractResult[]> (or the correct paginated envelope) instead of any. Then update consumers such as blockWorker and TransactionService to use IContractResult instead of any, removing ad‑hoc typing and stringly-typed property access, and add/adjust tests to cover the typed contract result shape and ensure we remain aligned with the mirror node API.

### Alternatives

_No response_

tgntr:
### Description

Adds typing to the Mirror Node contract-result responses. Replaces `any`/loose types on methods related to contract results with a proper interface hierarchy, and threads nullability (matching the Mirror Node API) through all consumers. No behavior change, except null `block_number`/`transaction_index` on `eth_getBlockReceipts` now serialize to 0x0 instead of throwing a `TypeError`; the single-receipt path is unchanged (still rejected upstream as an immature record).

**Types (`types/mirrorNode.ts`)**
- New hierarchy: `MirrorNodeContractResultBase` → `MirrorNodeContractResult` → `MirrorNodeContractResultDetails`; plus `MirrorNodeContractResultReceipt`, `MirrorNodeContractResultsPage`, `MirrorNodeContractResultLog`, `MirrorNodeContractResultStateChange`.
- Removed the dead `IContractResult` interface.
- Nullability aligned with the Mirror Node (`address`, `amount`, `gas_used`, `gas_limit`, `block_number`, `transaction_index`, `nonce`, `chain_id`, `gas_price`, `max_fee_per_gas`, `max_priority_fee_per_gas`, `r`, `s`, `type`, `v`, `failed_initcode`, `access_list`, `to`, `error_message`); `authorization_list?` moved to the base type.
- `ITransactionReceipt.contractAddress` → `string | null`.

**Client (`clients/mirrorNodeClient.ts`)**
- Typed returns: `getContractResults` → `MirrorNodeContractResult[]`; `getContractResult` / `getContractResultsDetails` / `getContractResultsByAddressAndTimestamp` → `MirrorNodeContractResultDetails | null`; `getContractResultsByAddress` / `getLatestContractResultsByAddress` → `MirrorNodeContractResultsPage | null`.
- Generics on `getPaginatedResults<T>` and `getContractResultWithRetry<T>`.

**Consumers (off `any`)**
- `blockWorker.ts`: typed `crPerTx` (`MirrorNodeContractResultReceipt`), `IReceiptRootHash.root` → `string | undefined`, `gas_used ?? 0` and `to === null` guards.
- `TransactionService.ts`: `gas_used` / `transaction_index` null guards in cumulative-gas handling.
- `transactionFactory.ts` / `transactionReceiptFactory.ts`: typed inputs; `getContractAddressFromReceipt` → `string | null`.
- `debug.ts`: `numberTo0x` → `nanOrNumberTo0x` for nullable `gasUsed`.
- `AccountService.ts`: `(nonce ?? 0) + 1` guard.
- `CommonService.ts` / `ICommonService.ts`: `getContractAddressFromReceipt` → `string | null`.

**Tests**
- `mirrorNodeClient.spec.ts` / `transactionFactory.spec.ts`: typed fixtures/casts, added HAPI null-field assertions.

### Related issue(s)

Fixes #4994

### Testing Guide

1. `npm run build` — clean compile.
2. `npm run lint` — ESLint clean.
3. `npm run test` — full test suite passes.

The issue proposed a new fully-typed `IContractResult` interface; instead this extends the existing `MirrorNodeContractResult` family (and removes the dead `IContractResult`), better matching the three distinct Mirror Node endpoint shapes.

`getContractAddressFromReceipt` on `CommonService` / `ICommonService` is dead code (no callers) and can be removed in a follow-up cleanup PR.

### Checklist


src/relay/lib/factories/transactionReceiptFactory.ts
```diff
@@ -106,7 +110,7 @@ class TransactionReceiptFactory {
     // Create the receipt object
     const receipt: ITransactionReceipt = {
       blockHash: toHash32(receiptResponse.block_hash),
-      blockNumber: numberTo0x(receiptResponse.block_number),
+      blockNumber: nanOrNumberTo0x(receiptResponse.block_number),
```
simzzz, line 113: Can you add a test that runs a receipt with null block_number/transaction_index through createRegularReceipt and checks for 0x0?
Otherwise if it gets swapped back to numberTo0x later for some reason the 500 quietly comes back.

simzzz (commented):
Looks good, added one suggestion for a test

simzzz (dismissed):

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -533,7 +538,7 @@ export class TransactionService implements ITransactionService {
     }
 
     const [resolvedToAddress, resolvedFromAddress] = await Promise.all([
-      this.common.resolveEvmAddress(contractResults[0].to, requestDetails),
+      contractResults[0].to === null ? null : this.common.resolveEvmAddress(contractResults[0].to, requestDetails),
```
quiet-node: Does it make sense to widen `CommonService.resolveEvmAddress` so it can accept `string | null` (it already handles a falsy address by returning early with `if (!address) return address`) and remove the repeated ternary, like `x.to === null ? null : resolveEvmAddress(...)`, from this spot and also from `blockWorker`? That would mirror the pattern we set up in `#5276` with `resolveAddress(string | null)`. If changing the method signature isn't possible for this PR, keeping the ternaries is totally fine.

src/relay/lib/factories/transactionReceiptFactory.ts
```diff
@@ -138,7 +142,7 @@ class TransactionReceiptFactory {
    * @param receiptResponse Mirror node contract result response
    * @returns {string} Contract address or null
```
quiet-node: lets update this return as well

quiet-node (commented):
LG couple of nits

quiet-node (approved):
LGTM great work!

natanasow (approved):

PR #5508, opened 2026-06-23
refactor: batch-fetch blocks in eth_feeHistory via MirrorNodeClient.getBlocksByRange

### LINKED ISSUE #5195: Refactor eth_feeHistory / FeeService.getFeeHistory to batch-fetch blocks via MirrorNodeClient.getBlocks (replace N× getBlock)
## Background
FeeService.getFeeHistory loads each block in range with repeated mirrorNodeClient.getBlock(blockNumber, …) calls (one HTTP request per block). A reviewer suggested using getBlocks once for the same range instead.

That change was deferred from the parent PR to avoid scope creep in #5066 .

## Constraints / why this is not a one-liner
1. getBlocks today issues a single get() with query params; it does not follow links.next for pagination. For ranges larger than the mirror default max page size, a naive single call may not return every block without pagination work.
2. Callers differ: some code expects the raw mirror payload (e.g. { blocks: [...], links: ... }), while shared pagination helpers elsewhere often normalize to a flat list of items. Any pagination change must stay compatible with all getBlocks callers (e.g. getLatestBlock, getBlock cache path, others) or introduce a dedicated API that does not break them.
3. Ordering / filtering: Fee history needs blocks for a contiguous numeric range (oldest → newest) with the same fields getFeeFromBlock / getGasUsedRatioForBlock need (timestamp, gas_used, hapi_version, etc.). The batch query must match that contract.

## Acceptance criteria
1. eth_feeHistory / getFeeHistory does not call getBlock once per block in the common case; it uses a batched mirror access path (or clearly documented fallback if the API cannot express the range).
2. Behavior matches current RPC output for baseFeePerGas, gasUsedRatio, oldestBlock, reward, and edge cases (e.g. newestBlockNumber + 1 “next fee” fetch if still required).
3. Tests updated or added: unit/integration coverage for multi-block fee history; regression for max-results cap and cache if applicable.
4. getBlocks pagination (if added) is covered by tests and documented for other callers; breaking changes to return shape are avoided or versioned.

quiet-node:
### Description

`eth_feeHistory` previously fetched its block range one block at a time, issuing an `mirrorNodeClient.getBlock` request per block in the range. This PR replaces that with a single batched fetch.

A new `MirrorNodeClient.getBlocksByRange(requestDetails, fromBlock, toBlock)` retrieves the inclusive range in one query (`block.number=gte:..&block.number=lte:..`), transparently following the mirror node `links.next` cursor so the full range is returned regardless of page size. `FeeService` indexes the result by block number and assembles the response from memory. RPC output is unchanged.

Pagination for the range is bounded by a new `FEE_HISTORY_BLOCK_PAGINATION_MAX` setting (default 20), independent of the global `MIRROR_NODE_PAGINATION_MAX`, so operators can size it for fee history alone. If the batched fetch fails, block resolution degrades to per-block lookups so the request still completes.

### Related issue(s)

Fixes #5195

### Testing Guide

1. Run the fee history unit/integration suites: `tests/relay/lib/services/feeService/FeeService.spec.ts`, `tests/relay/lib/eth/eth_feeHistory.spec.ts`, `tests/relay/lib/mirrorNodeClient.spec.ts`.
2. Call `eth_feeHistory` for a multi-block range and confirm `baseFeePerGas`, `gasUsedRatio`, `oldestBlock`, and `reward` match the prior behavior.
3. Confirm a single `/blocks?block.number=gte:..&block.number=lte:..` request is issued for the range rather than one request per block.

N/A

The per-block gas price lookup (`getGasPriceInWeibars`) is still one call per block; batching it is out of scope for this issue.

### Checklist


src/relay/lib/clients/mirrorNodeClient.ts
```diff
@@ -814,6 +814,42 @@ export class MirrorNodeClient {
...
+    fromBlock: number,
+    toBlock: number,
+  ): Promise<MirrorNodeBlock[]> {
+    const queryParamObject = {};
+    this.setQueryParam(queryParamObject, 'block.number', [`gte:${fromBlock}`, `lte:${toBlock}`]);
+    this.setQueryParam(queryParamObject, 'limit', ConfigService.get('MIRROR_NODE_LIMIT_PARAM'));
+    this.setQueryParam(queryParamObject, 'order', MirrorNodeClient.ORDER.ASC);
+    const queryParams = this.getQueryParams(queryParamObject);
+
+    return this.getPaginatedResults(
```
simzzz: One thing we lose by going through getPaginatedResults: getBlock caches each block under GET_BLOCK.{n}, and that cache is shared with eth_getBlockByNumber, AccountService, CommonService, and others. 

This path never reads or writes it. feeHistory(latest) also isn't response-cached (latest is in NON_CACHABLE_BLOCK_PARAMS), so back-to-back calls now re-fetch the whole range every time instead of reusing blocks we already pulled. 

Could we write the returned blocks into the GET_BLOCK cache here so the rest of the system still benefits?

src/relay/lib/clients/mirrorNodeClient.ts
```diff
@@ -814,6 +814,42 @@ export class MirrorNodeClient {
...
+      `${MirrorNodeClient.GET_BLOCKS_ENDPOINT}${queryParams}`,
+      MirrorNodeClient.GET_BLOCKS_ENDPOINT,
+      'blocks',
+      requestDetails,
+      [],
+      1,
+      ConfigService.get('FEE_HISTORY_BLOCK_PAGINATION_MAX'),
```
simzzz, line 849: If getBlocksByRange comes back empty or short for a range that actually has blocks
(mirror lagging, mid-reorg, or a node that ignores the gte/lte pair), every missing block falls through to a single getBlock in getFeeHistoryDataFromBlock. That's up to N+1 requests, more than we issued before this PR, and it lands right when the mirror node is already struggling. 

Could we fall back per-block only when the batch actually threw, not when it just returned fewer rows than the range?

src/relay/lib/services/ethService/feeService/FeeService.ts
```diff
@@ -192,9 +192,14 @@ export class FeeService implements IFeeService {
       oldestBlock: numberTo0x(oldestBlockNumber),
     };
 
-    // get fees from oldest to newest blocks
+    const blocksByNumber = await this.getBlocksInRange(oldestBlockNumber, newestBlockNumber, requestDetails);
```
simzzz: nit: since we're batching now, in the latest > newest branch further down we still do a separate getBlock for newest+1 after the loop. Could we request [oldest, newest+1] in that case and fold it into the batch? 

Saves a round-trip whenever someone passes an explicit (non-latest) newestBlock.

simzzz (commented):
Looks good, left a few suggestions

simzzz (dismissed):

src/relay/lib/clients/mirrorNodeClient.ts
quiet-node: Another good point however I decided to keep the per-block fallback on map miss rather than just on throw, because in the partial-response scenario (where batch succeeds but returns fewer blocks than the range), it could mean those blocks actually exist, mirror node just didn't return them. Falling back with getBlock gives us a real shot at correct data. If I skipped it and returned zeros, we'd silently serve wrong gas history. The extra request is worth it for correctness. The N+1 overhead only shows up when the batch returns completely empty without throwing, which is the rarest of the degraded-MN cases.

src/relay/lib/clients/mirrorNodeClient.ts
quiet-node: Great catch. Added cache in getBlocksInRange

src/relay/lib/services/ethService/feeService/FeeService.ts
quiet-node: good point. Fixed!

simzzz (approved):

natanasow (approved):

PR #5514, opened 2026-06-25
fix: remove RLP access list decode workaround for MirrorNode v0.156+

### LINKED ISSUE #5293: [Track] MN tracking: Return access list as array of objects in contract results
### Problem

Currently MirrorNode contract/results endoint returns access list as a rlp encoded hex string.
This format is:

 - difficult to read (requires end-client to decode it),
- incompatible with the format used for authorization list.

FOr that reason it would be the best if response format was changed on the MirrorNode side.

### Solution

For now we added support for both formats of this field (encoded and unencoded). But if MN will start returning unencoded response we can remove encoded support from our code base.

Tracked issue: https://github.com/hiero-ledger/hiero-mirror-node/issues/13343

PR introducing the fixme note that has to be removed after this issue is done; 
https://github.com/hiero-ledger/hiero-json-rpc-relay/pull/5271/changes

### Alternatives

_No response_

ValentinVPK:
### Description

  This PR removes the RLP access list decode workaround introduced in #5271. MirrorNode v0.156+ (hiero-ledger/hiero-mirror-node#13571) now returns `access_list` as a plain JSON array instead of an RLP-encoded hex
  string, making the decode machinery unnecessary.

  The `formatAccessList` function in `transactionFactory.ts` is simplified to directly consume the new array format, mapping MirrorNode's snake_case `storage_keys` field to the camelCase `storageKeys` required by the
  Ethereum JSON-RPC spec.

  The `MirrorNodeContractResult.access_list` type in `mirrorNode.ts` is updated from `string` to `{ address: string; storage_keys: string[] }[] | null` to reflect the current MirrorNode API contract.

  All test fixture mock objects that previously used `'0x'` (the old RLP-encoded empty access list) have been updated to `[]` to match the format MirrorNode v0.156+ actually returns.

  ### Related issue(s)

  Fixes #5293

  ### Testing Guide

  1. Build and start the relay against testnet:
  ```bash
  npm run build && HEDERA_NETWORK=testnet MIRROR_NODE_URL=https://testnet.mirrornode.hedera.com npm start

  2. Submit an EIP-2930 (type 1) transaction with a non-empty access list via eth_sendRawTransaction using a funded testnet ECDSA account.
  3. Confirm MirrorNode returns access_list as a JSON array:
  curl -s "https://testnet.mirrornode.hedera.com/api/v1/contracts/results/<TX_HASH>"

  4. Confirm the relay returns accessList with camelCase storageKeys:
  curl -s -X POST http://localhost:7546 \
    -d '{"jsonrpc":"2.0","method":"eth_getTransactionByHash","params":["<TX_HASH>"],"id":1}'

  Changes from original design (optional)

  N/A

  Additional work needed (optional)

  N/A


simzzz (approved):

src/relay/lib/types/mirrorNode.ts
```diff
@@ -260,8 +260,8 @@ export interface MirrorNodeContractResult {
   status: string;
   /** The failed contract init code, if applicable, otherwise null. */
   failed_initcode: string | null;
-  /** The access list for EIP-2930 transactions. */
-  access_list: string;
+  /** The access list for EIP-2930 transactions. MirrorNode v0.156+ returns this as an array of objects. */
```
quiet-node, line 263: Since this makes v0.156+ a hard requirement for correct access_list handling, should we also bump the documented floor in docs/configuration.md (still says v0.151.0)? An operator on a docs-compliant MN v0.151-v0.155 would get the RLP-hex form, which the new formatAccessList treats as a non-array and returns []. That silently drops the access list rather than warning. Would it make sense to update the doc floor (and maybe the relay.ts startup warning) as part of this change, or capture it in the pectra release notes?

src/relay/lib/factories/transactionFactory.ts
```diff
@@ -121,78 +117,22 @@ const formatAuthorizationList = (authorizationList: unknown): AuthorizationListE
...
-  if (typeof accessList === 'string' && isHex(accessList)) return formatEncodedAccessList(accessList);
-  return accessList && Array.isArray(accessList)
+const formatAccessList = (accessList: unknown): AccessListEntry[] =>
+  accessList && Array.isArray(accessList)
     ? accessList
         .filter((item: unknown) => item !== null && typeof item === 'object')
         .map((item: any) => ({
           address: formatAddress(item.address),
-          storageKeys: item.storageKeys ?? [],
+          storageKeys: item.storage_keys ?? [],
```
quiet-node, line 134: The old RLP path padded each storage key to 32 bytes; the new path trusts MN to send them pre-padded. That matches the testnet verification, so this looks correct. Would it make sense to add a one-line comment that MN v0.156+ guarantees 32-byte padded keys? Downstream ethers serialization asserts isHexString(key, 32) and throws "invalid slot" otherwise, so it would help future readers know why we intentionally skip normalization here.

quiet-node (commented):
Couple of items! Great work overall!

Ferparishuertas (approved):

PR #5526, opened 2026-07-03
perf: introduce worker context to reduce duplicate client instances

### LINKED ISSUE #5257: perf: Reduce resource duplication caused by per-worker MirrorNodeClient and CacheService instantiation
## Summary

Each worker thread independently instantiates a full `MirrorNodeClient`, `CacheService`, and `Registry`. With the default pool of 2–4 workers and 3 worker modules (`blockWorker`, `commonWorker`, `accountWorker`), the relay creates up to 13 independent MirrorNodeClient instances at peak, each with its own HTTP/HTTPS connection pools and Axios clients.

## Current Behavior

Worker modules are loaded once per worker thread. Their module-level initialization code runs in full isolation, so every worker executes:

```
// blockWorker.ts, commonWorker.ts, accountWorker.ts — identical pattern
const cacheService = CacheClientFactory.create(logger, register);
const mirrorNodeClient = new MirrorNodeClient(MIRROR_NODE_URL, logger, register, cacheService);
```

All three worker modules - `accountWorker`, `blockWorker`, and `commonWorker`, are statically imported at the top of workers.ts (lines 4–6). Because they are static imports, every Piscina worker thread loads all three modules at startup. Each module has top-level (module-scope) code that independently instantiates a `MirrorNodeClient`.

Per worker thread, this allocates 3× MirrorNodeClient instances (one per module), each with:
- 2× Axios instances (REST + Web3)
- 4× http.Agent / https.Agent instances, each maintaining an independent TCP socket pool
- 1× local LRU cache (when Redis is disabled)

With the default pool config (`WORKERS_POOL_MIN_THREADS=2`, `WORKERS_POOL_MAX_THREADS=4`):

Thread count | MirrorNodeClient instances | HTTP agents
-- | -- | --
2 workers | 7 (1 main + 2×3) | 28
4 workers | 13 (1 main + 4×3) | 52

### Inflated TCP connection count to Mirror Node

Each HTTP agent pool is independent, so effective max outbound connections scale as MIRROR_NODE_HTTP_MAX_SOCKETS × (number of agents). With 13 instances this multiplies the intended socket budget by ~13×, which can exceed upstream rate limits or trigger ephemeral port exhaustion under load.

### Silent metric loss within each worker thread

Each MirrorNodeClient constructor calls this.register.removeSingleMetric() before registering (mirrorNodeClient.ts:288,298). Because all three statically-imported worker modules share the same thread-local Registry singleton, each successive constructor silently de-registers the previous module's metrics and replaces them. The first two modules' clients retain references to orphaned, unregistered Histogram/Counter objects, meaning observations recorded against them are silently dropped and never reach the main thread's registry.

### Redundant initialization cost per worker

Each thread re-establishes HTTP keep-alive connections, sets up TLS, and parses the same configuration. This is work that is identical across all workers and could be avoided.

### Proposed Approach

Node.js worker threads run in isolated V8 heaps, objects backed by native handles (sockets) cannot be shared by reference across threads. The practical fix is to reduce from 3 clients per thread to 1.

Create a single MirrorNodeClient + CacheService once in the Piscina entry file (workers.ts) and pass the instances as parameters into each worker module's exported functions. Module-level code in workers.ts runs exactly once per thread at first import.

This approach:
- Reduces instances: 3N+1 → N+1 (13 → 5 for 4 workers)
- Fixes the metric clobber bug: one client per thread means no removeSingleMetric conflicts between modules
- Worker modules (`blockWorker`, `commonWorker`, `accountWorker`) become pure functions rather than owning hidden module-level singletons
- Keeping workers alive via WORKERS_POOL_IDLE_TIMEOUT_MS (default 5 min) amortizes the remaining per-thread initialization cost across many requests

tgntr:
### Description

This PR reduces resource duplication in the worker pool. Previously each of the three worker modules (`accountWorker`, `blockWorker`, `commonWorker`) instantiated its own `MirrorNodeClient`, `CacheService`, and `Registry` at module load, producing a lot of `MirrorNodeClient` instances (each with its own socket pools) at peak. This consolidates construction into a single `createWorkerContext` builder that wires one client/cache set per execution context, plus a `getWorkerContext` cache that reuses it — one per worker thread, and the relay's own client/cache on the main thread in local mode (`WORKERS_POOL_ENABLED=false`). Worker functions become pure and receive the context as a parameter.

Because the builder now runs on the main thread in local mode, the throwaway `TransactionPoolService` (required by `AccountService`'s constructor but unused for these tasks) is given a `new Registry()` rather than the shared one, so its remove-then-register of txpool metrics can't clobber the relay's live metrics on the scraped registry.

### Related issue(s)

Fixes #5257

### Testing Guide

1.`npm run build-and-test` — project builds and all tests pass.

N/A

N/A

### Checklist

### DIFF
--- a/src/relay/lib/services/workersService/workerContext.ts
+++ b/src/relay/lib/services/workersService/workerContext.ts
@@ -24,9 +24,6 @@ export interface WorkerContext {
   accountService: AccountService;
 }
 
-/** Created once per thread and shared by every context on that thread. */
-const logger: Logger = pino({ level: ConfigService.get('LOG_LEVEL') || 'trace' });
-
 /**
  * Single per-thread cache for the worker context.
  */
@@ -51,6 +48,7 @@ export function createWorkerContext(
   transactionPoolService?: TransactionPoolService,
   accountService?: AccountService,
 ): WorkerContext {
+  const logger: Logger = pino({ level: ConfigService.get('LOG_LEVEL') || 'trace' });
   const register = RegistryFactory.getInstance();
   if (!cacheService) {
     cacheService = CacheClientFactory.create(logger, register);
--- a/src/relay/lib/services/workersService/workerContext.ts
+++ b/src/relay/lib/services/workersService/workerContext.ts
@@ -16,7 +16,7 @@ import { TransactionPoolService } from '../transactionPoolService/transactionPoo
 /**
  * Clients and services shared by every worker task within one execution context.
  */
-export interface WorkerContext {
+export interface IWorkerContext {
   logger: Logger;
   cacheService: ICacheClient;
   mirrorNodeClient: MirrorNodeClient;
@@ -27,10 +27,10 @@ export interface WorkerContext {
 /**
  * Single per-thread cache for the worker context.
  */
-let cachedContext: WorkerContext | null = null;
+let cachedContext: IWorkerContext | null = null;
 
 /**
- * Creates a {@link WorkerContext}. Each argument is reused if supplied and instantiated otherwise, so
+ * Creates a {@link IWorkerContext}. Each argument is reused if supplied and instantiated otherwise, so
  * callers pass whatever they want to share (e.g. the relay's client in local mode); with no client, a
  * self-contained context is built around a fresh one.
  *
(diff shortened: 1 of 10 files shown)


src/relay/lib/services/workersService/workerContext.ts
```diff
@@ -0,0 +1,109 @@
...
+export interface WorkerContext {
+  logger: Logger;
+  cacheService: ICacheClient;
+  mirrorNodeClient: MirrorNodeClient;
+  commonService: CommonService;
+  accountService: AccountService;
+}
+
+/** Created once per thread and shared by every context on that thread. */
+const logger: Logger = pino({ level: ConfigService.get('LOG_LEVEL') || 'trace' });
```
natanasow: This runs on import time, but we don't need it if `WORKERS_POOL_ENABLED=false`.

We can just move the initialization inside `createWorkerContext`,

Ferparishuertas (approved):

natanasow (approved):

PR #5527, opened 2026-07-06
feat: compute baseFeePerGas as gas-used-weighted average of block tx prices

### LINKED ISSUE #5130: baseFeeperGas need to be a calculated value
x## Problem

baseFeePerGas in Ethereum is the fee applied at block level. On Hiero transactions can have different gasPrice within the same block.
Added to this, standard ethereum tooling uses latest baseFeePerGas on ethFeeHistory, to calculate next gasPrice.

A wrong calculated value can lead to transaction sending problem within tools, if value set is lower than network gasprice, or value 0 in empty block

Added to this, nowadays we set the value to the latest gasprice, indepently if the block information is the latest or historical.

## Solution

Implement  the following

- eth_getBlockByNumber or hash: The value for baseFeePerGas will be the average gasPrice applied. having known it should be a value, that multiplied to the total gasused its equal to totall charge hbar ( PLEASE review possible decimals issues, rounding) . In case of any empty block, the value will be the price of the blockNumber latest timestamp, echange value of GET
[/api/v1/network/fees](https://mainnet.mirrornode.hedera.com/api/v1/docs/#/network/getNetworkFees)

- eth_feeHistory. If latest is passed as param, latest = this moment gas price, to avoid any race condition. On other blocks we should answer with the latest gasprice applied, within the block ( latest transaction within the block). In case of any empty block, the value will be the price of the blockNumber latest timestamp, echange value of GET
[/api/v1/network/fees](https://mainnet.mirrornode.hedera.com/api/v1/docs/#/network/getNetworkFees)

ValentinVPK:
### Description
This PR fixes an issue where `baseFeePerGas` in `eth_getBlockByNumber`, `eth_getBlockByHash`, and
  `eth_feeHistory` always returned the governance-set fee-schedule floor rate, regardless of what
  transactions in the block actually paid.

  The fee schedule is a rejection floor only — transactions can and do pay above the minimum. This PR
  computes `baseFeePerGas` as the gas-used-weighted average of each transaction's `gas_price` (sourced from
  the mirror node's contract results, in tinybars, converted to weibars). When a block is empty or all
  contract results have a null `gas_price`, it falls back to the fee-schedule rate at the block's closing
  timestamp, preserving the previous behaviour for those cases.

  **Key changes:**
  - `blockWorker.ts` — adds `computeBlockGasPrice`, which computes the weighted average and replaces the
  previous flat `gasPrice` call for `eth_getBlockByNumber` / `eth_getBlockByHash`.
  - `mirrorNodeClient.ts` — adds `getLatestContractResultForBlock`, a single non-paginated GET
  (`limit=1&order=desc`) scoped to a block's timestamp range. Used by `FeeService` to retrieve the most
  recent transaction's gas price for `eth_feeHistory` without making N mirror node calls for an
  N-transaction block.
  - `FeeService.ts` — `getFeeHistoryDataFromBlock` now uses `getLatestContractResultForBlock` instead of
  always querying the fee schedule.
  
  ### Related issue(s)

  Fixes #5130

  ### Testing Guide

  1. Start the relay locally pointing at mainnet or testnet.
  2. Call `eth_getBlockByNumber` for a block known to contain `EthereumTransaction` entries and verify
  `baseFeePerGas` reflects the actual gas prices paid rather than always the floor rate.
  3. Call `eth_feeHistory` with `"0x5"` as block count and `"latest"` as newest block — verify
  `baseFeePerGas` is populated for all blocks and falls back gracefully to the fee-schedule rate for empty
  blocks.   

  ```bash
  curl -s -X POST http://localhost:7546 \
    -d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["latest",false],"id":1}' \
    | jq '.result.baseFeePerGas'

  curl -s -X POST http://localhost:7546 \
    -d '{"jsonrpc":"2.0","method":"eth_feeHistory","params":["0x5","latest",[]],"id":1}' \
    | jq '.result.baseFeePerGas'

Changes from original design (optional)

 N/A


Ferparishuertas (dismissed):

Ferparishuertas (dismissed):

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -365,6 +365,48 @@ async function prepareTransactionArray(
...
+  let totalGasUsed = 0;
+  for (const cr of validResults) {
+    const priceTinybars = parseInt(cr.gas_price!, 16);
+    const gasUsed = cr.gas_used!;
+    weightedSum += priceTinybars * gasUsed;
+    totalGasUsed += gasUsed;
+  }
+
+  const weightedAvgTinybars = Math.round(weightedSum / totalGasUsed);
+  return numberTo0x(weightedAvgTinybars * constants.TINYBAR_TO_WEIBAR_COEF);
```
tgntr: based on docs for `withHbarDisabled` function when Hbar is disabled the gas price is returned in weibars, probably worth double-checking whether we need to convert here

src/relay/lib/services/ethService/blockService/blockWorker.ts
```diff
@@ -365,6 +365,48 @@ async function prepareTransactionArray(
...
+  if (validResults.length === 0) {
+    return numberTo0x(await commonService.getGasPriceInWeibars(requestDetails, `lte:${blockTimestampTo}`));
+  }
+
+  let weightedSum = 0;
+  let totalGasUsed = 0;
+  for (const cr of validResults) {
+    const priceTinybars = parseInt(cr.gas_price!, 16);
+    const gasUsed = cr.gas_used!;
+    weightedSum += priceTinybars * gasUsed;
```
tgntr: should we use BigInt here to avoid overflow?

tgntr (approved):

natanasow (approved):

BartoszSolkaBD (approved):

PR #5531, opened 2026-07-08
chore: redact sensitive values in config and Redis error logging

ValentinVPK:
### Description

Improve Redis cache configuration and error logging.

  ### Related issue(s)

 #5535 

  ### Testing Guide

 Run the config-service unit tests: `npm run test`.

  N/A

  N/A


src/relay/lib/errors/RedisCacheError.ts
```diff
@@ -1,13 +1,24 @@
...
+ * Redacts credentials embedded in connection-string style URLs within a text value.
+ * Any userinfo segment between `scheme://` and `@` is replaced with `***`, so
+ * `redis://user:pass@host:6379` becomes `redis://***@host:6379`. Values without an
+ * embedded `user:pass@` segment (e.g. `redis://host:6379`) are returned unchanged.
+ *
+ * @param value - Arbitrary text that may contain one or more URLs with embedded credentials.
+ * @returns The text with any URL credentials redacted.
+ */
+export function redactUrlCredentials(value: string): string {
+  return value.replace(/([a-z][a-z0-9+.-]*:\/\/)[^\s/@]*@/gi, '$1***@');
```
tgntr: currently this regex stops at the first @, so for example `redis://user:p@ss@host` redacts to `redis://**@ss@host` and still leaks part of the password. should we drop the @ from the character class — `[^\s/@]*@` → `[^\s/]*@` so it consumes up to the last @ before the host and redacts the whole credential segment?

tgntr (commented):
LGTM, only one minor comment.

should we add a test to verify masked error logging?

src/relay/lib/errors/RedisCacheError.ts
ValentinVPK: Good catch. Fixed

tgntr (approved):

natanasow (approved):

BartoszSolkaBD (approved):

PR #5551, opened 2026-07-20
chore(ci): automate Solo, consensus node, and mirror node version updates

### LINKED ISSUE #5235: Automate Solo, consensus node, and mirror node version updates in CI
## Problem

With the migration from LocalNode to Solo in #5234, we are carrying forward a long-standing problem: there is no automated pipeline to keep the consensus node (CN) and mirror node (MN) versions up to date.

Previously with LocalNode, and now with Solo (via `falcon.yml`), the CN and MN versions are pinned in configuration files. These versions stay unchanged for extended periods, sometimes spanning several releases, until someone manually bumps them. When that manual upgrade finally happens, a large number of tests tend to break because of the accumulated gap between the versions we test against and the latest releases.

For reference, the current Solo configuration in `.github/falcon.yml` and the workflow files pin:

- Solo version: `0.68.0` (hardcoded in each workflow's `npx @hashgraph/solo@...` command)
- Network release tag: `v0.72.0`
- Mirror node version: `v0.151.0`

These will quickly fall behind as new versions are published.

## Proposal

Set up an automated job (e.g., a weekly cron-triggered GitHub Actions workflow) that:

1. Checks for the latest published Solo, CN, and MN versions
2. Updates the pinned versions in `.github/falcon.yml`, the workflow files (Solo version), and any other relevant config files
3. Opens a PR with the version bumps -> lets CI run the tests against the new versions

This would allow us to:

- Catch breaking changes early, while they are small and easier to debug
- Stay in sync with CN and MN releases instead of falling several versions behind
- Reduce the cost of upgrades by making them incremental rather than large jumps

ValentinVPK:
### Description

This PR introduces automated management of the Solo / consensus-node / mirror-node versions that our CI test network is pinned to.

Today those three versions are hardcoded and duplicated across every workflow that spins up a Solo-based local network, and keeping them current means editing each file by hand, which is error-prone and tends to lag behind upstream releases.
  
This PR does two things:

1. **Centralizes** the three pins into a single source of truth, `.github/network-versions.env`. Every workflow that deploys the local network now loads that file into `$GITHUB_ENV` (a "Load pinned versions" step) and references `env.SOLO_VERSION` / `env.NETWORK_TAG` / `env.MIRROR_NODE_TAG`. Reusable workflows keep their existing `inputs.X || env.X` override so callers (e.g. `manual-testing.yml`) can still target a specific version. 
2. **Adds a weekly bot** (`.github/workflows/bump-network-versions.yml`) that runs `.github/scripts/bump-network-versions.sh` to fetch the latest stable releases (Solo via the npm `latest` dist-tag; CN/MN via GitHub's `releases/latest`), rewrites only the changed lines in `network-versions.env`, and opens a PR. Because the PR is opened with a PAT, the full CI suite runs against the new versions **before** anything is merged, so a developer reviews and CI validates every bump.

### Related issue(s)

Fixes #5235

### Testing Guide

1. **Run the bump script locally against a throwaway copy** (real file is never touched thanks to the `VERSIONS_FILE` override):
```bash
     cp .github/network-versions.env /tmp/versions-test.txt
     VERSIONS_FILE=/tmp/versions-test.txt .github/scripts/bump-network-versions.sh
```    
Expected: it prints `updated …`/`current …` per version, rewrites only the value lines when a newer release exists, and reports "No version changes" on a second run. `git status` shows the real file unchanged.
2. **Confirm the refactor is inert:** the network-deploying workflows on this PR (acceptance, conformity, hoppscotch, dev-tool, release-acceptance, subgraph, dapp) run against the same versions as before and pass; the only change is where the numbers come from.
3. **After merge:** the `Bump Solo, Consensus Node, and Mirror Node Versions` workflow can be triggered manually from the Actions tab (`workflow_dispatch`) and runs automatically every Monday. When versions are behind, it opens/updates a PR on the `chore/bump-network-versions` branch; when everything is current, it does nothing.

N/A

N/A


tgntr (dismissed):

.github/network-versions.env
```diff
@@ -0,0 +1,15 @@
...
+#
+# Kept up to date by the automated bump-network-versions workflow:
+# the bot edits ONLY this file and opens a PR so CI runs against the new versions.
+#
+#   SOLO_VERSION    - @hashgraph/solo npm package version (no leading "v")
+#   NETWORK_TAG     - hiero-consensus-node release tag
+#   MIRROR_NODE_TAG - hiero-mirror-node release tag
+SOLO_VERSION=0.85.0
+NETWORK_TAG=v0.75.1
+MIRROR_NODE_TAG=v0.159.0
```
natanasow: nit: new line

.github/scripts/bump-network-versions.sh
```diff
@@ -0,0 +1,83 @@
...
+printf 'Changes:\n%s\n' "$summary"
+
+emit "changed=true"
+if [[ -n "${GITHUB_OUTPUT:-}" ]]; then
+  {
+    echo "summary<<SUMMARY_EOF"
+    echo "$summary"
+    echo "SUMMARY_EOF"
+  } >> "$GITHUB_OUTPUT"
+fi
```
natanasow: nit: new line

natanasow (dismissed):

natanasow (approved):

natanasow (dismissed):

Ferparishuertas (dismissed):

Ferparishuertas (approved):

natanasow (approved):

tgntr (approved):

natanasow (approved):

andrewb1269 (approved):

PR #5579, opened 2026-07-27
feat: opt-in transaction status tracing for eth_getTransactionReceipt

### LINKED ISSUE #5447: User Story: Improve error tracing for asynchronous transaction processing in Hiero JSON-RPC Relay
**As a** developer or dApp integrator using the Hiero JSON-RPC Relay (especially with `USE_ASYNC_TX_PROCESSING=true` or in scenarios with connection loss during submission),  
**I want** detailed, actionable error information for transactions that are rejected during precheck or at consensus time (or that time out), even when `eth_getTransactionReceipt` is queried after the initial fast `eth_sendRawTransaction` response,  
**so that** I can reliably debug failures, handle errors programmatically (e.g. retry logic, user feedback), and maintain a consistent Ethereum-like developer experience without losing critical Consensus Node (CN) or precheck error details that would otherwise be invisible in async mode.

---

### 1. Configuration & Feature Flag
- New config: `TX_STATUS_TRACING` (default: `false` – fully deactivated for backward compatibility and minimal overhead).
- When `true`, enable the tracing storage (independent of `USE_ASYNC_TX_PROCESSING` or `ENABLE_TX_POOL`).
- Optional: `TX_STATUS_TRACING_TTL_MINUTES` (default: `15`; `0` or `-1` for eternal/no cleanup).
- Storage backend: Follow the **exact same pluggable pattern** as the existing txpool (in-memory LRU by default; Redis-backed when `REDIS_ENABLED=true`).
- Implement as a **new, separate storage module** (`TransactionTracingStorage`) for extensibility (future audit logs, metrics, custom backends).
- No breaking changes when `TX_STATUS_TRACING=false`.

### 2. Transaction Status Tracking
The storage maps `transactionHash` → rlp encoded + status object (with timestamp, full error details, Hedera status codes).  
Status transitions (updated in the existing send/processing pipeline):

- `pending` – if txpool is enabled and tx is queued  
- `sent` – updated when successfully forwarded to Consensus Node  
- `rejected` – if any error is received from precheck, CN submission, or Mirror Node ingress (store full error payload + Hedera status code, e.g. `WRONG_NONCE`, `INVALID_ACCOUNT_ID`, `THROTTLED_AT_CONSENSUS`)  
- `timedout` – if a submission/processing timeout occurs  
- `validated` – once the transaction is successfully marked as validated/executed. If TTL is not eternal, clean up the entry.

### 3. Behavior on `eth_getTransactionReceipt` (core improvement)
- Normal flow unchanged: Query the Mirror Node first.
- If MN returns a receipt → return it as usual (`status: "0x1"` or `"0x0"`).
- If MN responds “not found”:
  - Check the tracing storage for the hash.
  - If status = `rejected` or `timedout` → **return a JSON-RPC error** (recommended – see Research section):
    ```json
    {
      "jsonrpc": "2.0",
      "id": 1,
      "error": {
        "code": -32003,
        "message": "Transaction rejected",
        "data": {
          "status": "WRONG_NONCE",
          "details": "Nonce too low or already used",
          "timestamp": "2026-05-28T15:40:00Z"
        }
      }
    }

If status = pending/sent → return null (standard pending behavior).
If status = validated but MN still doesn’t have it → fall back to null.

### 4. Admin / Debug API Exposure (when TX_STATUS_TRACING=true)

- New hedera specific admin methods (under existing admin namespace):
- admin_getTransactionTraceByHash(hash) – returns full stored status/details for a single tx hash.
- admin_getTransactionTraceById(id) – returns full stored status/details for a single tx id.

Only exposed/functional when TX_STATUS_TRACING=true.

### 5. Cleanup & Performance

- Automatic cleanup of validated entries after TTL (or immediately if eternal).
- Lightweight in-memory implementation (hashmap + TTL expiry).
- Redis version reuses exact same patterns as txpool.
- No performance impact on hot paths when feature is disabled.

### 6. Testing / Edge Cases

- Covers connection loss during async processing.
- Backward compatible with sync mode.
- Works with/without txpool and Redis.
- Full unit + integration test coverage for status transitions and receipt fallback.
- Update documentation (configuration.md, rpc-api.md, debugging guide).

tgntr:
### Description

Adds an opt-in tracing subsystem (`TX_STATUS_TRACING`, default off) that records each submitted transaction's lifecycle keyed by keccak256 hash. When enabled, a rejected or timed-out transaction surfaces a `-32003` error via `eth_getTransactionReceipt` (with `txHash` / `detail` / `hederaStatus` in `data`; `timedout` is marked `provisional`) instead of an indistinguishable `null`. Storage is the in-memory LRU by default, Redis when enabled. `admin_getTransactionTraceByHash` exposes raw records for inspection. With the flag off, nothing is stored and there is no hot-path change.

### Related issue(s)

Fixes #5447

### Testing Guide

1. `npm test` — tracing unit/integration specs (service, storage, receipt fallback, record sites, admin).

Note: clients only decode the `-32003` when the relay returns HTTP 200 for valid JSON-RPC errors
(`ON_VALID_JSON_RPC_HTTP_RESPONSE_STATUS_CODE=200`); at the default 400, ethers surfaces an opaque
transport error. See `docs/debugging-transactions.md`.

N/A

N/A

### Checklist

- [x] I've updated documentation (`configuration.md`, `rpc-api.md`, `debugging-transactions.md`)
- [x] I've done sufficient testing (unit, integration, client-library)


natanasow:
Reviewing it partially due to the size of the PR.

natanasow:
LGTM so far. Can we add unit tests to bump the coverage and make Codecov pass?

src/relay/lib/types/transactionTracing.ts
```diff
@@ -0,0 +1,131 @@
...
+   * @returns The record, or null if none exists.
+   */
+  get(hash: string): Promise<TransactionTraceRecord | null>;
+
+  /**
+   * Removes the record for the given hash.
+   *
+   * @param hash - The transaction's keccak256 hash (lower-cased 0x-prefixed hex).
+   */
+  delete(hash: string): Promise<void>;
```
ValentinVPK: We have delete() function here and it is implemented in all of the storages. However these methods are not invoked anywhere and we depend on the TTL. But with TX_STATUS_TRACING_TTL_MS=0 these records are not cleaned and the store grows without a bound.

src/relay/lib/services/transactionTracingService/transactionTracingService.ts
```diff
@@ -0,0 +1,202 @@
...
+        },
+      });
+    }
+
+    if (record.status === 'timedout') {
+      return new JsonRpcError({
+        code: TRANSACTION_REJECTED_CODE,
+        message:
+          'Transaction submission timed out and has not yet reached consensus; it may still be confirmed within the network validity window',
+        data: {
```
ValentinVPK: Why is `hederaStatus` not included here?

src/relay/lib/services/transactionTracingService/transactionTracingService.ts
```diff
@@ -0,0 +1,202 @@
...
+import { ConfigService } from '../../../../config-service/services';
+import { JsonRpcError } from '../../errors/JsonRpcError';
+import {
+  type ITransactionTracingService,
+  type ITransactionTracingStorage,
+  type TransactionTraceRecord,
+} from '../../types/transactionTracing';
+
+/** `-32003` / TRANSACTION_REJECTED, matching the code returned for consensus-time rejections. */
+const TRANSACTION_REJECTED_CODE = -32003;
```
ValentinVPK: Nit: we have the predefined object from JsonRpcError that exports this code, we can import it from there.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -945,6 +1020,10 @@ export class TransactionService implements ITransactionService {
     // zombie cleanup: max attempts reached, remove anyway
     // we don't know if the transaction was actually executed or not...
     this.logger.error('Zombie cleanup: transaction %s not found in Mirror Node after %d attempts', txHash, maxAttempts);
+    // Outcome remains unknown after polling exhausted - trace as timedout (provisional).
+    await this.transactionTracingService.recordTimedout(txHash, {
```
ValentinVPK: Do we need this `timedout` call? The transaction here is already send in the beginning of the function:
```
if (!error) {
    await this.transactionTracingService.recordSent(...);   // "sent" — the CN took it
  }
```

So the transaction is sent via the CN but if the mirror node is slow and can't fetch in time, we are going to set it as timedout which is not correct. I think this line should be deleted.

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -743,13 +788,19 @@ export class TransactionService implements ITransactionService {
...
       transactionBuffer,
       senderAddress,
       networkGasPriceInWeiBars,
       requestDetails,
     );
 
+    // A clean submission (no error) means the CN accepted the transaction - trace the sent state,
+    // capturing the transaction id so the record is resolvable by id.
+    if (!error) {
+      await this.transactionTracingService.recordSent(parsedTx.hash!, submittedTransactionId || undefined);
```
ValentinVPK: Should this be moved belo the release lock 2 because this lock is an operation that need to be triggered asap?

src/relay/lib/services/ethService/transactionService/TransactionService.ts
```diff
@@ -388,13 +419,23 @@ export class TransactionService implements ITransactionService {
 
       // save transaction to pool
       await this.transactionPoolService.saveTransaction(senderAddress, parsedTx, confirmedCount);
+
+      // Trace the pending state - only meaningful when the tx pool write actually happened.
+      if (ConfigService.get('ENABLE_TX_POOL')) {
+        await this.transactionTracingService.recordPending(parsedTx.hash!);
```
ValentinVPK: Should this operation be moved before the try catch because in the `finally` statement we have release of lock 1 that needs to happen asap?

src/relay/lib/services/transactionTracingService/LocalTransactionTracingStorage.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Delegates to an internal {@link LocalLRUCache} for per-entry TTL and a size bound. A private
+ * {@link Registry} is used for the internal cache so its `rpc_relay_cache` gauge does not clash with
+ * the shared relay cache's gauge on the main registry.
+ *
+ * Records are stored under a dedicated `hash:<hash>` key namespace.
+ */
+export class LocalTransactionTracingStorage implements ITransactionTracingStorage {
+  private static readonly HASH_KEY_PREFIX = 'txtrace:hash:';
+
+  private readonly cache: LocalLRUCache;
```
ValentinVPK, line 23: The local cache here is capped by the CACHE_MAX config which is 1000 by default. That means that more transactions than that get discarded and so the default transactaction status tracing ttl of 15 minutes is not accurate for the local branch. Is this on purpose or do we need to include a separate config for this? If it is on purpose maybe we could document it?

src/relay/lib/services/transactionTracingService/transactionTracingService.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  /**
+   * Returns the raw trace record for a hash, or null. Used by admin inspection.
+   *
+   * @param hash - The transaction's keccak256 hash.
+   * @returns The trace record, or null if none exists (or tracing is off).
+   */
+  async getByHash(hash: string): Promise<TransactionTraceRecord | null> {
+    if (!this.storage) return null;
+    return this.storage.get(this.normalizeHash(hash));
```
ValentinVPK: should we add try catch here?

src/relay/lib/services/transactionTracingService/RedisTransactionTracingStorage.ts
```diff
@@ -0,0 +1,51 @@
...
+import { type ITransactionTracingStorage, type TransactionTraceRecord } from '../../types/transactionTracing';
+
+/**
+ * Redis-backed implementation of {@link ITransactionTracingStorage}.
+ *
+ * Records are stored as JSON under a dedicated `txstatustrace:hash:<hash>` key. TTL is applied via the
+ * `SET ... EX` option (ms converted to seconds); a non-positive TTL persists the record indefinitely.
+ */
+export class RedisTransactionTracingStorage implements ITransactionTracingStorage {
+  private readonly hashKeyPrefix = 'txstatustrace:hash:';
```
ValentinVPK: why are the hash key prefixes different per the local and redis storages? Doesn't it make more sense to be the same?

ValentinVPK (commented):
Great work overall. Some nits and a couple of questions and comments

src/relay/lib/services/transactionTracingService/LocalTransactionTracingStorage.ts
tgntr: introduced a new env var TX_STATUS_TRACING_MAX_ENTRIES that overrides CACHE_MAX for local tracing service

src/relay/lib/types/transactionTracing.ts
tgntr: that's expected behavior for users who decide to set their TTL to 0

natanasow (dismissed):

ValentinVPK (approved):

natanasow (approved):

jasuwienas (approved):

PR #5621, opened 2026-08-10
fix: unify nonce error codes to -32000

### LINKED ISSUE #5609: [BUG] Nonce error codes (NONCE_TOO_LOW / NONCE_TOO_HIGH) use positive JSON-RPC codes (32001 / 32002)
## Description

`predefined.NONCE_TOO_LOW` and `NONCE_TOO_HIGH` are defined with **positive** JSON-RPC error codes (`32001` / `32002`). Ethereum's JSON-RPC error convention (EIP-1474) — which every client keys on — uses **negative** codes (server errors in `-32000…-32099`). A positive `32001` isn't recognized by clients that classify errors by code (viem, web3.js, wallets, custom retry/replacement logic): they fall back to a generic/unknown error instead of a nonce error, so a caller can't branch on "is this a nonce error?" (ethers v6 also matches the message text `"nonce too low"`, so it may be partially resilient.)

## Relationship to #4860

Open issue **#4860** ("Remove WRONG_NONCE Special Handling") targets the same path in the other direction, so please read the two together. Today, on a consensus `WRONG_NONCE`, the `eth_sendRawTransaction` error path (`sendRawTransactionProcessor()`, `TransactionService.ts:810-842`) polls the Mirror Node for the account nonce to decide `NONCE_TOO_LOW` vs `NONCE_TOO_HIGH`; #4860 would delete that poll, so `WRONG_NONCE` surfaces as a generic `TRANSACTION_REJECTED`. These need to be reconciled:

- **#4860's Mirror-Node-load concern is legitimate.** It's an extra round-trip on the `eth_sendRawTransaction` path, incurred exactly when nonce errors are most frequent (`WRONG_NONCE` is high-frequency and concentrated on a few high-throughput accounts), and it couples error handling to Mirror Node availability/latency.
- **But it conflates two separable things:** *(a)* how you **derive** too-low vs too-high (the Mirror Node poll — the load problem), and *(b)* what you **return** (the error's specificity and code). Removing (a) is reasonable; collapsing (b) to a generic `TRANSACTION_REJECTED` discards the too-low/too-high distinction that Ethereum tooling uses to auto-recover (resync-and-retry on too-low; wait/backoff on too-high) — hurting exactly the high-throughput senders who hit `WRONG_NONCE` most.
- **The two goals are compatible** — the distinction can be derived **locally**, without the Mirror Node poll: the precheck path (`precheck.ts:118`) already holds the account nonce it just checked against, and the post-consensus path can compare against the relay's own tracked pending nonce.

**Net:** adopt #4860's performance intent (drop the per-error Mirror Node poll) **and** this issue's correctness (keep specific `NONCE_TOO_LOW`/`NONCE_TOO_HIGH` with correctly-signed codes) — derive the distinction locally. Please don't ship #4860 as a plain "collapse to `TRANSACTION_REJECTED`."

## Steps to Reproduce

1. Send a transaction with a nonce **below** the account's current nonce via `eth_sendRawTransaction`.
2. Observe the JSON-RPC error response:

```json
{ "jsonrpc": "2.0", "id": 1, "error": { "code": 32001, "message": "Nonce too low. Provided nonce: 0, current nonce: 5" } }
```

**Expected:** a **correctly-signed negative** code (per EIP-1474, in `-32000…-32099`) with the nonce-specific message.
**Actual:** `code: 32001` (positive) — and `32002` for the too-high case.

## Root Cause Analysis

`src/relay/lib/errors/JsonRpcError.ts:107-116`:

```js
NONCE_TOO_LOW:  (nonce, currentNonce) => new JsonRpcError({ code: 32001, message: `Nonce too low. …` }),
NONCE_TOO_HIGH: (nonce, currentNonce) => new JsonRpcError({ code: 32002, message: `Nonce too high. …` }),
```

Every other server error in the file is negative (`-32602`, `-32001`, `-32003`, `-32005`, `-32015`, …); the only other positive code is the deliberate `code: 3` (execution-reverted convention). So the nonce codes read like a dropped minus sign — and note `-32001` (the signed form of `32001`) is itself a *different* EIP-1474 error ("resource not found") that the relay uses elsewhere, so `32001` is both wrong-signed and numerically collides with an unrelated code.

Thrown from:
- `src/relay/lib/precheck.ts:118` — pre-submission nonce check.
- `src/relay/lib/services/ethService/transactionService/TransactionService.ts:810-842` (in `sendRawTransactionProcessor()`) — after a consensus `WRONG_NONCE`, the relay polls the Mirror Node for the account nonce (`:829`) and maps it to `NONCE_TOO_LOW/HIGH` (`:836-837`). When that lookup fails or is disabled, it falls through to a generic, correctly-signed `TRANSACTION_REJECTED` (`-32003`, `:839`) — so callers get *either* a non-standard positive nonce code *or* a correctly-signed but non-nonce-specific rejection. (See [Relationship to #4860](#relationship-to-4860).)

## Recommended Fix

Change the two nonce errors to **distinct, unused negative codes** in `-32000…-32099`, keeping the messages/semantics.

**Why this path.** Correctly-signed, nonce-specific codes are what let code-keyed clients (viem, web3.js, wallets, backends) recognize the error and take the *right* recovery action — resync-and-retry on too-low, back off on too-high — which is the entire reason to surface a nonce error instead of a generic failure. It also matches **EIP-1474**, the convention every Ethereum client's error→class map keys on, so the relay stays a drop-in for standard Ethereum tooling. And it's the **minimal, lowest-risk** change: error-formatting only, with no execution/consensus/mirror-node behavior change. It's strictly better than the alternatives — *doing nothing* leaves clients unable to classify the error, while *collapsing to a generic `TRANSACTION_REJECTED`* (#4860's direction) is correctly-signed but discards the too-low/too-high distinction that drives recovery. Fixing the sign **and** keeping the specific errors captures both; deriving the distinction locally also meets #4860's performance goal (see [Relationship to #4860](#relationship-to-4860)).

- **Watch for code reuse** (the space is crowded): `-32001` is used by `RESOURCE_NOT_FOUND` (L151) and `FILTER_NOT_FOUND` (L272); `-32003` by `GAS_LIMIT_TOO_LOW` (L55) and `TRANSACTION_REJECTED` (L317-319). Pick codes not already in use and audit the existing reuse in the same pass.
- **Keep the messages too:** some clients (e.g. ethers) match on the message text (`"nonce too low"`) as well as the code, so retain the clear messages alongside the corrected code.
- **Derive too-low vs too-high locally** (not via the Mirror Node poll) — see [Relationship to #4860](#relationship-to-4860).

**Files to modify:** `src/relay/lib/errors/JsonRpcError.ts` (the two `code:` values) + any tests asserting `32001`/`32002`.

**Test considerations:** assert the negative code on both the precheck path (`precheck.ts`) and the WRONG_NONCE-mapped path (`TransactionService.ts`); add a regression test that no `predefined.*` server error has a positive code (except the `code: 3` revert convention).

## Related

- **#4860** — proposes the opposite change to this path; see [Relationship to #4860](#relationship-to-4860). These should be resolved together.
- Closed symptom reports (never fixed): #1791, #1803.
- Fits the HIP-1383 (Enhanced Error Messages) effort.

## Environment

Relay `main` (observed at commit `24e24d5e`). Reproducible on any network (Solo/testnet/mainnet) — this is error-formatting, independent of the consensus/mirror layer.

ValentinVPK:
### Description

This PR fixes an issue where `NONCE_TOO_LOW` and `NONCE_TOO_HIGH` were defined with **positive** JSON-RPC error codes (`32001` / `32002`). Clients that classify errors by code (viem, web3.js, wallets, custom retry logic) cannot recognize a positive code and fall back to a generic error, so callers can't branch on "is this a nonce error?".

Both errors now use **`-32000`**, matching geth, Besu, Alchemy and QuickNode. The too-low/too-high distinction is carried by the message, which is what ethers and viem already match on. Messages are unchanged.

Also adds a regression test asserting no `predefined.*` error uses a positive code, with `CONTRACT_REVERT` (`code: 3`, the Ethereum execution-reverted convention) as the single documented exception — this is the test class that would have caught the original dropped minus sign.

### Related issue(s)

Fixes #5609

### Testing Guide

1. Send a transaction via `eth_sendRawTransaction` with a nonce **below** the account's current nonce.

   **Before:** `{ "error": { "code": 32001, "message": "Nonce too low. Provided nonce: 0, current nonce: 5" } }`
   **After:** `{ "error": { "code": -32000, "message": "Nonce too low. Provided nonce: 0, current nonce: 5" } }`

2. Send a transaction with a nonce **far above** the current nonce (requires `USE_ASYNC_TX_PROCESSING=false`) and confirm `code: -32000` with `"Nonce too high. …"`.

3. Run the unit tests covering both paths — the precheck path and the post-consensus `WRONG_NONCE` path:

   ```bash
   npx ts-mocha --recursive 'tests/relay/lib/errors/JsonRpcError.spec.ts' 'tests/relay/lib/precheck.spec.ts' --exit
   npx ts-mocha --recursive 'tests/relay/lib/eth/eth_sendRawTransaction.spec.ts' --exit
   ```

### DIFF
--- a/src/relay/lib/errors/JsonRpcError.ts
+++ b/src/relay/lib/errors/JsonRpcError.ts
@@ -104,14 +104,14 @@ export const predefined = {
       code: -32602,
       message: `Missing value for required parameter ${index}`,
     }),
-  NONCE_TOO_LOW: (nonce, currentNonce): JsonRpcError =>
+  NONCE_TOO_LOW: (nonce: number, currentNonce: number): JsonRpcError =>
     new JsonRpcError({
-      code: 32001,
+      code: -32000,
       message: `Nonce too low. Provided nonce: ${nonce}, current nonce: ${currentNonce}`,
     }),
-  NONCE_TOO_HIGH: (nonce, currentNonce): JsonRpcError =>
+  NONCE_TOO_HIGH: (nonce: number, currentNonce: number): JsonRpcError =>
     new JsonRpcError({
-      code: 32002,
+      code: -32000,
       message: `Nonce too high. Provided nonce: ${nonce}, current nonce: ${currentNonce}`,
     }),
   NO_MINING_WORK: new JsonRpcError({
--- a/tests/relay/lib/errors/JsonRpcError.spec.ts
+++ b/tests/relay/lib/errors/JsonRpcError.spec.ts
@@ -221,5 +221,21 @@ describe('Errors', () => {
         });
       });
     });
+
+    it('every predefined error uses a negative code', () => {
+      const positiveCodeExceptions = ['CONTRACT_REVERT'];
+      const dummyArgs: unknown[] = ['1', 1, '1', 1];
+
+      const positiveCodes = Object.entries(predefined)
+        .filter(([name]) => !positiveCodeExceptions.includes(name))
+        .map(([name, entry]) => {
+          const error =
+            typeof entry === 'function' ? (entry as (...args: unknown[]) => JsonRpcError)(...dummyArgs) : entry;
+          return { name, code: error.code };
+        })
+        .filter(({ code }) => code >= 0);
+
+      expect(positiveCodes).to.deep.eq([], `these predefined errors must use negative JSON-RPC codes`);
+    });
   });
 });
--- a/tests/relay/lib/eth/eth_sendRawTransaction.spec.ts
+++ b/tests/relay/lib/eth/eth_sendRawTransaction.spec.ts
@@ -1099,8 +1099,7 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
             .to.be.rejectedWith(JsonRpcError)
             .and.eventually.satisfy(
               (error: JsonRpcError) =>
-                expect(error.code).to.equal(predefined.NONCE_TOO_HIGH(10, 5).code) &&
-                expect(error.message).to.include('Nonce too high'),
+                expect(error.code).to.equal(-32000) && expect(error.message).to.include('Nonce too high'),
             );
 
           // Verify accounts endpoint WAS called to get current nonce
@@ -1131,8 +1130,7 @@ describe('@ethSendRawTransaction eth_sendRawTransaction spec', async function ()
             .to.be.rejectedWith(JsonRpcError)
             .and.eventually.satisfy(
               (error: JsonRpcError) =>
-                expect(error.code).to.equal(predefined.NONCE_TOO_LOW(3, 8).code) &&
-                expect(error.message).to.include('Nonce too low'),
+                expect(error.code).to.equal(-32000) && expect(error.message).to.include('Nonce too low'),
             );
 
           // Verify accounts endpoint WAS called to get current nonce
--- a/tests/relay/lib/precheck.spec.ts
+++ b/tests/relay/lib/precheck.spec.ts
@@ -617,6 +617,7 @@ describe('Precheck', async function () {
         expectedError();
       } catch (e: any) {
         expect(e).to.eql(predefined.NONCE_TOO_LOW(parsedTx.nonce, mirrorAccount.ethereum_nonce));
+        expect(e.code).to.eq(-32000);
       }
     });
 


natanasow (approved):
LGTM

tgntr (approved):

jasuwienas (approved):

Ferparishuertas (approved):

PR #5654, opened 2026-08-19
test: add transactionIndex acceptance coverage for eth_* methods

### LINKED ISSUE #5243: [Track] eth_getBlockByNumber includes child (internal) transactions in block transaction list
## Summary

eth_getBlockByNumber returns child/internal transactions as top-level transactions in the block, violating the Ethereum JSON-RPC spec. Only parent transactions should appear. The transactionIndex field should be calculated based on the final "EVM Block" order of transactions.

## Reproduction

Block 0x593810f (Mainnet block 93552911) returns 5 transactions but should return 3:

```
curl --request POST \
  --url https://mainnet.hashio.io/api/ \
  --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","id":1,"params":["0x593810f",true]}'
```

Expected: 3 transactions - 1 EthereumTransaction, 1 HAPI ContractCall (parent), 1 synthetic crypto transfer

Actual: 5 transactions - the 3 above plus 2 child transactions

## Root Cause

The `/api/v1/contracts/results` endpoint correctly excludes child transactions by default (internal=false). However, the `/api/v1/contracts/results/logs` endpoint returns logs from all transactions, including those emitted by child transactions.

In `blockWorker.ts`, the `populateSyntheticTransactions` function creates a synthetic transaction for every unique `transaction_hash` found in the logs that isn't already in the contract results. Since child transaction hashes are absent from the contract results (filtered by internal=false) but present in the logs, they get synthetic transactions created for them.

## Affected functions

- getBlock() - feeds eth_getBlockByNumber / eth_getBlockByHash
- loadBlockExecutionData() - feeds eth_getBlockReceipts / raw receipt encoding
- getLogsWithParams() - feeds eth_getLogs

## Proposed fix

Add a parallel call to /contracts/results with internal=true to retrieve all contract result hashes (including children). Compute childHashes = allHashes − parentHashes, then pre-filter logs to exclude entries whose transaction_hash is in childHashes before passing to populateSyntheticTransactions.

ValentinVPK:
### Description

This PR adds acceptance coverage for `transactionIndex` across the eight `eth_*` methods affected by #5243.

Blocks containing child (internal) transactions used to list the children and skip the index slots they consumed, a three-transaction block reported `0x0, 0x1, 0x3`. Both defects were fixed in the mirror node (parent-only logs in 0.153, index numbering in 0.160.0), so no relay change is needed; these tests guard against regression.

The two pre-existing index tests compare the relay's `transactionIndex` back to the mirror node's `transaction_index`, which passes whether or not the index is correct. The new tests assert standalone properties: indexes are exactly `0..n-1` with no gaps, all eight methods agree on the same block, every index round-trips to the transaction the block lists there, no child transaction leaks into the block or its logs, and an out-of-range index returns `null`.

Added to `rpc_batch1.spec.ts` next to the existing index tests, reusing its `beforeAll` fixtures.

### Related issue(s)

Fixes #5243

### Testing Guide

1. Start a local node on mirror node ≥ 0.160.0: `MIRROR_IMAGE_TAG=0.161.1 npx @hashgraph/hedera-local start -d`
2. `cp tests/server/localAcceptance.env .env && npm run build`
3. `npm run acceptancetest:api_batch1` — the 9 tests under `Transaction related RPC Calls > transactionIndex` pass.


tgntr (approved):

tests/server/acceptance/rpc_batch1.spec.ts
```diff
@@ -867,6 +867,199 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+          const range =
+            `timestamp=gte:${mirrorBlock.timestamp.from}&timestamp=lte:${mirrorBlock.timestamp.to}` +
+            `&limit=100&order=asc`;
+
+          const parents = (await mirrorNode.get(`/contracts/results?${range}`)).results ?? [];
+          const all = (await mirrorNode.get(`/contracts/results?${range}&internal=true`)).results ?? [];
+          childCount = all.length - parents.length;
+
+          parentHashes = new Set<string>(parents.map((result: any) => result.hash));
+          logHashes = new Set<string>(
```
Ferparishuertas: Is this actually needed? it seems not used. Synthetics should already be included

tests/server/acceptance/rpc_batch1.spec.ts
```diff
@@ -867,6 +867,199 @@ describe('@api-batch-1 RPC Server Acceptance Tests', function () {
...
+              isParent || isSynthetic,
+              `${hash} is neither a parent contract result nor a synthetic log-only transaction, ` +
+                `so it is a child transaction that leaked into block ${mirrorContractDetails.block_number}`,
+            ).to.be.true;
+          }
+
+          assertContiguousIndexes(block, 'eth_getBlockByNumber');
+        });
+
+        it('@release should agree on transactionIndex in "eth_getTransactionByHash" and "eth_getTransactionReceipt"', async function () {
```
Ferparishuertas, line 972: I think we should check for absense of null indexes

tests/server/acceptance/rpc_batch1.spec.ts
ValentinVPK: Yes, removed logHashes extra logs call

tests/server/acceptance/rpc_batch1.spec.ts
ValentinVPK: Added a helper and inserted the check

natanasow (approved):

Ferparishuertas (approved):

PR #5712, opened 2026-09-01
fix: resolve synthetic transaction receipts by recorded consensus timestamp

### LINKED ISSUE #5695: `eth_getTransactionReceipt` returns `null` for synthetic transactions the relay has just listed in a block
## Description

Blockscout indexes a block in two stages: it calls `eth_getBlockByNumber(block, true)`, then requests `eth_getTransactionReceipt` for each transaction hash the block reported. For synthetic transactions — the Ethereum-shaped records that represent native Hedera transactions moving HTS tokens — the first call succeeds and lists the transaction, while the second returns `null` shortly afterwards. Blockscout treats the pair as a contradiction, synthesises `-32602 "Not Found"` internally, and discards the entire block. Every transaction in that block is lost from the explorer, not just the synthetic one.

The two calls resolve a transaction by different means. `eth_getBlockByNumber` queries the Mirror Node by consensus timestamp range. `eth_getTransactionReceipt` queries by transaction hash, and for a synthetic transaction the only viable route is the contract logs endpoint filtered on `transaction.hash` — `/contracts/results/{hash}` never serves synthetic records by design. Immediately after a block is imported there is a window in which the timestamp query already returns the transaction and the hash query does not, so the relay answers the two questions inconsistently.

The relay's existing retry does not help. `getContractResultsLogsWithRetry` fires only when a record is returned but partially populated — a null `transaction_index` or `block_number`, or a `0x` block hash. During this window the response is an empty list rather than an immature record, so no retry is attempted and the relay returns `null` on the first try.

The relay cannot distinguish that empty list from the ordinary answer for a hash that does not exist. An unknown hash and a not-yet-visible hash produce byte-identical responses, and unknown hashes are the common case on this endpoint because wallets poll it for transactions that are not yet mined. Retrying every empty result would therefore add latency to the busiest path in the relay in order to help a rare one.

## Steps to reproduce

1. Index a block containing a synthetic transaction with Blockscout, or issue the two calls manually against a block that has just been imported.
2. Call `eth_getBlockByNumber` with `showDetails = true` and note the synthetic transaction's hash in the `transactions` array.
3. Call `eth_getTransactionReceipt` with that hash within the visibility window.
4. The block call lists the transaction; the receipt call returns `null`.

Outside the window both calls agree, so the failure is only reproducible against a freshly imported block.

## Observed occurrences

Nine synthetic `CRYPTOTRANSFER` transactions taken from Blockscout indexer logs on 2026-08-17, 2026-08-18 and 2026-08-20. All are parent transactions with a single synthetic log each, spread across six payers and three tokens. All nine resolve correctly today, confirming the failure was transient rather than a permanently missing record.

```
0x9bca036bc5d34168f7b308bd4923b628a33b72349939819175a55496101eab02
0xcd7a40ee08d9b86c732d2980cd50361a81f9502fe61dd8e8aebed3a1dccf8ef6
0xb32998b8340a2f7305d5f09ed0d5054050451b5688e3200909761a25f99d530a
0x6d39a10a68cd4477e5fb5c45cde70b017bd414e47c73d4649cb5ddfe54da27f3
0x6b1bf812130f90a0d015685b27adb9e32df0a5abdc01781c7010904d638cf425
0x6e1c4ef77e8fefcbeb623da5212c7b12734f9c24d71ae2c76090e83ef0003aa8
0xe78130fde70feadf2c3ee451b5140e6b09c8b64226903c8f257295fe23fe3bb7
0x697090a651da149c6e7931204c3c3dc728c8fcecf6f51dde091c14b186e9a380
0xc2051cf4d88b3e8a6b9f5521b6ebb42188955fcf864a8968974d7254c2206206
```

Sampling the head of the chain put the rate at roughly one in fifteen thousand transactions, so this is infrequent but recurring — the occurrences above span three separate days.

## Expected behaviour

A transaction the relay has just reported as part of a block should be resolvable by its hash. `eth_getTransactionReceipt` should not return `null` for a hash the relay itself supplied moments earlier.

## Proposed fix

While serving a block, the relay already holds each transaction's consensus timestamp: the block path queries contract results by timestamp range, and that response carries both the hash and the exact `seconds.nanoseconds` timestamp. Record that mapping for the block's synthetic transactions, and when a by-hash receipt lookup would otherwise fail, resolve the transaction by its recorded timestamp instead. The timestamp route is the one that already works during the window, because it reads what was written first.

Consulting the recorded timestamp _before_ the by-hash lookups also removes work rather than adding it: a recorded entry means the transaction is synthetic, so both by-hash routes are known to be unproductive. Three Mirror Node calls become one.

Only synthetic transactions are recorded, identified by a record having no gas limit and no nonce or signature — a combination no valid Ethereum transaction can have. The relay's existing `isChildContractRecord` helper is unsuitable here: it also matches inner calls of real transactions, which resolve by hash perfectly well, and building a synthetic-shaped receipt for one of those would return wrong data. Sampling 200 contract results across mainnet and testnet, each fetched individually by hash, the stricter condition produced no false positives and no false negatives, while the child-record helper would have wrongly matched 30 inner calls.

Two safeguards keep the new path from making anything worse than it is today. The logs returned for a recorded timestamp are checked to belong to the requested hash before a receipt is built, and anything unexpected — no entry, no logs, a mismatch, or a failed query — falls through to the existing behaviour rather than failing the request.

### Notes on the implementation

The write happens on the main thread rather than inside the block worker, and the worker returns the hash-to-timestamp pairs as plain data. Worker threads cannot share an in-memory store with the thread that serves receipt requests, so writing inside the worker would leave the entries unreachable whenever `WORKERS_POOL_ENABLED` is on. Returning them as data works for both settings of that flag.

The index is held on `MirrorNodeClient`, which both the block path and the receipt path already receive from the composition root, so no new collaborator has to be threaded through the service graph. It holds its own bounded store rather than using the shared relay cache, whose default bound of 1000 entries would be exhausted by a busy block and would evict live block, balance and gas price data.

### Configuration

| Variable                         | Default  | Purpose                                                                                                |
| -------------------------------- | -------- | ------------------------------------------------------------------------------------------------------ |
| `TX_TIMESTAMP_INDEX_ENABLED`     | `true`   | Switches the behaviour off without a release if it misbehaves                                          |
| `TX_TIMESTAMP_INDEX_MAX_ENTRIES` | `10000`  | Bound on the in-memory index, independent of `CACHE_MAX`                                               |
| `TX_TIMESTAMP_INDEX_TTL_MS`      | `300000` | An entry only needs to outlive the gap between a block request and the receipt requests that follow it |

## MN Tracking

The underlying visibility window is tracked on the Mirror Node side as hiero-ledger/hiero-mirror-node#14133. The change proposed here is a relay-side mitigation and remains useful regardless of that issue's outcome, since it also removes two Mirror Node round trips from every synthetic receipt lookup.

ValentinVPK:
### Description

This PR fixes an issue where `eth_getTransactionReceipt` returns `null` for a synthetic transaction that the relay has just listed in a block, causing Blockscout to discard the whole block.

The two calls resolve a transaction by different means. `eth_getBlockByNumber` queries the Mirror Node by consensus timestamp range; `eth_getTransactionReceipt` queries by transaction hash, and for a synthetic transaction the only viable route is the contract logs endpoint filtered on `transaction.hash`, since `/contracts/results/{hash}` never serves synthetic records by design. Shortly after a block is imported there is a window in which the timestamp query already returns the transaction and the hash query does not, so the relay answers the two questions inconsistently. Blockscout treats that as a contradiction, synthesises `-32602 "Not Found"` internally, and drops every transaction in the block, not just the synthetic one. The existing retry does not help: it fires only on a partially populated record, and during this window the response is an empty list.

While serving a block the relay already holds each transaction's consensus timestamp, so this PR records that mapping for the block's synthetic transactions and consults it at the start of `eth_getTransactionReceipt`. A recorded entry means the transaction is synthetic, so the `/contracts/results/{hash}` leg is known to be unproductive and resolving the transaction drops from two Mirror Node calls to one. Anything unexpected — no entry, no logs, a hash mismatch, or a failed query — falls through to the existing path, so a request can never end up worse than it is today.

Synthetic records are identified by having no gas limit and no nonce or signature. This is confirmed in the Mirror Node source: `SyntheticContractResultServiceImpl` sets `contractResult.setGasLimit(0L)` and only creates a result when the record item is not a contract. The signature fields are load-bearing rather than redundant, since an ethereum transaction submitted with `gasLimit: 0` would otherwise match. The relay's existing `isChildContractRecord` helper is unsuitable here: it also matches HAPI-submitted `CONTRACTCALL` records, which carry no ethereum signature either but did execute and resolve by hash.

The block worker returns the hash-to-timestamp pairs as plain data and the recording happens on the main thread, so the behaviour is the same whether `WORKERS_POOL_ENABLED` is on or off. The index is hosted on `MirrorNodeClient`, which both the block path and the receipt path already receive, so no new collaborator is threaded through the service graph. It holds its own bounded store rather than using the shared relay cache, whose default bound of 1000 entries would be exhausted by a busy block and would evict live block, balance and gas price data.

New configuration, all documented in `docs/configuration.md`:

| Variable                         | Default  | Purpose                                                                                              |
| -------------------------------- | -------- | ---------------------------------------------------------------------------------------------------- |
| `TX_TIMESTAMP_INDEX_ENABLED`     | `true`   | Switches the behaviour off without a release                                                         |
| `TX_TIMESTAMP_INDEX_MAX_ENTRIES` | `10000`  | Bound on the in-memory index, independent of `CACHE_MAX`                                             |
| `TX_TIMESTAMP_INDEX_TTL_MS`      | `300000` | An entry only has to outlive the gap between a block request and the receipt requests that follow it |

Known limitations, both in the configuration docs: a by-hash request with no preceding block request has nothing recorded and behaves exactly as today, and entries are held per relay instance unless `REDIS_ENABLED=true`, so a multi-replica deployment without Redis may serve a block and its receipts from different instances.

The underlying visibility window is tracked upstream as hiero-ledger/hiero-mirror-node#14133. This change is a relay-side mitigation and remains useful regardless of that issue's outcome, since it also removes a Mirror Node round trip from every synthetic receipt lookup.

### Related issue(s)

Fixes #5695

### Testing Guide

45 unit tests cover the store implementations and factory, the classifier against captured Mirror Node shapes, the write side including the worker-pool-enabled path, and the read side including every fall-through case. To verify manually against mainnet:

1. Set `MIRROR_NODE_URL=https://mainnet.mirrornode.hedera.com` and start the relay with `LOG_LEVEL=trace`. No local node or operator key is needed; the read paths only reach the Mirror Node.
2. `POST http://localhost:7546` with `{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x5ee448e",true]}`. Block 99501198 contains one synthetic transaction, `0x5321e1d95502d892c5d911f33e82b4863133d75a2b9f93709bd1970eafda76d9`. The log should show `Caching txtimestamp:hash:0x5321e1d9…` against the `tx-timestamp-index-cache`.
3. `POST http://localhost:7546` with `{"jsonrpc":"2.0","id":1,"method":"eth_getTransactionReceipt","params":["0x5321e1d95502d892c5d911f33e82b4863133d75a2b9f93709bd1970eafda76d9"]}`. Expect a populated receipt with `"blockNumber": "0x5ee448e"`, a `resolved 0x5321e1d9… by recorded consensus timestamp` log line, one `contracts/results/logs?timestamp=eq:` request, and no `contracts/results/0x5321…` or `transaction.hash=` request.
4. Restart the relay to empty the index and repeat step 3 alone. The receipt should be identical, but the log now shows the `contracts/results/{hash}` 404 followed by the `transaction.hash=` query, confirming the previous behaviour is intact on a miss.
5. Repeat steps 2 and 3 with `WORKERS_POOL_ENABLED=true`. The behaviour should be unchanged: the worker's own log lines appear unformatted, and the `tx-timestamp-index-cache` write still appears on the main thread.


src/relay/lib/services/transactionTimestampIndexService/constants.ts
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: Apache-2.0
+
+/** Key namespace for hash to consensus timestamp entries, shared by the local and Redis implementations. */
+export const TIMESTAMP_INDEX_KEY_PREFIX = 'txtimestamp:hash:';
```
natanasow: I don't think we need a new file for just one constant. We can use the already existing constants file.

src/relay/lib/types/ITransactionTimestampIndex.ts
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
```
natanasow, line 1: We should rename the file to `ITransactionTimestampIndex` to keep the conventions.

natanasow (commented):
We should add an acceptance test as well.

src/config-service/services/globalConfig.ts
```diff
@@ -803,6 +803,21 @@ const _CONFIG = {
     required: false,
     defaultValue: 900000,
   },
+  TX_TIMESTAMP_INDEX_ENABLED: {
+    type: 'boolean',
+    required: false,
+    defaultValue: true,
+  },
+  TX_TIMESTAMP_INDEX_MAX_ENTRIES: {
```
Ferparishuertas, line 811: it cannot be 10000
10000 is the TPS per second, and a second can contain 10k crypto transfers
We have seen in blockscout 270 retries until not found is resolved. 2-5 mins. depending on the retries ( 500 ms)

Default value should be at least 1200000

ValentinVPK: So with our current default configurations we have:
- `ETH_GET_TRANSACTION_COUNT_MAX_BLOCK_RANGE` = 1000. A block with higher than that contract results will throw MAX_BLOCK_SIZE and the cache will not be written to
- `MIRROR_NODE_CONTRACT_RESULTS_PG_MAX` (25) * `MIRROR_NODE_LIMIT_PARAM` (100) = 2,500 contract results per block. So at 2s blocks that is 1,250 entries a second. With the 5 minutes TTL 375,000 (we can round it to 400,000)

If we were to assume the highest scenario possible (10K entries a second), over a 5 minute TTL we will have 3,000,000 entries for the cache.

Should we use the 400K value because it is connected to the other default values of the configurations or use the 3 million one as worst case scenario?

ValentinVPK: @Ferparishuertas So 1.2 million cache records is around 500 MB in memory which is high.

natanasow (approved):

Ferparishuertas (approved):

PR #5751, opened 2026-09-10
feat: add redis version compatibility testing

### LINKED ISSUE #5511: No regression test coverage for Redis version compatibility [needs discussion]
## Problem

The codebase has no tests that verify Redis-dependent components work correctly against a real, version-pinned Redis instance.

Redis is a shared dependency across four production subsystems:

- **Cache layer**: `redisCache.ts`, `safeRedisCache.ts`
- **Rate limiter**: `RedisRateLimitStore.ts`
- **Transaction pool**: `RedisPendingTransactionStorage.ts`
- **Lock service**: `RedisLockStrategy.ts`

Existing tests for these components either mock the Redis client or use test doubles. None spin up a real Redis process at a pinned version. This means a Redis version upgrade today has no automated safety net; any regression would only surface in production.

## Action required

Before doing anything, open an ADR and discuss with the team on how to approach testing for Redis version compatibility, particularly what mechanism would let us confidently catch regressions across all relay components before a version upgrade reaches production.

ValentinVPK:
### Description

This PR adds real Redis coverage for the two components that had none, and a way to check the relay against any Redis version on demand.

`RedisRateLimitStore` and `RedisLockStrategy` are the only files in the codebase that run Lua inside Redis, and both of their test suites replaced the client with a `sinon` stub — so that Lua had never been executed by any Redis, at any version, in any test. The existing mocked suites are kept, since they cover the fail-open branches a healthy server will not produce; the new suites run alongside them against a real server.

Also included: a one-line port fix for a suite that intermittently failed the relay test run, a `test:redis` script that runs only the Redis-backed suites, an on-demand **Redis Compatibility** workflow, and documentation for `REDISMS_VERSION`.

**Changes**

- `RedisRateLimitStore` — 9 tests: limit enforcement, per-IP and per-method isolation, window expiry, concurrent atomicity, and that the Lua reply is a number (`incrementAndCheck` compares with `result === 1`, so a string reply would silently disable rate limiting)
- `RedisLockStrategy` — 11 tests: mutual exclusion, handoff to a waiter, release ownership, TTL expiry, and zombie queue pruning
- `redisClientManager.spec.ts` — moved off port 6380, which `hbarSpendingPlanRepository.spec.ts` already used. The two together left an abandoned client reconnecting to the next suite's server, timing out a `before all` hook. Before the fix the full relay run gave 2474 passing / 1 failing; after, 2475 / 0
- `npm run test:redis` — runs the 13 Redis-backed suites (~10s vs ~3m for the full relay suite), discovered by searching for `useInMemoryRedisServer` so new suites are picked up automatically
- `.github/workflows/redis-compat.yml` — `workflow_dispatch` with a version input (blank = newest release). The Redis build is a separate named step from the test run, so a toolchain failure is distinguishable from a real incompatibility
- `docs/testing-guide.md` — when a real Redis is needed, the port list, and how to target a specific version

**Both new suites were mutation-tested.** Dropping `NX` from the lock's `SET`, removing its `PX`, skipping zombie pruning, removing the ownership check from the release Lua, flipping the rate limiter's `>` to `>=`, dropping its `EXPIRE`, and returning a string instead of a number each produced failures. Before this PR every one of those mutations passed the suite.

**Result:** all 356 Redis tests pass against Redis 8.10.1, unchanged from 7.2.15. The August incident that led to the current pin was a build failure — `redis-memory-server` compiles Redis from source and the runner lacks the toolchain Redis 8's bundled modules need — not a behavioural one. Setting `BUILD_ARGS=core` builds the server alone and avoids it.

### Related issue(s)

Fixes #5511

### Testing Guide

1. `npm run test:redis` — 356 passing.
2. `REDISMS_VERSION=8.10.1 BUILD_ARGS=core npm run test:redis` — 356 passing, against Redis 8. First run compiles Redis (~1 min), then caches.
3. Run the **Redis Compatibility** workflow from the Actions tab, leaving the version blank to pick up the newest release.
4. To confirm the new tests have teeth, remove `NX: true` from the lock's `set` in `RedisLockStrategy.ts` and re-run — 3 tests fail.

### DIFF
--- a/tests/relay/lib/services/rateLimiterService/RedisRateLimitStore.spec.ts
+++ b/tests/relay/lib/services/rateLimiterService/RedisRateLimitStore.spec.ts
@@ -298,14 +298,18 @@ describe('RedisRateLimitStore Real Redis Test Suite', function () {
 
   it('should set the expiry on the first request only', async () => {
     const key = uniqueKey();
+    const longWindowStore = new RedisRateLimitStore(redisClient, logger, 5000);
 
-    await store.incrementAndCheck(key, 5);
-    const ttlAfterFirst = await redisClient.ttl(key.toString());
+    await longWindowStore.incrementAndCheck(key, 5);
+    const ttlAfterFirst = await redisClient.pTTL(key.toString());
     expect(ttlAfterFirst).to.be.greaterThan(0);
 
-    await store.incrementAndCheck(key, 5);
-    const ttlAfterSecond = await redisClient.ttl(key.toString());
-    expect(ttlAfterSecond).to.be.at.most(ttlAfterFirst);
+    await new Promise((resolve) => setTimeout(resolve, 100));
+
+    await longWindowStore.incrementAndCheck(key, 5);
+    const ttlAfterSecond = await redisClient.pTTL(key.toString());
+    expect(ttlAfterSecond).to.be.greaterThan(0);
+    expect(ttlAfterSecond).to.be.lessThan(ttlAfterFirst);
   });
 
   it('should allow requests again once the window has elapsed', async () => {
--- a/tests/relay/lib/services/rateLimiterService/RedisRateLimitStore.spec.ts
+++ b/tests/relay/lib/services/rateLimiterService/RedisRateLimitStore.spec.ts
@@ -3,11 +3,12 @@
 import { expect } from 'chai';
 import { type Logger, pino } from 'pino';
 import { Counter, Registry } from 'prom-client';
-import { type RedisClientType } from 'redis';
+import { createClient, type RedisClientType } from 'redis';
 import * as sinon from 'sinon';
 
 import { RedisRateLimitStore } from '../../../../../src/relay/lib/services/rateLimiterService/RedisRateLimitStore';
 import { RateLimitKey } from '../../../../../src/relay/lib/types/rateLimiter';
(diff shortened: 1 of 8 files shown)


tests/relay/lib/services/rateLimiterService/RedisRateLimitStore.spec.ts
```diff
@@ -216,3 +217,132 @@ describe('RedisRateLimitStore Test Suite', function () {
...
+
+    for (let i = 0; i < 2; i++) {
+      await store.incrementAndCheck(chainId, 2);
+    }
+
+    expect(await store.incrementAndCheck(chainId, 2)).to.be.true;
+    expect(await store.incrementAndCheck(blockNumber, 2)).to.be.false;
+  });
+
+  it('should set the expiry on the first request only', async () => {
```
tgntr, line 299: Should we use pTTL with a short sleep and lessThan here? Right now ttl() returns whole seconds against a 1s window, so both reads are 1 — I mutated the Lua to call EXPIRE unconditionally and the test still passed.

ValentinVPK: Good catch. Implemented the solution.

tgntr (approved):

PR #5766, opened 2026-09-16
fix: reject non-primitive JSON-RPC id

### LINKED ISSUE #5692: JSON-RPC request deserializer crashes on non-primitive id types (boolean, array, object) instead of returning -32700 parse error
### Description

The relay's JSON-RPC request deserializer attempts to cast the `id` field into a strongly-typed value without validating its type first. When `id` is a boolean, array, or object, the server throws an unhandled exception and returns **HTTP 500 Internal Server Error** instead of the JSON-RPC 2.0 specification-compliant parse error (`-32700` with HTTP 200).

In severe cases, the exception propagates beyond the application layer and the **load balancer returns HTTP 502**, indicating the crash kills the request handling entirely.

### Evidence

| Payload | HTTP Status | Response | Spec-Compliant? |
|---------|-------------|----------|-----------------|
| `{"jsonrpc":"2.0","id":true,"method":"eth_blockNumber"}` | **500** | `Internal Server Error` | ❌ No |
| `{"jsonrpc":"2.0","id":false,"method":"eth_blockNumber"}` | **500** | `Internal Server Error` | ❌ No |
| `{"jsonrpc":"2.0","id":[1,2,3],"method":"eth_blockNumber"}` | **500** | `Internal Server Error` | ❌ No |
| `{"jsonrpc":"2.0","id":{"a":1},"method":"eth_blockNumber"}` | **500/502** | `Internal Server Error` / `error code: 502` | ❌ No |
| `{"jsonrpc":"2.0","id":1.337,"method":"eth_blockNumber"}` | 200 | Valid result | ⚠️ Non-standard but no crash |
| `{"jsonrpc":"2.0","id":null,"method":"eth_blockNumber"}` | 200 | Valid result | ✅ Yes |
| `{"jsonrpc":"2.0",id":-1,"method":"eth_blockNumber"}` | 200 | Valid result | ✅ Yes |
| `{"jsonrpc":"2.0","id":"test","method":"eth_blockNumber"}` | 200 | Valid result | ✅ Yes |

### Steps to reproduce

# Test 1: boolean true
  -d '{"jsonrpc":"2.0","id":true,"method":"eth_blockNumber","params":[]}'

# Test 2: boolean false
  -d '{"jsonrpc":"2.0","id":false,"method":"eth_blockNumber","params":[]}'

# Test 3: array
  -d '{"jsonrpc":"2.0","id":[1,2,3],"method":"eth_blockNumber","params":[]}'

# Test 4: object
  -d '{"jsonrpc":"2.0","id":{"a":1},"method":"eth_blockNumber","params":[]}'

### Additional context

| Payload       | HTTP Status | Spec-Compliant? |
| ------------- | ----------- | --------------- |
| `id: true`    | **500**     | ❌ No            |
| `id: false`   | **500**     | ❌ No            |
| `id: [1,2,3]` | **500**     | ❌ No            |
| `id: {"a":1}` | **500**     | ❌ No            |
| `id: null`    | 200         | ✅ Yes           |
| `id: -1`      | 200         | ✅ Yes           |
| `id: "test"`  | 200         | ✅ Yes           |

Expected (per JSON-RPC 2.0 spec):

{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error"}}

with HTTP 200.
Actual: HTTP 500 Internal Server Error.

  ### **Impact**

Violates JSON-RPC 2.0 specification (Section 5.1: parse errors must use -32700).
Unhandled exceptions may leak stack traces or consume server resources.
HTTP 500/502 responses create monitoring noise and suggest missing input validation in the request deserializer.
The id field is user-controlled input; lack of type validation is a code-level bug.

### Proposed Fix:

**_Validate the id field type before deserialization. Accept only string, number, or null. Return -32700 with HTTP 200 for all other types._**

### Hedera network

testnet

### Version

hiero-json-rpc-relay

### Operating system

Other

### One-Line Reproduction

```bash
  -d '{"jsonrpc":"2.0","id":true,"method":"eth_blockNumber","params":[]}'

### PR DESCRIPTION
**Pull request by tgntr**
### Description

This PR fixes an issue where a JSON-RPC `id` was only checked for presence, never for type, so a boolean, array or object `id` reached `jsonRespResult`/`jsonRespError`, which throw on it — including from the catch handlers meant to recover. Over HTTP the throw escaped as an unhandled exception and HTTP 500; over WebSocket it became an unhandled promise rejection and the client got no response at all. Both transports now return `-32600 Invalid Request` with `"id": null`, and a bad entry no longer fails the rest of its batch.

### Related issue(s)

Fixes #5692

### Testing Guide

1. `npm run test:server` (302 passing) and `npm run test:ws-server` (154 passing) — 24 new regression cases across both transports; the 14 negative ones were each verified to fail without the corresponding source change.
2. HTTP: `curl -X POST localhost:7546 -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":true,"method":"eth_chainId","params":[]}'` — expect HTTP 400 and `{"jsonrpc":"2.0","error":{"code":-32600,...},"id":null}` instead of HTTP 500 `Internal Server Error`.
3. WS: send `{"jsonrpc":"2.0","id":true,"method":"eth_chainId","params":[]}` over `ws://localhost:8546` — expect a `-32600` response with `"id": null`; previously no reply arrived at all.
4. Repeat both with `id` of `false`, `[1,2,3]` and `{"a":1}`, then confirm `id` of `null`, `-1`, `1.337` and `"test"` still succeed with the id echoed back.

### Checklist

### CODE REVIEW

File: src/ws-server/webSocketServer.ts [hunk 1]
```diff
@@ -227,6 +233,9 @@ export async function initializeWsServer(
 
           // process requests
           const requestPromises = request.map((item) => {
+            if (item?.id !== undefined && !isValidJsonRpcId(item.id)) {
```
ValentinVPK, line 236: Should we add tests for this new check?

src/ws-server/webSocketServer.ts
```diff
@@ -227,6 +233,9 @@ export async function initializeWsServer(
...
+              return jsonRespError(null, spec.InvalidRequest, requestDetails.requestId);
+            }
             if (ConfigService.get('BATCH_REQUESTS_DISALLOWED_METHODS').includes(item.method)) {
               return jsonRespError(
```
ValentinVPK, line 240: Here we are missing one case where `item.id` is missing (`undefined`) then `item.id` passes through the newly implemented guard above and now if the `item.method` is a part of the `BATCH_REQUESTS_DISALLOWED_METHODS` we pass an undefined value for the `item.id`. Can we do this instead?

```
if (ConfigService.get('BATCH_REQUESTS_DISALLOWED_METHODS').includes(item.method)) {
      return jsonRespError(
        isValidJsonRpcId(item.id) ? item.id : null,
        spec.BatchRequestsMethodNotPermitted(item.method),
        requestDetails.requestId,
      );
    }
```

src/ws-server/controllers/jsonRpcController.ts
ValentinVPK: Unrelated to the changes but still worth fixing. By this check `validateJsonRpcRequest` has passed and the `request.id` should not be any falsy value unless `0`, `""` or `null`. But here if the `request.id` is 0 for example we pass `null` to the `jsonRespError` instead of 0. We should remove the `|| null`

ValentinVPK: Unrelated to the changes but still worth fixing. By this check `validateJsonRpcRequest` has passed and the `request.id` should not be any falsy value unless `0`, `""` or `null`. But here if the `request.id` is 0 for example we pass `null` to the `jsonRespError` instead of 0. We should remove the `|| null`

ValentinVPK (commented):
Great work. I found one case that it is not handled and a couple of smaller bugs left.

ValentinVPK (approved):
LGTM

natanasow (approved):

PR #5767, opened 2026-09-17
fix: bound log filter topic positions and sub-topics at validation time

### LINKED ISSUE #5672: fix: bound log filter topic arrays at validation time
## Summary

Bound both dimensions of a log filter's `topics` array in the shared validator: at most four positions, and at most 100 sub-topics within a position. Today neither is checked at validation time.

## Problem / motivation

The `topics` validator confirms every entry is a valid 32-byte hash but bounds neither the number of positions nor the size of a nested position, and none of `eth_getLogs`, `eth_newFilter` or `eth_subscribe("logs")` caps them downstream.

Positions: an EVM log carries at most four topics and the filter is positional, so five or more positions cannot match any log. We discard everything past the fourth position when building the Mirror Node query, which produces results that are not merely wrong but inverted — `[null, null, null, null, "0x…"]` skips the nil positions, assigns the only real constraint to `topic4`, and `prepareLogsParams` never reads it. The query reaches the Mirror Node with no topic filter at all, so a filter that must match nothing returns every log in the range. go-ethereum rejects more than four positions outright.

Sub-topics: `addTopicsToParams` rejects a nested position longer than 100, but it only runs when a filter is read. `eth_newFilter` validates the block range and caches `topics` verbatim, so an arbitrarily large single position is accepted at creation and fails only when the filter is first read, via `eth_getFilterLogs` or `eth_getFilterChanges` — a permanently unusable filter occupying a cache entry for `FILTER_TTL`. `CACHE_MAX` bounds the LRU by entries, not bytes.

Reported in #5666. There is no upstream amplification, since the discarded positions never reach the Mirror Node, but filter state does scale with the input.

## Proposed solution

1. Bound both dimensions in the shared validator, so all three methods are covered in one place and creation agrees with read. Note the validator's test function operates on `param.flat()`: the position bound must apply to `param.length`, or one legal position holding five sub-topics is rejected while four positions pass.
2. Four is a fixed EVM limit rather than a tunable, and 100 matches what `addTopicsToParams` already enforces, so neither should become a config entry. Nothing that works today breaks: the 100 bound already fires on every read path, so the change only moves the rejection from read time to creation time.
3. Tests: four positions accepted, five rejected, a nested position over 100 rejected, and the null-padded five-position case above.

tgntr:
### Description

This PR bounds both dimensions of a log filter's `topics` array in the shared validator — at most four positions, at most 100 sub-topics per position — covering `eth_getLogs`, `eth_newFilter` and `eth_subscribe("logs")` in one place. Positions past the fourth were silently dropped when building the Mirror Node query, so `[null, null, null, null, "0x…"]` reached the Mirror Node unfiltered and returned every log in the range; the 100 sub-topic limit already existed in `addTopicsToParams` but only fired on read, letting `eth_newFilter` cache a permanently dead filter. The four-position limit is a constant, since it's fixed by the EVM. The sub-topic limit is `ETH_GET_LOGS_SUB_TOPICS_LIMIT` (default 100) and must not exceed the Mirror Node's `maxRepeatedQueryParameters`, and nothing that works today breaks — the change only moves rejection from read time to creation time.

### Related issue(s)

Fixes #5672

### Testing Guide

1. `npx ts-mocha --recursive 'tests/relay/lib/validators/validators.spec.ts' --exit` — four positions accepted, five rejected, the null-padded case, and 100 vs 101 sub-topics.
2. `eth_getLogs` with `"topics": [null, null, null, null, "0x790673a87ac19773537b2553e1dc7c451f659e0f75d1b69a706ad42d25cbdb55"]` — expect `-32602`, previously returned every log in the range.
3. `eth_newFilter` with one position holding 101 topics — expect `-32602` at creation, previously threw only on the first `eth_getFilterLogs`.

### Checklist


src/relay/lib/services/ethService/ethCommonService/CommonService.ts
ValentinVPK, line 516: Should we add the following check here before the `for` loop using the new constant implented?

```
if (topics) {
      if (topics.length > constants.LOG_TOPICS_MAX_POSITIONS) {
        throw predefined.INVALID_PARAMETER(
          'topics',
          `Expected at most ${constants.LOG_TOPICS_MAX_POSITIONS} topic positions, got ${topics.length}`,
        );
      }
      for (let i = 0; i < topics.length; i++) {
      .....
```

ValentinVPK (commented):
Great work overall. Just one question

ValentinVPK (dismissed):

BikramBiswas786:
Thank you for addressing the review comment in the second commit.

The added defense-in-depth check in CommonService.addTopicsToParams() is important because it protects callers that might bypass the shared validator.

I would like to clarify the security relevance of the original defect. This was not only an oversized-input or performance issue. Before the fix, an attacker-controlled filter with more than four topic positions could have its extra positions silently discarded while constructing the Mirror Node query.

For example:

[
  null,
  null,
  null,
  null,
  "0x790673a87ac19773537b2553e1dc7c451f659e0f75d1b69a706ad42d25cbdb55"
]

This filter should match no EVM log because EVM logs have at most four topic positions. Before validation was added, the fifth-position constraint could be dropped and the resulting Mirror Node query could become effectively unfiltered, returning logs outside the intended filter semantics.

That is a query-integrity issue caused by attacker-controlled input, not merely a performance concern. Applications relying on the relay to enforce event-filter semantics could process false-positive events.

Could we also add or confirm end-to-end regression coverage for all three externally reachable paths?

1. eth_getLogs rejects the five-position/null-padded filter with -32602 before any Mirror Node request is made.
2. eth_newFilter rejects more than four positions and more than 100 subtopics before creating or caching a filter.
3. eth_subscribe("logs") rejects the same inputs before creating a poller or subscription.
4. A valid four-position filter remains accepted.
5. The null-padded fifth-position case is explicitly retained as a regression test.

The fix correctly prevents the demonstrated query-integrity failure. I am not claiming High or Critical severity, funds loss, or consensus impact. I am asking that the issue be recorded as a confirmed security-relevant input-validation/query-integrity vulnerability rather than only as a performance or resource-hardening issue.
@tgntr @ValentinVPK

tgntr:
> Thank you for addressing the review comment in the second commit.
> 
> The added defense-in-depth check in CommonService.addTopicsToParams() is important because it protects callers that might bypass the shared validator.
> 
> I would like to clarify the security relevance of the original defect. This was not only an oversized-input or performance issue. Before the fix, an attacker-controlled filter with more than four topic positions could have its extra positions silently discarded while constructing the Mirror Node query.
> 
> For example:
> 
> [ null, null, null, null, "0x790673a87ac19773537b2553e1dc7c451f659e0f75d1b69a706ad42d25cbdb55" ]
> 
> This filter should match no EVM log because EVM logs have at most four topic positions. Before validation was added, the fifth-position constraint could be dropped and the resulting Mirror Node query could become effectively unfiltered, returning logs outside the intended filter semantics.
> 
> That is a query-integrity issue caused by attacker-controlled input, not merely a performance concern. Applications relying on the relay to enforce event-filter semantics could process false-positive events.
> 
> Could we also add or confirm end-to-end regression coverage for all three externally reachable paths?
> 
> 1. eth_getLogs rejects the five-position/null-padded filter with -32602 before any Mirror Node request is made.
> 2. eth_newFilter rejects more than four positions and more than 100 subtopics before creating or caching a filter.
> 3. eth_subscribe("logs") rejects the same inputs before creating a poller or subscription.
> 4. A valid four-position filter remains accepted.
> 5. The null-padded fifth-position case is explicitly retained as a regression test.
> 
> The fix correctly prevents the demonstrated query-integrity failure. I am not claiming High or Critical severity, funds loss, or consensus impact. I am asking that the issue be recorded as a confirmed security-relevant input-validation/query-integrity vulnerability rather than only as a performance or resource-hardening issue. @tgntr @ValentinVPK

The mechanism is accurate, but #5672 already documented it as an inverted filter rather than a perf item — it just isn't a security boundary: a caller can only broaden their own query over public logs they could already fetch with `topics: []`, so no privilege is crossed and one caller can't affect another's results. On coverage, (4) and (5) are already in the suite, and (1)–(3) share one validator whose wiring is covered by the existing `FilterObject` and `EthSubscribeLogsParamsObject` tests

src/relay/lib/validators/constants.ts
```diff
@@ -15,6 +15,7 @@ export const BLOCK_HASH_ERROR = `Expected ${HASH_ERROR} of a block`;
 export const TRANSACTION_HASH_ERROR = `Expected ${HASH_ERROR} of a transaction`;
 export const TOPIC_HASH_ERROR = `Expected ${HASH_ERROR} of a topic`;
 export const REWARD_PERCENTILES_ERROR = `Expected an array of up to ${mainConstants.FEE_HISTORY_REWARD_PERCENTILES_MAX_SIZE} numbers, each between 0 and 100`;
+export const TOPICS_ERROR = `Expected an array or array of arrays containing ${HASH_ERROR} of a topic, with at most ${mainConstants.LOG_TOPICS_MAX_POSITIONS} positions and at most ${mainConstants.LOG_TOPICS_MAX_SUB_TOPICS} topics per position`;
```
natanasow, line 18: `containing ${HASH_ERROR}` will resolve into `containing 'Expected 0x prefixed string representing the hash (32 bytes)' of topics` or am I wrong?

ValentinVPK (approved):

PR #5768, opened 2026-09-17
fix: enforce WS_SUBSCRIPTION_LIMIT for logs

### LINKED ISSUE #5679: fix: WS_SUBSCRIPTION_LIMIT is not enforced for eth_subscribe("logs")
## Summary

A client can create more subscriptions than `WS_SUBSCRIPTION_LIMIT` allows on the `logs` path, and separately the counter grows for subscriptions that were never created. Both come from the same split between where the limit is checked and where it is incremented, so moving the accounting into `SubscriptionService.subscribe` fixes both. Reported externally in #5638.

## Problem / motivation

`jsonRpcController` checks the counter before dispatch, while `incrementSubs` runs after `handleEthSubscribe`'s handler returns.

**The limit can be exceeded.** The `logs` branch awaits address validation between the two, so concurrent requests all read the pre-increment value and all pass. A batch makes this deterministic rather than racy, since items resolve through one `Promise.all`: a 20-item batch of unique topic filters creates 20 subscriptions against a limit of 10. Rapid individual frames do the same, as `ws` emits them within one tick. `newHeads` is synchronous through to the increment and caps correctly. Overrun is bounded — roughly 2× per connection, with the per-IP connection cap holding — and needs unique filters, as identical ones dedupe by tag.

**The counter counts subscriptions that do not exist.** `incrementSubs` also runs when `subscribe` returned an existing id, which with `WS_SAME_SUB_FOR_SAME_EVENT` on by default is any repeat subscribe to the same filter. Such a client holds one subscription but is eventually refused at the limit, with nothing to unsubscribe.

Release is sound either way: a failed subscribe throws before the increment, and connection close unsubscribes everything.

## Proposed solution

1. Enforce and increment inside `SubscriptionService.subscribe`, counting only a newly created `Subscriber`, so the check and the mutation are one synchronous step. Keep the pre-dispatch check only as a cheap early reject.
2. Rejected requests still return `predefined.MAX_SUBSCRIPTIONS`.
3. Tests: a 20-item batch of unique topic filters admits exactly the limit; rapid individual frames likewise; repeat subscribes to one filter leave the count at 1; `newHeads` unchanged.

tgntr:
### Description

`jsonRpcController` checked the subscription count before dispatch while `incrementSubs` ran after the handler returned, and the `logs` branch awaits address validation in between — so concurrent requests all read the stale count and all passed (a 20-item batch created 20 subscriptions against a limit of 10), and the counter also grew when `subscribe` returned an existing id, eventually refusing a client that had nothing to unsubscribe. Both now happen in `SubscriptionService.subscribe` as one synchronous step, counting only a newly created `Subscriber`; the pre-dispatch check stays as an early reject and refusals still return `MAX_SUBSCRIPTIONS`. `ConnectionLimiter`'s counter methods now take the `RelayWebSocket` instead of the `WsContext`.

### Related issue(s)

Fixes #5679

### Testing Guide

1. `npm run test:ws-server` — new `tests/ws-server/unit/subscriptionLimit.spec.ts`: a 20-item batch and 20 frames in one tick each admit exactly the limit, repeat subscribes to one filter leave the count at 1, `eth_unsubscribe` frees a slot, `newHeads` still caps.
2. `npm run acceptancetest:ws_batch3` against a local node.

### Checklist


src/ws-server/controllers/jsonRpcController.ts
```diff
@@ -130,7 +130,7 @@ export const getRequestResult = async (
 
   // Check if the subscription limit is exceeded for ETH_SUBSCRIBE method
   let response: IJsonRpcResponse;
-  if (method === WS_CONSTANTS.METHODS.ETH_SUBSCRIBE && !limiter.validateSubscriptionLimit(ctx)) {
+  if (method === WS_CONSTANTS.METHODS.ETH_SUBSCRIBE && !limiter.validateSubscriptionLimit(ctx.websocket)) {
```
ValentinVPK: Can we see if we need this if check at all? We have the new check in the `subscriptionService` that takes care of that. Also if we leave this check in the `jsonRpcController` if somebody that has already subscribed asks for subscribing again, will they get an error if they go over the limit where they should've gotten their old subscribtion?

ValentinVPK (commented):
Looking good. Just one question

ValentinVPK (approved):

natanasow (approved):

PR #5781, opened 2026-09-18
feat: aborted HTTP requests should not continue downstream

### LINKED ISSUE #5676: Aborted HTTP requests continue downstream eth_getLogs fan-out after client disconnect
## Summary

An HTTP client can start an expensive multi-address `eth_getLogs` request and close the connection almost immediately, but the relay continues executing the downstream Mirror Node work after the client has disconnected. The request lifecycle is not connected to the downstream Axios `AbortController` or to the address-level promises created by the log-fetch path.

This is an independent request-lifecycle cancellation issue. It is related to #5664 because the unbounded multi-address fan-out makes the post-disconnect work especially expensive, but it is not the same root cause: #5664 concerns missing concurrency/fairness control, while this report concerns failure to cancel work after the caller has abandoned the HTTP request.

## Observed impact

In a clean localhost reproduction, one `eth_getLogs` request containing 500 addresses was aborted by the client after approximately 50 ms. The relay continued processing the request and generated approximately 500 downstream address-level operations. In a five-request bounded variant, all five client connections were aborted after approximately 50 ms, while the relay continued generating downstream work and reached approximately 300 active upstream requests.

A stronger synchronized run used three waves of ten aborted requests, with 500 addresses per request and a 500-ms deterministic localhost upstream delay. All 30 client requests were abandoned, but the upstream recorded 3,678 requests by the recovery sample and reached a peak of 301 concurrent operations. Twelve unrelated `eth_chainId` controls succeeded in every wave, and twelve recovery controls succeeded after the workload.

The relay therefore continues consuming downstream request capacity after the initiating clients have disappeared. The result is a resource-retention and availability-risk issue, particularly when combined with large accepted fan-out requests.

## Expected behavior

When the HTTP client disconnects or the request is aborted, request-scoped downstream work should be cancelled where possible. Queued address-level operations should not continue indefinitely after the caller has abandoned the request.

## Actual behavior

The relay continues executing the already-created downstream operations after the HTTP client has disconnected. The internal `AbortController` in `MirrorNodeClient.request()` is not connected to the HTTP request lifecycle, and the outer address-level `Promise.all` remains active.

## Severity assessment

The evidence supports a deployment-dependent Medium-severity availability/resource-amplification classification. A Medium-to-High chained-impact argument is possible when this behavior is combined with the unbounded multi-address fan-out documented in #5664, but this report does not claim that the standalone issue caused OOM, permanent outage, production-wide failure, funds loss, consensus impact, or Critical severity.

### Steps to reproduce

1. Check out the tested repository revision:

   `4b9d38b0037c4d4a456cee2723c1ba2f7e95784b`

2. Start a deterministic Mirror Node-style HTTP stub on `127.0.0.1` that records request count and active concurrency, delays each request by 500 ms, and returns valid empty log responses.

3. Start the relay in read-only mode with the Mirror Node URLs pointed only to the localhost stub:

   ```text
   READ_ONLY=true
   REDIS_ENABLED=false
   RPC_HTTP_ENABLED=true
   RPC_WS_ENABLED=false
   MIRROR_NODE_URL=http://127.0.0.1:<stub-port>
   MIRROR_NODE_URL_WEB3=http://127.0.0.1:<stub-port>


4.Confirm that the fresh relay responds to a local eth_chainId request.

5.Construct one JSON-RPC request containing 500 distinct addresses:

{
  "jsonrpc": "2.0",
  "id": 77,
  "method": "eth_getLogs",
  "params": [
    {
      "fromBlock": "0x1",
      "toBlock": "0x2",
      "address": [
        "0x0000000000000000000000000000000000000001",
        "0x0000000000000000000000000000000000000002",
        "...500 total distinct addresses..."
      ],
      "topics": []
    }
  ]
}


6.Send the request to the local relay and abort the client connection after approximately 50 ms. For example, with curl:

curl -sS --max-time 0.05 \
  --data-binary @request.json \
  http://127.0.0.1:<relay-port>


7.Record the expected client-side timeout. The client should receive no completed JSON-RPC response because the connection was intentionally closed.

8.Immediately query the local stub statistics endpoint, then query it again approximately three seconds later. The downstream request count should continue increasing after the client abort.

9.Run an independent local control request:
curl -sS --max-time 20 \
  --data '{"jsonrpc":"2.0","id":99,"method":"eth_chainId","params":[]}' \
  http://127.0.0.1:<relay-port>

10.Repeat steps 5–8 with five concurrent aborted requests, or with three bounded waves of ten aborted requests, while continuing to record upstream active, peak, and total counters.

11.Verify that the relay remains alive and that the local control request eventually returns 0x127. Terminate only the locally started relay and stub processes after collecting the evidence.



### Additional context


# Additional context

```markdown
## Source-level cause

The HTTP JSON-RPC path awaits the relay method but does not pass the HTTP request's `aborted` or `close` lifecycle into a request-scoped cancellation signal. `MirrorNodeClient.request()` creates an internal `AbortController`, but that controller is aborted only after an Axios failure; it is not aborted when the caller closes the HTTP connection.

The multi-address log path then creates one downstream promise per accepted address and awaits the complete collection through an outer `Promise.all`. Consequently, client abandonment does not cancel queued or active address-level requests.

Relevant implementation areas:

- `src/server/koaJsonRpc/index.ts`: HTTP JSON-RPC dispatch and request handling.
- `src/server/koaJsonRpc/lib/utils.ts`: server timeout configuration.
- `src/relay/lib/clients/mirrorNodeClient.ts`: internal `AbortController` and downstream requests.
- `src/relay/lib/services/ethService/ethCommonService/CommonService.ts`: address-level fan-out and result aggregation.

## Verified local evidence

The tests used only localhost processes and a deterministic upstream stub. No mainnet, testnet, public Mirror Node, or production relay was contacted.

- Single aborted request: 500-address request; client aborted at approximately 50 ms; downstream work continued to approximately 508 total upstream requests; peak activity was 300.
- Five repeated aborted requests: 500 addresses per request; downstream total reached approximately 928 after three seconds; peak activity was 301; the relay returned a healthy `eth_chainId` response.
- Synchronized High-gate run: three waves of ten aborted requests; 500 addresses per request; 3,678 upstream requests by the recovery sample; peak activity was 301; all 12 unrelated controls succeeded in each wave; all 12 recovery controls succeeded.

## Security boundary and limitations

This report concerns resource retention and downstream work continuation after HTTP client abandonment. It does not claim that the tested profile caused unrelated-client denial, process crash, OOM, permanent outage, production-wide impact, funds loss, credential exposure, consensus impact, or Critical severity.

The issue is independently relevant because cancellation should be enforced at the HTTP request boundary even when other fan-out limits are present. The impact is greater when combined with #5664, but the missing cancellation propagation is a separate remediation concern.

## Recommended remediation

Create a request-scoped `AbortController` at the HTTP boundary and abort it when the request emits `aborted` or `close`, while distinguishing normal response completion from premature disconnect. Propagate the signal through every Mirror Node Axios request and queued address-level operation. Ensure that fan-out workers stop scheduling new operations after cancellation and that active requests receive the signal. Add regression tests covering one aborted request, repeated aborted requests, concurrent controls, upstream request counts, and post-abort recovery.

### PR DESCRIPTION
**Pull request by natanasow**
### Description

An HTTP client can start an expensive multi-address `eth_getLogs` request and close the connection almost immediately, but the relay continues executing the downstream Mirror Node work after the client has disconnected. The request lifecycle is not connected to the downstream Axios AbortController or to the address-level promises created by the log-fetch path.

### Related issue(s)

Fixes #5676 

### Testing Guide

1.
2.
3.

N/A

N/A

### Checklist

### CODE REVIEW

**Comment by natanasow** [comment]
## User Story: Aborted HTTP Requests Continue Downstream Work After Client Disconnect

## Description

When an HTTP client opens a JSON-RPC request against the relay and closes the connection before the response is written, the relay keeps executing the downstream Mirror Node work for that request. The HTTP request lifecycle is not wired to the `AbortController` used by `MirrorNodeClient.request()`, nor to the address-level promises created by the multi-address `eth_getLogs` fan-out path.

The result is that abandoned requests keep consuming upstream Mirror Node capacity, connection pool slots and relay CPU for the full duration of work that nobody will ever read.

## Current Behavior (Incorrect)

- **HTTP layer:** `KoaJsonRpc` does not observe `ctx.res` `close`. A client disconnect is invisible to the relay.
- **`RequestDetails`:** carries `requestId`, `ipAddress`, `connectionId` - no cancellation channel, so no downstream component can learn that the caller is gone.
- **`MirrorNodeClient.request()`:** creates a local `AbortController` that is only used for its own timeout. It is never linked to the caller, so an in-flight Axios request survives the client disconnect and its retries keep firing.
- **`CommonService.getLogsByAddress()`:** builds one promise per address and awaits `Promise.all`. Once created, every address is fetched regardless of the caller's state.
- **`MirrorNodeClient` sliced log fan-out:** the concurrency pool loop keeps scheduling new slices until every slice has been dispatched.

## Expected Behavior

**Cancel the work. Return nothing.**

- No new Mirror Node request is issued for an abandoned request.
- In-flight Mirror Node requests are aborted and their TCP connections dropped.
- Remaining addresses / slices in a fan-out are never scheduled.
- **Nothing is written back:** no JSON-RPC error object, no HTTP status, no headers. The socket is already gone, so there is nothing to deliver and nothing to invent.
- No new JSON-RPC error code and no new HTTP status code are introduced. The relay's public error surface is unchanged.
- The only trace of an abandoned request is a `debug` log line.

### Fully aligned with geth, Infura and Alchemy

This is exactly how the mainstream Ethereum JSON-RPC providers behave. None of them defines a "client aborted" error code, and none of them answers a caller that has hung up.

| Client | Cancellation mechanism | Sent to the client on disconnect |
|---|---|---|
| **geth** | A `context.Context` derived from the HTTP request. `net/http` cancels it when the client disconnects, and the handler chain unwinds with `context.Canceled`. | **Nothing.** |
| **Infura** | Connection teardown. | **Nothing.** |
| **Alchemy** | Connection teardown. | **Nothing.** |

## Impact

**Severity: MEDIUM**

- **Resource retention:** Mirror Node connection pool slots, sockets and relay CPU are held for work whose caller is gone.
- **Availability risk:** a client that repeatedly issues large `eth_getLogs` requests and hangs up immediately amplifies its own cost by orders of magnitude at near-zero cost to itself.
- **Upstream pressure:** Mirror Node absorbs the full fan-out of every abandoned request, degrading latency for legitimate traffic.
- **Observability noise:** abandoned requests surface as ordinary downstream errors/retries, hiding real failures in logs and metrics.

### Example Impact Scenario

A client sends 10 concurrent `eth_getLogs` requests with 500 addresses each and disconnects after 50 ms. The client pays for 10 aborted sockets; the relay and Mirror Node pay for ~5,000 upstream requests. Repeat the loop and upstream capacity is exhausted by traffic that has no consumer.

## Root Cause Analysis

There is no request-scoped cancellation channel between the HTTP server and the downstream clients. Each layer owns its own abort primitives, and none of them are connected. This is precisely the role `context.Context` plays in geth - the relay is missing the equivalent plumbing, not missing an error code.

## Proposed Solution

Introduce a request-scoped `AbortSignal` - the Node equivalent of geth's request `context.Context` - created by the HTTP layer and propagated through `RequestDetails` to every downstream consumer. Use **native platform cancellation semantics only**; add no relay-specific error type, error code or HTTP status.

### 1. HTTP layer - `src/server/koaJsonRpc/index.ts`

Create one `AbortController` per request. Abort it on `ctx.res` `close` when `writableFinished` is false, and abort it in `finally` once the request has completed, so the listener is released. Pass its signal into `handleSingleRequest` / `handleBatchRequest` and on into `RequestDetails`.

Abort with a native `AbortError` rather than a plain string, so awaiting callers unwind with a real error:

```ts
abortController.abort(requestAbortReason(CLIENT_DISCONNECTED_ABORT_REASON));
```

When the work unwinds because the caller is gone, write nothing at all:

```ts
} catch (error) {
  if (!isRequestAbortedError(error)) {
    throw error;
  }

  // The caller is gone, so there is no connection left to answer on. Write nothing back and let the
  // request unwind silently, the way geth, Infura and Alchemy treat a cancelled request: no JSON-RPC
  // error object and no HTTP status, because neither can be delivered.
  this.logger.debug(`Request abandoned by the client: requestId=%s, method=%s`, requestId, ctx.state.methodName);
  ctx.respond = false;
  ctx.state.status = CLIENT_DISCONNECTED;
  ctx.state.clientDisconnected = true;
}
```

`ctx.respond = false` makes Koa skip response handling entirely - no body is serialised, no status line and no headers are emitted.

Set `ctx.state.methodName` *before* awaiting the result, so an abandoned request is still identifiable in the access log.

### 2. `RequestDetails` - `src/relay/lib/types/RequestDetails.ts`

Add an optional `abortSignal`, defined as a non-enumerable property so `RequestDetails` stays structured-cloneable when dispatched to a worker thread.

### 3. Helper module - `src/relay/lib/utils/requestAbort.ts`

Thin wrappers over the platform primitives - no custom error class, no custom code:

```ts
export function requestAbortReason(message: string): DOMException {
  return new DOMException(message, 'AbortError');
}
export function getRequestAbortSignal(requestDetails?: CancellableRequest): AbortSignal | undefined;
export function isRequestAborted(requestDetails?: CancellableRequest): boolean;
export function throwIfRequestAborted(requestDetails?: CancellableRequest): void {
  getRequestAbortSignal(requestDetails)?.throwIfAborted();
}
export function isRequestAbortedError(error: unknown): boolean;
```

### 4. Downstream enforcement

- `MirrorNodeClient.request()`: bail out before issuing the call if already aborted; link the caller signal to the local timeout controller; let the native abort error propagate; skip Axios retries for cancelled requests (`Axios.isCancel(error) → false` in `retryCondition`).
- `MirrorNodeClient` sliced log fan-out: break out of the concurrency loop when the request is aborted, and check again before merging.
- `CommonService.getLogsByAddress()`: check before building the fan-out and again per address.
- `RpcMethodDispatcher`: log an abandoned request at `debug`, not `error`, and rethrow the cancellation rather than converting it into a `JsonRpcError`.

### 5. Metrics - `src/server/server.ts`

An abandoned request never received a status, so it must not be labelled with one. Koa leaves `ctx.status` at its default when `respond` is false, so recording it would inject a status into the histogram that was never sent:

```ts
if (!ctx.state.clientDisconnected) {
  methodResponseHistogram.labels(ctx.state.methodName, `${ctx.status}`).observe(ms);
}
```

`geth` likewise does not account for abandoned requests; it only logs them.

## HTTP batch requests: handled
A JSON-RPC batch is a single HTTP request carrying many entries, each of which can start its own Mirror Node fan-out. It is the worst case for this bug - one connection, one close event, many independent pieces of downstream work - so it is covered by the same request-scoped signal as a single request, with no separate mechanism.

##  WebSocket server: handled

Cancellation is not HTTP-only. The WebSocket server runs the same relay, the same dispatcher and the same `CommonService.getLogsByAddress` fan-out, so an abandoned `eth_getLogs` over a socket wastes exactly the same upstream capacity as one over HTTP. It is now covered by the same mechanism.

Termination can occur in three ways:

- the client closes the socket during a request
- inactivity TTL - `WS_MAX_INACTIVITY_TTL,` default 300,000 ms - calls `ctx.websocket.terminate()`
- connection limits terminate the socket

## Acceptance Criteria

### AC1: No new downstream work after disconnect
- **Given** a request whose client has already disconnected
- **When** a downstream Mirror Node call is attempted
- **Then** no HTTP request is issued to the Mirror Node
- **And** the call rejects with the native abort error

### AC2: In-flight requests are cancelled
- **Given** a Mirror Node request that is in flight
- **When** the caller aborts
- **Then** the upstream connection is closed before the response is written
- **And** exactly one upstream request was made - no retry after cancellation

### AC3: Fan-out stops early
- **Given** a multi-address `eth_getLogs` request
- **When** the client disconnects partway through the fan-out
- **Then** the number of upstream requests is strictly less than the number of addresses
- **And** the upstream count stops growing once the client is gone

### AC4: Nothing is returned for an abandoned request
- **Given** an abandoned request
- **When** the relay finishes unwinding it
- **Then** no JSON-RPC error object is produced
- **And** no HTTP status, status line or headers are written
- **And** the behavior is identical to geth, Infura and Alchemy: the caller receives no response
- **And** the event is logged at `debug` level, not `error`
- **And** the relay's public error surface is unchanged from before this story

### AC5: No regression for normal requests
- **Given** a request that is never abandoned, with or without an abort signal
- **When** it is executed
- **Then** every address is fetched and the response is identical to current behavior

### AC6: Worker-thread safety
- **Given** a `RequestDetails` carrying an abort signal
- **When** it is structured-cloned to dispatch a task to a worker thread
- **Then** the clone succeeds, `abortSignal` is `undefined` on the clone, and `isRequestAborted` reports `false`

### AC7: Batch requests
- **Given** a JSON-RPC batch request
- **When** the client disconnects
- **Then** the same signal cancels every item in the batch
- **And** no partial batch response is written

### AC8: Server-side timeouts are unaffected
- **Given** a request that is *not* abandoned but exceeds the relay's own Mirror Node timeout
- **When** the timeout fires
- **Then** the existing timeout error and status are returned exactly as today
- **And** it is not misreported as a client abort

### AC9: Metrics stay honest
- **Given** an abandoned request
- **When** the access-log middleware runs
- **Then** no entry is added to the method response histogram
- **And** the access log identifies the method and marks the request as disconnected

### AC10: WebSocket requests carry a cancellation signal
- **Given** a message received on a WebSocket connection
- **When** the relay is handed the request
- **Then** RequestDetails.abortSignal is present and not yet aborted

### AC11: Closing the socket cancels in-flight work

- **Given** a WebSocket request that is still executing
- **When** the client closes the connection
- **Then** its signal is aborted with a native AbortError
- **And** every other message still in flight on that socket is aborted as well
- **And** no further Mirror Node requests are issued for them

### AC12: An abandoned WebSocket request is not answered
- **Given** a WebSocket request abandoned by its client
- **When** it finishes unwinding
- **Then** nothing is sent on the socket
- **And** no unhandled rejection is left behind
- **And** the event is logged at debug, not error

PR #5785, opened 2026-09-18
fix: null contractAddress on receipts for calls that deploy no contract

### LINKED ISSUE #5743: `contractAddress` Field Populated with `to` Address on Reverted Message Calls
## Definition
For a reverted message call (not contract-creation), `eth_getTransactionReceipt` populates `contractAddress` with the call's own `to` address instead of `null`, deviating from the Ethereum JSON-RPC spec.

## Fresh Reproduction (2026-09-09)
```
Garbage calldata sent to 0x167 (guaranteed revert, not a creation call)
Raw receipt: status=0x0, contractAddress=0x0000000000000000000000000000000000000167,
             to=0x0000000000000000000000000000000000000167
contractAddress == to, both non-null, despite this being a message call, not CREATE.
```

## Suggested Fix
Set `contractAddress` to `null` for any receipt where `to` is non-empty, per spec.

### Hedera network

testnet

### Version

hiero-json-rpc-relay

### Operating system

None

tgntr:
### Description

This PR fixes an issue where `eth_getTransactionReceipt` returned the callee's address in `contractAddress` for plain message calls, instead of `null` as the Ethereum JSON-RPC spec requires. The receipt now only carries a `contractAddress` when the transaction actually deployed a contract — determined by the same `created_contract_ids` check the factory already uses to null out `to` — with the HTS system-contract branch kept ahead of it, since token creation through `0x167` is a CALL and never lands in `created_contract_ids`. Also removes the unused duplicate of `getContractAddressFromReceipt` in `CommonService` and its `ICommonService` declaration.

### Related issue(s)

Fixes #5743

### Testing Guide

1. `npm test` — unit coverage in `transactionReceiptFactory.spec.ts` covers a plain message call, a contract created only by the callee, a direct deployment, an HTS token creation, and a reverted HTS creation.
2. Against a running relay, send a plain HBAR transfer and a call to an existing contract, then `eth_getTransactionReceipt` on each — both should return `"contractAddress": null`.
3. Deploy a contract and fetch its receipt — `contractAddress` should be the new contract's address and `to` should be `null`.
4. Create a token via the HTS precompile and fetch its receipt — `contractAddress` should still be the new token's address.

### Checklist


ValentinVPK (commented):
To fix the failing conformity tests we should make a change in the following files:
`tests/server/acceptance/data/conformity/overwrites/eth_getTransactionReceipt/`
- get-legacy-receipt.io
- get-dynamic-fee.io

And change the output (<<) line to:
`"contractAddress":null`

natanasow:
we must add at least one acceptance test

ValentinVPK (approved):

